PR #13, opened 2015-09-03
Side-navigation.js delegate causes all menus to open

pat270:
On the http://liferay.github.io/lexicon/content/header/ page clicking the toggle button causes all sidebars to open due to event delegation. This fix makes the selector on the delegate more specific when using the default toggler class name.

### DIFF
--- a/src/content/header.html
+++ b/src/content/header.html
@@ -162,7 +162,7 @@ section: Components
 									<span>Squaredown</span>
 								</a>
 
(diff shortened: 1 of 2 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:3eb8f47063...bb699e2b23
View total diff: https://github.com/liferay/lexicon/compare/549965c241...bb699e2b23

PR #14, opened 2015-09-03
Added custom open class name to _toggleSimpleSidenav

pat270:
> We are using two sidebars in this moment,  so we need apply differents styles,  not always a “open" class maybe a parameter like:
> $('.sidenav-container').sideNavigation({
>            openedClass: ‘open-productmenu',

This feature only applies to the toggleSimpleSidenav method which is used when using the data attribute api.


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:ef54705567...9ae85f74f9
View total diff: https://github.com/liferay/lexicon/compare/bb699e2b23...9ae85f74f9

PR #15, opened 2015-09-03
List-group-heading combined with collapse js should have same bg colo…

pat270:
…r when focus or hovered

On page http://liferay.github.io/lexicon/content/blogs-appearance/ hovering or focusing List Group Heading causes background color to change. It should stay the same.

### DIFF
--- a/src/scss/atlas-theme/_list-group.scss
+++ b/src/scss/atlas-theme/_list-group.scss
@@ -18,7 +18,8 @@
 	}
 }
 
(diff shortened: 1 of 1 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:58c6809e04...5c760bbaa1
View total diff: https://github.com/liferay/lexicon/compare/9ae85f74f9...5c760bbaa1

PR #17, opened 2015-09-14
Namespace portlet topper

pat270:
Namespaced portlet topper in lexicon so it wouldn't conflict with https://github.com/natecavanaugh/liferay-portal/pull/3484. I'm leaving this in for now as a reference, but should be removed in the future.


natecavanaugh:
Patrick will resend a new pull removing all deprecated components.

PR #18, opened 2015-09-14
Rename cells to table-list-field and table-list-content, also vertica…

pat270:
…lly center table-list-content with top: 50%

http://liferay.github.io/lexicon/content/tables/ implementation requires a fixed width to be set on each table cell, except for the cell that contains the truncated text. This new implementation is more fluid and doesn't require the user to set any width on a cell at the expense of vertically aligning content using the css property vertical-align. @marcoscv-work also requested we change the class name checkbox-cell to something less specific.

### DIFF
--- a/src/content/tables.html
+++ b/src/content/tables.html
@@ -15,29 +15,29 @@ section: Elements
 		<table class="table table-list">
 			<thead>
 				<tr>
-					<th class="checkbox-cell"></th>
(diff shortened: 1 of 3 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:9cb0bee2d0...7a97112842
View total diff: https://github.com/liferay/lexicon/compare/817591eaef...7a97112842

PR #19, opened 2015-09-15
Side-navigation transition ending too early

pat270:
http://liferay.github.io/lexicon/content/blogs-appearance/

Side navigation transition ends too early when search input is focused, added a check in the transition end event listener to make sure it doesn't run unless the transition end event is from side-navigation menu.

### DIFF
--- a/src/js/side-navigation.js
+++ b/src/js/side-navigation.js
@@ -319,10 +319,11 @@
 				}
 			});
 
-			setTimeout(function() {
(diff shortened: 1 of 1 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:7204781087...af83525691
View total diff: https://github.com/liferay/lexicon/compare/7a97112842...af83525691

PR #27, opened 2015-09-23
Add helpers for aspect ratios with background images

pat270:
Added two helper classes for aspect-ratio. Aspect-ratio-bg-cover makes background-image fill the entire space while aspect-ratio-bg-center shows background-image at original size and centers it in its container.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/041a15e141...0e82a8877a

PR #33, opened 2015-09-30
Components should only list components that are in the guidelines pdf

pat270:
I tried to sort as best as I could based on Lexicon - Guidelines - v05.pdf.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/3cafe3f85d...8be5c332be

PR #34, opened 2015-10-01
Side Navigation delegate click should only be instantiated once per t…

pat270:
…oggler class

Side-navigation.js delegate click is instantiated multiple times if there are multiple side-navigations on a page with the same toggler class name.


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:9c198ee345...ff1b9b72c4
View total diff: https://github.com/liferay/lexicon/compare/8be5c332be...ff1b9b72c4

PR #36, opened 2015-10-01
Side-navigation and content should be equal height before opening

pat270:
Sidenav content overflows until the navigation is completely open. It should be equal height before opening. See http://liferay.github.io/lexicon/content/header/.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/2b362b1daf...875370fe44

PR #40, opened 2015-10-06
LPS-59013 hidden-xs class deleted

pat270:
Removes hidden-xs class on left angle bracket from header toolbar. Juan decided to keep it visible in mobile.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/fd1fb54e8b...b8e87c0e87

PR #42, opened 2015-10-12
Form Control Validation shouldn't bleed into nested form-groups

pat270:
A form-group nested inside a form-group (http://liferay.github.io/lexicon/content/form-validation/) with form validation will inherit styles from its parent.

https://issues.liferay.com/browse/LPS-54946

### DIFF
--- a/src/scss/lexicon-base/_form-validation.scss
+++ b/src/scss/lexicon-base/_form-validation.scss
@@ -1,13 +1,67 @@
-.validation-message {
-	.has-error & {
-		color: $state-danger-text;
+// Bootstrap Form Control Validation Reset
+// TWBS won't fix. See https://github.com/twbs/bootstrap/pull/17897
+
+.has-error,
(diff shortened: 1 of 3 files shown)


natecavanaugh:
Heya Patrick,
As we discussed, I'm going to close this one out because I think if the Bootstrap team thinks nested form-groups are invalid, we should treat them as such, and change how we're using them in portal.
So the next step will be to modify the cases in portal where we are using nested form groups (specifically in the .ftl files in /modules/apps/dynamic-data-mapping/dynamic-data-mapping-service/src/com/liferay/dynamic/data/mapping/dependencies/alloy) to not use the `field-wrapper` taglib, but to use a div with a different class (such as `.field-structure-container` or something).
Then we just need to move the styling for the nested fields from applying to the `field-wrapper` to applying to the `field-structure-container` elements.

Thanks Patrick,

PR #43, opened 2015-10-15
Add list-group-card

pat270:
Added component list-group-card to http://liferay.github.io/lexicon/content/list-groups/, I think this functions better than tabular-list-group with checkboxes, since it takes advantage of the :checked pseudo class. We should remove tabular-list-group or change the demo not to include checkboxes.


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:13e6dc8d86...554526cd7a
View total diff: https://github.com/liferay/lexicon/compare/4e5346e28e...554526cd7a

PR #48, opened 2015-10-20
Add background-color to card.

pat270:
http://liferay.github.io/lexicon/content/extending-cards/ have no background-color by default. It doesn't fit well if the body's background-color is another color other than white.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/b9f4eef60b...ce6a7fa989

PR #52, opened 2015-10-26
Atlas control-label class should be styled like label and fix html cl…

pat270:
…osing tag in docs

Hey @natecavanaugh, there are no examples in Lexicon where we use the class ```control-label``` without the label element, but in portal there are cases like ```<span class="control-label"> Radio </span>``` to label a group of radio buttons.


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:fd6d1df0e1...0c027f9ac9
View total diff: https://github.com/liferay/lexicon/compare/c80740ddd0...0c027f9ac9

PR #54, opened 2015-10-26
Move clamp css to its own file and update table example

pat270:
Updated http://liferay.github.io/lexicon/content/tables/ content to use something similar to portal.


natecavanaugh:
Heya Patrick,
Can you resend, there's a rebase error.
Thanks!

PR #56, opened 2015-10-27
Atlas header-toolbar-default divider shouldnt show if there is only o…

pat270:
…ne item

http://liferay.github.io/lexicon/content/header/#header-toolbar-default

Border shouldn't show if there is only one item in the first toolbar-group.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/0c027f9ac9...275a06ab7d

PR #57, opened 2015-10-27
Update headings selector and reorganize sections

pat270:
This update targets only section headings, the only place it doesn't work is http://liferay.github.io/lexicon/content/typography/#headings.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/0053f88580...8f9a0e7bc4

PR #72, opened 2015-11-13
Breaking: Remove tabular-list-group in favor of list-group-card

pat270:
http://liferay.github.io/lexicon/content/list-groups/

This feature removes tabular-list-group and uses list-group-card as the primary way of creating checkable list-group-items. List-group-card uses a checkbox hack to style the list-group-item based on the state of the checkbox (no js required). Clicking links inside the label won't cause the checkbox to be clicked.


natecavanaugh:
Heya Patrick,
Since this one has gone stale, I'm going to close it (sorry that the timing never worked out on merging it).
I think for this one, we'll need to keep it on the roadmap to change for 2.0.

Thanks Patrick,

PR #76, opened 2015-11-25
Breaking: Toggle Switch and Toggle Card should work with SVG Icons

pat270:
http://liferay.github.io/lexicon/content/toggles/

Changed markup/css of Toggle Switch and Card to accomodate SVG Icons
Reorganize demo page


natecavanaugh:
Heya Patrick,
Since this one has gone stale, I'm going to close it (sorry that the timing never worked out on merging it).

If you have a chance, could you take a look at implementing this change in a non-breaking way?

If not, we can keep it on the roadmap to change for 2.0.

Thanks Patrick,

PR #85, opened 2015-12-07
Update: Fixes #81 add support for placing text around toggle-switch

### LINKED ISSUE #81: Add support for placing text inside of the toggle's switch-handle
Right now, if you add text into the `.toggle-switch-handle` element, it will butt up against the toggle switch.
There are cases where someone may prefer to augment the `data-label-*` attributes with persistent text, or just use text with HTML elements inside of it, and forego the `data-label-*` attributes altogether.

The idea Patrick and I came up with was to support this by wrapping the text inside of a span with `toggle-switch-text` as the class name.
That will give it the same appearance as the `data-label-*` attributes.

pat270:
Resending because it was sent it against the wrong branch.

http://liferay.github.io/lexicon/content/toggles/

I ended up putting .toggle-switch-text outside of .toggle-switch-bar. This allows us to place text on every side of the switch.

### DIFF
--- a/src/content/toggles.html
+++ b/src/content/toggles.html
@@ -45,6 +45,8 @@ section: Components
 
 <div class="row">
 	<div class="col-md-12">
+		<h3>Toggle Switch with Data Attributes</h3>
+
 		<blockquote class="blockquote-sm blockquote-success" style="margin-top:15px;">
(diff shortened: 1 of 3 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:e98e6c9c9d...49d4861df8
View total diff: https://github.com/liferay/lexicon/compare/e63db3c224...49d4861df8

PR #90, opened 2015-12-14
Update tooltip sizes

pat270:
http://liferay.github.io/lexicon/content/popovers/

We have decided change the font size/line, because in portal the messages allocated inside tooltips are too long.

Moved $tooltip-inner-max-width to Lexicon Base
Renamed files from tooltips to tooltip to match Bootstraps naming
Added word-break to tooltip so long text wraps
Removed line-height from .tooltip-inner so multiple lines of text is spaced properly in tooltip

### DIFF
--- a/src/scss/atlas-theme/_atlas.scss
+++ b/src/scss/atlas-theme/_atlas.scss
@@ -31,6 +31,6 @@
 @import "toggle-cards";
 @import "toggle-switch";
 @import "toolbars";
-@import "tooltips";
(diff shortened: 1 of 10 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:6199cb25d9...6c2c7f561b
View total diff: https://github.com/liferay/lexicon/compare/c9ff283cb6...6c2c7f561b

PR #98, opened 2015-12-17
Update: Atlas split button dropdowns should be separated by a border

pat270:
http://liferay.github.io/lexicon/content/buttons/#split-button-dropdowns

primary, info, success, warning, danger should have a 1px separator

### DIFF
--- a/src/scss/atlas-theme/_button-groups.scss
+++ b/src/scss/atlas-theme/_button-groups.scss
@@ -8,27 +8,38 @@
 }
 
 .btn-group > .btn + .dropdown-toggle {
-	padding-left: $btn-padding-horizontal / 3;
-	padding-right: $btn-padding-horizontal / 3;
-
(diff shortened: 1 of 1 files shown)


natecavanaugh:
Heya @marcoscv-work,
I'm going to close this pull request out and not merge it right now. I was talking with @pat270 and it seems like this was meant only for the Staging portlet, but we wanted to get verification on that.

Here is what this pull does to all dropdowns if we apply it to everything:
screenshot 2015-12-23 10 44 29

which looks really strange.

Could you verify if this is intentional, or if the Staging split button is somehow different from all other split button dropdowns.

Thanks guys,

marcoscv-work:
Hi guys, yes @natecavanaugh by the moment it will be used only on staging portlet, but we are working on new patterns with this behaviour, it's not the most needed thing but I think we will need it soon.

Thanks guys!

Issue #116, opened 2016-01-20
Sidenav plugin cannot be destroyed and recreated


jbalsas:
The `Sidenav` plugin does not have an API to destroy instances. This is particularly harmful in SPA environments, where the plugin needs to be destroyed upon navigation and instantiated again.

In particular, https://github.com/liferay/lexicon/blob/master/src/js/side-navigation.js#L676 makes new instantiations fail because the reference to the toggler to be destroyed is kept in the document.

pat270:
Fixed at https://github.com/liferay/lexicon/commit/c800d4cf112b0169edf438a9b4cd00d0015eda08

PR #119, opened 2016-01-27
Update: Add destroy method

pat270:
http://liferay.github.io/lexicon/content/sidenav/

You can destroy the plugin using $('#Id').sideNavigation('destroy'). I also made the toggle and close buttons work when using delegate with regular sidenav. If everything is correct, we won't need to destroy sideNavigation plugin in portal and should work as is.


pat270:
#116

natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:b5dc288591...0a56b0b1b9
View total diff: https://github.com/liferay/lexicon/compare/9892d53bda...0a56b0b1b9

PR #120, opened 2016-01-27
Fix: <a class="badge badge-default"> text color should be blue on hover

pat270:
http://liferay.github.io/lexicon/content/badges-and-labels/#badges

`<a class="badge badge-default" href="#1">Badge Default</a>` text color should be blue on hover.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/6912e8e8d3...9892d53bda

PR #127, opened 2016-01-30
Fix: Data url shouldn't place content inside multiple sidebar-body wh…

pat270:
…en nested

This fixes a bug in the document library admin section when the product menu is closed and the browser is refreshed. Data url will load content into the left nav and the nav on the right.


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/7a9c7381de...1ad96ca2d6

PR #130, opened 2016-02-07
Fix issue 129 - We don't need to remove padding in more than 1300 px …

marcoscv-work:
Fix issue 129 - We don't need to remove padding in more than 1300 px width resolutions


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/b2c88637ea...c2898a8316

PR #134, opened 2016-02-17
Fix: Side Navigation lazy loading doesn't try to reload content that …

pat270:
…was never loaded

http://liferay.github.io/lexicon/content/sidenav/#lazy-loading-side-navigation-content

Turn off network 
Click Simple Sidenav 2 Toggler
Close Sidenav
Turn on network
Click Simple Sidenav 2 Toggler
Content should load

### DIFF
--- a/src/js/side-navigation.js
+++ b/src/js/side-navigation.js
@@ -648,11 +648,13 @@
 
 			var urlLoaded = sidenav.data('url-loaded');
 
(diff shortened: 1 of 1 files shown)


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/1fb376ec93...1c810cafea

Issue #132, opened 2016-02-12
modal styles causes document to scroll to top in Chrome


mjbradford89:
This can be reproduced at http://liferay.github.io/lexicon/content/modals/ and also in liferay-portal.  I think this is caused by the styles from the class `modal-open` that is appended to the body.

1. Navigate to http://liferay.github.io/lexicon/content/modals/.
2. Scroll down a bit and click the 'Large Modal' button or the 'Full Screen Modal' button.
3. Close the modal.

Result: The page is scrolled back to the top.

Note: It seems like for the bottom 3 buttons on the Modal demo page, the bootstrap modal component is resetting the scroll position after the modal is closed.

natecavanaugh:
Thanks @mjbradford89 

@pat270 this is being caused by the `modal-open` class being defined here:
https://github.com/liferay/lexicon/blob/master/src/scss/lexicon-base/_modals.scss#L83-L89

I can't remember why that was added. Do you happen to remember?

pat270:
Hey @natecavanaugh and @mjbradford89, those styles were added specifically for ios. Long content under the modal would be scrollable without position fixed because ios doesn't respect overflow hidden on the body. I'll find a work around.

PR #136, opened 2016-02-19
Fix: #132 add work around for ios not respecting overflow hidden on body

pat270:
Hey @natecavanaugh, I couldn't find a css only work around to this. I had to add an exception for ios in Bootstraps modal plugin.

### DIFF
--- a/src/js/modals-ext.js
+++ b/src/js/modals-ext.js
@@ -1,0 +1,106 @@
+(function($) {
+	// https://bugs.webkit.org/show_bug.cgi?id=153852
+
+	var iOS = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
+
+	$.extend($.fn.modal.Constructor.prototype, {
+		hideModal: function () {
+			var that = this;
+			this.$element.hide();
+			this.backdrop(function () {
+				that.$body.removeClass('modal-open');
+
+				// https://bugs.webkit.org/show_bug.cgi?id=153852
+
+				if (iOS && !(that.winScrollPos === 0)) {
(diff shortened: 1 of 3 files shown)


natecavanaugh:
Heya Patrick,
Thanks for working on this. I was doing some testing as well, and we actually always full screen our modals in Lexicon. Does the bug still happen if you don't swipe outside of the modal itself?

Also, I'm a little on the fence about us being responsible for fixing the iOS bug, since it's not all that bad of a bug (unless you know of some really bad side-effects).

So I'm leaning towards 1 of 2 solutions, and you tell me which you think we should go with:

1. Not fix it at all, and tell people that they can use the `position: fixed` hack (and let them decide if it's worth the consequences of having it scroll to the top, or roll their own type of fix).
or 
2. We can also keep the `.modal-open` change, but target iOS 7+ with the following media query:
```
@media \\0 screen {}
```
(which will also target Chrome 22-28, but those are far outside of our needs to support).

I'm okay with either, but I can see some danger in the second solution, since webkit could fix that bug without fixing the other, or vice-versa.
But if there are some negative or destructive downsides to the background scroll, then this is probably the approach we would go with.

Let me know what you think. Thanks Patrick :)

pat270:
Hey Nate,

Once you reach the end (top or bottom) of the modal content, ios will change focus to the body element and scroll that even for full screen modals. It can be a little confusing because you will be scrolling but nothing is happening visually.

I think I like option 1 with offering some solutions. If we do option 2, we will have bug reports for sure e.g. https://issues.liferay.com/browse/LPS-40955. I'll send another pull if thats good with you. Thanks.

PR #143, opened 2016-02-25
Use liferay-gulp-tasks for bringing in maven tasks instead

jbalsas:


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/jbalsas:34bd7e6fae...5201947332
View total diff: https://github.com/liferay/lexicon/compare/55f135bb26...5201947332

natecavanaugh:
Heya Chema,
I had to roll this one out, as it doesn't work for me.
As I mentioned in Slack, I believe it's an NPM 2 vs 3 thing, but what I believe is happening is that since NPM2 is nesting your gulp dependency, you're modifying one that I'm not.

Since npm's resolving mechanism makes it impossible to know for sure which version of gulp it's going to look up in your module vs the consuming module, I think that either passing a copy of the users gulp in should be required (or optional; but I lean towards required, simply to avoid possible mistakes or confusion).

The most common way I've seen this is with something like gulp-help, where you have to do

```js
var gulp = require('gulp');
gulp = require('gulp-help')(gulp);
```

However, I could also see having it as an option on the `registerTasks` method and doing something like:
```js
var gulp = opt_options.gulp || require('gulp');
```

But as I mentioned, I think it should be a requirement, since if the user ever changes their version of gulp (or had a different version already installed), then they'll be left wondering why the tasks aren't there.

What do you think, is this something we could fix in the liferay-gulp-tasks?

natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/jbalsas:34bd7e6fae...ada67dc941
View total diff: https://github.com/liferay/lexicon/compare/55f135bb26...ada67dc941

PR #144, opened 2016-02-26
Update: Collapsible Search should use event delegation

pat270:
http://liferay.github.io/lexicon/content/collapsible-search/

Browser loses reference to the event handler in single page applications, using the data-toggle attribute should delegate the event handler on the document element.

### DIFF
--- a/src/content/collapsible_search.html
+++ b/src/content/collapsible_search.html
@@ -31,28 +31,10 @@ section: Components (JS)
 			</form>
 		</div>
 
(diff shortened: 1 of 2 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:c7c2a5e64f...347b6ed94d
View total diff: https://github.com/liferay/lexicon/compare/ae98f5a290...347b6ed94d

PR #154, opened 2016-03-29
Fix: IE text in card not truncating

pat270:
http://liferay.github.io/lexicon/content/cards/#truncating-text-inside-horizontal-card-without-card-row-layout-fixed

card-clamp-container


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/15b87fd7ef...e7d4412ea3

PR #159, opened 2016-04-14
Breaking: Sidenav should instantiate on toggler for both apis

pat270:
http://liferay.github.io/lexicon/content/header/
http://liferay.github.io/lexicon/content/sidenav/
http://liferay.github.io/lexicon/content/documents-and-media/

Hey @natecavanaugh, this is a big one and will break things in portal. We will need to change the side navigation selector in portal.

### DIFF
--- a/src/content/documents_and_media.html
+++ b/src/content/documents_and_media.html
@@ -755,21 +755,19 @@ section: Example Pages
 		$('body').toggleClass('fixed');
 	});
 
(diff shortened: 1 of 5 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:4560a77fc0...960465e4c6
View total diff: https://github.com/liferay/lexicon/compare/3b5fbc07af...960465e4c6

PR #165, opened 2016-04-26
Remove card header footer section

pat270:
http://liferay.github.io/lexicon/content/cards/
http://liferay.github.io/lexicon/content/extending-cards/

Breaking: Remove card-col-# (not used in portal)
Breaking: Remove card-header, card-footer (not a big issue in portal)
Breaking: Replace card-section with card-block (not used in portal)
Breaking: Remove card focus styles (should be added in portal, cards dont share the same focus styles)
Site: Update docs and examples

### DIFF
--- a/src/content/cards.html
+++ b/src/content/cards.html
@@ -5,18 +5,24 @@ section: Components
 
 <div class="row row-spacing uxgl-vertical-card">
 	<div class="col-md-12">
-		<h3>Default Card</h3>
+		<h3>Card</h3>
 
 		<blockquote class="blockquote-sm blockquote-success">
-			<p>Display card content vertically.</p>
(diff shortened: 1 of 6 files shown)


natecavanaugh:
Heya @pat270 
Thanks for sending this over. I did have another question, which was how much this overlaps/conflicts with the cards in Bootstrap 4?

I don't want us to have to do too much refactoring now that we'll need to repeat later. Do you know how similar/different our cards our?

pat270:
Hey @natecavanaugh, only a couple classes overlap with bs4. I think card and card-block are the only classes that overlap. The components that we have can be used on top of bs4 without conflicts.

natecavanaugh:

PR #185, opened 2016-06-06
Npm publish

Robert-Frampton:
Hey @natecavanaugh 

Here is the release:npm task. I haven't actually run it all the way through since I don't have the right permissions (and because I didn't want to accidentally publish anything).

I left a lot of the logic that made it easier to publish multiple packages for when we eventually split it up in the future.

Also right now `includePaths` just returns the path to the src/scss directory. Not sure what other paths you think we should include there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ release
 lexicon*.zip
 bower_components
 maven-dist
+temp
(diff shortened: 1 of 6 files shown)


natecavanaugh:

PR #186, opened 2016-06-20
Update icon monospaced markup

pat270:
http://liferay.github.io/lexicon/content/icons/#monospaced-icons


natecavanaugh:
Heya Patrick,
Actually, I was wondering if it would be possible to support both cases (where `icon-monospaced` is wrapped around the item, as well as on the same item).

The reason I don't want to push this in as is, is because by breaking the API like this, we'd have to bump the version to 2.0, and that won't get into master.

Is it possible to leverage both uses?

pat270:
Hey Nate, I'll take a look and resend.

PR #188, opened 2016-06-22
Update: icon-monospaced should work on an icon if it is a direct parent

pat270:
http://liferay.github.io/lexicon/content/management-bar/#management-bar-default


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/2a6e09aed7...23cbc22780

PR #191, opened 2016-07-18
Fix: unclosed line

deyaeddin:
You have unclosed line 31 in file 
lexicon/src/scss/lexicon-base/variables/_management-bar.scss


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/930f571c1b...4f23e9af1f

PR #192, opened 2016-08-08
Update: Scrollbar in IE covers content

pat270:
IE should use classic scrollbar type


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/96631e169f...04a04f8a2a

PR #195, opened 2016-08-12
Breaking: Update imports

pat270:
Hey @natecavanaugh, I moved all the mixins to lexicon-base. Importing only atlas mixins should import all the mixins. There will need to be some changes to the imports partials in all the themes so we aren't importing mixins twice. I can do that once this is good.

I also changed the order of the Bourbon import so that Bootstrap's mixins can't overwrite Bourbon's mixins.

I added Lexicon Base variables to the variables files lexicon-base-variables.scss and atlas-variables.scss.


natecavanaugh:
Heya Patrick,
This is pretty close, but can you undo all of the changes related to variables?
This shouldn't be a breaking change (since it should be 100% compatible with the previous usage).
But it seems that this will break because of the Bootstrap variable ordering.

Can you send over just these changes for mixins?

Thanks Patrick,

PR #202, opened 2016-08-17
Update help icon

pat270:
http://liferay.github.io/lexicon/content/icons/#help-icon
http://liferay.github.io/lexicon/content/form-elements-input-groups/#input-group-default


natecavanaugh:
Heya Patrick,
I'm sending this one back, because I don't think you tested it in the final state.
Sass doesn't allow conditional imports, so the import inside of the function call fails.
You can try to build and it will also throw an error.

I think you can wrap the function definitions with a check to make sure it doesn't overwrite any existing ones, which is what we'd need to do inside of the mixins file.

Thanks Patrick,

PR #204, opened 2016-08-18
Breaking: Remove bourbon dependency

pat270:
Hey @natecavanaugh, I left the -ms-prefix on the transform properties that aren't inside animations so it will still work in ie9. It will be one less headache if someone chooses to make Lexicon work with ie9.

I was also able to remove the ie10 and ie11 flexbox hacks by not using Bourbon's flexbox mixins.

### DIFF
--- a/src/scss/_site.scss
+++ b/src/scss/_site.scss
@@ -75,9 +75,8 @@ section.row {
 	position: absolute;
 	right: 15px;
 	top: 50%;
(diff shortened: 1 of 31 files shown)


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/ee860a77a5...0965c26d43

PR #206, opened 2016-08-23
Fix: Word-wrap should work on card-col-content

pat270:
http://liferay.github.io/lexicon/content/cards/#horizontal-cards-with-fixed-width-content


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/0639ec0695...28fe38a361

PR #207, opened 2016-08-24
Update: Use variable -default-text to set panel-defautl heading and f…

pat270:
…ooter text
http://liferay.github.io/lexicon/content/panels/#panel-states


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/28fe38a361...3efb9a3049

PR #213, opened 2016-09-08
Update: Atlas table heading font weight and font size should only app…

pat270:
…ly to bootstrap tables

http://liferay.github.io/lexicon/content/tables/


natecavanaugh:
View total diff: https://github.com/liferay/lexicon/compare/76dbf7288c...1d34f140d6

PR #226, opened 2016-09-19
Update: $color should be added on lexicon-icon-outline and width and height are needed for ie

pat270:
Hey @natecavanaugh, I was getting `Bus error: 10` when trying to use certain svg icons with the svg-url method. It was happening when we replace spaces with %20. I ended up removing that and see no issues in any of the browsers.

I also added width and height to all the svg icons so it will work with ie and background-images.


natecavanaugh:
Heya Patrick,
I'm kicking this one back, as it seems like the inline attributes in the file is breaking in Chrome, and not allowing the background-size to be set.
Can you test and resend with any changes?

Thanks Patrick!

PR #229, opened 2016-09-21
Fix: Collapsible search should transition for browsers that support c…

pat270:
…ss transitions

Hey @natecavanaugh, $.support.transition is undefined when the collapsible search plugin first runs. I wasn't sure how to get around that other than using a timeout. I decided to go back to using $.support.transition directly rather than storing it in a variable.

### DIFF
--- a/src/js/collapsible-search.js
+++ b/src/js/collapsible-search.js
@@ -1,6 +1,4 @@
 +function($) {
-	var SUPPORTS_TRANSITIONS = $.support.transition;
-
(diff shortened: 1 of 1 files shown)


natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:629b63f185...d8b784cbed
View total diff: https://github.com/liferay/lexicon/compare/83e7f2d4f8...d8b784cbed

PR #232, opened 2016-09-27
Update svg scss map

pat270:
Hey @natecavanaugh, I investigated the background-size: 100%; issue. The browser sets the background width to 100% and height to auto. If you use background-size: auto 100% or background-size: contain, it sizes correctly.

I was documenting the Sass SVG functions and ran into an issue where str-replace() would error on large SVG's using double quotes. I removed double quote escaping from svg-url() so we can toss in custom SVGs as background images.


natecavanaugh:
@pat270 I'm about to merge this in, but wanted to give you a heads up that I renamed the `get-lx-svg-icon` to being just `lx-icon`, since we're of course "getting" the icon, and whether it's svg or not isn't important, so I think this would be better to reuse and it's shorter.

Let me know if you have any issues with this though.

Thanks Patrick,

natecavanaugh:

View just my changes: https://github.com/liferay/lexicon/compare/pat270:1eada18bd3...457464d7c2
View total diff: https://github.com/liferay/lexicon/compare/c7cbe7f545...457464d7c2

PR #239, opened 2016-10-05
Fix: Management Bar height should include borders

pat270:
http://liferay.github.io/lexicon/content/management-bar/


natecavanaugh:
merge

PR #240, opened 2016-10-06
Optimize nameplates

pat270:
http://liferay.github.io/lexicon/content/nameplates/

Hey @natecavanaugh, I replaced the selectors .nameplate-content h# and .nameplate-content small with .nameplate-heading and .nameplate-subheading, this saves several selectors. Those selectors aren't used in portal.

I also moved the behavior where nameplate-labels would expand to fill the width of its container in mobile to its own class. I could see some not wanting that behavior by default.


natecavanaugh:
Heya

pat270:
Hey @natecavanaugh, Your comment got cut off, can you paste it again?

PR #244, opened 2016-10-13
Update: Undo border styles when using nav-tabs in navbar

pat270:
Hey @natecavanaugh, We have an instance in portal where we are putting nav-tabs inside a navbar (edit forms sidebar), throwing off the navbar styles. I thought it's better to add a reset in Lexicon since there could be other places we do that.


natecavanaugh:
merge

PR #249, opened 2016-10-14
Fix: Checkbox and radio alignment in management bar

pat270:
http://liferay.github.io/lexicon/content/management-bar/


natecavanaugh:
merge

PR #257, opened 2016-10-27
Update: Optimize Modals for better IE9 support

pat270:
http://liferay.github.io/lexicon/content/modals/
http://liferay.github.io/lexicon/content/test-assorted/


natecavanaugh:
merge

PR #268, opened 2016-11-07
Fix: Chrome rendering issue with responsive-tables and td position relative

pat270:
Hey @natecavanaugh, I think there was a Chrome update that is causing this issue. It didn't appear in any of the previous versions of Lexicon and now it appears in them.

chrome-bug


natecavanaugh:
merge

PR #271, opened 2016-11-07
Fix: Really long text inside fieldsets should break to new line

pat270:
http://liferay.github.io/lexicon/content/form-elements/#disabled-fieldset

Hey @natecavanaugh, Really long text inside fieldsets should break to a new line. This is a workaround for a FF browser bug. I visited this a while ago in FF Developer Edition and noticed it was behaving like webkit, but now FF and developer edition are behaving the same.


natecavanaugh:
merge

PR #280, opened 2016-11-10
Update: Optimize Tooltip for better IE9 support and deprecate `$toolt…

pat270:
…ip-inner-max-width`

http://liferay.github.io/lexicon/content/popovers/


natecavanaugh:
merge

Issue #290, opened 2016-11-24
In Atlas theme, table-list class is not compatible with table-striped and table-hover classes


marcoscv-work:
Only in Atlas theme.

table-striped and table-hover provides striped lines and hover effect, those styles are lost when these classes are combined with table-striped.

PR #293, opened 2016-11-29
Fix: #290 table-striped and table-hover should work with table-list

pat270:
http://liferay.github.io/lexicon/content/tables/#table-list-striped


natecavanaugh:
merge

PR #294, opened 2016-12-02
Update: Add clearfix to sidebar-header

pat270:
http://liferay.github.io/lexicon/content/sidebar/#sidebar-default

Hey Nate, I added clearfix to sidebar-header to cover a case where there is no title just actions.


natecavanaugh:
merge

PR #301, opened 2016-12-14
Update: Badges check for atlas-theme instead of badge-default-bg sinc…

pat270:
…e there are other variables in the block


natecavanaugh:
merge

PR #308, opened 2017-01-30
New: Move label styles to vars, Input[disabled] styles to vars, Input[readonly] styles to vars

pat270:
Prepping for v2 styles, Sass doesn't output the style if the value is null.

http://liferay.github.io/lexicon/content/form-elements/#disabled-inputs
http://liferay.github.io/lexicon/content/form-elements/#read-only-states


natecavanaugh:
merge

PR #309, opened 2017-01-31
Update: Remove input label disabled opacity and rename input-disabled variables

pat270:
http://liferay.github.io/lexicon/content/form-elements/#disabled-inputs

Hey @natecavanaugh, I decided to rename some variables I added two commits ago to match Bootstrap's naming pattern since they are keeping it in BS4. I also removed a variable I added a commit ago? Opacity on the label cascades to the input when input is nested in label, two different values cause styling issues.


natecavanaugh:
merge

PR #310, opened 2017-01-31
Update: Atlas select icon color should use select icon color variables and add $input-textarea-height

pat270:


natecavanaugh:
merge

PR #319, opened 2017-02-15
New: Added option to customize sticker border radius and remove null value check

pat270:
http://liferay.github.io/lexicon/content/badges-and-labels/#stickers


natecavanaugh:
merge

Issue #331, opened 2017-02-28
Checkbox inputs don't align with their label when there is a lot of description text below


drewbrokke:
This behavior can be seen in Liferay Portal when reproducing this ticket: https://issues.liferay.com/browse/LPS-70956.

drewbrokke:
Pull request sent: https://github.com/pat270/lexicon/pull/80

PR #332, opened 2017-03-01
Fix: #331 Atlas Checkbox and Radio Label element should have position relative so the input will stay inline with label

pat270:
http://liferay.github.io/lexicon/content/form-elements/#checkboxes-and-radios-no-extra-mark-up


natecavanaugh:
merge

PR #336, opened 2017-05-18
Removes private key from package.json

zenorocha:


Robert-Frampton:
Hey @zenorocha, the private key doesn't actually need to be removed, there is a publish task in this repo that removes it prior to publishing. I'm assuming so that nothing is ever published by accident.

PR #358, opened 2017-08-29
Pager definitely is dropped

marcoscv-work:
I've confirmed with UX guys we could create the same pager with two "btn-outline-secondary" buttons with arrow icons inside. So definitely it can be dropped.


pat270:
Merged, Thanks!

pat270:
Merged, Thanks!

PR #364, opened 2017-08-29
Breadcrumb vertical/stacked is definitely dropped

marcoscv-work:
Confirmed by @victorvalle 

We can drop breadcrumb-vertical, to keep the correct functionality in liferay-portal we add this component to compatibility layer: https://github.com/liferay/clay/pull/363 and it will be changed in vertical-breadcrumb ADT:

image


pat270:
Merged, Thanks!

PR #399, opened 2017-12-19
Loading icon auto-center

marcoscv-work:


marcoscv-work:
In previous loading indicator version, it was centered by default so I think this could prevent bigger markup changes.
buttons change the display: block to block-inline, so all should work fine.

image

jbalsas:
Closing as I assume this has been merged :)

PR #406, opened 2018-01-09
LPS-76859 Added z-index to '.dropdown' to display ellipsis in front of cards in Web Content

jonathanmccann:
Not sure if pulls are supposed to be sent to the repository so please let me know if it needs to go to a specific reviewer.

/cc @fortunatomaldonado

### DIFF
--- a/src/scss/bootstrap/_dropdown.scss
+++ b/src/scss/bootstrap/_dropdown.scss
@@ -2,6 +2,7 @@
 .dropup,
 .dropdown {
   position: relative;
+  z-index: 1;
 }
 
 .dropdown-toggle {


pat270:
Hey @jonathanmccann @fortunatomaldonado ,

This is the right place to send Lexicon CSS/Clay bug fixes, but this issue looks more like a problem with Web Content's implementation of Lexicon CSS components. It should be fixed there.

fortunatomaldonado:
Hello @pat270,

What is happening is the user has the class 'card' in the web content and so this card is blocking the ellipsis from being seen or clicked. Without the card, the ellipsis works just as intended in every other web content. I added the z-index so that it will account for this scenario.
If you believe this is not the correct fix, I will look into the web content implementation of the components.
Thank you.

pat270:
This is fix is correct, just doesn't belong in Lexicon CSS. The only improvement I would suggest is to up the z-index a bit (maybe keep it under 400).

The problem is they are using a custom pattern that isn't provided by Lexicon CSS. They are using the `pull-right` class on the dropdown which uses `float: right;` and didn't provide extra styles to account for the possibility of user created content that might be `position: relative/absolute/fixed`. The markup below (not part of Lexicon CSS) will also hide the ellipsis.

```
<div style="background-color:red;width:100%;height:150px;position:relative;z-index:100;"></div>
```

fortunatomaldonado:
@pat270,
Thank you for the clarification!

PR #411, opened 2018-01-15
Ensure that NODE_ENV variable is set to 'test' on all environments to avoid inconsistencies during Metal's render/attach lifecycle

Robert-Frampton:


jbalsas:
Merging into `develop`! 🎉

PR #442, opened 2018-01-22
Remove web components examples from demos

carloslancha:


jbalsas:
Removing for now because we haven't fully figured out how css should work for web-components and the demos are a bit misleading.

PR #451, opened 2018-01-24
Fix render unnecessary custom-control-label-text on other components in ClayCheckbox

matuzalemsteles:
In use cases like ClayCard when it is not necessary to pass `$label` it is rendered unnecessarily. Only by placing inside the `{if}` conditional so that they are not rendered in these use cases. What do you think?

```soy
{let $spanLabelClasses kind="text"}
	custom-control-label-text
	{if $hideLabel}
		{sp}sr-only
	{/if}
{/let}

<span class="{$spanLabelClasses}">{$label}</span>
```


carloslancha:

carloslancha:
LGTM!

PR #455, opened 2018-01-24
Update some markups at claycss.com

matuzalemsteles:
Just updating some markups with the latest changes made in Clay.

cc @pat270


jbalsas:
Merging this! We definitely need to move over to using components here to avoid having to keep track of the markup changes in both places

Issue #450, opened 2018-01-24
Dropdown menu should support several lists inside


marcoscv-work:
Dropdown Clay component needs to use different context lists in combination with the rest of elements which are not a list.

This is a markup example:

__
```
<div class="dropdown-menu">
	<form>
		<div class="dropdown-section">
			<div class="input-group input-group-sm">
				<div class="input-group-item">
					<input class="form-control input-group-inset input-group-inset-after" type="text">
					<span class="input-group-inset-item input-group-inset-item-after">
						<button class="btn btn-unstyled" type="button">
							S-Icon
						</button>
					</span>
				</div>
			</div>
		</div>

		<ul class="unnamed-class">
			<li class="unnamed-class">
				<ul class="unnamed-class">
					<li class="dropdown-subheader">SUBHEADER</li>

					<li class="unnamed-class">
						<a class="dropdown-item" href="#1">Actions</a>
					</li>

					<li class="dropdown-divider"></li>

					<li class="unnamed-class">
						<a class="dropdown-item" href="#1">Actions</a>
					</li>
				</ul>
			</li>

			<li class="dropdown-divider"></li>

			<li class="unnamed-class">
				<ul class="unnamed-class">
					<li class="dropdown-subheader">SUBHEADER</li>

					<li class="unnamed-class">
						<a class="dropdown-item" href="#1">Actions</a>
					</li>

					<li class="dropdown-divider"></li>

					<li class="unnamed-class">
						<a class="dropdown-item" href="#1">Actions</a>
					</li>
				</ul>
			</li>
		</ul>
	</form>
</div>
```

carloslancha:
Related with #425. Fixed in same pr.

carloslancha:
Label this as `next release`

PR #461, opened 2018-01-25
Use new markup for ClayDropdown  | Fixes #425 and #450

### LINKED ISSUE #425: Dropdown menu with lists markup is not looking correctly
Dropdown menu markup should be compatible with lists markup to improve semantics. We are currently using <ul> lists in dropdown-menu, we tried to use "list-unstyled" to reset the ul native styles but it does not look like the best option.

Examples in components:
clay-dropdown/demos/
clay-card/demos/

image

image

carloslancha:


matuzalemsteles:

packages/clay-dropdown/src/ClayDropdownBase.soy
```diff
@@ -391,43 +367,78 @@
...
-			<li class="dropdown-subheader">
-				{$label}
-			</li>
-		{else}
-			<div class="dropdown-subheader">
-				{$label}
-			</div>
-		{/if}
-	{/if}
+	<div class="dropdown-subheader" role="presentation">
```
matuzalemsteles: hey @carloslancha, In cases of using `groups` the `$label` is **required** in soy but no error is charged, this ends up printing and leaving empty spaces.

screen shot 2018-01-25 at 11 29 15

carloslancha: Hey @matuzalemsteles we can't throw any errors 'cause is a deltemplate. I'm adding a conditional anyway. Thx!

PR #482, opened 2018-01-26
Fix the align of the dropdown when expanded in ClayDropdown | Fixes #452

### LINKED ISSUE #452: ClayDropdown is not being aligned in the best way
ClayDropdown is not being aligned correctly, just aligning in a single position in all use cases by ignoring the boundaries of the screen.

screen shot 2018-01-24 at 18 09 50

matuzalemsteles:


jbalsas:
Hey @matuzalemsteles, rather than listening the `rendered` on the `portal`,  could you try listening on the `expandedChanged` event in the component?

```javascript
this.on('expandedChanged', this.handleRenderedPortal_.bind(this));
```

matuzalemsteles:
hey @jbalsas, The problem is that I need to know when the `show` class is visible in the DOM, so I can switch to `Align.align`, so it can see **clientWidth** ... See #452 and https://github.com/metal/metal.js/issues/351

jbalsas:
Hey @matuzalemsteles, yeah, I'm here with @Robert-Frampton trying to come up with alternatives. 

We assume you've tried writing this directly in the `rendered` lifecycle of this component and it doesn't work either because the portal might not have been redendered yet?

matuzalemsteles:
Oh, Yeah. When I use `rendered` in` ClayDropdownBase` and make some changes to `expanded` that is used inside the `content` of ClayPortal, `rendered` is called but has not yet been reflected in the DOM which is kind of weird that a bad perception. But by listening to the `rendered` component of the ClayPortal component I can tell when the` content` was actually rendered in the DOM.

jbalsas:
Sounds good then, merging!

Issue #459, opened 2018-01-25
Charts design?


bryceosterhaus:
Hey, I was wondering what the goal for charts is. Are we expecting that clay-charts will match lexicon design? For example, I am looking at the pie chart at https://lexicondesign.io/docs/patterns/Charts/charts.html and noticed that the legend has the actual percentage value for the label. Is that something that will be available out of the box for charts? Also, for bar charts, will it work out of the box to have multiple bars and labels with different colors? 

These are the designs I am hoping for that would work out of the box for charts.
<img width="340" alt="screen shot 2018-01-24 at 4 10 54 pm" src="https://user-images.githubusercontent.com/6843530/35363881-316bac32-0121-11e8-8913-cc5829f3c088.png">
<img width="377" alt="screen shot 2018-01-24 at 4 10 44 pm" src="https://user-images.githubusercontent.com/6843530/35363885-35219b34-0121-11e8-9a73-cb11afd72447.png">

bryceosterhaus:
cc @jbalsas

jbalsas:
Hey @bryceosterhaus, it is definitely the goal of charts to match Lexicon design out of the box with minimal effort. We'll double-check with the Lexicon team and work on these changes based on their feedback.

Could you share some kind of simple snippet of how you're using it? Just to make sure we reproduce the right thing?

Thanks for letting us know!

bryceosterhaus:
We aren't doing anything special. We are just using it as is in JSX.

Very similar to
https://hosting-metalcharts.wedeploy.io/avdanced-examples/x-axis-tick-formatting/
and 
https://hosting-metalcharts.wedeploy.io/simple-examples/pie/

julien:
Hi @bryceosterhaus,

Concerning the labels, you could use the chart's [`legend`](https://github.com/liferay/clay/blob/master/packages/clay-charts/src/ChartBase.js#L896-L911) option, which will let you set the position's position.

We need to add the missing (`contents`) property to support legend templates like in this example: https://naver.github.io/billboard.js/demo/#Legend.LegendTemplate2, which would let us customize the labels formatting.

Concerning the chart designs, we might need to review them with the UX team since legends usually contain rectangles instead of circles, but thinking of a solution, we could also solve it by providing custom templates if is a good option.

julien:
Hey again @bryceosterhaus,

[...]

PR #494, opened 2018-01-30
Add missing contents property to legend option

julien:


julien:
@Robert-Frampton if you have time to review this, it'd be much appreciated. 
You can have a look at #459 for more context.

Thanks.

packages/clay-charts/demos/index.html
```diff
@@ -370,7 +399,18 @@ <h2>Bubble</h2>
 						document.querySelector('#vertical-small'),
 					];
 				}
-			}
+			},
+			legend: {
+				contents: {
+					bindto: '.pie-legend',
+					template(title, color) {
```
Robert-Frampton: Hey @julien 

Not sure if this is applicable or not, but should we consider bringing this svg template over to Clay somehow so that it can be leveraged? Maybe even if it was just provided as a utility.

packages/clay-charts/src/ChartBase.js
```diff
@@ -889,11 +889,15 @@ ChartBase.STATE = {
 	/**
 	 * Legend display options.
 	 * @instance
-	 * @memberof ChartBase
+	 * @memberof ChartBas
```
Robert-Frampton: Typo.

packages/clay-charts/demos/index.html
julien: Yes, it sounds like a good idea

packages/clay-charts/src/ChartBase.js
julien: Oops, thanks

packages/clay-charts/demos/index.html
julien: @Robert-Frampton, should I close this pull request until we have an idea of how we want to include this svg template?

Robert-Frampton: Hey @julien, since this is adding missing config, let's merge it and then do the abstraction in another issue.

julien:
@Robert-Frampton ok thanks!

PR #497, opened 2018-01-30
Remove unnecessary use of minValue and maxValue in the ProgressBar

matuzalemsteles:
In the API, it no longer uses `minValue` and `maxValue`.

### DIFF
--- a/packages/clay-progress-bar/demos/a11y.html
+++ b/packages/clay-progress-bar/demos/a11y.html
@@ -51,8 +51,6 @@
 
 		new metal.ClayProgressBar(
 			{
-				minValue: 0,
-				maxValue: 100,
 				spritemap: spritemap,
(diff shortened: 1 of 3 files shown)


jbalsas:
Hey @matuzalemsteles, why are those no longer necessary? How does the component work now?

matuzalemsteles:
hey @jbalsas, it no longer has the possibility of being able to choose the value of `minValue` and` maxValue` the logic is in Soy, just pass the `value`. This was corresponding to this PR https://github.com/metal/metal-clay-components/pull/252

julien:
@jbalsas, @matuzalemsteles if I remember correctly the `minValue` would always  be `0` and `maxValue` would always  be `100`.  This is something we saw with @carloslancha and UX.

PR #511, opened 2018-02-05
Use tiles only when passed to state | Fixes #495

### LINKED ISSUE #495: Enable chart background "patterns" on demand
At the moment in [clay-charts](https://github.com/liferay/clay/blob/develop/packages/clay-charts), if the default [patterns are available](https://github.com/liferay/clay/blob/develop/packages/clay-charts/src/ChartBase.js#L75-L81); they are used as "background patterns" (or tiles). 

We should change this behaviour, so that patterns are only used if one activates them.
We could add an `useTiles` option for example.

@matuzalemsteles if you're interested you can have a look at it, and if you need help feel free to ask me.

matuzalemsteles:
hey guys, this corrects the behavior of using the tile only when it is passed to the state I'm not sure if it was the correct way to fix this. Feel free to talk about what you think about it.

##### `color.tiles`
Some changes have been implemented in the use of the `color.tiles` API. It is only allowed to use an array with the names of the tiles. So we only need to use what is defined by lexicon and facilitates its use. See 9a3a75a.

> NOTE: This may seem strange because the API still supports function because it is required to be passed to the billboard.

### Method `constructColorsAndTilesConfig_`

This method may seem confusing at first glance. You must leave the `color` API empty whenever `tiles` are passed to make the `color` API more important than` colors`.

When we have many columns that can share the same `id` or charts it can happen that the billboard returns to` colors` the colors of the respective columns with its id and using in other charts or even the `tiles`. This function ensures that no matter what column id it will pass only what is needed.

### SVG

It was rather strange when I had to add the svg with the `patterns` in the DOM and then collect. With the change of the API ClayCharts would have to generate the svg and pass back to tiles so that billboard can process.

An `SVG` class has been created which has some abstract functions of svg elements so that you can create multiple elements from a single object. See 6dd953f and bc1b1d7

It's a way we can treat SVG internally, leave your thoughts on this implementation.

##### `create`

* **param** {Array} props

It receives an array of objects, the keys define which are the methods that should be called, their attributes and children.

```javascript
[
	{
		nodeName: 'pattern',
		attributes: {},
		children: [
		   {
		       nodeName: 'rect',
		       attributes: {}
		   }
		]
	}
]
```


julien:
Hi @matuzalemsteles,
I think the main goal is to convert the `tiles` option into a boolean with the following behaviour:

- If set to `true`, the charts use tiles we have defined (and provide).
- If set to `false`, the charts doesn't use tiles.

 Ideally we should think of a way to include the tiles only once; and not necesarrily create them via JS. If I recall we decided to have the tiles in `svg` format in case we would need to update them (us or someone from the design or UX team).

If you need help, don't hesitate to ask.

Thanks!

matuzalemsteles:
hey @julien, thanks for clarifying. I'll adjust so we can use it that way.

It just seems strange to have to put the SVG in the DOM and then move on to the billboard.

The SVG I think we should think more as we save and an option would be to use the [DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) and so we could save the SVG in strings and then rendering, is an option. But I do not know how this is feasible.

PR #513, opened 2018-02-06
Use tiles when it's true | Fixes #495

### LINKED ISSUE #495: Enable chart background "patterns" on demand

matuzalemsteles:
> For context, see #511.

hey @julien, I made some changes compared to the previous ones, the tiles are activated with `boolean` and for the time being, they get caught in the DOM and it gets caught when the chart is created. Ensuring that it is not always called to a change.

Just a problem, I think we should treat `colors` when the pieces are not passed, this time the default is to always use `color.pattern` for tile colors because there are sometimes when colors begin to be repeated due to `colors'.


julien:
@Robert-Frampton if you have time could you please review? Thanks.

PR #515, opened 2018-02-06
Add Chart examples to claycss.com (Fixes #456)

### LINKED ISSUE #456: Move charts demos to claycss.com
Currently, we host some docs on about our charting solution in https://hosting-metalcharts.wedeploy.io

We should migrate that into the website for consistency and easiness

/cc @julien, @matuzalemsteles

julien:


package.json
```diff
@@ -12,7 +12,10 @@
     "mcritic": "mcritic packages/ --ignore '**/{browserslist-config-clay-components,claycss.com,generator-metal-clay,node_modules}/**'",
     "prettier": "prettier-eslint packages/clay-*/src/*.js packages/clay-*/src/**/*.js",
     "start": "http-server . -p 4000",
-    "test": "npm run build && npm run jest && npm run a11y"
+    "test": "npm run build && npm run jest && npm run a11y",
+    "electric": "cd packages/clay && npm run build && cd ../clay-charts && npm run compile && npm run sass",
```
jbalsas: Hey @julien, even if we might incur into some overhead now, I assume at some point we'll want to add other components as well, so why not just have these do:

```javascript
...
"electric": "npm run build && cd packages/claycss.com && el",
"web": "npm run electric -- run"
"deploy": "npm run electric -- deploy"
...
```

packages/claycss.com/electric.config.js
```diff
@@ -32,10 +32,10 @@ module.exports = {
 		return generateIconData(data);
 	},
 	codeMirrorLanguages: ['xml', 'htmlmixed', 'soy'],
-	metalComponents: ['electric-quartz-components'],
+	metalComponents: ['electric-quartz-components', 'clay-charts'],
```
jbalsas: While we're at it, we might as well add the rest of clay components, ideally in a programatic way 😉

package.json
julien: Yes, good point.

packages/claycss.com/electric.config.js
julien: That sounds like a better idea, thanks.

julien:
Pull request submitted to https://github.com/liferay/clay/pull/521

PR #550, opened 2018-02-13
Create `data` API to replace `columns` in Charts Fixes #526

### LINKED ISSUE #526: Create `data` API to replace `columns` in Charts

I propose we:
    - An `object` with the data

julien:


julien:
I'm wondering if we couldn't get rid of the [`columns` property in `ChartBase`](https://github.com/julien/clay/blob/chart-new-features/packages/clay-charts/src/ChartBase.js#L766-L783) and only keep `data` instead of maintaining the two. /cc @Robert-Frampton @jbalsas

julien:
Hey @brunobasto and @bryceosterhaus,

Would it be an issue for you if we replaced `columns` with `data` ?

Thanks!

julien:
Hey guys,

I'm closing this pull request since we're also going to replace trailing underscores with leading underscores (see [#574](https://github.com/liferay/clay/pull/574)), but I still need your precious feedback when you can so feel free to comment even though this one is closed.

Thanks!

brunobasto:
Hey @julien, I think it should be fine for me.

julien:
Ok, Thanks @brunobasto

PR #559, opened 2018-02-15
Update ClaySelect markup and remove wrapperType to use the inline API| Fixes #558

### LINKED ISSUE #558: Update ClaySelect Markup
The class `.form-inline-group` was renamed to `.form-group-item` in [v2.0.0-beta.5 December 25, 2017](https://github.com/liferay/clay/blob/master/packages/clay/CHANGELOG.md#v200-beta5-december-25-2017). Hey @pat270 let me know if I'm wrong about this.

I do not know if it does but feel we have the `wrapperType` API, maybe we can just go with an `inline` to **true** to add the `.form-group-item` class. This way it does not risk breaking an API when a markup change and does not depend on the API input.

**Current**
```soy
class="
	{if $wrapperType}
		form-{$wrapperType}
	{else}
		form-group
	{/if}

	{if $elementClasses}
		{sp}{$elementClasses}
	{/if}
"
```

**New**
```soy
class="
	{if $inline}
		form-group-item
	{else}
		form-group
	{/if}

	{if $elementClasses}
		{sp}{$elementClasses}
	{/if}
"
```

I do not know if it's the best time to change the API. What do you think? @carloslancha

matuzalemsteles:


carloslancha:
Merged, thx!

PR #561, opened 2018-02-15
Adds disabled="true" to force your render separate from other attributes | Fixes #522

### LINKED ISSUE #522: Disabled attribute markup is incorrect for button
In the Clay Sample portlet, when you add a clay button taglib with attribute `disabled="true"`, the generated markup is `disabledtype=button`, so the button is not disabled. Tested on https://liferay-master.wedeploy.io/

matuzalemsteles:
This can fix issue #522, forcing the attribute to be rendered separate from others. We still have to investigate why this happens only in TagLibs.

### DIFF
--- a/packages/clay-button/src/ClayButton.soy
+++ b/packages/clay-button/src/ClayButton.soy
@@ -164,7 +164,7 @@
 	{/if}
 
 	{if $disabled}
-		disabled
+		disabled="true"
 	{/if}
 
 	{if $id}
--- a/packages/clay-button/src/ClayButton.soy
+++ b/packages/clay-button/src/ClayButton.soy
@@ -164,7 +164,7 @@
 	{/if}
 
 	{if $disabled}
-		disabled="true"
+		disabled="disabled"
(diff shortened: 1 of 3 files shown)


packages/clay-button/src/ClayButton.soy
```diff
@@ -164,7 +164,7 @@
 	{/if}
 
 	{if $disabled}
-		disabled
+		disabled="true"
```
carloslancha: According to disabled docs this should be `disabled="disabled"`.

Source: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/disabled

Thx!

matuzalemsteles: This makes more sense now. Thanks @carloslancha.

carloslancha:
Merged, thx!

yuchi:
Could the renderer being outputting XHTML/HTML 4? Boolean props are relatively new AFAI remember.

PR #562, opened 2018-02-15
Add onlySearch API to render only search | Fixes #543

### LINKED ISSUE #543: Implement Search only content markup in ManagmentToolbar
We need to implement in the component this feature https://github.com/liferay/clay/issues/512

The one that should be implement is the al "Always Open" example. Approved by UX.

matuzalemsteles:
hey @carloslancha, I thought to follow the line of adding a Flag, to say when you want only the search, not to have to create many conditions in the code just to know when the user only wants the search. Maybe we could follow the line of a new component within ManagementToolbar just for this use case but I do not know if it is needed at the moment. This is a short implementation.

### DIFF
--- a/packages/clay-management-toolbar/demos/a11y.html
+++ b/packages/clay-management-toolbar/demos/a11y.html
@@ -24,12 +24,20 @@
 		ClayManagementToolbar
 	</h1>
 
+	<div id="container0"></div>
 	<div id="container1"></div>
 	<div id="container2"></div>
 	<div id="container3"></div>
 
 	<script type="text/javascript">
 		var spritemap = '../../../node_modules/clay/lib/images/icons/icons.svg';
+
+		new metal.ClayManagementToolbar({
+			onlySearch: true,
+			searchFormName: 'mySearchName',
+			searchInputName: 'mySearchInputName',
+			spritemap: spritemap
+		}, '#container0');
 
 		new metal.ClayManagementToolbar({
 			creationMenu: {
--- a/packages/clay-management-toolbar/demos/index.html
+++ b/packages/clay-management-toolbar/demos/index.html
@@ -24,12 +24,20 @@
 		ClayManagementToolbar
 	</h1>
 
+	<div id="container0"></div>
 	<div id="container1"></div>
 	<div id="container2"></div>
 	<div id="container3"></div>
 
 	<script type="text/javascript">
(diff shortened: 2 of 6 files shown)


carloslancha:
Hey @matuzalemsteles I was waiting for UX to end the correct definition of this.

The thing is that there were some stuff that always were there like filters, order or creationButton, but now they're all optional, so I thing that search should appear alone in in the same way filters or creation button can do it: by passing only the params for that (filterItems, creationMenu, searchActionURL...).

So we need to change a little the template to only show previously always visible fields to only appear when params are there, instead defining "onlyWhatever" params. Makes sense?

matuzalemsteles:
hey @carloslancha, That makes sense, but we have some use cases to just leave the search, we would have to change some things just for this to work for him, classes that only make you feel when he is alone, at the same time we create a lot of conditions for it.

For example `creationMenu` has 3 use cases, it uses Dropdown when it has more than one item, it makes a link when it has only one item and when no item is passed the default is just a Button. For the last behavior to work as expected the user would have to pass only `creationMenu: {}` empty, it may seem a little strange.

We have `sorting` that appears by default and we have no APIs that we can indicate when it will appear or not.

Do you think it's worth it? to leave this behavior. I like the idea of rendering only the parts according to the params.

carloslancha:
Yeah @matuzalemsteles that's why I said that we need to change some stuff to fit with new requirements.

For example the `sorting` was always there before, now we need to modify it to render it only if you pass a sortingOrder (asc or desc) and remove the default value.

The same happens with creationMenu, until now it was always there, that's why if you didn't pass creationMenu it was rendered as only a button. Now we need to think about other API for this (yep, passing `{}` is not ok).

And probably with more options...

About the search we're doing stuff like that in other components, if there are classes that must only be there if only search is present then we probably would need something like `{let $searchOnly: not $filterItems and not $creationMenu and .... etc}` and play with that. What I mean is that yes, is a lot of conditioning but it's our logic, I don't think is a good idea to add that complexity to the API.

I wish this hadn't happened when we are on RC 😅 

(FYI you can follow this changes in the management toolbar requirements google doc)

matuzalemsteles:
Yeah, oh okay 😅. Thanks for clarifying. For a moment I thought we would just implement the search case. I will close this to work again on these changes we discussed.

PR #563, opened 2018-02-15
Add flag to indicate when focusable is true or not | Fixes #536

### LINKED ISSUE #536: IE: Tabbing through the document focuses SVG elements
Clay SVG's need the attribute `focusable="false"`

matuzalemsteles:


packages/clay-icon/src/ClayIcon.js
```diff
@@ -25,6 +25,15 @@ ClayIcon.STATE = {
 	 */
 	elementClasses: Config.string(),
 
+	/**
+	 * Flag to indicate when svg focusable or not.
```
carloslancha: Typo: Flag to indicate if the svg is focusable or not

packages/clay-icon/src/ClayIcon.soy
```diff
@@ -20,6 +21,12 @@
 			{/if}
 		"
 
+		{if $focusable}
```
carloslancha: Can this be written `focusable="{$focusable == true}"` ?

packages/clay-icon/src/ClayIcon.js
matuzalemsteles: That looks better, thanks!

packages/clay-icon/src/ClayIcon.soy
matuzalemsteles: No, the attributes do not allow **bool** to be used ☹️. To get as close to that `focusable="{$focusable == true ? 'true' : 'false'}"`(is that what you meant?). Sounds good to you?

carloslancha: Nah, leave it as it was 😄 Thx!

carloslancha:
Merged, thx!

PR #578, opened 2018-02-20
Create `data` API to replace `columns` in Charts

### LINKED ISSUE #526: Create `data` API to replace `columns` in Charts

julien:
Fixes #526


packages/clay-charts/demos/index.html
```diff
@@ -338,8 +338,41 @@ <h2>Bubble</h2>
...
+				},
+				{
+					data: [130, 150, 200, 300, 200, 100],
+					id: 'data2'
+				}
+			]
+		};
+
+		var lineChartWithData = new metal.LineChart({
+			columns: lineChartData.columns_1,
```
jbalsas: still columns ;)

packages/clay-charts/src/ChartBase.js
```diff
@@ -70,19 +71,26 @@ const ChartBase = {
 	 * @inheritDoc
 	 */
 	attached() {
-		const config = this._constructChartConfig();
+		const data = this.columns || this.data;
```
jbalsas: Why in this order and not `this.data || this.columns`

packages/clay-charts/demos/index.html
julien: Yes, weren't we still supposed to support columns as well as data?

packages/clay-charts/src/Geomap.js
```diff
@@ -47,7 +48,21 @@ class Geomap extends Component {
 		this.path = d3.geoPath().projection(this.projection);
 		this._selected = null;
 
-		d3.json(this.data, this._onDataLoad.bind(this));
+		this._onDataLoadHandler = this._onDataLoad.bind(this);
+
+		if (isString(this.data)) {
```
jbalsas: Why don't we have a similar `resolveData` here, like in `ChartBase`?

packages/clay-charts/demos/index.html
jbalsas: Yeah, but no need to showcase it in the demos :)

julien: Make sense, ok thanks

packages/clay-charts/src/ChartBase.js
```diff
@@ -731,21 +763,52 @@ ChartBase.STATE = {
...
+				data: Config.array().required(),
+				hide: Config.bool(),
+				id: Config.required().string(),
+				name: Config.string(),
+				regions: Config.array(),
+				type: Config.oneOf(types.all),
+				x: Config.string(),
+			})
+		),
+		Config.func(),
```
jbalsas: I don't think we want to allow the other types for the old API. `columns` should accept only the old column shape

packages/clay-charts/src/LineChart.js
```diff
@@ -16,20 +16,24 @@ LineChart.STATE = {
...
+				data: Config.array().required(),
+				hide: Config.bool(),
+				id: Config.required().string(),
+				name: Config.string(),
+				regions: Config.array(),
+				type: Config.oneOf(types.all),
+				x: Config.string(),
+			})
+		),
+		Config.func(),
```
jbalsas: Just accept the old type, not the new data options

packages/clay-charts/src/__tests__/AreaSplineChart.js
```diff
@@ -18,8 +18,8 @@ afterAll(() => {
 });
 
 describe('AreaSplineChart', function() {
-	it('should be pass correct type to billboard.js', () => {
-		new AreaSplineChart({
+	it('should be pass correct type to billboard.js', async () => {
+		await new AreaSplineChart({
 			columns: [],
```
jbalsas: We should update the tests to use the new `data` API instead of the deprecated `columns` one

packages/clay-charts/src/__tests__/Chart.js
```diff
@@ -17,23 +17,20 @@ afterAll(() => {
 	bb.generate.mockRestore();
 });
 
-describe('Chart', function() {
-	it('should be pass correctly formatted configuration options to billboard.js', () => {
-		const chart = new Chart({
+describe.only('Chart', () => {
```
jbalsas: Remove `.only`

packages/clay-charts/src/ChartBase.js
```diff
@@ -450,6 +452,18 @@ const ChartBase = {
 		}
```
jbalsas: There's something fishy here and in `_handleDataChanged`. It looks like this is being triggered when data changes because down there we do `this.columns = data`. This also means that we are not syncing data if this gets set via `columns` instead

We also risk getting in an infinite loop by triggering dataChanged and columnsChanged over and over, so I think the underlying model needs to be better thought out. 

Could you take a closer look?

packages/clay-charts/src/ChartBase.js
julien: Yes it make sense

packages/clay-charts/src/LineChart.js
julien: Ok

packages/clay-charts/src/__tests__/AreaSplineChart.js
julien: Ok

packages/clay-charts/src/__tests__/Chart.js
julien: Sorry

packages/clay-charts/src/ChartBase.js
julien: Yes, that's actually something I wanted to check with you. Thanks

julien:
Pull request sent to https://github.com/liferay/clay/pull/627

PR #599, opened 2018-02-21
Dropdown trigger need space between icon and label when both are passed | Fixes #597

### LINKED ISSUE #597: Dropdown trigger need space between icon and label when both are passed

carloslancha:


matuzalemsteles:

packages/clay-dropdown/src/ClayDropdown.soy
```diff
@@ -19,10 +19,12 @@
 	{@param? searchable: bool}
 	{@param? style: string}
 	{@param? triggerClasses: string}
+	{@param? triggerSize: string}
 
   {let $trigger kind="html"}
     {if $icon}
       {call ClayIcon.render}
+        {param elementClasses: $label ? 'inline-item inline-item-before' : '' /}
```
matuzalemsteles: hey @carloslancha, Since we added this with the icon before, would it make sense to add to an icon after the label?  `inline-item inline-item-after`.

packages/clay-dropdown/src/ClayDropdownBase.soy
```diff
@@ -218,6 +221,10 @@
 				{sp}{$classes}
 			{/if}
 
+			{if $size}
+				{sp}btn-{$size}
```
matuzalemsteles: Hey @carloslancha the use of `btn-sm` does not seem to have any effect and does not seem to work well with `btn-unstyled`. 

Hey @pat270 could you look at this if it would be correct?

carloslancha: btn-unstyled overwrites the size. It's something @pat270 might take a look, for this pr is ok

packages/clay-dropdown/src/ClayDropdown.soy
carloslancha: I'd not add it until somebody needs it

matuzalemsteles: Lexicon use cases seem to cover this.
screen shot 2018-02-21 at 11 22 46

Currently it does not have any spacing, since in ClayButton we have spacing, I think it makes sense to put, to ensure consistency of use of the Buttons.
screen shot 2018-02-21 at 11 24 34

packages/clay-dropdown/src/ClayDropdownBase.soy
matuzalemsteles: Yes, I opened an issue for this so we can continue with this pr. #600

packages/clay-dropdown/src/ClayDropdown.soy
carloslancha: Ok, makes sense

carloslancha: But we need to be aware that then everywhere will be a space between label and caret-down, like in management-toolbar, where right now there's no such space

carloslancha: Lexicon defined it right, but that doesn't fit with @pat270 markup, so it should have it, just pointing it out.

matuzalemsteles: Hmm, Yeah, I think we should look at this more carefully. Hey @pat270 would we cause many problems in adding this?

pat270: @matuzalemsteles I don't think it would cause major problems as long as spaces are stripped. If you wanted to use a space instead of the margin you could omit `inline-item-after`.

matuzalemsteles: Great, thanks @pat270 for the info. 

Hey @carloslancha what do you think?

PR #618, opened 2018-02-22
Migrate claycss to Electric 3 | Fixes #617

### LINKED ISSUE #617: Migrate site to Electric 3

carloslancha:
DO NOT MERGE

We're experiencing two issues.

1)
When we have a `.md` page that uses a soy component, for example `ClayChart` we get a compilation error at `metal:render:html` task:
```
events.js:182
      throw er; // Unhandled 'error' event
      ^
Error: TypeError: Cannot read property 'getTotalLength' of null
```

That `getTotalLength` is a method of billboard. This is the code we're using:

```
---
title: "Bar Chart"
description: ""
layout: "guide"
weight: 3
---

###### {$page.description}

<article id="1">

{call ClayChart.render}
	{param columns: [
		[
			'id': 'data1',
			'data': [100, 20, 30]
		],
		[
			'id': 'data2',
			'data': [20, 70, 100]
		]
	] /}
	{param type: 'bar' /}
{/call}

</article>
```

The funny part is that if we run the electric site removing first all those calls to ClayChart everything works fine and starts watching. If then we add again the ClayChart call and save the file the build starts again and... it works!!

2)
The other problem we found is that when we call a soy template from a `.md` file with quotes like:
```
{call guide.render data="all"}
   {param content kind="html"}Foo{/param}
{/call}
```
compilation fails (just when we run it from zero, not watching, the same as the preview case, if we add that while watching it compiles ok) giving us the error:
```
Error: Error in plugin 'metal-tools-soy'
Message:
    Compile error:
errors during Soy compilation
In file .temp/task/metal/site/pages/docs/components/charts/basic/bar_chart.soy:15:1: Malformed attributes in 'param' command text (kind=&quot;html&quot;).
{param content kind=&quot;html&quot;}foo{/param}
^

In file .temp/task/metal/site/pages/docs/components/charts/basic/bar_chart.soy:14:4: Malformed attributes in 'call' command text (data=&quot;all&quot;).
<p>{call guide.render data=&quot;all&quot;}
```


jbalsas:
@robframpton, could you take a look? 🙏

zenorocha:
@robframpton I can confirm that this is happening on the latest version of Electric.

@carloslancha I sent a pull request to your fork with some additional changes needed: https://github.com/carloslancha/clay/pull/3

robframpton:
With Electric 3.0, we're using Metal server side rendering. Which means if a component attempts to do anything with the DOM during any of the lifecycle methods there is a good chance you'll get an error. This is why we're doing checks [like this](https://github.com/carloslancha/clay/pull/3/files#diff-e2bcad6881830f4a6efcbdc763047204R12).

For the clay charts, you'll need to add something similar so that Billboard doesn't attempt to render during SSR.

robframpton:
In regards to the second problem, we didn't update the version of metal-tools-soy, so I'm not sure what's causing it.

PR #620, opened 2018-02-22
Update dropdown | Fixes #619

### LINKED ISSUE #619: form-groups in a dropdown
We need to use form-groups in a dropdown, we can place all form-groups in the same 'dropdown-section' but the form-group margin-bottom should be 1rem

Now:
image

form-group with margin-bottom: 1rem
image

marcoscv-work:


carloslancha:
@marcoscv-work could you please add an example?

pat270:
Hey @marcoscv-work,

`$dropdown-item-indicator-spacer-*` should only be used to space `.dropdown-item-indicator` in http://pat270.github.io/lexicon/content/dropdowns/#dropdown-menu-with-indicators. We also need to make sure the total space between `.dropdown-section .form-group` and `.dropdown-item` is also equivalent to `$dropdown-item-padding-y * 2` (1rem in Atlas, 0.5rem in Base).

pat270:
One thing I saw somewhere (I forget where) is that dropdown with search should be 32px tall? Should the form elements in #619 also be 32px tall?

victorvalle:
Lexicon team members have been talking and in advance search we want the fields to be 32px height. But we don't want it as a general rule right now for different reasons we would like to study further.

marcoscv-work:
So we could use "form-control-sm" by the moment to set the height to 32px, do we agree?

image

About the space between form-control
@pat270 Can you accommodate the var name and merge?

pat270:
I sent an update. I'm closing this in favor of https://github.com/liferay/clay/pull/636.

PR #625, opened 2018-02-22
Check if the dispose has already been done and does nothing | Fixes #582

### LINKED ISSUE #582: ClayAlert disposed instances issues
We need to look into ClayAlert, ClayToast and ClayStripe to investigate the error

```
    console.warn node_modules/metal-state/lib/State.js:788
      Error. Trying to access property "_visible" on disposed instance
    console.warn node_modules/metal-state/lib/State.js:788
      Error. Trying to access property "destroyOnHide" on disposed instance
```

matuzalemsteles:
hey @carloslancha, the messages from **warn** happened when the dispose was made on the components that used `ClayAlertBase`, the dispose happened from top to bottom and when it arrived in `ClayAlertBase` in some cases it had no more reference of `_visible`... The way to omit this is to check when it has already been disposed.

### DIFF
--- a/packages/clay-alert/src/ClayAlertBase.js
+++ b/packages/clay-alert/src/ClayAlertBase.js
@@ -52,15 +52,17 @@ class ClayAlertBase extends Component {
 	 * @private
 	 */
 	_defaultHideAlert() {
-		this._delayTime = 0;
-		this._visible = false;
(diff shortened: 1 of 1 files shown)


carloslancha:
Makes sense, thx @matuzalemsteles

PR #629, opened 2018-02-23
Create `data` API to replace `columns` in Charts

### LINKED ISSUE #526: Create `data` API to replace `columns` in Charts

julien:
Fixes #526


jbalsas:
Hey @julien, this looks almost good to go!

Have we added any test to check that both the `promise` and `string` flavours of the `data` API work as expected?

julien:
@jbalsas I'm on it ;) thanks

Issue #534, opened 2018-02-08
Create electric-clay-components to use on claycss.com


matuzalemsteles:
The [electric-quartz-components](https://github.com/deprecate/electric-quartz-components) has been depreciated and we should start using [electric-marble-components](https://github.com/electricjs/electric/tree/master/packages/electric-marble-components) at claycss.com.

jbalsas:
Actually, we should probably create electric-clay-components instead.

The reason is that we follow Lexicon style and marble is Wedeploy style, so we need to keep ours for the lexicon-oriented sites.

matuzalemsteles:
Oh, okay. Will we keep this package here or will we keep it at Electric? I think we need some components that exist in `electric-marble-components` to generate the API page more easily 🙂 so that we do not have to create many things. Maybe we can start working on this package.

Thanks for clarifying!

zenorocha:
Done on https://www.npmjs.com/package/electric-clay-components

PR #635, opened 2018-02-23
Use electric-clay-components instead of electric-quartz-components #534

zenorocha:


matuzalemsteles:
Hey @zenorocha, can you regenerate the `yarn.lock` with changes?

zenorocha:
Done @matuzalemsteles ;)

matuzalemsteles:
Thanks. Merged!

carloslancha:
Hey @zenorocha with this changes we get this error when we try to render a soy component in the electric site:

`Uncaught TypeError: comp.getRenderer(...).renderInsidePatch is not a function`

the code that causes it is:
```
{call ClayButton.render}
   {param label: 'whatever' /}
{/call}
```

carloslancha:
We also get this error
`Warning: Error: Invalid state passed to 'items'. Expected type 'array', but received type 'string'. Passed to 'ClayDropdown'.`

from this code:
```
```text/html
<clay-dropdown
	items='[{"label": "Action", "url": "#1"},{"separator": true, "type": : "group"},{"label": "Scope", "url": "#1"}]'
	spritemap="/vendor/lexicon/icons.svg"
	triggerLabel="Primary"
	triggerStyle="primary">
</clay-dropdown>
\```
```

There we want to render the use, not the result, so it shouldn't try to execute it.

zenorocha:
Hmm, this PR shouldn't introduce that error. If that's the case, it's better to revert the commit and use quartz components until the problem is figured it out.

PR #650, opened 2018-02-26
 Add and remove active class on check | Fixes #649

### LINKED ISSUE #649: ClayImageCard needs to add and remove active class on check
Since the Lexicon team moved the clickable area to the image, we need to add and remove the `active` on check.

matuzalemsteles:


carloslancha:
Hey @matuzalemsteles we shouldn't change public state from the component, but I can see the problem here. Tomorrow I'll talk with Chema about how we want to deal with controlled/uncontrolled components to move forward with this, ok?

Thx!

matuzalemsteles:
hey @carloslancha, that makes sense even, maybe we should go with an internal `Flag` so it can make that active switch, but I see more sense with `selected` or just letting in `soy` when it is **selected** and let dev do this switch as a practice.

carloslancha:
Depends on https://github.com/liferay/clay/issues/668 and https://github.com/liferay/clay/issues/670

PR #653, opened 2018-02-27
ClayCSS Update Several pages

pat270:


matuzalemsteles:

packages/claycss.com/src/pages/docs/components/form_custom.md
```diff
@@ -5,9 +5,10 @@ layout: "guide"
 weight: 100
 ---
 
-<article id="clay-custom-checkbox">
-
-### Custom Checkbox
+<article id="custom-checkbox">
```
matuzalemsteles: Hey @pat270, the `clay-custom-checkbox` he adds via CSS always on the second `WEB COMPONENTS` tab, it was a way that Chema found to have the HTML highlights and in the `Web components` tab name, we have a list huge in `_code.scss` to do this, maybe we can improve this.

pat270: I added the selector `li:nth-last-child(3) + li > button` to style the Web Components tab. The order should be Soy, Web Components, Html always. There will be problems if there are 3 item code example tabs that don't have Web Components as the second item (maybe in charts). We might be able to add some id at the top of the page so this only applies to the components library.

matuzalemsteles:

packages/claycss.com/src/pages/docs/components/form_custom.md
matuzalemsteles: Sounds good to me. Thanks @pat270 🙂

matuzalemsteles:
Merged, thanks!

PR #637, opened 2018-02-24
#600 Breaking Dropdown Action Togle and Quick Action Item require `.component-action`

pat270:

Issue #600, opened 2018-02-21
`btn-sm` is not working with `btn-unstyled`


matuzalemsteles:
Hey @pat270, in the case of using `btn-sm` with`btn-unstyled` does not seem to work very well, see #599 for an example.

pat270:
@matuzalemsteles I created `btn-unstyled` because we wanted to use `button` for interactive elements that didn't have any `href` attribute. It seemed easier to create a new class instead of using `btn-link` because there are a lot of css overwrites I needed to add to make it function like an `inline` element (e.g. `a` tag) and I would need to re-add `btn-link` styles to places where we would want it to function like Bootstrap 4 intended.

I prefer not to make `btn-unstyled` more complicated and would rather add some documentation on it. Is there a specific use case you can show me?

matuzalemsteles:
I understand, thanks for clarifying and I agree with you. A case use is in ClayDropdown, It may happen to have a button with `btn-unstyled` along with `btn-sm` but I do not know if we would have cases with that for now.

pat270:
I see, if that combo can be output I can just add support for it.

matuzalemsteles:
Sounds good to me.

carloslancha:
Hey @pat270 there's a case in blogs where they're using a dropdown with a small unstyled button, they can more or less do the trick by passing `elementClasses` but it would be better if we can add support for it. Thx!

pat270:
@carloslancha  When I created `.dropdown-action`, I ran under the assumption it would be the same color across the board https://lexicondesign.io/docs/patterns/buttons.html, then there became the sites stuff https://lexicondesign.io/docs/patterns/Sites/assetTitleBar.html.

It's possible to create that blue vertical ellipsis button using:
```
<div class="dropdown">
    <button class="btn btn-outline-primary btn-outline-borderless btn-monospaced btn-sm dropdown-toggle"></button>
    <div class="dropdown-menu"></div>
</div>
```
with the caveat that the icon size isn't 1rem.

I will probably need to revisit `.dropdown-action` colors and support for different `.btn-*` modifiers (e.g. `.btn-outline-primary.btn-outline-borderless`) so that you can create the action dropdowns in Lexicon Sites. Generally `.btn-unstyled` shouldn't be used unless you are trying to style a button the same style as a link in a component.

carloslancha:
@pat270 makes sense, yep :)

PR #683, opened 2018-03-01
#600 Breaking Dropdown Action Toggle and Quick Action Item require `.component-action`

matuzalemsteles:
See context #637

Hey @pat270, I do not know if this is expected for Dropdown Actions with `component-action`, the **focus** differs from others, are they expected?


pat270:
@matuzalemsteles do you have an example or screenshot of what is different about the focus?

matuzalemsteles:
Hey @pat270, Sorry for delay.

For example:
screen shot 2018-03-05 at 09 46 59
screen shot 2018-03-05 at 09 46 46

pat270:
@matuzalemsteles
Lexicon doesn't define any focus state in any of their docs. I changed it to match btn-secondary since they share the same colors. If you want it to be blue we can make it blue, but I still prefer we use the same btn classes across items in management bar (mentioned above) even for dropdown action. It allows us to scope any style changes just to management bar.

Or I can add component action to all the navbars (I implemented this pattern late after I saw more of the site designs).

matuzalemsteles:
Hey @carloslancha, I talked to @pat270 and not is need to use `component-action` in `ClayManagementToolbar`. I think this is ready to go, could you check to see if something is missing?

carloslancha:
@matuzalemsteles @pat270 what is the logic for `component-action` then? Where should and where shouldn't be?

pat270:
@carloslancha The logic for `component-action` is for actions in components, except in Management Bar, Navigation Bar, Application Bar, Pagination, Panels, Progress Bars etc. `component-action` was implemented after a lot of those components and there is no support to customize it at the moment.

carloslancha:

carloslancha:
Resent https://github.com/liferay/clay/pull/725

PR #704, opened 2018-03-07
Add pollingInterval option to Charts

julien:


packages/clay-charts/src/Chart.js
```diff
@@ -11,7 +12,8 @@ import templates from './Chart.soy.js';
 class Chart extends Component {}
 
 Object.assign(Chart.prototype, ChartBase);
-Chart.STATE = ChartBase.STATE;
+Object.assign(Chart.prototype, DataComponent);
```
jbalsas: Why not `Object.assign(Chart.prototype, DataComponent, ChartBase);` all in one line?

packages/clay-charts/src/Chart.js
```diff
@@ -11,7 +12,8 @@ import templates from './Chart.soy.js';
...
+Chart.STATE = Object.assign(ChartBase.STATE, DataComponent.STATE);
```
jbalsas: Should this rather be `Object.assign({}, ChartBase.STATE, DataComponent.STATE)`? Otherwise, aren't we mutating `ChartBase.STATE`?

packages/clay-charts/src/ChartBase.js
```diff
@@ -77,6 +76,7 @@ const ChartBase = {
 		}
 
 		this._resolveData(this.data).then(data => {
+			this._setupPolling();
```
jbalsas: This should go to `DataComponent`, as `ChartBase` has no knowledge at all about polling

packages/clay-charts/src/ChartBase.js
```diff
@@ -103,6 +103,10 @@ const ChartBase = {
 			return;
 		}
 
+		if (this._pollingInterval) {
+			clearInterval(this._pollingInterval);
+		}
```
jbalsas: This should go to DataComponent, as ChartBase has no knowledge at all about polling

packages/clay-charts/src/Geomap.js
```diff
@@ -19,6 +19,8 @@ class Geomap extends Component {
 			return;
 		}
 
+		this._setupPolling();
```
jbalsas: This should go to DataComponent, as ChartBase has no knowledge at all about polling

packages/clay-charts/src/Geomap.js
```diff
@@ -72,6 +74,10 @@ class Geomap extends Component {
 			return;
 		}
 
+		if (this._pollingInterval) {
+			clearTimeout(this._pollingInterval);
+		}
```
jbalsas: This should go to DataComponent, as ChartBase has no knowledge at all about polling

packages/clay-charts/src/Geomap.js
```diff
@@ -263,20 +262,10 @@ Geomap.STATE = {
...
-	 */
-	data: Config.oneOfType([
-		Config.func(),
-		Config.object(),
-		Config.string(),
-	]).required(),
 };
 
+Object.assign(Geomap.prototype, DataComponent);
+Object.assign(Geomap.STATE, DataComponent.STATE);
```
jbalsas: Same as before?

packages/clay-charts/src/ChartBase.js
julien: Ok, but `ChartBase` (or `Chart` if you prefer) will have to call this method defined in `DataComponent` since it can't be called `disposed`. Do you agree?

packages/clay-charts/src/Chart.js
julien: Yes.👍

packages/clay-charts/src/Chart.js
julien: Yes we are, good catch thanks

packages/clay-charts/src/Geomap.js
julien: Here, I'm mutating `Geomap.prototype` which is what I want and doing the same which `Geomap.STATE` which is another object so I don't really want to mix the two, but I'm not modifying `DataComponent` or `DataComponent.STATE`, did I miss anything?

julien: It probably makes more sense to have `Geomap` extend from `DataComponent` since itself extends `Component`, and just copy the `STATE` over. What do you think?

packages/clay-charts/src/Chart.js
```diff
@@ -8,10 +8,10 @@ import templates from './Chart.soy.js';
  * @extends Component
  * @extends ChartBase
  */
-class Chart extends Component {}
+class Chart extends DataComponent {}
 
 Object.assign(Chart.prototype, ChartBase);
-Chart.STATE = ChartBase.STATE;
+Chart.STATE = Object.assign({}, ChartBase.STATE, DataComponent.STATE);
```
jbalsas: I think that we probably don't need to merge `DataComponent.STATE` here, since metal will pick it up from the class chain. Could you double check?

packages/clay-charts/src/Geomap.js
```diff
@@ -263,20 +259,9 @@ Geomap.STATE = {
...
-	 * @default undefined
-	 */
-	data: Config.oneOfType([
-		Config.func(),
-		Config.object(),
-		Config.string(),
-	]).required(),
 };
 
+Object.assign(Geomap.STATE, DataComponent.STATE);
```
jbalsas: Since we already extend `DataComponent`, is this really necessary?

julien: No, you're right.

packages/clay-charts/src/Chart.js
julien: Yes you're right, this is not needed. Thanks.

packages/clay-charts/src/ChartBase.js
jbalsas: Maybe this could go inside the `_resolveData` method, right before resolving the `Promise`?

packages/clay-charts/src/ChartBase.js
```diff
@@ -103,6 +103,8 @@ const ChartBase = {
 			return;
 		}
 
+		this.cleanup();
```
jbalsas: Do we really need to add this additional lifecycle method? Wouldn't it be enough to just implement `disposed()` in `DataComponent`

julien: Yes, you're right; that's the thing to do, thanks.

PR #716, opened 2018-03-12
Returning empty string when no attribute is present | Fixes #715

### LINKED ISSUE #715: ClayComponent.getAttribute does not respect the spec when the attribute has no value
[the spec](http://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-666EE0F9) says that the return value must be the `empty string` when the attribute has no value.

However, clay returns `undefined` when there is no value for an attribute. This makes tools like Alloy UI to get confused and malfunction in some weird corner cases that @carloslancha has witnessed with me.

More info at https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#Notes.

4lejandrito:
As per: http://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-666EE0F9

### DIFF
--- a/packages/clay-component/src/ClayComponent.js
+++ b/packages/clay-component/src/ClayComponent.js
@@ -26,7 +26,7 @@ class ClayComponent extends Component {
 				attributeValue = domData.get(this.element, attributeName);
 			}
 
-			return attributeValue;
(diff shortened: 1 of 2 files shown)


carloslancha:

carloslancha:
Resent https://github.com/liferay/clay/pull/720

PR #747, opened 2018-03-15
Adds new JSDOC tags to ChartBase

matuzalemsteles:
When generating the API Docs, the methods of the ChartBase class are rendered out of context. I'm adding some JSDOC tags to try to adjust this.


jbalsas:
Looks good!

PR #758, opened 2018-03-23
LPS-78852 Fix assignment of the navigation element

jeyvison:


jbalsas:
Hey @jeyvison, I'm not sure this is the right fix. Please see my comment on https://issues.liferay.com/browse/LPS-78852. If there's still something on this side, please, open an issue in this repo with simple reproduction steps we can verify and add some additional information so we can actually know why the fix is needed.

Thanks!

Issue #788, opened 2018-04-11
Typography adjustments for adaptation to system font


atjuan:
The font values proposed are described in the following table:
https://drive.google.com/open?id=1gxsFP0A-hIiELwV9PuGYiiG0CX83Unt7nLtD-GqF4LA

PR #797, opened 2018-04-13
(#788) Atlas adjust font sizes

pat270:


jbalsas:
Hey @pat270, why use an `<h4>` for the modal dialog instead of the `<div>`, or what I assume would be better, an `<h1>`?

pat270:
@jbalsas The document with font updates also had specific `h` tags and I forgot to undo this change. I asked them to remove it since `h` tags should depend on the structure of the document which we have no control over.

jbalsas:
Sounds good, thanks @pat270! We might want to consider using a more semantic markup for this... 🤔

pat270:
I'm open to suggestions, just not sure what we can do since the markup would depend on the content and how it's structured.

jbalsas:
No rush, just thinking out loud :)

Issue #798, opened 2018-04-13
Label Lg space between text is too large when breaking into new line


pat270:
<img width="386" alt="label-spacing" src="https://user-images.githubusercontent.com/788266/38760546-cbd1717e-3f30-11e8-80eb-1b9d0c828e7f.png">

PR #799, opened 2018-04-16
(#798) Adjust spacing between words that break to new line in Label and Badges

pat270:
Also fix labels inside `.form-control-tag-group` to size and break properly


jbalsas:
@carloslancha, @matuzalemsteles, this might imply we need to update the cards markup. Could any of you double-check?

Thanks!

carloslancha:
@jbalsas Cards itself doesn't change, badges and labels does, so we need to update that.

carloslancha:
Great @matuzalemsteles thanks!

matuzalemsteles:

PR #814, opened 2018-04-18
IFI-219: Implement Predictive Chart

julien:


packages/clay-charts/demos/index.html
```diff
@@ -463,6 +468,64 @@ <h2>Bubble</h2>
...
+						'2018-05-01',
+						'2018-06-01',
+						'2018-07-01',
+						'2018-08-01',
+						'2018-09-01',
+						'2018-10-01',
+						'2018-11-01',
+						'2018-12-01'
+					],
+					id: 'x'
```
jbalsas: Is this a convention for time series? Where does it say that this becomes the x-axis data?

julien: Yes, in billboard.js if a column has the id `x` it is used as the x-axis data.

jbalsas: 👍

jbalsas: Isn't there a shortcut, or more meaningful API for it?

julien: I'll check that out right now, (I agree that it's not very meaningul)
Otherwise I was thinking we could maybe add a `timeseries` paremeter that would be used to  generated this column.

jbalsas: Yeah, that sounds better...

packages/clay-charts/src/PredictiveChart.js
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+
+	/**
+	 * Normalizes the data.
+	 * @memberof PredictiveChart
+	 * @private
+	 */
+	_normalizeData() {
+		for (let i = 0, l = this.data.length; i < l; i++) {
```
jbalsas: What about:

```javascript
this.data = this.data.map(
	dataSeries => {
		if (dataSeries.id !== 'x') {
			dataSeries.data = dataSeries.data.map(
				element => {
					return isNumber(element) ? {low: element, mid: element, high: element} : element;
				}
			);
		}

		return dataSeries;
	}
);
```

packages/clay-charts/demos/index.html
julien: 👍Ok, thanks

julien:
@jbalsas ,

Thanks.
Yes, of course I'll add those changes and add docs to the site.

jbalsas:
Oh, and please, remove the IFI- reference from the commits, as that's an internal thing only. Here, we should add the `Fixes #xyz` where `xyz` is the issue in this repo.

PR #896, opened 2018-05-14
legend.contents.bindto accepts object and string | Fixes #895

### LINKED ISSUE #895: legend.contents.bindto expects an string only, but also receive HTMLElement/Object
### Current API
Currently the ClayCharts API `legend.contents.bindto` expects only one String.

Consider:
```javascript
...
legend: Config.shapeOf({
		contents: Config.shapeOf({
			bindto: Config.string(),
...
```

### Expected API behavior
Billboard.js expects an `HTMLElement` and `string`.

Consider:
```javascript
...
legend: Config.shapeOf({
		contents: Config.shapeOf({
			bindto: Config.oneOfType([Config.string(), Config.object()])
...
```
https://naver.github.io/billboard.js/release/latest/doc/Options.html#.legend

matuzalemsteles:


julien:
LGTM. Thanks @matuzalemsteles

carloslancha:
Hey @matuzalemsteles!

Every component that can be used from `soy` would have any html param as `Config.any`, since for `soy` html are not objetcs but functions ;)

matuzalemsteles:
hey @carloslancha,

Yeah, I just forgot about soy 😅, thank you for reminding me. We really should stick with `Config.any` for the Soy components.

But about this use case specifically, I think we can go with `Config.object` as it is passed only to the billboard. What do you think?

julien:
@matuzalemsteles I think case, I'd say yes for using `Config.object`

carloslancha:
@matuzalemsteles @julien it's passed **to** billboard, but can be used **from** soy, I mean:
```
{let $myHtml kind="html"}
   <div>Lorem Ipsum</div>
{/let}

{call WhateverChart.render}
   {param legend: [
      'contents': [
         'bindto': $myHtml
      ],
      ...
    ]/}
{/call}
```
Just check if in that case `WhateverChart` component receives an object and not a function and if it does, go on with `Config.object()`

Thx guys!

PR #908, opened 2018-05-17
Changes the validation of grid.x.lines.value and grid.y.lines.value to any | Fixes #907

### LINKED ISSUE #907: grid.x.lines.value and grid.y.lines.value must accept multiple values
### Summary
There seems to be several use cases for `grid.x.lines.value` and `grid.y.lines.value` both accept Date, Number, String...

As we do not know all these use cases properly and the Billboard documentation does not inform them, we can go with the `Config.any()` validation.

matuzalemsteles:
hey @julien what do you think?


julien:
@matuzalemsteles looks good to me, thanks.

PR #909, opened 2018-05-18
Fixes the data that is returned in the click action of a quick action in the management toolbar | Fixes #901

### LINKED ISSUE #901: Clicking a quick action actually performs another one
When the position of a quick action is not the same as its position inside the dropdown, clicking on it fires the wrong action.

This can only be reproduced with the changes in https://github.com/4lejandrito/liferay-portal/pull/107.

More info at my desk :)

matuzalemsteles:
I ended up removing the way we got the index of the element in favor of adding a `data-clay-index` in the DOM and retrieving it in the click action of a quick action, I'm not sure if it's the most elegant way. Please leave your thoughts.


carloslancha:
Hey @matuzalemsteles !

Another idea

- Add a syncActionItems method to create a private quickActionItems array with only those actions marked as quickActions.
- Search for the correct item in that array.

```
	syncActionItems(newVal, prevVal) {
		if (newVal && newVal !== prevVal) {
			this._quickActionItems = newVal.filter(item => item.quickAction);
		}
	}

        ...

	_handleQuickActionClicked(event) {
		const element = event.delegateTarget;
		const elementIndex = this._getQuickItemIndex(element);
		const item = this._quickActionItems[elementIndex];

		return !this.emit({
			data: {
				item: item,
			},
			name: 'actionItemClicked',
			originalEvent: event,
		});
	}
```

With this we won't need to add data-attributes to the DOM. What do you think?

Thx!

matuzalemsteles:
hey @carloslancha, sounds good. My only concern is if we are doing too much to just get the index of the quick action element.

carloslancha:
@matuzalemsteles is really improbable that a dropdown would have more than 10-20 actionItems, so it's not a big deal I guess.

@jbalsas  ?

carloslancha:

carloslancha:
Merged, thx!

Issue #910, opened 2018-05-21
Toast Alert needs to be placed at the bottom left and default auto closing time set to 8 secs


victorvalle:
Please set alert toast to be appear at the bottom left corner of the screen and 8 secs closing time for automatic close ones

carloslancha:
As far as I know the position of the toast alert is not defined by the component but by where is used.

About the closing time, where is that documented?

Thx!

jbalsas:
The position is set in `clay-css` -> https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_alerts.scss#L142-L157

About closing time, not sure where it was, but we're currently employing 5s (maybe following the old pattern?) -> https://github.com/liferay/clay/blob/master/packages/clay-alert/src/ClayAlertBase.js#L37

carloslancha:
Oh, ok, is about that class then, not about the component.

About closing time I asume we're following the "old" patter, what I mean is... where's the new one documented? 😅

victorvalle:
lexicon-system/2.0/admin/patterns/alerts/system-messages/documentation/alerts-lexicon-documentation

pat270:
@victorvalle where should the alert be positioned on small screens? At the bottom still?

PR #915, opened 2018-05-21
Update: (#910) Alerts move `.alert-notifications-fixed` to the bottom…

pat270:
… left of the screen


jbalsas:
Hey @pat270, not sure you saw this... I can merge this but we'd still need to complete the solution. See  below image for a more clear example of the problem. You can see a hint of the second alert below the danger one.

screen shot 2018-05-24 at 18 43 03

Expected layout could be:

screen shot 2018-05-24 at 18 44 12

PR #916, opened 2018-05-21
Adds a new `delayTime` option and changes the ClayToast time to 8/s | Fixes #910

### LINKED ISSUE #910: Toast Alert needs to be placed at the bottom left and default auto closing time set to 8 secs

matuzalemsteles:


jbalsas:
Hey @matuzalemsteles, thanks for this one!

I'm always confused with `delayTime` and similar properties, as it's not clear if it means that it will take `delayTime` to appear or `disappear`... Better options that come to mind could be:
- `duration`
- `autoCloseDelay`
- Make `autoClose` a `boolean` or a `number` so if `true`, we use the default value, otherwise we use the one set in `autoClose`.

I like the last one best, as it doesn't require complicating the API more. But do you think, @matuzalemsteles , @carloslancha ?

carloslancha:
@jbalsas sounds good to me. @matuzalemsteles I asked UX to define better the time stuff in alerts, please wait for it :)

packages/clay-alert/src/ClayAlertBase.js
```diff
@@ -121,6 +109,26 @@ class ClayAlertBase extends Component {
...
+	_startTimer() {
+		if (
+			this.autoClose &&
+			(this.type === 'stripe' || this.type === 'toast')
+		) {
+			if (this.delayTime) {
+				this._delayTime = this.delayTime * 1000;
+			} else {
+				this._delayTime =
+					(this.element.querySelector('a') ? 10 : 5) * 1000;
```
carloslancha: Remove this. Delay time should be no longer calculated depending on a link exists inside the alert. That should be a decision of the user.

matuzalemsteles:
hey @jbalsas sounds good to me. I'm changing `_delayTime` to` _timeToDisappear` to make it clearer 🙂

Just leaving some thoughts of what I'm doing based on the recent changes in the documentation: I'm going with the idea of Chema, if `autoClose` is passed as boolean the default value 5 is set to `stripe/embedded` and 8 to `toast`. We let the dev set when there will be actions on the messages and they set the time according to Lexicon. What do you think?

packages/clay-alert/src/ClayStripe.js
```diff
@@ -44,13 +44,14 @@ class ClayStripe extends Component {
...
+	 * Flag to `true` to indicate whether the alert should be closed
+	 * automatically with the default time.
 	 * @default false
 	 * @instance
 	 * @memberof ClayStripe
-	 * @type {?bool}
+	 * @type {?(bool|number)}
 	 */
-	autoClose: Config.bool().value(false),
+	autoClose: Config.oneOfType([Config.bool().value(false), Config.number()]),
```
carloslancha: I think this format doesn't work, should be something like
 `Config.oneOfType([Config.bool()), Config.number()]).value(false)`

packages/clay-alert/src/ClayToast.js
```diff
@@ -44,13 +44,14 @@ class ClayToast extends Component {
...
+	 * Flag to `true` to indicate whether the alert should be closed
+	 * automatically with the default time.
 	 * @default false
 	 * @instance
 	 * @memberof ClayToast
-	 * @type {?bool}
+	 * @type {?(bool|number)}
 	 */
-	autoClose: Config.bool().value(false),
+	autoClose: Config.oneOfType([Config.bool().value(false), Config.number()]),
```
carloslancha: I think this format doesn't work, should be something like
 `Config.oneOfType([Config.bool()), Config.number()]).value(false)`

packages/clay-alert/src/ClayAlertBase.js
```diff
@@ -113,11 +101,30 @@ class ClayAlertBase extends Component {
...
+	}
+
+	/**
+	 * Sets the delayTime if passed, if it does not set the default, and starts.
+	 * @private
+	 */
+	_startTimer() {
+		if (
+			this.autoClose &&
+			(this.type === 'stripe' || this.type === 'toast')
```
carloslancha: Since this is configurable and this is `ClayAlertBase` I don't think we should check the type to start the timer, only if autoClose is not null.

packages/clay-alert/src/ClayAlertBase.js
```diff
@@ -113,11 +101,30 @@ class ClayAlertBase extends Component {
...
+		) {
+			if (isNumber(this.autoClose)) {
+				this._timeToDisappear = this.autoClose * 1000;
+			} else {
+				this._timeToDisappear = (this.type === 'toast' ? 8 : 5) * 1000;
```
carloslancha: We should not have here the defaults but in the specific components (ClayToast and ClayStripe). Due in those components `autoClose` can be `bool` or `number` the default should be specified in its templates.

packages/clay-alert/src/ClayAlertBase.js
```diff
@@ -148,13 +155,14 @@ ClayAlertBase.STATE = {
...
+	 * Flag to `true` to indicate whether the alert should be closed
+	 * automatically with the default time.
 	 * @default false
 	 * @instance
 	 * @memberof ClayAlertBase
-	 * @type {?bool}
+	 * @type {?(bool|number)}
 	 */
-	autoClose: Config.bool().value(false),
+	autoClose: Config.oneOfType([Config.bool().value(false), Config.number()]),
```
carloslancha: If like I suggest we configure the default times in the specific components this property should be only of type number or null.

packages/clay-alert/src/ClayStripe.js
matuzalemsteles: Really, thanks for your observation!

packages/clay-alert/src/ClayToast.js
matuzalemsteles: Really, thanks for your observation!

packages/clay-alert/src/ClayAlertBase.js
matuzalemsteles: Yeah, I've been thinking that this was redundant... remove it. thanks!

packages/clay-alert/src/ClayAlertBase.js
matuzalemsteles: hey @carloslancha I've been thinking of changing the `autoClose` approach to `bool` or `number` in the components (ClayToast and ClayStripe).

We can make it `Config.oneOf` with the two possive options that Lexicon allows, in the case of Stripe 5 and 10 in Toast 8 and 12. I do not know if it fits in the context of `autoClose` and if it will confuse the API, but we were able to restrict the possibility of developer putting a different value than is allowed.

Disregard the names 🙂
```javascript
autoClose: Config.oneOf(['default', 'action'])
	.setter(value => {
		if (value === 'default') {
			return 8;
		}

		return 12;
	}),
```

What do you think?

cc @jbalsas

carloslancha: Hey @matuzalemsteles the idea is to try being less restrictive with this. In this case Lexicon defines that the values should be 8 and 12, but will be the user the one who pass that config to the component, we need to be covered if one team decides that for them the time must be 16.

So here `autoClose` should be `Config.oneOf([Config.bool(), Config.number()]).value(false)`

I like the approach anyway ;)

Thx!

matuzalemsteles: That makes sense now, thanks for the clarifications.

carloslancha:
Merged, thx!

PR #934, opened 2018-05-23
 Adds the possibility of passing the `id` to the nav items and emits an event when there is a click action | Fixes #933

### LINKED ISSUE #933: it should be possible to receive the clicked item in the ClayNavigationBar item click event
There are cases in the portal that use `id` to make the screen transition via js and we should emit a click action event on any item.

> Forms case

matuzalemsteles:


packages/clay-navigation-bar/src/ClayNavigationBar.js
```diff
@@ -155,6 +175,7 @@ ClayNavigationBar.STATE = {
 		Config.shapeOf({
 			active: Config.bool().value(false),
 			href: Config.string().required(),
+			id: Config.string(),
```
carloslancha: We don't need the id here

carloslancha:

carloslancha:
Resent https://github.com/liferay/clay/pull/943

PR #959, opened 2018-05-30
Add events to ClayDropdown to close on mobile and when the esq is pressed and when the iframe is clicked

matuzalemsteles:
I've been working on looking at the best approach to close the dropdown when it has a click on the iframe.

#### Approach 1
We could listen to the click events of some iframe element on the screen, but we would have problems with some use cases:

* We do not know when there will be an iframe on screen (Modal case)
* The click event will only work if the iframe domain is the same as the navigation

#### Approach 2
We could go with the simpler approach of closing the dropdown when clicking on some dropdown item but we can not due to interactions in the dropdown.

#### Approach 3
We can listen to the `blur` event in `window` and check the `activeElement` if it is an iframe so we can close the dropdown.

I'm following with the 3 approach to cover use cases.

> NOTE: I ended up changing the way we are attaching the events, instead of adding them when created, is only adding when the dropdown is open, since it only becomes necessary at that moment. It helps in the case of when we have too many dropdowns on the screen not to be listening to events unnecessarily.

### DIFF
--- a/packages/clay-dropdown/src/ClayDropdownBase.js
+++ b/packages/clay-dropdown/src/ClayDropdownBase.js
@@ -14,6 +14,8 @@ import {EventHandler} from 'metal-events';
 import itemsValidator from './items_validator';
 import templates from './ClayDropdownBase.soy.js';
 
+const KEY_CODE_ESC = 27;
+
 /**
  * Implementation of the base for Metal Clay Dropdown.
  * @extends ClayComponent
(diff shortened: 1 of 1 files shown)


carloslancha:

carloslancha:
@matuzalemsteles I'm merging this but there's an issue we need to resolve when we have time. On mobile if you open the dropdown and click and drag outside, the dropdown shouldn't close and now it does.

http://recordit.co/xXAKUaBToj

matuzalemsteles:
Ah yeah, Good observation. Thanks @carloslancha!

PR #966, opened 2018-06-01
Prevent event default on toggler click | Fixes #965

### LINKED ISSUE #965: NavigationBar Toggler navigates on click
This is related to https://issues.liferay.com/browse/LPS-81781

carloslancha:


packages/clay-navigation-bar/src/ClayNavigationBar.js
```diff
@@ -1,3 +1,4 @@
+import 'clay-button';
```
jbalsas: Is this necessary?

PR #971, opened 2018-06-05
Component autocomplete

gcmznt:
This is the first implementation of the [autocomplete text field](https://lexicondesign.io/docs/patterns/Forms/text_input_variations.html#autocomplete-text-field)


carloslancha:

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+const itemShape = {
+	label: Config.string(),
+};
+
+ClayAutocomplete.STATE = {
+	dataSource: Config.oneOfType([
+		Config.arrayOf(Config.shapeOf(itemShape)),
+		Config.string(),
+	]),
+	suggestions: Config.arrayOf(Config.shapeOf(itemShape)).value([]),
```
carloslancha: All params that are meant to be used internally and not passed from the outside must be marked as internal `.internal()` and the name has to be preceded by `_`

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+	maxSuggestions: Config.number().value(5),
+	selectedSuggestionIndex: Config.number().value(0),
+	elementClasses: Config.string(),
+	hasFocus: Config.bool().value(false),
+	id: Config.string(),
+	query: Config.string().value(''),
```
carloslancha: `query` should be always the same value of `value`, so I think we can get rid of this param

packages/clay-autocomplete/src/ClayAutocomplete.soy
```diff
@@ -0,0 +1,68 @@
...
+		<input
+			class="form-control"
+			type="text"
+			value="{$value}"
+			data-oninput="{$_handleSearch}"
+			data-onfocus="{$_handleFocus}"
+			data-onblur="{$_handleBlur}"
+			data-onkeydown="{$_handleKeydown}"
+		>
+		<ul {$listAttributes}>
```
carloslancha: Can't we use ClayDropdown here?

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+/* eslint-disable require-jsdoc */
+
+/**
+ * Metal ClayAutocomplete component.
+ */
+class ClayAutocomplete extends Component {
+	created() {
+		this.provider = new DataProvider(this.dataSource, {
+			elements: this.maxSuggestions,
+			paramName: 'query',
```
carloslancha: This should be configurable so `ClayAutocomplete` should receive a `paramName` param too.

packages/clay-autocomplete/src/DataProvider.js
```diff
@@ -0,0 +1,58 @@
...
+
+/* eslint-disable require-jsdoc */
+
+function getDataFromArray(query, data, cb) {
+	cb(
+		fuzzy
+			.filter(query, data, {
+				// pre: '<strong>',
+				// post: '</strong>',
+				extract: el => el.label,
```
carloslancha: Which object property we want to extract should be configurable.

packages/clay-autocomplete/src/DataProvider.js
```diff
@@ -0,0 +1,58 @@
...
+			})
+			.map(e => ({label: e.string}))
+	);
+}
+
+function getDataFromURL(query, url, paramName = 'q', cb) {
+	fetch(`${url}?${paramName}=${query}`)
```
carloslancha: We need to manage all kind of urls, for example, what if the passed url already has queryParams? You can use `metal-uri` https://github.com/metal/metal-plugins/tree/master/packages/metal-uri to deal with this.

carloslancha:

jbalsas:
Superseeded for now by https://github.com/liferay/clay/pull/1292 (although we might want to take a step and extract the autocomplete feature as a standalone)

Issue #1046, opened 2018-07-02
Form padding in mobile must be 16px and not 24px


victorvalle:
![image](https://user-images.githubusercontent.com/21121115/42153442-6800af92-7de3-11e8-9a13-41f8f9ba60ae.png)

PR #1047, opened 2018-07-05
(#1046) Sheet should have 1rem padding in mobile and added .sheet-footer-btn-block-sm-down for block level buttons in mobile

pat270:


marcoscv-work:

jbalsas:
Thanks @pat270! We will need to increase the minor version after this change since it is adding a new API.

PR #1095, opened 2018-08-15
ClayCSS: (Fixes #1094) Atlas Nav removed unused variables `$nav-colla…

### LINKED ISSUE #1094: ClayCSS: Nav $nav-collapse-icon-closed and $nav-collapse-icon-open arent used anywhere

pat270:
…pse-icon-closed`, `$nav-collapse-icon-open`

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_navs.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_navs.scss
@@ -5,9 +5,6 @@ $nav-link-disabled-color: lighten(saturate(adjust-hue($secondary, -3), 5.39), 23
 $nav-link-padding-x: 1rem !default; // 16px
 $nav-link-padding-y: 0.625rem !default; // 10px
 
-$nav-collapse-icon-closed: "\f0da" !default;
-$nav-collapse-icon-open: "\f0d7" !default;
-
 // Nav Nested
 
 $nav-nested-spacer-x: 1rem !default; // 16px


marcoscv-work:
I think we definitely need some policy to deprecate/eliminate vars/mixins in order to do not make breaking changes.

Maybe we can move them o label them with comments or warns.

@pat270 @jbalsas what do you guys think?

jbalsas:
Hey @marcoscv-work, @pat270, yes please, let's do that! 👍 

We can't introduce breaking changes in `2.x.x`, and we should stick with this version for a while. My proposal would be:
- Create a `deprecated` folder mimicking the structure of `variables`, for example
- Move the deprecated variable to the corresponding file, in a section with a `// Deprecated` header
- Add a warning so if anyone uses it the compiler logs a `This variable is deprecated as of 2.12.1 and will likely be removed in 3.0.0. You should use $foo instead.`

Then, when we get to move to `3.0.0`, we can get rid of everything under `deprecated`... or at least consider it.

Having it in a separate folder and in a different `import 'variables/deprecated'` might help if someone is doing a custom build to remove the extra boilerplate.

What do you think?

pat270:
We can do something like that. The reason I submitted these is because they're not being used in Clay or Portal. They are left overs from LexiconCSS that I forgot to remove.

jbalsas:
Hey @pat270, that's fine, people could still be using those outside of clay and portal, in their own themes. Let's try to be stricter from now on so we don't unknowingly hurt our beloved devs :)

Let's use this low-risk removals to put in motion this deprecation workflow and start getting accustomed with it!

jbalsas:
Closing until we fix this to prevent merging it accidentally

PR #1098, opened 2018-08-16
Avoid empty tooltip

boton:


carloslancha:

carloslancha:
Thinking about this maybe is better to simple improve our selectors default value to not accept elements titles with empty title or data-title, so the user could know how his custom selector should work.

PR #1106, opened 2018-08-26
Adds Clay examples with React, Angular, Vue...

matuzalemsteles:
I'm moving the examples from Clay from https://github.com/matuzalemsteles/clay-examples to here, so we can keep.

I think it's important to have these examples so people know how to use it in other frameworks since we'll start investing in WebComponents.

Since we will start investing in WebComponents we may have to start evaluating the APIs we have to be compatible with these frameworks and not cause problems. The `style` API causes a lot of problems with WebComponents; these frameworks evaluate their entries to follow the HTML attribute standards. And we ended up using a lot in our components.


matuzalemsteles:
hey @jbalsas, Any way to bypass Codacy for this folder? I do not know if it would be necessary to check there.

jbalsas:
Hey @matuzalemsteles, how hard would it be to follow the tslint guidelines? I think we might as well produce nice code even for our examples... what do you think?

I'm sure you can simply add a `.tslint` file with specific configuration if you don't want to go around fixing those.

Could you give that a try?

matuzalemsteles:
Sure, no problem.

examples/angular-with-clay/tslint.json
```diff
@@ -0,0 +1,145 @@
...
+      true,
+      "check-open-brace",
+      "check-catch",
+      "check-else",
+      "check-whitespace"
+    ],
+    "prefer-const": true,
+    "quotemark": [
+      true,
+      "single"
```
matuzalemsteles: hey @jbalsas, I fix the formatting but Codacy still keeps giving error. It asks to change `' should be "` but the `.tslint` settings are as `single` this is strange, could you check where it is taking these settings from on Codacy?

jbalsas: Hey @matuzalemsteles, yeah, I'll take a look tomorrow to get this in one way or an other. Sorry for bothering with this...

Since we never really agreed on adding it (I did it on my own), we might as well disable it for now and add it on our own terms. What do you think?

/cc @carloslancha

matuzalemsteles: hey @jbalsas no problem. Maybe it's just some configuration for `typescript` I do not quite know how it works... But I think we can reevaluate with our rules...

jbalsas:
Pull request submitted to https://github.com/liferay/clay/pull/1140. [See changes here.](https://github.com/liferay/clay/compare/603bb04fe~1...603bb04fe) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #1116, opened 2018-08-28
Add id parameter to clay-chart templates | Fixes #1115

### LINKED ISSUE #1115: [clay-charts] Add 'id' parameter to template
Currently the .soy template in clay-charts does not allow receiving an `id` parameter and we need to fix that.

julien:


packages/clay-charts/src/Chart.soy
```diff
@@ -5,8 +5,16 @@
  */
 {template .render}
 	{@param? _loading:  bool}
+	{@param? id: string}
 
-	<div class="metal-chart-container">
+  {let $attrs kind="attributes"}
```
jbalsas: Looks like a wrong indentation here... 🤔

packages/clay-charts/src/Geomap.soy
```diff
@@ -6,5 +6,23 @@
 {template .render}
 	{@param? _height: string}
 	{@param? _width: string}
-	<div style="width:{$_width};height:{$_height}"></div>
+	{@param? id: string}
+
+  {let $attributes kind="attributes"}
```
jbalsas: Wrong indentation?

julien: Yes, I'll fix that
Thanks

packages/clay-charts/src/Chart.soy
julien: Yes

PR #1149, opened 2018-08-31
Focus dropdown on open and make the use of ClayPortal configurable | Fixes #1141

### LINKED ISSUE #1141: Dropdown menu does not work correctly on keyboard tabulation
Dropdown items are not accessible using tabs

dropdown

We need to tab through all dropdown items.

Bootstrap is a good about we need:
https://getbootstrap.com/docs/4.1/components/dropdowns/#single-button

carloslancha:


matuzalemsteles:

packages/clay-dropdown/src/ClayDropdown.soy
```diff
@@ -64,5 +65,6 @@
 		{param triggerClasses: $triggerClasses /}
 		{param triggerSize: $triggerSize /}
 		{param triggerTitle: $triggerTitle /}
+		{param usePortal: $usePortal ?: true /}
```
matuzalemsteles: hey @carloslancha can you do the same for `ClayActionsDropdown` and `ClayCreationMenuDropdown`?

packages/clay-dropdown/src/ClayDropdownBase.js
```diff
@@ -49,6 +60,30 @@ class ClayDropdownBase extends ClayComponent {
...
+
+			if (alignElement) {
+				this._alignedPosition = Align.align(
+					bodyElement,
+					alignElement,
+					Align.BottomLeft
+				);
+			}
+
+			bodyElement.querySelector('li a, li input').focus();
```
matuzalemsteles: hey @carloslancha we want to cover the `li button` cases too? If not, we have to add a condition if do not find anything inside.

packages/clay-dropdown/src/ClayDropdownBase.js
```diff
@@ -21,18 +21,29 @@ const KEY_CODE_ESC = 27;
...
+		}
 	}
 
 	/**
 	 * @inheritDoc
 	 */
-	created() {
-		this._eventHandler = new EventHandler();
+	rendered() {
+		if (!this.usePortal) {
```
matuzalemsteles: Just a suggestion: `rendered` is a hot path to updates, when `renderers` are always going to fall here (it will not align because of `this.expand`), but maybe we can improve this by calling in `syncExpanded`. What do you think?

packages/clay-dropdown/src/ClayDropdownBase.js
carloslancha: I think we don't have `li button` cases. Anyway, I'm extracting this to a method that could be overwritten to fit user puroposes and adding a safe condition . Thx!

carloslancha: Nvm, we're not adding this feature in this way, I'm opening a new issue for this.

PR #1160, opened 2018-09-03
Fixes #1153 - Make use of ClayPortal configurable in ClayDropdown

### LINKED ISSUE #1153: Make use of ClayPortal in ClayDropdown configurable
Move dropdown to portal layer can generate semantic/SEO problems because we are breaking the nested items so would be great if we could specify in component dropdown or component which use dropdown if they want to use portal layer or not.

image

carloslancha:


packages/clay-dropdown/src/ClayDropdownBase.js
```diff
@@ -49,6 +51,28 @@ class ClayDropdownBase extends ClayComponent {
...
+				this._alignElementSelector
+			);
+
+			if (alignElement) {
+				let bodyElement = this._getMenuElement();
+
+				this._alignedPosition = Align.align(
+					bodyElement,
+					alignElement,
+					Align.BottomLeft
```
matuzalemsteles: hey @carloslancha, can you add our recent change here? `this.preferredAlign`.

We also have to remove `_handleRenderedPortal` since we are no longer using it.

carloslancha: Something happened in the rebase of this. I'll resend, thx.

PR #1173, opened 2018-09-06
 Keep the expanded one up to date with ClayDropdownBase | Fixes #1172

### LINKED ISSUE #1172: ClayActionsDropdown, ClayCreationMenuDropdown and ClayDropdown do not keep the expanded updated with ClayDropdownBase

matuzalemsteles:


carloslancha:
Hey @matuzalemsteles 

I think it would be better if we use the same pattern used in ClayAlert. I mean, instead listening to the `expandedChange` event and update the parent component state with the value of the child component we should do:

- ClayDropdownBase `toggle` method should only emit a `toggle` event.

- ClayDropdown, ClayActionDropdown and ClayCreationMenuDropdown should also have a public `toggle` method that emits a `toggle` event.

- ClayDropdownBase, ClayDropdown, ClayActionDropdown and ClayCreationMenuDropdown should have a default listener to the `toggle` event to do what `toggle` method does right now.

- ClayDropdown, ClayActionDropdown and ClayCreationMenuDropdown should listen to ClayDropdownBase `toggle` event and fire `toggle` method.

Let me know if you have any thoughts or doubts about this 😄

matuzalemsteles:
hey @carloslancha

This sounds good, but I just do not like to go with this approach because we will end up duplicating the logic of having a `toggle` and `defaultToggle`... in each component.

We could abstract this into one class for everyone, even the DropdownBase, so we can remove this duplication. This would be solved easy with an HOC, but it does not work well with Soy... what do you think?

jbalsas:
Hey @matuzalemsteles, you're totally right about this. On the other side, [having clear patterns](https://twitter.com/jlongster/status/1041887130287521792) (even if duplicated) is actually better than a wrong abstraction.

I think for now it's fine to work on solidifying our patterns. Emitting events and having a `default` function do the work is for now working, so let's keep implementing it that way. 

In the future, we can explore how to turn that pattern into something else as we improve our knowledge of how the library is used and composed.

carloslancha:

matuzalemsteles:
Hey @jbalsas, Yes I agree that standards are important and we should stick to it.

But just to clarify, I do not want to remove the emission of events from the components, I find a very promising pattern of Metal ... Let me state what I mean in abstracting in the **class**.

```javascript
class WithToggle extends ClayComponent {
	attached() {
		this.addListener('toggle', this._defaultToggle, true);
	}

	_defaultToggle() {
		this.expanded = !this.expanded;
	}

	toggle() {
		return !this.emit({
			name: 'toggle',
		});
	}
}
```

So we can get everyone to extend this `class` and keep Carlos proposal up.

carloslancha:
Hey @matuzalemsteles that's a great idea, anyway I still think that the duplicated code is not long or complex enough to go with an abstraction and having to extend here. I'd really go with the duplication in this case ;)

matuzalemsteles:
hey @carloslancha no problem, we can move on with this.

matuzalemsteles:
hey @carloslancha, What do you think of adding a flag to **ClayDropdownBase** to say when `expanded` is controlled or uncontrolled? to remove the two renderings that will happen when ClayDropdown, ClayActionDropdown and ClayCreationMenuDropdown update their `expanded`.

So where we use only ClayDropdownBase we can tell him to control the `expanded`.

carloslancha:
If you prevent the toggle event in `ClayDropdown`, `ClayActionDropdown` and `ClayCreationMenuDropdown` the default listener in `ClayDropdownBase` will no be fired, so it won't be two renders I guess

carloslancha:

carloslancha:
Resent https://github.com/liferay/clay/pull/1191

PR #1181, opened 2018-09-12
Change the default value of actionItems to [] | Fixes #1178

### LINKED ISSUE #1178: ManagementToolbar ActionItems show if an empty array is passed
In the management toolbar, if an empty array is passed as `actionItems`, an empty dropdown is shown. Would it be possible to control this case?

```javascript
ManagementToolbar.STATE = {
    [...]
    actionItems: actionItemsValidator.value([]),
    [...]
};
```

screen shot 2018-09-11 at 13 22 21

matuzalemsteles:


packages/clay-management-toolbar/src/ClayManagementToolbar.soy
```diff
@@ -251,7 +251,9 @@
 				</li>
 			</ul>
 
-			{if $actionItems or $showInfoButton}
+			{let $actionItemsIsEmpty: length($actionItems) > 0 /}
```
jbalsas: Wouldn't this be actually the opposite? `hasActionItems`?

packages/clay-management-toolbar/src/ClayManagementToolbar.soy
```diff
@@ -200,20 +200,12 @@
  * This renders the component's active content.
  */
 {template .active}
+	{@param actionItems: list<?>}
```
matuzalemsteles: hey @jbalsas, I'm removing the typing here so that at https://github.com/liferay/clay/pull/1181/files#diff-54445d58e09b53d03b6d0558828824f6R136 can set the default value if not passed.

Since we are typing `actionItems` above, we can ensure that the values are passed correctly.

carloslancha:

PR #1188, opened 2018-09-19
Fixes #1105 - Update styles and markup for results bar

### LINKED ISSUE #1105: Styles and markup for results bar
We need to create the markup for the new management toolbar results bar.

Docs: https://docs.google.com/document/d/1R2l3wfdPQsoVSX2QDpmPooz52D6ei61PiHVMnI5wI90/edit

@pat270  Could you please take a look? Thx!

carloslancha:


matuzalemsteles:

packages/clay-management-toolbar/src/ClayResultsBar.soy
```diff
@@ -80,35 +79,58 @@
...
-					{param ref: 'label' + index($filterLabel) /}
-					{param size: 'lg' /}
-					{param spritemap: $spritemap /}
-				{/call}
-			{/foreach}
+			{call ClayLabel.render}
+				{param closeable: $closeable /}
+				{param elementClasses: 'component-label tbar-label' /}
+				{param events: ['close': $_handleFilterLabelCloseClicked] /}
+				{param label: $label /}
```
matuzalemsteles: I think we have to wrap this in a `.tbar-section`.

```soy
{param label kind="text"}
    <div class="label-section">{$label}</div>
{/param}
```

carloslancha: If you mean in a `label-section` I didn't add it 'cause that class does nothing and changes how the label component is supposed to work.

pat270: An empty block level element should be there because of `label-item` uses `flex-direction: column` and causes anything that is a direct descendant to become a block level element. As soon as you try and set elements inline, something like `Category: <strong>Productivity</strong>`, you will see it stacks by default.

The class does nothing, but gives something for devs to hook on to and I think it's better than an empty div.

matuzalemsteles: hey @pat270. Is there any problem for us to go without it?

pat270: @matuzalemsteles No problem without, but I see stuff like `.some-class > div` in portal and don't like it.

matuzalemsteles: Sorry @pat270 I had not seen your previous comment when I posted it 😅. (I think it did not update for me, so I deleted it...).

So we'd better get involved with the `.label-section`, so we do not give the possibility that others might do `.some-class> div` to try to fix this.

carloslancha: Ok guys, I'll add it tomorrow, thx!

matuzalemsteles:
Merged, thanks!

PR #1224, opened 2018-10-07
Remove redundant aria-label in clayButton

### LINKED ISSUE #1174: Redundant aria-label in clayButton
I think that setting the `aria-label` attribute's value according to the `label` param if `aria-label` param is blank https://github.com/liferay/clay/blob/master/packages/clay-button/src/ClayButton.soy#L52 creates duplicate information, because if a `label` set the content, it will be set both for `aria-label` and content will be read by screen readers twice.

darekkay:
If a `button` contains a label, `aria-label` should not be set.
If there is no `label` or `ariaLabel` present, use the `icon` as aria-label.

Closes #1174

### DIFF
--- a/packages/clay-button/src/ClayButton.soy
+++ b/packages/clay-button/src/ClayButton.soy
@@ -48,9 +48,7 @@
 
 		{if $ariaLabel}
 			aria-label="{$ariaLabel}"
-		{elseif $label}
-			aria-label="{$label}"
-		{elseif $icon}
+		{elseif $icon and not $label}
 			aria-label="{$icon}"
 		{/if}
 


matuzalemsteles:
hey @darekkay, first of all sorry for the delay and thank you very much for your contribution, we really appreciate it, feel free to get more involved. This sounds good to me but I want to hear from Marcos too so we can be sure we can merge this.

marcoscv-work:
Hi guys, sorry for the delay, I was in Devcon (Amsterdam) last week.

It looks great, but I think we are evaluating the possibility to remove title from icons, so we will not able tu use icons like a label, that's the reason I think in those cases we will need or aria-label, o an a hidden label with something like:

`<span class="sr-only">Button label</span>`

In this way I think we will need a way to manage aria-labels or hidden labels in icon-buttons

marcoscv-work:
We added the title to the icons because an automatic code validation required, but it never made sense because our icons should be totally hidden, they always are label visual helper. they must be totally expendable. It's like a decoration image with alt="", its meaning is nothing.

So I think we can remove all description labels and text from icons because our icons will always be accompanied with text in a link or inside a button, the case of button without text is the one making problems, so I propose make in a button the label or text required and hidden by class sr-only if is a type icon-button.

We have some issues and work lines open, so I propose create a research task considering UX / technical point of view and make as a result a final decision.

Probably the Lexicon icons/icon buttons definition should be more concrete: https://lexicondesign.io/docs/patterns/icons.html
https://lexicondesign.io/docs/patterns/buttons.html (icon buttons section)

jbalsas:
Pull request submitted to https://github.com/liferay/clay/pull/1325. <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

jbalsas:
Hey @matuzalemsteles, I've ammended the commit message and resent to https://github.com/liferay/clay/pull/1325 to see if CI is still happy with this before merging.

👌

PR #1255, opened 2018-10-24
Allow overriding of highlight color

sbaechler:
Currently there is no easy way of overriding the highlight color. I extracted the default mixin parameters to sass variables.


pat270:
New pull: https://github.com/liferay/clay/pull/1278

PR #1267, opened 2018-10-27
Fixes #1266 Reflow sidenav to prevent issue in IE11

### LINKED ISSUE #1266: Clay: IE11 SideNav does not fully close
https://issues.liferay.com/browse/LPS-86163

In IE11, there is a bug that does not allow the SideNav to fully close. This can be reproducible in Clay and Lexicon but it is very inconsistent. If you use 'enter' you can open and close the SideNav easily. It seems that reflowing the content will be able to fix this issue. 

If more information is needed please let me know.

fortunatomaldonado:
Fixes https://github.com/liferay/clay/issues/1266
https://issues.liferay.com/browse/LPS-86163

Being able to reflow the content will fix the issue in IE11. I was able to find examples of reflowing content which can be found [here](https://github.com/liferay/clay/blob/1.x/src/js/bootstrap/tab.js#L87). I was unable to get the reflow to work using offsetWidth and found using .css() worked. Adding the reflow to _renderUI will make sure this issue will not be reproducible for any sidenav created.
Let me know if there is a better way to cause the reflow and if there are any questions or comments.

Thank you.

### DIFF
--- a/packages/clay-css/src/js/side-navigation.js
+++ b/packages/clay-css/src/js/side-navigation.js
@@ -934,6 +934,8 @@
 
 				instance._renderNav();
 			}
+
+			container.css({display: ''});
 		},
 
 		_setScreenSize: function() {


pat270:
New pull: https://github.com/liferay/clay/pull/1271

Hey @fortunatomaldonado,
There needs to be some additional reflows done on `.control-menu` and `#wrapper` wherever we are instantiating the side navigation in DXP to completely fix the issue.

PR #1281, opened 2018-10-31
Fixes #1280 - react-clay-charts

### LINKED ISSUE #1280: Create clay-charts for React
For the use case of moving from metal-jsx to react, it'd be helpful to have a react version of clay-charts.

Doing this will require us to have 3 chart related packages

- `clay-charts` - the original metal soy/jsx versions
- `clay-charts-react` - the react version
- `clay-charts-shared` - the shared logic between the metal and react version

bryceosterhaus:


julien:
Thanks @bryceosterhaus!

PR #1290, opened 2018-11-06
Fixes 1289 | Fix installation issue with node v10.13.0

julien:


carloslancha:

julien:
> I don't thing regen all snapshots is the best solution. We need to find out why that changed.

These are generated files, if the tool that generated has changed it's format, I think we don't have many other options, but I may be wrong. I'll have a look and see what changed.

julien:
@carloslancha it looks like  https://github.com/liferay/clay/pull/1302 has a good explanation of what is going on, I'll close this one to avoid (more) confusion.

Thanks

PR #1292, opened 2018-11-07
Fixes #1285 - API proposal to ClayLabelsInputField component

### LINKED ISSUE #1285: Create ClayMultiSelect Component
### Initial implementation

- [x] Send API Proposal #1292 
- [x] Send final component #1351

matuzalemsteles:
> 🚨 This PR is not to be merged. Do not stick to the code now, but the idea behind. I will still send another PR to evaluate the implementation and our good practices, once this idea is valid.

## Disclaimer

I'm sending the API proposal to the new ClayLabelsInputField component. View the documents and Invision to analyze the use cases.

- [Invision](https://liferay.invisionapp.com/share/UKE9RIJDY#/screens/261808490_Label-Input-Field)
- [Doc](https://docs.google.com/document/d/1cIx6RkWNbnho_fOSR__3Fp5alIBE_ObhO-jecfXlA9I/edit?usp=sharing)

## Proposal

The component does not manipulate the data or changes internal states, it only triggers events of what is happening and letting the developer decide what to do with the information or validate the information for example. but always respecting the principles that the Lexicon is predicting.

### Events

The developer has the freedom to make decisions using the "life cycles" of the component, some of them are:

- **labelAdded** - The event is triggered when the user interacts with the input or dropdown by pressing comma or enter to create a new label or by clicking a dropdown item.
  - It's a good time for the developer to create their validation rules to add the label.
- **labelRemoved** - The event is triggered when the user interacts with the labels, removing them through the backspace or clicking the label's close button.
- **labelFocused** - The event is triggered when any label is focused.
- **queryChange** - The event is triggered when the user is typing in the input.
  - This is a good time to evaluate autocomplete or provide suggestions based on your criteria.

> I'm still not sure about the names of the events, if you think of something please comment.

### Helper methods

The component provides some auxiliary methods so that it can be used with events.

- **filterAutocomplete** - Filters an `Array<string>` that corresponds to `query`.
- **clearInput**- It just clears the input.

### Extension points

Lexicon predicts that developers can customize dropdown items with a list of contacts, phones ... or customize the labels for example. You can start by creating a variation using soy exclusivity `deltemplate` to achieve the expected results.

- **ClayInputField.Label**
- **ClayInputField.ItemDropdown**

> Both `deltemplates` are given the corresponding `data` parameter. causing the developer to decide the data API.

As the markup is still not 100% ready I added some css on the demo page to get closer to the Lexicon just for testing purposes, this will change when the markup is ready.

### DIFF
--- a/packages/clay-labels-input-field/src/ClayLabelsInputField.js
+++ b/packages/clay-labels-input-field/src/ClayLabelsInputField.js
@@ -1,0 +1,422 @@
+import 'clay-button';
+import 'clay-label';
+import {Config} from 'metal-state';
+import {EventHandler} from 'metal-events';
+import {match} from './utils';
+import ClayComponent from 'clay-component';
+import defineWebComponent from 'metal-web-component';
+import Soy from 'metal-soy';
+
+import templates from './ClayLabelsInputField.soy.js';
+
+/**
+ * Metal ClayLabelsInputField component.
+ * @extends ClayComponent
+ */
+class ClayLabelsInputField extends ClayComponent {
+	/**
+	 * Get the value typed in the input or through the filtered items.
+	 * @param {!Event} event
+	 * @protected
+	 * @return {string}
+	 */
+	_getLabelFromEvent(event) {
+		const {tagName} = event.target;
+
+		if (tagName === 'INPUT') {
+			return event.target.value.toLowerCase().replace(',', '');
+		} else if (tagName === 'A') {
+			const index = event.target.getAttribute('data-dropdown-item-index');
+			const element = this.filteredItems[Number(index)];
+
+			return element.originalString.toLowerCase();
+		}
+	}
+
+	/**
+	 * Continues the propagation of the Button clicked event.
+	 * @param {!Event} event
+	 * @protected
+	 * @return {Boolean} If the event has been prevented or not.
+	 */
+	_handleButtonClicked(event) {
+		return !this.emit({
+			name: 'buttonClicked',
+			originalEvent: event,
+		});
+	}
+
+	/**
+	 * Handle the click on the close label button.
+	 * @param {!Object} data
+	 * @protected
+	 * @return {Boolean} If the event has been prevented or not.
+	 */
+	_handleCloseButtonClick(data) {
+		return this._handleLabelRemoved(data.target.element);
+	}
+
+	/**
+	 * Handle the click on the dropdown item and the propagation of the labelAdded event.
+	 * @param {!Event} event
+	 * @protected
+	 * @return {Boolean} If the event has been prevented or not.
+	 */
+	_handleDropdownItemClick(event) {
+		return this._handleLabelAdded(event);
+	}
+
+	/**
+	 * Continues the propagation of the labelAdded event.
+	 * @param {!Event} event
+	 * @protected
+	 * @return {Boolean} If the event has been prevented or not.
+	 */
+	_handleLabelAdded(event) {
+		const label = this._getLabelFromEvent(event);
+
+		if (
+			label.trim() &&
+			!this.selectedLabels
+				.find((labelSelected) => labelSelected.label === label)
+		) {
+			return !this.emit({
+				data: {
+					value: label,
+				},
+				name: 'labelAdded',
+				originalEvent: event,
+			});
+		} else {
+			this.refs.input.value = label;
+		}
+	}
+
+	/**
+	 * Handle the focus on the selected labels and triggers the focused label event.
+	 * @param {!Event} event
+	 * @param {?Boolean} direction
+	 * @protected
+	 * @return {Boolean} If the event has been prevented or not.
+	 */
+	_handleLabelFocus(event, direction) {
+		if (this.selectedLabels.length) {
+			const {formGroupInput} = this.refs;
+			const items = formGroupInput.querySelectorAll('span[id="item-tag"]');
+
+			if (this._labelFocused) {
+				const index = this._labelFocused.getAttribute('data-tag');
+				const condition = direction
+					? Number(index) - 1
+					: Number(index) + 1;
+
+				if (condition > (items.length - 1)) {
+					this.refs.input.focus();
+					this._removeFocusedLabel();
+					return false;
+				} else if (condition >= 0) {
+					this._labelFocused.classList.remove('label-focus');
+					this._labelFocused = items[condition];
+					this._labelFocused.classList.add('label-focus');
+				} else {
+					return false;
+				}
+			} else {
+				this._labelFocused = items[items.length - 1];
+				this._labelFocused.classList.toggle('label-focus');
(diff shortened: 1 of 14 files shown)


packages/clay-labels-input-field/src/ClayLabelsInputField.js
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	/**
+	 * Helper method to filter a list based on a string.
+	 * @param {!Array<string>} data
+	 * @param {!string} query
+	 * @public
+	 * @return {Array} A list of items containing the corresponding characters
+	 */
+	filterAutocomplete(data, query) {
```
jbalsas: Let's extract this to something like `clay-dataprovider`. See [DataProvider](https://github.com/liferay/clay/pull/971/files#diff-fc5f1ca3d00e2b11a1a6029e3b277473R24) for some reference or [Kendo DataSource](https://demos.telerik.com/kendo-ui/datasource/index).

We should support both local and remote data. See also how we started doing it in [clay-charts-shared](https://github.com/liferay/clay/blob/master/packages/clay-charts-shared/src/data.js) and [ClayBase](https://github.com/liferay/clay/blob/master/packages/clay-charts/src/ChartBase.js)

packages/clay-labels-input-field/src/ClayLabelsInputField.soy
```diff
@@ -0,0 +1,171 @@
...
+		"
+
+		{if $id}
+			id="{$id}"
+		{/if}
+	{/let}
+
+	<div {$attributes}>
+		{if $label}
+			<label>{$label}</label>
```
jbalsas: I think `<label>` here will be missing a11y attributes like `for`

packages/clay-labels-input-field/src/ClayLabelsInputField.soy
```diff
@@ -0,0 +1,171 @@
...
+	{@param? _handleOnKeydown: any}
+	{@param? _removeFocusedLabel: any}
+	{@param? contentRenderer: string}
+	{@param? filteredItems: list<?>}
+	{@param? selectedLabels: list<?>}
+
+	<div class="input-group input-group-stacked-sm-down">
+		<div class="input-group-item">
+			<div data-keydown="{$_handleOnKeydown}" ref="formGroupInput" class="form-control dropdown-full form-control-tag-group">
+				{if $selectedLabels}
```
jbalsas: For easier integration, it might be good to have hidden input fields here with the label values. That way the component will be compatible with standard form submission.

We can't just select labels, they need to have values associated with them, so our input data should look more like:

```
data: [{
    value: '123',
    label: 'Element 1'
}, {
    value: '456',
    label: 'Element 2'
}]
```

On the UI, we should show the labels, but we would be updating and keeping a list of `<input type="hidden">` with the values of the selected items. There should also be a way to configure the name of the input so the server will be able to retrieve the data as a native array as expected.

packages/clay-labels-input-field/src/ClayLabelsInputField.soy
```diff
@@ -0,0 +1,171 @@
...
+	{@param spritemap: string}
+	{@param? _handleButtonClicked: any}
+	{@param? _handleCloseButtonClick: any}
+	{@param? _handleDropdownItemClick: any}
+	{@param? _handleOnInput: any}
+	{@param? _handleOnKeydown: any}
+	{@param? _removeFocusedLabel: any}
+	{@param? contentRenderer: string}
+	{@param? filteredItems: list<?>}
+	{@param? selectedLabels: list<?>}
```
jbalsas: Let's call this `selectedItems`?

packages/clay-labels-input-field/src/utils.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * If `query` matches to `string`, returns an object with the
+ * corresponding characters. If there is no match, return null.
+ * @param {!String} query
+ * @param {!String} string
+ * @return {null|Object}
+ */
+export const match = (query, string) => {
```
jbalsas: @gcmznt used [fuzzy](https://github.com/liferay/clay/pull/971/files#diff-fc5f1ca3d00e2b11a1a6029e3b277473R3). Do you think that might be an option so we don't have to code and maintain this?

victorvalle:
@jbalsas multiselect sounds good to me. It could be confused with other old pattern but we could even use this to replace that one.

jbalsas:
Maybe we simply call it `ClaySelect` like https://react-select.com/home?

packages/clay-labels-input-field/src/utils.js
matuzalemsteles: hey @jbalsas, Yeah, I thought of using it here, but I had trouble getting to the Lexicon markup result, with the fuzzy it concatenates in the characters that are match the `strong` tag, but we're having problems with the parameters that they accept HTML in Metal soy (https://github.com/metal/metal.js/issues/381). And this utility is the same fuzzy implementation, but without the option to concatenate some markup, just list which items are active... so we can get to the desired https://github.com/liferay/clay/pull/1292/files#diff-087a55086d39969ce453305b7bca6fcfR160

packages/clay-labels-input-field/src/ClayLabelsInputField.soy
matuzalemsteles: I had put them as items, but I had decided to change thinking it was not a good option 😅. Referring to items is better. Thanks Chema!

packages/clay-labels-input-field/src/ClayLabelsInputField.js
matuzalemsteles: I was really thinking about not dealing with the DataProvider here 😅, I'll separate a task to just deal with it. Thanks for the examples.

matuzalemsteles:
I like the idea of going with `Picker` but I think we would have to follow this line with the rest of the other components that are part of this group. Does it make sense?

packages/clay-labels-input-field/src/utils.js
jbalsas: Sounds good for now, then!

packages/clay-labels-input-field/src/ClayLabelsInputField.soy
jbalsas: 👍

packages/clay-labels-input-field/src/ClayLabelsInputField.js
jbalsas: Okay, let's get this component through and open an issue to abstract that!

victorvalle:
I prefer mutli select when it can select multiple things.

jbalsas:
Let's go with `ClayMultiSelect`, then! 👌

matuzalemsteles:
I'm closing this to send the final changes in another PR.

PR #1294, opened 2018-11-08
Fixes #1293 - Add tests for clay-charts-react

### LINKED ISSUE #1293: Unable to add tests for clay-charts-react
There are some issues with conflicting versions of babel and jest. So we should fall back to the same version that clay-charts is using.

bryceosterhaus:
There were a bunch of issues getting jest/yarn/lerna to work with different versions of babel so I ended up just falling back to the same babel version for all the packages. This allowed em to write some basic render tests for clay-charts-react


matuzalemsteles:
hey @bryceosterhaus, I think we can try to update the babel version to 7.x for all packages and see what happens. We have to analyze this, I had seen some cases where the babel transformed some code to specifications with use of `Symbol` for example, forcing us to use a polyfill for this 🤷‍♂️.

matuzalemsteles:

matuzalemsteles:
hey @bryceosterhaus, It would be something interesting then we think of a way to create tests to ensure the consistency of APIs that are passed to the charts in react and metal.

jbalsas:
Hey @bryceosterhaus, I think this is conflicting with our fix for `Node 10` from yesterday. Can you resolve those and push so we can merge?

This is looking cool! 👏

bryceosterhaus:
Updated!

PR #1296, opened 2018-11-09
Fixes #1242 - Use aria-label to announce icon name instead of title

### LINKED ISSUE #1242: IE11 - <a> element is hard to click with child lexicon-icon

EDIT : 

pat270:
Wrapped SVG Icons with a `span` tag and aria attributes so screen readers can announce icon names without the use of the `title` element inside the SVG (title actually didn't work in MacOS VoiceOver). This prevents the default browser tooltip from displaying in https://issues.liferay.com/browse/LPS-81456 while satisfying https://issues.liferay.com/browse/LPS-68726 and https://issues.liferay.com/browse/LPS-86160.


matuzalemsteles:
hey @pat270, I ran the CI again to check if it was just a flaky test. But it looks like some snapshots have not been updated. Could you also update the isomorphic snapshots?

packages/clay-icon/src/ClayIcon.soy
```diff
@@ -32,8 +44,12 @@
 		{/if}
 	{/let}
 
-	<svg {$attributes}>
-		<title>{$title ?: $id ?: $symbol}</title>
-		<use xlink:href="{$spritemap}#{$symbol}" />
-	</svg>
+	<span {$containerAttributes}>
+		<svg {$attributes}>
+			{if $title}
```
carloslancha: If we don't need title inside `svg` for accessibility anymore I'd remove it.

packages/clay-icon/src/ClayIcon.soy
```diff
@@ -6,11 +6,23 @@
 {template .render}
 	{@param spritemap: string}
 	{@param symbol: string}
+	{@param? ariaLabel: string}
```
carloslancha: Why can't we use just title as the ariaLabel?

packages/clay-icon/src/ClayIcon.soy
```diff
@@ -6,11 +6,23 @@
...
+	{@param? decorative: bool}
 	{@param? elementClasses: string}
 	{@param? focusable: bool}
 	{@param? id: string}
 	{@param? title: string}
 
+	{let $containerAttributes kind="attributes"}
```
carloslancha: To follow the same patter the attributes variable of the first element is called just `attributes`, so this one should be `attributes` and the other `svgAttributes`.

packages/clay-icon/src/ClayIcon.soy
```diff
@@ -6,11 +6,23 @@
...
+		{if $decorative}
+			aria-hidden="true"
+		{/if}
+
+		aria-label="{$ariaLabel ?: $symbol}"
+		class="lexicon-icon-container"
```
carloslancha: `elementClasses` should be applied to the first element of the component, so here.

packages/clay-icon/src/ClayIcon.soy
pat270: Wouldn't removing the title break the api for those that are already using ClayIcon?

packages/clay-icon/src/ClayIcon.soy
pat270: Using `title` instead of `ariaLabel` is confusing because if I'm passing a value to the title attribute, I fully expect it to be the SVG's title element.

packages/clay-icon/src/ClayIcon.soy
pat270: Wouldn't this be a breaking change since we are moving classes that were once on the SVG to the span?

carloslancha: Pattern we're following places the elementClasses in the first element always. IMHO markup change is already a breaking change so let's try to be consistent at least with the pattern :)

pat270:
@carloslancha We should keep the span around the svg icon separate from other containers. The alignment of the icons will be off in some places like `dropdown-item-indicator-end` and `collapse-icon-closed`.

carloslancha:
> @carloslancha We should keep the span around the svg icon separate from other containers. The alignment of the icons will be off in some places like `dropdown-item-indicator-end` and `collapse-icon-closed`.

I see, ok

pat270:
@carloslancha made most of the changes you requested, let me know if you see anything else.

packages/clay-icon/src/ClayIcon.soy
carloslancha: I don't mean removing title from API but from inside the svg.

```
<svg {$attributes}>
	<use xlink:href="{$spritemap}#{$symbol}" />
</svg>
```

packages/clay-icon/src/ClayIcon.soy
carloslancha: `title` is just that, the title of the icon, is semantically understandable, where it's placed shouldn't care the user IMHO.

pat270: User's do care. It's extremely frustrating to have an attribute named the same as an element that exists in the svg spec but is applied somewhere else. We even placed it in the title element and now we moving it to aria-label? What if we need to place something in the title element in the future what will we call it then?

carloslancha: My point is that this is no ClaySVG, is ClayIcon, user shouldn't care about the markup but that an icon is placed. `title` could refer to the container title attribute too, or `ariaLabel` to an svg attribute.

PR #1322, opened 2018-11-20
Fixes #1321 - (ClayCSS) Added new Lexicon Icon `document-pdf`

### LINKED ISSUE #1321: ClayCSS: New Lexicon SVG Icon Document PDF
https://issues.liferay.com/browse/LEXI-460

pat270:
https://issues.liferay.com/browse/LEXI-460


jbalsas:
Hey @pat270, could you also update https://github.com/liferay/clay/tree/master/clayui.com/static/images/icons with the icon, so it properly shows in the documentation? I think we might be missing some already. Could you please consolidate the icons folder so it shows all the currently present icons?

@diegonvs, since this lives in the same repository... could we [simply change the path](https://github.com/liferay/clay/blob/master/clayui.com/plugins/gatsby-remark-foreach-icons/index.js#L50) to point to the source icons folder? That way we don't need to duplicate it. That, or have a build step to copy one over the other?

diegonvs:
Hey @jbalsas, we can use the released `clay-css` to fetch these icons instead of manually add to clayui.com. Once `static` folder is where these files are exposed to production. We can try automate copying development clay-css/icons folder specifically to static when building. Thats good?

* I'm not sure If I can gquerying something on a plugin.

jbalsas:
Could you simply create an issue here so we can get back to it at some point?

For now, it should be fine to simply copy-paste the icons folder to get all the icons in sync.

diegonvs:
Sure

PR #1324, opened 2018-11-22
Fixes #1242 - (BREAKING CHANGE) Removes title from the svg element and focusable attribute. Use role=presentation instead and rely on usage to provide proper semantics when using/wrapping the icon

### LINKED ISSUE #1242: IE11 - <a> element is hard to click with child lexicon-icon

jbalsas:


carloslancha:

carloslancha:
Resent https://github.com/liferay/clay/pull/1326

PR #1326, opened 2018-11-22
Fixes #1242 - Regen snapshots

### LINKED ISSUE #1242: IE11 - <a> element is hard to click with child lexicon-icon

carloslancha:


packages/clay-icon/src/ClayIcon.js
```diff
@@ -26,15 +26,6 @@ ClayIcon.STATE = {
...
 	elementClasses: Config.string(),
 
-	/**
-	 * Flag to indicate if the svg is focusable or not
-	 * @default false
-	 * @instance
-	 * @memberof ClayIcon
-	 * @type {?bool}
-	 */
-	focusable: Config.bool().value(false),
```
matuzalemsteles: hey @carloslancha, @jbalsas instead of removing this here, we can trigger a warning when someone uses it, stating that it is being depreciated. What do you think? And we can remove it completely in the next release...

jbalsas: Hey @matuzalemsteles, that would be fine by me. Could you take this one and apply that change? The problem will be that we don't have an easy way to flag that from the template itself, so we will only be able to show the warning when used from js...

matuzalemsteles: Sure

matuzalemsteles:

PR #1331, opened 2018-11-26
 Fixes #1242 - Adds an alert informing the discontinuation of the title and focusable API

### LINKED ISSUE #1242: IE11 - <a> element is hard to click with child lexicon-icon

matuzalemsteles:


matuzalemsteles:
Just for info: A malicious package (`flatmap-stream`) that was dependent on the `event-stream` package was removed from NPM... Just updating `yarn.lock` to make sure we're removing it...

More information:
- https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream
- https://github.com/dominictarr/event-stream/issues/116

pat270:
@jbalsas @matuzalemsteles removing the `focusable="false"` attribute re-introduces svg icons receiving focus on tab in IE11 and Edge. This is an IE specific attribute.

packages/clay-icon/src/ClayIcon.js
```diff
@@ -26,15 +26,6 @@ ClayIcon.STATE = {
...
 	elementClasses: Config.string(),
 
-	/**
-	 * Flag to indicate if the svg is focusable or not
-	 * @default false
-	 * @instance
-	 * @memberof ClayIcon
-	 * @type {?bool}
-	 */
-	focusable: Config.bool().value(false),
```
pat270: See my comment below.

jbalsas:
Got it, we can leave that then. Does the rest look good to you @pat270?

pat270:
LGTM. Can we roll back https://github.com/liferay/clay/pull/1307 and do the same there?

PR #1332, opened 2018-11-27
Fixes #1318 Replace hardcoded rem values with $font-size-base

### LINKED ISSUE #1318: ClayCSS: Support Document Font Size to be 62.5% or 10px for easier rem/em conversion
Issue from Slack Chat:
> Hey, a question. Is there a reason for 1rem to be 16 and not 10px in Liferay 7.1 ga1 in the Classic Theme? (i know 16px is default but isn't 10px the best practice?)

> you’d set `font-size: 62.5%`, then have 1rem map to 10px so the conversion is simpler, 2.4rem == 24px

Changing this outright in Atlas would break any theme currently extending it.

1) We could create a function that returns a rem value based on the document's base font size and a pixel value. We would have to wrap every rem value we use in that function. We would also have to overwrite all Bootstrap 4 variables that use rem values with this new pattern. I'm also not sure if this will work 100%. I've done a limited test and it's working but there are probably edge cases.

2) Another option is to provide a base 10px or 62.5% variable theme in the compatibility layer.

3) The final option I can think of is to offer a third party variable theme with base 10px or 62.5%.

All 3 of these options will only provide support for Sass inside our themes. Any Sass provided by our widgets and other plugins will have to be customized.

sbaechler:
This allows user to change the base font size from 16px to anything without breaking the layout.

It uses the Bootstrap variable `$font-size-base` instead of a hardcoded rem value.

fixes #1318 
https://github.com/liferay/liferay-themes-sdk/issues/44


sbaechler:
I locked the event-stream dependency to 3.3.4. 3.3.6 contains malware that searches a user's computer for Bitcoin wallets and empties them.

pat270:
Hey @sbaechler,

Thanks for the pull, but this is essentially doing the same thing as hardcoding the rem value. These changes are assuming a value of 1rem for `$font-size-base` with the document's font-size being 16px. If we change the document's font-size to 10px it will scale everything the same as it is currently. If we change `$font-size-base` to 0.625rem (10px) it will also scale everything down (e.g., `$alert-padding-x` should be 16px but it is now 10px).

We also don't have control over the meaning of `$font-size-base`. Bootstrap defined it for us by setting it on the `body` element. If we want to change the document's base font-size we need to set it on the `html` element and ideally called something else like `$document-font-size-base`.

Right now I'm thinking about adding a variable theme to Clay CSS base and Atlas that is based on document font-size 10px. It would be an opt-in theme via a setting variable or copy paste.

sbaechler:
Hi Patrick

Let me explain the idea behind this.
In fact, it does make a difference. `$font-size-base` is the size of the body text. This can now be decoupled from the rem value.

If you want to define 1rem as 10px while keeping a base font size of 16px you can do

    $font-size-base: 1.6rem;

    html {
        font-size: 10px;
    }

This is what you actually have to do now if you upgraded form a Liferay 7 site to 7.1 since Liferay 7 had a 10px root font size and 7.1 has 16px.

I don't think Bootstrap will change this variable. The implications would be too big. But if you prefer, we could have a `$clay-font-size-base` variable to separate the Liferay styles from the public UI.

I would also change the line widths to Pixel values. Otherwise you get fractions which don't render nicely.

pat270:
Moving this conversation to https://github.com/liferay/clay/issues/1318 so that it's more visible.

pat270:
Closing this in favor of #1423

PR #1338, opened 2018-11-29
Fixes #1323 | Update lockfiles

### LINKED ISSUE #1323: Clayui.com - Automate icons exposure to static folder
As discussed on #1322

diegonvs:


clayui.com/package.json
```diff
@@ -3,9 +3,10 @@
   "name": "clayui.com",
   "license": "MIT",
   "scripts": {
+    "copy:icons": "cpx '../packages/clay-css/src/images/icons/*.svg' static/images/icons",
```
matuzalemsteles: hey guys, we need to exclusively copy the `lib/images/icons/icons.svg` file, we just need it.

diegonvs: I discussed internally with @matuzalemsteles and It makes sense, I'll send another PR 👍

diegonvs: We don't need exclusively copy the spritemap(`icons.svg`) because we need to get name of svg files too. When 'lib' folder isn't copied, the spritemap will be missed and if another SVG is used on client-side it will be missed

PR #1340, opened 2018-11-30
ClayCSS: Create markup for labels in input

pat270:


matuzalemsteles:
hey @pat270, the icons on the label are bigger, was that supposed to happen or do we have to change some markup?

See example:
screen shot 2018-12-03 at 10 34 48

pat270:
@matuzalemsteles The Lexicon team requested the size of the icons inside labels be 16px.  https://issues.liferay.com/browse/LEXI-393. They need to change the rules, change their mock ups, or make the times svg smaller.

victorvalle:
Hi! The icons inside 24px height labels have a size of 12x12 instead of 16x16.

Sorry for the delay

PR #1353, opened 2018-12-04
Fixes #1348 - Adding colon to alert titles

### LINKED ISSUE #1348: Alert titles don't follow lexicon spec
They should be followed by a ":". See https://issues.liferay.com/browse/LPS-87793.

matuzalemsteles:


jbalsas:
Hey @matuzalemsteles, @4lejandrito, I'm not convinced this belongs in here like this. One thing we've done in the past is to enforce Lexicon too much, which has rendered clay components uncustomizable.

As it is, an alert just takes a title. Why not simply pass this title on usage? We're currently using the `warning` keys in Liferay DXP, but could be using `warning-colon` instead.

If we want to keep it as is, then we should at least add some params here so the colon is not always appended. Otherwise, people might be forced to use something else down the line.

What do you think?

4lejandrito:
Hi @jbalsas,

I think if Clay is an implementation of Lexicon the colon should be here. I discussed this with @ambrinchaudhary and she thinks having the colon in the language message is not a good idea.

@ambrinchaudhary what do you think?

ambrinchaudhary:
I agree with @4lejandrito that if this is a Lexicon pattern, Clay should provide it. But we can do it customizable as @jbalsas said. We can create a boolean attribute like `titleColon` that by default is `true`, so in `ClayAlertBase.js` have something like this: `{$title}{if $titleColon}:{/if}`. I think this could be the best approach.

carloslancha:
Hi! I agree with @jbalsas I don't think this belongs here. What if any language doesn't use semicolon in that way, for example? And adding a parameter for every single little think we want to customize like this IMHO is overkill...

4lejandrito:
I agree with @carloslancha, some languages might not use colon for this purpose.

See https://github.com/jbalsas/liferay-portal/pull/1533.

Feel free to close this.

matuzalemsteles:
hey guys, since we're all agreeing. I'll close this and mark the issue as wontfix. Thanks

PR #1355, opened 2018-12-04
Fixes #1354 Clay CSS Buttons in Input Group should be more customizable

### LINKED ISSUE #1354: ClayCSS: Buttons in Input Group should be more customizable
CSS property values for Buttons in Input Groups are coupled to the `$btn-*` variables. Decoupling them will provide more customization options for buttons inside a form input for example.

pat270:


jbalsas:
Looks great, @pat270!

My main concern was that we were losing some defaults here, introducing a breaking change, but if I'm reading the code correctly, we're setting the default values of the new variables to the old `$btn-` variables, right?

I'm merging it in 👍

pat270:
@jbalsas yes default values were moved to variables. There were some redundant properties that are defined in the normal size and were being defined again in lg and sm sizes.

PR #1357, opened 2018-12-05
Fixes #1328 - Add the new API rowClasses

### LINKED ISSUE #1328: Add rowClasses and selectedRowClasses to ClayTable
Currently, [ClayTable](https://clayui.com/docs/components/table/table.html) renders a very restrictive markdown. More specifically, it [doesn't allow additional classes on the table rows](https://github.com/liferay/clay/blob/master/packages/clay-table/src/ClayTable.soy#L171).

We should at least add the following:
- `rowClasses: string`. List of classes to append to the `tr` element
- `selectedRowClasses: string`. List of classes to append to the `tr` element **when the row is selected**. These classes will be added in addition to the ones set in `rowClasses`

By configuring this, users would have more control on the row states that should facilitate certain animations or interactions.

table

matuzalemsteles:
hey guys, I've created a little demo to get what you're hoping for in issue #1328. @gcmznt, @FabioDiegoMastrorilli you can look if this solves your problem and would be the best approach... please leave your ideas.

cc @jbalsas


packages/clay-table/demos/a11y.html
```diff
@@ -220,6 +230,63 @@ <h1 class="page-title">
...
 			spritemap: spritemap,
 		}, '#container');
+
+		// Table customize
+		new metal.ClayTable({
+			tableClasses: 'table table-autofit table-responsive',
+			useDefaultClasses: false,
+			events: {
+				'cellContentClicked': (event) => {
+					event.data.event.target.closest('tr').classList.toggle('active');
```
jbalsas: It think this should work, but here, instead of traversing the dom, we could simply do:

```
this.items = items.map(
    item => Object.assign({}, item, event.data.id == item.id ? {rowClasses: 'table-row active'} : {rowClasses: ''});
);
```

What do you think @matuzalemsteles,  @gcmznt, @FabioDiegoMastrorilli?

matuzalemsteles:
hey guys, will we still proceed with this?

cc @jbalsas

matuzalemsteles:
hey @FabioDiegoMastrorilli can you take a look at this? if this will help you.

FabioDiegoMastrorilli:
> hey @FabioDiegoMastrorilli can you take a look at this? if this will help you.

Hello @matuzalemsteles :) it looks perfect to me. Thank you!

PR #1359, opened 2018-12-06
Fixes #1265 - Adds the new API icons and adds warning of discontinuation of icon and itemsIconAlignment

### LINKED ISSUE #1265: Update ClayDropdown Markup
We need to update the ClayDropdown markups with the recent changes in #1245.

matuzalemsteles:
This is a proposal not to have a breaking change now and we can send an API discontinuation warning that may no longer make sense of the change to get icons from both sides in the dropdown and independently.

One problem is that we can not tell when there is an icon left or right to add the `.dropdown-menu-indicator-(start|end)` classes along with `dropdown-menu`, I talked to @pat270 about this, his suggestion is maybe to be able to add the `li` instead of the root element, which may not be a good approach to this, but we'll have to see some way to deal with it. Please leave your ideas.

cc @carloslancha


matuzalemsteles:
I went ahead with this PR and made some modifications that makes it safer to be merged:

- I went back and I'm keeping the `itemsIconAlignment` API to be used in conjunction with the API of icons. It still becomes mandatory to be set to use `icons`.
- I added a new `left-right` flag in the `itemsIconAlignment` API to center icons on both sides.
- I came back with the icon discontinuation messages from to eliminate noises on the console for now.
- I went back and removed the discontinuation message from the `icon` API to remove noises on the console. (Just so people do not bother to change it for now.)

Remembering that this change has full compatibility with the API of `icon` and `itemsIconAlignment`.

PR #1373, opened 2018-12-07
Fixes #1226 - Add the new API tbarInlineBreakpoint to set breakpoint of the results bar

### LINKED ISSUE #1226: Management toolbar mobile viewports Clear action must be right aligned
The aligntment of the clear action should be right.
This action is now aligned next to the text

image

Please see: https://liferay.invisionapp.com/share/PVN3QSTHK65#/312223170_0604m

Do not modify other styling things, please.

matuzalemsteles:
hey guys, I do not know if this would be the best name, any suggestions?

### DIFF
--- a/packages/clay-management-toolbar/src/ClayManagementToolbar.js
+++ b/packages/clay-management-toolbar/src/ClayManagementToolbar.js
@@ -511,6 +511,15 @@ ClayManagementToolbar.STATE = {
 	spritemap: Config.string().required(),
 
 	/**
+	 * Sets the breakpoint of the results bar.
+	 * @default md
+	 * @instance
+	 * @memberof ClayManagementToolbar
+	 * @type {?(string|undefined)}
+	 */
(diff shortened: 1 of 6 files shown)


jbalsas:
I'm definitely not in love with that API 😂 Can you describe what it does in a way that's not technical? @marcoscv-work, @pat270, @carloslancha, ideas?

matuzalemsteles:
hey @jbalsas, Actually I'm not so sure about this if we really should expose as an API or make `tbar-inline-xs-down` as default. I do not see the use cases... just wanting to be in accordance with Lexicon 🙂. @pat270 do you see any case for this in that component?

pat270:
I'd say if you don't want to allow customizing just use `tbar-inline-xs-down`. The only reason you would use different breakpoints is if you wanted the Clear All button appear floated right or inline.

jbalsas:
I see, thanks @pat270! @matuzalemsteles, what do you think of simply hardcoding it for now then? If ayone asks, we can provide this easily in the future.

matuzalemsteles:
sure, np. I'll close this PR and send another one.

PR #1382, opened 2018-12-10
Fixes #1381 - Improvements clay multi select

### LINKED ISSUE #1381: ClayMultiSelect should be fully functional by default without having to handle events

matuzalemsteles:


packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -332,6 +343,16 @@ ClayMultiSelect.STATE = {
...
 
+	/**
+	 * Set the field name that contains the string to be added when
+	 * the item is selected.
+	 * @default originalString
+	 * @instance
+	 * @memberof ClayMultiSelect
+	 * @type {?string}
+	 */
+	filterField: Config.string().value('originalString'),
```
matuzalemsteles: hey @jbalsas, I do not know if I understood your intention of the `filterField`, but I'm using it for the context of setting the field when the items are filtered.

https://github.com/liferay/clay/pull/1382/files#diff-21f70f10741963ea56df2ba593022af3R64

jbalsas: Hey @matuzalemsteles, yeah, I think this does what's expected, but the initial value is odd. It should likely be `value`. I'd expect the `itemSelected` event from `AutoComplete` to emit the full item that was selected. This should be more than just a string, it should be the whole object. For example:

```js
const items = [{
    foo: 1,
    bar: 2
}, {
    foo: 3,
    bar: 4
}];

You'd then set `filterField = 'foo'`, and the value to add should then be `1` or `3` depending on which item got selected. 

Additionally, we should have a counterpart for the `label`. We can't simply derive it from the `value`, as these things might not always have a clear mapping. Imagine the following:

```js
const categories = [{
    categoryId: 10032,
    name: 'Food',
}, {
    categoryId: 203214,
    name: 'Animals',
}];
```

In this case, you'd likely want to set the field value to map to the `categoryId` field, and the label one to map the `name` field.

With this in mind, I'd propose something like:

```js
const categories = [{
    categoryId: 10032,
    name: 'Food',
}, {
    categoryId: 203214,
    name: 'Animals',
}];

new MultiSelect({
    [...]
    // Both could be string|function
    labelLocator: 'name', // could be (item) => item.name
    valueLocator: 'categoryId', // could be (item) => item.categoryId
    [...]
});
```

matuzalemsteles: Thanks @jbalsas for the clarifications, I'll make the modifications.

matuzalemsteles: hey @jbalsas just clarifying something related to the `filterField` API, we are only using to get the field of the filtered item (See https://github.com/liferay/clay/blob/master/packages/clay-data-provider/src/ClayDataProvider.js#L152). I'm assuming a dev can make its own filter system and disable the default.

matuzalemsteles: On the use of `labelLocator` and `valueLocator` of accepting functions is a bad thing because we need to pass to Soy and it does not deal well with this. I like the idea of ​​accepting only `string`, what do you think?

jbalsas: Hey @matuzalemsteles, yeah, I'm not sure I fully understand the `filterField` API. Could you explain in some detail? Maybe make a diagram? :D 

On the other APIs, what's the problem, where are we passing them down? We can keep them as strings for now, but the fact that we have problems supporting functions too is a possible sign that we might need to change something somewhere :)

matuzalemsteles: hey @jbalsas maybe I have to rename `filterField` haha... but I think I'm complicating things here, I was just needing it because Autocomplete was returning the object that was filtered... I'll just return the real object of the selected item... This will soften.

On `labelLocator` and `valueLocator` from what I understand it is working as a locator of the field's contents and field name where the data is, we always need it as `string` because we have to use it in Soy (https://github.com/liferay/clay/blob/master/packages/clay-multi-select/src/ClayMultiSelect.soy#L218 and https://github.com/liferay/clay/blob/master/packages/clay-multi-select/src/ClayMultiSelect.soy#L249) and we just need the field name to add the correct names to the `selectedItems` list (https://github.com/liferay/clay/pull/1382/files#diff-21f70f10741963ea56df2ba593022af3R104), even if we add the item object returned from the `itemSelected` event to `selectedItems` we will need the field name to create a new item for example, does it make sense? Do you have any idea how we can dodge this?

matuzalemsteles: hey @jbalsas, yesterday I was looking at this and I did the following to be able to accept the `string` and `function`, I kept your idea and I'm just not taking into consideration to modify the `selectedItems` API schema, that way I just do the mapping of `labelLocator` and `valueLocator`... What do you think?

jbalsas: What did you do? 😂

matuzalemsteles: haha 😅, I'm going to send what I had done, it would have been clearer.

jbalsas: 😂 👍

jbalsas:
I think this looks much better, @matuzalemsteles. I'll merge ahead so we can put it through some extensive testing directly on the use case.

One thing I noticed is that we create one `input` tag per `label` or selected value, rather than just one. That should work as a default Array on the server implementation, so it might be great, but we'll need to double-check.

Thanks @matuzalemsteles ! 👌

PR #1387, opened 2018-12-13
RFC - API proposal to ClayDatePicker component

matuzalemsteles:
> 🚨 This PR is not to be merged. Do not stick to the code now, but the idea behind. I will still send another PR to evaluate the implementation and our good practices, once this idea is valid.

## Disclaimer

This is an API proposal for the ClayDatePicker component, following the use cases contained in the document. As we do not have a markup ready yet I've improvised a markup and style for the demo, please disregard.

- [Doc](https://docs.google.com/document/d/1iPFtX7Zku7N-sElSfpGSBlfpYvMu4ATCkVVCU_oB3a4/edit)

## Proposal

The API is focused on making the datepicker internationalized, such as `weekdaysShort` and `months`, and can customize the date and time format.

We have an important point when we have the internationalization of the calendar, in some languages ​​we have to define `firstDayOfWeek` in the API, I'm not sure how this will be placed on the portal, please leave your ideas.

> Obs: No keyboard interaction has been created yet.

### Extension points

The extension point has not been thought yet, I'm considering it to be a `deltemplate`, but maybe we should go with an API that accepts `html` since dev can add anything inside and can add more than one function to listen the events.

- **ClayDatePicker.ExtensionPoint**

## References

- [Anti Design](https://ant.design/components/date-picker/)
- [Atlaskit](https://atlaskit.atlassian.com/packages/core/datetime-picker)
- [Blueprintjs](https://blueprintjs.com/docs/#datetime/datepicker)
- [Microsoft Fabric](https://developer.microsoft.com/en-us/fabric#/components/datepicker)
- [ReactDayPicker.js.org](http://react-day-picker.js.org/)


jbalsas:
Hey @matuzalemsteles, without going into too much detail, one thing I miss is something like a `useNative` API that would simply render a `<input type="date">` tag. Currently, both Chrome and Firefox ship very rich date pickers, and IOS and Android properly understand those as well, so we need a way to opt-in (or opt-out for specific browsers).

Maybe better than a `useNative` we could have a `preferNative`, that could be either a boolean or some kind of browserlist... if it's a boolean, when set to `true` we could do feature detection to render the simple html tag.

I'll look more into it, but I'd love if we can start leveraging standard components. It's a shame we'll need to implement this anyways for IE11 and Safari 🤷‍♂️

matuzalemsteles:
hey @jbalsas, Great idea, maybe the datepicker will become more necessary when we have a date range implemented here...

julien:
Hi @matuzalemsteles,

1) In terms of look and feel, I think this matches the design docs.

2) I also think it would be nice have an option to prefer the native date picker (i.e. `<input type="date">`) when possible.

3) Concerning the styles, wouldn't it be better to have a scss file that we later include in the demo rather than declaring the styles in the 
demo's html? (I imagine this is what's planned but I thought I'd ask)

4) Concerning the `firstDayOfWeek` property, it's declared like this: 

    ```javascript
    firstDayOfWeek: Config.oneOf([0, 1, 2, 3, 4, 5, 6]).value(0),
    ```
     Maybe we could use:

    ```javascript
    firstDayOfWeek: Config.inRange(0, 6).value(0),
    ```
    
    It's not really an improvement but it's a bit shorter to read

5) Concerning event listeners:

    Could we standardize the names?

    For example: 

    Some event handlers are named `_handleDotClicked` and others `_handlePreviousMonth` 
    (without `Clicked`)

6) Concerning the `placeholder` property, do you think it would be possible to have a default value,
   the same way we we have for `weekdaysShort` and `firstDayOfWeek`? 
	 
	 I'm not sure it makes sense, but maybe we can use the value of the `dateFormat` property?

7) Concerning the `years` property I was wondering if instead of being an `Array` we could use an `Object`:

    So instead of having this:

    ```
    years: Config.array().required(),
    ```

    We could have something like this:

    ```
    years: Config.shapeOf({
	    start: Config.number.required().value(new Date().getFullYear()),  
	    end: Config.number.required().value(new Date().getFullYear()),
    })
    // This `new Date()` should probably be declared before to avoid re-creating `Date` objects.
    ```

    This way we could have default values hen creating a new instance, so instead of using this:

    ```javascript
     new metal.ClayDatePicker({
         // ...
         years: [
             '2018',
             '2019',
             '2020',
             '2021',
             '2022',
             '2023',
             '2024',
         ]
     }, '#some-element');
     ```

      We can use this:

      ```javascript
      new metal.ClayDatePicker({
          // ...
          years: {
            start: 2018,
          },
      }, '#some-element'); // From 2018 to current year
      ```

      Or:

      ```javascript
      new metal.ClayDatePicker({
          // ...
          years: {
              end: 2024,
          },
      }, '#some-element'); // From current year to 2024
      ```

      Obviously we would need to check that `years.start < years.end` but it might make things easier?

8) I was also wondering if we'd rewrite this in React sooner or later?

These are just a few ideas I had when reviewing but I'm not saying this is the correct way to do it, feel free to tell me what you think.

julien:
Hey @matuzalemsteles 

Thanks for the quick reply; and to keep the conversation going here are my replies:

> I really do not know when exactly we're going to start writing the components in React, Forms team is using Metal and they would use that component... I'm really lost when we're gonna do this... 🤷‍♂️

Sorry for the confusion, I actually put that comment so we could discuss it all together, as it does seem like an important topic that we're all wondering about.

>  Actually I had done just for this demo case, I remove it once we have the component well defined. I do not know if my CSS would help Patrick to reuse. This marking is not the best either, we have to rewrite it.

Sounds great!

> I like this, I think we can go with this, but I wanted your help and from others to know if it makes sense that we block the visualization of the Data in years that are not listed, this would disable the next month or previous month when it reaches the limit... makes sense?

As far as I'm concerned, yes it makes sense.

matuzalemsteles:
hey guys, I'm closing this PR in favor of the Date Picker in React. #1440

PR #1394, opened 2018-12-14
Fixes #1390 - Render the id when wrapTable is false

### LINKED ISSUE #1390: ClayTable doesn't render the id when it is not wrapped
Steps to reproduce: just create a table with `wrapTable` set to false and see the resulting HTML. It has no id and makes SPA fail when refreshing the page, for example.

matuzalemsteles:


carloslancha:
Merged. Thx!

PR #1395, opened 2018-12-14
Fixes #1379 - Call ClayAlertBase close method

### LINKED ISSUE #1379: ClayToast and ClayStripe close methods doesn't work as expected
All the logic of the closing animation is in ClayAlertBase and it's not being called through close method in ClayToast and ClayStripe. We need to refactor all this to improve the logic.

matuzalemsteles:


carloslancha:
Hey @matuzalemsteles  With this we're abandoning state-event communication to directly call child component methods... So I definitely do not like this solution.

`close` method is emitting `hide` event, that has a default listener. That listener is what is not ok right now, it should change `visible` state to false so it can be propagated to ClayAlertBase, I guess.

ClayAlertBase maybe should move some logic to a syncVisible method or something like that.

Also maybe is a good idea that ClayAlertBase emit an event when, on `autoClose == true`, timer starts and stops so the user can prevent it too.

If you have any doubts please let me know ;)

Thanks!

matuzalemsteles:
hey @carloslancha, I do not know if this would be the best way now, but once we call the `close` method of **ClayAlertBase**, it will issue the `hide` event and the above components will do the same behavior as before... This is a quick fix until we can adjust what you asked for above.

carloslancha:
hey @matuzalemsteles this is not critical, I prefer to do it properly (afaik nobody is asking for this yet). Txh!

matuzalemsteles:
Oh okay, so I'm going to close it and we can reprioritize it later.

PR #1396, opened 2018-12-14
Fixes #1393 - Replace itemSelected in favor of itemClicked

### LINKED ISSUE #1393: ClayDropdown itemSelected event should be itemClicked
When an item in ClayDropdown is clicked the event `itemClicked` is fired. When an item is selected via keyboard and enter key is pressed event `itemSelected` is fired.

The behavior is the same so same event should be fired. We need to change `itemSelected` event to `itemClicked`.

matuzalemsteles:


carloslancha:
Hey @matuzalemsteles we need to change it also in the rest of dropdown components. Thx.

matuzalemsteles:
Oh I had forgotten the other components, thank you @carloslancha!

carloslancha:
Merged. Thx!

PR #1402, opened 2018-12-17
Fixes #1392 - Add actionHandler to ClayManagementToolbar

### LINKED ISSUE #1392: Add support to Clay Component to receive an actionHandler

carloslancha:


packages/clay-card/src/ClayUserCard.soy
```diff
@@ -5,6 +5,7 @@
  */
 {template .render}
 	{@param name: string}
+    {@param? _handleActionItemClicked: any}
```
jbalsas: SF?

packages/clay-component/src/ClayComponent.js
```diff
@@ -116,6 +116,11 @@ class ClayComponent extends Component {
 
 		const listeners = this.getRawListeners_(eventName);
 
+		if (this.actionHandler) {
```
jbalsas: Should the `actionHandler` be able to prevent events?

jbalsas:
We've had an offline conversation about this... the proposal is to move this to a `defaultEventHandler` for clarity, and tweak the `emit` method so that the lifecycle is as follows:
- An event is emitted
- We loop through all registered listeners
    - If they're default listeners, we store them for later
    - If they're regular listeners, we run them passing down the event facade
- If we have a registered `defaultEventHandler`, with the event method implementation, we run it passing down the event facade
- If no-one has prevented the default behaviour, we run the stored default listeners

carloslancha:
Resent https://github.com/liferay/clay/pull/1405

PR #1414, opened 2018-12-19
Fixes #1412 | Add clay-component as a dependency

### LINKED ISSUE #1412: Charts throws an error on npm run build if extends from ClayComponent instead Component

julien:


julien:
Closing for now since this doesn't seem to be the correct way to fix it.

That said if @bryceosterhaus has an idea, I'm all ears (or eyes in this case)

PR #1423, opened 2019-01-07
[DO NOT MERGE] Clay CSS with base 10px font

pat270:
This is my initial pull for base 10px font. The way to use it is to set the `$html-font-size` and calculate the rem value for `$clay-radix` using `16 / $html-font-size` (e.g. for base 12px, $clay-radix = 1.333333rem). This lets you change the document's base font size and preserve original component sizes for Bootstrap 4 and Clay CSS.

For now, I've created a variable theme for Clay CSS Base and Atlas that scales the font sizes to minimize any breakages, but we might be able to make this change directly into Clay CSS.

Before we merge something like this we need to consider how far we are going to support this feature. Should changing the base font size scale all components in DXP? We have hardcoded rem values in the theme as well as module CSS. We can provide support in the theme, but I'm not sure how to handle module specific CSS.

Should this even be a feature in Clay CSS or is this better as a separate plugin for Clay CSS?

### DIFF
--- a/packages/clay-css/src/scss/atlas/_variables.scss
+++ b/packages/clay-css/src/scss/atlas/_variables.scss
@@ -1,3 +1,5 @@
+@import "variables/_base-font-sizing";
+
 @import "variables/_globals";
 
 @import "variables/_alerts";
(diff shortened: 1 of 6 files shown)


jbalsas:
I'm closing this for now, as we continue to explore the implications of such a change and how to better reach a compromise here.

PR #1427, opened 2019-01-08
Color picker

bryceosterhaus:
**DO NOT MERGE**
_Waiting to extract css into clay-css_

This is my initial run through of the color picker component written in React. I wanted to get some feedback on how this looks so far. I decided to experiment and use React's new hooks API for the component, although I am also fine if we decide to just use class components instead. To test it out, just run `npm run start` to spin up the dev-server. You can also look at the demo app for how its used.

@pat270 take a look at the style sheet and let me know if you have any questions on extracting it into clay-css.

### DIFF
--- a/packages/clay-color-picker/.eslintrc
+++ b/packages/clay-color-picker/.eslintrc
@@ -1,0 +1,9 @@
+{
+	"plugins": ["react"],
+	"rules": {
+		"no-unused-vars": ["error", {"ignoreRestSiblings": true}],
+		"react/jsx-uses-react": "error",
+		"react/jsx-uses-vars": "error",
+		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"]
+	}
+}
--- a/packages/clay-color-picker/.eslintrc
+++ b/packages/clay-color-picker/.eslintrc
@@ -2,8 +2,9 @@
 	"plugins": ["react"],
 	"rules": {
 		"no-unused-vars": ["error", {"ignoreRestSiblings": true}],
+		"react/jsx-sort-props": "error",
 		"react/jsx-uses-react": "error",
 		"react/jsx-uses-vars": "error",
-		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"]
+		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"],
 	}
 }
--- a/packages/clay-color-picker/.eslintrc
+++ b/packages/clay-color-picker/.eslintrc
@@ -1,10 +1,12 @@
 {
-	"plugins": ["react"],
+	"plugins": ["react", "react-hooks", "sort-destructure-keys"],
 	"rules": {
 		"no-unused-vars": ["error", {"ignoreRestSiblings": true}],
 		"react/jsx-sort-props": "error",
 		"react/jsx-uses-react": "error",
 		"react/jsx-uses-vars": "error",
+		"react-hooks/rules-of-hooks": "error",
 		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"],
+		"sort-destructure-keys/sort-destructure-keys": 2
 	}
 }
--- a/packages/clay-color-picker/.eslintrc
+++ b/packages/clay-color-picker/.eslintrc
@@ -1,12 +1,12 @@
 {
 	"plugins": ["react", "react-hooks", "sort-destructure-keys"],
 	"rules": {
+		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"],
 		"no-unused-vars": ["error", {"ignoreRestSiblings": true}],
+		"react-hooks/rules-of-hooks": "error",
 		"react/jsx-sort-props": "error",
 		"react/jsx-uses-react": "error",
 		"react/jsx-uses-vars": "error",
-		"react-hooks/rules-of-hooks": "error",
-		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"],
 		"sort-destructure-keys/sort-destructure-keys": 2
 	}
 }
--- a/packages/clay-color-picker/README.md
+++ b/packages/clay-color-picker/README.md
@@ -1,0 +1,1 @@
+# clay-color-picker
--- a/packages/clay-color-picker/demo/index.html
+++ b/packages/clay-color-picker/demo/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<link
+			rel="stylesheet"
+			href="https://cdn.jsdelivr.net/npm/clay/lib/css/atlas.css"
+		/>
+	</head>
+
+	<body style="background: white; padding: 24px; max-width: 300px;">
+		<div id="root"></div>
+	</body>
+</html>
--- a/packages/clay-color-picker/package.json
+++ b/packages/clay-color-picker/package.json
@@ -25,7 +25,7 @@
 		"babel-preset-stage-0": "^6.24.1",
 		"css-loader": "^2.1.0",
 		"eslint-plugin-react": "^7.11.1",
-		"eslint-plugin-react-hooks": "0.0.0",
+		"eslint-plugin-react-hooks": "^0.0.0",
 		"eslint-plugin-sort-destructure-keys": "^1.1.0",
 		"html-webpack-plugin": "^3.2.0",
 		"sass-loader": "^7.1.0",
--- a/packages/clay-color-picker/package.json
+++ b/packages/clay-color-picker/package.json
@@ -25,6 +25,8 @@
 		"babel-preset-stage-0": "^6.24.1",
 		"css-loader": "^2.1.0",
 		"eslint-plugin-react": "^7.11.1",
+		"eslint-plugin-react-hooks": "0.0.0",
+		"eslint-plugin-sort-destructure-keys": "^1.1.0",
 		"html-webpack-plugin": "^3.2.0",
 		"sass-loader": "^7.1.0",
 		"style-loader": "^0.23.1",
--- a/packages/clay-color-picker/package.json
+++ b/packages/clay-color-picker/package.json
@@ -1,0 +1,35 @@
+{
+	"name": "clay-color-picker",
+	"version": "0.0.1",
+	"private": true,
+	"main": "lib/index.js",
+	"jsnext:main": "src/index.js",
+	"scripts": {
+		"compile": "./node_modules/.bin/babel -d lib/ src/ -s",
+		"prepublish": "npm run compile",
+		"start": "webpack-dev-server --mode development"
+	},
+	"author": "Bryce Osterhaus",
+	"dependencies": {
+		"prop-types": "^15.6.2",
+		"react": "16.7.0-alpha.2",
+		"react-dom": "16.7.0-alpha.2",
+		"tinycolor2": "^1.4.1"
+	},
+	"devDependencies": {
+		"babel-cli": "^6.24.1",
+		"babel-core": "^6.25.0",
+		"babel-loader": "^7.0.0",
+		"babel-preset-env": "^1.6.0",
+		"babel-preset-react": "^6.24.1",
+		"babel-preset-stage-0": "^6.24.1",
+		"css-loader": "^2.1.0",
+		"eslint-plugin-react": "^7.11.1",
+		"html-webpack-plugin": "^3.2.0",
+		"sass-loader": "^7.1.0",
+		"style-loader": "^0.23.1",
+		"webpack": "^4.22.0",
+		"webpack-cli": "^3.1.2",
+		"webpack-dev-server": "^3.1.10"
+	}
+}
--- a/packages/clay-color-picker/src/Basic.js
+++ b/packages/clay-color-picker/src/Basic.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Splotch from './Splotch';
+
+Basic.propTypes = {
+	colors: PropTypes.arrayOf(PropTypes.string),
+	label: PropTypes.string,
+	onChange: PropTypes.func,
+};
+
+/**
+ * Renders basic color picker
+ * @return {React.Component}
+ */
+function Basic({colors, label, onChange}) {
+	return (
+		<div className="basic-color-picker">
+			{label && (
+				<div className="label-container">
+					<label>{label}</label>
+				</div>
+			)}
+
+			<div className="splotch-grid">
+				{colors.map(hex => (
+					<Splotch
+						onClick={() => onChange(hex)}
+						key={hex}
+						value={hex}
+					/>
+				))}
+			</div>
+		</div>
+	);
+}
+
+export default Basic;
--- a/packages/clay-color-picker/src/Basic.js
+++ b/packages/clay-color-picker/src/Basic.js
@@ -24,8 +24,8 @@ function Basic({colors, label, onChange}) {
 			<div className="splotch-grid">
 				{colors.map(hex => (
 					<Splotch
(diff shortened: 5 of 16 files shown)


julien:
Hi @bryceosterhaus,

I personally like the idea of using hooks although I'm aware they're still in `-alpha`, I guess they will be released soon.

This PR taught me a few new things: 
	- I didn't know about `<React.Fragment>`
	- I had to translate `Splotch` 😸  but the name seems appropriate 

I also guess a few tests will be added.

Nice work!

jbalsas:
<blockquote class="twitter-tweet" data-lang="es"><p lang="en" dir="ltr">Heads up: react@16.8.0-alpha.0 and react-dom@16.8.0-alpha.0 are out — for people who want the latest build with Hooks enabled. (🚧The warnings about using alphas apply.)<br><br>We’re back from vacations and are focused on getting Hooks into a stable release as our biggest priority.</p>&mdash; Dan Abramov (@dan_abramov) <a href="https://twitter.com/dan_abramov/status/1083139151720001543?ref_src=twsrc%5Etfw">9 de enero de 2019</a></blockquote>

jbalsas:
Things we already discussed that would be needed before settling this in:
- [ ] Native HTML Form integration. Component should be backed by a configurable hidden input so native form submits carry on the value.
- [ ] Available `useNative` prop. When `true`, we should simply output an `input type="color"`. Not sure if we should feature-detect here or on the calling code. The idea would be to output proper input types for mobile devices, mainly.
- [ ] Hydration mechanism/constructor. For initial integration with SSR, we should be able to take an initial state markup that we could render on the server and instantiate a component on top. Something like `new ColorPicker('#myColorPicker')`
- [ ] WebComponent wrapper. We currently have this for all clay soy components. Not sure if our current mechanism would already work for it or not.

julien:
@jbalsas thanks for the extra info.

Concerning the WebComponent wrapper, do you think this really adds value?
It might be a good topic for a future discussion.

jbalsas:
It might be the case that we need to use it from somewhere else... like an exisiting metal+soy component. I think the `webcomponent` API could be a better starting point here (although it's just a feeling).

I also think we need to invest in that interface. As adoption for webcomponents grow, they should help with interoperability more and more.

bryceosterhaus:
>  Native HTML Form integration. Component should be backed by a configurable hidden input so native form submits carry on the value.

This should already exist, unless you had a different idea than https://github.com/liferay/clay/pull/1427/files#diff-b2320458fcbb4f56ba06dbe31e422568R133

> Available useNative prop. When true, we should simply output an input type="color". Not sure if we should feature-detect here or on the calling code. The idea would be to output proper input types for mobile devices, mainly.

Adding this now and will update.

> Hydration mechanism/constructor. For initial integration with SSR, we should be able to take an initial state markup that we could render on the server and instantiate a component on top. Something like new ColorPicker('#myColorPicker')

Are you expecting this to be written and added to this repo? I was anticipating this being a later issue when we try to integrate and add it to portal via a taglib.

> WebComponent wrapper. We currently have this for all clay soy components. Not sure if our current mechanism would already work for it or not.

I wasn't fully sure if we wanted to jump into this now or wait and see how we would use it first.

bryceosterhaus:
Oh gotcha, I misread the part about the constructor example. Ill look into that and webcomponents.

bryceosterhaus:
@jbalsas does it make more sense to have people intialize their components like this? Part of me would want to encourage this practice because it reduces the number of ways we have to build our components(common, umd, webcomponent). And it also enforces the user to know what they are using. As I have been looking around at other open source react components, this seems to be the standard.

```js
ReactDOM.render(
     React.createElement(MyComponent),
     document.getElementById(''myColorPicker')
);
```

jbalsas:
Hey @bryceosterhaus!

> @jbalsas does it make more sense to have people intialize their components like this? Part of me would want to encourage this practice because it reduces the number of ways we have to build our components(common, umd, webcomponent). And it also enforces the user to know what they are using. As I have been looking around at other open source react components, this seems to be the standard.

I think that like everything every approach has pros and cons :)

As a user of the library, I wouldn't necessarily enjoy having to know about `React` to use it. This is even more clear if you consider the Portal scenario as a possible client of the library. Having to know we're using `React` under the hood is not really our users concern. Sure, if someone is already using `React` the could leverage the simpler and more natural React component right away, but for other cases, having a higher level wrapper should be more ergonomic.

It should also protect us from ever wanting to switch to something else. I wouldn't want to fill our calling code of `ReactDOM` references if avoidable.

> I'm also running into an issue deciding how this component will be consumed. If it was consumed by another react component, the API is pretty straight forward by giving it both an onChange prop and also a value prop. However, if we also anticipate instantiating by doing something like new ColorPicker('#someId) then we can't use the same "controlled" approach. One thought is to create two builds, one that is consumed by other react components and then create an additional build which wraps that one with state. This second build would allow it to be instantiated as a standalone widget.

This comes down to an old dilemma we already faced with Clay. Today @andresfulla brought this same concern for React components. Making them controlled makes them less usable in isolation, as you always need to control them. Making them uncontrolled kind of encapsulates them better, so less integration is needed to use them, but is a bit of anti-pattern, I guess... 🤷‍♀️ 

In clay, what we (@carloslancha, @matuzalemsteles) ended up doing was to have them `uncontrolled` by default, but always firing events that could be cancelled in case you'd like to control them from up above. Not sure that would suit this case, though... 🤔

bryceosterhaus:
> It should also protect us from ever wanting to switch to something else.

This is a good point, I hadn't thought about that much and definitely is something to give more thought to. I also start to wonder if we are thinking too much about the "when we switch to something else" too much. That will always be the boogie man around the corner, but really we have no idea what that even looks like and it may just be easier rewrite code later than try to make it perfect now for the "next thing."

>Making them uncontrolled kind of encapsulates them better, so less integration is needed to use them, but is a bit of anti-pattern

Yeah this definitely seems like an anti-pattern for react components. I believe react will even through internal warning for uncontrolled components.

I'm thinking that its best to see this in a two step process. First step is that we create react components that behave and work in a manner that the react community would expect. Then secondly, we create a layer that allows us to be compatible with Liferay, whether that is taglibs, encapsulated js widgets, or what ever else it may be. This way we keep the react code looking and behaving like react and the middle layer is what would allow us to actually use these components in Liferay. Does that make sense to do it that way? Keeping the Liferay-usability layer separate from the components themselves? 

The other thought I start to have is, "why even choose react for Liferay?" 🤷‍♂️  But this is probably a better discussion to have once we are all together at the frontend-gathering. We can also table this entire issue until the gathering if that might be easier to all discuss in person. 🤓

jbalsas:
> I'm thinking that its best to see this in a two step process. First step is that we create react components that behave and work in a manner that the react community would expect. Then secondly, we create a layer that allows us to be compatible with Liferay, whether that is taglibs, encapsulated js widgets, or what ever else it may be.

This sounds reasonable enough. Let's see how this looks from Liferay's side before adding any unnecessary boilerplate over here.

@julien, @carloslancha, @greg-hurrell, @matuzalemsteles, care to give a final look at this? Still waiting on @pat270 to extract the CSS into `clay-css`.

matuzalemsteles:
hey guys, I want to leave just a few things I'm missing here.

- I'm missing JSDOC at least in the APIs of the `index.js` file, this will help the clayui.com plugin to create the API table.
- I wish we could continue to keep some patterns that exist in components made in Metal.js, just file naming details... the demos...

packages/clay-color-picker/src/Basic.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Splotch from './Splotch';
+
+Basic.propTypes = {
```
wincent: If you add babel-plugin-transform-class-properties (`@babel/plugin-proposal-class-properties` in Babel v7) then you can replace these with `static` property declarations, which might read a bit nicer. (Longer term, adopting Flow or TypeScript would obviate the need for propType declarations entirely).

packages/clay-color-picker/src/Custom.js
```diff
@@ -0,0 +1,245 @@
...
+							onChange(hex);
+						}}
+						key={i}
+						value={hex}
+					/>
+				))}
+			</div>
+
+			{editorActive && (
+				<React.Fragment>
```
wincent: Ideally we'd use the shorthand here:

```javascript
<>
  ...
</>
```

In practice, you only need to use `React.Fragment` if some piece of your build pipeline (lints, Babel etc) is old and doesn't handle the shorthand syntax. Not sure if there's anything in the package.json here that would prevent this here.

packages/clay-color-picker/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+	"plugins": ["react"],
+	"rules": {
+		"no-unused-vars": ["error", {"ignoreRestSiblings": true}],
+		"react/jsx-uses-react": "error",
+		"react/jsx-uses-vars": "error",
+		"no-mixed-spaces-and-tabs": ["error", "smart-tabs"]
```
wincent: We might want to add `react/jsx-sort-props` here (perhaps the "plugin:react/recommended" preset includes it already, and using the preset would probably be preferable); it looks like you've generally tried to keep things in alphabetical order, but there are some places where things are out of order.

packages/clay-color-picker/src/util.js
```diff
@@ -0,0 +1,41 @@
...
+		}
+	}
+
+	return join ? keyArray.join('') : keyArray;
+}
+
+/**
+ * Regex to only handle values of A,B,C,D,E,F,0,1,2,3,4,5,6,7,8,9
+ */
+export const HEX_REGEX = /[a-fA-F0-9]+/;
```
wincent: This regex might not be strict enough because it is unanchored; for example, if I paste "@Q#$#$a)*" in the text input, for example, it will match the regex because of the "a" in the middle of the string. This could be what we want but I haven't traced through how the match is used so I don't know. To go stricter you could either add anchors (eg. `/^...$/`) or anchors + slurp up whitespace (eg. `/^\s*...\s*$/`, or `trim()` the string first).

packages/clay-color-picker/src/Basic.js
bryceosterhaus: Since `Basic` is a function component I was unable to add it as a static property. If it was a class I would totally be with you on that.

packages/clay-color-picker/src/Custom.js
bryceosterhaus: Since we are still on babel 6, we have to use the more explicit form of Fragment.

bryceosterhaus:
@matuzalemsteles 

> I'm missing JSDOC at least in the APIs of the index.js file

I wonder if we want to find a tool that can derive the API from the propTypes of the exposed component. Otherwise we have to redundantly add jsdoc types. I agree we definitely need something for the clay site though.

> A test suite for the components. Do you plan to add later...

Yeah I am planning to add later, I just want to put this out there so that we can decide the general direction we want to go with these new components.

matuzalemsteles:
hey @bryceosterhaus,

> I wonder if we want to find a tool that can derive the API from the propTypes of the exposed component. Otherwise we have to redundantly add jsdoc types. I agree we definitely need something for the clay site though.

Yeah, I think I can work on that later so we do not need JSDOC anymore in the API but I still think we'd have to keep the descriptions to work as expected.

> Yeah I am planning to add later, I just want to put this out there so that we can decide the general direction we want to go with these new components.

ok, I've been looking at it today. For us to migrate little by little the components I think we will have to configure the jest in each React package, if you want to use enzyme (I think the most recommended) for the tests.

packages/clay-color-picker/src/Basic.js
wincent: Ah, sorry. In my first read-through I noticed you doing this on the `GradientSelector` class and then I came back and put my comment on the first instance of `propTypes` that I found, where it doesn't actually apply.

wincent:
I'd be cautious about investing too heavily in deriving anything from proptypes if there's a chance that they could be superseded by actual static types in the near-ish future, but in any case, if you want to look into some existing tooling you could check out [react-docgen](https://github.com/reactjs/react-docgen) (that happens to know how to deal with both proptypes and Flow types), and also see what Gatsby has done for their (pretty sweet) API docs; I [asked about this on Twitter a few weeks back](https://twitter.com/wincent/status/1077656693599809536) and got [pointed to this example](https://twitter.com/kylemathews/status/1077688076997144577).

I'm trying to also remember some project that a former colleague of mine (@mitermayer) built that updated JSDoc comments based on static type information, but I can't remember the name of it right now...

matuzalemsteles:
hey @wincent, thanks for the suggestions, this is a bit similar to what we did for the clayui.com site, but we used the `documentation` lib, it does the analysis only for the JSDOC, but I think we can migrate to analyze the proptypes and if we change for static types we will already be prepared. Thanks!

packages/clay-color-picker/README.md
```diff
@@ -0,0 +1 @@
+# clay-color-picker
```
matuzalemsteles: We can fill in `README.md` with the same information we have in the other components. See [ClayDatePicker](https://github.com/liferay/clay/blob/2e847b45db7e308799ed682029a995c1ab16f81f/packages/clay-date-picker/README.md)

packages/clay-color-picker/package.json
```diff
@@ -0,0 +1,37 @@
+{
+	"name": "clay-color-picker",
+	"version": "0.0.1",
+	"private": true,
+	"main": "lib/index.js",
+	"jsnext:main": "src/index.js",
+	"scripts": {
```
matuzalemsteles: We can add a `build` task, so when someone runs `yarn build` at the root, we can perform the build of all the components. You can see how I did it on [ClayDatePicker](https://github.com/liferay/clay/pull/1440/commits/a96bad47531059edb4262950aa19ddb2d7907de6)

packages/clay-color-picker/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
matuzalemsteles: We can keep the information pattern in `package.json` equal to what we did in the other components. See [package.json from ClayDatePicker](https://github.com/liferay/clay/blob/9630c7157d9cb0b8c270a26353ceb630f0045c08/packages/clay-date-picker/package.json)

packages/clay-color-picker/src/index.js
```diff
@@ -0,0 +1,278 @@
+import React, {useEffect, useRef, useState} from 'react';
```
matuzalemsteles: I think we can continue to maintain the filename nomenclature, following the pattern of the other components. Something like `ClayColorPicker.js`.

packages/clay-color-picker/demo/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
matuzalemsteles: I think we should do the same here, to follow the pattern of the other components. Here's how I did it on [ClayDatePicker](https://github.com/liferay/clay/blob/9630c7157d9cb0b8c270a26353ceb630f0045c08/packages/clay-date-picker/demos/a11y.html).

It would also be interesting to continue creating the `a11y.html` page to follow the pattern.

packages/clay-color-picker/webpack.config.js
```diff
@@ -0,0 +1,20 @@
+const path = require('path');
```
matuzalemsteles: For the demo to work here without having to give `yarn start`, we will have to make some modifications here. Here's how I did it on [ClayDatePicker](https://github.com/liferay/clay/blob/9630c7157d9cb0b8c270a26353ceb630f0045c08/packages/clay-date-picker/webpack.config.js).

bryceosterhaus:
rewrote with typescript here, https://github.com/liferay/clay/pull/1706

Issue #1424, opened 2019-01-07
ClayCSS: Add Time Picker input


pat270:
https://issues.liferay.com/browse/LEXI-481
Create a timer picker component that mimics the interaction that the browser native time picker  has.
<img width="658" alt="time-picker" src="https://user-images.githubusercontent.com/788266/50795773-2eb35b00-1284-11e9-9e24-76276b709e4d.png">

PR #1440, opened 2019-01-15
Fixes #1483, #1484, #1482 - Clay Datepicker and Clay Timepicker

### LINKED ISSUE #1483: Implement Clay Date Picker

matuzalemsteles:
hey guys, from the rewrite of this component in React, we can withdraw from here the Button, Select, Icon and Dropdown components of Datepicker but I still do not know how we will proceed with rewriting the existing components.


matuzalemsteles:
Just a little extra information about the missing tests here: Since we are experimenting with hooks (not yet stable) and implementing the new components in React, we would need to prepare the tests infrastructure to be able to use enzyme for example, so a temporary solution would be to configure the tests in each package, but hooks still do not support shallow rendering. I'm following https://github.com/facebook/react/issues/14091.

packages/clay-datepicker/README.md
```diff
@@ -0,0 +1,24 @@
+# clay-datepicker
+
+ClayDatepicker
+
+## Setup
+
+1. Install NodeJS >= v0.12.0 and NPM >= v3.0.0, if you don't have it yet. You
```
wincent: A question: could the Node requirement be bumped to v6? That's the oldest LTS release. [v0.12 hit End-of-life](https://github.com/nodejs/Release) at the end of 2016.

wincent:
Seems this `clay-datepicker` should be `clay-date-picker` (and `ClayDatePicker`) for consistency with `clay-color-picker` (https://github.com/liferay/clay/pull/1427) and what seems to be the pattern in the existing components.

packages/clay-datepicker/src/Button.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+Button.defaultProps = {
+	monospaced: false,
+	style: 'primary',
+	type: 'button',
+};
+
+export {Button};
+export default Button;
```
wincent: What's the rationale for exporting twice here?

packages/clay-datepicker/src/Helpers.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * @return {Array<int>}
+ */
+export function range({start, end}) {
+	return Array.from(
+		{
+			length: end - start,
+		},
+		(v, k) => k + start
```
wincent: Neat!

packages/clay-datepicker/src/ClayDatepicker.js
```diff
@@ -0,0 +1,384 @@
...
+
+	/**
+	 * The input value.
+	 * @type {?(Date|undefined)}
+	 */
+	const [inputValue, setInputValue] = useState(undefined);
+
+	/**
+	 * An array of the weeks and days list for the current month
+	 * @type {!Array<Array>}
```
wincent: Not sure what the limitations of the type syntax here are, but can we specify the type of the nested array?

packages/clay-datepicker/src/ClayDatepicker.js
```diff
@@ -0,0 +1,384 @@
...
+		setInputValue(date);
+	}
+
+	/**
+	 * Handles the input change.
+	 * @param {!Event} event
+	 */
+	function inputChange(event) {
+		const {value} = event.target;
+		const format = `${dateFormat} ${time ? timeFormat : ''}`;
```
wincent: Might be nice to trim off the unwanted(?) trailing space in the event that `time` is falsey. eg.

```suggestion
		const format = `${dateFormat} ${time ? timeFormat : ''}`.trim();
```

packages/clay-datepicker/src/ClayDatepicker.js
```diff
@@ -0,0 +1,384 @@
...
+
+	/**
+	 * Set the format of how the time will appear in the input element.
+	 * See available: https://momentjs.com/docs/#/parsing/string-format/
+	 * @default HH:mm
+	 */
+	timeFormat: PropTypes.string,
+
+	/**
+	 * Describe a brief tip to help users interact.
```
wincent: I'm not sure what this means exactly. Can you clarify?

packages/clay-datepicker/src/ClayDatepicker.js
```diff
@@ -0,0 +1,384 @@
...
+
+	/**
+	 * Short names of days of the week to use in the header
+	 * of the month. It should start from Sunday.
+	 * @default [S, M, T, W, T, F, S]
+	 */
+	weekdaysShort: PropTypes.arrayOf(PropTypes.string),
+
+	/**
+	 * List of years available for navigate that are added in the selector.
```
wincent: Grammar nit:

```suggestion
	 * List of years available for navigation within the selector.
```

packages/clay-datepicker/src/ClayDatepicker.js
```diff
@@ -0,0 +1,384 @@
...
+							</div>
+						)}
+					</React.Fragment>
+				</Dropdown>
+			)}
+		</div>
+	);
+}
+
+const DateNow = new Date();
```
wincent: Suggestion: move this down close to where it is actually used.

packages/clay-datepicker/src/DateNavigation.js
```diff
@@ -0,0 +1,153 @@
...
+	 * years in the range
+	 * @param {!number} month
+	 */
+	function handleChangeMonth(month) {
+		const date = moment(currentMonth)
+			.clone()
+			.add(month, 'M')
+			.toDate();
+
+		if (memoizedYears.find(elem => elem.value === date.getFullYear())) {
```
wincent: Maybe some harmless micro-optimization to do here and pull the `getFullYear()` outside of the `find`? ie.

```javascript
const year = date.getFullYear();
```

packages/clay-datepicker/src/Dropdown.js
```diff
@@ -0,0 +1,60 @@
...
+	return (
+		<div className={classNames} ref={elementRef}>
+			{children}
+		</div>
+	);
+}
+
+Dropdown.propTypes = {
+	containerRef: PropTypes.object,
+	children: PropTypes.element,
```
wincent: nit: can we alphabetize this?

packages/clay-datepicker/src/Helpers.js
```diff
@@ -0,0 +1,81 @@
...
+			weekArray.push(week);
+			week = [];
+		}
+		week.push(day);
+		if (dayArray.indexOf(day) === dayArray.length - 1) {
+			weekArray.push(week);
+		}
+	});
+
+	// unshift days to start the first week
```
wincent:
```suggestion
	// unshift days from start the first week
```

packages/clay-timepicker/src/ClayTimepicker.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * ClayTimepicker
```
wincent: Same thought here about `clay-time-picker` and `ClayTimePicker`.

packages/clay-datepicker/src/ClayDatepicker.js
```diff
@@ -0,0 +1,384 @@
...
+								<Weekday weeks={weeks}>
+									{({day, key}) => (
+										<Day
+											day={day}
+											daySelected={daySelected}
+											key={key}
+											onClick={handleDayClicked}
+										/>
+									)}
+								</Weekday>
```
wincent: The naming of `<Weekdays />` and `<Weekday />` here is a little confusing because they both render plural things. Is one of these a header of and the other is the list of dates?

packages/clay-datepicker/src/Button.js
matuzalemsteles: hey @wincent, Actually we ended up leaving this as standard in having both, this ended up staying in our `generator-metal-clay`.

packages/clay-datepicker/README.md
matuzalemsteles: Yes, I do not see any problem in being able to go to v6. We just have to change in all to maintain consistency.

packages/clay-datepicker/src/ClayDatepicker.js
matuzalemsteles: Yes we can, but it will not be very readable.

packages/clay-datepicker/src/ClayDatepicker.js
matuzalemsteles: I think I'll do the same thing I did in **InputDate**, I think it's better than having to call `trim()` for this.

```javascript
const format = time ? `${dateFormat} ${time}` : dateFormat;
```

packages/clay-datepicker/src/DateNavigation.js
matuzalemsteles: Oh yeah, thanks!

packages/clay-datepicker/src/ClayDatepicker.js
matuzalemsteles: Yes. This can be a bit confusing, do you have any suggestions?

- `<Weekdays />` - Lists weekdays short
- `<Weekday />` - Lists the days of the month

packages/clay-date-picker/src/ClayDatePicker.js
```diff
@@ -0,0 +1,384 @@
...
+			// date is enabled. Unnecessary rendering and processing.
+			if (!useNative) {
+				setWeeks(currentMonth, firstDayOfWeek);
+			}
+		},
+		[currentMonth]
+	);
+
+	const classNames = classnames('input-group', {
+		[id]: id,
```
bryceosterhaus: Shouldn't this id be put on the element and not a className?

packages/clay-date-picker/src/ClayDatePicker.js
```diff
@@ -0,0 +1,384 @@
...
+								}}
+								onMonthChange={setCurrentMonth}
+								spritemap={spritemap}
+								years={years}
+							/>
+							<Weekdays
+								firstDayOfWeek={firstDayOfWeek}
+								weekdaysShort={weekdaysShort}
+							>
+								{({weekday, key}) => (
```
bryceosterhaus: You may want to add this plugin to catch sorting destructured keys. https://github.com/mthadley/eslint-plugin-sort-destructure-keys

packages/clay-date-picker/src/ClayDatePicker.js
```diff
@@ -0,0 +1,384 @@
...
+								{time && (
+									<TimePicker
+										currentTime={currentTime}
+										onTimeChange={setCurrentTime}
+										spritemap={spritemap}
+										timeFormat={timeFormat}
+									/>
+								)}
+								{!time &&
+									React.Children.only(
```
bryceosterhaus: I had no idea about this API, very cool

packages/clay-date-picker/src/ClayDatePicker.js
```diff
@@ -0,0 +1,384 @@
...
+	 * See available: https://momentjs.com/docs/#/parsing/string-format/
+	 * @default YYYY-MM-DD
+	 */
+	dateFormat: PropTypes.string,
+
+	/**
+	 * Set the initial value of the input.
+	 * @default undefined
+	 */
+	defaultValue: PropTypes.oneOfType([
```
bryceosterhaus: Just want to double check with you, but does this component function as an "uncontrolled" component? I want to make sure we using similar wording as react's API. They distinguish value vs defaultValue based on the nature of it being "controlled".

Reference: https://reactjs.org/docs/uncontrolled-components.html#default-values

packages/clay-date-picker/src/DateNavigation.js
```diff
@@ -0,0 +1,154 @@
...
+ */
+function DateNavigation({
+	currentMonth,
+	months,
+	onDotClicked,
+	onMonthChange,
+	spritemap,
+	years,
+}) {
+	const memoizedYears = useMemo(
```
bryceosterhaus: 🔥 nice job on memo'ing

packages/clay-date-picker/src/DateNavigation.js
```diff
@@ -0,0 +1,154 @@
...
+
+	const memoizedMonths = useMemo(
+		() =>
+			months.map((month, index) => {
+				return {
+					label: month,
+					value: index,
+				};
+			}),
+		months
```
bryceosterhaus: does this need to be in brackets?

packages/clay-date-picker/src/DateNavigation.js
```diff
@@ -0,0 +1,154 @@
...
+
+		if (memoizedYears.find(elem => elem.value === year)) {
+			onMonthChange(date);
+		}
+	}
+
+	/**
+	 * Handles the previous month button
+	 */
+	function handlePreviousMonthClicked() {
```
bryceosterhaus: Out of curiosity, how come you chose to use a function declaration vs an arrow function? For simple functions I tend to prefer arrow functions because it takes up less lines. I don't have strong opinions on in, mostly curious.

something like 
```js
const handlePreviousMonthClicked = () => handleChangeMonth(-1);
```

cc: @wincent what do you think?

packages/clay-date-picker/src/Helpers.js
```diff
@@ -0,0 +1,81 @@
...
+	const firstDayOfMonth = new Date(d.getFullYear(), d.getMonth(), 1, 12);
+
+	firstDayOfMonth.setMonth(firstDayOfMonth.getMonth() + 1);
+	firstDayOfMonth.setDate(firstDayOfMonth.getDate() - 1);
+
+	return firstDayOfMonth.getDate();
+}
+
+/**
+ * Based on https://github.com/gpbl/react-day-picker/blob/master/src/Helpers.js#L55
```
bryceosterhaus: 👍 great documentation!

packages/clay-datepicker/src/ClayDatepicker.js
wincent: Something like:

- `<WeekdayHeader />` : for the list of "S M T W T F S" (currently, this is `<Weekdays />`).
- `<Weekday />`: for each of "S", "M" etc (currently this is `<Week />`; super confusing!).
- `<DaysRow />`:  for each row of "1 2 3 4 etc" (currently this is `<Weekday />`; again pretty confusing for me).
- `<DayNumber />`, maybe: for each of "1", "2" etc (currently this is `<Day />`).

Maybe not those exact names, but in a case like this where all the terms are very similar and repetitive, I think it's better to bias towards names that are longer and more literal and descriptive, otherwise it is easy to get confused.

packages/clay-date-picker/src/DateNavigation.js
wincent: First answer, the abstract one:

Generally when working with teams I've found that simple rules (because they are easy to understand, easy to apply) are often the best. They tend to lead to the most consistent result, and consistency is super important (even more important than having something "optimal", especially when there is not much difference between the simple-and-consistent version and the "optimal" version).

So in the past, a pattern I've followed is `function` by default, and arrow functions where I want to lock in a `this` value (for example, in event handlers in React components). This pattern is simple enough to explain and follow. In side projects or smaller teams I am much more likely to be nuanced, following a heuristic like the one you're advocating here ("if very simple, turn into a one-liner") which involves an element of judgment and subjectivity.

Second answer, the concrete one:

In this instance I'd probably use an arrow function. 😂  The function is long and using something that is shorter is going to improve the readability of it.

packages/clay-datepicker/src/Button.js
wincent: I created https://github.com/liferay/clay/pull/1453 so that we have a place to discuss this further.

packages/clay-datepicker/src/ClayDatepicker.js
matuzalemsteles: Thanks a lot for your suggestions, I like these names but only I would change `<DaysRow />` to `<DaysTable />` since it renders the table in rows and columns, what do you think?

packages/clay-date-picker/src/Button.js
```diff
@@ -0,0 +1,62 @@
...
+	size,
+	spritemap,
+	style,
+	type,
+}) {
+	const classNames = classnames('btn', {
+		'btn-monospaced': monospaced,
+		[`btn-${size}`]: size,
+		[`btn-${style}`]: style,
+		[className]: className,
```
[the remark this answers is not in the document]
matuzalemsteles: The last time I used it this way came `undefined` in the className when it was not passed, but it's been a long time 😅. Thanks for that.

packages/clay-date-picker/src/ClayDatePicker.js
matuzalemsteles: Oh, that went unnoticed was not supposed to be there. Thanks!

packages/clay-date-picker/src/ClayDatePicker.js
matuzalemsteles: Thanks for the tip!

packages/clay-date-picker/src/DateNavigation.js
matuzalemsteles: Too, but as they are array would have no problem. But I'll add in brackets to keep a pattern here.

packages/clay-date-picker/src/DateNavigation.js
matuzalemsteles: 🙂 I do not have strong arguments against this haha, but I always set default `function` I think it was more customs. But this case really is simpler arrow function, now other cases I think it is just changing for each other, so I think we should decide if inside hooks we will always use `arrow functions` or just `function` just to maintain a consistency . makes sense?

packages/clay-date-picker/src/Helpers.js
matuzalemsteles: 😅 I can not tell if this was good or bad haha. But I think I should better document what this function does...

packages/clay-date-picker/src/DateNavigation.js
matuzalemsteles: Thanks 🙂

packages/clay-date-picker/src/ClayDatePicker.js
matuzalemsteles: In fact, I did not know, from that moment on, that subsequent updates were out of control, thanks for the link. But the reason for this API escapes the idea of ​​being uncontrolled just to set an initial value, the rest is controlled by the component and we issue the `onValueChange`. I think this will cause confusion for React components... I'm going to remove that.

I will follow the same line as you are doing in the color picker.

packages/clay-date-picker/src/Helpers.js
bryceosterhaus: You might want to add a brief description of "how" its based on that code. If it was a straight copy of that code, its probably fine to have just the link.

packages/clay-date-picker/src/DateNavigation.js
bryceosterhaus: I'm fine either way, we could also just say "use discretion" and not be super firm one way or the other. In the end I would likely favor readability the most.

matuzalemsteles:
hey @bryceosterhaus about my experience with Hooks: I liked hooks mainly because of the idea of ​​being able to reuse logic this is indisputably essential (In the components in Clay for example, using `class` we were already starting to repeat many logics in several components, hooks will help enough to remove this...), I think I have not used all the power that the hooks have to give better feedback on, but I think we're figuring out how it's fitting...

packages/clay-datepicker/src/Helpers.js
matuzalemsteles: I do not know if I understood correctly what you mean. 😅

packages/clay-datepicker/src/Button.js
matuzalemsteles: Great!!

wincent:
> hey @bryceosterhaus about my experience with Hooks: I liked hooks mainly because of the idea of ​​being able to reuse logic this is indisputably essential (In the components in Clay for example, using class we were already starting to repeat many logics in several components, hooks will help enough to remove this...),

Will be interesting to see how this plays out @matuzalemsteles, because I agree with you that the primary selling point of hooks is extracting reusable logic into shareable, testable units. But having said that, there are many times when we architect things for re-use and then we never actually re-use them 😂 — and it is very easy to fall into the mindset of "when you have a hammer, everything looks like a nail". Having said that though, if we started using hooks pervasively, even when re-use is not a motivation, we *might* still get some benefit from the simplicity of doing something consistently, always using a single pattern. But I don't know yet, so I'll be interested to see where things settle.

matuzalemsteles:
hey @wincent yes, this is a concern. I see some places that the hooks can help us here, such as `Dropdown`, we have the logic of keyboard navigation in Dropdown but we have another one for MultiSelect to cover the use cases of there... we can have with hooks only the navigation logic for example.

I think we have fairly strong reuse of components here whenever possible, as Lexicon uses the Atomic Design approach fortifies that too... I think hooks will help to fortify logic reuse of components (I hope...) but I speak at the point of view over non-public APIs.

But I'm not just basing on hooks just in that thought, I think hooks have a peculiar thing about you doing small micro optimizations (I know that while having this possibility it can make things more complex too...) that in the end are always welcome.

In my point of view, I think going with `class` would be using a lot of HOCs that although it is an interesting approach, it creates a lot of classes and maybe a giant tree of wrappers... the functions covers that side but at the same time we will have many functions... I think we will see how this will fit better with time and until there continue to maintain organization and consistency.

matuzalemsteles:
hey @pat270, I ended up extracting the CSS that I did in the demo to ClayCSS, when you can have some time, can you look at this? Thanks!

matuzalemsteles:
hey guys, one person raised a concern for me today about ClayTimePicker using the native `type="time"` browser only, this is not supported in Safari and IE, so we would have to support an alternate time for those browsers.

https://caniuse.com/#feat=input-datetime

cc @victorvalle @jbalsas

jbalsas:
Hey @matuzalemsteles, the idea is to only use that on mobile browsers. IOS Safari actually supports it and offers a way better UX than our component, so this should be fine. We will never use the native input in Desktop devices.

matuzalemsteles:
hey @jbalsas thanks for the explanations, it's clearer to me now. So I'll wait for the #1424 to markup to continue with ClayTimePicker.

pat270:
@matuzalemsteles For the CSS, I wanted to start using the Sass map + mixin pattern for all new components. I also wanted to remove rename some classes like `.date-picker-caption` to `.date-picker-navigation` and name the current `.date-picker-navigation` to something else. How should we go about doing this? Do you want me to send a pull to you with changes?

matuzalemsteles:
hey @pat270, you can send a PR to me with the changes.

matuzalemsteles:
hey @bryceosterhaus, I ended up removing `defaultProps` from hooks functions, were not working well with typescript, has a issue related to React types https://github.com/DefinitelyTyped/DefinitelyTyped/issues/30695 to solve this but it seems that React has an RFC that will depreciate the `defaultProps` for functions but nothing merged yet (https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components).

matuzalemsteles:
hey guys, I think now it's for a final review, we have the markups and CSS ready and I've rewritten for typescript and standardization according to what was defined in the generator.

packages/clay-date-picker/src/ClayDatePicker.tsx
```diff
@@ -0,0 +1,379 @@
...
+	 */
+	const handleDocClick = () => setExpanded(false);
+
+	useEffect(() => {
+		// It is not necessary to update the weeks when the native
+		// date is enabled. Unnecessary rendering and processing.
+		if (!useNative) {
+			setWeeks(currentMonth);
+		}
+	}, [currentMonth]);
```
wincent: Why do we need to use `useEffect` for this? If I am reading this correctly, this will lead to a possible double render on first render and whenever the `currentMonth` changes and `!useNative` (because first we'll render/update, then `setWeeks` will update the state and we'll render a second time).

packages/clay-date-picker/src/ClayDatePicker.tsx
```diff
@@ -0,0 +1,379 @@
...
+				</div>
+			</div>
+
+			{!useNative && (
+				<Dropdown
+					containerRef={elementRef}
+					expanded={expanded}
+					onDocumentClick={handleDocClick}
+				>
+					<React.Fragment>
```
wincent: In an earlier PR a couple months back I asked whether we could use `<>`/`</>` shorthand for fragments and there was some dependency in the toolchain that wasn't up-to-date enough to cope with it (sadly, I can't remember which one it was); is that still the case now?

packages/clay-date-picker/src/DateNavigation.tsx
```diff
@@ -0,0 +1,150 @@
...
+	/**
+	 * Handles the next month button
+	 */
+	const handleNextMonthClicked = () => handleChangeMonth(1);
+
+	/**
+	 * Handles the change of the year and month of the header
+	 */
+	function handleFormChange(event: ChangeEvent<HTMLSelectElement>) {
+		if (event.currentTarget.form === null) return;
```
wincent: Kind of surprised that we don't have an eslint rule that prohibits conditionals without braces.

packages/clay-date-picker/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "main": "lib/ClayDatePicker.js",
+  "esnext:main": "src/ClayDatePicker.js",
+  "jsnext:main": "src/ClayDatePicker.js",
+  "files": ["lib", "src"],
+  "scripts": {
+    "build": "babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "tsc --project ./tsconfig.declarations.json",
+		"prepublishOnly": "npm run build && npm run build:types",
+		"start": "webpack-dev-server --mode development",
+		"test": "jest --config ../../jest.config.js"
```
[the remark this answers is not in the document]
wincent: Hmmm. I thought I left a comment linking to [this discussion](https://github.com/liferay/liferay-npm-tools/issues/32#issuecomment-475329129) about using Prettier to format JSON and Markdown etc, but it seems that GitHub ate it? Sorry if this ends up being a dupe.

matuzalemsteles:
hey @wincent, about CSS I do not know what motives it will take @pat270 to use this way, maybe it will respond better than I do. Recently it is changing some things...

packages/clay-date-picker/src/ClayDatePicker.tsx
matuzalemsteles: I think the limitation was babel 6 but I think we can use it now. But I'll remove the Fragment is no longer needed here after we've changed the Date Picker markup. I forgot to remove 😅

packages/clay-date-picker/src/ClayDatePicker.tsx
matuzalemsteles: We need `currentMonth` treated https://github.com/liferay/clay/pull/1440/files?file-filters%5B%5D=.ts#diff-1753466c95969933de66a3c9603a812bR31 before updating weeks. Maybe we can bring this method to DatePicker and treat the date before sending it to `setCurrentMonth` but we will have to call in several places and so we can update both at the same time and remove this `useEffect`, the problem is that I have to always make sure to call `setWeeks` when calling `setCurrentMonth`. What do you think?

packages/clay-date-picker/src/ClayDatePicker.tsx
```diff
@@ -0,0 +1,379 @@
...
+	 */
+	weekdaysShort?: Array<string>;
+
+	/**
+	 * List of years available for navigation within the selector.
+	 */
+	years?: IYears;
+}
+
+const DateNow = new Date();
```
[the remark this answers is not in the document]
matuzalemsteles: Well thinking about this, maybe we can always make sure to clone `initialMonth` before sending to `onValueChange` https://github.com/liferay/clay/pull/1440/files?file-filters%5B%5D=.tsx#diff-ff7375578ff0fc5080ff0275edb83851R262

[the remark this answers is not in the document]
wincent: Here's another one: I open portal at 11:45 PM on December 31st, the navigate another page with a date picker at 12:15 AM. Because my system administrator kindly configured a long timeout, I wasn't logged out due to inactivity while I celebrated the arrival of the new year, and this new date picker now starts off showing the previous month.

(I know, I know... but it is late on Friday and my mind is wandering.) 😂

packages/clay-date-picker/src/ClayDatePicker.tsx
wincent: Makes sense. I think it's worth trying to refactor in order to get rid of `useEffect`; it seems important to get this pattern right early on because we wouldn't want double-rendering to be a thing we always do whenever we have dependent state.

pat270:
> I also see that we have this pre-existing pattern of calling map-get a zillion times to assign a value to a variable that is probably used only once; do you know why we don't just do it inline?

@wincent that is a good point. I ended up using this pattern early on because I did re-use variables in some mixins and wanted to keep the pattern consistent. I started to move away from re-using variables inside mixins because I thought it's too confusing/limiting for devs not familiar with the code. I'm still trying to find the sweet spot between customization and ease of use. I eventually want to document all the code using http://sassdoc.com/?

> There are a few lengthy values like `rgba(lighten(saturate(adjust-hue($secondary, -2), 5.48), 37.06), 0.5)` that get repeated in a few places which might be nice to DRY up into named variables instead.

The Lexicon Team gave me some names for the colors recently. We can change it to that.

packages/clay-date-picker/src/ClayDatePicker.tsx
matuzalemsteles: Yeah, I think we should start writing down these good practices on some of our style guides...

packages/clay-date-picker/src/ClayDatePicker.tsx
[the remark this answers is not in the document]
matuzalemsteles: Yeah, this is a very particular ​​problem 😅, so I think we should set an `initialMonth` whenever we have a call to `setExpanded` if it is not passed. It's the best place I think to add that. what do you think?

matuzalemsteles:
hey @pat270 adding SASSDOC would be great, maybe we could add that documentation to clayui.com automatically in some way. It's a great content.

packages/clay-date-picker/src/DateNavigation.tsx
bryceosterhaus: I think there is a CSF rule against it, which means we should probably add it to our `eslint-config-liferay`

wincent:
@pat270: thanks for the context!

packages/clay-date-picker/src/ClayDatePicker.tsx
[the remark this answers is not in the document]
wincent: The most natural change would be switching `initialMonth = DateNow,` to `initialMonth = new Date(),`, right? I don't know if it will work, because `useCurrentMonth` is a little confusing to me now that I look at it.

If I'm reading `useCurrentMonth` correctly, the value we pass for `initialMonth` is going to be used in a callback that is passed as the initial state in a `useState()` call — question: why aren't we setting the result of calling `setDate()` as the initial value? Instead, we pass a callback that sets it.

That is, why is it?

```javascript
const useCurrentMonth = (value: Date) => {
	const setDate = (date: Date) =>
		moment(date)
			.clone()
			.set('date', 1)
			.set('hour', 12)
			.set('minute', 0)
			.set('second', 0)
			.set('millisecond', 0)
			.toDate();

	const [currentMonth, set] = useState<Date>(() => setDate(value));

	const setCurrentMonth = (value: Date) => {
		set(setDate(value));
	};

	return [currentMonth, setCurrentMonth] as [Date, (value: Date) => void];
};
```

instead of:

```javascript
const useCurrentMonth = (value: Date) => {
	const setDate = (date: Date) =>
		moment(date)
			.clone()
			.set('date', 1)
			.set('hour', 12)
			.set('minute', 0)
			.set('second', 0)
			.set('millisecond', 0)
			.toDate();

	const [currentMonth, set] = useState<Date>(setDate(value));

	const setCurrentMonth = (value: Date) => set(setDate(value));

	return [currentMonth, setCurrentMonth];
};
```

(I am just guessing, but I get the impression that you might have needed that `as` cast only because of passing the function instead of the value.)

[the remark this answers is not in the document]
matuzalemsteles: I think even changing `initialMonth = new Date()` might not take effect in some cases, because we are only initializing the `currentMonth` state only when the component is created. We would have to make sure to update the `initialMonth` even when it is already mount, we would have a case of the person being midnight at work, but a few minutes before he left the page with the Date Picker open, with the 31/03 at midnight he back to the page, but there still is with the 31/03, in case it should be 01/04.

About `useCurrentMonth` I ended up removing it, but it reflects on the other Hooks we have. The use of `as` is to infer the return of the function, even if we do something like:
```javascript
const setCurrentMonth = (value: Date) => set(setDate(value));
```

TypeScript is not inferring the return from `set` to `setCurrentMonth`, so we have an inference of the wrong kind.
```
IDay[][] | ((value: Date) => void)
```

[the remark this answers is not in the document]
wincent: As a user, though, I don't expect a mounted component that is already on the page to flip months at midnight; what I do expect, however, is that if I navigate to a new view and see a date picker, on the new view it will always start on the current month.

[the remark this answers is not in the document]
matuzalemsteles: Yes the user would not wait so long with an open page. But it was a case that a person is with the page open a few minutes before midnight, although it is a case that may be very difficult to happen.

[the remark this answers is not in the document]
wincent: Yeah, but I mean independently of whether I have the page open 5 seconds/minutes or 5 days/weeks, I don't expect a rendered date picker to start rendering a different month without me explicitly skipping to another month. (On the other hand, if I navigate to a new view, I'd expect a new date picker to start on whatever month is the current month.)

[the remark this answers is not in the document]
matuzalemsteles: Oh yeah, you're right.

matuzalemsteles:
> Edit: Forgot to mention for the lengthy values, can we merge this and create a separate issue for that? It was already merged in 2.x-develop that way I can update both of them at the same time.

Can you take care of creating a separate issue? as soon as that merges with branch develop, you can work on it.

hey @wincent and @bryceosterhaus I'm going to go ahead with this PR for branch develop, I believe that from this PR we can get some components that will be created in # 1711, # 1710. If have more changes to this PR I'll take care care of updating, if anyone has any other opinion or change request feel free to comment here.

wincent:
> I'm going to go ahead with this PR for branch develop

Sounds like a good idea. GitHub isn't actually very good for code review, and the things that make it not good for code review only get worse the longer a PR lives and the more activity it has. 😭

matuzalemsteles:
Merged!

Issue #1448, opened 2019-01-17
ClayMultiSelect: When you press backspace in input to focus on the tag the browser returns to the previous page in Firefox


matuzalemsteles:

matuzalemsteles:
This is not really a bug but an expected behavior of Firefox, see this for more information https://support.mozilla.org/en-US/questions/1057630.

I'll mark this as a wontfix, if anyone has any idea of ​​how to circumvent this behavior please leave your thoughts.

4lejandrito:
Hi @matuzalemsteles,

I understand firefox has this behaviour when the user is not typing text. The problem is that ClayMultiSelect is basically a "smart" input. 

When I type in it and press enter a label is created and, since I can keep typing to add more labels, I also expect the backspace to remove what I typed. 

If you try https://react-select.com/home it behaves the way I expect on both browsers (Firefox and Chrome).

I still think this is a bug.

Thanks!

julien:
I'm re opening this issue since it also happens with V3
https://issues.liferay.com/browse/LPS-89052

Issue #1447, opened 2019-01-17
ClayMultiSelect: Comma creates a tag with a comma at the end


ambrinchaudhary:
Typing `mytag,` should create a tag called `mytag` but it creates `mytag,`.

PR #1449, opened 2019-01-17
Fixes #1446, #1447, #1448 - Adds the last typed character to char when data is empty and Remove the comma from the value (ADDED API)

### LINKED ISSUE #1446: ClayMultiSelect: Comma does not create tags in Firefox
Typing a comma should create a tag with the text entered. This works in Chrome but not in Firefox.

matuzalemsteles:


packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -110,15 +110,16 @@ class ClayAutocomplete extends ClayComponent {
...
 		if (this.enableAutocomplete) {
 			this._handleFilteredItems(event, query);
 		}
 
 		return !this.emit({
 			data: {
-				value: event.target.value,
-				char: event.data,
+				value: query,
+				char: char,
```
wincent: We could use concise notation here, even more so if `query` were named `value`:

```javascript
data: {
    char,
    value,
}
```

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -207,7 +207,7 @@ class ClayMultiSelect extends ClayComponent {
 	 * @return {Boolean} If the event has been prevented or not.
 	 */
 	_handleOnInput(event) {
-		const value = event.data.value;
+		const value = event.data.value.replace(',', '');
```
wincent: What will this do if I do something like:

* Type "foobar", then move the cursor left three times and type ","? (I'm expecting it will appear to swallow the keypress)
* Paste a string like "a,b,c"? (I'm expecting it will blow away only the first comma)

I'm thinking that maybe a stricter check might be required.

matuzalemsteles: hey @wincent, thanks for your concerns with these cases, really they will do what you expect. Do you have any suggestions about this?

jbalsas: So, what should happen if someone copy-pastes something like `a`, or `a,b` or `a , b , something,`?

jbalsas: So, shouldn't we autocreate tags on pasting? I was under the impression that that would be a preferred approach.

victorvalle: Talking with David A. we show that is better to have the functionality that detects all commas and creates at tag per element before a comma.
In this way:

- a , b , something, ---> [a][b][something] please scape white spaces when after the last char previous to comma
- a , b , something ---> [a][b]something

wincent: Sounds good @victorvalle: as a user I'd probably find that the most pleasant experience. 🎉

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -302,6 +310,23 @@ class ClayMultiSelect extends ClayComponent {
...
+	/**
+	 * Analyze the string and separate the values that
+	 * get a comma at the end.
+	 * @param {!string} string
+	 * @param {!string} separator
+	 * @return {Object}
+	 */
+	_tokenize(string, separator = ',') {
+		const hasLastComma = string.endsWith(separator);
+		const values = string.split(/\s*(?:,|$)\s*/).filter(token => token);
```
wincent: Pro-Tip™ for filter functions like this; if all you want is a truthiness filter, you can also write it like:

```suggestion
		const values = string.split(/\s*(?:,|$)\s*/).filter(Boolean);
```

I am not 100% sure it's better, but I like the way it is terser avoids creating a new function.

matuzalemsteles: Ah yes, that's better. Thanks

matuzalemsteles:
hey guys, I think this is ready to go forward.

matuzalemsteles:
hey @carloslancha, can you look at this when you have time?

carloslancha:

PR #1451, opened 2019-01-18
Fixes #1459 - Use SPDX license identifier in package.json files

### LINKED ISSUE #1459: Yarn warns about non-SPDX license specifier during install
We are using "BSD" as the license name in our Clay package.json files, which causes Yarn to emit warnings like this:

    warning package.json: License should be a valid SPDX license expression

We should switch it to "BSD-2-Clause", as listed [here on the SPDX website](https://spdx.org/licenses/BSD-2-Clause.html) which corresponds (somewhat loosely) to our LICENSE.md.

Already submitted PR #1451 for this but creating this issue retroactively to match the standard practice in this repo.

Labeling as "bug" as that seems to be the closest label we have for this kind of hygiene issue.

wincent:
We are using "BSD" as the license name in our Clay package.json files, which causes Yarn to emit warnings like this:

    warning package.json: License should be a valid SPDX license expression

We should switch it to "BSD-2-Clause", as listed [here on the SPDX website](https://spdx.org/licenses/BSD-2-Clause.html) which corresponds (somewhat loosely) to our LICENSE.md.

Note that because this commit updates "packages/generator-metal-clay/app/templates/_LICENSE.md", future generated packages will automatically benefit from this change.

Related: https://issues.liferay.com/browse/IFI-412

### DIFF
--- a/packages/browserslist-config-clay/package.json
+++ b/packages/browserslist-config-clay/package.json
@@ -4,7 +4,7 @@
   "description": "Clay components Browserslist Shared Config",
   "main": "index.js",
   "repository": "https://github.com/liferay/clay/tree/master/packages/browserslist-config-clay",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "keywords": [
     "clay",
     "browserslist",
--- a/packages/clay-alert/package.json
+++ b/packages/clay-alert/package.json
@@ -2,7 +2,7 @@
   "name": "clay-alert",
   "version": "2.6.0",
   "description": "Metal Clay Alert component.",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "repository": "https://github.com/liferay/clay/tree/master/packages/clay-alert",
   "engines": {
     "node": ">=0.12.0",
--- a/packages/clay-autocomplete/package.json
+++ b/packages/clay-autocomplete/package.json
@@ -2,7 +2,7 @@
   "name": "clay-autocomplete",
   "version": "2.6.0",
   "description": "Metal ClayAutocomplete component",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "repository": "https://github.com/liferay/clay/tree/master/packages/clay-autocomplete",
   "engines": {
     "node": ">=0.12.0",
--- a/packages/clay-badge/package.json
+++ b/packages/clay-badge/package.json
@@ -2,7 +2,7 @@
   "name": "clay-badge",
   "version": "2.6.0",
   "description": "Metal Clay Badge component.",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "repository": "https://github.com/liferay/clay/tree/master/packages/clay-badge",
   "engines": {
     "node": ">=0.12.0",
--- a/packages/clay-button/package.json
+++ b/packages/clay-button/package.json
@@ -2,7 +2,7 @@
   "name": "clay-button",
   "version": "2.6.0",
   "description": "Metal Clay Button component.",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "repository": "https://github.com/liferay/clay/tree/master/packages/clay-button",
   "engines": {
     "node": ">=0.12.0",
--- a/packages/clay-card-grid/package.json
+++ b/packages/clay-card-grid/package.json
@@ -2,7 +2,7 @@
   "name": "clay-card-grid",
   "version": "2.6.0",
   "description": "Metal ClayCardGrid component",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "repository": "https://github.com/liferay/clay/tree/master/packages/clay-card-grid",
   "engines": {
     "node": ">=0.12.0",
--- a/packages/clay-card/package.json
+++ b/packages/clay-card/package.json
@@ -2,7 +2,7 @@
   "name": "clay-card",
   "version": "2.6.0",
   "description": "Metal ClayCard component",
-  "license": "BSD",
+  "license": "BSD-2-Clause",
   "repository": "https://github.com/liferay/clay/tree/master/packages/clay-card",
   "engines": {
     "node": ">=0.12.0",
--- a/packages/clay-charts-react/package.json
+++ b/packages/clay-charts-react/package.json
@@ -13,7 +13,7 @@
 		"test": "jest"
 	},
 	"author": "Bryce Osterhaus",
-	"license": "BSD",
+	"license": "BSD-2-Clause",
 	"dependencies": {
 		"clay-charts-shared": "^2.6.0",
 		"react-billboardjs": "^1.4.3"
--- a/packages/clay-charts-shared/package.json
+++ b/packages/clay-charts-shared/package.json
@@ -12,7 +12,7 @@
 		"copyAssets": "ncp src/svg lib/svg"
 	},
 	"author": "Bryce Osterhaus",
-	"license": "BSD",
+	"license": "BSD-2-Clause",
 	"dependencies": {
 		"d3": "^5.7.0",
 		"metal": "^2.16.6"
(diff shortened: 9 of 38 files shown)


wincent:
Just changed the base from "master" to "develop" and it applies cleanly but I'm going to double check and make sure there aren't any substantive differences between the two branches that would affect this commit; if necessary I'll re-roll it.

matuzalemsteles:
hey @gregory-bretall, Thanks for this change, I would recommend @jbalsas more to look at this.

jbalsas:
I think @wincent is right with this change, so we can go ahead and merge it. 

For the future, we should apply the new guidelines that @silverhook outlined in https://loop.liferay.com/es/home/-/loop/feed/12995353

@silverhook, would you mind guiding us through the process? I know you've done this a couple of time, but I hope we can see it through this time and apply it ourselves to our other projects.

Thanks!

wincent:
Rebased and updated commit message to reference the issue that I just created (#1459).

silverhook:
Hey! Glad to see this being tackled :D

# s/BSD-2-Clause/BSD-3-Clause

The license text that I find in the repo is actually closer to `BSD-3-Clause` (the hint is also in the number of bullet-points) – please use that instead.

You can check the diff to standard SPDX licenses simply from your browser using this useful tool: https://github.com/spdx/spdx-license-diff/

# General stuff

> Note that because this commit updates "packages/generator-metal-clay/app/templates/_LICENSE.md", future generated packages will automatically benefit from this change.

Sounds really cool. How would changing the lines affect that template though?

# Applying REUSE.software best practices

> For the future, we should apply the new guidelines that @silverhook outlined in https://loop.liferay.com/es/home/-/loop/feed/12995353
> 
> @silverhook, would you mind guiding us through the process? I know you've done this a couple of time, but I hope we can see it through this time and apply it ourselves to our other projects.

As that Loop post is not reachable from the outside, here is what it’s based on:

<https://reuse.software/>

Basically, the policy I talked about there is an implementation of the REUSE.software best practices. If you need to point people outside Liferay to the policy, you can simply direct them to REUSE.software (we’re helping with the practices as well, BTW). The specs/guides are quite well written (and if not, do [submit a bug report](https://github.com/reusesoftware/reuse/issues), so we can fix it together) and has some useful software as well – e.g. a linter.

In short, what would need to be done for this repo to be REUSE-compliant (and thus with our internal policies would be to 

## Include licensing header in source code

Prepend the following comment (feel free to use a commenting format that takes less lines) in all the source code:

``` js
/**
 * © 2018 Liferay, Inc. <https://liferay.com>
 *
 */
```

… just change the year to whatever year that file was in (if that’s too much work, `2014` as the year of the first commit should do). And then do *not* bother with updating the year in the future. When you create new files, just put the year when you created it.

## Include the license text

It should be perfectly enough to have the license text just in one location:

- either in `LICENSES/LicenseRef-BSD-3-Clause-Liferay.txt` – especially if there are other licenses in the repo (BTW, the `LicenseRef-` prefix is used in SDPX to indicate an non-standard license, we use it here together with the `-Liferay` suffix, because the BSD licenses need to be modified to include the copyright holder);
- or if it really just one license and we don’t expect to pull code that’s under a different license it can simply be in `LICENSE.txt`

In both cases the license text should read (this should be a SPDX-compatible version of what we already have):

```
Valid-License-Identifier: LicenseRef-BSD-3-Clause-Liferay
Valid-License-Identifier: BSD-3-Clause
License-Text:

Copyright (c) 2014, Liferay Inc. All rights reserved.

Redistribution and use in source and binary forms, with or without modification,
are permitted provided that the following conditions are met:

1. Redistributions of source code must retain the above copyright notice,
this list of conditions and the following disclaimer.

2. Redistributions in binary form must reproduce the above copyright notice,
this list of conditions and the following disclaimer in the documentation
and/or other materials provided with the distribution.

3. Neither the name of the copyright holder nor the names of its contributors
may be used to endorse or promote products derived from this software without
specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```


The thing is that if you use SPDX identifiers, it is already understandable which license we’re talking about, so no need to have it everywhere. But if the individual packages are often distributed separately or the license template in the generator has some other use, do let me know and we’ll think about this.

## Fix the README to point to the right license

Use either the SPDX’:
- Full name `BSD 3-Clause "New" or "Revised" License`; or easier still
- Short identifier: `BSD-3-Clause`

Ideally with a link to the license file.

## Run the `reuse` linter

Since this is the first project to try this, please run the `reuse` linter (at least `a0a6e398 ` commit or newer) on it and let me know of the outcome (or ping me, so I run it myself).

wincent:
@silverhook: Thanks for catching my mistake. Issue: #1461 and corresponding pull: #1462. Sorry for the noise.

> > Note that because this commit updates "packages/generator-metal-clay/app/templates/_LICENSE.md", future generated packages will automatically benefit from this change.
>
> Sounds really cool. How would changing the lines affect that template though?

That was a copy-paste mistake of the path; what I actually meant to write was:

> Note that because this commit updates "packages/generator-metal-clay/app/templates/**_package.json**", future generated packages will automatically benefit from this change.

Going to try the reuse linter now...

silverhook:
> Hey @silverhook, We currently publish our separate packages, so I do not know if we should include the license only in one place.

That’s a good reason then. In that case let’s just replicate the same set-up as in the root directory, also in the individual package directories, so it’s easy to find.

I suspect that is what the `packages/generator-metal-clay/app/templates/_LICENSE.md` is for, right? If so, it makes sense to do it through that.

> Hey @gregory-bretall I think we can start adding to the licensing header in source code in the components we are creating now (ClayDatePicker and ClayColorPicker) and separate another issue to work on the rest and do the suggested by @silverhook, what do you think ?

FWIW, I’m OK with taking this slowly. This would be the first project that we try to make comply with REUSE, so no haste is needed.

wincent:
> FWIW, I’m OK with taking this slowly. This would be the first project that we try to make comply with REUSE, so no haste is needed.

@silverhook: one pain point with the reuse linter is it requires Python 3.5 which may be a bit of a barrier. I wonder if we could get partway there with some a custom ESLint plug-in (or maybe one already exists), which would work with our existing toolchain. We already have a "eslint-config-liferay" package where that could be added, perhaps.

silverhook:
@wincent I’m pretty sure there is no other linter out there (yet) that checks for compliance with REUSE specs, so that would mean porting it ESLint.

Alternatively, I can run it locally in `venv`, if you point me to the repo and commit where it’s applied.

wincent:
@silverhook: I found an ESLint plug-in ([eslint-plugin-notice](https://www.npmjs.com/package/eslint-plugin-notice)) that would get us at least part of the way there and [set it up in a side project](https://github.com/wincent/js/commit/ec0103cc849b9f97a2e5587cab43b29d0f1b9bd8). Not sure if it would be adequate or not, but you can do things like define a pattern to check for the presence of the header, supply a template for adding a missing header, control whether that template gets prepended, appended, or overwrites an existing header, customize the error message that is shown for non-compliance etc.

PR #1452, opened 2019-01-18
Fixes #1458 - Re-enable and update generator-metal-clay tests

### LINKED ISSUE #1458: Re-enable and update generator-metal-clay tests
These were disabled in 2017 in commit 844f56f without explanation. I want to propose a change to the generator, so must first make the tweaks necessary here to get these tests running and passing again.

I actually already submitted #1452 for this, but creating this issue retroactively to conform to the usual pattern in this repo.

wincent:
These were disabled in 2017 in commit 844f56f35006377911427bd8ae325df3c without explanation. I want to propose a change to the generator, so I've made the tweaks necessary here to get these tests running and passing again.

Tested with `yarn test`.

Related: https://issues.liferay.com/browse/IFI-420

### DIFF
--- a/packages/generator-metal-clay/__tests__/test-app.js
+++ b/packages/generator-metal-clay/__tests__/test-app.js
@@ -4,13 +4,13 @@ let path = require('path');
 let assert = require('yeoman-generator').assert;
 let helpers = require('yeoman-generator').test;
 
-describe.skip('metal-clay-generator:app', function() {
+describe('metal-clay-generator:app', function() {
 	beforeAll(function(done) {
 		helpers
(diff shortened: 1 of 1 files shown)


wincent:
@carloslancha: I changed base from "master" to "develop".

matuzalemsteles:
This sounds good to me, I do not remember why we deactivated the tests, but I do not see any problem reactivating them. I would say the same comment I put on #1454 so we can move on.

wincent:
@matuzalemsteles: This one should be ready to go now too; I created #1458 and linking to it from commit message and the PR title.

PR #1456, opened 2019-01-18
Fixes #1455 - Add default type in ClayButton.soy

### LINKED ISSUE #1455: Add default type in Claybutton
The default type attr works fine in js (via metal config) but is missing in soy. 
Add default to soy to avoid differences between server and client render.

boton:


carloslancha:
hey @boton you need to regen the test snapshots and the expected soy output for isomorphic tests. Thanks!

boton:
@carloslancha I will try.
Thanks!

matuzalemsteles:
Merged, Thanks!

PR #1474, opened 2019-01-24
Fixes #1472 Fixing existing tests

### LINKED ISSUE #1472: multi-select, autocomplete and dataprovider should request data based on query
At the moment the data is only fetched on load and with an optional polling interval. It is always fetched without taking the user input into account.

It should fetch every time the input changes (debounced). Also polling is quite dangerous because it will cause unnecessary performance impact on the server.

We need this for https://issues.liferay.com/browse/LPS-87558.

4lejandrito:


matuzalemsteles:

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -109,10 +111,10 @@ class ClayAutocomplete extends ClayComponent {
 	 * @return {Boolean} If the event has been prevented or not.
 	 */
 	_handleOnInput(event) {
-		const query = event.target.value;
+		this._query = event.target.value;
 
 		if (this.enableAutocomplete) {
-			this._handleFilteredItems(event, query);
+			this.refs.dataProvider.updateData(this._query);
```
matuzalemsteles: hey @4lejandrito, we have to be careful for use cases that do not need a query in the URL or that do not need to make requests to each user input, in this way will always request a request unnecessarily.

packages/clay-data-provider/src/ClayDataProvider.js
```diff
@@ -257,15 +268,6 @@ ClayDataProvider.STATE = {
...
 	}),
 
-	/**
-	 * Flag to define how often to refetch data (ms)
-	 * @instance
-	 * @default 0
-	 * @memberof ClayDataProvider
-	 * @type {?(number|undefined)}
-	 */
-	requestPolling: Config.number().value(0),
```
matuzalemsteles: hey @4lejandrito I understand that for some use cases this would not do much feeling, but we may have cases where you want the data to be updated every minute for example... with DataProvider we intend to use it in other places, such as ClayCharts. I would like to keep this API to cover these cases.

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -18,6 +18,8 @@ class ClayAutocomplete extends ClayComponent {
 	 * @return {Boolean} If the event has been prevented or not.
 	 */
 	_handleDataChange(event) {
+		this._handleFilteredItems(event);
```
matuzalemsteles: If we always keep the call to `updateData` when the query changes, we will have the use case when we do not want to make a request for each query change, in this case we will always be issuing the `dataChange` event without them being altered

packages/clay-data-provider/src/ClayDataProvider.js
```diff
@@ -128,23 +122,23 @@ class ClayDataProvider extends ClayComponent {
...
 		this._isResolvedData = false;
-		this._pollingInterval = 0;
 		if (this._hasData(this.dataSource)) {
 			this._dataSource = this.dataSource;
 			this._isResolvedData = true;
 		} else {
-			this._fetchData();
+			this.updateData('');
 		}
+		this.updateData = debounce(
```
matuzalemsteles: I do not know if `debounce` is necessary when we do not want to make a request every time the query changes.

4lejandrito:
Hi @matuzalemsteles,

Thanks for looking into this.

I was under the impression that DataProvider was only used by the autocomplete. I understand you might want to cover future use cases with it, but I think you're expecting too much from a single component. Also, autocompletion requirements are very different from those of a dynamic _real time_ chart. 

The reason of our changes is that we need to query a backend service that provides autocompletion results based on the user input. The best way to provide accurate results is if we query the server every time the user stops typing for a small fraction of a second (this is what we use debounce for).

If we want to support charts and autocomplete, I would personally recommend having 2 different components. This way we wouldn't have dependencies between the 2 modes of operation and we would be able to evolve both components separately. Otherwise we might end up breaking autocomplete each time we need a new feature for charts and vice-versa.

I am happy to work on my changes to try to keep both modes working at the same time, but I think it is a waste of our efforts since at the moment there is no other consumer of DataProvider. 
 
What are your thoughts about this?

matuzalemsteles:
hey @4lejandrito, I understand your point of view, the DataProvider was extracted from Autocomplete initially, we separated it to start evolving to support our other components ((See https://github.com/liferay/clay/pull/1292#discussion_r235650633)), initially it is being used only by Autocomplete and MultiSelect .

Please, I am not disagreeing with the use of debounce or best practices to do autocomplete, instead I am thanking you for your contribution. But we need to make sure that this component is not designed for just the use case of Autocomplete, for example.

My recommendation to follow these changes is to adjust the modifications to support the use case of when we do not want to make a request for each query change, we can enable this functionality by adding a new API. What do you think? Thanks!

4lejandrito:
Hi @matuzalemsteles,

Thanks for your response. I've had a chat with @jbalsas and they will take over this. Feel free to get back to me if I can be of any help.

Regards!

packages/clay-data-provider/src/ClayDataProvider.js
```diff
@@ -14,31 +13,38 @@ import templates from './ClayDataProvider.soy.js';
...
 	 * @protected
 	 */
-	_fetchData(requestRetries = 0) {
-		if (
-			this.initialData &&
-			!this._pollingInterval &&
-			requestRetries === 0
-		) {
+	updateData(query, requestRetries = 0) {
+		if (this._hasData(this.dataSource)) {
```
carloslancha: @4lejandrito is this needed? We're already doing this in the `created` method.

4lejandrito: Hi @carloslancha, see https://github.com/carloslancha/clay/pull/8. With the last commit we can remove that if statement. Now we filter the items every time the query changes or every time the data changes.

4lejandrito: I think I accidentally pushed the commit to this PR too...

PR #1485, opened 2019-01-28
POC for first step to moving towards v3.x with React.

bryceosterhaus:
Thought I would start a PR for v3.x and we can further discuss on here.


wincent:
I think `tabWidth` won't affect the content of the files, because `"useTabs": true` is set. So the actual number is immaterial. (We should try to be consistent anyway, of course.)

README.md
```diff
@@ -12,7 +12,7 @@ The pillar to give colors and structure to Clay, where is the css and examples o
```
matuzalemsteles: hey @bryceosterhaus, It would be interesting to add a warning about the v3 version and explain about the branch v2 with components in Metal...

bryceosterhaus: Good idea! Ill add that in.

bryceosterhaus:
> I think it might be useful to keep the metal components around for testing purposes as we do this

I'm not totally sure how this would work out or if its practical, mainly because the focus and intent is not necessarily to keep the API line and compatible with the previous soy versions of these components. It could be interesting to keep the snapshot files to test against though. Any chance you have an example of how you did it at your last job? I'm sure if I saw an example I could get my head around it a bit more.

README.md
bryceosterhaus: Added the reference under the `Clay Components` header. Let me know if you think I should add more to it.

bryceosterhaus:
I also added a package that contains generic boilerplate for new react components

wincent:
> Any chance you have an example of how you did it at your last job? I'm sure if I saw an example I could get my head around it a bit more.

Unfortunately nothing open source, but we basically had a way of relating components together that were implemented in two languages, for example:

* `FancyButton.php`: implementation A.
* `FancyButton.examples.php`: usage examples that implement a `render` method.
* `FancyButton.react.js`: implementation B.
* `FancyButton.react.examples.js`: usage examples that implement a `render` method.

The files didn't literally need to be next to each other (they could be in different folders); we just needed a way of mapping them to one another.

So the test runner would just render all the examples and compare the resulting DOMs. In other words, no behavioral changes, just verifying the structure of the examples.

Back before React components could return fragments, you sometimes had to wrap things in a `<div>` or `<span>` that you didn't really need in the DOM for any semantic or logical purpose, but because it was the only way to return a collection of things from a component. So this meant we also needed a syntax for specifying DOM elements that would be ignored for the purposes of comparison. I don't think you'd need this now though.

One unrelated but nice thing about the example files is they served as a basis for populating a browsable component gallery.

jbalsas:
Thanks for starting this conversation @bryceosterhaus!

Since we're set on exploring this option, I think it should be about time to:
- Rename `master` branch to `2.x`
- Rename `develop` branch to `2.x-develop`
- Re-target this PR agains `develop` ;)

As @wincent mentioned, it's important that we move with care here. If we don't, we risk causing a lot of problems when teams start migrating to new components. One way or another we should be able to keep track of:
- Markup changes -> avoid at all costs (or as much as possible)
- API changes -> keep proper track of differences and document upgrade path for new ones

README.md
matuzalemsteles: Thanks, I would be more "aggressive" with the information, so as not to go unnoticed to someone already know.

I would put something like this below line 3 just as an add-on:
```
🚨This Branch contains the Beta version of Clay v3, if you are looking for the components in Metal.js visit the 2.x branch.
```

That's just a suggestion, but something we should maybe use more, Clay's blog, publish these kinds of decisions we make and the reasons, I think it's important to maintain our relationship with the devs.

README.md
```diff
@@ -92,55 +66,58 @@ If you want to contribute to this project and do not know where to start [good f
 
 2. Install global dependencies:
```
matuzalemsteles: hey @bryceosterhaus, I think we can remove those steps and just point to the [CONTRIBUTING.md](https://github.com/liferay/clay/blob/develop/CONTRIBUTING.md).

We can put the information here for those who want to contribute with version v2, something like this:
```
### Contributing to Clay v2

For contributions with Clay v2, go to branch v2 and follow the [contribution guide](https://github.com/liferay/clay/blob/2.x/CONTRIBUTING.md).
```

README.md
```diff
@@ -152,6 +129,7 @@ yarn site
 ```
```
matuzalemsteles: hey @bryceosterhaus, With the migration guide going to be a WIP process, we can add a category informing about the migration guide pointing to the documentation soon.

Something like:
```
## Migration Guides
Do you already have the Clay v2 components implemented in your project? These handy guides will help you perform the migration from Clay v2 to v3.

- [Migrate the Clay components from v2 to v3](https://next.clayui.com/docs/guides/migrating-from-v2-to-v3)
```

matuzalemsteles:
hey @jbalsas, If you can do that, I would be grateful, I do not see problems at least for me 🙂. I do not want to risk messing things up here 😅

bryceosterhaus:
@jbalsas I don't think I have permissions to do any branch related stuff so I will defer to letting you set that up. I think as long as we have the correct versions on the right branches, we should be able to start pretty bare with 3.x and then we can add more as we go, such as snapshots of the markup to compare against.

wincent:
Just to clarify about renaming branches. You can't "rename" a branch in a remote repo (branches are immutable, unless force-pushed). You *can* rename a local branch. What @jbalsas is suggesting is that we create new branches.

Current graph:

```
  * develop
  |
 /
* master
|
|
```

Create `2.x` maintenance branch from current master with `git checkout master && git checkout -b 2.x && git push origin 2.x`:

```
  * develop
  |
 /
* master, 2.x
|
|
```

Create `2.x-develop` branch from current `develop` with `git checkout develop && git checkout -b develop && git push origin 2.x-develop`:

```
  * develop, 2.x-develop
  |
 /
* master, 2.x
|
|
```

Now, maintenance commits go on `2.x-develop` and new work goes on `develop`:

```
  * develop
  |
  * React work
  |
  | * 2.x develop
  | |
  | * maintenance work 
  |/
  *
  |
 /
* master, 2.x
|
|
```

jbalsas:
So, I just created the new branches:
- [2.x](https://github.com/liferay/clay/tree/2.x)
- [2.x-develop](https://github.com/liferay/clay/tree/2.x-develop)

I haven't touched `develop` or `master` yet. The way I see these 2 changing is:
- `develop` will be updated and mean "3.x-develop" as soon as we merge this in
- `master` will mean `3.x` as soon as we do our initial `3.x` release

bryceosterhaus:
@wincent great description! This all sounds good to me. 

Let me know if there is anything else wanted or needed in this PR.

CONTRIBUTING.md
```diff
@@ -4,32 +4,24 @@ Want to contribute to Clay? There are a few things you need to know before you s
...
-	- [Proposing a Change](#proposing-a-change)
-	- [Sending a Pull Request](#sending-a-pull-request)
-- [Style Guides](#style-guides)
-	- [Git Commit Messages](#git-commit-messages)
-	- [JavaScript Style Guide](#javascript-style-guide)
-	- [CSS Guidelines](#css-guidelines)
-	- [Documentation Style Guide](#documentation-style-guide)
-- [Additional Notes](#additional-notes)
-	- [Issue Labels](#issue-labels)
+-   [What Should I Know Before I Get Started?](#what-should-i-know-before-i-get-started) - [Clay and Packages](#clay-and-packages)
```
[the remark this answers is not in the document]
bryceosterhaus: Good catch. It got messed up when prettier ran

bryceosterhaus:
@matuzalemsteles just updated and added the licensing stuff. I didn't look into adding it to clayui or clay-css though. Wasn't sure if those are different use cases.

As for the a yeoman generator vs the boilerplate, I think it'd be a good idea to create that. I might wait until after this initial PR though. I want to get a pretty generic base layer before adding. Once its merged I can start working on a generator to replace the boilerplate.

wincent:
@bryceosterhaus: Our internal ticket for the REUSE licensing stuff is https://issues.liferay.com/browse/IFI-425 — taking these steps in Clay is a great beginning. 👏

bryceosterhaus:
Updated commit messages and pushed

PR #1486, opened 2019-01-29
Fixes #1479 - Adds new API allowInputCreateItem

### LINKED ISSUE #1479: ClayMultiSelect: adding item only from autocomplete list or "select" button
There are cases in which we want to add items only from trusted sources. This means, we want to avoid that use could introduce anything in the multiselect (typing Enter or ","), and we want to allow only items selected from the Autocomplete list or the "select" button.
See use case in portal: https://issues.liferay.com/browse/LPS-87549 
UX proposal will be uploaded in https://issues.liferay.com/browse/LPS-89900

matuzalemsteles:


packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -213,11 +213,14 @@ class ClayMultiSelect extends ClayComponent {
...
-				event
-			);
+			if (this.allowInputCreateItem) {
+				return this._handleItemAdded(
+					value,
+					this._getItemSchema(value, value),
+					event
+				);
+			}
+			break;
```
wincent: With this change you are no longer returning a boolean in the event that the creation is disallowed, but the JSDoc says that we return a bool.

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -330,6 +337,15 @@ class ClayMultiSelect extends ClayComponent {
...
  */
 ClayMultiSelect.STATE = {
+	/**
+	 * Flag to indicate whether input can create item.
+	 * @default false
+	 * @instance
+	 * @memberof ClayMultiSelect
+	 * @type {?bool}
+	 */
+	allowInputCreateItem: Config.bool().value(true),
```
wincent: Naming suggestion: should this be of the form `allow${NOUN}`? Something like `allowItemCreation`? (I'm not familiar with our naming conventions yet, but `allowInputCreateItem` struck me as a bit awkward.)

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -243,7 +246,11 @@ class ClayMultiSelect extends ClayComponent {
 			event.preventDefault();
 			if (this._itemFocused) {
 				return this._handleItemRemoved(this._itemFocused);
-			} else if (value && event.data.eventFromInput) {
+			} else if (
+				value &&
+					event.data.eventFromInput &&
+					this.allowInputCreateItem
+			) {
```
wincent: I'm not sure about the UX here. From the user's perspective it may seem like the input is broken because hitting Enter has no visible effect.

matuzalemsteles: hey @wincent, Thanks for pointing this out, this really can be an impact to the user, Lexicon does not say something about it, but there seems to be use cases that this should be blocked, as commented in issue #1479, but I would like to know more about the Lexicon team, to get on with it.

cc @victorvalle

packages/clay-multi-select/src/ClayMultiSelect.js
matuzalemsteles: Yes this may seem a bit confusing, but I just want to point out that input can create items. But we can create items from autocomplete or select, for example... Maybe that name was not so clear. I think that just going with `allowItemCreation` might seem like we will disable the creation of items in any user interaction. Makes sense 🤔?

packages/clay-multi-select/src/ClayMultiSelect.js
matuzalemsteles: Oh yeah, you're correct about that. thanks.

packages/clay-multi-select/src/ClayMultiSelect.js
wincent: Not sure if too verbose, but maybe `allowItemCreationFromInput`?

matuzalemsteles: I do not see problem in going with that name despite being large but it is clearer and API is self explanatory.

ambrinchaudhary:
Hey @matuzalemsteles @wincent: 
Thank you for looking into this issue. The thing is I am working with UX team to define the best approach here, and once decided I will take care of implementing it. I'm sorry I didn't reflect this in the issue. But I will take in mind the work and review done here and continue from it.

matuzalemsteles:
hey @ambrinchaudhary, I think I misunderstand, but will you implement that in Clay? If so, we can close this PR and wait for the definition of the UX Team.

ambrinchaudhary:
Yes @matuzalemsteles, I will implement it in Clay, based on what you have already done and completing it with the UX definition. 
Thank you!

matuzalemsteles:
ok, Thanks @ambrinchaudhary, I will close this PR and let me know if you need my help.

PR #1488, opened 2019-01-29
Fixes #1487 Initialize contentRenderer attribute by default

### LINKED ISSUE #1487: ClayMultiSelect: contentRenderer not intialized.
The param contentRenderer should be initialized because if we do not provide a value (so contentRenderer is null) it does not pass the tests and occurs an error in the component rendering.

LuismiBarcos:


matuzalemsteles:
Merged, thanks @LuismiBarcos!

PR #1497, opened 2019-02-01
Fixes #1472 - Add support for debounce and for dataSource functions (BREAKING CHANGE)

### LINKED ISSUE #1472: multi-select, autocomplete and dataprovider should request data based on query

carloslancha:


matuzalemsteles:

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -281,7 +305,13 @@ ClayAutocomplete.STATE = {
 	 * @memberof ClayAutocomplete
 	 * @type {?bool}
 	 */
-	enableAutocomplete: Config.bool().value(true),
+	enableAutocomplete: Config.validator(value => {
+		if (value) {
```
matuzalemsteles: hey @carloslancha, I know we're issuing this warning not to cause a change break. Maybe we should keep this API working too so we do not break the applications of who it is using... makes sense?

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -242,19 +253,32 @@ ClayAutocomplete.STATE = {
...
+		Config.func(),
+		Config.object(),
+		Config.string(),
 	]).required(),
 
+	/**
+	 * Set the request debounce time
+	 * @instance
+	 * @default 200
+	 * @memberof ClayDataProvider
```
matuzalemsteles: Just certified to go according to the other standards we have. `@memberof ClayAutocomplete`

packages/clay-autocomplete/src/ClayAutocomplete.js
carloslancha: Since to use your own filtering strategy you must be listening the event and passing `filtereItems` again to the components... I think nobody would be just passing `enableAutocomplete` without listening, so this is probably useless and will not break a thing.

matuzalemsteles: Oh yeah, good point, really this will not break. Thanks @carloslancha.

carloslancha:
@matuzalemsteles @wincent requested changes sent. Thanks!

matuzalemsteles:
Merged, thanks!

PR #1518, opened 2019-02-08
Only render dropdown when expanded

travisrcory:
Only render dropdown items when expanded to improve performance.

### DIFF
--- a/packages/clay-dropdown/src/ClayDropdownBase.soy
+++ b/packages/clay-dropdown/src/ClayDropdownBase.soy
@@ -9,6 +9,7 @@
 	{@param? _handleButtonClick: any}
 	{@param? _handleItemClick: any}
 	{@param? _handleSearch: any}
+	{@param? _hasExpanded: any}
 	{@param? button: ?}
 	{@param? caption: string}
 	{@param? contentRenderer: string}
@@ -57,6 +58,7 @@
 			{param _handleButtonClick: $_handleButtonClick /}
 			{param _handleItemClick: $_handleItemClick /}
 			{param _handleSearch: $_handleSearch /}
+			{param _hasExpanded: $_hasExpanded /}
 			{param button: $button /}
 			{param caption: $caption /}
 			{param contentRenderer: $contentRenderer ?: '' /}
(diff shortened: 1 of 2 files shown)


matuzalemsteles:

packages/clay-dropdown/src/ClayDropdownBase.soy
```diff
@@ -257,7 +257,11 @@
 
 	{call ClayPortal.render}
 		{param content kind="html"}
-			{$dropdownMenu}
+			{if $expanded}
```
matuzalemsteles: hey @travisrcory, this sounds like a good idea. But I was curious when you refer to performance, what performance do you mean? The first render?

travisrcory: Hey @matuzalemsteles sorry for the confusion, yeah I was referring to the first render! Thanks for clarifying.

matuzalemsteles: Np, thanks!

matuzalemsteles: Just leaving one thought of me about this change:

- **Always rendered** - As soy takes care of doing the diff, the first rendering may be slow but the next interactions will be fast since the only change is just add to `show` for example.
- **Destroy every time** - When rendered only when it is expanded, the first rendering will be fast, but the subsequent interactions will be slow since we are adding and destroying every time it has an `expanded` change.

We may be trading for each other, I like the idea of ​​getting the first fast render, but I'd also like to get the subsequent quick interactions.

One thought of getting around this would be something like `lazy`, not rendering on the first render, rendering on the first interaction with the dropdown, and subsequent interactions just kept changing what is needed... We could say that this would be a "dynamic loading" triggered by interaction.

---

I noticed something in the tests with this change, when I'm with the scroll zero and open the dropdown it skips scroll, I can not say why...

matuzalemsteles:

matuzalemsteles:
hey @travisrcory First sorry for the delay with this, I tried to make some improvements here for this to work well, but I had some problems:

- On the first Dropdown load, when you click to open Dropdown depending on where the dropdown content is, it will tinker with the scroll leading to where the dropdown is and then the alignment is done. This creates a scrolling jump.

For now I will close this PR, I think we have to do a lot to not render the content in the first render.

travisrcory:
@matuzalemsteles no worries! Thanks for taking a look at this. Sounds good, I got busy with some other tasks as well.

PR #1524, opened 2019-02-11
Fixes #1523 - Improves Bulk Action Support in ManagementToolbar

### LINKED ISSUE #1523: Improvements on Bulk Support for Management Toolbar
Following the latest changes to add support for bulk action management in the management toolbar, we're missing the following:
- Message for all selected items should include a placeholder for the number of items: `all-x-items-selected`
- It should be possible to configure visibility of both the `clear` and the `selectAll` buttons
- When the creationMenu is a single item, the item data should be passed down to the component as well as the items

jbalsas:


matuzalemsteles:

carloslancha:
Hi, if we're trying to remove logic from here and make the component "dumber" I don't really get what "supportBulkActions" is.

Did "all items selected" message change? I asked three times for that and the answer was always "all selected" without the number of items.

@jbalsas

jbalsas:
Hey @carloslancha, it's currently impossible to hide the `Clear` button. The old behaviour was that neither the `Clear` nor the `SelectAll` buttons showed. Now, it's possible to hide one, but not the other. For those places not implementing the new feature, it should be possible to keep both hidden.

The `all-selected` message doesn't exist, it's a new one and we'd need to translate it. Showing `All Selected` in the toolbar is certainly a weird message, so since we need to create all this, `All <strong>15</strong> Items Selected` sounds like a doable/simple enough approach, don't you think?

carloslancha:
@jbalsas and why not adding a `showClearButton` param just like with the `showSelectAllButton` one? Maybe I'm missing something.

I'm ok with the `all-x-items-selected`, just wanted to confirm 'cause what is defined in the docs and what UX told me multiple times was that, `all-items-selected`.

Thx

jbalsas:
Hey @carloslancha, first of all, you're on vacation, so remove all notifications and enjoy Florida!!

When `supportBulkActions === false`, none of those 2 buttons should show, no matter the state. Having 2 separate flags means it's easier to render incorrect states from the invoking code. This maintains the component dumb and makes it easier on the implementing side to handle.

carloslancha:
Then I don't see the point of having `showSelectAllButton` param :sweat_smile: 

Ok... I'll turn off notificaciones @jbalsas :P

jbalsas:
Because when `bulkActions` are available, there's a state where the `Clear` button is visible, but the `SelectAll` isn't 😉

packages/clay-management-toolbar/src/ClayManagementToolbar.soy
```diff
@@ -416,17 +423,20 @@
 		maxSecondaryItems: int,
 		maxTotalItems: int,
 		primaryItems: list<[
+			data: any,
```
matuzalemsteles: hey @jbalsas, is giving error when compiling, can you add this parameter in the other templates above that also declare them?

packages/clay-management-toolbar/src/ClayManagementToolbar.soy
```diff
@@ -416,17 +423,20 @@
...
 			label: string,
 			separator: bool
 		]>,
 		secondaryItems: list<[
+			data: any,
 			disabled: bool,
 			href: string,
 			icon: string,
 			items: list<[
+				data: any
```
matuzalemsteles: SF

packages/clay-management-toolbar/src/ClayManagementToolbar.js
```diff
@@ -606,6 +607,15 @@ ClayManagementToolbar.STATE = {
...
 	spritemap: Config.string().required(),
 
+	/**
+	 * Flag to indicate that the toolbar supports bulk selection.
+	 * @default false
+	 * @instance
+	 * @memberof ClayManagementToolbar
+	 * @type {boolean}
+	 */
+	supportsBulkActions: Config.bool().value(false),
```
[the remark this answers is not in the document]
jbalsas: Hey @matuzalemsteles, the `showSelectAllButton` logic is a bit more complicated than that... it shouldn't show by default, even if the `bulkAction` feature is enabled, so `false` still sounds like the proper default. You can see the whole definition here https://docs.google.com/document/d/1Aov9rThS1bcVscmS36DyCbpHdrM0OcEJZhA4w4_kHR4/edit?usp=drive_web&ouid=118239731538492698708

Parts of it are implemented here, but as @carloslancha said, we've tried to keep this component dumb so it can be easily controlled from above in Portal.

matuzalemsteles:
Resent #1527

PR #1535, opened 2019-02-11
 Fixes #1526 - Controls the input focus

### LINKED ISSUE #1526: ClayMultiSelect: Focus
**Steps to Reproduce**
1. Click an empty Multiselect

*Current behavior*
Multiselect has no focus style

not focused

*Expected behavior*
Multiselect should have a focus indicator

matuzalemsteles:


packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -92,6 +92,24 @@ class ClayMultiSelect extends ClayComponent {
...
+		this._removeFocusedItem();
+		this._inputFocus = false;
+	}
+
+	/**
+	 * Handles the input focus
+	 * @protected
+	 */
+	_handleInputOnFocus() {
+		this._removeFocusedItem();
```
jbalsas: Why do we need to do this here?

matuzalemsteles: Yes, this is not necessary here, we are already removing it in `_handleItemFocus`. Thanks!

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -81,6 +83,19 @@ class ClayMultiSelect extends ClayComponent {
...
+	 * Handle the click on the document to control the focus of the element
+	 * @param {!Event} event
+	 * @protected
+	 */
+	_handleDocClick(event) {
+		if (this.refs.autocomplete.element.contains(event.target)) {
+			this._inputFocus = true;
+		} else {
+			this._inputFocus = false;
+		}
```
jbalsas:
```suggestion
this._inputFocus = this.refs.autocomplete.element.contains(event.target);
```

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -331,14 +365,32 @@ class ClayMultiSelect extends ClayComponent {
...
 		this._itemFocused = null;
+		this._eventHandler = new EventHandler();
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	attached() {
+		this._eventHandler.add(
+			dom.on(document, 'click', this._handleDocClick.bind(this), true)
```
jbalsas: Why do we need to listen for `click` on `document`?? Isn't this an input, can't we handle this with simple css? @pat270 ?

matuzalemsteles: Hey @jbalsas when the user is focused on the Label and it leaves with the click to another, the focus will remain and another case, it would be when I click directly on the label, the focus should happen.

jbalsas: 👍

packages/clay-multi-select/src/ClayMultiSelect.js
matuzalemsteles: Great suggestion!!

PR #1547, opened 2019-02-12
clay-badge in React+Typescript

bryceosterhaus:
We talked about the possibility of using TS for the clay components over last week. I wanted to get a POC of a small component in react+ts just to see what it is like. 

Let me know what you guys think!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,7 @@
 		"notice/notice": [
 			"error",
 			{
-				"templateFile": "copyright.js"
+				"template": "/**\n * © 2018 Liferay, Inc. <https://liferay.com>\n *\n * SPDX-License-Identifier: BSD-3-Clause\n */\n"
 			}
 		]
 	}
--- a/copyright.js
+++ b/copyright.js
@@ -1,5 +1,0 @@
-/**
- * © 2018 Liferay, Inc. <https://liferay.com>
- *
- * SPDX-License-Identifier: BSD-3-Clause
- */
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
 		"link": "lerna run link",
 		"lint": "eslint packages/**/*.js examples/**/*.js scripts/**/*.js",
 		"pa11y": "pa11y-ci ./packages/clay-*/demos/a11y.html",
-		"prettier": "prettier-eslint packages/**/**/*.js examples/**/**/*.js scripts/**/**/*.js",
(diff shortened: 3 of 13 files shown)


bryceosterhaus:
also note, I change the copyright eslint rule because it was erroring out in nested packages for not being able to properly locate the template file.

matuzalemsteles:
hey @bryceosterhaus I do not see problem in going with TypeScript in Clay (Actually I like it 🙂).

wincent:
This is a great exploration. One thought I had about this (and also on the generator PR #1548) is that we should probably try to hoist as much common configuration and tooling up into the root of the monorepo (such as the tsconfig.json file). This means that we have one place to update whenever we want to make changes, and no possibility of different packages getting out of sync. Especially in the case of typechecking, we really want to do it at the whole-repo level anyway, to ensure that all the packages interoperate correctly; maybe we could do it even without hoisting the config up, but I know for sure that we can do it if we do.

Another thought is that we should probably still go with Babel here (and upgrade to v7 while we're at it), and use TypeScript for type-checking alone. Ditto for using Yarn workspaces.

I've explored a fair bit of this in my monorepo side-project to see if I could get all the kinks ironed out, and I am pretty happy with the result. Basically have global/shared Prettier/Flow/TypeScript/ESLint/Babel config and top-level build scripts all working with Yarn workspaces, and the build includes ES modules, old-school JS, and Flow and TypeScript typedefs by default. Poke around [the repo](https://github.com/wincent/js) if you want to take a look.

bryceosterhaus:
@wincent hoisting config up to the root is a good idea but im concerned about the usability and the developer experience when working on a single package. If we hoist all of the config and dev dependencies, does that mean we would have to always run tests and builds from the root level? I noticed in your repo that you used the Makefile to build everything from root. I would prefer that devs could just work on individual packages and use the same workflow as if it was an independent module.

I'll continue to play around with this a bit and investigate it though. Ideally I would probably get the generator finished first and then use that to create new packages.

wincent:
@bryceosterhaus: We can still add commands to the packages' "package.json", but I think they should delegate to the top-level. The whole point of having things in the monorepo is that all the bits cohere together as a whole, and making sure config and shared tooling lives in one place is the way to do that.

The `Makefile` in my repo is just an implementation detail; we don't have to use a solution like that. (In my case, I never want to build a single package without verify that every package builds and everything works as a whole.)

bryceosterhaus:
Going to re-build this component from the component-generator after the generator is merged here https://github.com/liferay/clay/pull/1548.

PR #1548, opened 2019-02-13
Create a yeoman generator for React+Typescript components

bryceosterhaus:
Our previous generator was named `generator-metal-clay` so I renamed it and removed the metal reference and then added the react and ts support.


packages/generator-clay-component/app/templates/_package.json
```diff
@@ -0,0 +1,60 @@
...
+	},
+	"main": "lib/<%= componentName %>.js",
+	"esnext:main": "src/<%= componentName %>.js",
+	"jsnext:main": "src/<%= componentName %>.js",
+	"files": ["lib", "src"],
+	"scripts": {
+		"compile": "tsc",
+		"prepublish": "npm run compile",
+		"start": "webpack-dev-server --mode development",
+		"test": "jest"
```
wincent: We should probably build a ".d.ts" file as well (something like I am doing [here](https://github.com/wincent/js/blob/c65bc118839edfae0490313ad0090fe642a9d170/Makefile#L9) in my test side-project).

packages/generator-clay-component/app/templates/_README.md
```diff
@@ -0,0 +1,26 @@
...
+
+  ```
+  npm run build
+  ```
+
+4. Open the demo at demos/index.html on your browser.
+
+## Contribute
+
+We'd love to get contributions from you! Please, check our [Contributing Guidelines](CONTRIBUTING.md) to see how you can help us improve.
```
matuzalemsteles: We can update the CONTRIBUTING.md link.
```suggestion
We'd love to get contributions from you! Please, check our [Contributing Guidelines](https://github.com/liferay/clay/blob/master/CONTRIBUTING.md) to see how you can help us improve.
```

bryceosterhaus:
@wincent @matuzalemsteles 

I updated the structure a bit and moved some of the dependencies around. I couldn't find any great examples of this sort of monorepo that both the benefits of a monorepo and also the individual autonomy of the package. So I moved the testing infrastructure up to the root of the project and then also provided shared configs at the root.

@wincent I didn't fully copy your example you provided because I thought it required too much interdependency between packages for building. Reason I avoided that is because packages could technically be independent of other packages in this repo. Such as `clay-badge`, the only dependency would be on clay-css which isn't a build dependency.

Feel free to test it out by using the generator from the root of the repo.

I'm open to more suggestions in changing this around though. I'm not totally satisfied with this solution yet.

bryceosterhaus:
Decided to make additional changes... I wasn't quite happy with the last change.

Changes:
- Hoisted jest/ts/babel configurations to root level
- Each package has its own dependencies
- Each package has its own build/test scripts. These can be run by...
	- `cd ./packages/some-clay-component && yarn test`
	- `yarn workspace @clayui/come-clay-component run test`
	- `yarn test` from root. This runs `lerna run test` which recursively runs that script in each module.

This way each module is autonomous in development, but also can be developed in parallel to other components with lerna.

IMO this is the way that seems to make most sense to me.

Let me know if you guys have any concerns or thoughts

packages/generator-clay-component/__tests__/test-app.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+'use strict';
+
+let path = require('path');
+let assert = require('yeoman-generator').assert;
+let helpers = require('yeoman-generator').test;
```
wincent: These `let` look like they should all be `const`.

packages/generator-clay-component/__tests__/test-app.js
```diff
@@ -0,0 +1,45 @@
...
+	});
+
+	it('content of MyComponent.tsx', function() {
+		assert.fileContent(
+			'src/MyComponent.tsx',
+			/const MyComponent: React.FunctionComponent<Props> = \(\{/
+		);
+		assert.fileContent('src/MyComponent.tsx', /export default MyComponent/);
+	});
+});
```
wincent: Just as style feedback, I think in tests we should favor arrow functions over `function () {}` for brevity, and value of `this` is irrelevant here.

packages/generator-clay-component/__tests__/test-app.js
```diff
@@ -0,0 +1,45 @@
...
+			'src/__tests__/MyComponent.tsx',
+			'webpack.config.js',
+			'package.json',
+			'tsconfig.json',
+			'tsconfig.declarations.json',
+			'README.md',
+		]);
+	});
+
+	it('content of MyComponent.tsx', function() {
```
wincent: This should read grammatically as an assertion; instead of "it content of MyComponent.tsx" it should be something like: "it produces MyComponent.tsx with templated content" (for example).

```suggestion
	it('produces MyComponent.tsx with templated content', function() {
```

packages/generator-clay-component/app/index.js
```diff
@@ -0,0 +1,163 @@
...
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+'use strict';
+
+let _ = require('lodash');
+let chalk = require('chalk');
+let yeoman = require('yeoman-generator');
+let yosay = require('yosay');
```
wincent: More `let` to `const` here, I think.

packages/generator-clay-component/app/index.js
```diff
@@ -0,0 +1,163 @@
...
+
+module.exports = yeoman.generators.Base.extend({
+	initializing: function() {
```
wincent: Can we assume this will be running in the context of a recent enough Node version that we can use ES6 object concise methods? ie.

```js
module.exports = yeoman.generators.Base.extend({
        initializing() {
                // etc...
```

packages/generator-clay-component/README.md
```diff
@@ -0,0 +1,33 @@
...
+yo clay-component
+```
+
+The generator will ask you a few questions about the package. Once you've answered all of them it will automatically generate a starting structure for you, as well as download all the local dependencies.
+
+_Note: This generator is intended to be used from the root level of the `clay` repository._
+
+## License
+
+[BSD License](http://opensource.org/licenses/BSD-3-Clause) © Liferay, Inc.
```
wincent: May as well use the SPDX ID here too (BSD-3-Clause).

```suggestion
[BSD-3-Clause License](http://opensource.org/licenses/BSD-3-Clause) © Liferay, Inc.
```

packages/generator-clay-component/app/index.js
```diff
@@ -0,0 +1,163 @@
...
+			yosay(
+				'Welcome, let\'s generate a ' +
+					chalk.blue('Clay') +
+					' component!'
+			)
+		);
+	},
+
+	prompting: function() {
+		let done = this.async();
```
wincent:
```suggestion
		const done = this.async();
```

packages/generator-clay-component/app/index.js
```diff
@@ -0,0 +1,163 @@
...
+
+		let prompts = [
```
wincent:
```suggestion
		const prompts = [
```

packages/generator-clay-component/app/templates/_README.md
```diff
@@ -0,0 +1,25 @@
...
+<%= repoDescription %>
+
+## Setup
+
+1. Install `yarn`
+
+2. Install local dependencies:
+
+```
+yarn install
```
wincent: This could be just `yarn`, because "install" is the default.

packages/generator-clay-component/app/templates/_package.json
```diff
@@ -0,0 +1,58 @@
...
+		"react-test-renderer": "^16.6.1",
+		"react": "^16.8.1",
+		"source-map-loader": "^0.2.4",
+		"style-loader": "^0.23.1",
+		"ts-jest": "^23.10.4",
+		"typescript": "^3.0.3",
+		"webpack-cli": "^3.1.2",
+		"webpack-dev-server": "^3.1.10",
+		"webpack": "^4.29.0"
+	},
```
wincent: Pretty much all dev dependencies can and I would argue should be hoisted up to the repo root. The node modules resolution algorithm means that `require('foo')` will still be found when working inside a package directory. Rationale is that we ensure that we are using a standard set of dev dependencies across all packages without any divergent versions. I know that you want developers to be able to work within package directories, but they still need to clone the repo in order to do that, so declaring the dev dependencies at the top level doesn't make working locally any harder. Take a look at a typical monorepo project like Babel and you'll see that all the dev dependencies are [at the top](https://github.com/babel/babel/blob/master/package.json), with only a couple of exceptions.

If you're interested, I have a checker in that sample project that [checks to make sure that dev dependencies only get declared at the top](https://github.com/wincent/js/blob/851033a694bd4c75dec463c9974f83e5c9d69b91/scripts/check-dependencies.js#L205-L236). Other checks are for things like [missing deps, mismatched deps etc](https://github.com/wincent/js/blob/851033a694bd4c75dec463c9974f83e5c9d69b91/scripts/check-dependencies.js#L285-L288).

packages/generator-clay-component/app/templates/src/_Boilerplate.tsx
```diff
@@ -0,0 +1,22 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import * as React from 'react';
+import getCN from 'classnames';
```
wincent: Maybe we should bind this to `classNames`, which is what [the "classnames" README examples use](https://www.npmjs.com/package/classnames).

packages/generator-clay-component/app/templates/src/tests/_Boilerplate.tsx
```diff
@@ -0,0 +1,18 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import * as React from 'react';
+import * as TestRenderer from 'react-test-renderer';
+import <%= componentName %> from '../<%= componentName %>';
+
+describe('<%= componentName %>', function() {
```
wincent:
```suggestion
describe('<%= componentName %>', () => {
```
I see you're already using arrow functions in the `it` examples.

packages/generator-clay-component/app/templates/src/tests/_Boilerplate.tsx
```diff
@@ -0,0 +1,18 @@
...
+	it('should render', () => {
```
wincent: Style: generally "should" should be avoided in example descriptions. You can always express it more concisely by leading directly with a verb. This is useful because it makes it more likely that you can fit the description in a single line. So in this case it would be:

```suggestion
	it('renders', () => {
```

tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+		"strict": true,
+		"jsx": "react",
+		"target": "esnext",
+		"module": "esnext",
+		"moduleResolution": "node",
+		"esModuleInterop": true,
+		"removeComments": true,
+		"baseUrl": "./",
+		"paths": {
+			"@clay/*": ["packages/*/src"],
```
wincent:
```suggestion
			"@clayui/*": ["packages/*/src"],
```

To match the change you made elsewhere? AFAIK we've requested `@clay` but no idea whether that can/will happen.

package.json
```diff
@@ -6,58 +6,31 @@
 		"checkFormat": "npm run prettier -- --list-different",
 		"coveralls": "node node_modules/coveralls/bin/coveralls.js < coverage/lcov.info",
 		"format": "npm run prettier -- --write",
-		"jest": "cross-env NODE_ENV=test jest",
 		"lerna": "lerna bootstrap -- --no-optional --no-package-lock",
 		"link": "lerna run link",
 		"lint": "eslint packages/**/*.js examples/**/*.js scripts/**/*.js",
 		"pa11y": "pa11y-ci ./packages/clay-*/demos/a11y.html",
 		"prettier": "prettier-eslint packages/**/**/*.js examples/**/**/*.js scripts/**/**/*.js",
-		"start": "http-server . -p 4000",
```
matuzalemsteles: hey @bryceosterhaus, I know we had talked about the demos and that this would not be the best option, I've been thinking about testing https://storybook.js.org/, I've seen some use cases in a monorepo, I think it is a great experience. We can test it once we start having some components in the master.

package.json
```diff
@@ -6,58 +6,31 @@
...
-		"test": "npm run build && npm run jest && npm run pa11y",
+		"test": "lerna run test && npm run pa11y",
```
matuzalemsteles: Just putting my thoughts about it:

- One of the points I like about having the jest running at the root is that we will only have one jest startup running and we can reap the coverage much better than terms in each package its coverage... This facilitates the information on https://coveralls.io/github/liferay/clay.
- We will have many components here still, so running the test for each project will take a long time for you to know if you have not broken another component that depends on it.
- When running the tests in this way, we will have the high boot time in each package, since we raise the jest all the time to test the package, maybe increase the CI time.
- I like to run the tests at the root because I can select which tests I want to run when I'm working on just one project. `yarn jest packages/clay-badge` for example.

packages/generator-clay-component/app/templates/_package.json
bryceosterhaus: I had tried that originally but then I ran into issues with resolutions when I was in a package itself. Something couldve been out of wack so I will try that again though.

package.json
bryceosterhaus: Yeah I agree we definitely need easily accessible demos on the repo level. Storybook is pretty great, I have used it in the past and it was easy to use. One other thing I am investigating is the possibility of using "Framer X" and that could also possible be a way to run demos. I am working with some of the UX leads here in LA and gonna determine if framer fits our needs.

I think this can be a tracked in a differ issue though, so I have created one https://github.com/liferay/clay/issues/1562

package.json
bryceosterhaus: Ah yeah that makes sense. I think if we hoist devDependencies like Greg had suggested then we can also probably run a single instance of Jest at both the root and package levels. Ill look into this.

package.json
matuzalemsteles: Oh great, I tested a component of Metal.js at the time with `Framer X` worked up, even though it was not React, I built a layer to give compatibility... The big problem was the CSS that are scoped.

bryceosterhaus:
@matuzalemsteles @wincent just updated.

Here are the highlights
- Moved devDependencies to root. (we can still call the individual package scripts by using `yarn build` in the package itself)
- Refactored some things from review comments
- Cleaned up package.json dependencies at root
- Created clay-badge component via the generator to showcase use

packages/clay-badge/package.json
```diff
@@ -0,0 +1,31 @@
+{
+	"name": "@clayui/clay-badge",
```
matuzalemsteles: hey @bryceosterhaus as we now have the scope, I think it is no longer necessary to add `clay-`, just leaving `@clayui/badge`.

packages/generator-clay-component/app/templates/_package.json
wincent: I'd say it's worth it. Babel at least has [some docs](https://babeljs.io/docs/en/config-files#monorepos) on set-up for monorepos, although that might not help with the resolution issues you were having.

packages/generator-clay-component/app/templates/_package.json
```diff
@@ -0,0 +1,31 @@
...
+		"npm": ">=3.0.0"
+	},
+	"main": "lib/<%= componentName %>.js",
+	"esnext:main": "src/<%= componentName %>.js",
+	"jsnext:main": "src/<%= componentName %>.js",
+	"files": ["lib", "src"],
+	"scripts": {
+		"build": "babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "tsc --project ./tsconfig.declarations.json",
+		"prepublish": "npm run build && npm run build:types",
```
wincent: This should probably be "prepublishOnly", right?

https://docs.npmjs.com/misc/scripts

bryceosterhaus:
@matuzalemsteles @wincent made updates in regards to your comments, re-ran the generator for clay-badge and then also added ts parser for eslint.

wincent:
CC @marcoscv-work and @carloslancha in case you haven't seen this yet.

.eslintrc
```diff
@@ -5,25 +5,83 @@
...
+		],
+		"sort-keys": [
+			2,
+			"asc",
+			{
+				"caseSensitive": false
 			}
-		]
+		],
+		"sort-vars": 2
```
wincent: A lot of this custom stuff would ideally be merged up into the "eslint-config-liferay" package, but that is currently a little hard for us to update because it is not in the Liferay org repo even though that's where [the package metadata](https://www.npmjs.com/package/eslint-config-liferay) says it should be; I have an internal ticket open for sorting that out: https://issues.liferay.com/browse/IFI-482.

package.json
```diff
@@ -6,58 +6,56 @@
...
-		"jest": "cross-env NODE_ENV=test jest",
 		"lerna": "lerna bootstrap -- --no-optional --no-package-lock",
 		"link": "lerna run link",
-		"lint": "eslint packages/**/*.js examples/**/*.js scripts/**/*.js",
-		"pa11y": "pa11y-ci ./packages/clay-*/demos/a11y.html",
-		"prettier": "prettier-eslint packages/**/**/*.js examples/**/**/*.js scripts/**/**/*.js",
-		"start": "http-server . -p 4000",
-		"test": "npm run build && npm run jest && npm run pa11y",
+		"lint": "eslint \"**/*.{js,ts,tsx}\"",
+		"prettier": "prettier-eslint \"**/*.{js,ts,tsx}\"",
```
wincent: You can safely use single quotes here if you want to avoid the backslash escaping.

packages/clay-badge/demo/App.tsx
```diff
@@ -0,0 +1,36 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
```
wincent: I don't think we need to do the `* as` thing here (or in the other files either). If you search, you'll find a few discussions about what React exports (and what it "should export"); this is the current reality:

- https://github.com/facebook/react/blob/1d6b1660a29c452e954d053a14c5adc42ccf94ba/packages/react-dom/index.js#L14-L16
- https://github.com/facebook/react/blob/1d6b1660a29c452e954d053a14c5adc42ccf94ba/packages/react-dom/index.js#L14-L16

And the official docs ([ReactDOM](https://reactjs.org/docs/react-dom.html), [React](https://reactjs.org/docs/react-api.html)) just tell you to `import React`, `import ReactDOM`.

packages/clay-badge/src/__tests__/__snapshots__/ClayBadge.tsx.snap
```diff
@@ -0,0 +1 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
wincent: I would have expected there to be some content in here.

packages/clay-badge/webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const path = require('path');
+const HWP = require('html-webpack-plugin');
+
+module.exports = {
+	entry: path.join(__dirname, './demo/App.tsx'),
```
wincent: nit: if you're using `path.join`, you don't need the "./" in "./demo/App.tsx".

.eslintrc
bryceosterhaus: Great idea. I had just copied over some of the rules we used in Analytics cloud. We slowly added them as we worked on the project and found them beneficial.

packages/clay-badge/demo/App.tsx
bryceosterhaus: The reason I use `* as React` was due to the ts compiler complaining.

screen shot 2019-02-25 at 12 17 19 pm

I looked into it and we would have to add this option to our tsconfig, `"allowSyntheticDefaultImports": true`. Not sure if that is the route we want to go though. 

I was reading these issues about it, https://github.com/Microsoft/TypeScript/issues/3337 and https://github.com/Microsoft/TypeScript-React-Starter/issues/8

Any preference on it?

packages/clay-badge/src/__tests__/__snapshots__/ClayBadge.tsx.snap
bryceosterhaus: Updated this in the generator

packages/clay-badge/webpack.config.js
bryceosterhaus: Updated this in the generator

packages/clay-badge/demo/App.tsx
wincent: Ah, that's a bummer. I wanted to stick to the pattern in the React docs but `allowSyntheticDefaultImports` seems like a bit of a hack. So `* as React` is probably the least horrible option.

bryceosterhaus: Yeah thats what I figured as well. Ill just leave it as is for now, and then if for some reason TS fixes this, we can make the change everywhere.

.eslintrc
```diff
@@ -5,25 +5,83 @@
...
+				"args": "after-used",
+				"ignoreRestSiblings": false,
+				"vars": "all"
+			}
+		],
 		"notice/notice": [
 			"error",
 			{
-				"templateFile": "copyright.js"
+				"template": "/**\n * © 2018 Liferay, Inc. <https://liferay.com>\n *\n * SPDX-License-Identifier: BSD-3-Clause\n */\n"
```
matuzalemsteles: hey @bryceosterhaus, just remembering that the year should start in 2019. 🙂
```suggestion
				"template": "/**\n * © 2019 Liferay, Inc. <https://liferay.com>\n *\n * SPDX-License-Identifier: BSD-3-Clause\n */\n"
```

bryceosterhaus: So I was reading through the original requirements for licensing in this comment, https://github.com/liferay/clay/pull/1451#issuecomment-456384580

This portion stuck out to me

> just change the year to whatever year that file was in (if that’s too much work, `2014` as the year of the first commit should do). And then do _not_ bother with updating the year in the future. When you create new files, just put the year when you created it – e.g. for new files started this year, that would be 2019.

And I just realized with using the eslint rule, it will update the year everywhere. It sounds like the requirements for licensing is to add the year the file is created, but then to not update the year in subsequent years.

I'll update everything to 2019 since its the first time we are adding this header, but I'm gonna look at the eslint rule and see if we can configure it to not update in subsequent years.

bryceosterhaus: @matuzalemsteles just updated pr. The eslint plugin actually allows a `<%= YEAR %>` param in the template. This should solve the issue of new files and existing files with licenses. It only updates to the current year if its a new file.

matuzalemsteles: Oh yeah, great @bryceosterhaus! Thanks

bryceosterhaus:
Once the generator portion of this PR is good to go, I am going to re-generate the badge component based on the generator. I'll also squash all of the generator commits into a single commit.

matuzalemsteles:

packages/generator-clay-component/app/templates/_webpack.config.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const path = require('path');
+const HWP = require('html-webpack-plugin');
+
+module.exports = {
```
matuzalemsteles: hey @bryceosterhaus, as we will always repeat this for other components, maybe we can do the same thing we did in Clay v2 https://github.com/liferay/clay/tree/2.x-develop/packages/webpack-config-clay.

```javascript
const webpackCommonConfig = require('webpack-config-clay');

module.exports = Object.assign(webpackCommonConfig, {
	entry: './src/ClayBadge.js',
	output: Object.assign(webpackCommonConfig.output, {
		filename: './build/globals/clay-badge.js',
	}),
});
```
That's on Clay v2. Here it might look something like this:

```javascript
const webpackCommonConfig = require('webpack-config-clay');

module.exports = webpackCommonConfig;
```

What do you think?

bryceosterhaus: I'm not entirely sold on the idea of sharing configs through additional modules. I'm also not totally against it. I do think it might be beneficial to start out with more verbosity in the config and then if it becomes an issue later, we can abstract it.

Part of this reasoning comes from a thought of mine in that I think we to over-abstract too quickly. I personally think its easier to abstract later in the development process when you have more examples to work with.

matuzalemsteles: No problem, just a suggestion we can see how it will be in the future.

matuzalemsteles:
hey guys, I think we're ready to go with that forward, @wincent do you have anything else you want to comment on? I want to merge this so we can go ahead with #1440 and #1427 and apply the patterns we've added here.

wincent:
Nah, I don't have anything else. I think this is a good enough base upon which we can iterate further. I wanted to make sure we didn't rush this, in order to give people time to comment, but I figure that at this  point everybody who cares has probably already shared their opinions.

jbalsas:
I care and I haven't! (I'll just take a quick look right away, but I'm sure it'll be to my liking 🤓)

.eslintrc
jbalsas: We're in the process of migrating the config to its proper location so we can start adapting it... one thing I'm worried about is if these long list makes sense for all cases or just for the `jsx` one... should we have different config  bases, like `eslint-config-liferay` and `eslint-config-liferay-jsx`?

wincent: > One config to rule them all...

.travis.yml
```diff
@@ -24,4 +24,4 @@ language: java
 script:
```
jbalsas: If we're moving to `yarn`... shouldn't we use it to run the CI commands too?

packages/clay-badge/src/ClayBadge.tsx
```diff
@@ -0,0 +1,34 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface props {
```
jbalsas: I assume a good chunk of this would be common... I know this is likely a premature abastraction, but isn't there something like a declared base interface for this so in cases you don't want to declare anything else you can avoid the additional interface, and in case you want to add something, you just need to specify those?

bryceosterhaus: I'm not totally sure I am understanding your question here. You want to abstract the interface here?

packages/clay-badge/src/ClayBadge.tsx
```diff
@@ -0,0 +1,34 @@
...
+	label?: string | number;
+	className?: string;
```
jbalsas: Maybe this should be `classNames`?

packages/clay-badge/src/ClayBadge.tsx
jbalsas: Maybe this is classic early-abstraction, so it can wait. I was thinking...

```
// @types/baseprops.dts
export const interface baseprops {
    // not sure if there are other cases than the prop index
    [propName: string]: any;
}
```

```
import baseprops from '@types/baseprops';

interface props extends baseprops {
    className?: string;
    label?: string | number;
    type?: string;
}

const ClayBadge: React.FunctionComponent<props> = ({
...
});
```

jbalsas: I might've jumped the gun because of `className`, so it's likely I'm missing something, but just in case :)

packages/clay-badge/src/ClayBadge.tsx
bryceosterhaus: I think it would confuse people to have `classNames` on components and then the native `className`(no "s") on elements. I think we either keep it consistent, or we entirely divert and use something like `elementClasses` that we had in metal.

packages/clay-badge/src/ClayBadge.tsx
bryceosterhaus: `[propName: string]: any;` would essentially be a catch all for when people want to allow "...otherProps". Let me look into how the community handles this and maybe there is a more simple way. It may even be an anti-pattern, and instead we should extend the interface from the html element the `otherProps` are being applied to.

wincent:
Ah, a thing I forgot to mention. `.eslintrc` is deprecated; we should use `.eslintrc.js` instead.

packages/clay-badge/src/ClayBadge.tsx
bryceosterhaus: Rather than using the `any` type, we should extend from the HTMLElement's attributes interface.

See my update https://github.com/liferay/clay/pull/1548/commits/4e28a0e64e1b606b81bae796f7a462c2c2c2610d

bryceosterhaus:
Updated

bryceosterhaus:
@jbalsas , Let me know if there is anything else wanted in this PR.

matuzalemsteles:
hey guys, just an update about the `@clay` scope in npm: I received the response from the NPM team that it will not be possible for them to proceed even though the user did not reply to their emails because there are packages in the user account that are still being downloaded very frequently. I think we'll have to keep `@clayui`. 😞

cc @wincent, @bryceosterhaus, @jbalsas

bryceosterhaus:
I went ahead and created a new PR with a cleaned up commit history, seen here https://github.com/liferay/clay/pull/1648

jbalsas:
Closing in favour of #1648 to reduce noise here!

PR #1549, opened 2019-02-13
Fixes #1480 - Adds new allowedCharacters API in Autocomplete and MultiSelect

### LINKED ISSUE #1480: ClayMultiSelect: forbidden characters
Add the possiblity to define if there are characters that we do not want to allow in the input field. 
We have this in the TagSelector taglib in portal.

matuzalemsteles:


packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -461,6 +481,16 @@ ClayAutocomplete.STATE = {
...
+
+	/**
+	 * Validate the values of the input, use a Regex to validate the
+	 * input characters (e.g /[a-zA-Z0-9_]/g).
+	 * @default undefined
+	 * @instance
+	 * @memberof ClayAutocomplete
+	 * @type {?RegExp}
+	 */
+	validator: Config.instanceOf(RegExp),
```
jbalsas: I think validator should be at least either a regexp or a function. Otherwise, we impose a very heavy restriction on the type of validation that can occur. For better flexibility, or `_runValidation` should also handle async values, so it should return a promise, which would be what's expected of the validator when a function is passed.

packages/clay-autocomplete/src/ClayAutocomplete.js
```diff
@@ -148,11 +148,17 @@ class ClayAutocomplete extends ClayComponent {
 	 */
 	_handleOnInput(event) {
 		const {value} = event.target;
-		const char = event.data || value.substr(-1);
+		const newValue = this.validator ? this._runValidation(value) : value;
+		const char = newValue.substr(-1);
```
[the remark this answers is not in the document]
matuzalemsteles: Hey @jbalsas this is equivalent to `event.data` but `event.data` does not work in Firefox, it was causing some bugs as described here #1446.

packages/clay-autocomplete/src/ClayAutocomplete.js
matuzalemsteles: Accepting a function can give you more flexibility... I do not know if it would be the best way to deal with async code here, these checks have to be done sync, since the Metal does not control the input, the user can typing and we have to make the changes so as not to lose information that the user has entered. Maybe this name is confusing, the idea is not to deal with it after the user has finished typing.

matuzalemsteles: Actually if we go with the idea of ​​validating the final value of the input, we must follow this path. I will include this functionality in another PR, but I will change the name of this API so it will not be confusing and improve the description.

jbalsas: good compromise and way to move forward! 👍

PR #1561, opened 2019-02-15
Improves autocomplete feedback

matuzalemsteles:
hey @jbalsas, there are still some things to be done here, but it comes quite close to what have in the proposed documents, if you can get a look at it to check the behaviors and see how it's getting.

Feel free to leave your thoughts.

> As we are not sure yet whether we will continue with these improvements, DON'T MERGE!

### DIFF
--- a/packages/clay-autocomplete/demos/index.html
+++ b/packages/clay-autocomplete/demos/index.html
@@ -100,6 +100,7 @@
 									.then(res => res.results)
 									.catch((error) => console.log(error)),
 			extractData: (elem) => elem.name,
+			requestRetries: 0,
 		}, '#data-function-block');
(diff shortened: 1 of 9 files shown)


jbalsas:
Hey @matuzalemsteles, after the other merges this has some conflicts... could you fix those? I'll take a look at this next!

matuzalemsteles:
hey @jbalsas, it still lacks some work, it will probably conflict when #1590 is merged, but I think we can validate if this will go forward.

matuzalemsteles:
This PR needs to be reworked with recent changes over MultiSelect. I will send another PR by adding these improvements, see #1852 to track.

PR #1578, opened 2019-02-20
Fixes #1577 - Adds onClick only when the item is not disabled

### LINKED ISSUE #1577: Management toolbar action buttons trigger click event when disabled

matuzalemsteles:


jbalsas:
Hey @matuzalemsteles, this looks like a possible workaround, but I'm sure this issue will appear in other cases. Might not be worth spending too much time on it, but before going ahead and merge this... could you take a look at https://github.com/metal/metal.js/issues/134?

It's fishy that a disabled element is triggering the events, so it might just happen that we have some underlying bug causing it. Could you take a look and report back here so we can proceed one way or another?

matuzalemsteles:
hey @jbalsas was taking a look at this now, the implementation [done here](https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/domNamed.js#L294), does not cover all use cases that may occur:

- Our **onclick** is in a `Li` tag, the Metal does not stop adding listerns if the Button is disabled it still keeps propagating.
- Metal just skips the event if the event is on the Button (That is, he does not call the event that is attached to him), for example. Only these are supported `BUTTON, INPUT, SELECT, TEXTAREA, FIELDSET`.

I do not know if this was expected behavior just for those cases, I would have to go a little deeper.

jbalsas:
I think this might be good enough for now, but I'm not sure how it would react if the `disabled` property changes dynamically. Could you verify that the listener gets added/removed properly on `item.disabled` changes?

matuzalemsteles:
hey @jbalsas, I did some manual tests, it always has the delegate added, but it always checks when it has a click if there is any listener in the element to be called. This works fine. This happens here https://github.com/metal/metal.js/blob/master/packages/metal-dom/src/domNamed.js#L861

PR #1590, opened 2019-02-21
Fixes #1575 - Wrap autocomplete in MultiSelect with class .dropdown.dropdown-full

### LINKED ISSUE #1575: ClayMultiSelect: helpText is hidden when there are filtered items
This is visually weird. Everything under the multiselect component is moved because the helpText dissapears. See attached gif
claymultiselect_autocomplete_helptext

matuzalemsteles:


carloslancha:
Hey @matuzalemsteles I don't really like the idea of neither duplicating the logic of those classes in two components and that any other classes dependes on `elementClasses` to be passed or not.

I mean, for the `elementClasses`, that never should be an else statement for that param, i.e:

```
// Correct
form-group

{if $elementClasses}
   {$elementClasses}
{/if}

// Incorrect
{if $elementClasses}
   {$elementClasses}
{else}
   form-group
{/if}
```

Otherwise you're forcing the user to add some required classes to the `elementClasses` param when he wants to add a custom one. In the case above if you pass your custom `elementClasses` you may need to also pass `form-group`, which is logic of the component itself.

If you need to do something like this and provide support to wrap or not the component I suggest to use the pattern present in ClayTable (`wrapTable` stuff and so on)

Just let me know if you have any doubts! :D

Thx!

carloslancha:
Merged, thx.

PR #1593, opened 2019-02-21
Fixes #1592 - Adds collapse height when closing to ensure animation happens

### LINKED ISSUE #1592: Navigation Bar menu broken when click repeatedly will cause it to breaking
By repeatedly clicking on the toggler in ClayNavigationBar it breaking.

matuzalemsteles:
This is an assumption I can not debug to see if this happens, when you add `_setCollapseHeight` in the click and then remove it in `sync_isTransitioning` if they happen in a short period so that it does not cause an animation and breaks the menu.

### DIFF
--- a/packages/clay-navigation-bar/src/ClayNavigationBar.js
+++ b/packages/clay-navigation-bar/src/ClayNavigationBar.js
@@ -23,11 +23,11 @@ class ClayNavigationBar extends ClayComponent {
 	 */
 	// eslint-disable-next-line
 	sync_isTransitioning() {
-		if (this._isTransitioning && !this._visible) {
+		if (this._isTransitioning) {
 			this._setCollapseHeight();
-		} else if (this._isTransitioning && this._visible) {
-			this._setCollapseHeight();
-			this._removeCollapseHeight();
+			if (this._visible) {
+				this._removeCollapseHeight();
+			}
 		}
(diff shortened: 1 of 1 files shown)


packages/clay-navigation-bar/src/ClayNavigationBar.js
```diff
@@ -26,6 +26,7 @@ class ClayNavigationBar extends ClayComponent {
 		if (this._isTransitioning && !this._visible) {
 			this._setCollapseHeight();
 		} else if (this._isTransitioning && this._visible) {
+			this._setCollapseHeight();
 			this._removeCollapseHeight();
 		}
```
wincent: I can't comment on the correctness of this change, but the conditional might be more readable if you wrote it like this:

```javascript
if (this._isTransitioning) {
  this._setCollapseHeight();
  if (this._visible) {
    this._removeCollapseHeight();
  }
}
```

Obviously, readability is a bit subjective, but if you write it this way, the side-effect that is common to both branches is more obvious (ie. the `_setCollapseHeight` call) and also, the conditions under which it is called (ie. when `_isTransitioning`).

matuzalemsteles: Thanks for this, thanks @wincent!

PR #1594, opened 2019-02-21
Fixes #1585 - Adds the object validator to splineInterpolationType

### LINKED ISSUE #1585: Clay-Charts expects splineInterpolationType to be a string, but can sometimes be an object.
https://github.com/liferay/clay/blob/860f3be9e01cbab321bd7f2ff2abfdc931383b61/packages/clay-charts/src/ChartBase.js#L1265-L1274

In Clay-Charts v2.8.3, the `splineInterpolationType` prop in `ChartBase.js` will have an invalid props warning that an object was passed instead of a string. It looks like the props setter will return an object if `value` is valid, so it seems like there should add an additional type of `Config.object()` to allow for it.

matuzalemsteles:


packages/clay-charts/src/ChartBase.js
```diff
@@ -1267,9 +1267,12 @@ ChartBase.STATE = {
 	 * @default undefined
 	 * @instance
 	 * @memberof ChartBase
-	 * @type {?(string|undefined)}
+	 * @type {?(string|object|undefined)}
 	 */
-	splineInterpolationType: Config.string().setter(value => {
+	splineInterpolationType: Config.oneOfType([
+		Config.string(),
+		Config.object(),
```
wincent: nit: alphabetical order for this?

PR #1596, opened 2019-02-21
Fixes #1510 - ClayCSS 2.x Atlas Stickers invert background-color with color (Breaking)

### LINKED ISSUE #1510: Stickers must have inverted colors and not allow any text
Right now the stickers have a colored background and white icons, we need it to be inverted, the result should be a white background with a colored icon. 

Text (normally used for the user initials) won't be allowed in the stickers, this was a requirement of accessibility and international translations. 

[Lexicon Documentation](https://docs.google.com/document/d/1F5usuexmav7h3s9gKhUEgMhYYMMQiy_UUvHxSEx5bBE/edit#)

image

pat270:
ClayCSS Mixins added `clay-sticker-variant` for creating other sticker types

ClayCSS Stickers added new variant `.sticker-user-icon` and Sass map `$sticker-user-icon` which uses `clay-sticker-variant` mixin

@matuzalemsteles for `ClayUserCard` sticker we need to add `sticker-user-icon` to `elementClasses`
or
`ClaySticker` we will need to add a shape `user-icon` and change the shape in `ClayUserCard` to `user-icon`.

### DIFF
--- a/packages/clay-css/src/content/badges_and_labels.html
+++ b/packages/clay-css/src/content/badges_and_labels.html
@@ -662,17 +662,32 @@ section: Components
 			<p>Stickers are monospaced badges/labels.</p>
 		</blockquote>
 
-<span class="sticker sticker-primary">133</span>
-<span class="sticker sticker-secondary">133</span>
-<span class="sticker sticker-success">133</span>
-<span class="sticker sticker-info">133</span>
-<span class="sticker sticker-warning">133</span>
-<span class="sticker sticker-danger">133</span>
+<div>
+	<span class="sticker sticker-primary">133</span>
+	<span class="sticker sticker-secondary">133</span>
+	<span class="sticker sticker-success">133</span>
+	<span class="sticker sticker-info">133</span>
+	<span class="sticker sticker-warning">133</span>
+	<span class="sticker sticker-danger">133</span>
+	<div class="clay-site-light-container" style="vertical-align: baseline;">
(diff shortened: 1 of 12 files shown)


carloslancha:
Hey @pat270 @matuzalemsteles in order to make this more extensible I'd not use `sticker-user-icon` as a class if the only difference is that we're adding the border. I'd rather create a `sticker-border` or something like that to add it to the shape you want. What do you guys think?

matuzalemsteles:
hey @carloslancha I like to go with `sticker-border`, since it just adds the border, unless @pat270 has plans to add something more in the future.

pat270:
@carloslancha, @matuzalemsteles, @jbalsas 

I made a mistake in the early stages of v2 by removing the `user-icon` component. I think it's better to have `sticker-user-icon` because it future proofs the component for us. If the Lexicon Team decides to change the styles for a user icon we have a class that it's scoped to and can make changes to it independent of the `sticker` component.

It also makes it easier for devs to customize user icons via CSS in their projects. They don't need to worry about changing the sizes or colors of a `.sticker.sticker-circle.sticker-border` that isn't a user icon. This is something you can pass in via `elementClasses` and `ClaySticker` doesn't need to be updated. On a side note, I think we shouldn't be forcing a `shape` or a `style` by default. It makes it difficult to overwrite/customize.

Lastly, the Lexicon Team wants user icons to visually be 2px larger than a normal sticker. I achieved that by using a `box-shadow` instead of `border`.

PR #1608, opened 2019-02-26
Fixes #1604 - Remove the rule to check if the comma exists at the end to be labeled when it is paste

### LINKED ISSUE #1604: Multi-select must discard label if a non-valid character is found
Multi-select component has an open list of non-valid chars.
Text pieces containing non-valid chars should not shape a label.
Example:
Non-valid chars list: "-"
Text string to paste: a, b, c-d, f
Expected outcome: [a][b][f]
c-d must be discarded in the creation

I found this happening in blogs: so either the component is wrong or the component use in blog creation is not stating the non-valid chars list.

matuzalemsteles:


jbalsas:
Hey @matuzalemsteles, I'm not sure this is right. The current code is actually causing https://issues.liferay.com/browse/LPS-91089 because we end up with `tag,` as input value, when it should be `tag`. Could you check to see if this covers both cases before proceeding?

Thanks!

matuzalemsteles:
hey @jbalsas, we actually deleted the comma here https://github.com/liferay/clay/blob/2.x-develop/packages/clay-multi-select/src/ClayMultiSelect.js#L356, in theory the values ​​to be transformed into items should not go with comma.

I tried to reproduce the use case outside the Portal in the Clay demos, but I can not reproduce this problem but I see these problems inside the Portal, I do not understand 🤷‍♂️.

matuzalemsteles:
hey @jbalsas, I can reproduce this in the demos, this is being fixed in 70c1229 #1617.

diegonvs:

carloslancha:
Hey @matuzalemsteles is this still needed after merging https://github.com/liferay/clay/pull/1617 ?

matuzalemsteles:
hey @carloslancha yes, this adds the new rule when we add a set of text to the input so it makes the labels even if the last word does not have the comma and it also fixes the issues with the typing the spaces.

carloslancha:
Merged, thx!

PR #1614, opened 2019-02-27
Fixes #1610 - Emit rowContentClicked event when click the ClayTable row (ADDED API | BREAKING CHANGE)

### LINKED ISSUE #1610: ClayTable: managing the active state in a different way (on rows)
At the moment the state "active" on rows is managed passing "selectable" as true on the entire table component and defining item.selected as true.
That adds a checkbox that, if clicked, triggers the events `change`.

What would be nice to have are:
- a modifier on the row that responds to item.selected
- the possibility to detect a click on the entire row (now it is possible only handling the single cells via `_handleCellContentClick`) 

Thank you guys! :)

cc: @marco-leo

matuzalemsteles:


packages/clay-table/src/ClayTable.soy
```diff
@@ -235,13 +239,14 @@
...
 	{@param? _handleItemToggled: any}
+	{@param? _handleRowContentClick: any}
 	{@param? actionsMenuVariant: string}
 	{@param? groupName: string}
 	{@param? selectable: bool}
 	{@param? showActionsMenu: bool}
 	{@param? spritemap: string}
 
-	<tr>
+	<tr data-onclick="{$_handleRowContentClick}">
```
carloslancha: Nice @matuzalemsteles !

Could you please add to this `tr` the class `table-active` when `item.selected` is true?

Thx!

packages/clay-table/src/ClayTable.soy
```diff
@@ -246,7 +246,15 @@
 	{@param? showActionsMenu: bool}
 	{@param? spritemap: string}
 
-	<tr data-onclick="{$_handleRowContentClick}">
+	{let $attributes kind="attributes"}
+		{if $selectable and $item.selected}
```
carloslancha: not only on selectable, this should work always (even when there's no checkbox present, that's the use case for commerce). 

If you find it weird maybe we can implement something like `showCheckbox`  and use `selectable` as you pointed.

@FabioDiegoMastrorilli sounds good?

FabioDiegoMastrorilli: @carloslancha the second idea is the best: we do need to pass `selectable` as a parameter in order to emit an event but, at the moment, it also affect the html markup (showing a checkbox).
The fact is that we want the row to be selected but we don't need the checkbox

Thank you guys :)

matuzalemsteles: I'm going to add the `showCheckbox` API but keep the `selectable` for the context of whether or not the row is selectable. Thank you guys!

packages/clay-table/src/ClayTable.js
```diff
@@ -204,6 +240,15 @@ ClayTable.STATE = {
...
 	showActionsMenu: Config.bool().value(false),
 
+	/**
+	 * Flag to indicate if the checkbox is show or not when selectable is true.
+	 * @default true
+	 * @instance
+	 * @memberof ClayTable
+	 * @type {?bool}
+	 */
+	showCheckbox: Config.bool().value(true),
```
matuzalemsteles: hey @carloslancha I'm putting the `true` value as default to keep the same behavior as before and not cause some breaking change. In this way to hide `showCheckbox`, just set true `selectable` and `showCheckbox` to false.

carloslancha: Hey @matuzalemsteles if `selectable` default value is `false` I'd go with `showCheckbox` to false too. To avoid problems in portal we can set this to `true` by default in the taglib, so here we can have a more "logic" definition. What do you think?

matuzalemsteles: I do not know if this would cause some strange behavior in the Portal, I'm assuming that when selectable is false the checkbox will not be shown and when selectable is true it will be shown, this keeps the default as it was before, so will have no problems.

I think if we only keep in taglib as true, in the rehydration of the elements the checkbox may disappear.

makes sense?

carloslancha: `I think if we only keep in taglib as true, in the rehydration of the elements the checkbox may disappear`.

Nope, the Tag is in charge of create both server and client component instances with same values.

carloslancha:

carloslancha:
Merged, thx!

PR #1617, opened 2019-02-28
 Fixes #1581 - Compares the old value of selectedItems with the new value to set focus to the input

### LINKED ISSUE #1581: Clay Multi Select gets focus when other input elements change
This is how we reproduce it:

video

When any of the other inputs change the multiselect gets the focus. Please note that the whole thing is a Soy template.

This is the line that is run:

screenshot 2019-02-20 at 17 00 51

matuzalemsteles:
To understand more why add this, see the comments in #1581, please leave your thoughts on this, I would love to hear other ideas.

### DIFF
--- a/packages/clay-multi-select/src/ClayMultiSelect.js
+++ b/packages/clay-multi-select/src/ClayMultiSelect.js
@@ -188,14 +188,14 @@ class ClayMultiSelect extends ClayComponent {
 					this._performCall(this.valueLocator, itemSelected) === label
 			)
 		) {
-			const index = this.selectedItems.push(data);
-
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
hey @carloslancha was what I expected Metal to do, but it looks like **newVal** and **prevVal** are coming with equal values ​​whenever they are changed. I've tried to make the values ​​of `selectedItems` immutable by always creating a new array to lose the reference, but it does not solve the problem.

carloslancha:
Hey @matuzalemsteles could you please resolve the conflicts? Thx!

carloslancha:
Merged, thx!

PR #1632, opened 2019-03-07
Fixes #1631 - Add new spacers

### LINKED ISSUE #1631: New spacers for helpers classes
We need some extra spaces:

image

based on:
https://issues.liferay.com/browse/LPS-91500

marcoscv-work:


pat270:
Hey @marcoscv-work, 

According to that chart, it looks like Atlas `$spacer` should be 0.75rem? We need to change `3` to be `($spacer * 0.75)` or define `$spacer: 0.75rem !default` and use the values from the chart. Also we should probably add `6`, `7`, and `8` to Bootstrap as well. I can forsee this being used in places outside of Classic even if we try hard to contain them there.

marcoscv-work:
We have been taking a deeper look and we decided to keep using 1rem as a base and change the spacers multiplicators, new PR here:

https://github.com/liferay/clay/pull/1645

Thanks Patrick!

PR #1633, opened 2019-03-07
Fixes #1630 - Emit the onblur event of the input and clears the values ​​contained in the input as default

### LINKED ISSUE #1630: ClayMultiSelect enable onblur state
Collab team needs onblur state to be able to trigger this event and create their own logic.

As a default behavior we can discard every text that has not been transformed into a label. The behaviour must be possible to overwrite.

matuzalemsteles:


packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -107,11 +115,18 @@ class ClayMultiSelect extends ClayComponent {
 
 	/**
 	 * Handles the input blur
+	 * @param {!Event} event
 	 * @protected
+	 * @return {?Boolean} If the event has been prevented or not.
 	 */
-	_handleInputOnBlur() {
+	_handleInputOnBlur(event) {
 		this._removeFocusedItem();
```
carloslancha: Shouldn't this two lines be in the `_defaultInputOnBlur` method?

matuzalemsteles: We need to ensure that the focus of the item is always removed as much as the focus of the input, if we stay in `_defaultInputOnBlur` someone may override the behavior of clearing the input value and may break those behaviors that we want to always work. Regardless of whether someone overwrites or not. Makes sense?

matuzalemsteles:
It makes sense, I'll make those changes and send. Thanks @carloslancha

packages/clay-multi-select/src/ClayMultiSelect.js
carloslancha: Makes sense, but... my concern is that if I prevent the default event behavior I don't know if I should expect nothing but my own logic happen. What do you think?

matuzalemsteles: Maybe if the contained logic of the behavior had only in the `inputOnBlur`, but the removal of the focused item and focus the input I think its behavior is all distributed in the component, for example if I give prevent the default I would probably expect that I lose complete behavior and maybe not a single part of it. Makes sense for you?

carloslancha:
Merged, thx!

PR #1647, opened 2019-03-08
Fixes #1571 - Remove the <a> and <li /> tag from the deltemplate ClayAutocomplete.Item

### LINKED ISSUE #1571: Clay Multi Select filters out valid results
We have connected `clay-multi-select` to an autocomplete that returns results based on different fields (user name and user email). However we just want to display the user name in the labels. Since the email is not included in the label, [the filter function of the data provider](https://github.com/liferay/clay/blob/2.x-develop/packages/clay-data-provider/src/ClayDataProvider.js#L189) filters out the result which should be displayed.

For example if we search for `example` we expect the following result from the backend: `{user: 'someone', email: 'example@example.com'}` but we just want to display the name `someone`. Since `example` is not included in `someone` the suggestion is not displayed.

It would be nice if we could pass a boolean to disabled the filtering, since in our case is being done in the backend.

Thanks!

matuzalemsteles:


matuzalemsteles:
cc @4lejandrito

4lejandrito:
Really nice @matuzalemsteles. I will keep an eye on this because we will have to change https://github.com/liferay/liferay-portal/blob/6101d56327e3692828ec25730733c7f10fb4e12b/modules/apps/sharing/sharing-web/src/main/resources/META-INF/resources/Sharing.soy#L136-L151 when it gets merged.

carloslancha:
hey @matuzalemsteles @4lejandrito I have the same concerns Matu mentioned in the issue about someone needing to customize the `<li>` or the `<a>`. I'd suggest to follow the same patter we're using in `ClayDropdown` (see `ClayDropdownItem.soy`).

If you agree please close this pr. Thanks guys!

4lejandrito:
Hi @matuzalemsteles,

Both of your suggestions would be fine for us since they allow us to only override the content. The second one would allow for more advanced use cases but giving the user enough power to break the component's behaviour. 

In any case, this will make [our code](https://github.com/liferay/liferay-portal/blob/6101d56327e3692828ec25730733c7f10fb4e12b/modules/apps/sharing/sharing-web/src/main/resources/META-INF/resources/Sharing.soy#L136-L151) simpler.

Thanks!

carloslancha:
@matuzalemsteles I see that a bit overkill, but ok.

matuzalemsteles:
I will close this PR and send those changes to another one. Thanks guys!

PR #1648, opened 2019-03-08
Cleaned up commit history for pr#1548

bryceosterhaus:


jbalsas:
If everyone's happy with this, I'm happy with it! 👍

.eslintrc.json
```diff
@@ -0,0 +1,87 @@
...
+		"es6": true,
+		"jest": true
+	},
+	"parser": "@typescript-eslint/parser",
+	"parserOptions": {
+		"ecmaFeatures": {
+			"jsx": true
+		}
+	},
+	"extends": "liferay",
```
wincent: If you update to the latest version fo the config (v3.0.0) you can drop some of the rules because I rolled them in to the defaults (current "master" defaults [here](https://github.com/liferay/eslint-config-liferay/blob/master/index.js)).

For example, I think you could drop the `no-for-of-loops` and `notice` plug-ins (eslint-config-liferay includes it and will use the `copyright.js` file automatically if it exists).

I think you can also drop some rules:

- `max-len`, `eol-last`, `comma-dangle`, `no-mixed-spaces-and-tabs`, and (I think) `quotes`: all handled by the prettier preset, which is bundled with the liferay preset.
- `no-console`, `no-undef`: already in eslint:recommended, which the preset extends.
- `default-case`, `no-return-assign`, `object-shorthand`, `prefer-const`, `quote-props`: already in the preset, because I added them there.
- `no-unused-vars`: already in the preset, albeit with some different options; I'd suggest trying to see if dropping it is viable.

Basically, I think most of the plug-ins and rules can go away, except the React-related ones, and the sorting ones (which ended up being too invasive to bake into the preset).

If you're curious about which rules I added to the preset, I wrote it all up in https://github.com/liferay/eslint-config-liferay/issues/25 — one of the things I did was look at this PR and the existing config on the "develop" branch to see what could be merged "upstream" into the preset.

bryceosterhaus:
@wincent updated. I left in the `no-unused-vars` rule due to `ignoreRestSiblings` being used fairly often with react by doing `...otherProps`

wincent:
> I left in the no-unused-vars rule due to ignoreRestSiblings being used fairly often with react by doing ...otherProps

Good call. And I think we might want to make a React-variant of the config for precisely this kind of thing plus the other React stuff. I think in theory it is possible to create a `react.js` file in eslint-config-liferay and then you'd be able to extend "eslint-config-liferay/react" instead of "eslint-config-liferay"; I'll create an issue for setting that up: https://github.com/liferay/eslint-config-liferay/issues/36

PR #1673, opened 2019-03-14
Fixes #1571 - Adds a new extension point to only customize the content of the dropdown item

### LINKED ISSUE #1571: Clay Multi Select filters out valid results

matuzalemsteles:


carloslancha:
Merged. Thx.

PR #1680, opened 2019-03-18
Fixes #1679 - Get the attribute from delegateTarget

### LINKED ISSUE #1679: ClayAutocomplete - ClayMultiselect selection bug
Sometimes the selected item in the autocomplete list does not match with the label entered in the multiselect component. I have tried to find a pattern for this case and for now I just discovered that it fails when we select the first element matched in the autocomplete list. 

Steps to reproduce it: 

1. Enter a letter to filter elements and get some results in the autocomplete list.
2. Select the first element
3. Enter again the same letter for filtering
4. Select antoher element from the autocomplete list --> This creates the item selected in the step 2

Discovered in liferay portal: https://issues.liferay.com/browse/LPS-90863. See the attached file [here](https://issues.liferay.com/secure/attachment/303400/Autocomplete_selection_bug.mp4).

matuzalemsteles:


carloslancha:
Merged, thx!

PR #1688, opened 2019-03-20
Revert #1634 - Headers size

marcoscv-work:


marcoscv-work:
This is only for 2.x-develop branch

matuzalemsteles:
hey @marcoscv-work, why are we just keeping on 2.x? will we release the clay-css for 3.x as well?

marcoscv-work:
> hey @marcoscv-work, why are we just keeping on 2.x? will we release the clay-css for 3.x as well?

Hey @matuzalemsteles we want this change but right now in liferay portal we cannot solve some issues related with this change, so we can keep it in 3.x and revert it for 2.x. in this way we will not loss the change

matuzalemsteles:
Oh okay. Thanks @marcoscv-work for the info.

PR #1695, opened 2019-03-21
Fixes #1097 |  Validate title attribute to prevent display Tooltips with empty messages

### LINKED ISSUE #1097: Tooltip is displayed with empty content
I think the tooltip has never been displayed with empty content

carloslancha:


diegonvs:
Thanks you Carlos!

PR #1698, opened 2019-03-21
Fixes #1693 - Replace title ascending and descending by reverse-sort-direction

### LINKED ISSUE #1693: Order button tooltip text at managemnet toolbar
We have decided to change the tooltip text from "Ascending" or "Descending" to "Reverse sort direction". Please change it in Clay and Portal

matuzalemsteles:


packages/clay-management-toolbar/src/ClayManagementToolbar.soy
```diff
@@ -707,15 +707,9 @@
...
-										{msg desc=""}
-											descending
-										{/msg}
-									{else}
-										{msg desc=""}
-											ascending
-										{/msg}
-									{/if}
+									{msg desc=""}
+										reverse-sort-direction
```
matuzalemsteles: hey @carloslancha I'm not sure if we should also apply this same text to when the button is disabled, it's currently `order` (https://github.com/liferay/clay/pull/1698/files#diff-54445d58e09b53d03b6d0558828824f6R726).

hey @victorvalle you can check this out.

victorvalle:
So the button tooltip text must be always "Reverse sort direction"
If it makes sense when to show the tooltip on hover when disabled... I don't think so. But @marcoscv-work could give a more accurate answer.

PR #1703, opened 2019-03-22
 Fixes #1701 - Compare the label to see if it exists in the selected items

### LINKED ISSUE #1701: ClayMultiSelect: adds duplicated items if label and values are different
In portal we are using the multiselect component for categories. For them, the label is the name of the category and the value its id. (For tags, label and value is the same).

It happens that we can add the same category multiple times, because when checking for duplicates its comparing value with label. Check method `_handleItemAdded` in ClayMultiSelect: 

```
!this.selectedItems.find(
      itemSelected =>
       this._performCall(this.valueLocator, itemSelected) === label
)
```

**Solution:** 
compare same atribute, label or value. 
`this._performCall(this.labelLocator, itemSelected) === label`

matuzalemsteles:


carloslancha:
Merged, thx!

PR #1706, opened 2019-03-22
Copy ColorPicker and port to typescript

bryceosterhaus:
Still waiting on @pat270 for the css portion of this to go into `clay-css`. Wanted to get the TS version sent over for review though. Let me know what you guys think

### DIFF
--- a/packages/clay-color-picker/package.json
+++ b/packages/clay-color-picker/package.json
@@ -1,0 +1,41 @@
+{
+	"name": "@clayui/color-picker",
+	"version": "3.0.0",
+	"description": "ClayColorPicker component",
+	"license": "BSD-3-Clause",
+	"repository": "https://github.com/liferay/clay/tree/master/packages/clay-color-picker",
+	"engines": {
+		"node": ">=0.12.0",
+		"npm": ">=3.0.0"
+	},
+	"main": "lib/ClayColorPicker.js",
+	"esnext:main": "src/ClayColorPicker.js",
+	"jsnext:main": "src/ClayColorPicker.js",
+	"files": [
+		"lib",
+		"src"
+	],
+	"scripts": {
+		"build": "babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "tsc --project ./tsconfig.declarations.json",
+		"prepublishOnly": "npm run build && npm run build:types",
+		"start": "webpack-dev-server --mode development",
+		"test": "jest --config ../../jest.config.js"
+	},
+	"keywords": [
+		"clay",
+		"react"
+	],
+	"dependencies": {
+		"@types/tinycolor2": "^1.4.1",
+		"classnames": "^2.2.6",
+		"tinycolor2": "^1.4.1"
+	},
+	"peerDependencies": {
+		"react": "^16.8.1",
+		"react-dom": "^16.8.1"
+	},
+	"browserslist": [
+		"extends browserslist-config-clay"
+	]
+}
--- a/packages/clay-color-picker/package.json
+++ b/packages/clay-color-picker/package.json
@@ -27,9 +27,11 @@
 		"react"
 	],
 	"dependencies": {
-		"@types/tinycolor2": "^1.4.1",
 		"classnames": "^2.2.6",
 		"tinycolor2": "^1.4.1"
+	},
+	"devDependencies": {
+		"@types/tinycolor2": "^1.4.1"
 	},
 	"peerDependencies": {
 		"react": "^16.8.1",
--- a/packages/clay-color-picker/src/ClayColorPicker.tsx
+++ b/packages/clay-color-picker/src/ClayColorPicker.tsx
@@ -103,7 +103,7 @@ interface Props {
 	value?: string;
 }
 
-const ColorPicker: React.SFC<Props> = ({
+const ColorPicker: React.FunctionComponent<Props> = ({
 	ariaLabels = DEFAULT_ARIA_LABELS,
 	colors = [],
 	label,
--- a/packages/clay-color-picker/src/ClayColorPicker.tsx
+++ b/packages/clay-color-picker/src/ClayColorPicker.tsx
@@ -1,0 +1,260 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import './ClayColorPicker.scss';
+import Basic from './Basic';
+import Custom from './Custom';
+import React, {useEffect, useRef, useState} from 'react';
+import Splotch from './Splotch';
+import tinycolor from 'tinycolor2';
+import {sub} from './util';
+import {useHexInput} from './hooks';
+
+const ESC_KEY_CODE = 27;
+
+const DEFAULT_COLORS = [
+	'#000000',
+	'#5F5F5F',
+	'#9A9A9A',
+	'#CBCBCB',
+	'#E1E1E1',
+	'#FFFFFF',
+	'#FF0D0D',
+	'#FF8A1C',
+	'#2BA676',
+	'#006EF8',
+	'#7F26FF',
+	'#FF21A0',
+	'#FF5F5F',
+	'#FFB46E',
+	'#50D2A0',
+	'#4B9BFF',
+	'#AF78FF',
+	'#FF73C3',
+	'#FFB1B1',
+	'#FFDEC0',
+	'#91E3C3',
+	'#9DC8FF',
+	'#DFCAFF',
+	'#FFC5E6',
+	'#FFD9D9',
+	'#FFF3E8',
+	'#B1EBD5',
+	'#C5DFFF',
+	'#F8F2FF',
+	'#FFEDF7',
+];
+
+const DEFAULT_ARIA_LABELS = {
+	selectColor: 'Select a color',
+	selectionIs: 'Color selection is {0}',
+};
+
+interface Props {
+	/**
+	 * Labels for the aria attributes
+	 */
+	ariaLabels?: {
+		selectColor: string;
+		selectionIs: string;
+	};
+
+	/**
+	 * List of color hex values
+	 */
+	colors?: string[];
+
+	/**
+	 * The label describing the collection of colors in the menu
+	 */
+	label?: string;
+
+	/**
+	 * The input attribute for name
+	 */
+	name?: string;
+
+	/**
+	 * Callback for when the list of colors change
+	 */
+	onColorsChange?: (val: string[]) => void;
+
+	/**
+	 * Callback for when the selected color changes
+	 */
+	onValueChange?: (val: string) => void;
+
+	/**
+	 * Determines if the hex input should render
+	 */
+	showHex?: boolean;
+
+	/**
+	 * Determines if the native color picker should be used
+	 */
+	useNative?: boolean;
+
+	/**
+	 * Value of the selected color hex
+	 */
+	value?: string;
+}
+
+const ColorPicker: React.SFC<Props> = ({
+	ariaLabels = DEFAULT_ARIA_LABELS,
+	colors = [],
+	label,
+	name,
+	onColorsChange = () => {},
+	onValueChange = () => {},
+	showHex = false,
+	useNative = false,
+	value = '#FFFFFF',
+}) => {
+	const containerRef = useRef<HTMLDivElement>(null);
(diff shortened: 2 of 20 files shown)


packages/clay-color-picker/src/ClayColorPicker.tsx
```diff
@@ -0,0 +1,260 @@
...
+						showHex ? ' input-group-prepend' : ''
+					}`}
+				>
+					<span className="input-group-text input-group-text-secondary">
+						<div className="open-control open-checkbox">
+							<Splotch
+								aria-label={ariaLabels.selectColor}
+								className="btn btn-secondary"
+								onClick={() =>
+									useNative && valueInputRef.current
```
matuzalemsteles: hey @bryceosterhaus, I see a lot of the use of arrow functions and many logics being done directly in `onXXXX`... I personally prefer to call a function with a better description of what they are doing, so I can better understand the code, even though the Attempt here is to minimize the code, we can impact the maintainability and readability and maybe it did not have as much effect on production, I think declaring these functions before JSX is easy to understand what the component is doing.

packages/clay-color-picker/package.json
```diff
@@ -0,0 +1,41 @@
...
+		"prepublishOnly": "npm run build && npm run build:types",
+		"start": "webpack-dev-server --mode development",
+		"test": "jest --config ../../jest.config.js"
+	},
+	"keywords": [
+		"clay",
+		"react"
+	],
+	"dependencies": {
+		"@types/tinycolor2": "^1.4.1",
```
matuzalemsteles: I'm not sure if we should add as a dependency. I do not know what the pattern is out there about the types... Can you check this?

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -0,0 +1,249 @@
...
+import React, {useState} from 'react';
+import Splotch from './Splotch';
+import tinycolor from 'tinycolor2';
+import {useHexInput} from './hooks';
+
+/**
+ * Renders custom color icon
+ * @return {React.Component}
+ */
+const CustomColorIcon: React.SFC = () => {
```
matuzalemsteles: I'm thinking if maybe we can remove the `return` in these cases from functions with just the JSX and use this as default, something like this:
```typescript
const CustomColorIcon: React.SFC = () => (
  <svg ...>
     ...
  </svg>
);
```

what do you think?

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -0,0 +1,249 @@
...
+	return (
+		<svg
```
matuzalemsteles: We can use the lexicon icons directly without having to add the svg hard code. Maybe we can create an Icon component soon, since we can use it a lot.

See example https://github.com/liferay/clay/pull/1440/files?file-filters%5B%5D=.tsx#diff-4fb77ee218c192d6a19607263ff8d304

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+interface RGBInputProps {
+	onChange: ({}) => void;
+	name: string;
+	value: number;
+}
+
+/**
+ * Renders input that displays RGB values
+ * @return {React.Component}
```
matuzalemsteles: We no longer need JSDOC, now that we're using TypeScript.

packages/clay-color-picker/src/ClayColorPicker.tsx
wincent: As far as reading the code goes, I don't really need a name here to add my understanding because its obvious that the function in `onClick={() => {}}` is a `handleClick` kind of function. The main issue, for me, is that anonymous functions are less helpful in stack traces. Of the different ways of naming the functions, what do you prefer?

```javascript
// Old school: may require a `.bind()` if you need to use `this`
onClick={function handleClick() {
}}

// Arrows: above...
const handleClick = () => { /* ... */ };
// then...
onClick={handleClick}
```

packages/clay-color-picker/package.json
wincent: Should probably be a `devDependency`; these packages don't do anything at runtime.

packages/clay-color-picker/src/ClayColorPicker.tsx
matuzalemsteles: Although the name of the property say much I do not agree that it says everything the question is that we have arrows above, can simplify the name of the functions and add descriptions if necessary to improve understanding, in my view I think add the whole logic above JSX, creates a mess and can make a salad, I prefer to add less logic over the JSX and separate a little. But I prefer arrow functions above 🙂

bryceosterhaus: I'd be hesitant to say we always need to put function handlers above and non-anonymous. A lot of times we might have something as simple as `onClick={() => setActiveIdHook(id)}` and I would hate to have to always move that to the top. Also we will likely tend to just call the function `handleXXX`. Even for the sake of stack traces, I think its helpful to have the name of the function, but also if its a simple enough function it probably isnt a use deal? For example, do we really think this [inline anonymous function](https://github.com/liferay/clay/pull/1706/files#diff-2919976302e4677a0f68885ac3357a9aR32) really needs to be moved to the top? Seems a bit overkill for me and will make readability a bit more difficult.

Now, trying to find a middle ground since I also understand your concern about too much logic in the jsx, what if we decided something along the lines of _"if its an implicit return, inlining the function is fine"_. Granted, that is very subjective and probably can't be automated with current eslint rules. Or we say if the function is more than 3 lines, move to the top?

packages/clay-color-picker/src/Custom.tsx
bryceosterhaus: How do you feel about the pattern of having to pass `spriteMap` into the top level component? I saw you did it here, https://github.com/liferay/clay/pull/1440/files?file-filters%5B%5D=.tsx#diff-f83ebf28c6e7caeca167ae49475dcec9R20. Seems a bit clunky to me, since the the intent of the component is to be based on clay-css's icons. If we follow this pattern, every component that uses a clay icon, we have to provide a `spriteMap` prop at the top level component.

I was thinking in `clay-icon` we can reference clay-css's icon sprite sheet directly, that way we only have to reference the symbol. Then once `clay-icon` is merged(I'm working on it now), we can update this hardcoded value for that component.

Thoughts?

matuzalemsteles: - reference to clay-css: The concern about this was that there could be cases of loading more than one `icons.svg` in different parts of the application and in the portal and the network overhead it can have, can make the pages slow, this file is a bit large and inside the portal we are loading it.
- href hard code: we had these concerns before about adding the spritemap hard code, the concern was that in the portal the address may be different and we must give freedom to modify this URL. Because of this we had to add `spritemap` to all our components.

I think we can try to minimize that impact a bit, ClayIcon may have support for getting the `context` spritemap for full React applications, you can add the `provider` to your root component to pass the spritemap information, so you do not have to go through for all components.

packages/clay-color-picker/src/ClayColorPicker.tsx
matuzalemsteles: hey @bryceosterhaus I understand that we have cases that are not really necessary, especially when we have only one function being called. Actually we have to have the good sense of not being able to abuse to adds JSX, I do not know if the rule of going from 3 lines should be the right one to decide this, I think it goes from case to case, we get some cases we could be calling a function that does not make much sense being in the prop `onClick` for example but can also have a function that makes full sense is there.

Summarizing what I mean in the end is that separating the two environments in my view helps more to understand the code before arriving in JSX. But I'm not saying we should remove all the logics that could be there.

packages/clay-color-picker/src/Custom.tsx
bryceosterhaus: Oh good idea with context. I'll submit a PR for icon today to get some feedback on that idea. Gonna see about allowing either a prop or context

jbalsas: In a way, that's how @wincent and @julien ended up doing [some of it in AlloyEditor](https://github.com/liferay/alloy-editor/blob/master/src/components/buttons/button-icon.jsx#L16).

One thing that's wrong there, and might need to rethink here as well is the `href` on svgs. SVGs are a basic static resource to host on a CDN. Unfortunately, svg does not support CORS by spec (at least last time I checked), so the following:

```
<svg href="cdn.foo.com/icons.svg">...</svg>
```

Will throw a CORS error when accessed from a different domain (`main.foo.com`). To still be able to load the svg spritemap, we're using [svg4everybody](https://github.com/jonathantneal/svg4everybody), which will use AJAX (granted the server is properly configured) to overcome this limitation. That, however, won't save us from the initial browser-triggered request, so the final change we might need to do is to use `data-href` instead of `href`.

At least, those are the things that we need to consider... another option would be to fin an alternative way to do this :)

jbalsas: For additional reference:
- https://bugs.chromium.org/p/chromium/issues/detail?id=470601
- https://www.w3.org/Graphics/SVG/WG/track/actions/3781

bryceosterhaus: @matuzalemsteles check out https://github.com/liferay/clay/pull/1725/files for clay-icon with context. You can look at the demo file to see how it is used.

jbalsas: More references:
- https://issues.liferay.com/browse/LPS-71763
- https://issues.liferay.com/browse/LPS-72903

Please, keep these in mind so we don't regress there

matuzalemsteles: thanks @jbalsas for context, one problem with using `href` or `data-href` (I'm not sure about `date-href`) is that it does not work in Safari 😞.

Thanks @bryceosterhaus I'm going to take a look, just a few concerns I have now, is that maybe it's more a context for the application, some applications may have many contexts aligned, I do not really like it, but if Clay comes to use context, maybe we can create a context for Clay and use it on every component you need. Something like `ClayContext`. Just to not have a connection for each component.

jbalsas: I meant with the `<use>` tag. And the `data-href` needs to be used in conjunction with something like `svg4everybody` properly configured to fetch from `data-href` and avoid the browser initiating a request.

Another option we explored was to "force" adding the spritemap in the page and then just refer to the ids locally

matuzalemsteles: Oh okay, makes sense now 🙂

bryceosterhaus: @matuzalemsteles 
>  some applications may have many contexts aligned, I do not really like it, but if Clay comes to use context, maybe we can create a context for Clay and use it on every component you need. Something like ClayContext. Just to not have a connection for each component.

Even if an app uses many contexts, this shouldn't be too big of a problem since you can create an global context that consumes the other contexts. This could be something we could try for a global ClayContext. Each component can independently create their own, but also we can join them all together in a separate util that provides a global one.

Also, by allowing for context here, its just another feature and they can always use the spriteMap prop if they prefer.

matuzalemsteles: Good idea, I had not thought about it.

wincent: Long thread, but yes, this kind of pattern (having to pass something down all the way from the top to a deeply nested component that renders an icon and is used everywhere) is an ideal use case for context. We should definitely explore that.

packages/clay-color-picker/src/ClayColorPicker.tsx
wincent: The `> 3 lines` heuristic sounds pretty good to me, especially if we pair it with guidance to use your judgment on a case-by-case basis. I know we like having simple rules that we can apply uniformly everywhere, but sometimes having an element of flexibility is actually for the best.

matuzalemsteles: It seems good to me too, to go with that simple rule but tied to your case-by-case judgment, it would be interesting if we could add those decisions somewhere so that the next people will look at it. Be a JS style guide...

bryceosterhaus:
Updated. Still waiting on CSS to go into `clay-css` and I will eventually remove the hard coded SVG once `Icon` is in upstream.

bryceosterhaus:
new PR with patrick's css changes. https://github.com/liferay/clay/pull/1857

PR #1715, opened 2019-03-26
Fixes #1713 ClayCSS Update User Card Icon sizing and add a 5 column grid for User Card

### LINKED ISSUE #1713: User Card Update
https://issues.liferay.com/browse/LEXI-503

We need to update the user card to avoid text inside the sticker. 

1. Stickers shouldn't allow text in order to improve accessibility and language changes.
2. As in the latest documentation update the sticker size should go from a minimum size of 48px to a maximum size of 80px.
3. Use the User Icon inside the sticker a a placeholder when the User Image is not provided.

pat270:


matuzalemsteles:

packages/clay-css/src/content/test_card_view_template.html
```diff
@@ -468,10 +468,10 @@ <h4 class="component-title">
 	<div class="container-fluid container-fluid-max-xl container-view sidenav-content">
 		<form>
 			<ul class="card-page">
-				<li class="card-page-item-header">
+				<li class="card-page-item card-page-item-header">
 					<h1 class="card-section-header">Group 1</h1>
 				</li>
-				<li class="card-page-item-directory">
+				<li class="card-page-item card-page-item-directory">
```
matuzalemsteles: hey @pat270 we have visual difference between the `.card-page-item-directory` and `.card-page-item-asset` classes?

packages/clay-css/src/scss/components/_cards.scss
```diff
@@ -418,16 +418,24 @@
...
 .card-page {
 	@extend %clay-custom-grid-wrapper;
-}
 
-.card-page-equal-height {
-	.card-page-item {
-		display: flex;
-		flex-direction: column;
-	}
+	&.card-page-equal-height {
```
matuzalemsteles: What use cases should we add this to ClayCardGrid?

packages/clay-css/src/content/test_card_view_template.html
pat270: @matuzalemsteles no visual difference at this time. It's there just in case the Lexicon Team decides to deviate the layout for `card-type-asset` from `card-type-directory`. If it's easier to manage we can do something like `card-page-item` on the `li` and add a column class on `card-page card-page-4-col`, `card-page card-page-5-col`, etc. I will need to add these though.

packages/clay-css/src/scss/components/_cards.scss
pat270: This is UX preference. I can't think of a rule that would cover every case. In Portal, I would say all the cards should have this class except for http://pat270.github.io/lexicon/v2.10.0/content/test-card-custom-fields-view/

packages/clay-css/src/content/test_card_view_template.html
matuzalemsteles: Thanks @pat270, for a moment we are always adding `.card-page-item.card-page-item-directory` we can keep this for now and when we have any changes on this, we will change.

packages/clay-css/src/scss/components/_stickers.scss
```diff
@@ -18,12 +18,6 @@
 	text-align: center;
 	vertical-align: middle;
 
-	.lexicon-icon {
```
matuzalemsteles: hey @pat270 removing it means we will no longer have icons in Sticker?

pat270: @matuzalemsteles This broke icon alignment in stickers in IE11. We still have icons in stickers.

carloslancha:
Merged, thx!

PR #1720, opened 2019-03-26
 Fixes #1618 - stop sending two events by clicking the checkbox and radio

### LINKED ISSUE #1618: Checkbox, Radio and Toggle components of Clay lose focus and get it back when they're clicked
From [JIRA](https://issues.liferay.com/browse/LPS-91383):

```
Steps to reproduce

Go to https://clayui.com/docs/components/forms/radio-check-toggle.html.
Add focus to one of the following components: Checkbox, Radio and Toggle.
Click on one of them without release the mouse button.
Check the focus is lost.
Release the mouse button and the focus is back to the clicked component.
```

This issue is causing an error in another components like ClayDropdown using radio, when we listen to the component `itemClicked` event the event is emitted twice.

A temporary solution proposed by @matuzalemsteles was change the input type radio, and checkbox z-index to value `2`.

@NemethNorbert @natocesarrego @matuzalemsteles

matuzalemsteles:


matuzalemsteles:
hey @pat270 this markup change of Checkbox would cause a break in ClayCard, I'm trying to do something that will not change the markup, but this is very complicated since it is a default browser behavior, I can not help it. Even avoiding sending two events, I can not mark the checkbox as checked through JS in case the markup remains as it is so as not to break change. The way I did ends up working fine for use cases with Dropdown, but for Card the interface breaks.

PR #1725, opened 2019-03-27
Fixes #1711 - Create clay-icon component

### LINKED ISSUE #1711: Create clay-icon component for v3.0.0
Need to implement in React

bryceosterhaus:


matuzalemsteles:

packages/clay-icon/src/ClayIcon.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface Props extends React.SVGAttributes<SVGSVGElement> {
+	className?: string;
+	spriteMap?: string;
+	symbol: string;
+}
+
+const ClayIcon: React.SFC<Props> = ({
+	className,
+	spriteMap,
```
matuzalemsteles: I think we can instantiate `useContext` here to remove the condition, I do not know if it would work or would break some Hooks rule.
```suggestion
	spriteMap = useContext(ClayIconSpriteContext),
```

packages/clay-icon/src/ClayIcon.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const ClayIconSpriteContext = React.createContext('');
+
+interface Props extends React.SVGAttributes<SVGSVGElement> {
+	className?: string;
+	spriteMap?: string;
+	symbol: string;
+}
+
+const ClayIcon: React.SFC<Props> = ({
```
matuzalemsteles: If this component is ready to go, just change the `SFC` to `FunctionComponent`.

packages/clay-icon/src/ClayIcon.tsx
matuzalemsteles: We could add a warning to help dev too, for example if spritemap is not passed via prop or does not have the provider in the application, it did not render the component and emit a error in the console for the developer.

That would be just a suggestion if we could adopt on Clay, use the https://www.npmjs.com/package/tiny-invariant lib, helps create those messages and it is also interesting to make these conditions, but we should ensure to add only in dev environment.

bryceosterhaus:
Just updated. I added a few things

- `svg4everybody`, not entirely sure if this is the best way to go about it but thought it would be appropriate here. Maybe we should make it configurable though.
- `warning` helper for users to know you need at least one means of spriteMap
     - `babel-plugin-dev-expression` used to optimize the `warning` for production builds
- `custom-types` directory to specify any updated or additional types we need for 3rd party packages.

Let me know what you think!

babel.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = api => {
 
 	return {
 		ignore: ['node_modules', '**/__snapshots__/**', '**/__tests__/**'],
+		plugins: ['babel-plugin-dev-expression'],
```
wincent: One thing about `babel-plugin-dev-expression` is that it transforms to a check against `process.env.NODE_ENV`, right? Does this mean that we have an implicit dependency on some other plugin (maybe "babel-plugin-transform-node-env-inline") to provide a `process.env` stub in the browser environment?

matuzalemsteles:

packages/clay-icon/src/ClayIcon.tsx
```diff
@@ -0,0 +1,50 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import classNames from 'classnames';
+import svg4everybody from 'svg4everybody';
```
matuzalemsteles: I'm not sure if this package should be here, since we already have it in the Portal.

custom-types/svg4everybody/index.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// Copied code from https://github.com/DefinitelyTyped/DefinitelyTyped
+// Added `attributeName` to interface to account for version 2.1.9
```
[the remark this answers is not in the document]
bryceosterhaus: Yeah I am gonna open a PR today and update the types for the newest version

packages/clay-icon/src/ClayIcon.tsx
bryceosterhaus: @jbalsas do you think this package should be included to Icon? Other possibilities would be remove it entirely or add a config to enable it.

babel.config.js
bryceosterhaus: Yeah I actually just remembered that after I left work last night. I'm gonna look into what the best solution is.

custom-types/svg4everybody/index.d.ts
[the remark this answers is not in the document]
bryceosterhaus: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34326

Not sure how long it usually takes to get merged and published.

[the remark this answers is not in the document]
wincent: Great. Seems it got merged pretty quickly, and I see a release of `@types/svg4everybody` 2 days ago, so that's probably it.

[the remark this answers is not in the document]
bryceosterhaus: Updated to use the published version

babel.config.js
bryceosterhaus: I ended up just adding `babel-plugin-transform-node-env-inline`. I'm open ears for other ideas, but this seems the most straight forward. I really like the idea of trying to use `warning()` within the clay library, I think we can really benefit in creating helpful warnings.

bryceosterhaus:
@matuzalemsteles @wincent updated pr.


@jbalsas what're your thoughts on using `svg4everybody` within this component itself? If you run the demos, you'll see it working for cdn resources.

matuzalemsteles:

packages/clay-icon/src/ClayIcon.tsx
```diff
@@ -0,0 +1,56 @@
...
+	...otherProps
+}) => {
+	let spriteMapVal = React.useContext(ClayIconSpriteContext);
+
+	if (spritemap) {
+		spriteMapVal = spritemap;
+	}
+
+	warning(
+		spritemap,
```
matuzalemsteles: We should check `spriteMapVal` as well.

packages/clay-icon/demo/App.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import ClayIcon, {ClayIconSpriteContext} from '../src/ClayIcon';
+
+import 'clay-css/lib/css/atlas.css';
+
+const spritemap = 'node_modules/clay-css/lib/images/icons/icons.svg';
+
+const App: React.SFC = () => (
```
matuzalemsteles: We must use `React.FunctionComponent` here as well.

bryceosterhaus:
updated

matuzalemsteles:

packages/clay-icon/src/ClayIcon.tsx
jbalsas: Based on https://github.com/jonathantneal/svg4everybody/blob/master/dist/svg4everybody.js#L99, I think a config to enable it should be fine, otherwise we risk having multiple intervals running and making requests if someone else initializes it on their own.

Other option would be to investigate alternative ways to incorporate the svg spritemap to avoid this issue completely, which could be an unexpected and welcomed outcome ;)

bryceosterhaus:
@matuzalemsteles I went ahead and removed `svg4everybody` since there is hesitation on whether that is the right solution for portal or not. A user can always add it to their app if they need that support. I also went ahead and squashed some of the middle commits since they referred to svg4everybody. 

I think this component should be good to go now.

matuzalemsteles:
Thanks a lot @bryceosterhaus! Merged!

PR #1726, opened 2019-03-27
Fixes #1618 - stop sending two events by clicking the checkbox and radio

### LINKED ISSUE #1618: Checkbox, Radio and Toggle components of Clay lose focus and get it back when they're clicked

matuzalemsteles:


packages/clay-dropdown/src/ClayDropdownItem.soy
```diff
@@ -181,7 +184,10 @@
 			{/if}
 		"
 
-		data-onclick="{$_handleItemClick}"
+		{if not $disabled}
+			data-onchange="{$_handleItemClick}"
```
carloslancha: onChange to a `li` element?

matuzalemsteles: `li` will receive the `change` event (via bubbling) when the input of the checkbox changes (being triggered by clicking on the label or directly in the input), I had to add to `li` because of the rules we have on top of it .

matuzalemsteles:
hey @carloslancha can you look at this? it is important that this also enters the Monday release.

packages/clay-dropdown/src/ClayDropdownItem.soy
```diff
@@ -186,8 +190,12 @@
...
+		{/if}
+
 		data-onkeydown="{$_handleItemKeyDown}"
```
jbalsas: Shouldn't `onKeyDown` receive the same treatment?

packages/clay-dropdown/src/ClayDropdownItem.soy
```diff
@@ -278,7 +286,10 @@
...
```
jbalsas: Why not `onkeydown`?

packages/clay-dropdown/src/ClayDropdownItem.soy
matuzalemsteles: Even if we move to the condition it has no effect, we have to do the `disabled` treatment inside the function to ignore the tab and find a simpler way to know this information as well.

packages/clay-dropdown/src/ClayDropdownItem.soy
matuzalemsteles: For some reason we do not have `delegateTarget` when we only use `onkeydown`.

PR #1738, opened 2019-03-28
Fixes #1710 - Update date-picker to use @clayui/button

### LINKED ISSUE #1710: Create clay-button component for v3.0.0
Need to implement in React

bryceosterhaus:


packages/clay-button/src/ClayButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+	<button
+		{...otherProps}
+		className={classNames(className, 'btn', `btn-${displayType}`, {
+			'btn-block': block,
+			'btn-monospaced': monospaced,
+			'btn-sm': small,
+		})}
+		type={type}
+	>
+		{children}
```
matuzalemsteles: I would probably expect the Button to provide itself the option of rendering an icon on the left or right side without having to pass an icon component.

We can provide the two if necessary, a `ButtonLite`(ButtonBase...) and `Button` with the same behaviors expected by the Lexicon.

bryceosterhaus: So one of the reasons I decided not to add `Icon` in this component is because there was no specific css classnames that required `Button` to have knowledge of `Icon`. I also wanted to keep the dependency footprint as small as possible and this also requires the usage to be as explicit as possible.

You can get the same outcome of using an Icon by passing it as a child to Button.
```jsx
<Button>
      <Icon>
      {'Click Me!'}
</Button>
```

packages/clay-button/src/ClayButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import * as React from 'react';
+import classNames from 'classnames';
+
+export const DISPLAY_TYPES = ['primary', 'secondary', 'link', 'unstyled'];
+
+interface Props extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+	block?: boolean;
+	displayType?: typeof DISPLAY_TYPES[number];
```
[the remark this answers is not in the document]
bryceosterhaus: The main reason I use a constant is so that users can import the types and see what is available. If the user is using TS, they wont need it because they an infer what is available, but someone is using regular js, they can import to see what types are available. This may be overkill though...

packages/clay-button/src/ClayButton.tsx
matuzalemsteles: Yes I really like to use it that way, we just have to be careful not to break the expected behaviors so much and let Dev have to do a lot to achieve the behavior.

We should just remember to add these changes in the "Migrate the Clay components from v2 to v3" tab.

packages/clay-button/src/ClayButton.tsx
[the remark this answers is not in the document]
matuzalemsteles: I understand your concern about this, but we do not have to worry when people are not using TS if we can not fill our components with utilities... we can delegate that responsibility to documentation.

packages/clay-button/src/ClayButton.tsx
bryceosterhaus: Yeah thats actually a good reminder that we should probably document the changes from v2 somewhere with each new component. Should this just be added to `clayui.com`?

matuzalemsteles: I'm thinking of creating a guide for migrating from v2 to v3, it will be available in the new clayui.com document in #1525 and maybe a redirect link from version v2 to v3.

bryceosterhaus: That sounds good to me. For the time being, where should we put these bits of into? Start a google doc?

matuzalemsteles: Yes I think we should start to add this information somewhere as we create new components. I think we can create the `migrate-the-clay-components-from-v2-to-v3.md` file and add it to the root folder maybe, then we can move to clayui.com, I think we'll have more control if this file is here.

bryceosterhaus: 👍 I went ahead and added that file, we can always remove it later.

bryceosterhaus:
updated

migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -0,0 +1,23 @@
+# Migrating clay components from v2 to v3
+
+Noteable Changes
+
+## clay-button
+
+-   Removed icon dependency within button itself
+
+```js
```
matuzalemsteles: Maybe using `diff` would have made it clearer, what do you think?

```diff
<Button
-	label="Icon positioned at right"
-	icon="plus"
-	iconAlignment="right"
-	spritemap={spritemap}
-/>
+>
+      {'Icon positioned at right'}
+      <Icon spritemap={spritemap} symbol="plus"/>
+</Button>
```

bryceosterhaus: Personally I would probably avoid using diffs in documentation just because I think they can be heard to read. But if you think it'd be a big help, I'm happy to do whatever.

migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -0,0 +1,23 @@
+# Migrating clay components from v2 to v3
+
+Noteable Changes
+
+## clay-button
```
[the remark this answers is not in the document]
bryceosterhaus: Yeah I agree. I think we should use this file to _over provide_ information and ideas behind changes from the devs themselves. And then later I think we need to take this information and condense it and filter it a bit to add on the official docs site.

migrate-the-clay-components-from-v2-to-v3.md
wincent: I think it mostly comes down to size: in small examples like this, I can easily see what has changed from v2 to v3 and seeing the examples written out in full is nice and easy. On the other hand, if the examples are large, then it will be hard for me to see exactly what's different without the aid of a diff. Hopefully we won't have too many that are like that, but we can evaluate those when we come to them.

migrate-the-clay-components-from-v2-to-v3.md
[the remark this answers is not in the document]
matuzalemsteles: yeah yeah, are some of my thoughts on this, especially in the component documentation part, it will be manual work but it is necessary to build good documentation and show our thoughts. I'm excited about that 🙂

migrate-the-clay-components-from-v2-to-v3.md
matuzalemsteles: @bryceosterhaus I do not see problem in not using in some cases like this, but for me it becomes clearer when I see a diff, some not so much ... but I agree with @wincent on bigger changes (I also hope that we do not change many things) maybe is necessary.

bryceosterhaus:
@matuzalemsteles updated with diff

PR #1746, opened 2019-03-29
Fixes #1737 - Create checkbox component

### LINKED ISSUE #1737: Create clay-checkbox component for v3.0.0

bryceosterhaus:


packages/clay-checkbox/src/ClayCheckbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+					{...otherProps}
+					checked={checked}
+					className={classNames('custom-control-input', className)}
+					ref={inputRef}
+					type="checkbox"
+					value={checked.toString()}
+				/>
+
+				{label && (
+					<span className="custom-control-label">
```
matuzalemsteles: This should be out of the condition, just your content, we should also add an API or use the children so we can add other content that is not text, we will be using it in a ClayCard for example.

See https://github.com/liferay/clay/blob/2.x-develop/packages/clay-checkbox/src/ClayCheckbox.soy#L93-L108

packages/clay-checkbox/src/ClayCheckbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+						<span className="custom-control-label-text">
```
matuzalemsteles: At some point we will need to add the `sr-only` class. Previously we had `showLabel` API for this.

bryceosterhaus: I talked to @pat270 about this and we settled on removing `sr-only` and having people rely on using aria attributes.

packages/clay-checkbox/src/ClayCheckbox.tsx
bryceosterhaus: I was trying to avoid using multiple inter-dependent props such as `showLabel`, `label`, and `children`. By using just label, we can assure its the only place to add content. One way to allow for more than text would be to check the type of the label and conditionally render the class names. What do you think?

packages/clay-checkbox/src/ClayCheckbox.tsx
matuzalemsteles: Hmm, we have to be careful about that. We have to add this information in the migration guide and also add as a good practice in the new documentation.

packages/clay-checkbox/src/ClayCheckbox.tsx
matuzalemsteles: Hmm, just be concerned that the API name is not self explanatory for this use case. But it sounds good to me to.

bryceosterhaus: Ill add it right now and test it out. Maybe be some edge case where a user wants both text and markup. Ill update the pr and let you know

bryceosterhaus:
@matuzalemsteles updated. I included a aria-label example in the demo. Once the `v2-v3.md` file is in, I will also update that doc.

matuzalemsteles:

packages/clay-checkbox/src/ClayCheckbox.tsx
```diff
@@ -0,0 +1,70 @@
...
+				/>
+
+				<span className="custom-control-label">
+					{label && (
+						<span className="custom-control-label-text">
+							{label}
+						</span>
+					)}
+
+					{children}
```
matuzalemsteles: By markup it is expected that the content stays outside the `span`.

bryceosterhaus:
@matuzalemsteles updated and moved markup outside of span

bryceosterhaus:
@matuzalemsteles updated and fixed conflicts

matuzalemsteles:
Merged!

PR #1748, opened 2019-04-01
Fixes #1732 - Replace h1 in favor of h5 in the main row of the table

### LINKED ISSUE #1732: Dataset display table row primary text using h1
Text in table rows must not be a heading

image

matuzalemsteles:


jbalsas:
Hey @matuzalemsteles, I'm not sure this is the right fix... have you cleared this out with @marcoscv-work or @eduardoallegrini?

The problem with those headers is that they convey the wrong semantic meaning creating accessibility problems. I'm not sure that just changing `h1` by `h5` would address this.

matuzalemsteles:
hey @jbalsas, thanks for clarifying this, I've been assuming they were problems with typography and not accessibility, I tried to get answers about this change but I could not.

matuzalemsteles:
I'll close this and send another PR by replacing with the tag `<p />`

PR #1750, opened 2019-04-01
Fixes #1743 ClayCSS undo Bootstrap's `text-align: left` on the `body`…

### LINKED ISSUE #1743: RTL is not working correctly with text align
Bootstrap adds a text-align: left; in the body.
https://github.com/twbs/bootstrap/blob/3e05d016cf18e99725b9e77491a501c97b8dc7ac/scss/_reboot.scss#L53

So the RTL is not working fine.

We could try with something like:
```
body[dir="rtl"] {
	direction: rtl;
	text-align: right;
}
```

Original issue:
https://issues.liferay.com/browse/LPS-92808

pat270:
… element so content with mixed languages will be properly aligned

Fixes #1743 ClayCSS Globals added option to configure `$body-text-align` and reset Bootstrap's `text-align: left` on the `body` element


jbalsas:
LGTM!

Issue #1731, opened 2019-03-28
heading problem at list row


victorvalle:
We have a heading problem in list title where the title is coded as an h4.

![image](https://user-images.githubusercontent.com/21121115/55140852-55718700-5139-11e9-87d9-d4b3f97c369c.png)

Title size should be 14px semibold (600) #272833
Moreover subtitle is wrong and should be 14px regular (400) #6B6C7E

matuzalemsteles:
The heading was removed from the documentation through #1755. @pat270 Are CSS changes already fixed? The site already has some time that has not been updated with recent clay-css changes.

pat270:
@matuzalemsteles @victorvalle The style issues aren't in Clay CSS. The reason why Clayui.com looks different is because the markup is only using `<li class="list-group-item list-group-item-flex"></li>` instead of:
```
<ul class="list-group">
    <li class="list-group-item list-group-item-flex"></li>
</ul>
```

matuzalemsteles:
Thanks @pat270 I'll take care of updating this on clayui.com

PR #1755, opened 2019-04-01
Fixes #1734, #1731 - Update ClayList markup

### LINKED ISSUE #1734: Page examples - list view template headings use
the use of headings is wrong

image

H1 should be dedicated in this case to "My Application Name" text in the header
There is no need of using any other heading

Section list must not use h3
List title must not use h4

matuzalemsteles:


carloslancha:
Merged, thx!

PR #1758, opened 2019-04-01
Fixes #1757 - Create @clay/label

### LINKED ISSUE #1757: Create label component

bryceosterhaus:
Once https://github.com/liferay/clay/pull/1738/commits/64058ec39c06dee1c7dbde8dfb194c902fe4714c is merged, I will go through and add additional docs for this component for v2-v3 migration

### DIFF
--- a/packages/clay-label/src/ClayLabel.tsx
+++ b/packages/clay-label/src/ClayLabel.tsx
@@ -12,8 +12,19 @@ type DisplayType = 'secondary' | 'info' | 'warning' | 'danger' | 'success';
 
 interface Props
 	extends React.BaseHTMLAttributes<HTMLAnchorElement | HTMLSpanElement> {
+	/**
+	 * HTML properties that are applied to the 'x' button.
+	 */
 	closeButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
+
+	/**
+	 * Determines the style of the label.
+	 */
 	displayType?: DisplayType;
+
+	/**
+	 * Flag to indicate if the label should be of the `large` variant.
+	 */
 	large?: boolean;
 }
 
--- a/packages/clay-label/src/ClayLabel.tsx
+++ b/packages/clay-label/src/ClayLabel.tsx
@@ -6,6 +6,7 @@
 
 import * as React from 'react';
 import classNames from 'classnames';
+import Icon from '@clayui/icon';
 
 type DisplayType = 'secondary' | 'info' | 'warning' | 'danger' | 'success';
 
@@ -46,7 +47,7 @@ const ClayLabel: React.FunctionComponent<Props> = ({
 						className="close"
 						type="button"
 					>
-						{closeButtonProps.children}
+						<Icon symbol="times" />
 					</button>
 				</span>
 			)}
--- a/packages/clay-label/src/ClayLabel.tsx
+++ b/packages/clay-label/src/ClayLabel.tsx
@@ -9,8 +9,9 @@ import classNames from 'classnames';
 
 type DisplayType = 'secondary' | 'info' | 'warning' | 'danger' | 'success';
 
-interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
-	closeButtonProps?: React.HTMLAttributes<HTMLButtonElement>;
+interface Props
+	extends React.BaseHTMLAttributes<HTMLAnchorElement | HTMLSpanElement> {
+	closeButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
 	displayType?: DisplayType;
 	large?: boolean;
 }
@@ -24,8 +25,18 @@ const ClayLabel: React.FunctionComponent<Props> = ({
 	large = false,
 	...otherProps
 }) => {
-	const content = (
-		<>
+	const TagName = href ? 'a' : 'span';
+
+	return (
+		<TagName
+			{...otherProps}
+			className={classNames('label', className, {
+				'label-dismissible': closeButtonProps,
+				'label-lg': large,
+				[`label-${displayType}`]: displayType,
+			})}
+			href={href}
+		>
 			<span className="label-item label-item-expand">{children}</span>
 
 			{closeButtonProps && (
@@ -39,24 +50,8 @@ const ClayLabel: React.FunctionComponent<Props> = ({
 					</button>
 				</span>
 			)}
-		</>
+		</TagName>
 	);
-
-	const attrs = {
-		...otherProps,
-		className: classNames('label', className, {
-			'label-dismissible': closeButtonProps,
-			'label-lg': large,
-			[`label-${displayType}`]: displayType,
-		}),
-		href,
-	};
-
-	if (!href) {
-		delete attrs.href;
-	}
-
-	return React.createElement(href ? 'a' : 'span', attrs, content);
 };
 
 export default ClayLabel;
--- a/packages/clay-label/src/ClayLabel.tsx
+++ b/packages/clay-label/src/ClayLabel.tsx
@@ -1,0 +1,62 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import classNames from 'classnames';
+
+type DisplayType = 'secondary' | 'info' | 'warning' | 'danger' | 'success';
+
+interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+	closeButtonProps?: React.HTMLAttributes<HTMLButtonElement>;
+	displayType?: DisplayType;
+	large?: boolean;
+}
+
+const ClayLabel: React.FunctionComponent<Props> = ({
+	children,
+	className,
+	closeButtonProps,
+	displayType = 'secondary',
+	href,
+	large = false,
+	...otherProps
+}) => {
+	const content = (
+		<>
+			<span className="label-item label-item-expand">{children}</span>
+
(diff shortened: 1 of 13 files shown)


packages/clay-label/src/ClayLabel.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type DisplayType = 'secondary' | 'info' | 'warning' | 'danger' | 'success';
+
+interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+	closeButtonProps?: React.HTMLAttributes<HTMLButtonElement>;
+	displayType?: DisplayType;
+	large?: boolean;
+}
+
+const ClayLabel: React.FunctionComponent<Props> = ({
```
matuzalemsteles: I'm finding this component a little tricky and maybe we do not need to call `React.createElement` directly. We can simplify a little:

Just a suggestion, what do you think? 🙂
```javascript
const ClayLabelContent = ({
	children,
	closeButtonProps,
}) => (
	<>
		<span className="label-item label-item-expand">{children}</span>
		{closeButtonProps && (
			<span className="label-item label-item-after">
				<button
					{...closeButtonProps}
					className="close"
					type="button"
				>
					{closeButtonProps.children}
				</button>
			</span>
		)}
	</>
);

const ClayLabel = ({
	children,
	className,
	closeButtonProps,
	displayType = 'secondary',
	href,
	large = false,
	...otherProps
}) => {
	const classNames = cN('label', className, {
		'label-dismissible': closeButtonProps,
		'label-lg': large,
		[`label-${displayType}`]: displayType,
	});

	if (href) {
		return (
			<a
				href={href}
                                 className={classNames}
				{...otherProps}
			>
				<ClayLabelContent closeButtonProps={closeButtonProps}>
					{children}
				</ClayLabelContent>
			</a>
		);
	}

	return (
		<span {...otherProps} className={classNames}>
			<ClayLabelContent closeButtonProps={closeButtonProps}>
				{children}
			</ClayLabelContent>
		</span>
	);
}
```

bryceosterhaus: Hey Matu, what exactly do you find confusing? Is it particularly the call to `React.createElement`? I was trying to avoid multiple returns and also avoiding duplicating the markup for the `content`.

matuzalemsteles: Hey Bryce, yeah I understand what you wanted to do, the big reason is that I do not really like having to call `React.createElement` directly, I'd rather let the babel presets to react transform to `React.createElement`, in this case, and I find it a bit strange that we have a condition that removes the href from the above object (it seems a bit hack), I think we have other ways of doing this without having to compromise a little beauty of our code in the development environment.

I think we do not need to focus as much on eliminating as many rows (we can then improve our build steps for distribution) by just getting in the middle so we do not remove some beauty from the code.

> Just making sure that we can have other ways of doing this.

bryceosterhaus: Good point. I actually was able to simplify it by simply using `const TagName = href ? 'a' : 'span';` I think this way makes most sense to me.

matuzalemsteles: Oh it looks much better now, good move with `TagName`.

matuzalemsteles:

packages/clay-label/src/ClayLabel.tsx
```diff
@@ -0,0 +1,57 @@
...
+			<span className="label-item label-item-expand">{children}</span>
+
+			{closeButtonProps && (
+				<span className="label-item label-item-after">
+					<button
+						{...closeButtonProps}
+						className="close"
+						type="button"
+					>
+						{closeButtonProps.children}
```
matuzalemsteles: hey @bryceosterhaus, since we have a `closeButtonProps` API I think it makes sense to use` ClayIcon` directly here to live up to its name and usage.

bryceosterhaus: Good call, I didn't think about that. Just updated

packages/clay-label/src/ClayLabel.tsx
```diff
@@ -46,7 +47,7 @@ const ClayLabel: React.FunctionComponent<Props> = ({
 						className="close"
 						type="button"
 					>
-						{closeButtonProps.children}
+						<Icon symbol="times" />
```
matuzalemsteles: As we offer the option to pass **spritemap** to `ClayIcon`, I think it makes sense to add this API to ClayLabel since we are depending on the Icon, if we go without it, we will force the developer to use only the context.

packages/clay-label/src/ClayLabel.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import * as React from 'react';
+import classNames from 'classnames';
+import Icon from '@clayui/icon';
+
+type DisplayType = 'secondary' | 'info' | 'warning' | 'danger' | 'success';
+
+interface Props
+	extends React.BaseHTMLAttributes<HTMLAnchorElement | HTMLSpanElement> {
+	closeButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
```
matuzalemsteles: I had forgotten about this, but wanted us to continue adding the API descriptions here.

packages/clay-label/src/ClayLabel.tsx
bryceosterhaus: Im a bit worried to go down this path of adding a `spritemap` prop to every component that we create that consumes icon. One primary reason it seems awkward is because you would naturally expect `ClayLabel` or any`Clay*` component to use the icons that come from clay-css. One thought is to automatically include the clay icons within `ClayIcon` itself, but thats sort of unclear because of the way people host and consume `clay-css` is different.

Any idea how we could address this? Do you agree its weird to pass spritemap everywhere? My hope was context would fix that problem, but that just means every React app using clay now requires ClayIconContext at the top.

packages/clay-label/src/ClayLabel.tsx
bryceosterhaus: oh good call, I had forgotten too. I'll send a new PR to update all the existing attributes that were merged and then I will update this PR and any others that I have open.

matuzalemsteles: Okay, thanks Bryce! I'll update the clay-link as well.

packages/clay-label/src/ClayLabel.tsx
matuzalemsteles: One of the problems is that if we do not pass the **spritemap** to non-React applications (e.g Taglibs), I would not be able to use `context` and the only way I see it at the moment would be using the `spritemap` API. Unfortunately I can not think of another way to resolve this outside of that scope.

In taglibs devs do not need to pass the spritemap because we already defined this in the call to the component, but we still need this API, I do not know if we can do something different, in the layers that we will deliver to the Portal we can wrap the components in a type of HOC that uses the context and these layers will have the spritemap API but not ClayLabel for example.

bryceosterhaus: yeah thats true about the taglib side. I guess its probably fine to just keep adding spritemap props for now, and then when we get to the other layer to be compatible liferay, we can re-visit it then.

Ill update this PR with spritemap.

bryceosterhaus:
updated

matuzalemsteles:
hey @bryceosterhaus can you take care of the errors in the CI and add some notes in the migration guide before merge?

bryceosterhaus:
updated and added docs

bryceosterhaus:
Any idea how to solve the CI issue where it can't find `@clayui/icon`?

wincent:
@bryceosterhaus: it could be that the `jest.config.js` needs something like this:

```
  moduleNameMapper: {
    '@clayui/(.+)$': '<rootDir>packages/$1/src',
  },
```

matuzalemsteles:
hey @bryceosterhaus and @wincent, I ended up getting these errors too, this is due to the typescript configurations that are trying to find the lib folder of the respective packages, as we do not have this build process in CI will cause these errors, I correct them in PR #1812, I will separate this fix and send it by another PR so that you can give it a rebase over it.

wincent:
@matuzalemsteles: I find that pretty puzzling. You definitely shouldn't have to do a build in order to run the tests. And the top-level tsconfig doesn't mention lib folders, except to exclude them, right?

matuzalemsteles:
Yes that is, the issue is that the typescript that is resolving the packages, as we have stated in the typescript configuration to fetch the packages within `"@clayui/*": ["./packages/clay-*"]` only , it will look for the `main` field in `package.json` which is pointing to the lib folder, it is a pity that the typescript settings currently do not allow you to configure it to resolve different main fields, this would be of great help we. I'm crawling this problem at https://github.com/Microsoft/TypeScript/issues/21423.

We can go back to the point of pointing directly to the `src` folder but we will have to add an `index.ts` file to export the default file.

wincent:
Thanks for the explanation. That might explain why I didn't run into this difficulty in another monorepo project with TypeScript: there, each package has a "main" of "lib/index.js", and every package has a "src/index.ts", with this tsconfig:

```
{
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "node",
    "allowJs": true,
    "noEmit": true,
    "strict": true,
    "isolatedModules": true,
    "esModuleInterop": true,
    "baseUrl": "packages",
    "paths": {
      "@my-project/*": ["*/src"]
    }
  },
  "files": ["types/global.d.ts"],
  "include": ["packages/*/src/**/*.ts"]
}
```

matuzalemsteles:

matuzalemsteles:
hey @bryceosterhaus you can take care of the rebase so we can merge.

PR #1779, opened 2019-04-03
Fixes #1772 - Create @clay/link

### LINKED ISSUE #1772: Create link component

matuzalemsteles:


migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -18,3 +32,60 @@ Noteable Changes
...
+<ClayLink
+	href="#link-styles"
+-	label="Default"
+-/>
++>
++	Default
++</ClayLink>
+```
+
+### API Changes
```
bryceosterhaus: This is a nice touch. 👍

packages/clay-link/package.json
```diff
@@ -0,0 +1,31 @@
...
+		"build": "babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "tsc --project ./tsconfig.declarations.json",
+		"prepublishOnly": "npm run build && npm run build:types",
+		"start": "webpack-dev-server --mode development",
+		"test": "jest --config ../../jest.config.js"
+	},
+	"keywords": ["clay", "react"],
+	"dependencies": {
+		"classnames": "^2.2.6"
+	},
```
bryceosterhaus: I updated the generator in this commit, https://github.com/liferay/clay/pull/1781/commits/7013a80bf26af2ae429cc4d807528e642e98cfe3, we need to make sure to include clay-css as a devDependency since we use it for the demo App.js

packages/clay-link/src/ClayLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import * as React from 'react';
+import classNames from 'classnames';
+
+type DisplayType = 'secondary';
+
+type ButtonDisplayType = 'link' | 'primary' | 'secondary';
+
+interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+	buttonDisplayType?: ButtonDisplayType;
```
bryceosterhaus: Hmm not sure about this. Seems weird to allow people to style a link like a button. I noticed on the clayui site we don't have this listed and on the lexicon site it even specifies that we need to make distinctions between links and buttons. I'm guessing you just copied this over from v2 Link?

packages/clay-link/package.json
matuzalemsteles: We have `clay-css` as` devDependencies` in the root folder, would not this resolve our issue?

bryceosterhaus: Oh I had never noticed that. we might be fine then.

packages/clay-link/src/ClayLink.tsx
bryceosterhaus: For cases like that since it is more of a one-off case, what do you think about just isolating that functionality to that component itself? So for that button, rather than use `ClayLink`, you would just manually add an anchor tag with the button classes. That way it is isolated to its use case and then we can keep our generic components more in sync with lexicon/clay docs.

matuzalemsteles: Yes I think it is better, I will remove this, I do not remember if we had other cases for this, but if it does appear we can think of other solutions.

bryceosterhaus:
LGTM

PR #1781, opened 2019-04-03
Fixes for issue #1774 - Create radio-group component

bryceosterhaus:


matuzalemsteles:

packages/clay-radio-group/src/ClayRadioGroup.tsx
```diff
@@ -0,0 +1,131 @@
...
+	className,
+	inline,
+	name,
+	onSelectedValueChange,
+	selectedValue,
+	...otherProps
+}) => {
+	return (
+		<div {...otherProps} className={classNames(className)}>
+			{children.map((child, i) => {
```
matuzalemsteles: It may be safer to use the `React.Children.map` API.

bryceosterhaus: I was looking at the react docs and it seemed like `React.Children.map` was to be used specifically when you want to pass the `this` instance to each child. Maybe I'm wrong though, did you have a different understanding of its purpose?

matuzalemsteles: I think the documentation passes a different feeling about its use, but we can pass a function and clone the element. I say it may be safer because it traverse the children doing some treatments, I think looking at the code looks better. https://github.com/facebook/react/blob/master/packages/react/src/ReactChildren.js#L311-L346

bryceosterhaus:
@matuzalemsteles updated

packages/clay-radio-group/src/ClayRadioGroup.tsx
wincent: I think the idea is that "children" is actually an opaque data structure so the `React.Children` utils isolate you from having to worry about whether you have an array, an array of arrays, a string etc. Funnily enough, I think the [Flow documentation on Children](https://flow.org/en/docs/react/children/) might actually be the best summary of the kinds of shapes that children can take.

matuzalemsteles: Oh thanks @wincent for the info.

bryceosterhaus:
@matuzalemsteles I didn't write up the 1-to-1 API changes because its a different component itself, Radio vs RadioGroup. I didn't want users to think it was a straight port.

matuzalemsteles:
Makes sense, I'll take care of updating this document after we have clayui.com to direct the dev to the component page to learn more about it and how to use it.

PR #1784, opened 2019-04-03
Fixes #1773 - Create ProgressBar component

### LINKED ISSUE #1773: Create ProgressBar component

bryceosterhaus:


matuzalemsteles:

packages/clay-progress-bar/demo/App.tsx
```diff
@@ -0,0 +1,50 @@
...
+				{'50%'}
+			</ClayProgressBar>
+
+			<ClayProgressBar feedback status="success" value={100}>
+				{'100% as feedback'}
+			</ClayProgressBar>
+
+			<ClayProgressBar
+				feedback={value === 100}
+				status={value === 100 ? 'success' : 'warning'}
```
matuzalemsteles: Why do not these rules stay within the component?

packages/clay-progress-bar/src/ClayProgressBar.tsx
```diff
@@ -0,0 +1,64 @@
...
+					aria-valuemax={100}
+					aria-valuemin={0}
+					aria-valuenow={value}
+					className="progress-bar"
+					role="progressbar"
+					style={{width: `${value}%`}}
+				/>
+			</div>
+
+			{addon && <div className="progress-group-addon">{addon}</div>}
```
matuzalemsteles: Why do not we set the value in `10%` format when **children** are not past? so we cover the case of a dev not having to pass that format to the **children**.

packages/clay-progress-bar/src/ClayProgressBar.tsx
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	value: number;
+}
+
+const ClayProgressBar: React.FunctionComponent<Props> = ({
+	children,
+	className,
+	feedback,
+	status,
+	value,
```
matuzalemsteles: We can give you a warning (the same as we did for ClayIcon) to ensure that are passing a value in the range of 0 to 100. What do you think?

packages/clay-progress-bar/demo/App.tsx
bryceosterhaus: We can add them to the component, but I wasn't sure if people would ever use "success" or "warning" purely for the styling and not the semantic.

packages/clay-progress-bar/src/ClayProgressBar.tsx
bryceosterhaus: great idea!

packages/clay-progress-bar/src/ClayProgressBar.tsx
bryceosterhaus: 👍

packages/clay-progress-bar/demo/App.tsx
matuzalemsteles: Good question, I have not come across a case yet of using ClayProgressBar.

bryceosterhaus:
I went ahead and made it a bit smarter.

- I removed `status` and now show "success" if the value is 100 and added a `warn` prop so that the user can manually signify an error.
- I defaulted to wrapping the content in `feedback`. I think this is better as an opt out rather than an opt in.
- Defaulted the content to be the value + '%'
- Added dev warning for a value less than 0 or greater than 100

migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -88,4 +88,28 @@ To get to the behavior of having a ClayLink with image, use the composition with
...
+ClayLink has become simpler with v3, defaulting to wrap content with `progress-group-feedback` class and having the display status be naturally inferred from the `value`.
+
+If you want to use an icon for the progress value, you will need to manually compose it.
+
+```diff
+<ClayProgressBar
+-	spritemap: {'/images/icons/icons.svg'}
+-	status: {'warning'}
++	warn
+	value: {70}
```
matuzalemsteles: I think this is wrong
```suggestion
	value={70}
```

migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -88,4 +88,28 @@ To get to the behavior of having a ClayLink with image, use the composition with
 -/>
 +	<ClayImage src="image.jpg" alt="my image" />
 +</ClayLink>
-```
\ No newline at end of file
+```
+
+## ClayProgressBar
+
+ClayLink has become simpler with v3, defaulting to wrap content with `progress-group-feedback` class and having the display status be naturally inferred from the `value`.
```
matuzalemsteles:
```suggestion
ClayProgressBar has become simpler with v3, defaulting to wrap content with `progress-group-feedback` class and having the display status be naturally inferred from the `value`.
```

bryceosterhaus:
updated

packages/clay-progress-bar/src/ClayProgressBar.tsx
```diff
@@ -0,0 +1,78 @@
...
+	className,
+	feedback = true,
+	value = 0,
+	warn,
+	...otherProps
+}) => {
+	warning(
+		value >= 0 || value <= 100,
+		'ClayProgressBar requires `value` to be in the range of 0 to 100'
+	);
```
wincent: Instead of just warning, we should probably actually clamp the value, otherwise we would end up rendering with `width: 1000%` and such.

```typescript
export default function clamp(
  value: number,
  minimum: number,
  maximum: number,
): number {
  if (minimum <= maximum) {
    return Math.min(Math.max(value, minimum), maximum);
  } else {
    return NaN;
  }
}
```

bryceosterhaus:
@wincent updated

matuzalemsteles:

bryceosterhaus:
@matuzalemsteles updated

packages/clay-progress-bar/src/ClayProgressBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+	value = 0,
+	warn,
+	...otherProps
+}) => {
+	warning(
+		value >= 0 && value <= 100,
+		`ClayProgressBar requires \`value\` to be in the range of 0 to 100, it received ${value}.`
+	);
+
+	value = Math.min(Math.max(value, 0), 100);
```
matuzalemsteles: This is just a micro-optimization 🙂 But we can use `useMemo`.

```suggestion
	value = useMemo(() => Math.min(Math.max(value, 0), 100), [value]);
```

wincent: It's probably a dangerous habit to get into to apply microoptimizations "just because". In this case, the `useMemo` version is harder to read, and it seems unlikely that it would deliver any measurable performance benefit.

See [this benchmark](https://jsperf.com/clamp-micro/1), for example:

Math_min___Math_max_·_jsPerf

Clearly, it's so fast that the difference is just statistical noise.

matuzalemsteles: hey @wincent, Thanks for bringing this test, although it is relative and can vary... I agree that even if the use of `useMemo` comes to bring some benefit it will be something minimal or symbolic... (in this case)

bryceosterhaus:
Gonna merge, I updated migration guide to reflect the changes.

PR #1792, opened 2019-04-04
Fixes #1767 - Allow only items from autocomplete and emit event on add error

### LINKED ISSUE #1767: Multiselect: Add support for only accept autocomplete items insertion (don't allow input items)

carloslancha:


matuzalemsteles:

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -362,13 +360,19 @@ class ClayMultiSelect extends ClayComponent {
 				};
 			})[0];
 
-		item = filteredItem || item;
+		if (this.allowOnlyItemsFromAutocomplete) {
```
matuzalemsteles: Is this condition that guarantees that an item will not be created from the input? When I try to create a label through the comma it deletes all the content, I think I would have to keep the text. Probably because of this https://github.com/liferay/clay/blob/2.x-develop/packages/clay-multi-select/src/ClayMultiSelect.js#L115

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -362,13 +360,19 @@ class ClayMultiSelect extends ClayComponent {
 				};
 			})[0];
 
-		item = filteredItem || item;
```
matuzalemsteles: hey @carloslancha I was thinking today, it would be interesting to add some comments which use case the `filteredItem` covers, I took a while to understand this, probably someone else will also take a while. Just to leave the reasons why you added this.

carloslancha: I think now is clearer. You just need to read JSDOCS form creatable param

carloslancha:
Hey @matuzalemsteles take a look at this please. @jbalsas will be releasing a new clay version on monday. Thanks! ;)

matuzalemsteles:

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -109,10 +106,14 @@ class ClayMultiSelect extends ClayComponent {
...
 		if (char === ',' || words.length > 1) {
 			words.forEach(word => {
-				this._addLabelItem(this._createItemObject(word, word));
-			});
+				let added = this._addLabelItem(
+					this._createItemObject(word, word)
+				);
 
-			inputValue = '';
+				if (added) {
```
matuzalemsteles: It seems to have some pretty strange cases when I type 'qwe' and adds the item, then I type again and add a comma and type a new one after the comma it will add but the values ​​will still remain there:

matuzalemsteles: We must make sure to delete the commas to avoid problems, like adding an existing value, adding the comma and then pressing enter.

packages/clay-multi-select/demos/index.html
```diff
@@ -171,9 +178,19 @@ <h2>With only data remote</h2>
 			helpText: 'You can use a comma to enter tags',
 			label: 'Tags',
 			spritemap,
-			id: 'uno',
 		}, '#disabled-autocomplete-block');
 
+		// Allow only items from autocomplete
+		new metal.ClayMultiSelect({
+			allowOnlyItemsFromAutocomplete: true,
```
matuzalemsteles:
```suggestion
			creatable: false,
```

carloslancha: Yep, forgot that

packages/clay-multi-select/src/ClayMultiSelect.js
carloslancha: Do you think we can iterate over this? Or must be fixed on this pr?

matuzalemsteles: I think it's a bit critical to iterate, I think it's best to fix this in PR.

matuzalemsteles:

matuzalemsteles:
Resent #1853

PR #1811, opened 2019-04-05
Fixes #1809 - Create @clayui/image component

### LINKED ISSUE #1809: Create clay-image component

matuzalemsteles:


packages/clay-image/src/ClayImage.tsx
```diff
@@ -0,0 +1,21 @@
...
+import * as React from 'react';
+import classNames from 'classnames';
+
+const ClayImage: React.FunctionComponent<
+	React.ImgHTMLAttributes<HTMLImageElement>
+> = ({className, ...otherProps}) => (
+	<img
+		{...otherProps}
+		className={classNames(className, {
+			'img-fluid': !className,
```
matuzalemsteles: hey @bryceosterhaus I did this thinking of other use cases that do not use `img-fluid`, like the images for the sticker. Leave your thoughts 🙂

bryceosterhaus: Maybe instead of checking className, we just add a boolean for fluid? I'm imagining where someone wants to add a custom class so that they can target a style for their app.

bryceosterhaus: you know, do you think this component is actually really necessary? Seems a bit overkill since there are no clay specific classes needed for images.

matuzalemsteles: > Maybe instead of checking className, we just add a boolean for fluid? I'm imagining where someone wants to add a custom class so that they can target a style for their app.

I'm not sure, I wanted to leave default by not needing an API... but maybe make sense. 😅

> you know, do you think this component is actually really necessary? Seems a bit overkill since there are no clay specific classes needed for images.

I thought a little bit about it, maybe it does not make sense now but we're using image in some components so I'm going through the idea of ​​being able to concentrate the markup of images only here, if we have more use cases in the future then we will change in one place and everyone will receive the modifications.

bryceosterhaus: > maybe it does not make sense now but we're using image in some components so I'm going through the idea of ​​being able to concentrate the markup of images only here

I think this falls into the category of a ["premature abstraction"](http://wiki.c2.com/?PrematureAbstraction). If we continued with this idea of wanting to start abstracting early for cases in the future, we could make the same argument to start doing this for every html element, but I think you'd agree that would be a bit far fetched. Plus just from a practical side, it'll be very easy to forget to import `@clayui/image` when you can just use `` in its place. I agree in that this may reduce code duplication in the future if for some reason we start adding more clay specific stuff to , but since we aren't in that place yet I would say lets table it. Even if there was something like `clay-img` I would rather just use that class everywhere since there isn't any configuration we can provide via props. That make sense at all? Sometimes I feel like I can't string a coherent sentence together 😂

matuzalemsteles: Yeah indeed 🙂, initially it is much easier to use ``... I understand your concern I will close PR and if in the near future we need, we will come back to talk about this component. Thanks @bryceosterhaus

bryceosterhaus: Sounds good! And by the way, totally feel free to push back on any components or functionality I create as well, I think the more conflicting opinions and discussion we have will only make for a better library!

matuzalemsteles: yeah I think the same way, I'm happy about it 🙂. But we agree with the current components 🚀

PR #1812, opened 2019-04-05
Fixes #1808 - Create @clayui/sticker component

### LINKED ISSUE #1808: Create clay-sticker component

matuzalemsteles:


packages/clay-checkbox/package.json
```diff
@@ -17,7 +17,7 @@
 	],
 	"scripts": {
 		"build": "babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
-		"build:types": "tsc --project ./tsconfig.declarations.json",
+		"build:types": "tsc --project ./tsconfig.declarations.json --emitDeclarationOnly --allowJs false",
```
bryceosterhaus: I think we can just add `allowJs: false` to `tsconfig.eclarations.json` at the root level of the project. We shouldn't need to add this to each build script.

packages/clay-sticker/src/ClaySticker.tsx
```diff
@@ -0,0 +1,99 @@
...
+			[`sticker-${size}`]: size,
+		})}
+	>
+		<span
+			className={classNames({
+				'inline-item': icon,
+				'sticker-overlay': !icon,
+			})}
+		>
+			{icon ? <ClayIcon spritemap={spritemap} symbol={icon} /> : children}
```
bryceosterhaus: How do you feel about de-coupling Icon from sticker and make is so that it'll be composed together? My thinking is that for cases where the Icon is the primary content, let the user compose that, however if the icon is integral to the functionality(like the 'X' on ClayLabel) then we can make it a dependency.

So things generic items like Button, Badge, Sticker, etc would all be independent of Icon and if a user wants to use Icon inside of that component, they can compose it together.

packages/clay-checkbox/package.json
matuzalemsteles: I had forgotten that. Much better!

packages/clay-sticker/src/ClaySticker.tsx
matuzalemsteles: Hey @bryceosterhaus I thought a lot about this before submitting, one of the deadlock it has been was that I need to replace the `sticker-overlay` class with `inline-item` I could add an API to create this condition but I think it would be unnecessary , because its use case would only be for `Icon` (I'm still creating a dependency of Icon within the component) and maybe in that way it can create a degree of dependency to make the composition with `Icon` work fine (We can decrease this adding in the documentation as a guide to good practices...).

But I really like to remove the dependencies and make it more flexible, but I still do not know how we could dribble this case here.

bryceosterhaus: What if we checked the contents of `children` and conditionally used `inline-item` if it was an svg?

matuzalemsteles: We would have to go through the complete Fiber tree to see if this component exists, I think it's not worth it and it's very costly depending on the size of the application. We could check the **props** of **children** if there is a `symbol` but I find this check too weak.

I was thinking of checking the `type` of the **children** to know the name of the component but the` type.name` is according to the name of the function and that name may disappear when minified and receive optimizations for production.

bryceosterhaus: I talked with @pat270 about it and he decided that the behavior is weird on the css side to require `inline-item` for icons in a sticker. Take a look at https://github.com/liferay/clay/issues/1825. I think we can go ahead with the composition route with this new issue resolving the `inline-item` issue.

matuzalemsteles: Perfect, I will update with these changes.

PR #1816, opened 2019-04-05
Fixes #1709 - Migrate ClayAlert to React

### LINKED ISSUE #1709: Create clay-alert component for v3.0.0
Need to implement clay-alert in React.

bryceosterhaus:


matuzalemsteles:

packages/clay-alert/src/ClayAlert.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * Determines the style of the alert.
+	 */
+	displayType?: DisplayType;
+
+	/**
+	 * Flag to indicate the stripe variant.
+	 */
+	stripe?: boolean;
```
matuzalemsteles: I'm wondering if we'd better keep the API with that name or `type` (`variant`) and set the `typo` to stripe (`type="stripe"`), it seems more obvious that the stripe is a variation of just defining stripe in the component, I would probably fetch it by variant when I was using this component, even if only have one variant

bryceosterhaus:
@matuzalemsteles good call on the naming. I went ahead and switched it to `variant`. I also added the `autoClose` functionality.

I wasn't quite sure on the animations though, where were you seeing those at? I didnt see it in the v2.0 component.

matuzalemsteles:

packages/clay-alert/src/timerHook.tsx
```diff
@@ -0,0 +1,39 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+
+export default (autoClose?: boolean | number, onClose = () => {}) => {
```
matuzalemsteles: I think we need to improve on the logic here of timer administration, some things in the demo do not work so well.

bryceosterhaus:
@matuzalemsteles yeah something is actually off here with this implementation. I think I need to read up more on hooks, the useEffect hook is working differently than I'd expect.

bryceosterhaus:
@matuzalemsteles  Updated

matuzalemsteles:

packages/clay-alert/src/ClayAlert.tsx
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import classNames from 'classnames';
+import Icon from '@clayui/icon';
+
+const useAutoClose = (autoClose?: boolean | number, onClose = () => {}) => {
+	const timer = React.useRef<number | undefined>(undefined);
```
matuzalemsteles: hey @bryceosterhaus just out of curiosity why are we storing the timer inside a `useRef`?

packages/clay-alert/src/ClayAlert.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+		startTimer = () => {
+			timer.current = setTimeout(
+				onClose,
+				autoClose === true ? 8000 : autoClose
+			);
+		};
+	}
+
+	React.useEffect(() => {
```
matuzalemsteles: I think we can still call `cancelTimer` on the return of `useEffect` to clear the timer, when the component is dismounting and if it has not finished the time.

packages/clay-alert/src/ClayAlert.tsx
```diff
@@ -0,0 +1,177 @@
...
+				clearTimeout(timer.current);
+
+				timer.current = undefined;
+			}
+		};
+
+		startTimer = () => {
+			timer.current = setTimeout(
+				onClose,
+				autoClose === true ? 8000 : autoClose
```
matuzalemsteles: By default we should leave the timeout in 5s and we should freeze the timer when you have a `hover` on alert.

packages/clay-alert/src/ClayAlert.tsx
bryceosterhaus: This is sort of a paradigm shift that comes from classes to hooks. Because call components are just functions now, we need a way to keep a value around. I got the idea from other examples I saw that used `setTimeout`. I then read this in the docs and helped me make sense of it. https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

packages/clay-alert/src/ClayAlert.tsx
bryceosterhaus: Yeah I am doing that inside of the `if (autoClose && onClose)` condition. Are you thinking we should move it outside of that condition?

packages/clay-alert/src/ClayAlert.tsx
bryceosterhaus: I had gotten the default value of 8 seconds from [here](https://github.com/liferay/clay/blob/2.x-develop/packages/clay-alert/src/ClayToast.soy#L16).

And I went ahead with canceling and restarting the timer on hover since it seemed weird that if a user hovers with only 200ms remaining, it would close super quick. I thought it would be a little more user friendly to clear and restart on hover. What do you think?

bryceosterhaus:
Didnt see your comment about animations till now. Ill go ahead and update with this

> The animations are only in cases when we do not destroy the component and we add a show class. It is for the case of the destroyOnHide API when it is false.

bryceosterhaus:
@matuzalemsteles Actually, in regards to the animation, I am thinking we leave that up to the consumer. Since the parent component will determine when its rendered, it should also likely be up to the parent to decide how it gets animated. If we control the animation within the alert component itself, it means the user would always render it and toggle a "visible" prop rather than conditionally render it. In soy it was a bit different because every component was its own single widget which made sense for it to handle its visibility by itself. When we bring this into portal, we will likely handle this interaction from the portal layer for rendering our react taglibs.

packages/clay-alert/src/ClayAlert.tsx
matuzalemsteles: Hm, thanks for the information, it seems clearer now the use of `useRef` I had another thought about it.

packages/clay-alert/src/ClayAlert.tsx
matuzalemsteles: Sorry Bryce I had confused about it, that's correct. 😅

packages/clay-alert/src/ClayAlert.tsx
matuzalemsteles: No problems over the 8 seconds, let's keep do not remember why we changed to that value.

The bad thing is that if you have 200ms missing and it restarts it will stay for a lot longer, I do not know which would be the best case for that, we have pros and cons in both use cases.

@victorvalle can you take a look at this?

matuzalemsteles:
hey @bryceosterhaus this is a good question regarding the animations my concern about this is that we can be leaving a lot for the developer to customize the compositions to get the expected result and make it a bit frustrating to use the component.

packages/clay-alert/src/ClayAlert.tsx
victorvalle: We choose 8s as a good value reading other guides.
I don't understand why it restarts sorry. When you are hovering it the timer should freeze.

wincent: I was going to say the "freeze" thing too. As a user, I expect the thing to stick around indefinitely if I am hovering over it, but I *don't* expect the timer to reset. ie. if I start hovering with 2 seconds left on the timer, and I move my mouse away 30 seconds later, I expect the toast to disappear 2 seconds after my mouse leaves.

bryceosterhaus: 👍 ill go ahead and make it a pause rather than reset.

bryceosterhaus:
Added the pause functionality. Just waiting to reach a conclusion on animations from the other ticket, https://github.com/liferay/clay/pull/1842#discussion_r274654886

bryceosterhaus:
@matuzalemsteles updated and move demo to storybook. I also talked to @pat270 and there are no default animations for alert right now. Portal has specific animations for them, but we should leave that on the portal side. Let me know if you see any issues with this component.

matuzalemsteles:

bryceosterhaus:
@matuzalemsteles good catch on the dependency. Just removed it.

matuzalemsteles:
Thanks @bryceosterhaus. Merged!

Issue #1744, opened 2019-03-29
Typescript does not solve the modules correctly with dependencies in monorepo


matuzalemsteles:
This bug sounds pretty familiar with what we had when we started building the components of Clay, so I'm going to leave a checklist of things that we should look out for to create an integration and not have any problems ahead. (I wait)

- [ ] The resolved module should not be duplicated, we will have more cases ahead, when we have cyclic dependency in some cases where the `a` package depends on the `b` and `c` package and the `c` package depends on the `b` package, we have to be careful that there is no duplication of code.
- [ ] The `yarn start` command that is running webpack should resolve the packages correctly and point to the `tsx` files. (Probably we'll have to make some modifications to make this happen, typescript is trying to solve the package.json `main` by picking `lib/ClayXXX`...)
- [ ] We must ensure that the `main` attribute of `package.json` is pointing to files compiled in the `lib` folder. (One solution is to modify the attribute to point to `tsx` files in development environment...)
- [ ] We have to make sure that the tests can resolve the dependencies correctly for the respective `tsx` files of each package.

wincent:
Not sure if this is helpful, but when I rolled out Typescript in a monorepo I had to [use the moduleNameMapper config option](https://github.com/wincent/js/commit/fe2d7318dc94354306331eb9f5b0d191a831fd9a) to get Jest to be able to resolve the packages, and the analog here would be something like:

```
    "moduleNameMapper": {
      "clay/(.+)$": "<rootDir>packages/$1/src"
    },
```

Not sure whether this is useful either, but I also note that here we have this in our "tsconfig.json":

```
	"include": ["packages"],
```

but [I used](https://github.com/wincent/js/blob/8ab4976de2fa895ed4dc97a6ab885f6688bca690/tsconfig.json#L17):

```
  "include": ["packages/*/src/**/*.ts"]
```

matuzalemsteles:
hey @wincent thanks for this, I'll test. Some of the problems we have is that the Typescript loader resolves only to the files declared in compiled `main` (See https://github.com/Microsoft/TypeScript/issues/21423). This testing paths with:

```
"paths": {
	"@clayui/*": ["node_modules/@clayui/*"],
}
```

A hack solution would be to change the `main` path to point to the `src` files in the development environment. (As long as we do not have the resolution of https://github.com/Microsoft/TypeScript/issues/21423).

Another solution is to change the path to point to src and add an `index.ts` to export the main files.

```
"paths": {
[...]

Issue #1170, opened 2018-09-04
Improve the clayui.com deploy process


matuzalemsteles:
### Disclaimer
We are currently deploying deploy when it is in `master` or when we need to visualize something, we can improve this by automating our process and defining a detailed workflow.

### Proposal

I thought the following flow, but we can change.

`PR -> Deploy to preview [1] -> Merged in develop branch [2] -> Merged in master branch [3]`

1. Once a PR is submitted with changes to clayui.com the bot will create a preview of the website from that PR commit. This will generate a specific URL for this PR. (Something similar to what reactjs.org does)
2. The `dev` environment for clayui.com, as soon as something is merged into the branch develop a deploy for that environment (` dev.clayui.com? `) Is done. It will help others know what's coming and see changes coming soon in the next release, this also helps us identify problems sooner.
3. Our production environment, nothing new here but can be automated with a bot.

All this workflow we can define with a bot without having to do something manual. It starts ah take away sensation that clayui.com is outdated, with the help also of #1111 and other tasks to be automated....

> **OBS**: I do not know if it would be possible to do scenario 1. with wedeploy... Maybe we can opt for [netlify](https://www.netlify.com/) to do this part of the process, something reactjs.org uses for this step.

matuzalemsteles:
cc @jbalsas @marcoscv @pat270

jbalsas:
Hey @matuzalemsteles, `WeDeploy` introduced [environments](https://wedeploy.com/blog/introducing-wedeploy-environments/). I think we could have
- `dev`
- `prod`
- ideally a new environment per PR, but if not, one called `pr`

@zenorocha, is it possible to create environments programatically so we could automate this? That would be amazing! :)

jbalsas:
I think I can answer that myself, @zenorocha:

> There is currently no way to create new environments from the CLI. To do this, you must go use the Console.

jbalsas:
@matuzalemsteles, let's explore the WeDeploy environments idea... for automation on PRs, maybe we could:
- Create a `pr` environment with a different setup
- Configure gatsby in an inteligent way, so rather than going to `clay-dev.wedeploy.io`, every PR gets sent to `clay-dev/wedeploy.io/pr-123
- Get the bot (git-hook?) to do the automatic push to the `dev` and `prod` branches on merges on `develop` and `master`.

How does that sound?

matuzalemsteles:
hey @jbalsas Awesome!! I think we can explore that.

[...]

PR #1838, opened 2019-04-10
Add Storybook for displaying demos and UI-kit

bryceosterhaus:
Hey all, thought I would send a POC for this. The idea is to use storybook for demos and it would actually allow us to build a static site for documentation.

Here is a demo of both badge and button on storybook. Checkout the ClayButton default story and play around with the "knobs" tab. Thought it was pretty cool that it'd allow us to interactively play with our components. https://storybook-static-apenpbqdiu.now.sh/?path=/story/claybadge--default

If we do decide this is the route we want to go, I will update the PR to add stories for all of our existing components and I will remove our `demo` directories.


matuzalemsteles:

matuzalemsteles:
hey @bryceosterhaus the experience is very good for development. By me we can remove the demos and just keep the Storybook. This will add much more to our documentation.

It's a perfect replacement for our current demos, since we have the a11y addon and also the demos get much more organized and easy to navigate. Now would it be possible to run a11y tests on CI?

bryceosterhaus:
I also investigated using netlify to auto deploy storybook builds. Seems pretty cool. We would be able to automatically have static sites generated for every PR, which would allow quick user testing. We can also automate it so that each branch or tag could have a build and then set a custom domain.

matuzalemsteles:
That sounds great, it's something we want to do with clayui.com (#1170).

bryceosterhaus:
If no one has any issues with using storybook, I'm gonna go ahead and replace all of our existing demos with this. We can decide on the auto-deployment portion later.

matuzalemsteles:
hey @bryceosterhaus, I think you can go with it, I see no reason why we will not add the Storybook.

bryceosterhaus:
@matuzalemsteles updated to add all demos. I also had to modify some file structure and config so that the modules properly built. See https://github.com/liferay/clay/issues/1744 for reference

matuzalemsteles:

packages/clay-sticker/package.json
```diff
@@ -8,27 +8,34 @@
...
+	"keywords": [
+		"clay",
+		"react"
+	],
 	"dependencies": {
 		"classnames": "^2.2.6"
 	},
 	"devDependencies": {
-		"@clayui/icon": "^3.0.0"
+		"@clayui/icon": "v3.0.0"
```
matuzalemsteles: Should this be `v3.0.0` anyway?

bryceosterhaus: good catch, just updated.

PR #1842, opened 2019-04-11
Fixes #1829 - Migrate ClayModal to React

### LINKED ISSUE #1829: Migrate ClayModal to React

matuzalemsteles:


packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+	/**
+	 * Url to place an iframe in the body of the modal.
+	 */
+	url?: string;
+
+	/**
+	 * Flag to indicate if the modal is open or closed.
+	 */
+	visible: boolean;
```
bryceosterhaus: By having this component manage its own visibility, we are now required to always render its contents right? I'm wondering if its best to always let the parent component manage the child's visibility. This is larger infra question as a whole though and one I am not quite sure I know the "right" answer. Normally I would always default to having parent components manage visibility and rendering, but idk if that is right.

Anyone else have any thoughts on this? @wincent @jbalsas @diegonvs

packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+	const [visibleClassShow, setVisibleClassShow] = useState<boolean>(false);
+
+	const modalDialogElementRef = useRef<HTMLDivElement | null>(null);
+
+	/**
+	 * A hook that takes care of controlling click, keyup and keydown events
+	 * respectively close the modal after a click on the overlay, close the
+	 * modal by pressing the ESC key and control the focus within the Modal.
+	 */
+	useModal(modalDialogElementRef, visible, onClose);
```
bryceosterhaus: Since this hook is concerned with interactions, could we name it `useUserInteractions` or something like that? `useModal` is a big ambiguous to me.

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import classNames from 'classnames';
+import React, {FunctionComponent, HTMLAttributes} from 'react';
+import warning from 'warning';
+
+export interface FooterContainerProps extends HTMLAttributes<HTMLDivElement> {
+	position?: number;
+}
+
+const ButtonWrap: FunctionComponent<HTMLAttributes<HTMLDivElement>> = ({
```
bryceosterhaus: Should this exist in `@clayui/button` instead? Or is it only ever used in Modal?

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: Maybe this name can be a bit confusing, but who really manages the visibility of the component is Father, it defines through the click of its button to open or close the modal through the callback `onClose`, we need to have the modal content "visible" in order to create the animations in this case, such as adding the class `show` when visible is true.

I like this style to set the visibility of the component, I think we more freedom to do things within the component, such as animation.

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: Yeah, I did not have the creativity to name that hook 😅. `useUserInteractions` sounds good.

packages/clay-modal/src/Footer.tsx
matuzalemsteles: I'm not sure if we have use cases outside Modal for the `btn-group-item`, I think it makes sense to keep it there. I'll change that. Maybe we can do the same thing that was done on the radio.

packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+	/**
+	 * The path to the SVG spritemap file containing the icons.
+	 */
+	spritemap?: string;
+
+	/**
+	 * Status messages.
+	 */
+	status?: Status;
```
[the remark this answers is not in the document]
matuzalemsteles: I'm not sure if the best would be `displayType` for such, this is a slightly different case, `status` seems clearer about `displayType` in that case. But I think it does feel like going with `displayType` because the properties that are passed to it is more consistent with `displayType`.

packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+		};
+	}, [elToMountTo]);
+
+	return createPortal(children, portalRef.current);
+};
+
+const delay = (fn: Function) => {
+	const callbackId = setTimeout(() => {
+		fn();
+		clearTimeout(callbackId);
```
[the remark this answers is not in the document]
matuzalemsteles: yes, I really do not think we need to worry about clearing the timeout. I'll remove that. Thanks!

packages/clay-modal/src/ClayModal.tsx
bryceosterhaus: @jbalsas @wincent any thoughts on this? Just making sure this didn't get lost in the mix of GH notifications.

migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -94,6 +94,59 @@ To get to the behavior of having a ClayLink with image, use the composition with
...
++	onClose={() => setVisible(false)}
++	renderFooter={
++		<>
++			<ClayModal.FooterContainer>
++				<ClayButton>{'Cancel'}</ClayButton>
++			</ClayModal.FooterContainer>
++			<ClayModal.FooterContainer>
++				<ClayButton>{'Save'}</ClayButton>
++			</ClayModal.FooterContainer>
++		</>
```
wincent: I don't think this is a very ergonomic API. We should be able to grab the children directly and wrap them inside the component without requiring the user to do it.

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+	warning(
+		count <= 2,
+		`ClayModal -> props \`renderFooter\` has received more ${count} elements, only 2 is allowed, works fine use only 2 elements of <FooterContainer />`
+	);
+
+	return (
+		<div className="modal-footer">
+			{React.Children.map(childrens, (child: any, i) =>
+				React.cloneElement(child, {position: i})
+			)}
```
wincent: Related to my point early on, this is where we would wrap in a `FooterContainer` for the user instead of having the user do it themselves.

packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+	useEffect(() => {
+		elToMountTo.appendChild(portalRef.current);
+		return () => {
+			elToMountTo.removeChild(portalRef.current);
+		};
+	}, [elToMountTo]);
+
+	return createPortal(children, portalRef.current);
+};
```
wincent: In theory there should be only one modal on the screen at a time, but the code here doesn't seem to do anything to enforce that. You can instantiate multiple modals, and each will append a new div and create a portal.

Yet further down, we are adding and removing classes to/from the `document.body`, which is effectively a singleton. So we have an impedance mismatch, with some parts of the code behaving like a multiple instantiation is ok, and others behaving like they are operating inside a singleton. We should probably look at fixing that.

packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+	const [visibleClassShow, setVisibleClassShow] = useState<boolean>(false);
+
+	const modalDialogElementRef = useRef<HTMLDivElement | null>(null);
+
+	/**
+	 * A hook that takes care of controlling click, keyup and keydown events
+	 * respectively close the modal after a click on the overlay, close the
+	 * modal by pressing the ESC key and control the focus within the Modal.
+	 */
```
wincent: This description seems to belong with in `./Hooks.ts`, and not here. We document the implementation, not the usage site. Also, if it is only used here, it should probably be defined here (legit reasons to have something live in a separate module: you want to expose it for testing; you want to re-use it).

packages/clay-modal/src/ClayModal.tsx
wincent: Thanks @bryceosterhaus, I missed the original mention. I tend to agree with you that the parent should be responsible for choosing when to render. It is a bit strange to have something with `visible={false}` render something into the DOM. I don't think the modal should receive a prop from the outside; rather I would expect it to be: parent renders -> on mount, modal animates in -> on close, modal animates out, then calls parent -> parent unmounts dialog. Modal will obviously need internal state to manage the transitions: mounting/animating -> displayed -> closing/animating.

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+		</div>
+	);
+};
```
wincent: I get the feeling like we might be trying to do a bit too much here, when all we really want to confirm is that we have 2 or fewer children.

packages/clay-modal/src/ClayModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+interface Props extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Callback called to close the modal.
+	 */
+	onClose: () => void;
+
+	/**
+	 * Renders the footer with the FooterContainer component utilities.
+	 */
+	renderFooter?: React.ReactElement;
```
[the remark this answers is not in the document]
matuzalemsteles: I think it makes more sense to accept function when we need to pass some extra information so it can be rendered, I do not see any other benefit besides that but I can be totally mistaken.

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+		(children.type as any) === React.Fragment ||
+			typeof (children.type as any) === 'function',
+		'ClayModal -> Wrap the `<FooterContainer />` elements inside a Fragment to work as expected.'
+	);
+
+	const count = React.Children.count(childrens);
+
+	warning(
+		count <= 2,
+		`ClayModal -> props \`renderFooter\` has received more ${count} elements, only 2 is allowed, works fine use only 2 elements of <FooterContainer />`
```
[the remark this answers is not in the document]
matuzalemsteles: Yes, I'm going to remove that. I think we can leave as an error, although it is not a serious mistake, but if it is not followed you will probably have problems visually.

packages/clay-modal/src/__tests__/ClayModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+			return element;
+		});
+	});
+
+	afterEach(() => {
+		// @ts-ignore
+		ReactDOM.createPortal.mockClear();
+		// @ts-ignore
+		// eslint-disable-next-line no-console
+		console.error.mockClear();
```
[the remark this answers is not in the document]
matuzalemsteles: Good idea.

migrate-the-clay-components-from-v2-to-v3.md
```diff
@@ -94,6 +94,59 @@ To get to the behavior of having a ClayLink with image, use the composition with
 
 -   Added ability to utilize context for passing spritemap down instead of having to pass the prop everywhere.
 
+## ClayModal
+
+Of more freedom to customize `Footer` but still requiring some rules to make everything work well, always use the `<ClayModal.FooterContainer />` component to align the left and right buttons.
```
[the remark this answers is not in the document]
matuzalemsteles: What I wanted to say would be: The dev might have more control over Footer (but not so much...), previously the footer was defined by an API. That's why more freedom. I will improve this.

migrate-the-clay-components-from-v2-to-v3.md
matuzalemsteles: I also think. The issue that we have a limitation on markup, we need to define two elements that define the left and right side of the footer, in this case the two `FooterContainer` elements here (Abstracts this information, but I'm not sure if it was the right choice or if I need give more context in our documentation...), so I can add more than one ClayButton inside those containers. Even if I can eliminate the use of `FooterContainer` I would not have been able to identify this.

Maybe I could do something like:
```jsx
<>
  <>
    <ClayButton>{'Save'}</ClayButton>
    <ClayButton>{'Cancel'}</ClayButton>
  </>
  <>
    <ClayButton>{'...'}</ClayButton>
    <ClayButton>{'...'}</ClayButton>
  </>
</>
```

We can use the Fragment to differentiate the sides, but I do not know if this would be a recommended standard, I did not get to see such a case.
> I have not tested this yet

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+interface FooterProps {
+	children: React.ReactElement<FunctionComponent<FooterContainerProps>>;
+}
+
+/**
+ * Handles the childrens of the footer to format the markup and add
+ * the necessary classNames.
+ */
+const Footer: FunctionComponent<FooterProps> = ({children}) => {
+	let childrens = children;
```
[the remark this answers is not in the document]
matuzalemsteles: I think it would not be necessary to unwrap all Fragments, the first case is necessary just not to break the markup. Also not sure if anyone would use more than one fragment here.

migrate-the-clay-components-from-v2-to-v3.md
wincent: The thing about getting people to group using nested fragments is that is very implicit and probably not the friendliest API.

At `$previousJob` we had a `<LeftRight />` component for this, that expected two children. Was a bit nicer to use than another alternative, which was separate props for left and right:

```javascript
<Thing
   primary={<OkButton />}
   secondary={<CancelButton />}
/>

<Thing>
   <LeftRight>
     <CancelButton />
     <OkButton />
   </LeftRight>
</Thing>
```

Right now I can't remember whether it automatically flipped for RTL locales. It probably did.

packages/clay-modal/src/ClayModal.tsx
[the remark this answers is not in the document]
wincent: In [your other comment](https://github.com/liferay/clay/pull/1842#discussion_r278552133) you said:

> The dev might have more control over Footer (but not so much...), previously the footer was defined by an API. That's why more freedom. I will improve this.

Which makes it sound like you want more flexibility. There's obviously a trade-off here between control (doing only what is narrowly and explicitly defined in Lexicon) and flexibility (being able to go beyond what's explicitly specified). We should just try to be clear about which end of that spectrum we want to stand at.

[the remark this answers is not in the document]
matuzalemsteles: I do not quite understand what my comment has to do with this thread 😅 (maybe I'm understanding this wrong) but yes you're right about that, we should be careful with the way we want to position, when I say this is the in case you had no control before, now we are with a little freedom for you to set your buttons components or turn into something else ... without needing the dependency of a defined API to do such but still continue on the surveillance of the specification (which is the result of the rules).

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: I'm not sure if we would have a case of opening two modal at the same time, otherwise, we would have no problem with manipulating the `document.body`.

I'm going to improve this from the `backdrop` Portal, but I do not know if we can control Modal content as a singleton. Maybe dev will have to handle this in your application. I was thinking of creating some utility that the dev could call via context to manipulate the modal in its application.

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: Yeah I'll move this to `Hooks`, but although I agree with you regarding separating the module just to expose it to test or reuse it, I think it's beneficial to separate the component's responsibility (like that case) into a better understanding of code and maybe help in maintaining, debugging, and removing a lot of information that lies in the component.

packages/clay-modal/src/Footer.tsx
matuzalemsteles: You say about the warnings? or wrappings that we are doing?

packages/clay-modal/src/ClayModal.tsx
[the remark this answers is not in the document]
wincent: Well it's about where we fall on the spectrum:

before: no control -> in this PR: some more control -> allowing a function: even more control

Just wanting us to be intentional about where we land. My sense is that in this project, because it self-identifies as an implementation of Lexicon, that we have generally leant towards the left side. My habit has long been to lean in the other direction (not by adding a million options, but by doing things like allowing functions in certain places because that effectively constitutes an "escape hatch"). The problem with allowing more control, of course, is that you can wind up in a situation [like this](https://xkcd.com/1172/) where you lose the freedom to change things that people come to depend on.

packages/clay-modal/src/ClayModal.tsx
wincent: > I'm not sure if we would have a case of opening two modal at the same time

While we are responsible for how we use it in Portal, as an open source library, there are no guarantees about how others will do so. Ideally, we'd have something that is "fool-proof", and my comment was trying to get at the fact that it currently is not fool-proof (because it assumes singleton semantics without actually enforcing them).

packages/clay-modal/src/ClayModal.tsx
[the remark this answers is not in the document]
matuzalemsteles: Hmm yeah, makes sense. We were really leaning towards the Lexicon, I think we are taking short steps to give more flexibility and doors of escape like this, I think it is good to go slowly as we go slowly, and we can reach a middle ground without extrapolating.

packages/clay-modal/src/ClayModal.tsx
wincent: Yeah it's a balancing act. Nobody wants to read a 4,000-line file. But neither do they want to jump around between 20 files that all work together to implement a single feature. Have to find the right equilibrium. There was an interesting thread on Twitter about this a while back: https://twitter.com/sophiebits/status/1108855125135626241 — (and [this follow-up](https://twitter.com/sophiebits/status/1108899054325518336) in which she clarifies the sort of things that she thinks should go together usually).

packages/clay-modal/src/Footer.tsx
wincent: I just meant that overall it was a lot of massaging (unwrapping, checking etc) for what is essentially just `React.Children.map` + `React.cloneElement`. Let's see what it looks like after you've made the changes that you were going to make anyway.

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: Yeah, you're right about that.

migrate-the-clay-components-from-v2-to-v3.md
matuzalemsteles: Hmm, it's a good solution too. But this case the `<LeftRight />` had a pre-defined markup? I say in all the application.

Maybe it makes more sense to go with the left and right props to simplify and remove the `<FooterContainer />` elements. Although we can use `<LeftRight />`, for those who wish to use more than two buttons can add fragments, makes sense?

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: This is very interesting, I'll think more about it.

migrate-the-clay-components-from-v2-to-v3.md
wincent: `<LeftRight />` was just something like:

1. Assert that `props.children` contains exactly two children.
2. Render (something like):

```
<div>
  <div class="left">
    {child1}
  </div>
  <div class="right">
  </div>
</div>
```

wincent: (Updated because I accidentally submitted.)

matuzalemsteles: I've been thinking and adding `<Footer />` element with the props left and right and leaving the prop `footerContent` from ClayModal just by accepting a `React.ReactElement`.

```
<Footer
  left={<ClayButton />}
  right={(
    <>
      <ClayButton />
      <ClayButton />
    </>
  )}
/>

<ClayModal
  footerContent={<any>}
/>
```

wincent: Only thing I would want to check with Lexicon is whether it might be better to use semantic names like "primary" and "secondary" instead of spatial ones like "left" and "right". For example, if we expect to flip the order when in a RTL locale, then the semantic names might be less confusing. Up to you though.

matuzalemsteles: Hm yeah, hey @victorvalle any thoughts on this?

matuzalemsteles: hey @wincent, I've been thinking a bit about our conversations from yesterday and maybe it makes more sense to give more "freedom", that's just a rough sketch but I want your opinion on that.

```jsx
<Modal size="lg" status="danger">
  <Modal.Header onClose={} spritemap={}>
    Title
  </Modal.Header>
  <Modal.Body url={...}>
    <h1>Body</h1>
  </Modal.Body>
  <Modal.Footer
    primary={}
    secondary={}
  />
</Modal>
```

pat270: @wincent @matuzalemsteles ClayCSS has:
```
<div class="modal-footer">
  <div class="modal-item-first"></div>
  <div class="modal-item></div>
  <div class="modal-item-last"></div>
</div>
```

See http://pat270.github.io/lexicon/v2.10.0/content/modals/#modal-variants (modal success button). Modal Danger has a markup error.

packages/clay-modal/src/ClayModal.tsx
matuzalemsteles: hey @wincent the bad to go with the approach of animating the component in mount and unmount is that I have no control to suspend the unmount of the component until the animation finishes. Maybe I can control the onClose method to fire only when I finish the animation, but if I follow the approach I spoke in the other thread I will not have control over the onClick of the footer buttons.

Maybe we can accept a function in the `children` of `Modal` by passing the controlled method of onClose.

```jsx
<ClayModal onClose={onClose}>
 {handleClose => (
    <ClayModal.Footer
      primary={<ClayButton onClick={handleClose}>}
    />
  )}
</ClayModal>
```

What do you think?

migrate-the-clay-components-from-v2-to-v3.md
matuzalemsteles: Thanks @ pat270 I had forgotten about this, does it make sense then to name the props as `first`, `middle` and `last`?

pat270: @matuzalemsteles yes, I think `first`, `middle`, `last` makes sense.

packages/clay-modal/src/ClayModal.tsx
wincent: Yeah, I don't think you would want a way to "suspend" an unmount; the parent should be the one that controls when its children are mounted. That's why I said above that the modal would do the animation, then notify the parent, which actually does the unmounting. So what you say about calling `onClose` only at the end sounds like the right approach.

Having to pass through the `handleClose` callback probably isn't super ergonomic, but it could work. Did you look at prior art like [react-modal](https://github.com/reactjs/react-modal)? I see they have `onAfterOpen`, `onRequestClose` callbacks (and I also see that they do have an `isOpen` prop, although I haven't looked into the internals of it). I don't know anything about the project, so I can't vouch for its quality, but it might provide some food for thought.

matuzalemsteles: I took a look internally you can create an animation from their API, I think that's why they have many API's to customize the timeout, onAfterOpen to cover these cases... They have that power too due to the `isOpen` the even if we had `visible`, the component that controls when it will disassemble and mount.

I like the idea of ​​going with `handleClose`, I managed to make it work fine. I think it's something close to what Formik does with its handler functions, despite being another context and making more sense there.

matuzalemsteles:
hey guys, I've made some great updates after the reviews and this reflects our conversations.

- Use composition to mount the Modal component

I'll still add interaction tests, and when we get good with this component, I'll modify the migration guide.

wincent:
Found another interesting example of prior art to examine for ideas: https://github.com/reach/reach-ui/blob/master/packages/dialog/src/index.js (also has an `isOpen` prop... 🤔).

matuzalemsteles:
hey @wincent I had taken as base also the components in react native. It uses a lot of `visible` (although it is another context) https://facebook.github.io/react-native/docs/modal.

matuzalemsteles:
hey @bryceosterhaus you can take a look here and let me know what you think of these recent changes.

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,23 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React, {FunctionComponent} from 'react';
+
+export interface FooterProps {
+	first?: React.ReactElement;
```
bryceosterhaus: Should add some docs here since this is a public component

packages/clay-modal/src/Footer.tsx
```diff
@@ -0,0 +1,23 @@
...
+	last?: React.ReactElement;
+	middle?: React.ReactElement;
+}
+
+const Footer: FunctionComponent<FooterProps> = ({first, last, middle}) => (
+	<div className="modal-footer">
+		<div className="modal-item-first">{first && first}</div>
```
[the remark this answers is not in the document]
matuzalemsteles: ah yeah, ideally `div` should be rendered even if they have no content. I will remove the `&&` check.

matuzalemsteles:
@bryceosterhaus Update, I also added the missing migration guide.

PR #1857, opened 2019-04-15
Clay Color picker

bryceosterhaus:


matuzalemsteles:

packages/clay-color-picker/src/hooks.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	return [inputValue, handleNewInputValue];
+}
+
+const ESC_KEY_CODE = 27;
+
+/**
+ * Hook for closing dropdown when user hits ESC key or clicks outside the menu.
+ */
+export function useDropdownCloseInteractions(
```
matuzalemsteles: hey @bryceosterhaus I've been thinking of being able to share this rule with the other components, we have Dropdown here and DatePicker, both have the same behavior. But I wish it were not a dependency of a single component and that we have to import from it to use and also does not have to be a package of utilitarios of Clay, I do not want that we commit to him to make available, only for internal use that we can modify whenever I needed it, I've been thinking of something like the `shared` folder something like React does internally in its codebase. What do you think?

packages/clay-color-picker/src/GradientSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+interface GradientSelectorProps {
+	color: tinycolor.Instance;
+	hue?: number;
+	onChange?: (saturation: number, visibility: number) => void;
+}
+
+/**
+ * Renders GradientSelector component
+ */
+const GradientSelector: React.FunctionComponent<GradientSelectorProps> = ({
```
matuzalemsteles: I think we have some problems with using mouse position when I try to move to the lower right corner.

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -0,0 +1,240 @@
...
+										h: hue,
+										s: saturation,
+										v: visibility,
+									})
+								);
+							}}
+						/>
+
+						<div className="clay-color-map-values">
+							{rgbArr.map(([val, key]) => (
```
matuzalemsteles: It seems to have some problems when I try to type in the first input:

- Before the value is 0, I type 2 it forces the number to the 255 digit (Because we do not have 02 and this is invalid) but this happens only in the first RGB input.
- When I try to erase the value, it forces the values.

I'm not sure if the best would force or correct the values ​​when the user is typing, maybe doing nothing or adding feedback showing that this RBA is invalid is the best, the experience is bad as long as we want to try to enter a value and it gets forcing.

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+							setNewColor(tinycolor({h: hue, s, v}));
+						}}
+						value={hue}
+					/>
+
+					<div className="clay-color-footer">
+						<div className="form-group">
+							<div className="input-group">
+								<div className="input-group-item">
```
matuzalemsteles: When I am typing the first 3 digits it forces it to a color.

packages/clay-color-picker/src/hooks.ts
bryceosterhaus: Yeah thats a great idea. I think we could get away with either a generic `@clayui/shared` package or a hooks specific package `@clayui/shared-hooks`

bryceosterhaus: Before we extract it though, I think its best to keep this here and then once dropdown is created, we can reconcile the two and move it to shared.

packages/clay-color-picker/src/GradientSelector.tsx
bryceosterhaus: Yeah the problem is that both corners are the same hex value. Let me see if I can figure something out with this, its a bit tricky,

packages/clay-color-picker/src/Custom.tsx
bryceosterhaus: Yeah there seem to be a handful of interactive issues all stemming from `onChange`. I may need to set those to `onBlur` instead or I need to debounce setting the new color.

packages/clay-color-picker/src/hooks.ts
matuzalemsteles: I think we can only have a `shared` to add components (which are just small rules), utils and hooks if necessary, but I did not want it to be handled as a package, if doing that we would have to publish, now I wanted to avoid if that.

> Before we extract it though, I think its best to keep this here and then once dropdown is created, we can reconcile the two and move it to shared.

Yes No problems.

bryceosterhaus: Oh gotcha, i didn't see your reference to not publishing it. Yeah I think that is fine for now, not sure if we can avoid publishing it in the future though since it will be controlled by semver somewhere.

matuzalemsteles: Yes I thought about it, maybe we can handle it at build time so it does not have to be published. The concern I have in publishing it will be another package that will be available for people to use, so we would have to follow it as if it were another component keeping the semver and the good maintenance practices of not breaking, I wish we had the freedom to change those files without having to worry about it.

matuzalemsteles:

packages/clay-color-picker/src/GradientSelector.tsx
matuzalemsteles: It seems a lot better now, there's a case that might not be a problem. I can now move to the lower right corner, but when I move in the range hue, the Pointer goes back to the other side, this will happen with any update on the component.

bryceosterhaus: I think that issue might be unavoidable without writing specific code to handle `#000000` differently than others. What do you think of leaving that in now, since it doesn't effect the actual value, and then addressing it later if it becomes more of an issue?

matuzalemsteles: Yeah, no problem, it does not affect user usability.

bryceosterhaus:
Lets go ahead and merge it. I am gonna be working writing up some tests for a few components today using `react-testing-library`, and I can iron out things I find while implementing that.

matuzalemsteles:
ok. Merged!

PR #1867, opened 2019-04-16
 Fixes #1852 - Adds new APIs to handle DataProvider loading and error states

### LINKED ISSUE #1852: Improves Autocomplete feedback
Implement proposals for feedback improvements for the user in autocomplete.

- [Document](https://docs.google.com/document/d/1ZkpPajJBLioHBd2sEwE7Sb0FPdYr9ypwdL2ebHed1aQ/edit?ts=5c659b11)

matuzalemsteles:
hey @pat270 I'm having trouble setting the markup to add Loading with MultiSelect.


pat270:
@matuzalemsteles do you have a link to what it's supposed to look like? I can't find any documentation on it.

matuzalemsteles:
hey @pat270 this is a new proposal to improve the feedback of loading to the user, you can find the doc here https://docs.google.com/document/d/1ZkpPajJBLioHBd2sEwE7Sb0FPdYr9ypwdL2ebHed1aQ.

pat270:
Can you use in the `dropdown-menu`:

`<li><span class="disabled dropdown-item">Loading...</span></li>`
and 
`<li><span class="disabled dropdown-item">No results found</span></li>`

matuzalemsteles:
Thanks @pat270, I think we can with `disabled` if we need to change their behavior in the future we can remove from here. Another issue is that LoadingIndicator is sibling with the input, in Autocomplete it looks visually right but with MultiSelect it is getting bordered and the background is different compared to the MultiSelect focus.

pat270:
@matuzalemsteles I didn't see the loading indicator part at the bottom of the doc. I'll need to add some CSS to make it size properly, but the markup for autocomplete should be something like:
```
<div class="input-group">
		<input class="form-control input-group-inset input-group-inset-after" type="text" value="">
		<span class="input-group-inset-item input-group-inset-item-after">
			<span aria-hidden="true" class="loading-animation"></span>
		</span>
		<ul class="autocomplete-dropdown-menu dropdown-menu show">
			...
		</ul>
	</div>
</div>
```
I haven't tested multi select yet, but I'll send a pull under this after I get done with another task.

matuzalemsteles:
@pat270 no problem, thank you very much for your help.

wincent:
> but the markup for autocomplete should be something like

One thing about nesting the `input` and the `ul.autocomplete-dropdown-menu` at the same level is that the dropdown could get cropped if the whole component is nested inside an `overflow: hidden` container, right? I'm not sure how we've handled this kind of thing previously on this branch, but in the React version the typical way to handle it would be to use a [portal](https://reactjs.org/docs/portals.html) to locate the drop-down elsewhere in the DOM.

matuzalemsteles:
hey @wincent, yeah we're doing the same for the Metal component, using portal (it's the same as React's).

pat270:
@wincent that is correct. The markup in my example is for integrating Bootstrap's Dropdown plugin with MultiSelect.

PR #1877, opened 2019-04-18
Fixes #1875 Card Type Template and Template Card should reset styles from other cards

### LINKED ISSUE #1875: ClayCSS Card Type Template and Template Card should reset styles from other cards
Current limitations on `ClayCard` requires Clay CSS to support cards like:
```
<div class="card card-type-asset card-type-template file-card template-card"></div>
<div class="card card-type-asset card-type-template image-card template-card"></div>
<div class="card card-horizontal card-type-directory card-type-template template-card-horizontal"></div>
```
where Template Card styles should win over Asset Cards. See #1865.

pat270:
@matuzalemsteles @carloslancha  I had to add `aspect-ratio-item` to FileCard, ImageCard, and UserCard so I have something common to select when overriding styles. There were also mixed tabs and spaces so I converted everything to tabs.


matuzalemsteles:
Thanks @pat270, can you take care of updating snapshots and isomorphic tests before we proceed?

PR #1878, opened 2019-04-18
Fixes #1835 - Add test suite to DatePicker

### LINKED ISSUE #1835: Add tests for DatePicker

matuzalemsteles:


bryceosterhaus:
@matuzalemsteles great job, I really like these tests. I was uneasy about them at first, but after reading through them this seems like a much more solid way of testing our features.

Only question I would have would be regarding the `data-testid` attributes in production. Since this is a library and not an app, we should probably remove those before building. I saw on the `react-testing-library` docs they recommended using https://www.npmjs.com/package/babel-plugin-react-remove-properties. What do you think?

matuzalemsteles:
hey @bryceosterhaus yes I've been thinking about testing our components better, we were using a lot of snapshots, many cases make sense, but were not guaranteeing a good coverage of tests in cases of components with interactions, I want us to feel more secure with them.

>Only question I would have would be regarding the data-testid attributes in production. Since this is a library and not an app, we should probably remove those before building. I saw on the react-testing-library docs they recommended using https://www.npmjs.com/package/babel-plugin-react-remove-properties. What do you think?

Perfect, I'll add, I had not seen it. Thanks Bryce!

bryceosterhaus:
Looks good to me! I am gonna go through and update some tests today using this new approach.

PR #1901, opened 2019-04-23
Update unit tests

bryceosterhaus:


packages/clay-color-picker/src/Hue.tsx
```diff
@@ -45,13 +45,12 @@ const Hue: React.FunctionComponent<HueProps> = ({
 	}, [value]);
 
 	useEffect(() => removeListeners, []);
-
+	// console.log(x);
```
matuzalemsteles: I think you forgot to remove this 😅

matuzalemsteles:

packages/clay-color-picker/tests-util.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+interface FakeMouseEventInit {
```
matuzalemsteles: Great job with that.

packages/clay-color-picker/src/index.tsx
```diff
@@ -202,7 +202,6 @@ const ColorPicker: React.FunctionComponent<Props> = ({
 										onChange={onValueChange}
 										onColorsChange={onColorsChange}
 										spritemap={spritemap}
-										value={value}
```
matuzalemsteles: Do we have any reason why we removed this? I can not tell if the reason was this but we are causing a regression, now I can no longer type in the input in the footer within the dropdown.

bryceosterhaus: Good catch. I had to make sure and check the activeElement for this. The reason I removed the `value` prop is because it isn't necessarily accurate because it depends which index is selected from the custom colors.

matuzalemsteles: Makes sense, thanks for the information.

matuzalemsteles:

PR #1912, opened 2019-04-26
Fixes #1819 - Create DropDown component

### LINKED ISSUE #1819: Migrate DropDown to React

bryceosterhaus:
*Note:* I am going to wait until this PR is agreed upon before updating the migration docs. That way it can be added once finalized.


packages/clay-drop-down/src/Item.tsx
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export type DropDownItemType = FormItem | BaseDropDownItem;
+
+interface Props {
+	data: DropDownItemType;
+	spritemap?: string;
+}
+
+const DropDownItem: React.FunctionComponent<Props> = ({data, spritemap}) => {
```
matuzalemsteles: I was a little curious why you did not follow with the idea of ​​composition (renders props...). My concern I had with our Metal component was that it was getting complex as more use cases appeared, some people wanted to have more control over dropdown items.

One suggestion I have about using the composition, I think we can give off some of the Lexicon and offer escape doors to "get away". This is just a suggestion and a sketch:

#### `<Dropdown />`
Just the basic component that will control the opening and closing of Dropdown, click away, ESC... and maybe listen to the items click events directly and onChange for checkbox and radio cases...

```jsx
<Dropdown
  onItemClick={...} // maybe
>
  {children}
</Dropdown>
```

#### `<Dropdown.Search />`, `<Dropdown.Caption />`, `<Dropdown.Action />`

```jsx
<Dropdown>
  <Dropdown.Search ... />
  <Dropdown.Menu>...</Dropdown.Menu>
  <Dropdown.Caption ... />
  <Dropdown.Action>
    <ClayButton ... />
  </Dropdown.Action>
</Dropdown>
```

#### `<Dropdown.Menu />`
Just to wrap the components of items.

```jsx
<Dropdown.Menu>
  <Dropdown.Item>Item</Dropdown.Item>
</Dropdown.Menu>
```

#### `<Dropdown.Item />`
These are our standard variations, checkbox, radio, default... Nothing new, we can keep our APIs here.

```jsx
<Dropdown.Item>Item</Dropdown.Item>
<Dropdown.Divider />
<Dropdown.Item variant="checkbox" value="checkbox">Checkbox</Dropdown.Item>
```

#### `<Dropdown.Group />`
Just to create group of items, not to need a props of `items` inside the `<Dropdown.Item />` element.

```jsx
<Dropdown.Group label="Group 1">
  <Dropdown.Item>Item</Dropdown.Item>
</Dropdown.Group>
```

What do you think?

bryceosterhaus: I think this is a great idea. I originally set out to just copy the soy version, but I think this makes more sense and lets us manage less logic on our end. Ill go ahead and update and see how it works.

bryceosterhaus:
@matuzalemsteles updated! I went ahead and went with the composition route and also added metal-position for alignment

bryceosterhaus:
See storybook [here](http://5cc77869367f7f00088ac51d--relaxed-ptolemy-6978dd.netlify.com/?path=/story/claydropdown--default)

packages/clay-drop-down/src/hooks.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Hook for closing dropdown when user hits ESC key or clicks outside the menu.
+ */
+export function useDropdownCloseInteractions(
+	elementRef: React.RefObject<HTMLElement>,
+	setActive: (val: boolean) => void
+) {
+	React.useEffect(() => {
+		const handleClick = (event: MouseEvent) => {
```
matuzalemsteles: It seems that when I click on the trigger to close the modal it closes and opens, for some reason it is losing its active state.

packages/clay-drop-down/src/Menu.tsx
```diff
@@ -0,0 +1,29 @@
...
+const DropDownMenu: React.FunctionComponent<
+	React.HTMLAttributes<HTMLUListElement> & {
+		hasLeftSymbols?: boolean;
+		hasRightSymbols?: boolean;
+	}
+> = ({children, className, hasLeftSymbols, hasRightSymbols, ...otherProps}) => {
+	return (
+		<ul
+			{...otherProps}
+			className={classNames('dropdown-menu list-unstyled show', {
```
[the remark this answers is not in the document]
bryceosterhaus: My bad, this file shouldn't exist any longer. All of the menu logic will be in `index.tsx` to handle all of the positioning.

packages/clay-drop-down/src/hooks.ts
bryceosterhaus: good catch, i forgot to test that and I think I know the cause.

packages/clay-drop-down/src/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+};
+
+interface DropDownMenuProps extends React.HTMLAttributes<HTMLUListElement> {
+	alignElement: HTMLElement | null;
+	hasLeftSymbols?: boolean;
+	hasRightSymbols?: boolean;
+	ref?: React.Ref<HTMLUListElement>;
+}
+
+const DropDownMenu: React.ComponentType<DropDownMenuProps> = React.forwardRef(
```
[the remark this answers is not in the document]
bryceosterhaus: This should be the only instance of Menu. I will update

bryceosterhaus:
@matuzalemsteles updated. 

One thing I was thinking through was the "shared hooks" idea. Since this component and color picker could use the same hook for closing the element when outside is clicked.

I know you didn't want to create another package for that, but I wasn't entirely sure how we would go about doing that since we will need to version them somehow. Unless we always bump all clay component versions when a single component is updated. Otherwise, we may have color-picker at v3.0.0 and then drop-down at v3.0.5 and that minor version may change the shared hooks. I'm not sure I see another solution other than creating a separate versioned package for shared hooks.

matuzalemsteles:
hey @bryceosterhaus, the idea of ​​not having published is exactly that we can break the components, hooks... of shared without having to version the code, we will always be guaranteeing that it will have compatibility with our internal components, as we will have total control of it.

For example, we can have the `shared` folder coexisting with the clay packages side by side, and we can import the components normally as if importing any other, in the end it will not be a package, but only a folder. For ease we can add the aliases in the TypeScript configs but we still have to look at what the final builds look like.

packages/clay-drop-down/src/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+			hasLeftSymbols,
+			hasRightSymbols,
+			...otherProps
+		},
+		ref?: React.Ref<HTMLUListElement>
+	) => {
+		return (
+			<ul
+				{...otherProps}
+				className={classNames('dropdown-menu list-unstyled show', {
```
matuzalemsteles: As I had said in the other comments, we should have inside the `ul` only `li` tags.

For example:
```html
<div aria-labelledby="theDropdownToggleId" class="dropdown-menu">
	<div class="alert alert-fluid alert-info" role="alert">
		You can customize this menu or see all you have by pressing "more".
	</div>
	<ul class="list-unstyled">
		<li><a class="dropdown-item" href="#1">D Structure</a></li>...
	</ul>
	<div class="dropdown-caption">Showing 7 of 203 Structures</div>
	<div class="dropdown-section">
		<button class="btn btn-block btn-secondary">More</button>
	</div>
</div>
```

bryceosterhaus:
@matuzalemsteles updated. Thanks for catching that, I had forgotten to verify ul > li.

wincent:
If I understand the proposal correctly, I don't like the idea of having a `shared` folder that is not a package for a few reasons.

One is that it would introduce another dependency pattern into the monorepo (ie. instead of just having packages that depend on packages, you would additionally have packages that depend on files outside of their package directories via relative requires). This is non-obvious and potentially confusing.

The other is that this dependency would be implicit (ie. not expressed in the `package.json` dependencies) and therefore brittle: it would rely on us carefully including redundant copies of the shared files in the published packages (ie. via maintaining a correct "files" property, for example). In other words, there are more ways that things could go wrong.

A final reason is that once the packages are installed into `node_modules`, those dependencies aren't "shared" any more, but duplicated. A bundler is unlikely to deduplicate them because it won't know they are actually the same, so I guess that just leaves gzip.

I think just having a separate package would be the most robust solution. A separate question is whether you want to versioning all the packages independently or in lock-step, and the answer to that may depend on whether you view "Clay" as a cohesive set of components that should always be used together, or as a toolbox of individual tools that you want people to mix and match. The former seems to make more sense to me.

matuzalemsteles:
hey @wincent, I understand your concern that we are sending duplicate code and other factors, but my biggest concern is making a package published is that we will be opening the doors so that they can use them and then we will have to follow semver and the our new practices... The idea behind not publishing is that we can break and cover only our use cases and not have to get a issue saying that we broke a team because we modified something...

wincent:
We could alleviate the concern about undesired use by naming the package "clay-internals" or something similar. There's really no limit to how "ugly" you can make it to discourage external use. eg.

```javascript
import {useFoo} from 'clay-internals/__private__/hooks/useFoo';
```

(etc). I get that you don't want to be on the hook 😂 for maintaining something re-usuable that you don't want others to use, but the mechanism you're proposing is very unconventional and comes with considerable costs.

matuzalemsteles:
haha seems to be a "good solution" I would not mind naming with something like that, I understand those costs and it really may not be a good idea to go with them.

bryceosterhaus:
I agree with @wincent, I think its best to just publish it. This way we also don't have to think about it differently than our other packages we maintain. One solution to keeping it from being another publicly consumed packaged would be to just specify in the README that its not intended for consumption, and then keep any docs for it in a DOCS.md file. This way both NPM and GH will show the README, and see that its not supposed to be publicly used.

matuzalemsteles:
hey @bryceosterhaus we can go with that forward, I think it's cool to go with the idea of ​​@wincent for the information to get more explicit in `import`.

jbalsas:
Hey @matuzalemsteles, @bryceosterhaus, @wincent, I know I'm late to the party, but just in case.

**All of our code needs to be public and accesible**. We can't depend on local paths or local build processes other than to generate derived sources (icon spritemaps, for example).

The bundling/building responsibility is not ours and how Clay is built and consumed should be left to the consumers. Hiding some of our internals might make it impossible for us to properly assemble Clay and its components to our liking at the other end.

matuzalemsteles:
hey @jbalsas, I understand that we should leave our code accessible and public but the idea of ​​`shared` is to leave our hooks and components that we reuse internally to build the components, with `useTransition`, `useDropdownInteraction` for example. I think if we leave this as a public we will be tying ourselves to semver inside and out, this can limit us to make changes, we have to have an internal layer like this, so that we can break things and cover only our cases of use.

jbalsas:
Hey @matuzalemsteles, it we're talking runtime code, it needs to be out in the open, because how it gets bundled is not up to us entirely but to the consumer. If it were a build thing, then it's fine, that would be internal, but the code that runs can't be hidden away.

matuzalemsteles:
Oh, sorry @jbalsas, I misinterpreted it, sorry for that. But you say about names in import? not to use `__private__` or anything like that.

jbalsas:
Hey @matuzalemsteles, I don't mind using weirdly convoluted names for things we want to hide away. 

One thing to consider though, is that other teams might need to build their own components. If we heavily rely on some internals, they will surely need those too to make their own developments behave in similar ways. With that in mind, we should be very careful about what we tag as `internal_only`, because we should not be just serving the components, but also the building blocks that make creating other components possible.

packages/clay-drop-down/src/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+			<span ref={triggerElementRef} style={{display: 'inline-block'}}>
+				{React.cloneElement(trigger, {
+					className: 'dropdown-toggle',
+					onClick: () => onActiveChange(!active),
+				})}
+			</span>
+
+			{active && (
+				<Portal>
+					<div
```
matuzalemsteles: hey @bryceosterhaus I've been thinking about making this wrapper a component too `<ClayDropDown.Menu />`, we'll have some components that we do not want to trigger for example, just your content. What do you think?

bryceosterhaus: Yeah thats a good point. I'll either look into adding that to this PR or wait and do it in another refactoring ticket that pulls it out of other components as well.

matuzalemsteles: It looks better, we can go ahead and merge this component.

bryceosterhaus: 👍 I just need to update this PR with the transitions from diego's PR. Ill update today and let you know

wincent:
> With that in mind, we should be very careful about what we tag as `internal_only`, because we should not be just serving the components, but also the building blocks that make creating other components possible.

Another thing to bear in mind: it is much easier to give something than to take it away. That is, if something is marked as "internal"/"private", we're always free to promote it to "public" API. But conversely, things that are public are really hard to turn into private, because those are breaking changes, requiring a pretty painful and lengthy process of deprecation and eventual removal (or failing that, a long-term maintenance of something that you'd really rather get rid of or change).

In other words, I wouldn't agonize over too much over making things private: the cost of making a mistake and labeling something as "internal"/"private" that should really be public is pretty low because it is so easily corrected. The inverse mistake (of making an API public that shouldn't be) is much more costly.

bryceosterhaus:
@wincent thats a great point, I think it'd be very wise for us to try and always start with limited functionality, whether that be the API of a component or what packages are public, and then slowly add the "must need" things. I think it'd be a good idea to re-visit this before releasing 3.0 and go through each package and ask whether it should be public or not.

bryceosterhaus:
@matuzalemsteles updated. I went ahead and refactored a bit so that `<DropDown.Menu>` is easily re-usable. I went ahead and added it to date-picker and color-picker as well.

matuzalemsteles:

packages/clay-drop-down/src/Menu.tsx
```diff
@@ -0,0 +1,110 @@
...
+	 */
+	hasRightSymbols?: boolean;
+
+	/**
+	 * Callback function for when active state changes.
+	 */
+	onSetActive: (val: boolean) => void;
+}
+
+const DropDownMenu = React.forwardRef<HTMLDivElement, Props>((
```
matuzalemsteles: @bryceosterhaus we would have to deal with focus control later, but as it is composition we have to think of a way to do that.

packages/clay-drop-down/src/Menu.tsx
```diff
@@ -0,0 +1,110 @@
...
+	 * Align.RightCenter = 2;
+	 * Align.BottomRight = 3;
+	 * Align.BottomCenter = 4;
+	 * Align.BottomLeft = 5;
+	 * Align.LeftCenter = 6;
+	 * Align.TopLeft = 7;
+	 *
+	 * Defaults to BottomLEft
+	 */
+	alignmentPosition?: number;
```
matuzalemsteles: Would not it be better to expose an enum? so you do not need to export `Align` to pass the correct information and can set the best type. What do you think?

```typescript
enum AlignmentPosition {
  TopCenter = 0,
  TopRight = 1,
  RightCenter = 2,
  BottomRight = 3,
...
}
```

packages/clay-drop-down/src/Menu.tsx
bryceosterhaus: I have a few ideas how this might be fixed, let me see if I can add something

packages/clay-drop-down/src/Menu.tsx
bryceosterhaus: I went directly with the `Align` export from metal-postion because that is the consumer of that prop. I didn't want to create another definition for it and would rather just pass metal-position's align straight through.

matuzalemsteles: No problem, I think in this way the experience improves who is using Typescript and React for having a full type definition and if we change it in the future we would not have problems with export.

packages/clay-drop-down/src/Menu.tsx
matuzalemsteles: Great! This may help in the creation of Autocomplete, I am waiting for the DataProvider and Dropdown to come in to send the proposal.

bryceosterhaus: Wasn't super happy with the solution I came up with. Gonna go ahead and merge for now, and still look into ideas for this.

bryceosterhaus: https://github.com/liferay/clay/issues/1987

PR #1948, opened 2019-05-01
Fixes #1941 - Implement panel component with collapse ability

### LINKED ISSUE #1941: Migrate clay-collapse to React

bryceosterhaus:
I haven't added the animation for the height expanding yet. Wanted to make sure this was the route we wanted to go first.


matuzalemsteles:

matuzalemsteles:
hey @bryceosterhaus this component looks good to me but I'm not sure if we should follow with Panel implemented with Collapse or just a hook for the Collapse, I do not know what use cases it had or if it was used.

But it's fine with me if we go with that, since it no longer makes sense in the context of React to write as it was before.

bryceosterhaus:
@matuzalemsteles I think we just keep collapse prop for panel. We can extract the transition height logic to its own hook and share it throughout clay though. I will wait for Diegos PR for navigation bar since he has a similar hook for transitioning height.

bryceosterhaus:
@matuzalemsteles updated.

I went ahead and extracted `useTransitionHeight` from nav bar and put it into a `@clayui/shared` package. I went with the shared package because we need to publicly expose it as well as have versions for it. I'm not too concerned about the overhead in maintaining it since it is just another package similar to the others we are creating and should go through the same workflow. For now I think we isolate shared hooks there and avoid publicly promoting it as a package that other people should consume, this way we can keep from the overhead of adding features to this specific package.

packages/clay-shared/src/useTransitionHeight.ts
```diff
@@ -4,10 +4,32 @@
...
+function setCollapseHeight(collapseElementRef: React.RefObject<any>) {
+	if (collapseElementRef && collapseElementRef.current) {
+		collapseElementRef.current.setAttribute(
+			'style',
+			`height: ${collapseElementRef.current.children[0].clientHeight}px`
+		);
+	}
+}
+
+export function useTransitionHeight(
```
bryceosterhaus: I decided to make this explicit to only "height" right now. In the future I can imagine this being expanded to all transitions, but wanted to narrow our scope now and not make it generic until we actually need it for other use cases.

packages/clay-shared/README.md
```diff
@@ -0,0 +1,17 @@
+# clay-shared
+
+Shared utilities and hooks used across clay components
+
```
matuzalemsteles: hey @bryceosterhaus you can put some warnings that this package for a moment is only for internal use and that the API's are unstable and can be modified at any time and that "use at your own risk".

packages/clay-shared/src/useTransitionHeight.ts
matuzalemsteles: Yeah, makes sense. You did well.

packages/clay-shared/README.md
bryceosterhaus: Good call, I went ahead and updated the README.

PR #1962, opened 2019-05-06
Fixed #1946 - Create MultiSelect component in React

### LINKED ISSUE #1946: Migrate clay-multi-select to React

bryceosterhaus:


bryceosterhaus:
@matuzalemsteles updated and added autocomplete functionality. Once this is approved I will add docs after.

Also, I'm gonna spawn off another ticket for keyboard functionality such as down arrows and enter key. This will likely be a hook that comes from autocomplete or the dropdown component.

matuzalemsteles:
Great @bryceosterhaus. I'll take a look at this later. I'm trying to run the storybook in Netlify, but it's always failing by 70%, I do not know why this is happening.

On the keyboard functionality, I think it would be better to come from DropDown and may also be related to the focus in #1987.

matuzalemsteles:

packages/clay-multi-select/src/MultiSelect.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+				if (validItems.length === items.length) {
+					setIsValid(false);
+				}
+			}
+		}
+	};
+
+	return (
+		<div className="form-group" ref={forwardRef}>
```
matuzalemsteles: @bryceosterhaus I've been thinking we could do a bit differently, instead of having all the components encapsulated here without the outside world being able to tweak it without an API, we can do what we were doing with `Autocomplete` and `DropDown`, that means we will write this component on a lower level.

One thought about this I have is:

- **Validation** can we create any integration with Formik in any way?
- The validation components, we can take to `<ClayForm />`
- `<label>{label}</label>` can we have a `ClayForm.Label`?
- `<div className="input-group input-group-stacked-sm-down">` can we take it to `Clay Form.InputGroup`?

We have used these classes a lot to handle inputs, I think we can move to `ClayForm` and let people think about creating composites and we can make this component high-level, building with them.

Just a mental sketch:
```jsx
const [items, setItems] = useMultiSelect(['one']);

return (
	<ClayMultiSelect>
		<ClayForm.Group>
			<ClayForm.Label>{label}</ClayForm.Label>
			<ClayForm.InputGroup>
				<ClayForm.FormControl>
					{items.map((item, i) => (
						<ClayLabel
							closeButtonProps={{
								onClick: () => {
									setItems([
									...items.slice(0, i),
										...items.slice(i + 1),
									]);
								},
							}}
							key={i}
							spritemap={spritemap}
						>
							{item}
						</ClayLabel>
					))}
					<ClayForm.Input />
				</ClayForm.FormControl>
				<ClayAutocomplete.DropDown>
					{'...'}
				</ClayAutocomplete.DropDown>
				<ClayForm.Feedback />
			</ClayForm.InputGroup>
		</ClayForm.Group>
	</ClayMultiSelect>
);
```

This would be something close, I think it would have some design challenges yet, we will have to deal with a use case that implies in the markup that is to add the loading indicator... It would be something that someone would write composing with the other components to get to that result, but we can deliver all this with a high-level as well.

packages/clay-multi-select/src/MultiSelect.tsx
```diff
@@ -0,0 +1,196 @@
...
+	items: string[];
+	label?: React.ReactText;
+	onItemsChange: (val: string[]) => void;
+	onInputChange?: (val: string) => void;
+	spritemap?: string;
+	validateOnBlur?: boolean;
+	validationFn?: (val?: string) => boolean;
+}
+
+const MultiSelect: React.FunctionComponent<IProps> = ({
```
matuzalemsteles: I've missed some interactions in this component, it's probably something we should deliver on a high-level component.

- Press backspace in the input when it does not have a value, it takes to the focus of the Label, pressing again, erases the label.
- Navigate the label with Tab or keyboard
- When deleting an item, you must return the focus to the input...

This is the technical document with use cases and interactions: https://docs.google.com/document/d/1cIx6RkWNbnho_fOSR__3Fp5alIBE_ObhO-jecfXlA9I/edit#heading=h.jkrp0t7cj15n

jbalsas: For future reference, we should ping @drakonux (and open review requests) as we implement the components.

Also, maybe we should start committing the specification documents so we can keep track of them here and have them versioned... wdt?

packages/clay-multi-select/src/MultiSelect.tsx
bryceosterhaus: @matuzalemsteles I was actually just thinking about this the other day. `MultiSelect` is essentially a high-level component of `Input`. What do you think about just getting rid of this package and making it a high level version of Input? I think same sort of logic would likely apply to autocomplete. More so works for multi-select since there is no specific class names for it.

`ClayInput`
`ClayInputWithAutoComplete`
`ClayInputWithMultiSelect`

What do you think?

packages/clay-multi-select/src/MultiSelect.tsx
bryceosterhaus: Do we have docs like this for all clay components?

jbalsas: Yeah, we most certainly do, we just need to ask for them and maybe keep them closer to heart.

This is something we've often discussed regarding our workflow. Documents in google docs are live and subject to change without notice. This has resulted in doubts as to wether a component was wrongly implemented or if the definition changed underneath. If we could keep the definitions here and version them through git, that might alleviate some of the pain and also help collect our requirements.

What do you think?

matuzalemsteles: I would think it would be great, keep the documents in git and versioned, we need to versioned these documents to not get some of the problems we had in the 2.x life, I got a lot of complaints because Clay was "inconsistent" with Lexicon (this was a problem when we adopt monolithic components...). But do you say keep our own documents or in conjunction with Lexicon? One good thing is that we can collaborate much more closely with them if we keep those components together.

packages/clay-multi-select/src/MultiSelect.tsx
matuzalemsteles: Hmm makes sense, I like that. Sounds like a good idea, I think we can test it.

bryceosterhaus: Ill send a separate PR today with that general idea for multiselect

bryceosterhaus:
New spin on multi-select seen here, https://github.com/liferay/clay/pull/2123.

I added the keyboard functionality for backspace. I will add arrow and enter keys with a different PR.

bryceosterhaus:
Closing in favor of https://github.com/liferay/clay/pull/2123

PR #1964, opened 2019-05-07
Revert the use of the atlas by link

matuzalemsteles:
hey @pat270 I'm reversing the use of Atlas direct from css, it was breaking the site's css in production. We can analyze this better in 3.x.


pat270:
@matuzalemsteles `atlas.css` needs to be moved above `{this.props.headComponents}`.  I don't know why, but the order of the files in `<head>` are different in production vs development.

PR #1965, opened 2019-05-07
Clayui.com `atlas.css` should be above `this.props.headComponents` Ga…

pat270:
…tsby switches up the load order for files in <head> in production

@matuzalemsteles I fixed the CSS issue in #1964.


matuzalemsteles:
@pat270 thanks for this.

PR #1966, opened 2019-05-07
Add docs for color-picker interfaces

bryceosterhaus:


packages/clay-color-picker/src/Splotch.tsx
```diff
@@ -7,9 +7,19 @@ import React from 'react';
 import tinycolor from 'tinycolor2';
 
 interface Props extends React.HTMLAttributes<HTMLButtonElement> {
+	/**
+	 * Flag that indicates whether the splotch is the active one selected
+	 */
 	active?: boolean;
-	className?: string;
```
bryceosterhaus: Not necessary since this comes from the `extends`

PR #1968, opened 2019-05-08
LEXI-576 Add ta-IN flags for Tamil using the Hindi-India flag (hi-IN)

dgomezg:
Fixes #LEXI-576

### DIFF
--- a/clayui.com/static/images/icons/flags-ta-IN.flag
+++ b/clayui.com/static/images/icons/flags-ta-IN.flag
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="512" height="512" x="0px" y="0px"
+	 viewBox="0 0 512 512" enable-background="new 0 0 512 512" xml:space="preserve">
+<rect y="64" fill="#C0CFD8" width="512" height="384"/>
+<rect x="16" y="80.3" fill="#F5984C" width="480" height="112"/>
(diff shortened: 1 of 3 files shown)


pat270:
@dgomezg Thanks for the pull, but I think this is the wrong place for this. Clay CSS already has an Indian flag and Portal should reuse this if possible (it should be). If not we need to allow for this otherwise we will have a lot of duplicate flags. Let me get back to you.

dgomezg:
Thanks @pat270 for your review and explanation. The reason of duplicating the flag is because I need to add a new language to the portal and I was told that, for the new language to work, we need also to add the language flag to clay. And I copied the flag from India available with filename the Hindi ISO code (hi-in), using the Tamil ISO code (ta-in). 

My guess was that liferay portal was looking for the SVG flag using the ISO code for the language. Maybe I need some more info on how the language selector works under the hood to look for the language flag and how it is related to clay.

Also, I checked the other case in which we use the same flag for two different languages (or variations) is with Serbian-Cyrillic (sr_RS) and Serbian-Latin (sr_RS_latin). In this case, we already have the flags duplicated in Clay, both in `packages/clay-css/src/images/icons` and in `clayui.com/static/images/icons`. Also the entry in `packages/clay-css/src/scss/functions/_global-functions.scss` is duplicated for 'sr-rs' (line 480) and 'sr-rs-latin' (line 478).

Sorry for the duplication, I agree that if is possible it's better to reuse the same flag with a different flag, I will be grateful if you can provide me more info on how to do it or help me with that.

Also, maybe we should address the duplication of flags for Serbian variants.

Thanks in advance for your help.

jbalsas:
This is as good a moment as any other to promote http://www.flagsarenotlanguages.com :)

/cc @victorvalle

pat270:
@dgomezg You're right, I was looking for duplicate flags in the Clay CSS repo and I missed that one. I took a look at the template for `site-navigation-language-web` and it looks like Portal doesn't support it. We'll duplicate flags until we fix that.

pat270:
New pulls
3.x: https://github.com/liferay/clay/pull/1973
2.x: https://github.com/liferay/clay/pull/1974

PR #1969, opened 2019-05-08
Next.clayui.com prep CSS to toggle between Base/Atlas and clean up stuff and some

matuzalemsteles:


pat270:
@matuzalemsteles I was thinking of it from a testing point of view just to make sure everything plays nicely together. There is one Clay jquery plugin that we have `side-navigation.js`. If we want to demo that one on this site we need jquery until we have a replacement. Other than that we can remove them if you like. The only thing we need is `svg4everybody` (not included yet). Svg's using the `use` tag don't work in IE or Safari.

matuzalemsteles:
Hm yeah I think we still need `svg4everybody`. On the `side-navigation.js` we can work later to replace it with a React maybe... but for now I'm going to go with that.

jbalsas:
I think we're going to attempt to remove the `side-navigation.js` plugin pretty soon, so I wouldn't worry about it.

PR #1971, opened 2019-05-09
Clayui.com adds theme toggle

pat270:
There's probably a more elegant way to do this, but I don't know react.

### DIFF
--- a/clayui.com/src/components/Sidebar/Search.js
+++ b/clayui.com/src/components/Sidebar/Search.js
@@ -22,27 +22,25 @@ class Search extends Component {
 
 		return(
 			<div className="page-autocomplete">
-				<div className="form-group">
-					<div className="input-group">
-						<input
-							className="form-control"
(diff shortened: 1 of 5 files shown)


matuzalemsteles:

marcoscv-work:

marcoscv-work:
Hey @pat270 
Why do we need to add this control? do you want it for development reasons?
If we really need to switch atlas/base, could we use a mechanism only visible for us?

pat270:
@marcoscv-work I would say because a lot of our customers base their theme on Styled instead of Classic? I remember seeing something about extending Classic Theme not being supported by our tools.

marcoscv-work:
Hey guys, sorry for the delay..

I suppose people in USA is asking for this, is it right @pat270 ?
My only concern is where should we place this control to switch base/atlas
I this is a strategic decision too, so @jbalsas could you help us to make a decision?

PR #1975, opened 2019-05-09
Fixes #1953 - Require checked and onChange props for ClayCheckbox

### LINKED ISSUE #1953: Explicitly defined props and interfaces
There're some components, like clay-checkbox, that should have `disable` prop. Now we can pass this kind of props directly to the components due we're rendering `...otherProps` but... my question is:

Should all this known props be defined as API? For me it's a yes.

Another thing is that I see some props defined inside the component (i.e. `checked` in `clay-checkbox` but is not defined in the Props Interface. Should we define all component props in those interfaces? For me it's also a yes, but I don't know if I'm missing something.

Thanks guys!

bryceosterhaus:


bryceosterhaus:
@matuzalemsteles had to rebase. Went ahead and merged after rebase.

carloslancha:
Hey @bryceosterhaus is `onChange` param really required? What if I just want to place a checkbox that will be handle on the server after a form is send? The same with `checked`, we're forcing the developer to pass a param that may do nothing or may not be really required (like a false `checked`)

For me, this params are still optional. What do you think?

PR #1977, opened 2019-05-09
Fixes #1976 - Create DataProvider component and useResource hook in React

### LINKED ISSUE #1976: Migrate clay-data-provider to React

matuzalemsteles:
This component has more features compared to the previous component and much more possibilities and also follows some concepts that I inspired in React Apollo...

For those who use `render props` you can use the DataProvier component directly but anyone who needs to manipulate something in more detail can use `useResource` hook. As we do not yet have stable Suspense it's a great starting point here for consuming data and getting network status. Maybe we can integrate with Suspense when it's available...

You may want to read more about how everything works, I recommend starting with the `types.ts` file, having a lot of explanation about the APIs and can give an overview of everything.

I have not done the tests and the documentation yet, after we have discussed more about this component I will write.

### Cache

The requests have a cache and follow a policy that is set via `fetchPolicy`, the caches follow the LRU rule and can be stored inside a Context or Store of the application itself to centralize the data or remain with the cache during navigation... can do this just by passing the reference to the `storage` API.

### Retry

This retry was based on the Apollo Retry plugin (https://www.apollographql.com/docs/link/links/retry), many of its concepts are being applied here. In summary you can [avoid thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem) by setting the API, by default it is already enabled.

### Network Status

The component provides network status information such as loading, error, polling, refetch... if anyone uses this information to further customize the component to provide feedback to the user. You can achieve this by enabling the `notifyOnNetworkStatusChange` API, which will cause new renderings as the network status changes.

### DataProvider component

It was built for `renders props` makes it easier to compose with other components that need some data out, see the examples in the Storybook.

### `useResource` hook

Virtually all cache, retry, fetch... logic is inside the hook, so if you do not need the DataProvider component, the hook can be used instead, though strongly recommending the use of the component.

### ProTips

That way we can create a very interesting application using maybe just the DataProvider with `storage`, maybe we can create a hook that can read the cache data so that parts of the application can use this information without needing the DataProvider component.

We can imagine this with an implementation of React Apollo... If you pass the same storage reference for all your DataProvider, it will consume the cache only from that location and if you have more than one DataProvider consuming the same data, since the user interacts with some of them the data can be cached and when the user interacts with the other it will consume the information from the cache.


matuzalemsteles:
hey guys, one thing I was thinking about at the end of the week, add the possibility of being able to set a data response schema, so I quickly looked at this lib https://github.com/victorvoid/dinoql, you can handle your local data/JSON, as if they were a GraphQL query. I particularly like to write querys using the GraphQL syntax, they are simple to manipulate the data ... using this will also help us to improve the cache rule. The API could look something like this:

```jsx
<ClayDataProvider
  link="https://clay.dataprovider"
  storage={store}
  schema={`
    requests {
      users {
        name
      }
    }
  `}
>
...
</ClayDataProvider>
```

We can adopt as an experimental API (`unstable_schema`) so we can collect feedback, we will have better certainty if this is the right direction.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+		if (storage[SYMBOL_DATA_PROVIDER]) {
+			return storage[SYMBOL_DATA_PROVIDER] as TSymbolData;
+		}
+
+		return (storage[SYMBOL_DATA_PROVIDER] = new LRU(storageMaxSize));
+	};
+
+	let networkStatus = useRef<NetworkStatus>(NetworkStatus.Unused).current;
+
+	let pollingIntervalIdRef = useRef<null | NodeJS.Timeout>(null).current;
```
bryceosterhaus: Since we access `.current`, we should probably remove the `ref` from the variable declaration. I think it makes most sense for all ` **Ref` variables to be the ref object itself.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+
+	let retryDelayIntervalIdRef = useRef<null | NodeJS.Timeout>(null).current;
```
bryceosterhaus: ^

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+
+	const cache = useRef<TSymbolData>(getStorageCache()).current;
```
bryceosterhaus: ^

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+		if (retryDelayIntervalIdRef) {
+			clearInterval(retryDelayIntervalIdRef);
+		}
+	};
+
+	const getRetryDelay = (retryAttempts: number) => {
+		const {
+			delay: {jitter, initial},
+		} = fetchRetry;
+		const baseDelay = jitter ? initial : initial / 2;
```
bryceosterhaus: nitpick: need a new line above this

I believe the general rule of thumb is 
        if ordering matters -> new line between
        if ordering doesn't matter -> new line not needed

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+	};
+
+	const handleFetchRetry = (err: any, retryAttempts: number) => {
+		const {attempts} = fetchRetry;
+
+		cleanRetry();
+
+		if (attempts > 0 && retryAttempts < attempts) {
+			const delay = getRetryDelay(retryAttempts);
+			console.warn(
```
bryceosterhaus: ^

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+				`DataProvider: Trying ${retryAttempts +
+					1} of ${attempts} will happen in ${delay}ms`
+			);
+			retryDelayIntervalIdRef = setInterval(() => {
```
bryceosterhaus: ^

packages/clay-data-provider/src/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+interface IChildrenProps {
+	data: any;
+	error: string | boolean;
+	loading: boolean;
+	networkStatus?: NetworkStatus;
+	refetch: Function;
+}
+
+interface Props extends IDataProvider {
+	children?: (data: IChildrenProps) => React.ReactNode;
```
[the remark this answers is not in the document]
matuzalemsteles: Oh yeah, makes sense. I will do this.

packages/clay-data-provider/src/useDebounce.ts
```diff
@@ -0,0 +1,27 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {useEffect, useState} from 'react';
+
+// Credits to Gabe Ragland
+// (https://dev.to/gabe_ragland/debouncing-with-react-hooks-jci)
+const useDebounce = (value: any, delay: number) => {
```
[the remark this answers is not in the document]
matuzalemsteles: Yeah, I'll add there too.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,284 @@
...
+
+	const getStorageCache = () => {
+		if (storage[SYMBOL_DATA_PROVIDER]) {
+			return storage[SYMBOL_DATA_PROVIDER] as TSymbolData;
+		}
+
+		return (storage[SYMBOL_DATA_PROVIDER] = new LRU(storageMaxSize));
+	};
+
+	let networkStatus = useRef<NetworkStatus>(NetworkStatus.Unused).current;
```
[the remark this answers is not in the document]
matuzalemsteles: Probably not, even if it does not have some value in `useRef` it will always return` .current`. But I'm not sure about that, in case just modifying `.current` directly.

packages/clay-data-provider/src/useResource.ts
matuzalemsteles: Hm yeah, makes sense.

matuzalemsteles:
@bryceosterhaus, updated, I also added tests and documentation to the migration guide. But still need to add the documentation on next.clayui.com, I'll work on this in another PR.

wincent:
TL;DR: (Just some comments on the `useResource` hook.) Can we do a pass over this considering naming, what things should actually be functions, and places where making function arguments explicit could improve readability?

---

I notice that "evaluate" is misspellled as "evalute" in a few places in the `useResource` hook, but I wonder why we're using the verb "evaluate" at all: all code "evaluates" something, so I feel like these functions could have more descriptive names in many cases that better describe what they actually do. For example, `evaluteSetCache()` might be better as `setCache()` (or `cacheSet()`).

Some of them, I am not even sure why they are functions. For example: why is `evaluteCachePolicy` a function? All it does is look at the `fetchPolicy` variable that's in scope; if you're putting it into a function so that it can be lazily evaluated seems like overkill because computing it is basically free. 

In other words, why this?

```javascript
const evaluteCachePolicy = () => {
	switch (fetchPolicy) {
		case FetchPolicy.CacheFirst:
		case FetchPolicy.CacheAndNetwork:
			return true;
		case FetchPolicy.NoCache:
		default:
			return false;
	}
};
```

Instead of this?

```javascript
const shouldUseCache = (
        fetchPolicy === FetchPolicyCacheFirst ||
        fetchPolicy === FetchPolicy.CacheAndNetwork
);
```

Also, I think that readability is impaired by having all these functions operate on values that they get implicitly from their scopes as opposed to be being passed in explicitly as parameters. For example, in `evaluateUrl`, which is only used in one place, takes no parameters but relies on `fetchOptions`, `link`, and `variables`. It has two `as string` casts which might not be necessary because we discard the refinement that we do in `handleFetch` (ie. the `typeof link` -> `'string'` check that it does) before calling it. It may be that by inlining it you don't need those casts any more.

Another one which I am not sure why it is a function is `getStorageCache()`. It is called exactly once. Why isn't this?

```javascript
const getStorageCache = () => {
	if (!storage[SYMBOL_DATA_PROVIDER]) {
		storage[SYMBOL_DATA_PROVIDER] = new LRU(storageMaxSize);
	}

	return storage[SYMBOL_DATA_PROVIDER] as TSymbolData;
};

const cache = useRef<TSymbolData>(getStorageCache()).current;
```

Just this?

```javascript
if (!storage[SYMBOL_DATA_PROVIDER]) {
        storage[SYMBOL_DATA_PROVIDER] = new LRU(storageMaxSize);
}

const storageCache = storage[SYMBOL_DATA_PROVIDER];
const cache = useRef<TSymbolData>(getStorageCache()).current;
```

`cacheSatisfies()` seems a bit of a misleading name. To me, it sounds like it tells me whether a value is present in the cache, but what it actually does is give me a cache key that I can use to look something up in the cache. If that's the case, why isn't it called `getCacheKey()`? And there again, if it took explicit parameters (`link` and `variables`) then I could see at a glance what the determinants of the cache key are, but by passing them implicitly via scope that is hidden from me. Another gotcha with `cacheSatisfies()` is that it uses `JSON.stringify()` to compute the cache key, which means that it is sensitive to the insertion order of properties in the `variables` object, and given that the `useResource` hook can't (or rather shouldn't) rely on variables objects always being constructed in the same way, this could lead to spurious cache misses — so we might want to use a "stable" stringify method instead, that produces the same output regardless of how the input object was constructed.

`getCache()` sounds like it should return an instance of a cache, but what it actually does is (try to) read something out of it. Here a better name might be `readFromCache()`, `lookup()` or something else that you come up with.

`performFetch()` is a fairly descriptive name (although with some caveats; see below), but `performCache()` isn't clear. What does it mean to "perform" a cache? Rather than trying to come up with a good name for this (hard), I would just inline it — it is only used in one place (`performFetch()`) — and it will probably be come much more readable.

The reason I note a caveat with the `performFetch()` name is that it makes it sound like it does a fetch. But it may not. The method which actually does the fetching is `handleFetch()`, which itself sounds like it handles the *result* of a fetch rather than actually initiating it. Naming is hard, but I'd like to think we can find some better names here: it may be that we keep `performFetch()` even though it's not perfect, but rename `handleFetch()` to something more dumb and descriptive, like `getFetchPromise()`. Or maybe `performFetch()` becomes `maybeFetch()` and `handleFetch()` becomes `actuallyFetch()`, `reallyFetch()` or `doFetch()`. Like I said, naming is hard, but maybe one of those suggestions gives you a possible idea. If in doubt, I think "dumb" and descriptive names are best.

In general, I think a good heuristic to use here is to look at all the functions which are called from exactly one place — there are quite a few of them — and ask whether factoring them out like that is actually beneficial, or whether it has had the opposite effect because it requires us to come up with a non-confusing name for them (and naming is hard).

(Tangent: One idea I'd explore to see if it makes the code cleaner is to wrap some of this up in a `cache` object (that wraps what is currently called `cache` in the current version of this file) and then use calls to `cache.get()` and `cache.set()` everywhere, with all of the "cache policy" checks actually buried inside `get()` and `set()`, so that maybe the rest of the code can be written as though it always goes through the cache.)

Finally, the whole `if (!firstRenderRef)` is a bit scary and I'm not sure I understand it. Can you explain what's going on there?

matuzalemsteles:
hey @wincent thank you very much for the comments, I went ahead and made some modifications, I separated the logic from the cache to the `useCache` hook, I think this way it would be more readable I hope. I also renamed the `handleFetch` method to `doFetch`, `performFetch` to `maybeFetch` and changed some other methods.

> Another gotcha with cacheSatisfies() is that it uses JSON.stringify() to compute the cache key, which means that it is sensitive to the insertion order of properties in the variables object, and given that the useResource hook can't (or rather shouldn't) rely on variables objects always being constructed in the same way, this could lead to spurious cache misses — so we might want to use a "stable" stringify method instead, that produces the same output regardless of how the input object was constructed.

I had forgotten about this detail, I ended up adding the `fast-json-stable-stringify` lib to be able to do this in place.

> (Tangent: One idea I'd explore to see if it makes the code cleaner is to wrap some of this up in a cache object (that wraps what is currently called cache in the current version of this file) and then use calls to cache.get() and cache.set() everywhere, with all of the "cache policy" checks actually buried inside get() and set(), so that maybe the rest of the code can be written as though it always goes through the cache.)

I ended up going this way by creating the `useCache` hook and I think it separates logic more and leaves less noise from functions in `useResource` hook.

> Finally, the whole if (!firstRenderRef) is a bit scary and I'm not sure I understand it. Can you explain what's going on there?

This is an attempt to check when the first render is, I'm still not sure with it, I had to do this because two requests were happening when the component is mounted. I tried to have only a `useEffect` but I still need both of them to make the first request when the component is rendered and the next ones when the `variables` change and network status have to be `refetch` as soon as the first request will be `loading`.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,236 @@
...
+				return null;
+		}
+
+		timeout(fetchTimeout, promise)
+			.then(fetchOnComplete)
+			.catch(err => handleFetchRetry(err, retryAttempts));
+	};
+
+	const maybeFetch = (status: NetworkStatus) => {
+		const cacheData = cache.get();
```
matuzalemsteles: @wincent this may seem a little strange because it is not going to `key` here, but I ended up implicitly leaving the key in `useCache` because we can have the cache disabled and I did not want to have to use `stringify` at all times unnecessarily that a variables change.

packages/clay-data-provider/src/types.ts
```diff
@@ -0,0 +1,178 @@
...
+	/**
+	 * This API is used in conjunction with variables API, if it is always
+	 * changing its value set a debounce time to make a new request.
+	 *
+	 * Set a value in ms.
+	 */
+	fetchDelay?: number;
+
+	/**
+	 * Set ups the request options.
```
bryceosterhaus: "Set ups" doesn't make much sense here. Maybe we could do "Options passed to request configuration"

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,236 @@
...
+const getQueryParams = (variables: any): string | null => {
+	if (!variables) {
+		return null;
+	}
+
+	const keys = Object.keys(variables);
+
+	return keys
+		.map((key, index) => {
+			const param = `${key}=${variables[key]}`;
```
bryceosterhaus: Do we need to escape `variables[key]`? Thinking this might lead to xss issues.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,236 @@
...
+		}
+	};
+
+	const getRetryDelay = (retryAttempts: number) => {
+		const {delay = {}} = fetchRetry;
+		const {initial = 300, jitter = true} = delay;
+
+		const baseDelay = jitter ? initial : initial / 2;
+
+		let value = Math.min(baseDelay * 2 ** retryAttempts);
```
bryceosterhaus: Shouldn't `Math.min` take at least 2 arguments?

matuzalemsteles: Oh yeah, I would add a minimum time that retry can reach, I just forgot to remove it. Thanks

packages/clay-data-provider/src/useResource.ts
matuzalemsteles: Hm yeah, I'll go ahead and use `metal-uri` since it handles some things, so it will be easier than implementing some logic here.

matuzalemsteles:
@bryceosterhaus update.

packages/clay-data-provider/src/__tests__/index.tsx
```diff
@@ -0,0 +1,316 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import DataProvider from '../';
+import LRU from 'lru-cache';
+import React from 'react';
+import {cleanup, render, wait} from 'react-testing-library';
+import {FetchMock} from 'jest-fetch-mock';
```
bryceosterhaus: Is this even needed? I noticed it wasn't called anywhere and the fetchMock global is already instantiated in `setupTests.ts`

packages/clay-data-provider/src/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+interface IChildrenProps {
+	data: any;
+	error: string | boolean;
+	loading: boolean;
+	networkStatus?: NetworkStatus;
+	refetch: Function;
+}
+
+interface Props extends IDataProvider {
+	children: (data: IChildrenProps) => React.ReactElement;
```
bryceosterhaus: Even though this prop is an internal react one, I think it might be helpful to add a JSDoc to explain why its declared here, whether that be because you need a certain type of child or if you want to ensure its required.

packages/clay-data-provider/src/types.ts
```diff
@@ -0,0 +1,182 @@
...
+	 * @example
+	 * {
+	 *   name: 'Matu',
+	 * }
+	 *
+	 * output for the link:
+	 * '/?name=Matu'
+	 */
+	variables?: TVariables;
+}
```
bryceosterhaus: After seeing all these types, I'm wondering if we should consider some rules in naming. I noticed we will do `Props`, `FetchPolicy`, `IDataProvider`, `TVariables`. I like the `I**` and `T**` scheme, wonder if it'd be best to just enforce this across clay, which means we would need to change `Props` to `IProps`

packages/clay-data-provider/src/useCache.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import LRU from 'lru-cache';
+import stringify from 'fast-json-stable-stringify';
```
bryceosterhaus: Whats the reason for bringing in this dependency vs JSON.stringify?

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,228 @@
...
+	const handleFetchRetry = (err: any, retryAttempts: number) => {
+		const {attempts = 5} = fetchRetry;
+
+		cleanRetry();
+
+		if (attempts > 0 && retryAttempts < attempts) {
+			const delay = getRetryDelay(retryAttempts);
+
+			// eslint-disable-next-line no-console
+			console.warn(
```
bryceosterhaus: Can we use the `warning` package here since we have been using that in other packags?

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,228 @@
...
+					1} of ${attempts} will happen in ${delay}ms`
+			);
+
+			retryDelayIntervalId = setInterval(() => {
+				doFetch(retryAttempts + 1);
+			}, delay);
+		} else {
+			setNetworkStatus(NetworkStatus.Error);
+			// eslint-disable-next-line no-console
+			console.error('DataProvider: Error making the requisition', err);
```
bryceosterhaus: Should we throw an error here instead or is this more of a warning? If its just a warning, lets use the `warning` package.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,228 @@
...
+		}
+	};
+
+	const cleanPoll = () => {
```
bryceosterhaus:
```jsx
const cleanPoll = () => pollingIntervalId && clearInterval(pollingIntervalId);
```

packages/clay-data-provider/src/useResource.ts
```diff
@@ -0,0 +1,228 @@
...
+		if (pollingIntervalId) {
+			clearInterval(pollingIntervalId);
+		}
+	};
+
+	const fetchOnComplete = (result: any) => {
+		// Should clear retry interval if any of the
```
bryceosterhaus: +1 for comments

packages/clay-data-provider/src/__tests__/index.tsx
```diff
@@ -0,0 +1,316 @@
+/*global fetchMock*/
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import DataProvider from '../';
+import LRU from 'lru-cache';
```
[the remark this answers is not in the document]
matuzalemsteles: hey @bryceosterhaus actually need, this was just for internal testing. We are [using here](https://github.com/liferay/clay/pull/1977/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-2c5fc8804dfb823c64d0915a3e5b725eR29).

packages/clay-data-provider/src/__tests__/index.tsx
matuzalemsteles: Yeah, this is actually a hack, for some reason typescript is not recognizing the typos of `setupTests` and was crashing at compile time due to` fetch` mocks.

packages/clay-data-provider/src/types.ts
matuzalemsteles: Yeah I was thinking this today, we're naming our types in several ways and it gets a bit confusing. I also like it, so Props was assuming it was a default, but we can leave `IProps` too.

packages/clay-data-provider/src/useCache.ts
matuzalemsteles: This was a problem pointed out by Greg, because we used to validate the cache, as we are using the `variables` to check if the cache satisfies, the order of the object can influence this check, so this lib helps with that.

packages/clay-data-provider/src/useResource.ts
matuzalemsteles: For this case I think not, because I just want to give a warning and only then I will throw an error if this persists. The `warning` package will throw an error.

packages/clay-data-provider/src/useResource.ts
matuzalemsteles: I think we can use the `warning` package here, we may have some confusion with this package, in fact it throws an `error`.

packages/clay-data-provider/src/index.tsx
matuzalemsteles: Yeah, I'll take care of adding that.

packages/clay-data-provider/src/types.ts
matuzalemsteles: I would keep `FetchPolicy` and `NetworkStatus` if it is an `enum`.

wincent: A pattern I've seen used pretty successfully is:

- No prefixes by default. Used for things that are "concrete", known types, which is usually what you're dealing with. This keeps thing simple.
- `T` and `T*` for generic types.
- `I*` for interfaces (should be rare).

matuzalemsteles: I see that a lot in codebase too. I like to use `I*` for interfaces is that it separates the TypeScript information from a variable with the same name and helps to identify the type. One thing I like about Flow is that we can make it explicit in the import that we are importing the type, I think for these cases it works very well.

```javascript
import type {TypeName} from '...';
```

bryceosterhaus: That pattern sounds good to me. I think there are probably some eslint + ts rules we can take advantage of here as well.

packages/clay-data-provider/src/useCache.ts
bryceosterhaus: ah gotcha, we actually had that same issue on Faro awhile back and did the same thing. Had forgotten about it till now.

packages/clay-data-provider/src/types.ts
matuzalemsteles: Oh, it would be nice to add eslint + ts rules to this, is there something we can set for nomenclature? I think we can already add that in #2022.

matuzalemsteles:
hey @bryceosterhaus I did the rebase and updated some things that was giving error in lint.

PR #1978, opened 2019-05-10
Fixes #1866 | Create stories for clay-navigation-bar

### LINKED ISSUE #1866: Create NavigationBar component

diegonvs:


packages/clay-navigation-bar/src/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+				'navbar-underline',
+				'navigation-bar',
+				{
+					'navigation-bar-light': !inverted,
+					'navigation-bar-secondary': inverted,
+				}
+			)}
+		>
+			<div className="container-fluid container-fluid-max-xl">
+				<ClayLink
```
matuzalemsteles: hey @diegonvs, this should probably be a `ClayButton` or `button`, since we are not passing `href` and we are controlling your click. What do you think?

diegonvs: Using ClayButton the element looks more dimmed. I consider ClayLink more suitable for the trigger, even element isn't the more semantic for it.

What do you think @pat270?

With ClayButton, tested with all displayTypes:
Screen Shot 2019-05-13 at 14 41 01

With ClayLink
Screen Shot 2019-05-13 at 14 41 13

pat270: @diegonvs You can use `<button class="btn btn-unstyled nav-link">` as long as ClayButton supports it. Those classes should make it look the same as the link.

PR #1979, opened 2019-05-10
Delete flags-ta-IN.flag from clayui.com folder

matuzalemsteles:
hey @pat270 do we need this file format?


pat270:
@matuzalemsteles no we don't. I was assumed it was an SVG.

matuzalemsteles:
ok, thanks @pat270

PR #1980, opened 2019-05-10
Proposal: Use backticks everywhere for quotes

bryceosterhaus:
Wanted to throw this out there and see what you guys thought. I like the idea of using backticks everywhere because it is essentially one less thing to worry about. By using them everywhere there is no need to jump in the mental mindset of if we need to concat a string or not. I don't really see any downsides to this, but I am open to see what you guys think.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,6 +28,7 @@
 		],
 		"no-unused-vars": 0,
 		"prefer-template": 2,
+		"quotes": [2, "backtick"],
 		"react-hooks/rules-of-hooks": "error",
 		"react/jsx-boolean-value": 2,
 		"react/jsx-handler-names": 2,
--- a/babel.config.js
+++ b/babel.config.js
@@ -12,23 +12,23 @@ module.exports = api => {
 			production: {
 				plugins: [
 					[
-						'react-remove-properties',
+						`react-remove-properties`,
 						{
-							properties: ['data-testid'],
+							properties: [`data-testid`],
 						},
 					],
 				],
 			},
 		},
-		ignore: ['node_modules', '**/__snapshots__/**', '**/__tests__/**'],
+		ignore: [`node_modules`, `**/__snapshots__/**`, `**/__tests__/**`],
 		plugins: [
-			'babel-plugin-dev-expression',
-			'transform-inline-environment-variables',
+			`babel-plugin-dev-expression`,
+			`transform-inline-environment-variables`,
 		],
 		presets: [
-			'@babel/env',
-			'@babel/preset-react',
-			'@babel/preset-typescript',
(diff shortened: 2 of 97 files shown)


wincent:
I'm not much of a fan of backticks-for-everything:

- In an ideal world, I'd like to write JS the same way in all Liferay projects both inside and outside portal, but I don't foresee us ever being able to switch to backticks in portal (because it has no analog in Java, for one thing).
- Based on my sampling of projects out in the ecosystem, I get the impression that very few projects have adopted the pattern of backticks-for-everything (yet... maybe they will in the future). Ideally, we'd try to write the "idiomatic JS" that we can, whatever "idiomatic" happens to mean at a given time (in the absence of robust countervailing reasons not to).
- I feel like we're trading the "no need to jump in the mental mindset of if we need to concat a string or not" for the one of "am I in the Clay repo or not".

Having said that, I value consistency and automated enforcement of whatever convention we choose, so I could live with backticks if I had to.

julien:
Using backticks for everything seems a bit like an overkill for me since they aren't always needed.

It might be irrelevant but when I see backticks I expect that something is going on: concatenation, multiline string, etc...

```javascript
const foo = `Hello`; // Very personal comment, but IMO this looks "uglier" than using quotes
const bar = 'Hello there';
const baz = "Hello again";
```

I also think that it's difficult to apply the rule "use pattern X everywhere", in some situation it just makes things more confusing and harder to maintain. (for e.g. Would you use arrow functions everywhere? I certainly wouldn't).

But if that's how we want to do it, I'm OK.

matuzalemsteles:
hey guys, I also do not have a strong opinion on this, but I agree with @julien and @wincent about this, I'd find it super weird if I saw this in an import for example:

```js
import foo from `foo`
```

But I also see the use of the backticks to be used only when it is needed. But I also do not see problem if we go with that.

bryceosterhaus:
I can understand the concerns about it feeling weird or looking off. Part of me thinks we feel that way because of the order in which the features were added to JS. We are so familiar with quotes and then the addition of back-ticks makes us think they are separate. I also sort of think its weird, but when I take a step back I really think its just not being comfortable with it yet. Its sort of like getting used to using prettier, initially I wanted to fight some of the rules but after using it for a few weeks I found all my previous issues trivial.

However, I think @wincent raises a good point in that its not standard across projects and might be weird to use the mental space of jumping in and out of clay. One way to alleviate this is to just continue using single quotes yourself and then rely on tooling to auto-format it, this was how I intended to use it. **But IMO, this is a good reason to not use back-ticks everywhere**. I do hope that one day it'll be normal to use back-tick everywhere, but until then we can just use both.

Thanks everyone for the feedback!

PR #1981, opened 2019-05-11
[DO NOT MERGE] ClayCSS Scss documentation

pat270:
I added the Sassdoc module to the test site and started documenting variables, functions, and mixins. This is my initial test to see if this is something worth while. Let me know what you guys think and if we should go a different direction.

I'm not sure how to integrate this with clayui.com yet. It's just one big index.html file with some js and images.

You can view the output at http://pat270.github.io/lexicon/v3.0.0-sassdoc/


matuzalemsteles:
hey @pat270 this is really cool! I've been looking quickly at the documentation (http://sassdoc.com/node/#using-raw-data), we can use the SassDOC API just to generate the raw annotation data, and from that, we'll build the React components to interpret this data.

Maybe we can add a tab in the CSS documentation, (DESCRIPTION / API), in the API tab we can add this documentation.

pat270:
@matuzalemsteles that sounds promising. Sassdocs as of now doesn't have a way of documenting map parameters inside mixins. I had to hack around until I could get something presentable. Maybe we could add our own annotation to generate those. See Map Keys and Value Types below:
mixin-map-parameters

matuzalemsteles:
Hmm, we can check if the sassdoc parser brings this information to us, so it will be easier.

pat270:
@matuzalemsteles ah I see, that makes more sense. Are we good with documenting Clay CSS in this way? I've documented the mixins and functions and want to merge it before we start updating anything.

matuzalemsteles:
hey @pat270 yes, I see no problem in documenting this way. Maybe it's better to send in another PR, you want to give merged on sassdoc's website?

PR #1982, opened 2019-05-13
Fixes #1923 - Create storybook demo for clay-select

### LINKED ISSUE #1923: Create clay-select component for v3.0.0

carloslancha:


packages/clay-select/src/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+				disabled={disabled}
+				id={id}
+				multiple={multiple}
+				name={name}
+				onChange={onChange}
+			>
+				{options.map((option, index) => (
+					<option
+						key={index}
+						selected={option.selected ? true : undefined}
```
matuzalemsteles: Is this condition necessary? would not it be better just to pass `option.selected`?

packages/clay-checkbox/src/index.tsx
```diff
@@ -60,13 +60,10 @@ const ClayCheckbox: React.FunctionComponent<Props> = ({
...
 			{...containerProps}
-			className={classNames(
-				'custom-control custom-checkbox',
-				containerProps.className,
-				{
-					'custom-control-inline': inline,
-				}
-			)}
+			className={classNames('custom-control custom-checkbox', {
+				[`${containerProps.className}`]: containerProps.className,
```
bryceosterhaus: Instead of putting this in an object, I think you can just do 

```js
classNames(
     'custom-control custom-checkbox',
     containerProps.className,
     {
          'custom-control-inline': inline
     }
);
```

packages/clay-select/src/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface Props extends SelectHTMLAttributes<HTMLSelectElement> {
+	/**
+	 * Props to add to the outer most container.
+	 */
+	containerProps?: React.HTMLAttributes<HTMLDivElement>;
+
+	/**
+	 * Flag to indicate if the select is disabled or not.
+	 */
+	disabled?: boolean;
```
bryceosterhaus: You probably don't need to specify this since it already extends from `SelectHTMLAttributes<HTMLSelectElement>` which should contain that.

packages/clay-select/src/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Id attribute of the select.
+	 */
+	id?: string;
```
bryceosterhaus: Same as above, you probably don't need to specify this since it already extends from `SelectHTMLAttributes<HTMLSelectElement>` which should contain that.

packages/clay-select/src/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Flag to indicate if multiple options can be selected.
+	 */
+	multiple?: boolean;
+
+	/**
+	 * Name attribute of the select.
+	 */
+	name?: string;
```
bryceosterhaus: Same as above, you probably don't need to specify this since it already extends from `SelectHTMLAttributes<HTMLSelectElement>` which should contain that.

packages/clay-select/src/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+				disabled={disabled}
+				id={id}
+				multiple={multiple}
+				name={name}
+				onChange={onChange}
+			>
+				{options.map((option, index) => (
+					<option
+						key={index}
+						selected={option.selected}
```
bryceosterhaus: Rather than having `option.selected`  how do you feel about `selectedValue` as a prop on `ClaySelect`? And then in this arg we would have `selected={option.value === selectedValue}`?

I did something like this with Radio, https://github.com/liferay/clay/blob/develop/packages/clay-radio-group/src/index.tsx#L110

I'd be curious too if React specifically recommends a paradigm for this.

packages/clay-select/src/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Label of the select.
+	 */
+	label: string;
+
+	/**
+	 * Flag to indicate if multiple options can be selected.
+	 */
+	multiple?: boolean;
```
bryceosterhaus: Same as above, you probably don't need to specify this since it already extends from `SelectHTMLAttributes<HTMLSelectElement>` which should contain that.

matuzalemsteles:

matuzalemsteles:
Resent #2051

PR #1995, opened 2019-05-17
WIP: Create ClayAutocomplete and add experimental <ClayForm.Input />

### LINKED ISSUE #1939: Migrate clay-autocomplete to React

matuzalemsteles:
hey @bryceosterhaus this is an draft of how I'm creating Autocomplete, it still depends on ClayDataProvider (#1977) to enter the branch develop.

What I'm thinking more or less for ClayMultiSelect:

> I left some comments in the example.
```jsx
<ClayMultiSelect
	onValueChange={setValue}
	value={value}
>
	{({selectedItems, onLabelClick, onItemClick, onValue, value}) => (
		{/* The prop component is to replace the markup of autocomplete by the MultiSelect, */}
		{/* they are different the same happens to the LoadingIndicator, you may have */}
		{/* to do something with the Input as well. */}
		<ClayAutocomplete component={ClayMultiSelect.AutocompleteMarkup}>
			{/* This component is just an abstraction of using the selectedItems.map, */}
			{/* I think it's more beautiful like that, but it's just for examples. */}
			<Map
				data={selectedItems}
				keyExtractor={item => item.label}
				render={item => (
					<ClayLabel  closeButtonProps={{ onClick: () => onLabelClick(item) }}>{item.label}</ClayLabel>
				)}
			/>
			<ClayForm.Input
				onChange={event => onValue(event.target.value)}
				value={value}
				ref={alignElementRef}
			/>
			<ClayDataProvider
				link="https://rickandmortyapi.com/api/character"
				notifyOnNetworkStatusChange
				variables={{name: value}}
			>
				{({data, error, loading, networkStatus}) => (
					<>
						<ClayAutocomplete.DropDown
							active={(!!data && !!value) || networkStatus === 1}
							alignElementRef={alignElementRef}
						>
							<ClayDropDown.ItemList>
								{/* This is an implementation to make the loading appear only the */}
								{/* first time and only when the request takes 500ms. */}
								<LoadingWithDebounce
									loading={loading}
									networkStatus={networkStatus}
									render={
										<If
											conditional={!error}
											then={
												<Map
													data={data.results || []}
													keyExtractor={item => item.id}
													emptyComponent={
														<ClayDropDown.Item className="disabled">
															{'No Results Found'}
														</ClayDropDown.Item>
													}
													render={item => (
														<ClayDropDown.Item
															onClick={() => onItemClick(item)}
														>
															{item.name}
														</ClayDropDown.Item>
													)}
												/>
											}
											else={
												<ClayDropDown.Item className="disabled">
													{'No Results Found'}
												</ClayDropDown.Item>
											}
										/>
									}
								/>
							</ClayDropDown.ItemList>
						</ClayAutocomplete.DropDown>
						{loading && <ClayAutocomplete.LoadingIndicator component={ClayMultiSelect.LoadingIndicatorMarkup} />}
					</>
				)}
			</ClayDataProvider>
		</ClayAutocomplete>
	)}
</ClayMultiSelect>
```

This is just a thought, the use of `component` props are useful to replace the markup of the component. For example:

```jsx
1.
<ClayAutocomplete />

2.
<ClayAutocomplete component={ClayMultiSelect.Autocomplete} />
```

output:
```jsx
1.
<div
	class="input-group"
>
	<div class="input-group-item">...</div>
</div>

2.
<div class="form-control form-control-tag-group">
</div>
```

We may not need `ClayAutocomplete` being declared and we can add within ClayMultiSelect implicitly.

The components in the example of `If` and `Map` are just abstractions, I see that it is more beautiful to look at them and it seems less messy for me... but just examples to demonstrate. Leave your thoughts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
 		"@storybook/react": "^5.0.6",
 		"@types/classnames": "^2.2.7",
 		"@types/jest": "^23.3.5",
+		"@types/lru-cache": "^5.1.0",
 		"@types/react": "^16.8.0",
 		"@types/react-dom": "^16.8.0",
 		"@types/react-test-renderer": "^16.8.0",
--- a/packages/clay-autocomplete/README.md
+++ b/packages/clay-autocomplete/README.md
@@ -1,0 +1,17 @@
+# clay-autocomplete
+
+ClayAutocomplete component
+
+## Setup
+
+1. Install `yarn`
+
+2. Install local dependencies:
+
+```
+yarn
+```
+
+## Contribute
+
+We'd love to get contributions from you! Please, check our [Contributing Guidelines](https://github.com/liferay/clay/blob/master/CONTRIBUTING.md) to see how you can help us improve.
--- a/packages/clay-autocomplete/package.json
+++ b/packages/clay-autocomplete/package.json
@@ -1,0 +1,30 @@
+{
+	"name": "@clayui/autocomplete",
+	"version": "3.0.0",
+	"description": "ClayAutocomplete component",
+	"license": "BSD-3-Clause",
+	"repository": "https://github.com/liferay/clay/tree/master/packages/clay-autocomplete",
+	"engines": {
+		"node": ">=0.12.0",
+		"npm": ">=3.0.0"
+	},
+	"main": "lib/index.js",
+	"module": "src/index.tsx",
+	"jsnext:main": "src/index.tsx",
+	"files": ["lib", "src"],
+	"scripts": {
+		"build": "cross-env NODE_ENV=production babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "cross-env NODE_ENV=production tsc --project ./tsconfig.declarations.json",
+		"prepublishOnly": "npm run build && npm run build:types",
+		"test": "jest --config ../../jest.config.js"
+	},
+	"keywords": ["clay", "react"],
+	"dependencies": {
+		"classnames": "^2.2.6"
+	},
+	"peerDependencies": {
+		"react": "^16.8.1",
+		"react-dom": "^16.8.1"
+	},
+	"browserslist": ["extends browserslist-config-clay"]
+}
--- a/packages/clay-autocomplete/package.json
+++ b/packages/clay-autocomplete/package.json
@@ -20,6 +20,9 @@
 	},
 	"keywords": ["clay", "react"],
 	"dependencies": {
+		"@clayui/drop-down": "^3.0.0",
+		"@clayui/form": "^3.0.0",
+		"@clayui/loading-indicator": "^3.0.0",
 		"classnames": "^2.2.6"
 	},
 	"peerDependencies": {
--- a/packages/clay-autocomplete/src/Context.ts
+++ b/packages/clay-autocomplete/src/Context.ts
@@ -1,0 +1,27 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+
+interface IContext {
+	/**
+	 * Status of a data fetch is happening, this will help
+	 * <ClayAutocomplete.Input /> add necessary markups.
+	 */
+	loading: boolean;
+
+	/**
+	 * Changes the loading status and causes a new rendering.
+	 */
+	onLoadingChange: (loading: boolean) => void;
+
+	/**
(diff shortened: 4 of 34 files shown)


bryceosterhaus:
For the `<Map>` and `<If>` components, I think those are a bit overkill and will require more overhead from us for maintaining those. I would rather just stick with vanilla jsx such as `{ifCondition && <div />}` or `{items.map(() => <div />)}`.

As for the render props method of composing components, I'm not totally against it but I do think we should try to avoid it and use hooks instead. Hooks aren't necessarily a 1-to-1 replacement for render props, but I think we can try to be strategic with trying to leverage hooks over render props unless its absolutely necessary to use render props.

Also, when we start to add components like `<If>` and `<Map>` in conjunction with render props, it starts to move so much of the code to the right and we hit the line break fairly quickly. So it starts to look uglier IMO.

matuzalemsteles:
hey @bryceosterhaus I agree with you about `If` and `Map`, this was just a thought to escape some of the noise that vanilla jsx does...

On the use of render props, I also agree that we should offer in place the hooks in those cases, the problem is that not always people will be using Hooks, so we must provide both, but this particular case do not know if it is necessary because still we need to implement a markup, we can even implement the markup and offer the hook, but I think it would be a step further to compose the component and arrive at a result... But we can offer both at the same time for those who use only hooks. What do you think?

bryceosterhaus:
>  the problem is that not always people will be using Hooks, so we must provide both

Hmm, this would concern me if we are aiming to cater to both people who use hooks and ones who do not. From my understanding, the react team is pushing hooks as the way of doing things going forward and I think it'd make sense for us to follow that path.

@jbalsas is providing both with and without hooks a concern for us?

jbalsas:
> @jbalsas is providing both with and without hooks a concern for us?

That depends. If we think it's a reasonable request, we could. If we think hooks is the best way for this, then we can simply advocate for it. We do need to just be careful with these decisions and base them on our mutual agreement on how we would better serve the teams building components on top of ours or using ours. 

If we only want to allow hooks for some things because we think that's the right approach, I think that should be fine. We should attempt to guide and educate teams in good usage and best practices whenever we can.

/cc @wincent

wincent:
Maintaining two parallel implementations seems pretty fraught to me. I'd start with hooks and if and when a use case comes up that is not adequately met by them, then I'd want to consider our options.

matuzalemsteles:
My concern is that we may be forcing the use of hooks for this, as far as I know the React team is not forcing this or killing the `class` components are optional ways of working with React, maybe hook is the most victorious here , but I do not know if I agree if we should force people to use it.

I do not know if the cost will be higher, I'm taking the DataProvider as an example, all its logic is over `useResource` and the component uses render props to be able to provide another way to use it, so the maintenance is all over the hook. I may be wrong as to other use cases, but that is what I would do to distribute the two.

I think we can use our communication channels (documentation, blog) to highlight the use of hooks and educate devs to use good practices, I think of keeping both because we can work with developers carefully, we have many people leaving from Metal to React soon into the Portal, Hook is a very different paradigm for some people. So I'd take it easy with that.

bryceosterhaus:
>  React team is not forcing this or killing the class components are optional ways of working with React

From my understanding, hooks does not mean that you can no longer use class components. If we focus using hooks, users can always take those and wrap them in a component to compose a render props component or something that fits their paradigm. We will be locking users into using react v16.8+ but I think that is an okay requirement to have.

I agree with @wincent, I think we should start primarily with hooks, and then for some reason if there is an edge case that doesn't work for someone, we can investigate and determine if it'd be worth maintaining another delivery method for that specific use case.

Also, see [this](https://reacttraining.com/blog/using-hooks-in-classes/) for reference on using hooks in class components. I really don't think hooks should be too much of an issue for consumers.

matuzalemsteles:
> Also, see this for reference on using hooks in class components. I really don't think hooks should be too much of an issue for consumers.

Hmm, that's a pretty interesting dribble.

> We will be locking users into using react v16.8+ but I think that is an okay requirement to have.

Yeah, I guess that would not be a problem, anyway I think it's already a requirement for people to use this version to use Clay because of internal hooks.

> I agree with @wincent, I think we should start primarily with hooks, and then for some reason if there is an edge case that doesn't work for someone, we can investigate and determine if it'd be worth maintaining another delivery method for that specific use case.

I think I'll get on with it. I just want to make sure we're not forcing people to use hooks even though the world is already doing 🙂

matuzalemsteles:
hey guys thanks for the feedbacks, I'll close this and send it back in another PR.

Issue #1495, opened 2019-01-31
Remove non-default export from components


matuzalemsteles:
See the discussions at #1453 for context.

This is a list of components that we need to add warnings when someone is using non-default export from components.

- [ ] clay-alert
- [ ] clay-autocomplete
- [ ] clay-badge
- [ ] clay-button
- [ ] clay-card-grid
- [ ] clay-charts
- [ ] clay-checkbox
- [ ] clay-collapse
- [ ] clay-component
- [ ] clay-data-provider
- [ ] clay-dataset-display
- [ ] clay-icon
- [ ] clay-label
- [ ] clay-link
- [ ] clay-list
- [ ] clay-loading-indicator
- [ ] clay-modal
- [ ] clay-multi-select
- [ ] clay-navigation-bar
- [ ] clay-pagination-bar
- [ ] clay-pagination
- [ ] clay-portal
- [ ] clay-progress-bar
- [ ] clay-radio
- [ ] clay-select
- [ ] clay-sticker
- [ ] clay-table
- [ ] clay-tooltip

jbalsas:
Hey, since we're moving on with `3.x`... should we even try to take care of this for the `2.x` branch? Shouldn't we simply apply this for the `3.x` branch directly?

matuzalemsteles:
hey @jbalsas, I do not know if I understand. But the idea is to add the warnings to `2.x` which are components of Metal.js and remove in later versions of `2.x`. And `3.x` are we already removing with the React components or should we leave `2.x` as it is and move on only with `3.x`?

jbalsas:
That's what I was thinking... we should never remove the default export sin the `2.x` branch. As moving to `3.x` is already a clear breaking change and will be signaled both by the version and maybe the package scope... I'm trying to think if warning about this in `2.x` is even worth it.

So yeah, what I'm trying to say is, we could just leave `2.x` as is and properly apply the new pattern for `3.x` only. 🤷‍♀️

matuzalemsteles:
ok, makes sense. hey @wincent, do you have any opinion on this?

wincent:
I think there's not much point in *removing* the existing non-defaults in the 2.x.  (Removing from the generator is probably fine though, and we already did that in #1453). But it *may* be worth adding the deprecation warnings when they're used, which is what I think you created this PR for.

[...]

PR #2017, opened 2019-05-22
Add ADR for component offering

bryceosterhaus:
@jbalsas here is what we had talked about earlier. Let me know if I missed anything.


bryceosterhaus:
Not sure `component structure` is the best way to describe this but I couldn't think of anything better at the time.

adr/component_structure.md
```diff
@@ -0,0 +1,102 @@
...
+
+```jsx
+<ClayDropDownWithPagination
+	delta={7}
+	header="Favorites"
+	items={items} // array of objects with `label` and `href`
+	message={'Showing {0} of {1} Structures'}
+	title="Folder"
+/>
+```
```
jbalsas: For the sake of fairness, I think all here should be mapped correctly to a corresponding composable example. I think that way it would be easier to spot the possible benefit or drawbacks of each approach for different cases.

```jsx
<ClayDropDown>
        <ClayDropDown.Title>Folder</ClayDropdown.Header>

 	<ClayDropDown.ItemList header="Favorites">
		{items.filter((item, i) => i < delta).map((item, i) => (
			<ClayDropDown.Item href={item.href} key={i}>
				{item.label}
			</ClayDropDown.Item>
		))}
	</ClayDropDown.ItemList>

 	{items.length > delta && 
 	 	<ClayDropDown.Pagination message="Showing {0} of {1} Structures">
 	 	 	 	<ClayButton label="More" onClick="{...}" />
 	 	</ClayDropDown.Pagination>
 	}
</ClayDropDown>
```

adr/component_structure.md
```diff
@@ -0,0 +1,102 @@
...
+
+	<ClayDropDown.ItemList>
+		{items.map((item, i) => (
+			<ClayDropDown.Item href={item.href} key={i}>
+				{item.label}
+			</ClayDropDown.Item>
+		))}
+	</ClayDropDown.ItemList>
+</ClayDropDown>
+```
```
jbalsas: Similarly to the example below. Wouldn't it be interesting to show how we think this exact same component feels like on high-level? That way we can compare and maybe come up with clearer guidelines on what would work and what wouldn't

adr/component_structure.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Consequences
+
+Benefits:
+
+-   Gives users high-level components to import and use right away.
+-   Gives users low-level components to build more complex interactions.
+-   Gives maintainers leverage by encouraging users to build their own custom components built on our building blocks rather than baking it into one of our high-level versions.
+
```
jbalsas: - Increased consistency for high-level components usage
- Reduced support costs (less versions of the same thing)

adr/component_structure.md
```diff
@@ -0,0 +1,102 @@
...
+Drawbacks:
+
+-   More lines of code to maintain.
```
jbalsas: - Possible lock-in scenarios when using high-level components. Teams might need to move from a high-level component to a composition scenario in case they need to add something not already covered
- Increased uncertainty... what approach should a developer use and when?

adr/component_structure.md
```diff
@@ -0,0 +1,102 @@
+# Create both low-level and high-level components that are exported from each package.
+
+## Context
+
+The primary issue we are facing right now is determining how "high-level" the component should be. For example, should we create a single component with many props, or should we create many components to be composed together.
```
jbalsas: Our 2 main goals with Clay3 are:
- Provide enough flexibility and low-level components for teams to quickly build and deliver applications without being blocked by us
- Provide quality high level components that allow teams to quickly build applications with the best possible interactions (UX) that are consistent, performant, accessible and extensible without the need of repeating themselves over and over again.

adr/component_structure.md
bryceosterhaus: I probably wouldn't include that here because it'd be too difficult to condense into an example without adding all of the business logic. Plus this example isn't to prove a point that composable is better, its just to point out that you are able to compose components together.

bryceosterhaus:
@jbalsas good feedback, I went ahead and updated the doc

adr/component_structure.md
jbalsas: Ok, as long as that's clear I'm fine with it 👍

adr/component_offerings.md
```diff
@@ -0,0 +1,134 @@
...
+
+	<ClayDropDown.Caption>
+		{`Showing ${page * delta} of ${items.length} structures`}
+	</ClayDropDown.Caption>
+
+	<ClayDropDown.Action onClick={() => setPage(page + 1)}>
+		{'Show More'}
+	</ClayDropDown.Action>
+</ClayDropDown>
+```
```
wincent: There are some other patterns that we could consider (or at least list) too. Given some low-level components `<Foo.A>`, `<Foo.B>` etc, depending on how many high-level embodiments we wish to offer, we might want them to be:

- `<FooVariantA />`, `FooVariantB />` etc: ie. the option you've laid out above.
- `<Foo a={...} b={...} />`: a single high-level component that has a bunch of options (I think this is your `ClayCard` example higher up).
- `<Foo mode={....} />`: some "mode" or "type" enum that sets the high-level behavior (I don't think this one has been mentioned so far).

And there are ultimately so many ways to extend and compose in React (whether they be higher-order components, render props, function children, context wrappers etc), that an analysis of individual cases might suggest a bunch of alternatives.

adr/component_offerings.md
```diff
@@ -0,0 +1,134 @@
...
+
+Finally, both the high-level and low-level components will be exported from each package. The low-level component will be exported as default, but the high-level ones will be exported by name. The high-level components should be put into a
+
+For example:
+
+```jsx
+export {ClayDropDownWithPagination};
+
+export default ClayDropDown;
+```
```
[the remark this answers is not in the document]
julien: If it was an option, I'd avoid using default exports ... 
If I remember correctly we've talked about this several times and it might be a good moment to think about it

https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html
https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/
https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

I'm not saying I agree 100% with everything that's said in the URLs linked above, but I think we should talk about this just to know what we want to go with.

adr/component_offerings.md
julien: As far as I'm concerned, I don't know which pattern is the best but I'd rather focus on small components that can be composed; at least that's what I've been used to do with React.

That said; we should take into account the fact that the teams using these components have been used to Clay's "high level" components; how are we going to "sell" them the fact that they are going to have to write much more code than before to get basically the same result? (Even if they asked for React in the 1st place).

adr/component_offerings.md
```diff
@@ -0,0 +1,134 @@
...
+### What does this look like?
+
+#### Composable:
+
+![Composable Drop Down](./images/component_offerings_2.png)
+
+```jsx
+<ClayDropDown>
+	<ClayDropDown.Search
+		onChange={() => {}}
```
julien: In this example I'm assming that the `onChange` and `value` attributes are optional? 
(Might be totally off topic but I thought I'd ask)

adr/component_offerings.md
```diff
@@ -0,0 +1,134 @@
...
+
+	<ClayDropDown.Action onClick={() => setPage(page + 1)}>
+		{'Show More'}
+	</ClayDropDown.Action>
+</ClayDropDown>
+```
+
+You'll notice that the high-level component handles much of the interactions internally. It will handle the click of the `show more`, paginate, and update language as necessary. In fact, even this high-level component will be made from the smaller low-level building blocks.
+
+Finally, both the high-level and low-level components will be exported from each package. The low-level component will be exported as default, but the high-level ones will be exported by name. The high-level components should be put into a
```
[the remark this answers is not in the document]
julien: I'm not sure if I understand correctly.
Are we saying we're going to have two versions of a component, i.e. a high level version and a low level version?
If that's the case it's sounds wrong to me, as we don't want to maintain two versions of the same thing.

adr/component_offerings.md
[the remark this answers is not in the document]
wincent: > If I remember correctly we've talked about this several times

Yep, that I am aware of least here:

- https://github.com/liferay/clay/issues/1495
- https://github.com/liferay/clay/pull/1453

adr/component_offerings.md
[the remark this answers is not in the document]
matuzalemsteles: @julien I think the previous paragraph explains this better but we will have the high-level components following the Lexicon standards with their behaviors and it is building with the low-level ones (For example, Autocomplete has some minimal low level components for deliver markup and its complete logic is made up of DataProvider, LoadingIndicator, Input... see #1995 for example, but this can all be composed and still opens options to customize my own logic with my use cases and we can still get all this composition and deliver wrapped in a high-level component).

Then the cost will be less to keep the two in my view. Since the high-level are composed of low level, we are only offering this option for people who want something fast with the behaviors provided by Lexicon for example.

adr/component_offerings.md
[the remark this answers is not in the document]
matuzalemsteles: If we banish standard export, we will have to deal with component naming to differentiate a high-level component for a low-level component (I am still assuming we will have these two for each component ClayDropdown, Card...), I not think of good names for differentiate them.

I agree on not using standard export the benefits are numerous but I still see this problem we will have to deal with here.

adr/component_offerings.md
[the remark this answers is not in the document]
julien: @matuzalemsteles thanks for making that clear. 
That doesn't sound like a bad idea or "duplication" to me.

adr/component_offerings.md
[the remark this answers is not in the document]
julien: > If we banish standard export

I don't actually wan't to "banish" them. There'll still probably be some situations in which we'll  need or want to use `export default` ... 

But as a norm/standard, I'd rather do this: 

```javascript
export ClayDropDown;
// ...
export ClayDropDownWithPagination;
```

Than this:

```javascript
export {ClayDropDownWithPagination};
export default ClayDropDown;
```

Once again, I'm just giving my opinion.

> we will have to deal with component naming to differentiate a high-level component for a low-level component 

I agree, naming is (and will be hard) whether we use `export default` or not ;)

adr/component_offerings.md
julien: @matuzalemsteles thanks for the comment.

 > I think our "sell" here for the low-level components are for just cases where the team needs to create other behaviors or customize their components in more detail.

From what I've experienced that has happened with most of the components for most of the teams. (I don't have exact numbers but we could actually look at the closed issues and PRs).

adr/component_offerings.md
bryceosterhaus: These would be required props to make it functional. This gives the user flexibility to do whatever they need when the input value changes.

However, in a high-level component, this would be handled internally.

adr/component_offerings.md
bryceosterhaus: @wincent 
> And there are ultimately so many ways to extend and compose in React (whether they be higher-order components, render props, function children, context wrappers etc), that an analysis of individual cases might suggest a bunch of alternatives.

You're totally right in that there are tons of different ways to accomplish this. From my preliminary brainstorming, I think the first option is the clearest straight forward path and most descriptive for users. I don't really like the "many options" or "mode" or "type" methods due to the amount of branching that ends up happening in the components. For example, if "searchable" is an option, we now have a large chunk of conditional code in a single component. Personally, I prefer having really descriptive components, even if that means we have to reduplicate some code. I find it a lot easier to maintain components that strive to implement a single variant.

@matuzalemsteles 
> it is still unclear in this ADR how the high-level components will be, we will deliver only a single high-level component following the Lexicon standards with their interactions

I would avoid aiming for 'only a single high-level component'. This is going to be harder to decide though and I don't think there is necessarily a clear path on how to decide where we break a component at. I think it might be best to aim each high-level component at a single variation, rather than trying to have multiple functionalities in the same component. For example, `ClayDropDown` might have `ClayDropDownBasic`, `ClayDropDownWithForm`, and `ClayDropDownWithPagination`. There may be some overlap, but I think we can aim each high-level component at a specific use case.

That make more sense?

bryceosterhaus: @julien I agree in that we should focus on maintaining the building blocks, however, we also need the high-level components because we run the risk of many implementations of the same basic functionality. Now if we can also try and make out high-level components simple and straight forward, we may have more components to maintain, but they should also be easier to maintain since we have the core building blocks.

bryceosterhaus:
@wincent I think you bring up a good point in exposing our low-level components and I'm not totally sure I know the best solution for that. I do think by giving access to our low-level components, it relieves pressure for us to do all the leg work for custom implementations, but with that it brings the added weight for us in maintaining and supporting the low-level layer.

adr/component_offerings.md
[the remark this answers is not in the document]
bryceosterhaus: I don't really have any strong opinions on this, `default` for the low-level just felt right at the time 🤷‍♂

bryceosterhaus:
Updated with some of the things noted and tried to clarify

adr/component_offerings.md
julien: @bryceosterhaus 👍 Thanks for making that clear

adr/component_offerings.md
julien: @bryceosterhaus that sounds good to me then.

matuzalemsteles: I think it makes sense to deliver components that deliver specific cases, but I'm not sure we should call them High-level. That's a bit my vision now.

* Intermediate layer are simple components that deliver only specific use cases. Composed of low-level components.
* High-level components are composed of low-level and intermediate layer.

Even if we do not get to deliver high-level, I think we should improve our speech in this sense, we have this intermediate layer following specific cases of the Lexicon component... For example the image below, when I see something high-level I consider the whole component `ClayDropdown` delivering everything the Lexicon expects of that component. They would be our v2 components for example but maybe there will be a higher or no cost to maintain.

Layer Components@2x

bryceosterhaus: @matuzalemsteles yeah I think its totally fine to name it something else, it was just the clearest way to me at the time to differentiate between the two types of components we will build. As far as us communicating this to users, I'm not sure we even need to go into this type of depth with them either. They will just want to know how to use a certain lexicon use case as quickly as possible. The low-level stuff really only gets brought up when they want a more custom implementation.

The main point of this ADR is to provide a reason why developers need to create two sets of components when working on a package, the `<ClayDropDown.Menu>` type components and the `<DropDownWithPagination>` type. Not sure we need to be too concerned with users reading our ADRs.

jbalsas: I agree with @bryceosterhaus here. We don't even to give them names or market them any differently.

I think the consensus and the conclussion here is that we want to follow [atomic design](http://atomicdesign.bradfrost.com/chapter-2/) principles (which is what inspires Lexicon too) and:
- Start off very small composable atoms
- Compose them as necessary to offer molecules, organisms and templates as deemed fit to help developers implement Lexicon patterns.

@bryceosterhaus, what do you think?

matuzalemsteles: I do not mean that users need to read our ADRs, I really did not want to encourage it (but whoever wants to go deeper into Clay is a starting point) when I say we should improve our discourse, it will be how we will present the two levels of components for people, for being something unusual perhaps for some people, especially the intermediary layer.

Since we will no longer have a high-level component (a complete component with all Lexicon definitions) how will we address Taglibs? or is it to address here in that ADR?

jbalsas: Hey @matuzalemsteles, we need to have and provide all levels that make sense to us. That includes what you call high-level components, but we can simply call them components. I don't see the need to differentiate them. 

We just need to add the proper documentation and examples for all of our components and how to leverage them. No need to expose this anymore than that.

matuzalemsteles: hey @jbalsas about following based on `atomic design`, I think that in a way we ended up following these principles making the components available... but I'm not sure if it would apply inside a component, maybe yes. I can not think well of so many variations (molecules, organisms and templates) so within the component...

When I talk about high-level and low-level is something common that some lib deliver, for example Tensorflow.js delivers its low-level layer with [Tensorflow Core](https://github.com/tensorflow/tfjs- core) and its high-level layer with [Tensorflow Layer](https://github.com/tensorflow/tfjs-layers). Enabling what we were initially talking about flexibility for low-level layers and fast delivery to the high-level layers.

I'm just calling it that because it's easier to identify the components 😅 and understand what they actually do. I understand that we should only make them available as components, but the point is that we have to explain why we are doing this and when we explain we will have to separate things to create clarity for people.

bryceosterhaus:
@matuzalemsteles @wincent @jbalsas @julien 

Any other specific changes you think we should make to this ADR? Or any other ADR's that we need to spawn off from this one? Sometimes its a bit difficult to track the convo via GH comments, not sure if I missed anything.

julien:
@bryceosterhaus I don't think we need more changes on this one.

We all seem to think conversations on GitHub are rather hard to track to say the least ;)  
But this was an interesting one. 👍

For the moment there's nothing I can think of, but I'm pretty sure something will pop up soon. 
(To give you an example, some people have raised their doubts about communication between React and non-React components but I'm not sure this fits in the "scope" of clay).

wincent:
@bryceosterhaus: My understanding of the ADR pattern is that you can merge the PR and that doesn't mean the decisions are forever set in stone and the document is immutable; we can update the document as our thinking evolves and we iterate further, right? If that understanding is correct, I don't see any reason why not to merge.

matuzalemsteles:
I think I'm fine with the low-level and high-level definitions for the components. I do not know if it would be necessary but maybe we should create some ADR to discuss delivery to the Taglibs, this is still not clear to me if we follow that decision here.

> (To give you an example, some people have raised their doubts about communication between React and non-React components but I'm not sure this fits in the "scope" of clay).

@julien I think WebComponents would help with this, but React does not work well with them and makes it very difficult to implement it needs to make new life cycles in the component classes so that it can serve the components like WebComponents, but we do not need to think about it now , I did not want to encourage this for people to use these components within the Metal.js application for example, I want we to encourage teams to migrate to React.

hey @wincent, actually I think if we go ahead with this, we can not iterate over it, if we want to go back on these decisions or because we think this is not complete, we have to create another ADR that explains the reasons they will lead overriding the old decision, adding new cases and a new decision, and moving that ADR to a separate folder within the `/adr/old)` folder, this helps keep track of decisions over time and keep a history.

wincent:
> hey @wincent, actually I think if we go ahead with this, we can not iterate over it, if we want to go back on these decisions or because we think this is not complete, we have to create another ADR that explains the reasons they will lead overriding the old decision, adding new cases and a new decision, and moving that ADR to a separate folder within the /adr/old) folder, this helps keep track of decisions over time and keep a history.

That seems like quite a heavy-weight process, then. I can understand why you might want that level of rigor and procedure in a large project with many stakeholders and consumers (extreme example: TC39), but Liferay itself is the principal user of Clay, and I worry that adding too much "red tape" is going to slow us down without a commensurate benefit. Git repos have history, so I'm not sure we need separate documents to show how things have changed over time. Just my 2¢ though.

bryceosterhaus:
@wincent technically that repo would define a good ADR as one that is immutable, but I agree with you and think thats a bit too rigid for us and our use case. I'd be fine with updating them since we have git to help us track the changes, however if for some reason that is a problem in the future we can always pivot to immutable ones. I think especially for our sake, its good to be flexible and we should likely expect things to change as we evolve and understand our process and use case more, even if that means dropping ADRs altogether. Any tooling or rules we follow should be here to help us work well together and work efficiently, if things get in the way of that then we should pivot.

TL;DR: Mutable ADRs and if it becomes an issue then we pivot.

What do you think about that, @matuzalemsteles?

matuzalemsteles:
Yeah, I think I agree with you on this we can test this now and see how it goes and find out how it worked for us.

But just letting my point on git keep history, we can have 100 commits to change in an ADR file, one point said is that you will not know how many versions had this file (probably 100 🙂), for me that's a deeper view I personally would like to see a macro view and find out that it had two versions.

bryceosterhaus:
@matuzalemsteles yeah thats a good point. I'll go ahead and merge now and then we can see how it goes moving forward, I'm definitely flexible on how this works.

And thanks all for the feedback on this! If you have any other ADR ideas, please send them in even if you aren't totally sure about it, I think its a great way to spur discussion.

wincent:
One thing about the use of Git history is that it makes it really easy for me to see what changed, even if there are 100 revisions (which of course, there won't be). For example, I can do `git diff @{2.weeks.ago} -- some-file` to see what happened over a time range, or I can look at the series of changes I'm interested in over and arbitrary time range with `git log --reverse -p --since 2019-03-02 -- some-file`. Those are just some random examples, but the point is that Git is built as a *version control* tool, so it's really good at visualizing changes over time.

Having said that, if we want to keep a cleaner history for this kind of thing that's kind of up to us: we can make a million commits in a PR hashing out details and the merge the entire branch, warts and all, or we can rebase/squash/force-push (etc) the PR before merging to produce whatever final state we desire. If we do that right then that should produce the "macro" level view you're wanting, @matuzalemsteles (ie. one PR for each significant revision to the document, with hopefully not much noise in the form of trivial "fix typo" PRs).

matuzalemsteles:
> Having said that, if we want to keep a cleaner history for this kind of thing that's kind of up to us: we can make a million commits in a PR hashing out details and then merge the entire branch, warts and all, or we can rebase/squash/force-push (etc) the PR before merging to produce whatever final state we desire. If we do that right then that should produce the "macro" level view you're wanting, @matuzalemsteles (ie. one PR for each significant revision to the document, with hopefully not much noise in the form of trivial "fix typo" PRs).

Yeah, I agree about using `squash` for every PR that's merged into an ADR, it probably sounds like a new version.

PR #2007, opened 2019-05-20
Card and CardGrid :)

diegonvs:

PR #2019, opened 2019-05-23
Fixes #1947 | Create ClayTable component

### LINKED ISSUE #1947: Migrate clay-table to React

diegonvs:
- [x] Component
- [x] Tests
- [x] Stories
- [x] Migration guide

- What do think folks about using [Structural testing](https://storybook.js.org/docs/testing/structural-testing/) with Storybook? I think that is another way to cover all our Jest Snapshots :)
- Sometimes, for a better experience in stories, some components with state are needed. So I'm creating some of them inside `fixtures` folder. However we could have a global package for fixtures for using in storybook without publishing them. What do you think?
- Another question is: Would be a good practice cover all possibilities using helper classes on our Composable Components?


packages/clay-table/src/Cell.tsx
```diff
@@ -0,0 +1,100 @@
...
+	const TagName = headingCell ? 'th' : 'td';
+
+	return (
+		<TagName
+			{...otherProps}
+			className={classNames(className, {
+				'table-cell-expand': expanded,
+				[`table-cell-${cellDelimiter}`]: cellDelimiter,
+				[`table-column-text-${columnTextAlignment}`]: columnTextAlignment,
+				'table-focus': headingCell,
```
matuzalemsteles: `table-focus` modifier class is not required in a `th`, it has no effect.

packages/clay-table/src/Cell.tsx
```diff
@@ -0,0 +1,100 @@
...
+				[`text-${align}`]: align,
+			})}
+		>
+			{headingTitle
+				? React.Children.map(children, (child, i) =>
+						React.cloneElement(
```
matuzalemsteles: We do not need to clone the `children` here.

packages/clay-table/src/Cell.tsx
```diff
@@ -0,0 +1,100 @@
...
+	headingCell?: boolean;
+
+	/**
+	 * Applies a style of heading inside a child of table
+	 * header cell element.
+	 */
+	headingTitle?: boolean;
+}
+
+type CellDelimiterType = 'start' | 'end';
```
matuzalemsteles: We have some in common with the type `RowDelimiterType`, we can leave it generic and use in both. Something like `TDelimiter`.

packages/clay-table/src/__tests__/index.tsx
```diff
@@ -0,0 +1,227 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import * as TestRenderer from 'react-test-renderer';
```
matuzalemsteles: I think we can only use `react-testing-library` from now on. What do you think @bryceosterhaus?

packages/clay-table/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+type VerticalAlignmentType = 'bottom' | 'middle' | 'top';
+
+const ClayTable: React.FunctionComponent<Props> & {
+	Body: TableBodyType;
+	Cell: TableCellType;
+	Head: TableHeadType;
+	Row: TableRowType;
+} = ({
+	bodyVerticalAlignment,
```
matuzalemsteles: I'm just a bit reluctant with this bunch of APIs that we have to set up the classes, this is pretty cool but I'm worried because it raises our maintenance layer above the API's, that's a bit worrisome, I'd try to put it on the balance and not need to add everything we have in `classes`. One thing that can help us with that and add only those that the Lexicon provides and `className` will do the rest.

packages/clay-table/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+	hover,
+	noWrap,
+	responsive,
+	responsiveSize,
+	striped,
+	tableVerticalAlignment,
+	...otherProps
+}) => {
+	return (
+		<div
```
matuzalemsteles: I'm not sure but the table does not always need a wrap.

matuzalemsteles:
hey @diegonvs I would also try to narrow down the stories with the full use cases by adding many rows and columns, I think it would be enough to just show the low-level components and maybe a complete storie composing them all. Since we are discussing the high-level in #2017 we will probably have to reorganize our stories to list examples with the low-level and high-level ones.

packages/clay-table/src/__tests__/index.tsx
bryceosterhaus: Yeah I think it'd be a good rule of thumb to opt for always using `react-testing-library` and especially testing any functional parts of the component. I'm growing less and less confident in snapshots since its so easy to just "update all"

packages/clay-table/src/Body.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export type TableBodyType = React.FunctionComponent<BodyProps>;
+
+const Body: React.FunctionComponent<BodyProps> = ({
+	children,
+	className,
+	...otherProps
+}) => {
+	return (
+		<tbody {...otherProps} className={className}>
```
bryceosterhaus: No need to declare className here. It'll already be there with `...otherProps`

packages/clay-table/src/Head.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export type TableHeadType = React.FunctionComponent<HeadProps>;
+
+const Head: React.FunctionComponent<HeadProps> = ({
+	children,
+	className,
+	...otherProps
+}) => {
+	return (
+		<thead {...otherProps} className={className}>
```
bryceosterhaus: no need to declare className here

packages/clay-table/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+type ResposiveSizeType = 'lg' | 'md' | 'sm' | 'xl';
+
+type VerticalAlignmentType = 'bottom' | 'middle' | 'top';
+
+const ClayTable: React.FunctionComponent<Props> & {
+	Body: TableBodyType;
+	Cell: TableCellType;
+	Head: TableHeadType;
+	Row: TableRowType;
```
bryceosterhaus: Rather than export these types from their components you can just do something like

```tsx
Body: typeof TableBody;
Cell: typeof TableCell;
Head: typeof TableHead;
Row: typeof TableRow;
```

packages/clay-table/src/index.tsx
diegonvs: What do you think about it @bryceosterhaus @pat270? Would be a good idea leave some APIs using css utilities on the component?

packages/clay-table/src/__tests__/index.tsx
wincent: Yep. I think a snapshot is mostly only useful as a smoke test (ie. did it render at all?) and as an *informational document of* (not a meaningful assertion about) what the component currently produces.

packages/clay-table/src/Body.tsx
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+
+interface BodyProps
+	extends React.TableHTMLAttributes<HTMLTableSectionElement> {}
+
+export type TableBodyType = React.FunctionComponent<BodyProps>;
```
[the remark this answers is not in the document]
diegonvs: Refactoring to the pattern of #2007

packages/clay-table/src/Cell.tsx
```diff
@@ -0,0 +1,100 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface CellProps extends TableCellBaseProps {
```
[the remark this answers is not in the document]
diegonvs: Sorting

packages/clay-table/src/Cell.tsx
```diff
@@ -0,0 +1,100 @@
...
+	 * first or last cell to maintain table styles on the left and right side.
+	 */
+	cellDelimiter?: CellDelimiterType;
+
+	/**
+	 * Aligns horizontally contents inside the Cell.
+	 */
+	columnTextAlignment?: ColumnTextAlignmentType;
+
+	children?:
```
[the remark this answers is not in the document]
diegonvs: Nope

packages/clay-table/src/Cell.tsx
diegonvs: Yes!

packages/clay-table/src/Cell.tsx
diegonvs: Solved!

packages/clay-table/src/index.tsx
diegonvs: `.table-responsive and .table-responsive-{sm|md|lg|xl} should be placed in a div that wraps the table element. See https://github.com/twbs/bootstrap/issues/24638 for more info.
`

diegonvs: @pat270 apparently this issue was fixed by https://github.com/twbs/bootstrap/pull/25058. Would be a good idea remove this wrap?

diegonvs: https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables

packages/clay-table/src/index.tsx
diegonvs: Fixed!

packages/clay-table/src/Head.tsx
diegonvs: Fixed

packages/clay-table/src/Body.tsx
diegonvs: Fixed

packages/clay-table/src/Cell.tsx
diegonvs: Removed!

packages/clay-table/src/index.tsx
diegonvs: .div wrapper removed!

diegonvs:
I gonna reopen this PR to solve merge conflicts and git history.

diegonvs:
reopened at https://github.com/liferay/clay/pull/2026

PR #2020, opened 2019-05-23
Fixes sidebar scrolling and adding badges and button documents

matuzalemsteles:
hey guys, I went ahead and I'm starting to use the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/) system, which we had discussed in #1712 and by Slack.

This is part of the migration progress of documents spoken in #1903.

PR #2022, opened 2019-05-24
Clean up imports and hooks

bryceosterhaus:
Just a few things thought it'd be good to fix and make consistent.


matuzalemsteles:
Great, thanks for this @bryceosterhaus. Perhaps we could already add in this PR also the normalization of typo nomenclatures? which we talked about in #1977.

Another thing, I started using the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/) system in #2020, I think we can start using here as well.

wincent:
Nice. So has something changed since we discussed the `import *` thing [here](https://github.com/liferay/clay/pull/1548#discussion_r259716545)? Back then it seemed it wasn't going to work without `allowSyntheticDefaultImports`, but now it does?

We should also take a look at our import sorting and spacing as it doesn't seem entirely consistent. I'm seeing some `import` statements with what seem to be unnecessary blank lines between them (eg. packages/clay-navigation-bar/src/index.tsx)  and others with everything together. Likewise, I mostly see things sorted case-insensitively according to the left-most export name, but then I see instances like packages/clay-pagination/stories/index.tsx where that isn't the case (eg. "React" is being sorted before "number"):

```javascript
import ClayPagination from '../src';
import React, {useState} from 'react';
import {number} from '@storybook/addon-knobs';
import {storiesOf} from '@storybook/react';
```

My preferred style would actually be to sort by the module name (ie. the bit in the string) because that is going to be more stable over time. For example, the instance above, if we're sorting by export name and I decided that I only wanted to import `useState` and not `React`, that would change the sort key from "R" to "u", requiring it to be moved after "storiesOf". On the other hand, if we sort by module name, changing the exports that I import from it doesn't change the sort order.

matuzalemsteles:
Hm yeah, good example @wincent. I had not thought about that. I still do not know when we use the blanks in the imports and what would be the best rules for this, to set a pattern. Normally I follow the order of separating local imports from external (`node_modules`).

bryceosterhaus:
> Nice. So has something changed since we discussed the import * thing here? Back then it seemed it wasn't going to work without allowSyntheticDefaultImports, but now it does?

TBH, not sure why this started working and totally forgot we discussed it. Overtime I realized that it just wasn't an issue anymore, perhaps it was fixed in ts when we updated to `3.3.4000`

As for the import ordering, in previous projects I have used [this eslint plugin](https://github.com/marudor/eslint-plugin-sort-imports-es6-autofix). Originally at LR we tried to keep the external deps(`from 'react'`) separate from the internal ones (`from './SomeComponent'`) but we realized it actually didnt matter much. Once we adopted the plugin, we were more comfortable to just let the tool order them for us all at the top of the file. I'm 100% for just letting a tool manage it and not have to worry about it.

> My preferred style would actually be to sort by the module name (ie. the bit in the string) because that is going to be more stable over time.

I'd be fine doing this so long as we can leverage a tool that will automatically sort for us. Do you know of another eslint plugin or rule that might do this for us?

wincent:
> Do you know of another eslint plugin or rule that might do this for us?

Sadly not, but there might be one. I've always just relied on the power of telepathy to sort imports (this is an advanced Vim skill that I learned after 12 years of study). 😂

matuzalemsteles:
I agree to use some tool for this, sometimes I'm wasting time on it 🙂.

> Sadly not, but there might be one. I've always just relied on the power of telepathy to sort imports (this is an advanced Vim skill that I learned after 12 years of study). 😂

Haha, I think we'll take a while...

bryceosterhaus:
@matuzalemsteles @wincent updated. I went ahead and added some rules from `@typescipt/eslint-plugin` to help us. Check out the new commits to see what was changed.

bryceosterhaus:
Gonna go ahead and merge to avoid and conflicts with other PRs since this touches so many files.

@wincent feel free to still make comments and I can submit another PR with any changes needed.

PR #2023, opened 2019-05-27
docs(CONTRIBUTING): Add new Git Commit Messages rules and use of TypeScript

matuzalemsteles:
Please check if there is a misspelling 😁

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,11 +122,8 @@ The header is mandatory and the scope of the header is optional.
 - **feat**: A new feature
 - **fix**: A bug fix
 - **docs**: Documentation-only changes
-- **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
 - **refactor**: A code change that neither fixes a bug nor adds a feature
-- **perf**: A code change that improves performance
-- **test**: Adding missing or correcting existing tests
-- **chore**: Changes to the build process or auxiliary tools and libraries such as documentation generation
+- **chore**: Changes in the build process, auxiliary tools, libraries, tests or formatting
 
 #### Scope
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Want to contribute to Clay? There are a few things you need to know before you s
 
 Here's a set of guidelines to contribute to Clay and its packages. Use your common sense, and feel free to propose changes to this document in a pull request.
 
-For contributions with Clay v2, go to branch `2.x-develop` and follow the [contribution guide](https://github.com/liferay/clay/blob/2.x-develop/CONTRIBUTING.md).
+**For contributions to Clay v2, go to branch `2.x-develop` and follow the [contribution guide](https://github.com/liferay/clay/blob/2.x-develop/CONTRIBUTING.md).**
 
 #### Table of Contents
 
@@ -88,7 +88,7 @@ Suggestions for enhancement or API changes are tracked as [Github issues](https:
 
 This section shows you how to send a pull request. The main staff are always monitoring pull requests, so rest assured that they will review yours. Follow these guidelines so the maintainers can complete a thorough review of your request.
 
-> Since Travis is failing on executing **a11y tests** due to out of date suid you need to execute `npm run pa11y` locally before sending any PR to verify we're following accessibility standards.
+> Be sure to check the accessibility tab of the Storybook to ensure accessibility tests are going pass.
 
 #### Before Sending a Pull Request
 
@@ -107,10 +107,30 @@ This section shows you how to send a pull request. The main staff are always mon
 
 This section shows you how to write commit messages in Clay. Follow these guidelines to help us maintain order and make it easier to locate your changes.
 
-#### Before You Make a Commit Message
+Each commit message consists of a header, a body and a footer. The header has a special format that includes a type, a scope and a subject:
 
--   Make sure that the problem or change you are making doesn't already exists.
--   Start the commit message with the word **Fixes**, and always follow it with a reference to the issue: - Fixes #0 - bla bla - Fixes #0, #1 - bla bla
+```
+<type>(<scope>): <subject>
+```
+
+The header is mandatory and the scope of the header is optional.
+
+> Clay follows the "[Conventional Commits](https://www.conventionalcommits.org/)" specification.
+
+#### Type
+
+- **feat**: A new feature
+- **fix**: A bug fix
+- **docs**: Documentation-only changes
+- **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
+- **refactor**: A code change that neither fixes a bug nor adds a feature
+- **perf**: A code change that improves performance
+- **test**: Adding missing or correcting existing tests
+- **chore**: Changes to the build process or auxiliary tools and libraries such as documentation generation
+
+#### Scope
+
+The scope could be anything specifying place of the commit change. For example `feat(@clayui/dropdown)`, `feat(@clayui/css)`, `fix(next.clayui.com)`, `docs(Badge)`, `fix(useCache)`, etc...
 
 ### JavaScript Style Guide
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -107,10 +107,32 @@ This section shows you how to send a pull request. The main staff are always mon
...
 
--   Make sure that the problem or change you are making doesn't already exists.
--   Start the commit message with the word **Fixes**, and always follow it with a reference to the issue: - Fixes #0 - bla bla - Fixes #0, #1 - bla bla
+```
+<type>(<scope>): <subject>
+```
+
+The header is mandatory and the scope of the header is optional.
+
+> Clay follows the "[Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)" specification.
```
wincent: Just link to https://www.conventionalcommits.org/ — it will redirect to the latest, which is already: https://www.conventionalcommits.org/en/v1.0.0-beta.4/

CONTRIBUTING.md
```diff
@@ -107,10 +107,32 @@ This section shows you how to send a pull request. The main staff are always mon
...
+- **refactor**: A code change that neither fixes a bug nor adds a feature
+- **perf**: A code change that improves performance
+- **test**: Adding missing or correcting existing tests
+- **chore**: Changes to the build process or auxiliary tools and libraries such as documentation generation
+
+#### Scope
+
+The scope could be anything specifying place of the commit change. For example `feat(@clayui/dropdown)`, `feat(@clayui/css)`, `fix(next.clayui.com)`, `docs(Badge)`, `fix(useCache)`, etc...
+
+You can use `*` when the change affects more than a single scope.
```
wincent: Is that `*` thing officially part of the spec? I don't see it mentioned. If it affects multiple places, shouldn't you just omit the scope, or pick something broader?

CONTRIBUTING.md
```diff
@@ -107,10 +107,32 @@ This section shows you how to send a pull request. The main staff are always mon
...
+
+The header is mandatory and the scope of the header is optional.
+
+> Clay follows the "[Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)" specification.
+
+#### Type
+
+- **feat**: A new feature
+- **fix**: A bug fix
+- **docs**: Documentation only changes
```
wincent: Documentation-only

CONTRIBUTING.md
```diff
@@ -4,7 +4,7 @@ Want to contribute to Clay? There are a few things you need to know before you s
 
 Here's a set of guidelines to contribute to Clay and its packages. Use your common sense, and feel free to propose changes to this document in a pull request.
 
-For contributions with Clay v2, go to branch `2.x-develop` and follow the [contribution guide](https://github.com/liferay/clay/blob/2.x-develop/CONTRIBUTING.md).
+**For contributions with Clay v2, go to branch `2.x-develop` and follow the [contribution guide](https://github.com/liferay/clay/blob/2.x-develop/CONTRIBUTING.md).**
```
wincent: with → to

CONTRIBUTING.md
```diff
@@ -122,19 +144,18 @@ See guidelines for [contribution with CSS](https://github.com/liferay/clay/blob/
...
-```javascript
-/**
- * Data to add to the element.
- * @default undefined
- * @instance
- * @memberof ClayActionsDropdown
- * @type {?object}
- */
-data: Config.object();
+Clay uses [TypeScript](https://www.typescriptlang.org/) for typed, thereby overwriting [JSDOC](http://usejsdoc.org/) writing but adding descriptions to `interface`, `type`, `enum`, etc... to make it easier to use components.
```
wincent: This reads a bit strangely. I suggest getting rid of "for typed", and saying something like:

> Clay uses TypeScript, obviating the need to write explicit JSDoc annotations for parameter and return types. We still write descriptions, however, for `interface`, `type`, `enum` etc to make it easier to use components.

CONTRIBUTING.md
```diff
@@ -122,19 +144,18 @@ See guidelines for [contribution with CSS](https://github.com/liferay/clay/blob/
...
+
+```typescript
+interface IProps {
+	/**
+	 * Set to true means that network status information will be passed
+	 * via `renders props` and will also cause new renderings as
+	 * networkStatus changes, when false rendering does not
+	 * happen again.
```
wincent: The English here is a bit weird so I'd try to pick another example. For example, maybe a nice short one from clay-drop-down?

```javascript
	/**
	 * Default position of menu element. Values come from `metal-position`.
	 */
	alignmentPosition?: number;
```

CONTRIBUTING.md
matuzalemsteles: Hum yeah, good question. Actually this is not in the specification but I see that this is more explicit than omit the scope, I think of it as a glob 🙂

wincent: I think it's pretty visually noisy and redundant.

wincent: And not spec compliant... 😁 

> A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis

`*` is not a noun

matuzalemsteles: Oh okay, I'll remove 😅

CONTRIBUTING.md
```diff
@@ -107,10 +107,30 @@ This section shows you how to send a pull request. The main staff are always mon
 
 This section shows you how to write commit messages in Clay. Follow these guidelines to help us maintain order and make it easier to locate your changes.
 
-#### Before You Make a Commit Message
+Each commit message consists of a header, a body and a footer. The header has a special format that includes a type, a scope and a subject:
```
bryceosterhaus: What is the body and header portion?

matuzalemsteles: I think this more or less reflects the structure of the header, footer and body.

```markdown
<type>[optional scope]: <description> // <-- Header

[optional body]

[optional footer]
```

I don't know if that's what you wanted to know.

bryceosterhaus: I think I would just remove the portion about the header/body/footer and only specifically mention the type, scope, and subject. Perhaps something like

```
Each commit message should include a typecheck, scope, and subject. It should be formatted as follows:
```

wincent: It's also possible that all of this could be automatically enforced [via bot](https://github.com/probot/semantic-pull-requests).

matuzalemsteles: @bryceosterhaus, in the https://www.conventionalcommits.org/ specification, they say more about it, but we can just go with the header.

Oh I had forgotten about the @wincent bot, we can add it here. What do you think?

julien:
Hey @matuzalemsteles this looks really nice.

I just had a question, I see this is the "scopes" we're planning to use

> feat: A new feature
fix: A bug fix
docs: Documentation-only changes
style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
refactor: A code change that neither fixes a bug nor adds a feature
perf: A code change that improves performance
test: Adding missing or correcting existing tests
chore: Changes to the build process or auxiliary tools and libraries such as documentation generation

I think that if they are that many scopes it's a bit easy to get lost; but that's just my 2 cents.

I'd limit it to `feat`, `fix`, `chore` and maybe `docs` but I'll wait for the rest of the team to chime in.

wincent:
`feat` and `fix` are the only scopes that are *required* by the conventional commits spec. The others are suggestions based on [Angular's patterns](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines). One tool to enforce following the patterns is called "Conform" and they use [this list of scopes](https://github.com/talos-systems/conform/blob/ebed4b31cc2e6f10914850ffba9cd73fca803333/.conform.yaml#L9). Here's another one: [conventional-changelog's list](https://github.com/conventional-changelog/commitlint/blob/d636811b4537904231c483a05e91aaf96deb7f6e/%40commitlint/config-conventional/index.js#L20). Yet another list: ["standard-commit" scopes](https://github.com/stasson/standard-commit/blob/c62e67dfcda8a4afc3ae4f3e5a6727bda09f15f4/src/lib/prompt.ts#L264).

Looking at those, it seems the ones that @matuzalemsteles listed are pretty common across the board, even though I hope most of ours will end up being "fix", "feat", "chore", "refactor", "docs", "test", and we probably won't need to go off into the esoteric ones very often.

matuzalemsteles:
Yes I followed with the Angular notes and https://github.com/commitizen/cz-cli. We probably do not need `perf` and maybe `style` I think we sometimes use a SF, it would be useful for that.

julien:
OK guys thanks for clarifying that - let's start simple and if we need to add more "scopes" we can do that progressively. 

@matuzalemsteles for the "SF" cases, I'd use the `chore` scope - but we can decide whatever is best.

matuzalemsteles:
Thanks @julien, I think we can remove from the `style` list and use @julien's proposal,`perf` and maybe `test` can be replaced by `chore`. What do you think?

Final list:

- feat
- fix
- docs
- refactor
- chore

bryceosterhaus:
That list sounds good to me!

matuzalemsteles:
I went ahead and changed the list. Feel free for merge if you're okay with it.

PR #2024, opened 2019-05-27
docs(README): Update Lexicon links and replace clayui.com with next.clayui.com

matuzalemsteles:


README.md
```diff
@@ -16,9 +16,9 @@ The pillar to give colors and structure to Clay, where is the css and examples o
 
 All components that follow clay-css marking and Lexicon use cases, developed on [React.js](http://reactjs.org)
 
-### [Clayui.com](./clayui.com)
+### [Next.clayui.com](./next.clayui.com)
 
-Contains the source code and documentation powering [clayui.com](https://clayui.com/).
+Contains the source code and documentation powering [next.clayui.com](https://next.clayui.com/).
 
 ## Browser support
```
matuzalemsteles: I'm not sure about listing this "Browser Support" section. What do you think?

bryceosterhaus: I think it'd be helpful to have, so long as we can keep it accurate and up to date. We would also need to make sure that we clearly distinguish the browser support for clay-css vs clay components.

matuzalemsteles: Yeah, that's important to separate from clay-css vs. clay. Maybe I'll say this more to Clay, because if anyone wants to use Clay in their application they will have to deal with the polyfills so that everything works fine.

matuzalemsteles: I'll keep this up and we can discuss this later.

PR #2030, opened 2019-05-30
docs(stickers): Adds CSS documentation for stickers

diegonvs:


next.clayui.com/content/docs/css/components/stickers.md
```diff
@@ -0,0 +1,381 @@
+---
+title: "Stickers"
+description: "Stickers are a visual way to quickly identify content in a different way than badges and labels."
```
bryceosterhaus: I'd probably leave off "in a different way than badges and labels" because it assumes the user knows about badges and labels and also doesnt actually really give any more clarity as to what it does. Should be fine with just "Stickers are a visual way to quickly identify content"

next.clayui.com/content/docs/css/components/stickers.md
```diff
@@ -0,0 +1,381 @@
...
+
+<div class="nav-toc">
+
+- [Colors](#colors)
+- [Position](#position)
+- [Sizes](#sizes)
+- [Variations](#variations)
+    - [Overlay](#overlay)
+	- [Outside](#outside)
+	- [User Icon](#user-icon)
```
bryceosterhaus: Should these two be indented at the same level as overlay?

next.clayui.com/content/docs/css/components/stickers.md
```diff
@@ -0,0 +1,381 @@
...
+    <div class="aspect-ratio">
+        <img alt="thumbnail" class="aspect-ratio-item aspect-ratio-item-fluid" src="/images/thumbnail_hot_air_ballon.jpg">
+        <span class="sticker sticker-bottom-right sticker-danger">PDF</span>
+    </div>
+</div>
+```
+
+# Sizes
+
+Stickers come in 4 sizes `sm`, normal, `lg`, and `xl`. Create your own custom size with the `sticker-size` mixin.
```
bryceosterhaus: maybe swap "normal" for "default"

next.clayui.com/content/docs/css/components/stickers.md
diegonvs: Thanks for the tip, I just copied from lexicon's website

next.clayui.com/content/docs/css/components/stickers.md
diegonvs: That's weird, on my editor everything looks fine: 
Screen Shot 2019-05-30 at 17 01 17

bryceosterhaus: Do you have mixed tabs and spaces by chance? Sometimes it helps to toggle on whitespace in your editor too `"editor.renderWhitespace": "all",`

wincent: Prettier can format Markdown, although I am not sure whether we have any "weird" constructs in our Markdown that might cause issues.

matuzalemsteles: Prettier considers html markup? if you add whitespace between the second-level markups and more, this breaks the gatsby.

bryceosterhaus: Prettier works with normal markdown but I'm not positive about gatsby's version which I think is [MDX](https://www.gatsbyjs.org/docs/mdx/#whats-mdx)?

diegonvs: I'm already using `"editor.renderWhitespace": "all"`.

next.clayui.com/content/docs/css/components/stickers.md
diegonvs: good!

next.clayui.com/content/docs/css/components/stickers.md
diegonvs: Fixed just removing white spaces then tabbing again.

matuzalemsteles: yeah, we used a mix of the two, MDX and MD, we had to do that because many MDX were leaving the build and watch slow, we have cases that do not have to be interpreted by an MDX.

wincent: I see only three files with an ".mdx" extension, so can we apply Prettier to the Markdown ones? On the other hand, if we have ".md" files which are actually MDX, then maybe we should give them that extension so as to be explicit.

matuzalemsteles: hey @wincent  I would not do that 😅,for two reasons the `MDX` files are interpreted by the `gatsby-mdx` plugin which in the end treats them as React components this requires more processing and when we were using `clayui.com` using just `.mdx` we had almost 90 files or more, generated a huge slowness. Processing files with `.md` with the `gatsby-transformer-remark` plugin is much lighter, the idea I had is that `.mdx` files are only for cases where you need to render a React component in our case. pages of the React Components and we can stay with the `.md` files for the CSS pages that are simpler.

Maybe just format the `.md` files?

wincent: > Maybe just format the .md files?

That's exactly what I've been suggesting. But if I understand you correctly, you're saying that many of the ".md" files are actually MDX files?

matuzalemsteles: Oh no, sorry for generating a confusion here 😅. I had understood that it was for us to change the files to `.mdx`, which was to say in the end that both are different and that the `.mdx` files are heavier and that they should only be used if need be.

wincent: Ok, so to summarize:

- Prettier can format the `.md` for us, which will help us avoid these little whitespace issues coming in.
- It cannot format `.mdx`, but that's fine because there are only three of them in the repo.
- We are in no hurry to switch existing `.md` to `.mdx` because that would only slow things down.

matuzalemsteles: Yes, great. Just an addendum, maybe we can increase the number of `.mdx` files because we will add more documents from the components.

PR #2034, opened 2019-05-31
feat(Autocomplete): Create ClayAutocomplete component

matuzalemsteles:
I ended up creating the `ClayForm` component to handle `Input`, but I'm still not sure if it would be necessary, what do you think?

I also did some fixes on `useResource` hook that I ended up having to deal with. I still have to add tests, as soon as we're okay with it, I'll work on it.

### DIFF
--- a/packages/clay-autocomplete/src/DropDown.tsx
+++ b/packages/clay-autocomplete/src/DropDown.tsx
@@ -32,12 +32,12 @@ const DropDown: React.FunctionComponent<IProps> = ({
 	onSetActive = () => {},
 }) => {
 	const {containerElementRef} = useContext(Context);
+	const menuElementRef = React.useRef<HTMLDivElement>(null);
 
 	if (!alignElementRef) {
 		alignElementRef = containerElementRef;
 	}
 
-	const menuElementRef = React.useRef<HTMLDivElement>(null);
 	const alignElementWidth =
 		alignElementRef.current && alignElementRef.current.clientWidth;
 
--- a/packages/clay-autocomplete/src/DropDown.tsx
+++ b/packages/clay-autocomplete/src/DropDown.tsx
@@ -1,0 +1,62 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayDropDown from '@clayui/drop-down';
+import Context from './Context';
+import React, {useContext} from 'react';
+
+export interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * HTML element that the menu should be aligned to
+	 */
+	alignElementRef?: React.RefObject<HTMLElement>;
+
+	/**
+	 * Flag to indicate if menu is showing or not.
+	 */
+	active?: boolean;
+
+	/**
+	 * Callback function for when active state changes.
+	 */
+	onSetActive?: (val: boolean) => void;
+}
+
+const DropDown: React.FunctionComponent<IProps> = ({
+	active = false,
+	alignElementRef,
+	children,
+	onSetActive = () => {},
+}) => {
+	const {containerElementRef} = useContext(Context);
+
+	if (!alignElementRef) {
+		alignElementRef = containerElementRef;
+	}
+
+	const menuElementRef = React.useRef<HTMLDivElement>(null);
+	const alignElementWidth =
+		alignElementRef.current && alignElementRef.current.clientWidth;
+
+	return (
+		<ClayDropDown.Menu
+			active={active}
+			alignElementRef={alignElementRef}
+			autoBestAlign={false}
+			className="autocomplete-dropdown-menu"
+			onSetActive={onSetActive}
+			ref={menuElementRef}
+			style={{
+				maxWidth: 'none',
+				width: `${alignElementWidth}px`,
+			}}
+		>
+			{children}
+		</ClayDropDown.Menu>
+	);
+};
+
+export default DropDown;
--- a/packages/clay-autocomplete/src/LoadingIndicator.tsx
+++ b/packages/clay-autocomplete/src/LoadingIndicator.tsx
@@ -1,0 +1,58 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import ClayLoadingIndicator from '@clayui/loading-indicator';
+import Context from './Context';
+import React, {useContext, useEffect} from 'react';
+
+const LoadingIndicatorMarkup: React.FunctionComponent<
+	React.HTMLAttributes<HTMLDivElement>
+> = ({children, className, ...otherProps}) => (
+	<div
+		{...otherProps}
+		className={classNames(
+			'input-group-inset-item input-group-inset-item-after',
+			className
+		)}
+	>
+		<span className="inline-item inline-item-middle">{children}</span>
+	</div>
+);
+
+export interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Div component to render. It can be a one component that will replace the markup.
+	 */
+	component?:
+		| typeof LoadingIndicatorMarkup
+		| React.ComponentType<any>
+		| React.ComponentType<React.HTMLAttributes<HTMLDivElement>>;
+}
+
+const LoadingIndicator: React.FunctionComponent<IProps> = ({
+	className,
+	component: Component = LoadingIndicatorMarkup,
+	...otherProps
+}) => {
+	const {onLoadingChange} = useContext(Context);
+
+	useEffect(() => {
+		onLoadingChange(true);
+
+		return () => {
+			onLoadingChange(false);
+		};
+	}, []);
+
+	return (
+		<Component {...otherProps} className={className}>
+			<ClayLoadingIndicator small />
+		</Component>
+	);
+};
+
+export default LoadingIndicator;
--- a/packages/clay-autocomplete/src/LoadingIndicator.tsx
+++ b/packages/clay-autocomplete/src/LoadingIndicator.tsx
@@ -27,10 +27,7 @@ export interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
 	 * Div component to render. It can be a one component that will replace the markup.
 	 */
-	component?:
-		| typeof LoadingIndicatorMarkup
-		| React.ComponentType<any>
-		| React.ComponentType<React.HTMLAttributes<HTMLDivElement>>;
+	component?: React.ComponentType<any>;
 }
 
 const LoadingIndicator: React.FunctionComponent<IProps> = ({
--- a/packages/clay-autocomplete/src/index.tsx
+++ b/packages/clay-autocomplete/src/index.tsx
@@ -28,7 +28,7 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
 	 * Div component to render. It can be a one component that will replace the markup.
 	 */
-	component?: typeof AutocompleteMarkup | React.ComponentType<any>;
+	component?: React.ForwardRefExoticComponent<any>;
 }
 
 const ClayAutocomplete: React.FunctionComponent<IProps> & {
--- a/packages/clay-autocomplete/src/index.tsx
+++ b/packages/clay-autocomplete/src/index.tsx
@@ -4,18 +4,67 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import * as React from 'react';
 import classNames from 'classnames';
+import Context from './Context';
+import DropDown from './DropDown';
+import Input from './Input';
+import LoadingIndicator from './LoadingIndicator';
+import React, {useRef, useState} from 'react';
 
-interface Props extends React.HTMLAttributes<HTMLDivElement> {}
+const AutocompleteMarkup = React.forwardRef<
+	HTMLDivElement,
+	React.HTMLAttributes<HTMLDivElement>
+>(({children, className, ...otherProps}, ref) => (
+	<div
+		{...otherProps}
+		className={classNames('input-group', className)}
+		ref={ref}
+	>
+		<div className="input-group-item">{children}</div>
+	</div>
+));
 
-const ClayAutocomplete: React.FunctionComponent<Props> = ({
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Div component to render. It can be a one component that will replace the markup.
+	 */
+	component?: typeof AutocompleteMarkup | React.ComponentType<any>;
+}
+
+const ClayAutocomplete: React.FunctionComponent<IProps> & {
+	DropDown: typeof DropDown;
+	Input: typeof Input;
+	LoadingIndicator: typeof LoadingIndicator;
+} = ({
+	children,
 	className,
+	component: Component = AutocompleteMarkup,
 	...otherProps
 }) => {
+	const containerElementRef = useRef<HTMLDivElement>(null);
+	const [loading, setLoading] = useState(false);
+
 	return (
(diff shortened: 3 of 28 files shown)


packages/clay-shared/src/useDebounce.ts
```diff
@@ -11,15 +11,24 @@ import {useEffect, useState} from 'react';
...
-			clearTimeout(handler);
-		};
-	}, [value]);
+			return () => {
+				clearTimeout(handler);
+			};
+		},
+		// This is required when the `object` has lost the
+		// reference plus the values are the same, `useEffect`
+		// uses `Object.is` or equivalent underneath the cloths.
```
wincent: I think you mean "under the covers".

packages/clay-shared/src/useDebounce.ts
```diff
@@ -11,15 +11,24 @@ import {useEffect, useState} from 'react';
...
+		// For some reason the reference is being lost.
+		typeof value === 'object'
+			? [...Object.keys(value), ...Object.values(value)]
```
wincent: I'm not sure I get what is happening here. The array you are constructing is also an object so I would expect `Object.is` to treat it just like `value`... It would be good to really understand what's happening.

matuzalemsteles: @wincent I spent a few hours debugging the core of React to understand what was happening on Friday, it's pretty much what I'm guessing, possibly the value is missing the reference because within React it stores within the Fibers the effects with the inputs of the previous and next effects, in this case the array with the `value` that is an object... to call the `useEffect` it check if there are inputs in [updateEffectImpl](https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberHooks.js#L870-L889) and then below it decides whether to throw the effect or not checking if the inputs are equal or not, in the end it compares each array item within the implementation [areHookInputsEqual](https://github.com/facebook/react/blob/master/packages/react-dom/src/server/ReactPartialRendererHooks.js#L112-L118) which [in the end uses a polyfill of Object.is](https://github.com/facebook/react/blob/master/packages/shared/objectIs.js#L14-L18) comparing the value of the old and new object. What happened was that the values ​​were the same and did not change the order but in the end it was not the same reference when you compare `a === b`.

I'm not 100% sure if the reference loss is going on inside React for some reason or it's on our side.

packages/clay-shared/src/useDebounce.ts
matuzalemsteles: Oh yeah 😅

packages/clay-shared/src/useDebounce.ts
wincent: Got it. That makes sense. So given an array it is going to compare the values rather than comparing the containing array itself. And because you are using `Object.keys` (ie. immutable strings) and `Object.values` (ie. references), so they end up always being equivalent, unless one of the `key` or `value` items change.

I'm not 100% sure what you mean by "missing the reference" or "lost the reference", but you just mean the object identity has changed, right?

matuzalemsteles: Yes, this may be more of a hack using `Object.values` and `Object.keys` but it's really what I wanted here, just dribbling the identity loss but verifying that the keys and values ​​still remain the same.

I tried to check if it was a problem in React, but it's a bit tricky to understand all the ways, I would spend a few days 😅... but I'm also not sure if `Object.is` would be the best to check this in React but I also understand that it is cheaper than quitting by checking the values ​​in depth.

> I'm not 100% sure what you mean by "missing the reference" or "lost the reference", but you just mean the object identity has changed, right?

 yeah, sorry for not being clear.

packages/clay-autocomplete/src/DropDown.tsx
```diff
@@ -0,0 +1,62 @@
...
+	children,
+	onSetActive = () => {},
+}) => {
+	const {containerElementRef} = useContext(Context);
+
+	if (!alignElementRef) {
+		alignElementRef = containerElementRef;
+	}
+
+	const menuElementRef = React.useRef<HTMLDivElement>(null);
```
bryceosterhaus: For sake of readability, I'd probably try to keep hooks towards the top of the function declaration.

I'm open to other ideas, but this is generally what I see in other projects.

packages/clay-autocomplete/src/LoadingIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+		<span className="inline-item inline-item-middle">{children}</span>
+	</div>
+);
+
+export interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Div component to render. It can be a one component that will replace the markup.
+	 */
+	component?:
+		| typeof LoadingIndicatorMarkup
```
bryceosterhaus: Is it necessary to define this here if we already declare `React.ComponentType<any>`?

packages/clay-autocomplete/src/LoadingIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+		| React.ComponentType<any>
+		| React.ComponentType<React.HTMLAttributes<HTMLDivElement>>;
```
bryceosterhaus: ^

packages/clay-autocomplete/src/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+	>
+		<div className="input-group-item">{children}</div>
+	</div>
+));
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Div component to render. It can be a one component that will replace the markup.
+	 */
+	component?: typeof AutocompleteMarkup | React.ComponentType<any>;
```
bryceosterhaus: isnt `typeof AutocompleteMarkup` is a subset of `React.ComponentType<any>`?

packages/clay-autocomplete/src/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+	...otherProps
+}) => {
+	const containerElementRef = useRef<HTMLDivElement>(null);
+	const [loading, setLoading] = useState(false);
+
+	return (
+		<Component
+			{...otherProps}
+			className={className}
+			ref={containerElementRef}
```
bryceosterhaus: Since `Component` requires a `ref`, should the IProps interface require something like `React.ForwardRefExoticComponent<IProps & React.RefAttributes<HTMLDivElement>>`

packages/clay-autocomplete/stories/index.tsx
```diff
@@ -0,0 +1,131 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayAutocomplete from '../src';
+import ClayDropDown from '@clayui/drop-down';
+import fuzzy from 'fuzzy';
```
bryceosterhaus: Since this is only used in our `stories`, do you think we really need it at all? We can just use `string.match()`

packages/clay-form/src/Input.tsx
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	allowedCharacters?: RegExp;
+
+	/**
+	 * Input component to render. Can either be a string like 'input' or a component.
+	 */
+	component?:
+		| 'input'
+		| React.ComponentType<any>
+		| React.ComponentType<React.InputHTMLAttributes<HTMLInputElement>>;
```
bryceosterhaus: This is a subset of `React.ComponentType<any>`

packages/clay-form/src/Input.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import classNames from 'classnames';
+import React, {useMemo} from 'react';
+
+export interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
+	/**
+	 * Gets the accepted characters of the input
+	 * element values
+	 */
+	allowedCharacters?: RegExp;
```
bryceosterhaus: Is this type of functionality really needed for our use case? We could likely just leave this to the user by having them manage their own `value` and `onChange`. I may be missing something here though...

packages/clay-form/src/index.tsx
```diff
@@ -0,0 +1,21 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import Input, {IProps as IInputProps} from './Input';
+
+const ClayForm: React.FunctionComponent<
```
bryceosterhaus: Rather than rolling out our own Form component and other elements inside of the form, what do you think about just using `formik`? I believe the personalization and segments editor in portal is already using it.

bryceosterhaus: Anecdote: We rolled our own form component and elements for both Loop and Faro, and it was one of the hardest and most annoying areas of the app we had to maintain and fix issues in.

matuzalemsteles: I agree with you, it's too bad to deal with Forms, the idea of `<ClayForm />` is to just provide a few markups with a few features and it's easy to integrate with `formik`, the part I have not yet started exploring.

Maybe it's something to explore in upcoming releases or add along with these new features.

packages/clay-form/src/Input.tsx
matuzalemsteles: I ended up bringing this in when we introduced MultiSelect to v2, that was difficult there, now with low-level I really do not see if there's any more need, but I think it would be interesting to consider some features and within Clay if we delegate always for developers to deal with this, with low-level they may have to deal with all logic always this can generate overload for those who consume.

packages/clay-autocomplete/src/DropDown.tsx
matuzalemsteles: Makes sense, I think it's worth adopting as a practice.

packages/clay-form/src/index.tsx
bryceosterhaus: Yeah I think it might be best to keep our Form and form elements as simple as possibly where we simply add the appropriate classes, type requirements, and then pass everythign else in via `...otherProps` unless there is a functionality difference with clay. This gives us freedom to default to the user using their own callbacks or formik.

packages/clay-form/src/Input.tsx
bryceosterhaus: Yeah I agree. I think we can keep the low-level simple and then anything we find our users re-implementing and creating, we can then create a high-level component for ease of use.

packages/clay-autocomplete/src/LoadingIndicator.tsx
matuzalemsteles: Yes, I think maybe not, I've seen this double statement in some codebase but I can not understand the real reason, if it's just enforcing the use of `LoadingIndicatorMarkup` for example. I'll remove that.

packages/clay-autocomplete/src/index.tsx
matuzalemsteles: Hm yeah, thanks for that.

packages/clay-autocomplete/stories/index.tsx
matuzalemsteles: I'd rather keep `fuzzy`, it handles the match by character, I'd probably have to create some more logic to handle it.

packages/clay-form/src/index.tsx
matuzalemsteles: Yeah, I think we can then add a use case of Formik with ClayForm in Storybook.

packages/clay-form/src/Input.tsx
matuzalemsteles: Yeah, I'm going to remove this functionality for now and depending on usage we'll add a new API or high-level if need to.

packages/clay-autocomplete/stories/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+		<ClayAutocomplete>
+			<ClayAutocomplete.Input
+				onChange={event => setValue(event.target.value)}
+				value={value}
+			/>
+			<ClayAutocomplete.DropDown
+				active={(!!resource && !!value) || initialLoading}
+			>
+				<ClayDropDown.ItemList>
+					<LoadingWithDebounce
```
matuzalemsteles: I've been wondering if this could turn out to be a high-level or low-level component, part of a specification to give feedback to the user, will probably be heavily used. What do you think @bryceosterhaus?

matuzalemsteles: I've been wondering if this could turn out to be a high-leve or low-level component, part of a specification to give feedback to the user, will probably be heavily used. What do you think @bryceosterhaus?

packages/clay-autocomplete/stories/index.tsx
bryceosterhaus: Oh so is this logic a clay requirement? I would figure that the matching logic of the autocomplete would be up to the user.

matuzalemsteles: It's something provided by the Lexicon, but this should be customized so I'm thinking of making it as a low-level component `<ClayAutocomplete.Item value={item.name} match={value} />` replacing `<ClayDropDown.Item>`. What do you think?

bryceosterhaus: Oh yeah thats a good idea. Im assuming the `match` prop would be a regex that makes the portion of the text bold?

packages/clay-autocomplete/stories/index.tsx
bryceosterhaus: Not sure if it should be its own component or if we just create a shared hook. I know once the suspense API is fully released that will help with the integration point of using loading indicators. I'd probably be fine if it was its own component though, we can always try it out and see how it works.

packages/clay-autocomplete/stories/index.tsx
matuzalemsteles: Not really 😁. It would be the value entered in the input that will be used to compare, we would use `fuzzy`. What do you think?

packages/clay-autocomplete/stories/index.tsx
matuzalemsteles: Hm I'm not sure if Suspense would help with this, this rule is very specific, the loading screen in DropDown should only appear when the data is passing 500ms. But I think we can run out of it for now and document this use case in next.clayui.com.

packages/clay-autocomplete/stories/index.tsx
bryceosterhaus: Oh I see, we could do that. One thought I have is, what if a user doesn't want the bold matching? This makes me think I would rather have a call back for formatting, this would then allow people to do other sorts of highlighting, like uppercase.

```jsx
<ClayAutocomplete.Item value={item.name} formatter={(val) => val.toUpperCase()} />
```
But now that I type that out, I wonder why we even pass in props...
```jsx
<ClayAutocomplete.Item>{item.name.toUpperCase()}</ClayAutocomplete.Item>
```

matuzalemsteles: I'm not sure about this but it would be interesting to add the `formatter` option, but we would probably have to create our own internal logic to allow this or offer the same API as `fuzzy` but would not allow too many things to just add wrappers between the texts.

> But now that I type that out, I wonder why we even pass in props...

If following this example I do not know if it would be necessary to use `ClayAutocomplete.Item`, I would probably only use `ClayDropDown.Item`.

bryceosterhaus: I'd be fine with just using `ClayDropDown.Item`.

matuzalemsteles: Summarizing just to see if I understood what we're waking up here:

- Create `<ClayAutocomplete.Item value={item.name} match={value} />` for just that case in bold.
- If someone wants to create something custom, use `<ClayDropDown.Item />`

Am I right or missing something? 🙂

bryceosterhaus: Yeah that sounds good to me! And we can always change it in the future if needed too.

bryceosterhaus:
LGTM. I'll let you merge if you are waiting on anything else.

PR #2038, opened 2019-06-04
docs(popovers-and-tooltips): Adds css documentation for popovers-and-tooltips

diegonvs:


next.clayui.com/content/docs/css/components/popovers-and-tooltips.md
```diff
@@ -0,0 +1,469 @@
+---
+title: "Popovers and Tooltips"
+description: "Popovers and Tooltips are patterns used to show helpful information in two different ways."
+---
+
+<div class="nav-toc">
+
+- [Popover](#popover)
```
bryceosterhaus: Rather than repeat "Popover" here, I think it makes more sense to do something like

```md
- [Popover](#popover)
    - [Position](#popover-position)
        - [Top](#popover-top)
        - [Right](#popover-right)
        - [Bottom](#popover-bottom)
        - [Left](#popover-left)
```

Same for tooltip section, and same changes for the headings below.

next.clayui.com/content/docs/css/components/popovers-and-tooltips.md
```diff
@@ -0,0 +1,469 @@
...
+- [Tooltip](#tooltip)
+    - [Tooltip Position](#popover-position)
+        - [Tooltip Top](#tooltip-top)
+        - [Tooltip Right](#tooltip-right)
+        - [Tooltip Bottom](#tooltip-bottom)
+        - [Tooltip Left](#tooltip-left)
+
+</div>
+
+# Popover
```
matuzalemsteles: Headers should start with `h2`. Same for others.
```suggestion
## Popover
```

next.clayui.com/content/docs/css/components/popovers-and-tooltips.md
diegonvs: I didn't know that it will works but worked well! Thanks for the tip Bryce.

next.clayui.com/content/docs/css/components/popovers-and-tooltips.md
```diff
@@ -0,0 +1,469 @@
...
+<div class="nav-toc">
+
+- [Popover](#popover)
+    - [Position](#popover-position)
+        - [Top](#popover-top)
+        - [Right](#popover-right)
+        - [Bottom](#popover-bottom)
+        - [Left](#popover-left)
+- [Tooltip](#tooltip)
+    - [Position](#popover-position)
```
bryceosterhaus: should be `#tooltip-position`

bryceosterhaus: Actually I don't think any of these work since the headers don't contain the `popover` or `tooltip` portions. You may actually have to have the prefix, unless markdown provides a syntax for custom `id` for headers.

PR #2051, opened 2019-06-10
feat: create ClaySelect with low-level and high-level components

matuzalemsteles:


matuzalemsteles:
I'm merging this so we can have it available for the first alpha. Feel free to leave your thoughts.

PR #2054, opened 2019-06-11
docs(next.clayui.com): Uses code-like font instead of a normal

diegonvs:


matuzalemsteles:
hey @diegonvs I've been looking at this again, I think we'd better leave it as it is. Anyway, we'll go back with that again if we think it's best.

PR #2061, opened 2019-06-13
docs(icon): Adds documentation for Icon React component

diegonvs:


diegonvs:
Depends on #2063

next.clayui.com/content/docs/components/icon.mdx
```diff
@@ -0,0 +1,45 @@
...
+For Icons, we use SVG elements that link to an SVG sprite. So, it becomes necessary to pass a parameter that points to a path where this [`spritemap`](#api-spritemap) is located.
+
+- See the complete list of symbols and flags that are offered by Clay's spritemap [here](/docs/css/components/icons.html)
+
+<Icon />
+
+### Using Context
+
+<div class="clay-site-alert alert alert-info">
+    For V3, was added the ability to utilize context for passing spritemap down instead of having to pass the prop everywhere.
```
bryceosterhaus:
```suggestion
    For V3, an API was added to utilize context for passing spritemap down instead of having to pass the prop everywhere.
```

next.clayui.com/src/components/clay/Icon.js
```diff
@@ -0,0 +1,41 @@
...
+
+import ClayIcon, {ClayIconSpriteContext} from '@clayui/icon';
+import Editor from './Editor';
+import React from 'react';
+
+const spritemap = '/images/icons/icons.svg';
+
+export const Icon = () => {
+	const scope = {ClayIcon, spritemap};
+	const code = `const Component = () => {
```
bryceosterhaus: Can we indent this template to be on the same line? Otherwise it looks a bit weird.

packages/clay-icon/src/index.tsx
```diff
@@ -28,7 +28,7 @@ const ClayIcon: React.FunctionComponent<IProps> = ({
 	spritemap,
 	symbol,
 	...otherProps
-}) => {
+}: IProps) => {
```
bryceosterhaus: I dont think this should be necessary since we already declare `IProps` in line 26.

matuzalemsteles: hey @bryceosterhaus this is more because of react-docgen to be able to generate the API's in doc, it can not infer the props from `React.FunctionComponents <IProps>` for example. Probably this could be an improvement there...

next.clayui.com/content/docs/components/icon.mdx
diegonvs: Solved!

next.clayui.com/src/components/clay/Icon.js
diegonvs: When Identing the template in this way,
Screen Shot 2019-06-17 at 19 11 01

the output in the example will look a bit weird, see:
Screen Shot 2019-06-17 at 19 11 09

diegonvs: A possible workaround for this is:
Screen Shot 2019-06-17 at 19 13 38

What do you think? @matuzalemsteles @bryceosterhaus

packages/clay-icon/src/index.tsx
diegonvs: Yes, considering Matu comment I gonna stay with this type.

diegonvs:
Updated @matuzalemsteles  @bryceosterhaus

packages/clay-icon/src/index.tsx
bryceosterhaus: dang, thats unfortunate about react-docgen. I wonder if we could send a feature for this or if its on the roadmap. Otherwise, we should decide how we want to declare these types since this area will sort of be a "gotcha"

next.clayui.com/src/components/clay/Icon.js
bryceosterhaus: hmm, still looks weird to me.

What if we did something like this and move it outside of the function.
```tsx
const iconWithContextCode = `const Component = () => {
	return (
		<ClayIconSpriteContext.Provider value={spritemap}>
			<ClayIcon symbol="add-cell" />
		</ClayIconSpriteContext.Provider>
	);
}

render(<Component />)`;

export const IconWithContext = () => {
	const scope = {ClayIcon, ClayIconSpriteContext, spritemap};

	return <Editor code={iconWithContextCode} scope={scope} />;
};
```

packages/clay-icon/src/index.tsx
matuzalemsteles: Yeah, I think we can try to send a resource there, I'm not sure if we can implement our own resolver in that case, if so, we can solve it from here but I think it should be a react-docgen resource. I'll try to check this out and come back here again.

I did not want to waste too much time on it, so it was the quickest solution I had found, but it sure should not be anything permanent like that.

next.clayui.com/src/components/clay/Icon.js
matuzalemsteles: I think this might look better!

diegonvs: LGTM

PR #2064, opened 2019-06-14
forwardRef for clay charts

gcmznt:


matuzalemsteles:
hey @gcmznt thanks for this, we're making some changes to ClayCharts in #2046, I think we can add that there since @bryceosterhaus is working on it.

matuzalemsteles:
hey @gcmznt I'm closing this up once Bryce just added in #2046, thanks for that again.

bryceosterhaus:
@gcmznt @matuzalemsteles I was thinking about this more last night after I added the forwardRefs. Do you think the forwarded ref should point to the chart's html element, or the instance of the billboard chart? I think it might make more sense to point to the chart instance rather than the html element since its likely rare you would ever want to interact with the root element of the chart.

matuzalemsteles:
Hmm, I think it makes sense to provide ref to the instance of the chart. I do not know if it really makes sense for the html element, maybe we can provide the two information in ref.

PR #2065, opened 2019-06-14
docs(select): Update lockfile

diegonvs:


diegonvs:
Depends on #2063

next.clayui.com/content/docs/components/select.mdx
```diff
@@ -0,0 +1,50 @@
...
+<div class="nav-toc">
+
+- [Variants and Best Pratices](#variants-and-best-pratices)
+    - [SelectWithOption](#selectwithoption)
+- [API](#api)
+
+</div>
+
+<div class="clay-site-alert alert alert-info">
+	Check the <a href="https://liferay.design/lexicon/">Lexicon</a> <a href="https://liferay.design/lexicon/core-components/forms/selector/">Selectors Pattern</a> for a more in-depth look at the motivations and proper usage of this component.
```
bryceosterhaus: Any idea if its possible to use markdown inside of html blocks? Im assuming not, but wasn't sure if we could just use a markdown link instead of an anchor tag.

next.clayui.com/content/docs/components/select.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+If you use ClaySelect only for simple cases that do not need props for [`options`](#api-options), you can use `<ClaySelectWithOption />` which will have the same result as the previous version.
+
+<SelectWithOption />
+
+## API
+
+### Select and Select.Option Variants
+
+These components don't have an API. Use the same attributes of a `<select>` element for `ClaySelect` and the same attributes of a `<option>` element for `ClaySelect.Option` variant.
```
bryceosterhaus: I wouldn't say "they don't have an API", Rather just say the API is identical to the `<select>` element and then link to the react api reference for form elements. Same for `<option>`.

next.clayui.com/content/docs/components/select.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+<div class="clay-site-alert alert alert-info">
+	Check the <a href="https://liferay.design/lexicon/">Lexicon</a> <a href="https://liferay.design/lexicon/core-components/forms/selector/">Selectors Pattern</a> for a more in-depth look at the motivations and proper usage of this component.
+</div>
+
+## Variants and Best Pratices
+
+### Default
+
+This pattern is completely different from a dropdown. A selector is a native component with a data provider. A dropdown, however, opens a panel that could contain multiple and different natural elements.
```
bryceosterhaus: I'm not sure the references to dropdown are helpful here. A user may have no idea or concept of clay's dropdown. Rather I think we can reference the native component portion and link to both clay's styling of the native <select> element.

next.clayui.com/src/components/clay/Select.js
```diff
@@ -0,0 +1,70 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClaySelect, {ClaySelectWithOption} from '@clayui/select';
+import Editor from './Editor';
+import React from 'react';
+
+export const Select = () => {
+	const scope = {ClaySelect};
+	const code = `const Component = () => {
```
bryceosterhaus: Can we put this code reference all on the same line indentation? If need be, we can move this to a constant above `Select`

next.clayui.com/src/components/clay/Select.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+render(<Component />)`;
+
+	return <Editor code={code} scope={scope} />;
+};
+
+export const SelectWithOption = () => {
+	const scope = {ClaySelectWithOption};
+	const code = `const Component = () => {
```
bryceosterhaus: ^ same as above

packages/clay-select/src/SelectWithOption.tsx
```diff
@@ -18,7 +18,7 @@ interface IProps extends React.ComponentProps<typeof Select> {
 const SelectWithOption: React.FunctionComponent<IProps> = ({
 	options = [],
 	...otherProps
-}) => (
+}: IProps) => (
```
bryceosterhaus: Shouldn't need this since we have line 18.

yarn.lock
```diff
@@ -8034,13 +8034,6 @@ detect-conflict@^1.0.0:
   resolved "https://registry.yarnpkg.com/detect-conflict/-/detect-conflict-1.0.1.tgz#088657a66a961c05019db7c4230883b1c6b4176e"
```
bryceosterhaus: how come this file is updated? We didnt update any package.json dependencies did we?

next.clayui.com/content/docs/components/select.mdx
diegonvs: Considering our current implementation on Gatsby isn't possible
Screen Shot 2019-06-17 at 19 20 32
Screen Shot 2019-06-17 at 19 20 11

next.clayui.com/content/docs/components/select.mdx
diegonvs: Yes, I agree. More polished

next.clayui.com/content/docs/components/select.mdx
diegonvs: Solved!

next.clayui.com/src/components/clay/Select.js
diegonvs: See https://github.com/liferay/clay/pull/2061#discussion_r294540203

next.clayui.com/src/components/clay/Select.js
diegonvs: See https://github.com/liferay/clay/pull/2061#discussion_r294540203

packages/clay-select/src/SelectWithOption.tsx
diegonvs: It's used for react-docgen

yarn.lock
diegonvs: Re-running `yarn` again

diegonvs: You're correct. I have no idea why this lockfile was updated.

diegonvs: Thanks for the tip!

diegonvs:
Updated @bryceosterhaus and @matuzalemsteles

next.clayui.com/content/docs/components/select.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+### SelectWithOption
+
+If you use ClaySelect only for simple cases that do not need props for [`options`](#api-options), you can use `<ClaySelectWithOption />` which will have the same result as the previous version.
+
+<SelectWithOption />
+
+## API
+
+### Select and Select.Option Variants
```
bryceosterhaus: We can probably remove this section since it doesn't really provide any extra information that the user wouldn't already assume.

PR #2071, opened 2019-06-18
docs(next.clayui.com): Adds references for WAI-ARIA best practices on CSS documentation

diegonvs:
While @eduardoallegrini don't start working reviewing ARIA best practices on our documentation I think that is okay.


matuzalemsteles:
cc @pat270

PR #2076, opened 2019-06-18
fix: update components to include 'types' and 'ts:main' package keys

bryceosterhaus:
I was testing some of these components in faro and realized we had issues with our [mainFields](https://webpack.js.org/configuration/resolve/#resolvemainfields) in non typescript projects. I removed the "module" field since that was the primary field that webpack looked for, this way projects consuming our components will default to use the `lib` version. I also specified `types` to be picked up for projects that use ts.

I additionally added `ts:main` field. I am a little unsure about this and I am welcome to suggestions, but we needed a way for our storybook config to pick up the `src` files and not the built `lib` versions. By specifying `ts:main` in each package and also in our storybook webpack config, this allows us to use our source files in storybook.

Let me know what you guys think.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -14,6 +14,9 @@ module.exports = ({config}) => {
 		test: /\.(ts|tsx)$/,
 		use: [require.resolve('awesome-typescript-loader')],
 	});
+
+	config.resolve.mainFields = ['ts:main', 'module', 'main'];
+
 	config.resolve.extensions.push('.ts', '.tsx');
 	return config;
 };
--- a/next.clayui.com/gatsby/onCreateWebpackConfig.js
+++ b/next.clayui.com/gatsby/onCreateWebpackConfig.js
@@ -26,6 +26,7 @@ module.exports = ({actions, stage}) => {
 			alias: {
 				$components: path.resolve(__dirname, '../src/components'),
 			},
+			mainFields: ['ts:main', 'module', 'main'],
 			modules: [path.resolve(__dirname, 'src'), 'node_modules'],
 		},
 	});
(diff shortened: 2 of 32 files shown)


.storybook/webpack.config.js
```diff
@@ -14,6 +14,9 @@ module.exports = ({config}) => {
 		test: /\.(ts|tsx)$/,
 		use: [require.resolve('awesome-typescript-loader')],
 	});
+
+	config.resolve.mainFields = ['ts:main', 'module', 'main'];
```
matuzalemsteles: Maybe in the place of `ts:main` would it be `source`? it may be clearer that it is not a file to be consumed by something from outside.

bryceosterhaus: The reason I went with `ts:main` is because I wanted it to be somewhat similar to `main` otherwise I thought someone might think its something entirely different.

matuzalemsteles: Oh yeah, makes sense. I do not see problem also in going with `ts:main`, just giving another name if `ts:main` is still doubtful.

bryceosterhaus: I still think their might be a better solution for this out there, but I just couldn't find anything for it. I'm sure someone has run into this issue before, I just couldn't find an example.

 Do you know of any lerna + ts libraries out there that are good examples?

@wincent any chance you've seen this before?

matuzalemsteles: Hm yeah, I did a little research but I also did not find anything related to this, it is quite difficult to see such setup with lerna+ts.

bryceosterhaus:
gonna merge since this is needed for a release. We can totally change this later if needed though.

.storybook/webpack.config.js
wincent: > @wincent any chance you've seen this before?

I'm afraid not.

PR #2083, opened 2019-06-19
fix(clay css) #1424 Add Clay Time (Time Picker)

pat270:
https://issues.liferay.com/browse/LEXI-481

### DIFF
--- a/packages/clay-css/src/content/form_elements_date_picker.html
+++ b/packages/clay-css/src/content/form_elements_date_picker.html
@@ -162,17 +162,41 @@ section: Components
 							</div>
 
 							<div class="date-picker-calendar-footer">
-								<div class="time-picker">
+								<div class="clay-time">
 									<div class="input-group">
 										<div class="input-group-item input-group-item-shrink">
-											<span class="input-group-text">
+											<div class="input-group-text">
 												<svg class="lexicon-icon lexicon-icon-time" focusable="false" role="presentation">
 													<use xlink:href="{{rootPath}}/images/icons/icons.svg#time"></use>
 												</svg>
-											</span>
+											</div>
 										</div>
 										<div class="input-group-item">
-											<input class="form-control" name="timer" type="time" value="00:00">
+											<div class="form-control">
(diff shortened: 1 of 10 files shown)


matuzalemsteles:

packages/clay-css/src/content/form_elements_date_picker.html
```diff
@@ -162,17 +162,41 @@ <h3>Date Picker</h3>
...
+											</div>
 										</div>
 										<div class="input-group-item">
-											<input class="form-control" name="timer" type="time" value="00:00">
+											<div class="form-control">
+												<div class="clay-time-edit">
+													<input class="clay-time-hours form-control-inset" id="timePicker12hTimeZoneIcon" maxlength="2" name="uniqueNameHours" type="text" value="12">
+													<span class="clay-time-divider">:</span>
+													<input class="clay-time-minutes form-control-inset" maxlength="2" name="uniqueNameMinutes" type="text" value="00">
+													<input class="clay-time-ampm form-control-inset" maxlength="2" name="uniqueNameAmpm" type="text" value="PM">
```
matuzalemsteles: I'm wondering if it should really be an input or select, I'm not sure what the default is. But it seems strange that this is an input. Any idea?

pat270: @matuzalemsteles We could use select but the browser pops up the select menu which wasn't in the Lexicon Designs. The browser default is a text input that changes the value with the up and down arrow keys.

It doesn't matter to me what type of element we use as long as the Lexicon Team doesn't care.

matuzalemsteles: hmm ok, hey @victorvalle what do you think about it?

marcoscv-work: I think it's fine and we can go ahead and merge it! :-)

matuzalemsteles: 🙂yeah, I think we can review this later so we do not block this PR.

PR #2086, opened 2019-06-19
docs(card): add ClayCard component docs

matuzalemsteles:


matuzalemsteles:
hey @bryceosterhaus I'm going ahead with this, since this is simple. If you have anything else to add. We can add these changes at any time.

Issue #397, opened 2017-12-12
ClayButton borderless


carloslancha:
We need `btn-borderless` class to fit with Lexicon style:

![captura de pantalla 2017-12-12 a las 15 33 55](https://user-images.githubusercontent.com/5803434/33889904-e6d82588-df51-11e7-8b47-3f33602bf55d.png)

https://lexicondesign.io/docs/patterns/buttons.html

Right now we have this:

![captura de pantalla 2017-12-12 a las 15 34 44](https://user-images.githubusercontent.com/5803434/33889941-03ceb65c-df52-11e7-927f-6e680790df5d.png)

pat270:
@carloslancha We have this `btn btn-outline-secondary btn-borderless` and `btn btn-link` the only thing is btn-link isn't bolded like in the screen shot.

carloslancha:
@pat270 lexicon doesn't define the outlines styles so in ClayButton we're just adding the class `btn-` + specified style (`primary`, `secondary`, etc). Couldn't be possible to style borderless to work just in that way?

pat270:
We should also be considering the things Bootstrap has provided. `.btn-outline-*` is shipped by default, I can't get rid of it. Clay is already bloated as is and adding another button variant doesn't seem like a good idea to me. Where are the use cases exactly?

yuchi:
Bystander’s 2cents again: both `primary` and `secondary` are semantic adjectives, while `borderless` is not. That’s confusing.

If you called it `btn-tertiary` the confusion here would have been prevented.

natecavanaugh:
I have to agree with @yuchi that `borderless` isn't a semantic name, but I'm not sure if tertiary works (because it implies that the button is lower in the hierarchy, but reading the docs, it may be considered on it's own, less important level).

Some other possibilities might be (if they're not already being used) `btn-label` or `btn-subtle`.

But as I'm writing this, I think Pier had it right, `btn-tertiary` may be the best one, unless we can think of some actual functionality that these buttons have.

yuchi:
Thanks @natecavanaugh. You are indeed right that “tertiary” does imply a lower level in the hierarchy and the docs do not imply that. Anyway that specific proposal was just to give an example.

pat270:
I like the name `btn-tertiary` much better than `borderless` as well. I'm still hesitant to add a new set of buttons when we already have a set of unused buttons that are styled like the `borderless` buttons with the exception of colored borders. Maybe I can name it `btn-outline-borderless`?

[...]

PR #2112, opened 2019-06-22
feat(button): Adds borderless and outline properties

diegonvs:


matuzalemsteles:
I'd rather take it easy with that and review Issue #397 before merging.

jbalsas:
Hey @matuzalemsteles, I reviewed our conversation in #397 and the final outcome was:

>After checking with @victorvalle and the Lexicon team, Borderless is indeed a variation, not a new type of button. They will proceed to document it in a different way and we can proceed to remove it from the clay component implementation.
>
>If someone wants a borderless button, as @pat270 mentioned, she can use the outline modifiers accordingly, or it will be styled by default based on context (like in a toolbar).

I think that should put `borderless` on the same level as other variants we already support such as `block`, `monospaced` or `small`. 

What do you think?

matuzalemsteles:
I really do not see a problem putting it in, but what I've been thinking about ending up adding a lot of API to our components (even though they're the simplest) is that they can get so bloated that any future change would make it very painful to discontinue many if necessary.

So what I'm thinking is: what sets our API limit we can add is what the Lexicon is predicting... does that make sense to you?

packages/clay-button/src/index.tsx
```diff
@@ -26,34 +31,44 @@ interface IProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
...
 	<button
-		{...otherProps}
-		className={classNames(className, 'btn', `btn-${displayType}`, {
+		className={classNames(className, 'btn', {
 			'btn-block': block,
 			'btn-monospaced': monospaced,
+			'btn-outline-borderless': borderless,
 			'btn-sm': small,
+			[`btn-${displayType}`]: displayType && !outline,
+			[`btn-outline-${displayType}`]: displayType && outline,
```
matuzalemsteles: For `borderless` to have effect it needs `btn-outline-*`and should disable `btn-${displayType}`.
```suggestion
			[`btn-outline-${displayType}`]: displayType && (outline || borderless),
```

Issue #2107, opened 2019-06-21
Add keyboard functionality for autocomplete


bryceosterhaus:
We need to enable arrow keys and return key at a minimum for autocomplete and dropdown instances.

andresfulla:
I have a use case related to this. I'm using `@clayui/dropdown` component. It's related to how the focus behaves on this element.

When the user is focused on the component rendered by the `trigger` prop, then opens the dropdown by hitting the spacebar, (once tab is pressed again) the next focused element is not inside the dropdown body. The focus just moves out of the dropdown.

I'd like the focus to move into the elements inside the dropdown and only then move to whatever is next to the dropdown.

bryceosterhaus:
Fixed https://github.com/liferay/clay/pull/2122. Will also be implemented with high-level autocomplete component

bryceosterhaus:
Fixed https://github.com/liferay/clay/pull/2122. Will also be implemented with high-level autocomplete component

PR #2122, opened 2019-06-25
add useArrowKeysForList and useEnterKey hooks

bryceosterhaus:
The aim of these hooks is to address https://github.com/liferay/clay/issues/2107. To see how they work, checkout the autocomplete stories. 

My initial thought is to get these reviewed and approved and if they look good, add them to the multiselect high-level component for an initial testing ground.

### DIFF
--- a/packages/clay-autocomplete/stories/index.tsx
+++ b/packages/clay-autocomplete/stories/index.tsx
@@ -6,10 +6,10 @@
 
 import ClayAutocomplete from '../src';
 import ClayDropDown from '@clayui/drop-down';
-import React, {useState} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 import {FetchPolicy, NetworkStatus} from '@clayui/data-provider/src/types';
 import {storiesOf} from '@storybook/react';
-import {useDebounce} from '@clayui/shared';
+import {useDebounce, useKeyHandlerForList} from '@clayui/shared';
 import {useResource} from '@clayui/data-provider';
 
 import '@clayui/css/lib/css/atlas.css';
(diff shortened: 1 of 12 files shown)


packages/clay-autocomplete/stories/index.tsx
```diff
@@ -52,10 +54,23 @@ const ClayAutocompleteWithState = () => {
...
+	useArrowKeysForList(
+		activeIndex,
+		resource ? resource.results.length : 0,
+		setActiveIndex,
+		inputRef.current
+	);
+
+	useEnterKey(
+		() => alert(resource.results[activeIndex].name),
+		inputRef.current
```
matuzalemsteles: This seems strange, just out of curiosity, why did we pass the reference from the input and not from `<ClayAutocomplete.DropDown />`, had problems with this?

packages/clay-autocomplete/stories/index.tsx
```diff
@@ -52,10 +54,23 @@ const ClayAutocompleteWithState = () => {
...
 	const error = networkStatus === 5;
 
+	useArrowKeysForList(
+		activeIndex,
+		resource ? resource.results.length : 0,
+		setActiveIndex,
+		inputRef.current
+	);
+
+	useEnterKey(
```
matuzalemsteles: I'm thinking maybe this is unnecessary, we can have the same effect by adding `onEnter` to `<ClayAutocomplete.Item /> `and React will handle events... what do you think?

packages/clay-shared/src/useArrowKeysForList.ts
```diff
@@ -0,0 +1,50 @@
...
+				event.keyCode === DOWN_ARROW_KEY_CODE
+			) {
+				event.preventDefault();
+
+				const nextIndex =
+					activeIndex === totalItems - 1 ? 0 : activeIndex + 1;
+				const prevIndex =
+					activeIndex === 0 ? totalItems - 1 : activeIndex - 1;
+
+				onIndexChange(
```
matuzalemsteles: We'll probably have to focus on the element too so that the dropdown scroll can track the active item.

wincent:
I'm a bit worried about the pattern here of attaching event listeners to actual DOM elements, effectively bypassing React's synthetic event system (which isolates us from a lot of cross-browser quirks, and is performant by default due to transparent event delegation).

Have you taken a look at some prior art, like how Reach UI's "Combobox" is implemented?

- https://ui.reach.tech/combobox#combobox
- https://github.com/reach/reach-ui/tree/master/packages/combobox

Ryan Florence, the author, has put in a huge amount of effort to making all the components accessible.

packages/clay-shared/src/useArrowKeysForList.ts
```diff
@@ -0,0 +1,50 @@
...
+					event.keyCode === UP_ARROW_KEY_CODE ? prevIndex : nextIndex
```
matuzalemsteles: I think we also have to cover another use case, when I'm active in the first item and do a `prevIndex` I wanted the focus to be turned to the input.

bryceosterhaus:
@wincent tbh not sure why I ended up just drilling down by using event listeners rather than react's event system. Gonna look into this.

Also, Ryan's code  comments are always great, https://github.com/reach/reach-ui/blob/master/packages/combobox/src/index.js#L680.

bryceosterhaus:
@matuzalemsteles @wincent updated and changed it up a bit. Trying to take advantage of returning a handler from a hook and using react's internal event API. Check out the demo on autocomplete and let me know what you think.

bryceosterhaus:
@matuzalemsteles let me know if you have any final thoughts on this. I cleaned it up a bit and made an explicit story for demoing keyboard functionality. 

I'm sure there might be some edge cases we want to iron out or different components this can be applied to, but I think those fixes might be best done in conjunction with the creation of the high-level components in which the use case is more prominent.

matuzalemsteles:
Yes, I think we can go through it as we build the high-level components so we can get more use cases. Just a few points I've noticed before moving on with it.

- I have to press more than once when the active item is on the last list to return to the first one.
- We should reset the active item when DropDown closes or when I click outside of DropDown.

bryceosterhaus:
@matuzalemsteles good idea, Ill go ahead and make those changes before merging.

Issue #2126, opened 2019-06-26
[next.clayui.com] Page visual simplification


marcoscv-work:
![image](https://user-images.githubusercontent.com/7963804/60187876-a83d0680-982e-11e9-8f7f-f680526d8188.png)

PR #2132, opened 2019-06-27
Fix(next.clayui.com): (#2126) Unused var

marcoscv-work:


jbalsas:
Hey @matuzalemsteles, answered my own question about netlify... the `deploy-preview` task gives us the urls per-pull-request... https://deploy-preview-2132--next-clayui.netlify.com/ in this case... which is awesome!! :clap:

The storybook ones seem to hang in there for a while, not sure what's going on there... I'll wait for some more minutes before merging this in. 

As we discussed, this is just an initial step to simplify the information we have and the places we need to tend to. As we have more clarity on where we're going with the project we can grow from this minimal setting :)

PR #2133, opened 2019-06-27
chore(pagination): Adds Pagination Bar story on pagination

diegonvs:
Hey folks, this is just a draft of a composed Pagination Bar component. Makes sense create another package `@clayui/pagination-bar` for hosting a high-level component more detailed or just leave the Pagination Bar as a composition of Pagination, Button and Dropdown?

### DIFF
--- a/packages/clay-pagination/src/ClayPaginationWithBar.tsx
+++ b/packages/clay-pagination/src/ClayPaginationWithBar.tsx
@@ -121,8 +121,10 @@ export const ClayPaginationWithBar: React.FunctionComponent<IProps> = ({
 
 			<div className="pagination-results">
 				{sub(labels.paginationResults, [
-					activePage,
-					Math.ceil(totalItems / perPage),
+					(activePage - 1) * perPage + 1,
+					activePage * perPage < totalItems
+						? activePage * perPage
+						: totalItems,
 					totalItems,
 				])}
 			</div>
--- a/packages/clay-pagination/src/ClayPaginationWithBar.tsx
+++ b/packages/clay-pagination/src/ClayPaginationWithBar.tsx
@@ -12,13 +12,38 @@ import ClayPagination from '../src/index';
 import React, {useState} from 'react';
 import {noop, sub} from '@clayui/shared';
 
-const defaultDeltas = [10, 20, 30, 50];
+const defaultDeltas = [
+	{
+		value: 10,
+	},
+	{
+		value: 20,
+	},
+	{
+		value: 30,
+	},
+	{
+		value: 50,
+	},
+];
+
+interface IDelta {
+	/**
+	 * Path or URL to be used for some SPA focused use cases.
+	 */
+	href?: string;
+
+	/**
+	 * Value of the delta
+	 */
+	value: number;
+}
 
 interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
 	 * Possible values of items per page.
 	 */
-	deltas?: Array<number>;
+	deltas?: Array<IDelta>;
 
 	/**
 	 * The page numbers that should be disabled. For example, `[2,5,6]`.
@@ -68,7 +93,7 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
 	 * Initializes delta. Default is `10`.
 	 */
-	initialSelectedDelta?: number;
+	initialSelectedDelta?: IDelta;
 
 	/**
 	 * The size of pagination element.
@@ -100,7 +125,7 @@ export const ClayPaginationWithBar: React.FunctionComponent<IProps> = ({
 	labels = DEFAULT_LABELS,
 	onDeltaChange = noop,
 	onPageChange = noop,
-	initialSelectedDelta = defaultDeltas[0],
+	initialSelectedDelta = deltas[0],
 	initialActivePage = 1,
 	size,
 	spritemap,
@@ -109,7 +134,7 @@ export const ClayPaginationWithBar: React.FunctionComponent<IProps> = ({
 }: IProps) => {
 	const [active, setActive] = useState<boolean>(false);
 	const [activePage, setActivePage] = useState<number>(initialActivePage);
-	const [perPage, setPerPage] = useState<number>(initialSelectedDelta);
+	const [perPage, setPerPage] = useState<number>(initialSelectedDelta.value);
 
 	return (
 		<div
@@ -136,18 +161,23 @@ export const ClayPaginationWithBar: React.FunctionComponent<IProps> = ({
 				}
 			>
 				<ClayDropDown.ItemList>
-					{deltas.map((delta: number, i) => (
+					{deltas.map(({href, value}: IDelta, i: number) => (
 						<ClayDropDown.Item
 							data-testid={`dropdownItem${i}`}
+							href={href}
 							key={`dropdownItem${i}`}
-							onClick={() => {
-								setPerPage(delta);
-								if (onDeltaChange) {
-									onDeltaChange(delta);
-								}
-							}}
+							onClick={
+								href
+									? undefined
+									: () => {
+											setPerPage(value);
+											if (onDeltaChange) {
+												onDeltaChange(value);
+											}
+									  }
+							}
 						>
-							{sub(labels.selectPerPageItems, [String(delta)])}
+							{sub(labels.selectPerPageItems, [String(value)])}
 						</ClayDropDown.Item>
 					))}
 				</ClayDropDown.ItemList>
--- a/packages/clay-pagination/src/ClayPaginationWithBar.tsx
+++ b/packages/clay-pagination/src/ClayPaginationWithBar.tsx
@@ -15,12 +15,22 @@ import {noop, sub} from '@clayui/shared';
 const defaultDeltas = [10, 20, 30, 50];
 
 interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Possible values of items per page.
+	 */
 	deltas?: Array<number>;
 
 	disabledPages?: Array<number>;
 
+	/**
+	 * The number of pages to show on each side of the active page before
+	 * using an ellipsis dropdown.
+	 */
 	ellipsisBuffer?: number;
 
+	/**
+	 * Function used to create the href provided for each page link.
+	 */
 	hrefConstructor?: (page?: number) => string;
 
 	labels?: {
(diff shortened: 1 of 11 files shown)


packages/clay-pagination/stories/index.tsx
```diff
@@ -53,4 +56,79 @@ storiesOf('ClayPagination', module)
...
+						activePage={activePage}
+						onPageChange={page => page && setActivePage(page)}
+						spritemap={spritemap}
+						totalPages={Math.ceil(numberOfItems / perPage)}
+					/>
+				</div>
+			);
+		};
+
+		return <PaginationBar numberOfItems={500} />;
```
bryceosterhaus: I might expect it would also be helpful to provide an API that looks like...

```
<PaginationBar
	deltas? // array,
	hrefConstructor? // func,
	onDeltaChange? // func,
	onPageChange? // func,
	selectedPage // number,
	selectedDelta // number,
	size? // 'sm' || 'lg',
	totalItems // number
/>
```

`deltas`: This allows for a custom list of numbers. We would have a default as well
`hrefConstructor`: This gives the user the ability to either create URLs for the pagination items
`onDeltaChange` & `selectedDelta`: This helps enforce a controlled mechanism. Also helpful for the parent component to know what the selected delta is.
`onPageChange` & `selectedPage`: This helps enforce a controlled mechanism. Also helpful for the parent component to know what the page is.
`size`: I think there are two sizes right? Or maybe that is just in faro. 
`totalItems`: I prefer 'total' over 'numberOf' since it is a bit more explicit.

matuzalemsteles:
I think I'm also inclined to just have a `@clayui/pagination` package and instead of delivering it as `ClayPaginationBar` would be a `ClayPaginationWithBar` as a high-level.

bryceosterhaus:
@matuzalemsteles what do you think about `<ClayPaginationWithDelta />` since its not necessarily _with_ a bar, its with the ability to change the delta of each page.

matuzalemsteles:
hey @bryceosterhaus using `ClayPaginationWithDelta` maybe would be more certain, but using` Bar` would be the best due that best understood for mere mortals. I see `delta` as something more technical in my vision 🤷‍♂️.

bryceosterhaus:
@matuzalemsteles yeah I was sort of thinking that as well, even though Bar might be a bit strange, for our use case it is likely best because its the name we are used to.

diegonvs:
+1 for `ClayPaginationWithBar`

packages/clay-pagination/stories/index.tsx
diegonvs: Yes, We have three different `sizes`. As documented here: https://next.clayui.com/docs/css/components/pagination.html#sizes but I think that is a good idea include `size` property on low-level :)

I agree with `totalItems` too

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+						<ClayIcon
+							spritemap={spritemap}
+							symbol="caret-double-l"
+						/>
+					</ClayButton>
+				}
+			>
+				<ClayDropDown.ItemList>
+					{deltas.map((item, i) => (
+						<ClayDropDown.Item
```
matuzalemsteles: Maybe we have to modify `ClayDropDown.Item` to default to it being a `button`. The interaction seems strange when it is `span`, semantically wrong.

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+	selectedDelta = defaultDeltas[0].value,
+	selectedPage = 1,
+	size,
+	spritemap,
+	totalItems,
+	...otherProps
+}: IProps) => {
+	const [active, setActive] = useState<boolean>(false);
+	const [activePage, setActivePage] = useState<number>(selectedPage);
+	const [perPage, setPerPage] = useState<number>(selectedDelta);
```
matuzalemsteles: Maybe it's best to make this component uncontrolled. If `selectedDelta` changes it will not reflect here.

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+	onPageChange = noop,
+	selectedDelta = defaultDeltas[0].value,
+	selectedPage = 1,
+	size,
+	spritemap,
+	totalItems,
+	...otherProps
+}: IProps) => {
+	const [active, setActive] = useState<boolean>(false);
+	const [activePage, setActivePage] = useState<number>(selectedPage);
```
matuzalemsteles: The same here.

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+import classNames from 'classnames';
+import ClayButton from '@clayui/button';
+import ClayDropDown from '@clayui/drop-down';
+import ClayIcon from '@clayui/icon';
+import ClayPagination from '../src/index';
+import React, {useState} from 'react';
+import {noop, sub} from '@clayui/shared';
+
+const defaultDeltas = [
```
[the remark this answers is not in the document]
diegonvs: Makes sense, removing!

packages/clay-pagination/src/ClayPaginationWithBar.tsx
diegonvs: We decided to rename these APIs as initializers :)

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+							{sub(labels.selectPerPageItems, [item.label])}
+						</ClayDropDown.Item>
+					))}
+				</ClayDropDown.ItemList>
+			</ClayDropDown>
+
+			<div className="pagination-results">
+				{sub(labels.paginationResults, [
+					activePage,
+					Math.ceil(totalItems / perPage),
```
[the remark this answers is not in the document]
diegonvs: Solved!

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+const defaultDeltas = [10, 20, 30, 50];
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Possible values of items per page.
+	 */
+	deltas?: Array<number>;
+
+	disabledPages?: Array<number>;
```
matuzalemsteles: Add some documentation here.

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+	 * using an ellipsis dropdown.
+	 */
+	ellipsisBuffer?: number;
+
+	/**
+	 * Function used to create the href provided for each page link.
+	 */
+	hrefConstructor?: (page?: number) => string;
+
+	labels?: {
```
matuzalemsteles: Same here

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+		paginationResults: string;
+
+		perPageItems: string;
+
+		selectPerPageItems: string;
+	};
+
+	onDeltaChange?: (page?: number) => void;
```
matuzalemsteles: Same here

packages/clay-shared/src/index.tsx
```diff
@@ -4,7 +4,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+export const noop = () => {};
```
matuzalemsteles: I'm a bit hesitant about this, I do not want us to always use this here, if we're not going to end up having some components with dependency on `shared` that is only using the `noop`.

packages/clay-pagination/stories/index.tsx
```diff
@@ -53,4 +88,90 @@ storiesOf('ClayPagination', module)
 				totalPages={totalPages}
 			/>
 		);
+	})
+	.add('w/ bar', () => {
```
matuzalemsteles: I think it would be best to put `w / bar low-levels` or something of a similar to differentiate with the bass example.

packages/clay-pagination/src/ClayPaginationWithBar.tsx
matuzalemsteles: Just one caveat to this, we must adjust so that the `Button` does not go without the style.

packages/clay-pagination/src/ClayPaginationWithBar.tsx
```diff
@@ -0,0 +1,184 @@
...
+						/>
+					</ClayButton>
+				}
+			>
+				<ClayDropDown.ItemList>
+					{deltas.map((delta: number, i) => (
+						<ClayDropDown.Item
+							data-testid={`dropdownItem${i}`}
+							key={`dropdownItem${i}`}
+							onClick={() => {
```
matuzalemsteles: Maybe this should be optional, because we can use the pagination-bar with links instead of controlling the state here, doing this we are always assuming that the data will be controlled through the client side in the DXP this does not happen so often the vast majority it happens only on the server to change.

So the `deltas` API should accept an API similar to what we had in v2: https://github.com/liferay/clay/blob/2.x/packages/clay-pagination-bar/src/ClayPaginationBar.js#L107-L113

I would even choose to change the `deltas` API to `entries` seems more understanding. We must accept these two use cases.

cc @bryceosterhaus

matuzalemsteles: cc @diegonvs and @bryceosterhaus you can expose your thoughts on this here!

bryceosterhaus: I'd prefer the `deltas` name for API since its more explicit to what it is. `entries` is fairly generic. As far as onClick, I think it'd be ideal to allow either `onClick` type callbacks for more SPA focused use cases and also provide an `href` prop to account for routing use cases.

matuzalemsteles: ok, we can go with `deltas` for now so if we think we are having confusion we can change.

> As far as onClick, I think it'd be ideal to allow either onClick type callbacks for more SPA focused use cases and also provide an href prop to account for routing use cases.

Yeah, we should support both here.

PR #2134, opened 2019-06-27
fix(clay-data-provider): remove metal-uri dependency in favor of URL

bryceosterhaus:
Since we have a [polyfill for searchParams](https://github.com/liferay/liferay-portal/blob/b628b3a178accd00caa4908222c1a122f8b539df/modules/apps/frontend-compatibility/frontend-compatibility-ie/build.gradle#L293-L307) in portal, can we can just use URL instead?

cc: @jbalsas

### DIFF
--- a/packages/clay-data-provider/package.json
+++ b/packages/clay-data-provider/package.json
@@ -30,7 +30,6 @@
 		"classnames": "^2.2.6",
 		"fast-json-stable-stringify": "^2.0.0",
 		"lru-cache": "^5.1.1",
-		"metal-uri": "^3.1.2",
 		"warning": "^4.0.3"
 	},
 	"devDependencies": {
--- a/packages/clay-data-provider/src/useResource.ts
+++ b/packages/clay-data-provider/src/useResource.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import Uri from 'metal-uri';
 import warning from 'warning';
 import {
 	FetchPolicy,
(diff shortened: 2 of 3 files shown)


jbalsas:
We have `URLSearchParams` polyfill, but not `URL`, which is also not supported in IE11. I think that makes no difference though.

I'd say let's go with this and we can add both `URL` and `URLSearchParams` to Liferay Portal for IE11 only, maybe via https://github.com/lifaon74/url-polyfill.

/cc @wincent, @julien?

bryceosterhaus:
@jbalsas oh my bad, I had only looked at `URLSearchParams` which is the `searchParams` of URL, but I wasn't aware that even URL isn't support. Need to do my research next time...

wincent:
> I'd say let's go with this and we can add both `URL` and `URLSearchParams` to Liferay Portal for IE11 only, maybe via https://github.com/lifaon74/url-polyfill.

For reference, the main ticket we have covering expanding our polyfill coverage in IE11 is this one: https://issues.liferay.com/browse/LPS-95425

I'll add a link back here from there.

matuzalemsteles:
hey guys, from what I'm seeing from this discussion, this sounds good, so I'm going ahead with this PR.

kresimir-coko:
If this has been merged does it mean I can use `new URL()` in my utils that I'm modernizing, for example https://github.com/markocikos/liferay-portal/pull/112

wincent:
> If this has been merged does it mean I can use `new URL()` in my utils that I'm modernizing

Just to reflect that parallel discussion that we had on Slack about this, technically, we shouldn't ship anything to liferay-portal that relies on `URL` until the polyfilling for IE is done (it hasn't been yet).

PR #2149, opened 2019-07-01
fix(clay-css): SVG Icons renamed `announcement`, `sticky`…

### LINKED ISSUE #2119: Priority Thread Icons need to be renamed
This issue is related to this ticket and the comment from Victor Valle: https://issues.liferay.com/browse/LEXI-300

We need a better naming for the icons: 

announcement -> megaphone-full
sticky -> pin-full
urgent -> bell-full

pat270:
…, and `urgent` to `megaphone-full`, `pin-full`, and `bell-full` respectively

fix(clay-css): Mixin `clay-get-icons` build Sass map of icons available

fixes #2119 BREAKING CHANGE

https://issues.liferay.com/browse/LEXI-300


wincent:
nit: the "BREAKING CHANGE" bit doesn't need to go in the title, [only in the "body"](https://www.conventionalcommits.org/en/v1.0.0-beta.4/).

marcoscv-work:
so @pat270 @wincent do we need to rename the commit?

wincent:
Renaming it would be nice, for consistency, but it is not the end of the world if you don't do it.

More important, I'd say, is retitling the PR, which you can do with the "Edit" button (no need to `git commit --amend` and force-push). But even that isn't 100% necessary — I'm just thinking forward to the time when we'll use the same script that we are using in the themes toolkit and AlloyEditor etc to generate changelogs (that one expects PR titles to be in "Conventional Commits" format, and if they are, the generated changelog turns out better).

pat270:
@wincent @marcoscv-work I updated the commit message. Thanks!

marcoscv-work:

PR #2159, opened 2019-07-02
chore(breadcrumbs): Adds stories for navigation

diegonvs:


packages/clay-navigation/src/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react';
+import {Breadcrumb} from './Breadcrumb';
+
+const ClayNavigation: React.FunctionComponent<
+	React.HTMLAttributes<HTMLDivElement>
+> & {
+	Breadcrumb: typeof Breadcrumb;
+} = ({className, ...otherProps}) => {
+	return (
+		<div {...otherProps} className={classNames(className)}>
```
bryceosterhaus: Is this just essentially a placeholder for now?

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import classNames from 'classnames';
+import React from 'react';
+
+export interface IBreadcrumbItem {
+	active?: boolean;
+
```
bryceosterhaus: no need for line breaks here unless you are adding jsdocs

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
```diff
@@ -0,0 +1,46 @@
...
+ */
+import ClayButton from '@clayui/button';
+import ClayDropDown from '@clayui/drop-down';
+import ClayIcon from '@clayui/icon';
+import React, {useState} from 'react';
+import {IBreadcrumbItem} from './BreadcrumbItem';
+
+export interface IBreadcrumbEllipsisProps {
+	items?: Array<IBreadcrumbItem>;
+
```
bryceosterhaus: no need for line breaks here unless you are adding jsdocs

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+import warning from 'warning';
+import {BreadcrumbEllipsis} from './BreadcrumbEllipsis';
+import {BreadcrumbItem, IBreadcrumbItem} from './BreadcrumbItem';
+
+interface IProps extends React.HTMLAttributes<HTMLOListElement> {
+	ellipsisBuffer?: number;
```
bryceosterhaus: We will need JSDocs for these props since this is the public facing API

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+	ellipsisProps?: Object;
+
+	items: Array<IBreadcrumbItem>;
+
+	spritemap: string;
+}
+
+const getBufferList = (
+	items: Array<IBreadcrumbItem>,
```
bryceosterhaus: Generally I think we tend to opt for `IBreadcrumbItem[]` for array types. I may be wrong, or we might need to add an eslint rule for this.

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+	start: number,
+	end: number,
+	ellipsisProps: Object,
```
bryceosterhaus: instead of an explicit prop here, can we just use `...otherProps` since bufferList implies it is for an ellipsis

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+		: removedItems;
+};
+
+const getBreadcrumbItems = (
+	ellipsisBuffer: number,
+	items: Array<IBreadcrumbItem>,
+	ellipsisProps: Object,
+	spritemap: string
+) => {
+	const activeIndex = items.findIndex(item => item.active);
```
bryceosterhaus: Im not sure this is supported in IE11. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Browser_compatibility

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+	...otherProps
+}: IProps) => {
+	warning(
+		items.findIndex(item => item.active) != -1,
+		'ClayNavigation.Breadcrumb expects at least one `active` item on `items`.'
+	);
+
+	return (
+		<ol {...otherProps} className={classNames('breadcrumb', className)}>
+			{(ellipsisBuffer
```
bryceosterhaus: could we move this logic outside of the return just to be a little more clear? We could set it to `breadCrumbItems` and then just only have to use `breadCrumbItems.map(`. I think this will help readability a bit.

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
```diff
@@ -0,0 +1,46 @@
...
+	spritemap: string;
+}
+
+export const BreadcrumbEllipsis: React.FunctionComponent<
+	IBreadcrumbEllipsisProps
+> = ({items, spritemap, ...otherProps}) => {
+	const [active, setActive] = useState(false);
+	return (
```
bryceosterhaus: line break above this. Essentially we put line breaks where "order is important", so if we move the return above the `const`, it'd break.

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+						items,
+						ellipsisProps,
+						spritemap
+				  )
+				: items
+			).map((item: IBreadcrumbItem | React.ReactElement, i: number) =>
+				React.isValidElement(item) ? (
+					React.cloneElement(item, {key: `ellipsis${i}`})
+				) : (
+					<BreadcrumbItem
```
bryceosterhaus: What about `onClick` instead of href? I could imagine that would be used as well.

packages/clay-navigation/src/__tests__/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+			<ClayNavigation.Breadcrumb
+				ellipsisBuffer={1}
+				ellipsisProps={{style: {fontSize: '15px'}}}
+				items={items}
+				spritemap="path/to/spritemap"
+			/>
+		);
+
+		expect(container).toMatchSnapshot();
+	});
```
bryceosterhaus: Can we add some more tests for interactions, such as dropdown interactions. Also, if we add `onClick` functionality, we could test that interaction as well.

packages/clay-navigation/src/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import classNames from 'classnames';
+import React from 'react';
+import {Breadcrumb} from './Breadcrumb';
+
+const ClayNavigation: React.FunctionComponent<
+	React.HTMLAttributes<HTMLDivElement>
+> & {
+	Breadcrumb: typeof Breadcrumb;
+} = ({className, ...otherProps}) => {
+	return (
```
matuzalemsteles: I would do the same thing we are doing for `ClayForm`, the content of this is not necessary.

packages/clay-navigation/stories/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+				},
+				{
+					href: '#11',
+					label: 'Eleven',
+				},
+			]}
+			spritemap={spritemap}
+		/>
+	))
+	.add('should throw a warning', () => (
```
matuzalemsteles: I didn't quite understand why.

packages/clay-navigation/src/Breadcrumb.tsx
matuzalemsteles: Just one comment would not be enough?

packages/clay-navigation/src/Breadcrumb.tsx
matuzalemsteles: I think they are just different ways of doing this, I like `Array<...>` because it is more explicit to me and goes along with the other syntaxes like `React.FunctionComponent<...>`.

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,122 @@
...
+					items={removedItems}
+					key="breadcrumbEllipsis"
+					spritemap={spritemap}
+					{...ellipsisProps}
+				/>,
+		  ]
+		: removedItems;
+};
+
+const getBreadcrumbItems = (
```
matuzalemsteles: I see this and `getBufferList` is akin to the pagination rule, so I would choose to try to use the same function for the two and add some new params to customize some components... I think it would be best not to have rules duplicated in different places.

packages/clay-navigation/src/Breadcrumb.tsx
matuzalemsteles: I think we can keep both. But we have to think if keeping `href` is probably for cases that would have been a navigation using the browser and not a navigation through the React Router for example. One thought about this is:

- If you allow `href`, it may be cases where people will only use the component as if it were a taglib. Maybe I was wrong, but in v2 we added the href for those cases.
- If you allow `onClick`, it may be cases where you have control and want to do a navigation using the React Router for example.

So the question is, can we just allow `onClick` and force people to use components for full React applications rather than for a loose component, such as a taglib. But if you follow this path it can not be used in a taglib.

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
```diff
@@ -0,0 +1,46 @@
...
+			trigger={
+				<ClayButton className="breadcrumb-link" displayType="unstyled">
+					<ClayIcon spritemap={spritemap} symbol="ellipsis-h" />
+					<ClayIcon spritemap={spritemap} symbol="caret-bottom" />
+				</ClayButton>
+			}
+			{...otherProps}
+		>
+			<ClayDropDown.ItemList>
+				{items &&
```
matuzalemsteles: I'd rather leave the default value of items an empty array and so we do not need the condition here.

packages/clay-navigation/src/Breadcrumb.tsx
bryceosterhaus: @matuzalemsteles yeah thats fine, we likely should just pick one or the other and stick with it for consistency

packages/clay-navigation/src/Breadcrumb.tsx
bryceosterhaus: For a react app with routing, I think the user would also just use an `href` and then let the touer handle changing the state. The more I think about it, the more `onClick` is likely used less and maybe wouldn't be used at all.

bryceosterhaus: So I guess im actually thinking the opposite, and only allowing href and keep them as anchor tags

packages/clay-navigation/src/Breadcrumb.tsx
matuzalemsteles: Sure, I think we can only use `Array<...>` for consistency, is there any way to add that in lint?

packages/clay-navigation/src/Breadcrumb.tsx
matuzalemsteles: Oh no 😅, I'm not sure if `href` would help with the React Router because you need to use the `Link` component (at least that's how I used it 🙂).

So I would probably use `onClick` and then call `navigate` (I think this is Reach Router 🙂).

packages/clay-navigation/src/Breadcrumb.tsx
bryceosterhaus: Yeah I think we can add, https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md#generic

packages/clay-navigation/src/Breadcrumb.tsx
bryceosterhaus: Oh good point with `Link`. Maybe we need to add ability to pass in a custom component as well, since that is likely a important use case.

The reason I would want to avoid `onClick` only is because you wouldn't be able to right click open or hold cmd and click to open in a new tab. href is important for accessibility.

matuzalemsteles: > Oh good point with Link. Maybe we need to add ability to pass in a custom component as well, since that is likely a important use case.

Hmm, I had not thought about it, but I find it interesting.

> The reason I would want to avoid onClick only is because you wouldn't be able to right click open or hold cmd and click to open in a new tab. href is important for accessibility.

Yes you are right, we can keep both, I think both end up being important.

packages/clay-navigation/src/index.tsx
diegonvs: Nope, I miss it.

packages/clay-navigation/src/BreadcrumbItem.tsx
diegonvs: Solved!

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
diegonvs: Solved!

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: Added 
```javascript
/**
 * The number of Breadcrumb Items to show on each side of the active Breadcrumb Item before
 * using an ellipsis dropdown.
 */
```

It's okay?

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: I vote for Array<anything> too :)

diegonvs: https://github.com/liferay/clay/pull/2169

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: `otherProps` is already being used on `<ol>` root

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: Replacing findIndex with a filter length.

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: Makes sense :)

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
diegonvs: I didn't know that. Thanks for the tip!

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: Was added a property called `onItemClicked`

packages/clay-navigation/src/__tests__/index.tsx
diegonvs: For sure

packages/clay-navigation/stories/index.tsx
diegonvs: Removed.

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: Done!

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
diegonvs: Done!

packages/clay-pagination/src/index.tsx
```diff
@@ -157,7 +105,7 @@ const ClayPagination: React.FunctionComponent<IProps> = ({
 						href={
 							hrefConstructor && hrefConstructor(page as number)
 						}
-						key={page as number}
+						key={`page${index}` as string}
```
matuzalemsteles: I think it would be better for the key to be `page`. It may happen that `page${index}` change when any item becomes a DropDown, page does not.

packages/clay-pagination/src/__tests__/index.tsx
```diff
@@ -118,7 +118,10 @@ describe('ClayPagination', () => {
 
 		fireEvent.click(getByText(container, '...') as HTMLElement, {});
 
-		fireEvent.click(getByText(document.body, '4') as HTMLElement, {});
+		fireEvent.click(
```
matuzalemsteles: Out of curiosity, why did you change that?

diegonvs: Source formatting

diegonvs: Source formatting

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,84 @@
...
+	 * Use this property for defining `otherProps` that will be passed to ellipsis dropdown trigger.
+	 */
+	ellipsisProps?: Object;
+
+	/**
+	 * Property to define Breadcrumb's items.
+	 */
+	items: Array<IBreadcrumbItem>;
+
+	spritemap: string;
```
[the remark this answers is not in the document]
diegonvs: For `spritemap`? In some components, we left out JSDoc of `spritemap`

packages/clay-pagination/src/index.tsx
diegonvs: What do you think about `paginationItem${index}`?

diegonvs: Solved internally :)

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+					data-testid={`testId${label}`}
+					href={href}
+					role="button"
+				>
+					{label}
+				</a>
+			);
+		}
+
+		if (onItemClick) {
```
matuzalemsteles: I think we can do the same here that was done in `href` just render the `ClayButton` if the item is not active.

packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,86 @@
...
+				BreadcrumbEllipsis,
+				ellipsisProps,
+				items.indexOf(activeItems[0]),
+				spritemap
+		  )
+		: items;
+
+	return (
+		<ol {...otherProps} className={classNames('breadcrumb', className)}>
+			{breadCrumbItems.map((item: any, i: number) =>
```
bryceosterhaus: Can we fix this `any` type here?

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
```diff
@@ -0,0 +1,61 @@
...
+import ClayDropDown from '@clayui/drop-down';
+import ClayIcon from '@clayui/icon';
+import React, {useState} from 'react';
+import {IBreadcrumbItem} from './BreadcrumbItem';
+
+export interface IBreadcrumbEllipsisProps {
+	/**
+	 * Property to define BreadcrumbEllipsis's items.
+	 */
+	items?: Array<IBreadcrumbItem>;
```
bryceosterhaus: Is this really an optional param? Meaning, would we ever expect or want this component not to receive items?

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * Path to the location of the spritemap resource.
+	 */
+	spritemap: string;
```
bryceosterhaus: This should be optional since spritemap can come via context

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+	/**
+	 * Label of the Breadcrumb item
+	 */
+	label: string;
+
+	/**
+	 * Callback for when a Breadcrumb item is clicked.
+	 */
+	onItemClick?: (event: React.SyntheticEvent) => void;
```
bryceosterhaus: Since its already called a `BreadcrumbItem`, can we just call this `onClick` to reduce name redundancy?

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export const BreadcrumbItem = ({
+	active,
+	href,
+	label,
+	onItemClick,
+	...otherProps
+}: IBreadcrumbItem) => {
+	const ItemElement = ({
```
bryceosterhaus: How come this is defined separately as its own function? We are only calling this once, which means we can likely just move this to the jsx markup

packages/clay-shared/src/getEllipsisItems.tsx
```diff
@@ -0,0 +1,79 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import React from 'react';
+
+const getBufferList = (
```
bryceosterhaus: Instead of having a ton of arguments, lets just make this accept a config object. Usually is there are more than 2 or 3 args, its easier to pass in an object since its more explicit.

packages/clay-shared/src/getEllipsisItems.tsx
```diff
@@ -0,0 +1,79 @@
...
+	start: number,
+	end: number,
+	EllipsisComponent: React.ComponentType<any>,
+	ellipsisProps?: Object,
+	spritemap?: string
+): Array<Object | number | React.ReactElement> => {
+	const removedItems = items.slice(start, Math.max(end, start));
+
+	if (removedItems.length > 1) {
+		return [
```
bryceosterhaus: How come we are returning an array here?

packages/clay-shared/src/getEllipsisItems.tsx
```diff
@@ -0,0 +1,79 @@
...
+				{...ellipsisProps}
+			/>,
+		];
+	}
+
+	return removedItems;
+};
+
+export const getEllipsisItems = (
+	ellipsisBuffer: number,
```
bryceosterhaus: Same as above, lets move all these args into a config object.

packages/clay-navigation/src/Breadcrumb.tsx
diegonvs: Fixed!

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
diegonvs: I left this optional due I was initializing with an empty array and I was taking as a base PaginationEllipsis that this property is defined as optional. Perhaps, make sense turning this one obligatory.

packages/clay-navigation/src/BreadcrumbEllipsis.tsx
diegonvs: Makes sense. I'm currently updating all components that receive `spritemap` to be optional and update docs.

packages/clay-navigation/src/BreadcrumbItem.tsx
diegonvs: Refactoring, I don't remember why I did that but possibly this problem was solved.

packages/clay-navigation/src/BreadcrumbItem.tsx
diegonvs: Yes, makes sense.

packages/clay-shared/src/getEllipsisItems.tsx
diegonvs: Done!

packages/clay-shared/src/getEllipsisItems.tsx
diegonvs: Since the return must be an array, makes sense encapsulate the component in an array previously.

packages/clay-shared/src/getEllipsisItems.tsx
diegonvs: Done!

packages/clay-navigation/stories/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+import '@clayui/css/lib/css/atlas.css';
+import ClayNavigation from '../src';
+import React from 'react';
+import {number} from '@storybook/addon-knobs';
+import {storiesOf} from '@storybook/react';
+
+const spritemap = require('@clayui/css/lib/images/icons/icons.svg');
+
+storiesOf('ClayNavigation', module)
+	.add('default', () => (
```
matuzalemsteles: You can rename it to `/w breadcrumb`? Since we can have several other components here.

packages/clay-navigation/stories/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+				},
+				{
+					href: '#11',
+					label: 'Eleven',
+				},
+			]}
+			spritemap={spritemap}
+		/>
+	))
+	.add('without anchors', () => {
```
matuzalemsteles:
```suggestion
	.add('/w breadcrumb without anchors', () => {
```

packages/clay-navigation/stories/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+import '@clayui/css/lib/css/atlas.css';
+import ClayNavigation from '../src';
+import React from 'react';
+import {number} from '@storybook/addon-knobs';
+import {storiesOf} from '@storybook/react';
+
+const spritemap = require('@clayui/css/lib/images/icons/icons.svg');
+
+storiesOf('ClayNavigation', module)
```
matuzalemsteles: I would be curious to know how `Breadcrumbs` works when controlled with `onClick`, we may have an example of usage just like pagination.

packages/clay-navigation/stories/index.tsx
diegonvs: makes sense. Maybe in the future, we can create a folder here.

packages/clay-navigation/stories/index.tsx
diegonvs: Sent!

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+			<a
+				className="breadcrumb-link"
+				data-testid={`testId${label}`}
+				href={href}
+				role="button"
+			>
+				{label}
+			</a>
+		) : onClick ? (
+			<ClayButton
```
matuzalemsteles: I was observing one thing when I see the buttons the mark up should be as follows:

```jsx
<ClayButton
	className="breadcrumb-link"
	data-testid={`testId${label}`}
	displayType="unstyled"
	onClick={onClick}
	title={label}
>
	<span className="breadcrumb-text-truncate">
		{label}
	</span>
</ClayButton>
```

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+		{!active && href ? (
+			<a
+				className="breadcrumb-link"
+				data-testid={`testId${label}`}
+				href={href}
+				role="button"
+			>
+				{label}
+			</a>
+		) : onClick ? (
```
matuzalemsteles:
```suggestion
		) : !active && onClick ? (
```

packages/clay-navigation/src/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+		{...otherProps}
+	>
+		{!active && href ? (
+			<a
+				className="breadcrumb-link"
+				data-testid={`testId${label}`}
+				href={href}
+				role="button"
+			>
+				{label}
```
matuzalemsteles:
```jsx
<span className="breadcrumb-text-truncate">
    {label}
</span>
```

Issue #2137, opened 2019-06-28
Move form related elements to @clayui/form package


bryceosterhaus:
`RadioGroup`, `Checkbox`, `Autocomplete` are all components that should be rolled into the `@clayui/form` package since they are all form related html elements. Multi-select is another component that is going into `@clayui/form`.

Happy to have my opinion on this changed though.

matuzalemsteles:
hey @bryceosterhaus I've been thinking more about this, maybe it makes sense to move `@clayui/date-picker`, `@clayui/color-picker` and `@clayui/time-picker` are essentially part of a Form and can be composed by others in the case of their input. What do you think?

wincent:
As some level, all web apps *are* forms.  😂

A good rule of thumb for whether something should actually live inside the `form` package might be whether it corresponds, at least conceptually, to an HTML element that falls into [the "form content" category](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Form-associated_content); copy-pasting from that MDN page, the list is:

- `<button>`
- `<fieldset>`
- `<input>`
- `<keygen>`
- `<label>`
- `<meter>`
- `<object>`
- `<output>`
- `<progress>`
- `<select>`
- `<textarea>`

bryceosterhaus:
> As some level, all web apps are forms.

I started to realize this with the PR I was working on. I agree in that we should probably only add html elements that fall into the form content category, however something like `<button>` makes sense to be a separate package to me, since often in a webapp we use it as an action and not necessarily always a form submission.

Overall, I would lean towards avoiding a hard and fast rule since we could probably end up just adding everything. For all of the picker components, I think I would opt to keep them separate since they are fairly complex. This nuance is the main reason I decided to start small and only [move checkbox and radio for now](https://github.com/liferay/clay/pull/2162)

matuzalemsteles:
Makes sense, in letting separate the more complex components. I think we can go back to this whenever we have a real need and benefit.

bryceosterhaus:
going to close for now. Possibly re-visit in future if we see a need though.

PR #2162, opened 2019-07-03
refactor: move checkbox and radiogroup into form

bryceosterhaus:
Addresses this, https://github.com/liferay/clay/issues/2137.

Not totally sold on this, but curious to get everyones thoughts on movign checkbox and radio into form. I left autocomplete out since it isn't necessarily a form element by itself.


packages/clay-form/src/__tests__/RadioGroup.tsx
```diff
@@ -5,7 +5,8 @@
  */
 
 import * as TestRenderer from 'react-test-renderer';
-import ClayRadioGroup from '..';
+import ClayRadio from '../Radio';
+import ClayRadioGroup from '../RadioGroup';
```
matuzalemsteles: I've been thinking about this, I do not know if it's very interesting to have a `ClayForm.RadioGroup` and `ClayForm.Radio`... So I'm thinking of the following, maybe we can have a generic `ClayForm.Group` that can handle the features of a `RadioGroup` and the simplest cases of a `Group`. Maybe it can be used up together with a group of checkboxes. What do you think?

bryceosterhaus: Not sure it makes sense to use `Group` for both use cases since `RadioGroup` has custom logic and props that a normal group wouldn't have. `RadioGroup` clones each child element and also allows controls the value of the selected radio.

matuzalemsteles: Yes I thought about this problem, I was just wondering if we could make it configurable, because soon we will have a `Group` for a use case of markup I predict ... so it would have more configurations and also seems a bit strange terms `Radio `and `RadioGroup`.

bryceosterhaus: "Radio group" is pretty common in terms of naming for what we are aiming for here. Even mdn docs reference it as a radio group. I think its best to keep it separate from a normal group we define via clay-css.

matuzalemsteles: Okay, makes sense. Let's get on with it. We can review this at any time.

PR #2166, opened 2019-07-03
feat(clay-dropdown): add the API `useSearchableWithForm` to ClayDropdownBase

### LINKED ISSUE #2164: ClayDropdown with `searchable`: breaks when inside a form tag.
This is a different case from the Forms team, the DropDown is within the context of a form and when it is rendered by the server using the taglibs, the searchable is inside a `<form>` tag and the DropDown is inside another tag `<form>` this is breaking some rules.

The quick hint here is to add a new API, like `wrapSearchable` or `useSearchableWithForm`, which says whether you want to wrap the input with a form or not.

matuzalemsteles:
Fixes #2164


matuzalemsteles:
I'm moving on with this to make a release.

PR #2167, opened 2019-07-03
fix(@clayui/modal): stop closing the modal when the element is removed from the modal content

### LINKED ISSUE #2165: Bug: Clay modal closes when clicking on an element that removes itself
This is with @clayui/modal v3.0.0-milestone.1

**Demo:** https://codesandbox.io/s/clay-modal-close-bug-3rhwc

**To reproduce:**
1. Open modal
1. Click on one
1. Click on two

The "Two" button onClick hides itself which causes the modal to close even though it doesn't explicitly call the close function. I believe the modal is closing since it thinks the click happened outside of the modal which automatically closes the modal.

matuzalemsteles:
Fixes #2165

~~This was a solution that I found most appropriate at the moment, since the "document" becomes the `modal-backdrop` for the case of the modal I think it is safe to check if the click was in it so we could close the modal. But leave your thoughts.~~

### DIFF
--- a/packages/clay-modal/src/Hook.ts
+++ b/packages/clay-modal/src/Hook.ts
@@ -82,7 +82,8 @@ const useUserInteractions = (
 		if (
 			elementRef.current &&
 			event.target !== null &&
-			!elementRef.current.contains(event.target as HTMLDivElement)
+			!elementRef.current.contains(event.target as HTMLDivElement) &&
+			document.contains(event.target as HTMLElement)
 		) {
 			onClick();
 		}


packages/clay-css/src/scss/components/_modals.scss
```diff
@@ -1,3 +1,7 @@
+.modal {
+	pointer-events: none;
```
matuzalemsteles: hey @pat270 let me know if I'm adding in the right place or if we will have implications with that.

matuzalemsteles: After conversation with @pat270 this would not be recommended as it can not work with bootstrap modal plugin. So I followed another, simpler approach that solves the problem 🙂.

matuzalemsteles:
@bryceosterhaus is ready for a review.

packages/clay-modal/src/Hook.ts
```diff
@@ -82,7 +82,8 @@ const useUserInteractions = (
 		if (
 			elementRef.current &&
 			event.target !== null &&
-			!elementRef.current.contains(event.target as HTMLDivElement)
+			!elementRef.current.contains(event.target as HTMLDivElement) &&
+			document.contains(event.target as HTMLElement)
```
bryceosterhaus: This would always evaluate to `true` if the element exists in the DOM right?

matuzalemsteles: Yes, this ends up covering the case of when the element was removed from the DOM and is nowhere else, as in the case of #2165.

PR #2169, opened 2019-07-04
style(typescript): prevent `type[]` and use `Array<type>`

diegonvs:
As discussed here: https://github.com/liferay/clay/pull/2159#discussion_r300099627

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,6 +20,7 @@
 		"sort-imports-es6-autofix"
 	],
 	"rules": {
+		"@typescript-eslint/array-type": [2, "generic"],
 		"@typescript-eslint/no-unused-vars": 2,
 		"@typescript-eslint/interface-name-prefix": ["error", "always"],
 		"no-unused-vars": 0,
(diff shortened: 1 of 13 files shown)


diegonvs:
Hey all, would be a good idea to discuss it on https://github.com/liferay/liferay-frontend-guidelines? I don't see Typescript things there.

matuzalemsteles:
It does not make sense to keep it there because we do not approve the use of Typescript and this is a long speech 🙂, so I prefer to keep it here until we have a well defined guide. Everything is fine at the beginning yet...

wincent:
> It does not make sense to keep it there because we do not approve the use of Typescript and this is a long speech 🙂, so I prefer to keep it here until we have a well defined guide. Everything is fine at the beginning yet...

I actually think it does make sense to have a document on TypeScript over in the liferay-frontend-guidelines, because the question of "can I use TypeScript?" keeps coming up. At this point the document would really only have to answer that question (ie. saying that we don't broadly support it yet, but are using it in Clay v3, and will evaluate wider usage in the future). It wouldn't need to go into details of things like `Array<>` vs `[]`. I made [a PR](https://github.com/liferay/liferay-frontend-guidelines/pull/56) for that.

Issue #2116, opened 2019-06-24
[next.clayui.com] Use the correct focus style


marcoscv-work:
Use the new accessible focus style for page elements like:
- Logo
- Horizontal navigation
- Vertical navigation
- etc.

![image](https://user-images.githubusercontent.com/7963804/60011666-987cc100-967a-11e9-82f4-c55d00f7a615.png)

PR #2173, opened 2019-07-05
Fix(next.clayui.com): Add focus state in sidebar (#2116)

marcoscv-work:


eduardoallegrini:
Just started reviewing :)

PR #2175, opened 2019-07-05
 refactor(@clayui/time-picker): rewrites the timer to mimic browser time behavior

### LINKED ISSUE #2106: Rewrite ClayTimePicker with new markup
We have to rewrite Time Picker to mimic the native browser picker, see the new markup in #2083 (#1424).

matuzalemsteles:
Fixes #2106

I'm not quite sure about some behaviors mainly for the case of the `AM/PM` picker, it does not seem so intuitive. I think there is room for some internal improvements in the code, so leave your thoughts. I've tried at most to mimic the behaviors of the native time picker, so it ends up having lots of use cases.

I have not done the tests yet nor the documentation, as soon as we are well with it, I will do.


packages/clay-date-picker/src/TimePicker.tsx
```diff
@@ -24,30 +23,44 @@ const ClayDatePickerTimePicker: FunctionComponent<IProps> = ({
 	timeFormat,
 	timezone,
 }) => {
+	const [values, setValues] = useState<Input>({
+		ampm: '--',
```
bryceosterhaus: nitpick: could just make a constant fro all the uses of `'--'`

packages/clay-date-picker/src/TimePicker.tsx
```diff
@@ -24,30 +23,44 @@ const ClayDatePickerTimePicker: FunctionComponent<IProps> = ({
...
+		const date = moment(currentTime, timeFormat);
+		const hours = values.hours === '--' ? '--' : date.hour();
+		const minutes = values.minutes === '--' ? '--' : date.minutes();
+
+		setValues(prevValues => ({
+			...prevValues,
+			hours: String(hours),
+			minutes: String(minutes),
+		}));
+	}, [currentTime]);
```
bryceosterhaus: we should probably also check for `timeFormat` here as well.

packages/clay-time-picker/src/index.tsx
```diff
@@ -4,57 +4,422 @@
...
+		if (intrinsicValue > config.max) {
+			return String(config.min);
+		} else if (intrinsicValue < config.min) {
+			return String(config.max);
+		}
+
+		return newValue;
+	};
+
+	const handleKeyDown = (
```
bryceosterhaus: It might beneficial to abstract this into a hook. Probably not needed for this PR, but it might be nice in the future.

packages/clay-time-picker/src/index.tsx
```diff
@@ -4,57 +4,422 @@
...
+		};
+	}, []);
+
+	const visibleActionReset =
+		actionVisible &&
+		(values.ampm !== DEFAULT_VALUE ||
+			values.hours !== DEFAULT_VALUE ||
+			values.minutes !== DEFAULT_VALUE);
+
+	return (
```
bryceosterhaus: Theres quite a bit of logic and function creation above the actual return, we might want to further abstract to help readability of the logic flow.

packages/clay-time-picker/src/index.tsx
```diff
@@ -4,57 +4,422 @@
...
+								setActionVisible(false);
+								setIsFocused(false);
+							}
+						}}
+						ref={elementRef}
+					>
+						<div className="clay-time-edit">
+							<input
+								className="clay-time-hours form-control-inset"
+								id="timePickerHours"
```
bryceosterhaus: I'm hesitant to add hard coded `id`'s like this, especially since someone may want two time pickers on the same page and then the id's would clash.

packages/clay-time-picker/src/index.tsx
```diff
@@ -4,57 +4,422 @@
...
+								maxLength={2}
+								name="hours"
+								onFocus={() => {
+									elementFocusedRef.current = {
```
bryceosterhaus: since this ref isnt actually an element, what do you think about calling it just `focusedRef`?

packages/clay-time-picker/src/index.tsx
```diff
@@ -4,57 +4,422 @@
...
+									)
+								}
+								ref={hoursRef}
+								type="text"
+								value={values.hours}
+							/>
+							<span className="clay-time-divider">{':'}</span>
+							<input
+								className="clay-time-minutes form-control-inset"
+								id="timePickerMinutes"
```
bryceosterhaus: ^ see above comment about ids

packages/clay-time-picker/src/index.tsx
```diff
@@ -4,57 +4,422 @@
...
+								maxLength={2}
+								name="minutes"
```
bryceosterhaus: Same sort of argument about `id`s could be applied to `name` attributes as well, hesitant to hard code these values for a library.

packages/clay-time-picker/src/index.tsx
matuzalemsteles: I'm not so sure, maybe we can create a hook that can handle focus, something like I'm seeing being done at https://github.com/facebook/react/pull/15849, I'm trying to get close to that with #1987 , but we will probably still maintain this function because we have things unique only to that component.

packages/clay-time-picker/src/index.tsx
matuzalemsteles: That makes sense but anyway we need to make this configurable via prop.

packages/clay-time-picker/src/index.tsx
bryceosterhaus: ah gotcha, sounds good.

matuzalemsteles:
hey @bryceosterhaus I made some changes here. Ready for a next look.

matuzalemsteles:
Yeah, I'm going to add the tests still in that PR and documents.

matuzalemsteles:
hey @bryceosterhaus just a status on this. I needed to modify several rules in the Date Picker so it works fine with Time Picker, I'm having to remove the `timeFormat` API from the Date Picker for everything to work fine, but I still have to refactor some things. This is taking longer than I expected 🙂.

matuzalemsteles:
hey @bryceosterhaus, I think it's fine now to take a look, I made some modifications to DatePicker.

packages/clay-date-picker/src/Hooks.ts
```diff
@@ -35,18 +35,34 @@ const useCurrentTime = (format: string) => {
...
 	);
 
-	function setCurrentTime(hours: number, minutes: number): void {
-		set(
-			moment()
+	function setCurrentTime(
+		hours: number | string,
+		minutes: number | string
+	): void {
+		const currentTime = [hours, minutes];
```
bryceosterhaus: Whats the advantage of using an array here? Wouldn't it be simpler to just do something like

```js
function setCurrentTime(
	hours: number | string,
	minutes: number | string
): void {
	if (typeof hours !== 'string') {
		hours = moment()
			.set('h', hours)
			.format('H');
	}

	if (typeof minutes !== 'string') {
		minutes = moment()
			.set('m', minutes)
			.format('m');
	}

	set(`${hours}:${minutes}`);
}

```

packages/clay-date-picker/src/TimePicker.tsx
```diff
@@ -35,27 +32,28 @@ const ClayDatePickerTimePicker: FunctionComponent<IProps> = ({
...
-		const minutes =
-			values.minutes === DEFAULT_VALUE ? DEFAULT_VALUE : date.minutes();
+		const time = currentTime.split(':');
 
 		setValues(prevValues => ({
 			...prevValues,
-			hours: String(hours),
-			minutes: String(minutes),
+			hours: String(time[0]),
+			minutes: String(time[1]),
```
bryceosterhaus: I'm a little hesitant at the back and forth use between `String()` and `Number()`. Any way we can keep just one type consistent?

matuzalemsteles: Yeah, I do not like it either, but it was necessary because in the end we're using the moment and the type is `number`, I could leave everything `string` to avoid it everywhere but I would have this problem there.

packages/clay-date-picker/src/Hooks.ts
matuzalemsteles: Yeah this really is not necessary, I ended up putting it there to build other cases but I just forgot to remove that logic, thanks for that.

matuzalemsteles:
hey @pat270 we can have some class we can add as select in the input when they are focused, just to try to get closer to the behavior of the native time picker.

packages/clay-date-picker/src/TimePicker.tsx
bryceosterhaus: ah gotcha, makes sense

matuzalemsteles:
@bryceosterhaus I added here the `useFocusManagement` hook and I organized the commits. @pat270 will work on adding the `focus` class to the inputs with `form-control-inset`. So I guess we can go without that and add it later.

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
```diff
@@ -357,13 +357,9 @@ describe('IncrementalInteractions', () => {
...
 			const input: any = getByLabelText(ariaLabels.input);
-			const inputTimePicker: any = getByTestId('time-picker-input');
+			const hoursEl = getByTestId('hours') as HTMLInputElement;
 
-			fireEvent.change(inputTimePicker, {
-				target: {
-					value: '00:20',
-				},
-			});
+			fireEvent.keyDown(hoursEl, {key: '10'});
```
bryceosterhaus: nitpick, could you store the key value as a constant? Otherwise, I have no idea what key '10' is.

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
```diff
@@ -381,17 +377,13 @@ describe('IncrementalInteractions', () => {
...
+			const hoursEl = getByTestId('hours') as HTMLInputElement;
 
 			fireEvent.click(dayNumber);
 
-			fireEvent.change(inputTimePicker, {
-				target: {
-					value: '00:20',
-				},
-			});
+			fireEvent.keyDown(hoursEl, {key: '10'});
```
bryceosterhaus: ^

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
matuzalemsteles: Not sure about this, the key is just the number 10. 🙂

bryceosterhaus: does 10 specify a certain key? Like "enter" or "tab"?

bryceosterhaus: If not, we could also do something like `const ANY_KEY_CODE = '10'`

matuzalemsteles: Oh no, for that case I'm just wondering what number was pressed and not the key number. Maybe you are confusing it with `keyCode`.

bryceosterhaus: ahh yeah im mistaking it with keyCode

PR #2182, opened 2019-07-08
chore(table): Update snapshots

### LINKED ISSUE #2181: ClayTable bottom border doesn't work for some use cases without positioning `table-responsive` classes in a parent `div`
Maybe this error is being caused by https://github.com/twbs/bootstrap/issues/24638.
You can reproduce the error taking a look on https://storybook-clayui.netlify.com/?path=/story/claytable--default

Disabling the following styles from `table-responsive` we can fix this error:

```css
display: block;
overflow-x: auto;
```

/cc @pat270 @brunofarache

diegonvs:
Fixes #2181


bryceosterhaus:
LGTM

Issue #2141, opened 2019-06-28
high-level InputWithAutocomplete


bryceosterhaus:
Sample API?
```jsx
<Form.ClayInputWithAutocomplete
	items // array of strings
	match // regex or function?
	onChange // func
	value // string
/>
```

bryceosterhaus:
https://github.com/liferay/clay/pull/2186

Issue #2152, opened 2019-07-01
Extensibility for Multi-Select component


bryceosterhaus:
Conversation originates here, https://github.com/liferay/clay/pull/2123#issuecomment-507420983.

TLDR; how can we easily add extension points for multi-select without customizing the API to specifically for each individual use case. See https://docs.google.com/document/d/1cIx6RkWNbnho_fOSR__3Fp5alIBE_ObhO-jecfXlA9I/edit#heading=h.vtnk7gaj2m17 for extension points in multi-select

jbalsas:
This is a conversation I started with @wincent offline some weeks ago. I don't think there's a lot of room for innovation here.

We probably need to follow a `*Renderer` approach where the Renderer prop can take:
- A component. This works in the case where the developer is "instantiating" the component directly and thus can inject its own component.
- A string referencing a component in a component registry. This covers the case where the component (multi-select) is instantiated somewhere else and the developer is merely adding her renderer and configuring the component without access to the component initialization.

This case would be the same for Tables and other high level components.

matuzalemsteles:
> A string referencing a component in a component registry. This covers the case where the component (multi-select) is instantiated somewhere else and the developer is merely adding her renderer and configuring the component without access to the component initialization.

Hmm, that sounds interesting, I've been thinking something similar a couple of weeks ago as well, but I'm not sure I understand what you meant by the `component registry`... but I think it resembles what I'm thinking.

In high-level components, you typically do not take advantage of `children` because they are configured via props, but at the same time still holds markup rules that people can not touch, maybe that's natural 🤷‍♂️. But we can work on a strategy for the high-level components that need to be touched, for example:

```jsx
<DropDownWithPagination>
  <MyCheckbox override="checkbox" />
  <MyDropDown.ItemList override="item-list" />
  <MyDropDown.Item override="item" />
</DropDownWithPagination>
```

You can replace low-level components or other components within the high-level components, they can still receive the same properties that low-levels are receiving and follow with the idea of ​​keeping a registry of the internal components to replace, the `override` would be used to reference this.

Perhaps in terms of design this does not look so beautiful, but it can soften the creation of several `*renderer` props for some components that may have many.

bryceosterhaus:
> A string referencing a component in a component registry. This covers the case where the component (multi-select) is instantiated somewhere else and the developer is merely adding her renderer and configuring the component without access to the component initialization.

[...]

PR #2186, opened 2019-07-09
feat(clay-form): create high-level InputWithAutocomplete component

bryceosterhaus:
Fix for https://github.com/liferay/clay/issues/2141

### DIFF
--- a/packages/clay-form/src/InputWithAutocomplete.tsx
+++ b/packages/clay-form/src/InputWithAutocomplete.tsx
@@ -1,0 +1,129 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayAutocomplete from '@clayui/autocomplete';
+import ClayDropDown from '@clayui/drop-down';
+import React, {useEffect, useRef, useState} from 'react';
+import {useKeyHandlerForList} from '@clayui/shared';
+
+interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
+	/**
+	 * Flag to indicate is loading indicator should be shown.
+	 */
+	loading?: boolean;
+
+	/**
+	 * Message displayed in dropdown when laoding is indicated.
+	 */
+	loadingMessage?: string;
+
+	/**
+	 * Items that will show up in the dropdown list
+	 */
+	items: Array<string> | Array<Object>;
+
+	/**
+	 * Function that returns the value that should be showed in the list.
+	 */
+	itemSelector?: (val: string | Object) => string;
+
+	/**
+	 * Callback for when input value changes
+	 */
+	onItemSelect: (val: string | Object) => string | Object;
+
+	/**
+	 * Callback for when input value changes
+	 */
+	onValueChange: (val: string | Object) => void;
+
+	/**
+	 * Flag to indicate if loading message should show in drop down
+	 */
+	showLoadingMessage?: boolean;
+}
+
+export const ClayInputWithAutocomplete: React.FunctionComponent<IProps> = ({
+	items = [],
+	itemSelector = val => val as string,
+	loading,
+	loadingMessage = 'Loading...',
+	onItemSelect,
+	onValueChange,
+	showLoadingMessage = true,
+	value = '',
+	...otherProps
+}) => {
+	const activeListItemRef = useRef<HTMLLIElement>(null);
+	const inputRef = useRef<HTMLInputElement>(null);
+	const [activeIndex, setActiveIndex] = useState(-1);
+	const [active, setActive] = useState(!!value);
+
+	const keyHandler = useKeyHandlerForList({
+		activeListItemRef,
+		index: activeIndex,
+		inputRef,
+		onIndexChange: setActiveIndex,
+		onIndexSelect: (index: number) => onItemSelect(items[index]),
+		totalItems: items.length,
+	});
+
+	useEffect(() => {
+		setActive(!!value);
+	}, [value]);
+
+	useEffect(() => {
+		setActiveIndex(-1);
+	}, [active]);
+
+	return (
+		<ClayAutocomplete>
+			<ClayAutocomplete.Input
+				{...otherProps}
+				onChange={e => onValueChange(e.target.value)}
+				onKeyDown={keyHandler}
+				ref={inputRef}
+				value={value}
+			/>
+
+			{((loading && showLoadingMessage) || !!items.length) && (
+				<ClayAutocomplete.DropDown
+					active={active}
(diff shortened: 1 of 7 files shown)


packages/clay-form/src/InputWithAutocomplete.tsx
```diff
@@ -0,0 +1,102 @@
...
+				onChange={e => onValueChange(e.target.value)}
+				onKeyDown={keyHandler}
+				ref={inputRef}
+				value={value}
+			/>
+
+			<ClayAutocomplete.DropDown active={active} onSetActive={setActive}>
+				<ClayDropDown.ItemList>
+					{items.map((item, i) => (
+						<ClayAutocomplete.Item
```
matuzalemsteles: I was going to say to make it extensible but I remembered that we have a speech at #2152, so I think we can add that later 🙂.

packages/clay-form/src/InputWithAutocomplete.tsx
```diff
@@ -0,0 +1,102 @@
...
+									? activeListItemRef
+									: undefined
+							}
+							value={itemSelector(item)}
+						/>
+					))}
+				</ClayDropDown.ItemList>
+			</ClayAutocomplete.DropDown>
+
+			{loading && <ClayAutocomplete.LoadingIndicator />}
```
matuzalemsteles: Another thing that has to enter the high levels is when loading is enabled, we should also show in the DropDown the message of `Loading...` as in the example of stories of autocomplete.

This is the additional document to handle the use cases of loading. https://docs.google.com/document/d/1ZkpPajJBLioHBd2sEwE7Sb0FPdYr9ypwdL2ebHed1aQ/edit?ts=5c659b11

bryceosterhaus: Is there any way to centralize these docs or make them more findable? I have no idea how to find them currently, even when I search google drive nothing shows up.

Could we link them to tickets?

matuzalemsteles: Yes that's a bit chaotic, sometimes I do not find the documents either... I just had access because of the LEXI tickets they added. I think it would be best to have such documents in github 🙂.

bryceosterhaus:
@matuzalemsteles updated

packages/clay-form/src/InputWithAutocomplete.tsx
```diff
@@ -0,0 +1,113 @@
...
+								ref={
+									activeIndex === i
+										? activeListItemRef
+										: undefined
+								}
+								value={itemSelector(item)}
+							/>
+						))}
+
+					{loading && (
```
matuzalemsteles: I think maybe it would be better to add a new API to activate the message to cover the case that it only appears when the request exceeds 500ms.

I think we can add two checks here with this API, the default value of this new one is `null`, if `loading` is true it shows the message and if the other API is `true` or` false` just show the message accordingly new API.

bryceosterhaus: I went with the strict boolean since for this component itself I wanted to keep it straightforward for a simple use case. For use with a datasource, the consuming component would be the one to implement a 500ms delay. This is also the same reason the items prop is an array and not necessarily a datasource.

matuzalemsteles: Yes I agree with you I would like to keep our high-levels simple also so as not to become something complex and look like this https://github.com/liferay/clay/blob/2.x/packages/clay-autocomplete/src/ClayAutocomplete.js#L454-L795 🙂, but I think in this case, if we implement this loading here, we have to add the possibilities that people can do it, even if the data comes from outside for example.

bryceosterhaus: > if we implement this loading here, we have to add the possibilities that people can do it

Totally agree, but I think if we add the debounce and delay as an internal API for this component, its now assuming the consumer is using async data with a delay or its requiring them to pass a config to disable the delay. By not including it, its not strictly up to the consumer how they want to enable and use the loading flag, they can leverage it any way they would like since it is just a flag. Component's I've consumed in the past with internal delays or timers seem to always be tricky or require specific configuration to get it working how I'd want. From my perspective, I think its best to give users API that doesn't rely internal timers or delays, unless its an explicit async component.

We could think about creating a `AutoCompleteWithDataSource` type high-level component that would strictly be async.

matuzalemsteles: Oh no, I think you got it wrong, I do not want us to add some timer or delay, I understand your concern.

I still want us to keep the API `loading` and add one more, just to the loading of` loadingMessage` prop, so the dev will be able to decide if it will add the delay to the messages and activate the loading to the loadingIndicator as soon as it happens a new requisition.

bryceosterhaus: hmm, not sure I am tracking with you. Could you write up a code example API you are expecting? The `Loading...` message only appears if `loading={true}` and the input is focused and the dropdown is active.

matuzalemsteles: Np, this is an example of usage:

```jsx
<ClayInputWithAutocomplete
    loading={true}
    loadingMessage="Loading..."
    showLoadingMessage={true}
/>
```

in impl:

```jsx
export const ClayInputWithAutocomplete: React.FunctionComponent<IProps> = ({
    loading,
    loadingMessage = 'Loading...',
    showLoadingMessage = null, // is boolean or null
}) => {
    return (
        ...
            {((loading && !showLoadingMessage) || showLoadingMessage !== null) && (
                <ClayAutocomplete.Item value={loadingMessage} />
            )}
        ...
        ...
        {loading && <ClayAutocomplete.LoadingIndicator />}
    );
};
```

In the end I can just set `loading` to true and it will show the two feedback flags or I can control the two separately, set `loading` to `true` and `showLoadingMessage` to false and if pass the 500ms without resolving the request I set `showLoadingMessage` to true. Is that clearer?

Another question for the use of `<ClayAutocomplete.Item value={loadingMessage} />` the markup of the message is a bit different, you can check in v2 version. https://github.com/liferay/clay/blob/2.x/packages/clay-autocomplete/src/ClayAutocomplete.soy#L306-L311

bryceosterhaus: Gotcha, so overall you are just expecting an API to toggle the display of `loadingMessage`?

bryceosterhaus:
@matuzalemsteles updated. I also slightly changed autocomplete item so that we don't always render the div container. I opened up the API to allow for not using "match".

packages/clay-form/src/InputWithAutocomplete.tsx
matuzalemsteles: Yeah

PR #2189, opened 2019-07-10
feat(@clayui/shared): add experimental useFocusManagement hook

### LINKED ISSUE #1987: Focus accessibility for dropdown in v3.x

matuzalemsteles:
Fixes #1987

This is a draft from a focus manager, possibly a substitute for `useKeyHandlerForList`, which may cover more cases in Modal, DropDown and perhaps DatePicker. I have taken as a basis the initial and experimental implementation of the FocusManager API (https://github.com/facebook/react/pull/15849) that may be acceptable.

In terms of design the use of `createScope` becomes more feasible to dribble element problems in React Portal, I tried to do the same as FocusManager API in having a `FocusScope` but here I only have access to DOM elements and not the fibers that would become easier but I want to hear other ideas.

### DIFF
--- a/packages/clay-autocomplete/src/Item.tsx
+++ b/packages/clay-autocomplete/src/Item.tsx
@@ -9,9 +9,9 @@ import fuzzy from 'fuzzy';
 import React from 'react';
 
 interface IProps extends React.ComponentProps<typeof ClayDropDown.Item> {
-	anchorRef?: React.Ref<HTMLAnchorElement>;
-
 	forwardRef?: React.Ref<HTMLLIElement>;
+
+	innerRef?: React.Ref<HTMLAnchorElement>;
 
 	/**
 	 * Match is the string that will be compared with value.
@@ -28,8 +28,8 @@ interface IProps extends React.ComponentProps<typeof ClayDropDown.Item> {
 const optionsFuzzy = {post: '</strong>', pre: '<strong>'};
 
 const ClayAutocompleteItem: React.FunctionComponent<IProps> = ({
-	anchorRef,
 	forwardRef,
+	innerRef,
 	match = '',
 	value,
 	...otherProps
@@ -39,7 +39,7 @@ const ClayAutocompleteItem: React.FunctionComponent<IProps> = ({
 	return (
 		<ClayDropDown.Item
 			{...otherProps}
-			anchorRef={anchorRef}
+			innerRef={innerRef}
 			ref={forwardRef}
 			tabIndex={-1}
 		>
--- a/packages/clay-autocomplete/src/Item.tsx
+++ b/packages/clay-autocomplete/src/Item.tsx
@@ -9,6 +9,8 @@ import fuzzy from 'fuzzy';
 import React from 'react';
 
 interface IProps extends React.ComponentProps<typeof ClayDropDown.Item> {
+	anchorRef?: React.Ref<HTMLAnchorElement>;
+
 	forwardRef?: React.Ref<HTMLLIElement>;
 
 	/**
@@ -26,6 +28,7 @@ interface IProps extends React.ComponentProps<typeof ClayDropDown.Item> {
 const optionsFuzzy = {post: '</strong>', pre: '<strong>'};
 
 const ClayAutocompleteItem: React.FunctionComponent<IProps> = ({
+	anchorRef,
 	forwardRef,
 	match = '',
 	value,
@@ -34,7 +37,12 @@ const ClayAutocompleteItem: React.FunctionComponent<IProps> = ({
 	const fuzzyMatch = fuzzy.match(match, value, optionsFuzzy);
 
 	return (
-		<ClayDropDown.Item {...otherProps} ref={forwardRef}>
+		<ClayDropDown.Item
+			{...otherProps}
+			anchorRef={anchorRef}
+			ref={forwardRef}
+			tabIndex={-1}
+		>
 			{match && fuzzyMatch ? (
 				<div
 					dangerouslySetInnerHTML={{
--- a/packages/clay-drop-down/src/Item.tsx
+++ b/packages/clay-drop-down/src/Item.tsx
@@ -20,6 +20,8 @@ interface IProps
 	 */
 	disabled?: boolean;
 
+	anchorRef?: React.Ref<any>;
+
 	forwardRef?: React.Ref<HTMLLIElement>;
 
 	/**
@@ -45,6 +47,7 @@ interface IProps
 
 const ClayDropDownItem: React.FunctionComponent<IProps> = ({
 	active,
+	anchorRef,
 	children,
 	className,
 	disabled,
@@ -69,6 +72,7 @@ const ClayDropDownItem: React.FunctionComponent<IProps> = ({
 				})}
 				href={href}
 				onClick={onClick}
+				ref={anchorRef}
 			>
 				{symbolLeft && (
 					<span className="dropdown-item-indicator-start">
--- a/packages/clay-drop-down/src/Item.tsx
+++ b/packages/clay-drop-down/src/Item.tsx
@@ -20,14 +20,14 @@ interface IProps
 	 */
 	disabled?: boolean;
 
-	anchorRef?: React.Ref<any>;
-
 	forwardRef?: React.Ref<HTMLLIElement>;
 
 	/**
 	 * Path for item to link to.
 	 */
 	href?: string;
+
+	innerRef?: React.Ref<any>;
 
 	/**
 	 * Path to icon spritemap from clay-css.
@@ -47,12 +47,12 @@ interface IProps
 
 const ClayDropDownItem: React.FunctionComponent<IProps> = ({
 	active,
-	anchorRef,
 	children,
 	className,
 	disabled,
 	forwardRef,
 	href,
+	innerRef,
 	onClick,
 	spritemap,
 	symbolLeft,
@@ -72,7 +72,7 @@ const ClayDropDownItem: React.FunctionComponent<IProps> = ({
 				})}
 				href={href}
 				onClick={onClick}
-				ref={anchorRef}
+				ref={innerRef}
 			>
 				{symbolLeft && (
 					<span className="dropdown-item-indicator-start">
--- a/packages/clay-shared/src/useFocusManagement.ts
+++ b/packages/clay-shared/src/useFocusManagement.ts
@@ -1,0 +1,247 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {useEffect, useRef} from 'react';
+
+type FocusManager = {
+	identifier: string;
+	portal?: boolean;
+	value: HTMLElement;
+};
+
+const TAB_KEY_CODE = 9;
+const FOCUSABLE_ELEMENTS = [
+	'a[href]',
+	'[contenteditable]',
+	'[tabindex]:not([tabindex^="-"])',
+	'area[href]',
+	'button:not([disabled]):not([aria-hidden])',
+	'embed',
+	'iframe',
+	'input:not([disabled]):not([type="hidden"]):not([aria-hidden])',
+	'object',
+	'select:not([disabled]):not([aria-hidden])',
+	'textarea:not([disabled]):not([aria-hidden])',
+];
+
+export function useFocusManagement(loop: boolean = false) {
+	const focusManagerRef = useRef<Array<FocusManager>>([]);
+	const lastPositionRef = useRef<number | null>(null);
+	const nextElementOutsideScopeRef = useRef<HTMLElement | null>(null);
+
+	const createScope = (
+		ref: any,
+		identifier: string,
+		portal: boolean = false
+	) => {
+		const elementIndex = focusManagerRef.current.findIndex(
+			el => (el && el.identifier) === identifier
+		);
+		const schema = {
+			identifier,
+			portal,
+			value: ref,
+		};
+
+		if (ref) {
+			if (elementIndex === -1) {
+				focusManagerRef.current.push(schema);
+			} else {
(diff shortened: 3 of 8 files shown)


packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+				element = focusableElements[item];
+				return true;
+			}
+
+			return false;
+		};
+
+		while (element && !match()) {
+			element = element.parentNode as HTMLElement;
+		}
```
wincent: Are we sure that we want to implement our own focus management when there seems to be a good chance that the API you linked to in the upstream PR will get merged? (I'm guessing there's a good chance, based on the interaction I see on the PR with React team members.)

My concern here is that by doing all this raw DOM interaction we're jettisoning one of the main benefits of React, which is abstraction from the DOM. That abstraction is good because it can shield us from performance, security and compatibility issues.

For example, here, you have a `querySelectorAll()` which is going to match very broadly and deeply (traversing the entire element subtree), and then in this `while` loop you walk up a level and will re-run the `querySelectorAll()` over the entire subtree that you already traversed.

Also, you are passing an array to `querySelectorAll()`, but it takes a string. It works because the array gets silently stringified to "a,b,c", but the fact that you need the `any` cast is a smell. Is passing the array instead of a string by accident or by design? We can't tell by looking at the code.

I imagine if we look in detail through all this we'll find more places where bypassing React's abstraction-from-the-DOM is going to cost us. For example, using local event listeners, which jettisons the event delegation you get for free with React (and which also abstracts away platform-specific differences in event handling).

I'm just wondering whether we'd be better off avoiding this work, and potentially advocating for the upstream PR instead.

matuzalemsteles: > Are we sure that we want to implement our own focus management when there seems to be a good chance that the API you linked to in the upstream PR will get merged? (I'm guessing there's a good chance, based on the interaction I see on the PR with React team members.)

That was one of the reasons I took the implementation of them, we really need a focus manager for our components now, anyway we'll have to implement one for our components if we do not go with it... but the idea is that this implementation is temporary once the upstream PR is merged and released out of experimental use we will remove ours and replace it with the React API (there is no way to fight with the implementation of them since they are consuming the information directly from the fiber... 🙂).

> Also, you are passing an array to querySelectorAll(), but it takes a string. It works because the array gets silently stringified to "a,b,c", but the fact that you need the any cast is a smell. Is passing the array instead of a string by accident or by design? We can't tell by looking at the code.

Yeah, design only 🙂 but we can use string if this has a real problem.

> ...I'm just wondering whether we'd be better off avoiding this work, and potentially advocating for the upstream PR instead.

Just reinforcing I fully support the upstream PR, I do not want to fight with the implementation of them... I totally understand your concern about getting through the DOM almost every time, but if we look at this approach well I can decrease its use a little, since it will call this method only when it reaches the ends of the scope, even if it implements something not to call every time and only calls one, we will have problems in renderings that can kill those elements at any moment, so it would be out of control...

About listeners for sure we are not using the benefit of React, but I would not worry so much, since we are taking care to remove and add only on one element, this does not become as expensive as we imagine.

Just to illustrate the implementation of the Focus Manager 🙂, it may help to understand its design and show that we are still using some React Benefits.

useFocusManagement@2x

`focusManagerRefs` is handled by `createScope` calls that are added to the `refs` of the elements that the dev wants to control, ie it is creating its scope, see the example in the Autocomplete on storybook.

It will look outside `focusManagerRefs` when it reaches the endpoints and if the `loop` is disabled, this becomes necessary since we need to focus the scope and let the browser handle it and add the event only when necessary to control the foci to go back to the scope, this is for cases when the elements are inside a portal.

packages/clay-autocomplete/src/Item.tsx
```diff
@@ -34,7 +37,18 @@ const ClayAutocompleteItem: React.FunctionComponent<IProps> = ({
 	const fuzzyMatchResult = fuzzy.match(match, value, optionsFuzzy);
 
 	return (
-		<ClayDropDown.Item {...otherProps} ref={forwardRef}>
+		<ClayDropDown.Item
+			{...otherProps}
+			component={props => (
+				<a
```
bryceosterhaus: Instead of specifying a new element here, can you just do this?

```jsx
<ClayDropDown.Item
    {...otherProps}
    href="javascript:;"
>
```

Also, I'm not a huge fan of the `javascript:;` pattern. Can just use a button if its for onClick?

packages/clay-drop-down/src/Item.tsx
```diff
@@ -15,6 +15,11 @@ interface IProps
 	 */
 	active?: boolean;
 
+	/**
+	 * Span component to render. It can be a one component that will replace the markup.
+	 */
+	component?: 'span' | React.ComponentType<any>;
```
bryceosterhaus: I think this API might be a bit overkill for clay's use case. We coudl probably get away with using only an anchor or a  button.

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {useEffect, useRef} from 'react';
+
+type FocusManager = {
```
bryceosterhaus: Why use a `type` vs an `interface`? Mostly asking because I have a hard time knowing what to use sometimes.

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+	key: string;
+	portal?: boolean;
+	value: HTMLElement;
+};
+
+const TAB_KEY_CODE = 9;
+const FOCUSABLE_ELEMENTS = [
```
bryceosterhaus: Did you copy this list of elements from somewhere? If so, I think it'd be helpful to document that here in a comment.

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+	'select:not([disabled]):not([aria-hidden])',
+	'textarea:not([disabled]):not([aria-hidden])',
+];
+
+export function useFocusManagement(loop: boolean = false) {
+	const focusManagerRef = useRef<Array<FocusManager>>([]);
+	const lastPositionRef = useRef<number | null>(null);
+	const nextElementOutsideScopeRef = useRef<HTMLElement | null>(null);
+
+	const createScope = (ref: any, key: string, portal: boolean = false) => {
```
bryceosterhaus: One nitpick I might add is changing the name away from `key` since that is typically used when talking about the key prop in react. Maybe we could go with `identifier` or `name`?

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+			focusManagerRef.current.splice(elementIndex, 1);
+		}
+	};
+
+	const getFocusableElementOutsideScope = (
+		element: HTMLElement,
+		backwards: boolean
+	) => {
+		const firstElement = element;
+		element = element.parentNode as HTMLElement;
```
bryceosterhaus: new line above needed

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+			const activeElementFocus = focusableElements.findIndex(
+				el => el === firstElement
+			);
+			const item = backwards
+				? activeElementFocus - 1
+				: activeElementFocus + 1;
+
+			if (focusableElements[item]) {
+				element = focusableElements[item];
+				return true;
```
bryceosterhaus: new line above

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+		const firstPosition = lastPositionRef.current === 0;
+		const endPosition =
+			lastPositionRef.current !== null &&
+			size === lastPositionRef.current;
+
+		// Get the parent elements or focusable children of the
+		// non-portal element to move focus.
+		if (
+			!loop &&
+			(((lastPositionRef.current === null || firstPosition) &&
```
bryceosterhaus: This condition is pretty tough to follow, can we break it up into more readable chunks?

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -0,0 +1,224 @@
...
+		// Hack to check if element is visible
+		if (!value.offsetParent) {
+			moveFocusInScope(backwards);
+		} else {
+			value.focus();
+		}
+	};
+
+	useEffect(
+		() => () => {
```
bryceosterhaus: just a thought, I was a little confused when I first saw this, but then I realized you are wanting this to run when it "unmounts". It might help a bit if we avoid implicit returns for useEffect, just to make it really clear its an unmount function.

packages/clay-autocomplete/src/Item.tsx
matuzalemsteles: Oh yes, the use case for this was that I needed the `ref` of anchor and not of `li`, but I think we can replace for a `Button` really is more sensible here.

packages/clay-drop-down/src/Item.tsx
matuzalemsteles: Hmm, maybe it is, I think to cover the case above, I think we can remove the `ItemElement` wrapper and let those who use the low level component decide to use a `a`, `button` element... via `children `, I think it is more flexible so, when the element is not passed the classNames (`dropdown-item`) can live in `li`. What do you think?

packages/clay-shared/src/useFocusManagement.ts
matuzalemsteles: Oh yeah, this can lead to confusion I'm also not completely sure about this, but I see that it makes sense to use:

- `type` for cases where it is not necessary a complete definition of something or that also do not need to use extends and functionalities that the `interface` allows me.
- `interface` I see more common usage for our `IProps` and when I need to extend other interfaces or when I create signatures for `class`, for example.

But I'm not 100% sure about this, sometimes I'm still in doubt, so I'm not sure when to use it or not in some cases. But what I'm trying to do now, I usually always use `interface`.

packages/clay-shared/src/useFocusManagement.ts
matuzalemsteles: In fact it was a mix, some that I added and others that I picked up from somewhere that I do not remember anymore 😅, this ended up coming from the v2 version I was using for Modal.

packages/clay-shared/src/useFocusManagement.ts
matuzalemsteles: Hm yeah, I think we can switch to `identifier` I like that.

packages/clay-shared/src/useFocusManagement.ts
matuzalemsteles: Hmm yeah, I think I can simplify it more, I'll think about it.

packages/clay-shared/src/useFocusManagement.ts
matuzalemsteles: Oh yeah, this may sound messy, you're right I'll avoid this. I've seen in some cases people creating hooks to avoid these kinds of thing, like `useUnmount` and `useMount`.

packages/clay-drop-down/src/Item.tsx
bryceosterhaus: Yeah that sounds good to me

packages/clay-shared/src/useFocusManagement.ts
wincent: I usually default to `type`, but TSLint [alleges that](https://palantir.github.io/tslint/rules/interface-over-type-literal/):

> Interfaces are generally preferred over type literals because interfaces can be implemented, extended and merged.

typescript-eslint had a similar rule, [prefer-interface](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-interface.md), but that one got deprecated in favor of the apparently less opinionated [consistent-type-definitions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md). That one enables you to always prefer `type`, or always prefer `interface`, and the only guidance it gives you is:

> If you specifically want to use an interface or type literal for stylistic reasons, you can disable this rule.

I think there are two reasonable positions on this one:

- Use `type` for everything, except when you need an `interface`.
  - Benefit: using the simpler concept with terser syntax in most places.
- Use `interface` for everything.
  - Benefit: less to think about, especially if the linter reminds us when we forget.

On the balance, it seems like @matuzalemsteles policy ("usually always use `interface`") is probably the "best", although I suspect it doesn't matter at a cosmological scale. 😂

packages/clay-drop-down/src/Item.tsx
matuzalemsteles: I think it will not work very well 🙂 so that it works well I would have to remove the icons from the `Item` because the icons have to live inside the elements, be it `a`, `button` or whatever. I think I will keep it as is and remove the `component` API I will try to modify cover the above use case by adding another new API to add ref to the anchor or button.

matuzalemsteles:
hey @bryceosterhaus update.

I added the focus manager to `ClayDropDownWithBasicItems` and made some modifications to` useFocusManagement` to avoid looping, zero the last position and other small details. I also arranged the commits.

packages/clay-drop-down/src/Item.tsx
```diff
@@ -69,6 +72,7 @@ const ClayDropDownItem: React.FunctionComponent<IProps> = ({
 				})}
 				href={href}
 				onClick={onClick}
+				ref={anchorRef}
```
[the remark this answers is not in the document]
matuzalemsteles: Oh yeah, I forgot to comment on that, I didn't find a good name. I like `innerRef`.

Issue #2153, opened 2019-07-02
High-level component for DropDown


bryceosterhaus:
Possible API

```jsx
<ClayDropdown
	items={[
		{
			label: 'clickable',
			onClick: () => {},
		},
		{type: 'divider'},
		{
			href: '/some/path',
			label: 'linkable',
		},
	]}
	trigger={<ClayButton>{'Click Me'}</ClayButton>}
/>;
```

PR #2193, opened 2019-07-12
feat(clay-drop-down): create DropDownBasic high-level component

bryceosterhaus:
Unsure about the name of `ClayDropDownBasic`... But here is my initial thought for a simple high-level dropdown. Particularly I think this high-level will help in the implementation of the simple dropdown actions for cards.

fixes https://github.com/liferay/clay/issues/2153


packages/clay-drop-down/src/DropDownBasic.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface IItem {
+	href?: string;
+	label?: string;
+	type?: 'divider';
+	onClick?: () => void;
+}
+
+interface IProps {
+	items: Array<IItem>;
```
matuzalemsteles: We can only add a few comments in the API to move on.

bryceosterhaus:
updated docs and renamed. went ahead and merged

PR #2194, opened 2019-07-12
feat(card): create high-level CardWithUser component

bryceosterhaus:
Actions dropdown will be added later after high-level dropdown is merged, everything else should be good to go for this though.

I wasn't totally sure where to find docs on the cards with user images, but I think I was able to figure it out.


packages/clay-card/src/CardWithUser.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import ClayCard from './Card';
+import ClayForm from '@clayui/form';
+import ClayIcon from '@clayui/icon';
+import ClayLabel from '@clayui/label';
+import ClaySticker, {DisplayType as StickerDisplayType} from '@clayui/sticker';
+import React from 'react';
+
+interface IProps {
+	description: string;
```
matuzalemsteles: I would add some comments here, before going ahead.

bryceosterhaus:
updated docs

PR #2199, opened 2019-07-16
feat(multi-step-nav): create multi-step-nav component

bryceosterhaus:
This is the low-level implementation for now. If this general pattern looks good, I will also build high-level on top of this.


packages/clay-multi-step-nav/stories/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+			active: value === 2,
+			complete: value > 2,
+			onClick: () => setValue(2),
+			subTitle: 'SubThree',
+			title: 'Three',
+		},
+		{
+			active: value === 3,
+			complete: value > 3,
+			onClick: () => setValue(3),
```
matuzalemsteles: Perhaps the use of `onClick` in each item can be removed from here and we pass it on the` map` with the `index` parameter.

packages/clay-multi-step-nav/stories/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+		},
+		{
+			active: value === 2,
+			complete: value > 2,
+			onClick: () => setValue(2),
+			subTitle: 'SubThree',
+			title: 'Three',
+		},
+		{
+			active: value === 3,
```
matuzalemsteles: I have a little bit of concern about defining `active` and `complete` directly on the items, I think it would be best to just set it directly in the `map` interaction.

jbalsas:
Hey @andresfulla, I think I remember seeing this used in the early `A/B testing` mockups. Can you confirm if this would address your use case as well?

interaminense:
Hey @bryceosterhaus, in the Analytics project, we need passing a Body component in every steps. Take a look in this image: 

Screen Shot 2019-07-17 at 14 34 31

Using a low-level component will be great for us because it will be necessary if you look at the image!!

bryceosterhaus:
@interaminense is this already possible with just clay-css or does it require custom styling?

packages/clay-multi-step-nav/stories/index.tsx
bryceosterhaus: Since this is just the low-level API for now, I think it makes more sense to pass it into the items directly. This allows greater flexibility on the low-level API by allowing users to mix with steps with both `onClick` and `href`. Likely in the high-level implementation, we would pass it in with the `map` and `index` params though.

packages/clay-multi-step-nav/stories/index.tsx
bryceosterhaus: Similar to my comment above, I think it makes sense to do it hear at the low-level. Once its implemented into a high-level component, it'll likely be integrated.

packages/clay-multi-step-nav/stories/index.tsx
matuzalemsteles: Hmmm, I'm not so sure about this but I think we can go with that.

interaminense:
With clay markup today, content customization is required so that the cards are aligned in the center as shown in the image. But I believe this is something specific to this screen, and that we must customize on our side.

matuzalemsteles:
hey @bryceosterhaus I think we can hold this component a little until the release of `milestone.2` since it will be available only in `milestone.3` or we can launch now... what do you think?

bryceosterhaus:
@matuzalemsteles we can go ahead and hold it a bit longer. I will work on the high-level version(s) and add to this PR. I'll ping you here when I update

bryceosterhaus:
@matuzalemsteles added high-level component. Let me know if you think it needs more functionality.

I also came across an improvement for storybook by adding a decorator to allow for hooks within the stories itself. Let me know if you see any issues with that.

bryceosterhaus:
also btw, I don't see any reason why we can't include this in milestone.2 if looks good to you. We can just move that ticket over.

packages/clay-multi-step-nav/src/MultiStepNavWithBasicItems.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+interface IProps {
+	activeIndex: number;
+	onIndexChange: (val: number) => void;
+	spritemap?: string;
+	steps: Array<ISteps>;
+}
+
+export const ClayMultiStepNavWithBasicItems: React.FunctionComponent<
```
matuzalemsteles: Probably we have to add this in the high-levels: When there are a certain amount of steps they should live inside the DropDown, as well as the cases of pagination and breadcrumb... Maybe we can add something to the low levels too...

bryceosterhaus: 👍

bryceosterhaus:
@matuzalemsteles updated

PR #2200, opened 2019-07-16
feat(clay-card): create high-level ClayCardWithNavigation component

bryceosterhaus:


packages/clay-card/src/Card.tsx
```diff
@@ -56,7 +57,7 @@ const ClayCard: React.FunctionComponent<IProps> & {
 		user: displayType === 'user',
 	};
 
-	const TagHeaderName = interactive ? 'a' : 'div';
+	const TagHeaderName = interactive ? (onClick ? 'span' : 'a') : 'div';
```
matuzalemsteles: I think it's more semantic to render a `button` when it has an `onClick`.

packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -0,0 +1,79 @@
...
+	children,
+	description,
+	horizontal = false,
+	href,
+	onClick,
+	spritemap,
+	title,
+}) => {
+	return (
+		<ClayCard
```
matuzalemsteles: I think we'll have to add a `tabindex="0"` in the anchor so that css does not remove the color in `horizontal` mode, when it does not have, it's leaving the `inherit` color.

packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -0,0 +1,79 @@
...
+							{description}
+						</ClayCard.Description>
+					</>
+				)}
+
+				{horizontal && (
+					<>
+						<span className="autofit-col">
+							<ClaySticker inline>
+								<ClayIcon spritemap={spritemap} symbol="page" />
```
matuzalemsteles: I think we'd better leave the `symbol` configurable.

packages/clay-card/src/Card.tsx
bryceosterhaus: oh true, not sure why i didn't think of that.

packages/clay-card/src/CardWithNavigation.tsx
bryceosterhaus: Should it always be there or only when href is present?

bryceosterhaus: And by href present, I mean when its an anchor element.

matuzalemsteles: I think that only when have an anchor can we add a `tabindex`.

bryceosterhaus:
@matuzalemsteles updated

I took a look at clay-css site and looks like interactive cards have `tabindex={0}` regardless of the element.

Also, I tried to add the button for the container element for the card but it messed up the width and size of the card. @pat270 what do you think about this? If we use an `onClick` instead of an `href` should the card container be a `button`?

matuzalemsteles:
> I took a look at clay-css site and looks like interactive cards have tabindex={0} regardless of the element.

Maybe it's okay to have `tabindex` on all items, but I think it makes sense for interactive cards.

bryceosterhaus:
@matuzalemsteles updated

I talked to Patrick and decided its best for now to use anchor tags with "javascript:;" for now until some updates can be made for button tags. If we just use a div or span for "onClick" we don't get the accessibility of being able to hit enter or space to activate the "onClick"

packages/clay-card/src/Card.tsx
```diff
@@ -82,7 +85,9 @@ const ClayCard: React.FunctionComponent<IProps> & {
 				'template-card-horizontal': horizontal && interactive,
 				'user-card': isCardType.user,
 			})}
+			href={interactive ? href : undefined}
```
matuzalemsteles: I'm not sure about this, why not just leave `href`?

packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayCard from './Card';
+import ClayIcon from '@clayui/icon';
+import ClaySticker from '@clayui/sticker';
+import React from 'react';
+
+interface IProps {
+	description?: React.ReactText;
```
matuzalemsteles: I would add some docs here also before going ahead.

packages/clay-card/src/Card.tsx
bryceosterhaus: Because we default `href = 'javascript:;'` in the destructuring of the props. So we don't want to apply any href if the element is not an anchor. This helps align with the line `const TagHeaderName = interactive ? 'a' : 'div';` as well

bryceosterhaus:
@matuzalemsteles updated docs

packages/clay-card/src/Card.tsx
matuzalemsteles: Hmm, ok makes sense.

Issue #2197, opened 2019-07-15
Create CardWithFolder high-level component


bryceosterhaus:
```jsx
<CardWithFolder
	actions // comes from dropdown
	onSelectChange // callback
	selected // boolean
	title // string
/>
```

Issue #2196, opened 2019-07-15
Create CardWithFile high-level component


bryceosterhaus:
```jsx
<CardWithFile
	actions // comes from dropdown
	description // string
	label={{
		displayType // string
		value // string
	}}
	onSelectChange // callback
	selected // boolean
	title // string
/>
```

Issue #2195, opened 2019-07-15
Create CardWithImage high-level component


bryceosterhaus:
```jsx
<CardWithImage
	actions // comes from dropdown
	description // string
	imageSrc // string path to image
	label={{
		displayType // string
		value // string
	}}
	onSelectChange // callback
	selected // boolean
	title // string
/>
```

PR #2208, opened 2019-07-23
feat(card): create CardWithFile high-level component

bryceosterhaus:
Fixes https://github.com/liferay/clay/issues/2195
Fixes https://github.com/liferay/clay/issues/2196 
Fixes https://github.com/liferay/clay/issues/2197

### DIFF
--- a/packages/clay-card/src/CardWithFile.tsx
+++ b/packages/clay-card/src/CardWithFile.tsx
@@ -1,0 +1,202 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import ClayCard from './Card';
+import ClayForm from '@clayui/form';
+import ClayIcon from '@clayui/icon';
+import ClayLabel from '@clayui/label';
+import ClaySticker from '@clayui/sticker';
+import React, {ImgHTMLAttributes} from 'react';
+import {ClayDropDownWithBasicItems} from '@clayui/drop-down';
+
+interface IProps {
+	/**
+	 * List of actions in the dropdown menu
+	 */
+	actions?: React.ComponentProps<typeof ClayDropDownWithBasicItems>['items'];
+
+	/**
+	 * Description of the file
+	 */
+	description?: React.ReactText;
+
+	/**
+	 * Flag to indicate if `aspect-ratio-item-flush` class should be
+	 * applied to the image.
+	 */
+	flushHorizontal?: boolean;
+
+	/**
+	 * Flag to indicate if `aspect-ratio-item-vertical-flush` class should be
+	 * applied to the image.
+	 */
+	flushVertical?: boolean;
+
+	/**
+	 * Path or URL to file
+	 */
+	href?: string;
+
+	/**
+	 * Object of props for `<img />` or string path to image
+	 */
+	imgProps?: React.ImgHTMLAttributes<HTMLImageElement> | string;
+
+	/**
+	 * List of labels that are applied to the file
+	 */
+	labels?: Array<
+		React.ComponentProps<typeof ClayLabel> & {value: React.ReactText}
+	>;
+
+	/**
+	 * Callback for when item is selected
+	 */
+	onSelectChange?: (val: boolean) => void;
+
+	/**
+	 * Flag to indicate if card is selected
+	 */
+	selected?: boolean;
+
+	/**
+	 * Path to clay icon spritemap
+	 */
+	spritemap: string;
+
+	/**
+	 * Values used in displaying bottom-left icon
+	 */
+	stickerProps?: {
+		content: React.ReactNode;
+		displayType: React.ComponentProps<typeof ClaySticker>['displayType'];
+	};
+
+	/**
+	 * Name of icon
+	 */
+	symbol?: string;
+
+	/**
+	 * Name of the file
+	 */
+	title: string;
+}
+
+export const ClayCardWithFile: React.FunctionComponent<IProps> = ({
+	actions,
+	description,
+	flushHorizontal,
+	flushVertical,
+	href,
+	imgProps,
+	labels,
+	onSelectChange,
+	selected = false,
+	spritemap,
+	stickerProps,
+	symbol = 'documents-and-media',
+	title,
+}) => {
+	const headerContent = (
+		<ClayCard.AspectRatio className="card-item-first">
+			{!imgProps && (
+				<div className="aspect-ratio-item aspect-ratio-item-center-middle aspect-ratio-item-fluid card-type-asset-icon">
+					<ClayIcon spritemap={spritemap} symbol={symbol} />
+				</div>
+			)}
+
+			{imgProps && (
+				<img
+					className={classNames(
+						'aspect-ratio-item aspect-ratio-item-center-middle aspect-ratio-item-fluid',
+						typeof imgProps !== 'string' && imgProps.className,
+						{
+							['aspect-ratio-item-flush']: flushHorizontal,
+							['aspect-ratio-item-vertical-flush']: flushVertical,
+						}
+					)}
+					{...(typeof imgProps === 'string'
+						? {src: imgProps}
+						: imgProps)}
+				/>
+			)}
+
+			<ClaySticker
+				displayType={
+					stickerProps ? stickerProps.displayType : 'primary'
+				}
+				position="bottom-left"
+			>
+				{stickerProps ? (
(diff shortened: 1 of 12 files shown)


packages/clay-card/src/CardWithFile.tsx
```diff
@@ -0,0 +1,152 @@
...
+	labels,
+	name,
+	onSelectChange,
+	selected = false,
+	spritemap,
+}) => {
+	const headerContent = (
+		<ClayCard.AspectRatio className="card-item-first">
+			<div className="aspect-ratio-item aspect-ratio-item-center-middle aspect-ratio-item-fluid card-type-asset-icon">
+				<ClayIcon spritemap={spritemap} symbol="documents-and-media" />
```
matuzalemsteles: We will have to make `symbol` configurable here, the user or the product that using can modify it.

packages/clay-card/src/CardWithFolder.tsx
```diff
@@ -0,0 +1,99 @@
...
+	name,
+	onSelectChange,
+	selected = false,
+	spritemap,
+}) => {
+	const content = (
+		<ClayCard.Body>
+			<div className="autofit-col">
+				<ClaySticker inline>
+					<ClayIcon spritemap={spritemap} symbol="folder" />
```
matuzalemsteles: I would say the same here

packages/clay-card/src/Description.tsx
```diff
@@ -38,22 +43,24 @@ const ClayCardDescription: React.FunctionComponent<ICardDescriptionProps> = ({
 	truncate = true,
 	...otherProps
 }: ICardDescriptionProps) => {
-	const {interactive} = React.useContext(Context);
-
-	const interactiveTag = interactive ? 'span' : 'div';
-
-	const TagName = href ? 'a' : interactiveTag;
+	const OuterTag = CARD_TYPE_ELEMENTS[displayType] as ('span' | 'div' | 'h3');
```
matuzalemsteles: Just curious, why isn't TypeScript inferring the output here?

bryceosterhaus: It will infer the type of the value, `string`, but it wouldn't pick up the explicit string values.

packages/clay-card/src/CardWithFolder.tsx
bryceosterhaus: I'd be hesitant to make it configurable since its explicitly the "CardWithFolder" component. If we remove that symbol, it starts to be much more vague.

packages/clay-card/src/CardWithFile.tsx
bryceosterhaus: Hmm, what would you imagine users changing it to? I was looking in the docs and it seemed explicit that this was the symbol someone would see.

packages/clay-card/src/Description.tsx
matuzalemsteles: Oh yeah, this is true.

packages/clay-card/src/CardWithFile.tsx
matuzalemsteles: The icons may be different, so we as a library should let the product team set this up easily without holding them back so much that they have to issue an issue here just to change the icon I want to leave this part of the high-level not too rigid in order not to reach that level, to certain things that we can make configurable.

But in the documentation has the description of making this configurable, are you looking at this document https://docs.google.com/document/d/1st2moaRVphrsFrDJUz-ic26aH-V46YkQam75X6BHPY0/edit#heading=h.eevh9ngq3hi8?

packages/clay-card/src/CardWithFolder.tsx
matuzalemsteles: Yeah, you're right about that, you might want to rename it `CardWithHorizontal` to follow what Lexicon said: https://docs.google.com/document/d/1st2moaRVphrsFrDJUz-ic26aH-V46YkQam75X6BHPY0/edit#heading=h.x89wughhxazs

What do you think?

packages/clay-card/src/CardWithFile.tsx
bryceosterhaus: ah, I wasn't able to find that doc. that should help. thanks

bryceosterhaus:
@matuzalemsteles made some changes. I also implemented the ability to use images for files, since file and image cards are essentially the same. I also went through that doc and updated the cards to be slightly more flexible for config. Let me know what you think

packages/clay-card/src/CardWithFile.tsx
```diff
@@ -0,0 +1,202 @@
...
+	const headerContent = (
+		<ClayCard.AspectRatio className="card-item-first">
+			{!imgProps && (
+				<div className="aspect-ratio-item aspect-ratio-item-center-middle aspect-ratio-item-fluid card-type-asset-icon">
+					<ClayIcon spritemap={spritemap} symbol={symbol} />
+				</div>
+			)}
+
+			{imgProps && (
+				<img
```
matuzalemsteles: I'm not sure `CardWithFile` will accept images in case it should only accept icons, but I understand that in the end it looks like `CardWithImage` maybe we can leave a more generic name to merge the two into one.

bryceosterhaus: hmm, yeah probably smart to make the name more generic.

`ClayCardWithBasicInfo`
or
`ClayCardWithEntity` (kinda weird cause we already have *WithUser)

We could try to merge this with `ClayCardWithUser` and make it `ClayCardWithEntity`. Not sure if we want to keep the distinction though.

Any other names come to mind?

matuzalemsteles: Hmm I don't know if I like `ClayCardWithEntity` does not seem so obvious to me its use, maybe going with `ClayCardWithInfo` it is not quite basic so in fact we can remove the word `Basic`.

bryceosterhaus:
@matuzalemsteles i went with `CardWithInfo`. Simple enough and flexible enough for future improvements

packages/clay-card/src/Description.tsx
wincent: @bryceosterhaus, @matuzalemsteles.

There are two ways to force it to infer a better type here.

Best way is to use [`as const`](https://github.com/Microsoft/TypeScript/pull/29510) like this:

```javascript
const CARD_TYPE_ELEMENTS = {
	subtitle: 'span',
	text: 'div',
	title: 'h3',
} as const;
```

This stops TypeScript from widening the type to `string` and you can drop the cast.

The other way that I can think of is to explicitly annotate the type, but that is very repetitive and I wouldn't suggest it:

```javascript
type T = {
	subtitle: 'span',
	text: 'div',
	title: 'h3',
};

const CARD_TYPE_ELEMENTS: T = {
	subtitle: 'span',
	text: 'div',
	title: 'h3',
};
```

[Playground link](https://www.typescriptlang.org/play/index.html#code/MYewdgzgLgBAwgQQEoBED6AVAmgBQKJp4AyeAsngHIYDKMAvDAN4BQAkBAK4BGUAllABsApgC4YAcggAHAIZhxAGjZQhADyhjxAE14A3Rcv7DNACwDMBgL4BuZsygBPKUPgyATlpRCIwN7yl84Ci80gIyDhhOLgziKuriMAA+EnyCQgnJktypwuK2zABmHGDAgWAwBWAAjAAUOqHhkc5icO6e3r7+ZcENEVEAlEzMMCMwoJCwAPIcKm4YMgDm9PDI6Nj4hCTkVNQA2vVSYX3OALowMhAwNZKy8kkSOvr34ubi-bajMG5CUBxu5dNZvMFrZLHZHM4YBhlix2NkjKIJNI5AZWHENA89KicoiXhYlDY7ONoCtUJhcARiGRKDRyfgUGJoQxYZweAjNMj5Eo0WoMdosdycaZ8cxCYViqVeOAKmAAEx1EKHRpRFptLw+PwBKVgHpK45CQYsT7EqYzIRzRbLRBk9aUrY06h0vAofaKo5NIQnD6jb6-f4wQHm4GgongEnWtYUzbUnZoOCTCjUDCMmFsVlCpG3bG8zSPbHsiSvAnnS4m-JFEplGXyg7ulWuDzqzpaoJu5XOQ3DUYmgNmi1LBgRp3R7a0+OJjCu3oer1dka+v4AvvB0XMIA)

matuzalemsteles: > Best way is to use as const like this...

This is a good approach. Thanks @wincent

wincent: I sent [a PR](https://github.com/liferay/clay/pull/2241) with the suggested fix.

PR #2222, opened 2019-07-25
chore(form): update snapshots

bryceosterhaus:
Not sure if I missed anything here. I was hesitant on some of the naming patterns for the components themselves. Let me know what you think.

We might also want to consider defining explicit `children` for some of these helpers, but not sold on it yet.

### DIFF
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import ClayForm, {ClayInput} from '@clayui/form';
 import GradientSelector from './GradientSelector';
 import Hue from './Hue';
 import Icon from '@clayui/icon';
@@ -47,12 +48,12 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
 	}, [value]);
 
 	return (
-		<div className="form-group">
-			<div className="input-group">
-				<div className="input-group-item">
-					<input
-						className="form-control input-group-inset input-group-inset-before"
+		<ClayForm.Group>
+			<ClayInput.Group>
+				<ClayInput.GroupItem>
+					<ClayInput
 						data-testid={`${name}Input`}
+						insetBefore
 						onChange={event => {
 							const newVal = Number(event.target.value);
 
@@ -64,12 +65,12 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
 						type="text"
 						value={inputValue}
 					/>
-					<label className="input-group-inset-item input-group-inset-item-before">
+					<ClayInput.GroupInsetItem before tag="label">
 						{name.toUpperCase()}
-					</label>
-				</div>
-			</div>
-		</div>
+					</ClayInput.GroupInsetItem>
+				</ClayInput.GroupItem>
+			</ClayInput.Group>
+		</ClayForm.Group>
 	);
 };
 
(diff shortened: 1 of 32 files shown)


packages/clay-form/src/Form.tsx
```diff
@@ -0,0 +1,119 @@
...
+import Checkbox from './Checkbox';
+import Input from './Input';
+import Radio from './Radio';
+import RadioGroup from './RadioGroup';
+
+const ClayForm: React.FunctionComponent<
+	React.HTMLAttributes<HTMLDivElement>
+> & {
+	Checkbox: typeof Checkbox;
+	FeedbackGroup: typeof FeedbackGroup;
```
matuzalemsteles: Normally I try to follow the order of declaration than read logic, I prefer to see the declared methods first and then consume, but it's just my opinion.

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -64,12 +65,12 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
 						type="text"
 						value={inputValue}
 					/>
-					<label className="input-group-inset-item input-group-inset-item-before">
+					<ClayForm.Input.GroupInsetItem before tag="label">
```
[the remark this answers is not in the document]
bryceosterhaus: I went with tag because it wasn't necessarily a `div` or `label`. We also use it as a `span` for dropdown search. Since there was multiple elements possible, I figured it was best to put as a prop.

packages/clay-form/src/Form.tsx
bryceosterhaus: Hmm, not sure what you mean by this. Can you share a code example?

bryceosterhaus:
@matuzalemsteles updated with your suggestion of renaming the exports to avoid multiple namespaces

packages/clay-color-picker/src/Custom.tsx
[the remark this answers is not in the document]
matuzalemsteles: Hmm, yeah it makes sense. Ok, I think we can go with that for now, it may slightly diminish this impression I have. Now that it will shorten the namespaces.

packages/clay-form/src/Form.tsx
matuzalemsteles: Oh sorry for that, What I mean is: I almost always try to prioritize the organization of methods, variables... in order of declaration ie I declare first before using. Example https://github.com/liferay/clay/blob/master/packages/clay-data-provider/src/useResource.ts

I think this eslint rule should explain better than me https://eslint.org/docs/rules/no-use-before-define 😅

wincent: I [asked a related question on Twitter](https://twitter.com/wincent/status/1153033906922360835) and was surprised by the variation in the responses (not so much that there were people with different opinions, but rather that there were so many people who thought their opinions were the only ones that were sane — I guess I should have expected that 😂).

matuzalemsteles: Ohh my god! I think we can come to a consensus here haha

bryceosterhaus:
@matuzalemsteles updated and sorted component names to be defined before its use.

bryceosterhaus:
@matuzalemsteles any idea why tests are failing here?

matuzalemsteles:
@bryceosterhaus I thought it was the instability in Travis that they was having yesterday, but it seems not, I will check it now.

bryceosterhaus:
@matuzalemsteles are these netlify fails an issue?

matuzalemsteles:
hey @bryceosterhaus, it's probably not a bug originated from your PR but it's weird that it's just breaking here. Maybe it's the components that have been modified, you can try lifting the site with these changes locally and see what happens?

bryceosterhaus:
@matuzalemsteles turns out I had to also update the components in gatsby to properly build. Do you see any issue with this?

matuzalemsteles:
> @matuzalemsteles turns out I had to also update the components in gatsby to properly build. Do you see any issue with this?

Oh yes, really the log in Netlify didn't seem so clear. I see no problem, you can move on.

PR #2223, opened 2019-07-25
refactor(@clayui/drop-down): remove wrapper from span on trigger and keep ref

matuzalemsteles:
hey @bryceosterhaus I'm to adding this to fix the PaginationBar problem that was running out of style and we just got rid of `span` and also I'm just rearranged the files to be consistent with the rest of the other components.

When that comes in, I'll be starting to release.


packages/clay-drop-down/src/DropDown.tsx
```diff
@@ -75,23 +75,29 @@ const ClayDropDown: React.FunctionComponent<IProps> & {
...
+			{React.cloneElement(trigger, {
+				className: classNames(
+					'dropdown-toggle',
+					trigger.props.className
+				),
+				onClick: () => onActiveChange(!active),
+				ref: (node: HTMLButtonElement) => {
+					triggerElementRef.current = node;
+					// Call the original ref, if any
+					const {ref} = trigger as any;
```
bryceosterhaus: Is this line necassary, or can we just do this below? Main purpose is to get rid of the `any`

```js
if (typeof trigger.ref === 'function')
```

matuzalemsteles: In the end it will give the same error, because there is no `trigger.ref`.

bryceosterhaus: hmm I see, so if there is no trigger ref, wont the destructure of it error?

matuzalemsteles: I think there will always come an object in the trigger that is the fiber elements and when you don't have a ref will probably declare the ref with undefined.

bryceosterhaus: if trigger is undefined, neither will work. But if trigger is an object and ref is undefined, both should work.

this should be the same logically
```js
if (typeof trigger.ref === 'function')
// vs 
const {ref} = trigger;
if (typeof ref === 'function')
```

Unless I am missing something here..

matuzalemsteles: Yes both will work you are correct but still we will need `any`. That is what I mean.

so:

```js
if (typeof (trigger as any).ref === 'function')
// ----- 
const {ref} = trigger as any;
if (typeof ref === 'function')
```

bryceosterhaus: oh gotcha. didnt know you were referring to the type. yeah for that case I guess its fine. Maybe include an inline comment explaining why we use `any` though. Typically its probably a good practice to add a comment anytime we use `any`

matuzalemsteles: Oh yeah, good ideia.

packages/clay-drop-down/src/DropDown.tsx
```diff
@@ -91,7 +91,11 @@ const ClayDropDown: React.FunctionComponent<IProps> & {
 				onClick: () => onActiveChange(!active),
 				ref: (node: HTMLButtonElement) => {
 					triggerElementRef.current = node;
-					// Call the original ref, if any
+					// Call the original ref, if any.
+					// Type is any because the React.ReactElement
+					// interface does not necessarily state that trigger
+					// will always be a fiber under the hood, so we have
+					// to declare any to get away from it.
```
wincent: I don't think "fiber" has anything to do with this, right? The problem is that `React.ReactElement` ([defined here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bb36642c70f366af454e1bf67471fbfdf0b9bdc8/types/react/index.d.ts#L99-L103)) only contains `type`, `props` and `key` properties.

I sent [this PR](https://github.com/liferay/clay/pull/2243) as a possible fix. ¯\\_(ツ)_/¯

matuzalemsteles: > I don't think "fiber" has anything to do with this, right? The problem is that React.ReactElement (defined here) only contains type, props and key properties.

Yeah, what I mean is, they don't declare the properties that the fibers have. Maybe not to encourage bad practices.

wincent: What do you mean by "fiber"? In the context of React, I take it mean:

> Fiber is reimplementation of the stack, specialized for React components. You can think of a single fiber as a virtual stack frame.

(that description taken from [here](https://github.com/acdlite/react-fiber-architecture#what-is-a-fiber)).

matuzalemsteles: Oh yeah, maybe I'm not being so clear about referencing fibers, but the elements have access to "fiber" directly from React, so it's the tree information, probably I could browse the tree consuming this fiber information (accessing child, sibling...) to find the elements you want for example, what React uses when some new update (it is likely) happens and also information such as effect time duration... etc... We could even use this information to improve the `useFocusManagement` system to not have problems with elements within a Portal...

This is more or less what I have when I get any children, by nature of JavaScript I could access any information but it is a bad practice for most cases and an unstable and unpublished React API... regarding the types of TypeScript this is not declared...
Screen Shot 2019-08-05 at 13 25 31

(Look inside the `_owner` property)

PR #2225, opened 2019-07-25
feat(@clayui/management-toolbar): add high-level ClayManagementToolbar component

matuzalemsteles:
I ended up following the same idea we did in v2 and I left the same API for most, yet I'm worried about the bunch of props we have to cover all the use cases it supports, that's one of my concerns when we have high-level components that are very large and with many use cases like this.

Leave your thoughts on this component if we will follow up with any different ideas for it.

There is still some testing and documentation to add to this component, so once we are fine with the API I will add and also has some improvements to make in this component, such as using `sub`, create `DropDownWithAdvancedItems` component.

- [x] Tests
- [x] Docs


packages/clay-management-toolbar/src/ResultsBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+	spritemap,
+	totalItems,
+}) => (
+	<nav className="tbar tbar-inline-xs-down subnav-tbar subnav-tbar-primary">
+		<div className="container-fluid container-fluid-max-xl">
+			<ul className="tbar-nav tbar-nav-wrap">
+				<ResultsBarItem expand={!(filterLabels.length > 0)}>
+					<span className="component-text text-truncate-inline">
+						<span className="text-truncate">
+							{`${totalItems} results for "`}
```
diegonvs: What do you think about wrap this text with a `sub`?

packages/clay-management-toolbar/src/SearchForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const SearchForm: React.FunctionComponent<IProps> = ({
+	children,
+	method = 'get',
+	onlySearch = false,
+	...otherProps
+}) => {
+	const content = (
+		<form {...otherProps} method={method} role="search">
```
diegonvs: Would be a good idea ship this code to `ClayForm`?

packages/clay-management-toolbar/src/__tests__/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import ClayManagementToolbar from '..';
+import React from 'react';
+import {cleanup, render} from '@testing-library/react';
+
+describe('ClayManagementToolbar', () => {
+	afterEach(cleanup);
+
+	it('renders', () => {
```
diegonvs: Need we to include more interaction tests? Especially for event callbacks

packages/clay-management-toolbar/src/DefaultState.tsx
```diff
@@ -0,0 +1,258 @@
...
+	onSortingButtonClick = () => {},
+	searchValue = '',
+	selectable = false,
+	showCreationMenu = true,
+	showInfoButton = false,
+	showSearch = false,
+	sortingOrder,
+	sortingURL,
+	spritemap,
+	titles = {
```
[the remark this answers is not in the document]
matuzalemsteles: `titles` has another context here, it is used to reference titles for tooltips and we also have `labels` here to sync with DropDown.

packages/clay-management-toolbar/src/Item.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLLIElement> {
+	expand?: boolean;
```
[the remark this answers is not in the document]
matuzalemsteles: Oh yeah, I just forgot to remove it. Thanks

packages/clay-management-toolbar/src/ItemList.tsx
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLUListElement> {
+	expand?: boolean;
+}
+
+const ItemList: React.FunctionComponent<IProps> = ({children, expand}) => (
```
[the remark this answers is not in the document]
matuzalemsteles: This does not have much need here since `undefined` in the condition is false too so it would be something more aesthetic.

packages/clay-management-toolbar/src/ResultsBar.tsx
matuzalemsteles: Yeah, I haven't added `sub` yet, that's one of the things I said in the PR description.

packages/clay-management-toolbar/src/SearchForm.tsx
matuzalemsteles: Not sure about that, I still don't see a real need to create a component to render just the form.

packages/clay-management-toolbar/src/__tests__/index.tsx
matuzalemsteles: Yeah, there's still a lot to work on testing, I'll add after we are fine with the component approach, I mentioned above in the PR description.

FabioDiegoMastrorilli:
Hello @matuzalemsteles :) in commerce we have a scenario where we need to use a custom handmade dropdown in order to show and apply custom filters.

Do you think it would be possibile (and compliant with clay) passing a custom filter component as a prop?

cc: @marco-leo

matuzalemsteles:
Oh thanks @FabioDiegoMastrorilli for bringing this use case, I was thinking about it yesterday when I was also looking at the App Builder use case. I'll think of something, maybe instead of being an API for rendering DropDown items maybe we can change For a `filterRender` that accepts the DropDown element, I find it easier to be flexible. @Bryceosterhaus thoughts?

bryceosterhaus:
@matuzalemsteles if customization is defined by lexicon, I think we can add the API. If it doesn't fall under lexicon's specification, we should probably push users to utilize the low-level layer

marco-leo:
Hi @bryceosterhaus @matuzalemsteles we do need this extension point for Commerce, will save us lot of future works, I will make sure this is added as Lexicon extension point.

Thanks

matuzalemsteles:
hey @bryceosterhaus I think we can allow devs to customize DropDown content by passing `filterRender` for example, I think if we force devs to use low levels they will have a lot of work to do with management toolbar, just to add more things or modifying, this component may change a lot and be more flexible, so we can focus and not get as stuck as before.

One day I saw the App Builder team discussing how it would be best to create just a part of the management toolbar for their use cases and something that I realized we had spent a lot of time discussing about the state of the art. of how to make the component or structure instead of worrying about other things that may be more important in your product, since this is a base part of portlets for example. So I think we can start to be a little more flexible and adjust, it's probably something Lexicon will be doing now and so we can make people happier...

drakonux:
Hi everyone! We have received a [request](https://issues.liferay.com/browse/LEXI-663) from team-commerce to evaluate an improvement in the management toolbar based on the scenario they currently have. 

So, I'm working on that now. I'll keep you posted.

matuzalemsteles:
Oh thanks @drakonux, so this would probably be on On Hold?! I think we can go ahead with this and then we can apply the changes that @drakonux requests. @bryceosterhaus?

packages/clay-form/src/InputWithAutocomplete.tsx
```diff
@@ -33,7 +35,7 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	/**
 	 * Callback for when input value changes
 	 */
-	onItemSelect: (val: string | Object) => string | Object;
+	onItemSelect?: (val: string | Object) => void;
```
bryceosterhaus: Would we ever actually have a case where a user wouldn't pass an `onItemSelect`?

packages/clay-management-toolbar/package.json
```diff
@@ -0,0 +1,34 @@
...
+	"scripts": {
+		"build": "cross-env NODE_ENV=production babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "cross-env NODE_ENV=production tsc --project ./tsconfig.declarations.json",
+		"prepublishOnly": "npm run build && npm run build:types",
+		"test": "jest --config ../../jest.config.js"
+	},
+	"keywords": ["clay", "react"],
+	"dependencies": {
+		"classnames": "^2.2.6",
+		"@clayui/label": "^3.0.0-milestone.1"
```
bryceosterhaus: Should this be milestone.2 now?

packages/clay-management-toolbar/package.json
```diff
@@ -0,0 +1,34 @@
...
+	},
+	"devDependencies": {
+		"@clayui/css": "^3.0.0-milestone.1"
```
bryceosterhaus: ^

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+import ClayButton from '@clayui/button';
+import ClayForm from '@clayui/form';
+import ClayIcon from '@clayui/icon';
+import ClayLink from '@clayui/link';
+import Item from './Item';
+import ItemList from './ItemList';
+import React from 'react';
+import {ClayDropDownWithBasicItems} from '@clayui/drop-down';
+import {IItem} from '@clayui/drop-down/src/DropDownWithBasicItems';
```
bryceosterhaus: Can we strip out `/src/DropDownWithBasicItems` here?

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,248 @@
...
+	labels = {
+		checkbox: 'Select Items',
+		clearSelection: 'Clear',
+		selectAll: 'Select all',
+	},
+	onCheckboxChange = () => {},
+	onClearSelectionClick = () => {},
+	onInfoButtonClick = () => {},
+	onQuickActionClick = () => {},
+	onSelectAllClick = () => {},
```
bryceosterhaus: All of these default no-op functions seem strange. Do we intend for people to use this component without providing these callbacks? It seems like there would be missing functionality if they don't provide a proper callback.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,248 @@
...
+								? labels.checkbox
+								: undefined
+						}
+						onChange={onCheckboxChange}
+					/>
+				</Item>
+				<Item>
+					<span className="navbar-text">
+						{selectedItems === totalItems ? (
+							'all-selected'
```
bryceosterhaus: We'll need to pass a configurable object of language keys for these sorts of messages. This way we can account for other languages.

bryceosterhaus: This applies to any explicit english message as well.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,248 @@
...
+													className="nav-link nav-link-monospaced"
+													disabled={item.disabled}
+													displayType="unstyled"
+													onClick={onQuickActionClick}
+													title={item.label}
+												>
+													<ClayIcon
+														spritemap={spritemap}
+														symbol={
+															item.symbolLeft as string
```
bryceosterhaus: ^

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,248 @@
...
+											{item.href && !item.disabled ? (
+												<ClayLink
+													className="nav-link nav-link-monospaced"
+													href={item.href}
+													title={item.label}
+												>
+													<ClayIcon
+														spritemap={spritemap}
+														symbol={
+															item.symbolLeft as string
```
bryceosterhaus: How come this doesnt get properly inferred?

packages/clay-management-toolbar/src/DefaultState.tsx
```diff
@@ -0,0 +1,258 @@
...
+	disabled = false,
+	filterItems,
+	labels = {
+		checkbox: 'Select Items',
+	},
+	onCheckboxChange = () => {},
+	onCreationMenuClick = () => {},
+	onInfoButtonClick = () => {},
+	onSearchValueChange = () => {},
+	onSortingButtonClick = () => {},
```
bryceosterhaus: Same as previous comment above. Are these no-ops necessary?

packages/clay-management-toolbar/src/DefaultState.tsx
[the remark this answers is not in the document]
bryceosterhaus: I wonder if it might be best to be explicit like `titleKeys` or `titlesLang`

packages/clay-management-toolbar/src/ResultsBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+							className="component-label tbar-label"
+							displayType="unstyled"
+							spritemap={spritemap}
+						>
+							{value}
+						</ClayLabel>
+					</ResultsBarItem>
+				))}
+				<ResultsBarItem>
+					{clearResultsURL ? (
```
bryceosterhaus: We tend to do this `Link || Button` logic quite a bit. I wonder if we could create some conditional component that could handle the logic for us.

packages/clay-management-toolbar/src/SearchForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+			{children}
+		</form>
+	);
+
+	if (onlySearch) {
+		return content;
+	}
+
+	return (
```
bryceosterhaus: Might make sense to condense to `onlySearch ? content : <div className="container-fluid container-fluid-max-xl">{content}</div>`

packages/clay-management-toolbar/src/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+	showResultsBar?: boolean;
+}
+
+const ClayManagementToolbar: React.FunctionComponent<IProps> = ({
+	active = false,
+	className,
+	clearLabel = 'Clear',
+	clearResultsURL,
+	filterLabels,
+	onClearButtonClick = () => {},
```
bryceosterhaus: no-op needed?

bryceosterhaus:
@matuzalemsteles yeah I think we can keep going ahead with this and add extensibility later whenever it gets baked out a bit more.

packages/clay-form/src/InputWithAutocomplete.tsx
matuzalemsteles: Oh sorry, actually I had done this to try using Autocomplete instead of the search but had trouble adjusting the markup. I will remove this, I do not remember why I had to add this as optional. But I think we can keep the `void` return here still.

packages/clay-management-toolbar/package.json
matuzalemsteles: Yeah, I had started working on it before `milestone.2`. Thanks

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: Yeah, I think we can remove it now since we won't have `DropDownWithBasicItems` anymore.

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: Yes this is really weird, this is a very large component with many use cases and n features so it's possible that you just render a Search for example without needing all the other components just defining it as the management toolbar will be rendered through the API.

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: Yes, I think I'll add it together with the `labels` API.

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: This is due to the API that `item.symbolLeft` is optional but I think I can check for `symbolLeft` before using it here. Maybe here https://github.com/liferay/clay/pull/2225/files/502893529d4160abe9d2faee596ee4c89e270126#diff-ceb43593f9c76576fc542f9f9c9f3c7bR182

packages/clay-management-toolbar/src/DefaultState.tsx
matuzalemsteles: I would say the same as above.

packages/clay-management-toolbar/src/ResultsBar.tsx
matuzalemsteles: I think that makes sense, it is likely that we will use more and many other places. But I think it would be better to add it to `shared` and not expose it.

packages/clay-management-toolbar/src/SearchForm.tsx
matuzalemsteles: Yes, that seems clearer. Thanks!

packages/clay-management-toolbar/src/index.tsx
matuzalemsteles: I would say the same as above.

packages/clay-management-toolbar/src/DefaultState.tsx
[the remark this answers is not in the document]
matuzalemsteles: Hmm, maybe `titlesLang` gets more context here, I think we can go with that.

packages/clay-management-toolbar/src/ActiveState.tsx
bryceosterhaus: If thats the case, we should still likely still keep these functions as undefined, and then wrap its uses with an if check to make sure its defined. Otherwise, if we have `onInfoButtonClick` defaulted to a no-op, we will show that button but have no functionality for it.

matuzalemsteles: Hm makes sense. I will take care of adding them to conditional on their features.

packages/clay-management-toolbar/src/index.tsx
matuzalemsteles: hey @bryceosterhaus i won't add this on conditional to show up `ResultsBar` because we have `showResultsBar` we will be imposing that just setting `showResultsBar` to true, developers know what they want.

packages/clay-shared/src/LinkOrButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayButton from '@clayui/button';
+import ClayLink from '@clayui/link';
+import React from 'react';
+
+export const LinkOrButton: React.FunctionComponent<any> = ({
```
matuzalemsteles: hey @bryceosterhaus i'm creating the component but i haven't used it yet. There is a case that I will always pass the `displayType` to` ClayButton` and sometimes I don't want to pass to `ClayLink`, this was breaking some things. So we have to think of a way to structure this. Or we can really separate the props from each other. Any thoughts?

matuzalemsteles:
@bryceosterhaus update.

packages/clay-drop-down/src/index.tsx
```diff
@@ -5,8 +5,8 @@
  */
 
 import ClayDropDown, {Align} from './DropDown';
-import {ClayDropDownWithBasicItems} from './DropDownWithBasicItems';
+import {ClayDropDownWithBasicItems, IItem} from './DropDownWithBasicItems';
```
bryceosterhaus: Rather than exporting `IItem` its probably best to do something like `React.ComponentProps<typeof ClayDropDownWithBasicItems>['items'][number];`

I know its a bit longer, but I think its a little more clear. Especially because `IItem` comes from `DropDownWithBasicItems`.

What do you think?

packages/clay-form/src/InputWithAutocomplete.tsx
```diff
@@ -49,11 +51,12 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 const ENTER_KEY_CODE = 13;
 
 export const ClayInputWithAutocomplete: React.FunctionComponent<IProps> = ({
+	containerRenderer,
 	items = [],
 	itemSelector = val => val as string,
 	loading,
 	loadingMessage = 'Loading...',
-	onItemSelect,
+	onItemSelect = () => {},
```
bryceosterhaus: Since `onItemSelect` is required, we don't need this no-op.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+interface IProps extends IPropsBase {
+	actionItems?: Array<IActionItems>;
+	clearSelectionURL?: string;
+	onClearSelectionClick?: (
+		e: React.MouseEvent<HTMLButtonElement, MouseEvent>
+	) => void;
+	onQuickActionClick?: (
+		e: React.MouseEvent<HTMLButtonElement, MouseEvent>
+	) => void;
+	onSelectAllClick?: (
```
bryceosterhaus: Rather than re-declare this type over and over, we could cache it to a type. `type ClickCallback = //...`

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+	filterItems,
+	labels = {
+		allSelected: 'All Selected',
+		checkbox: 'Select Items',
+		clearSelection: 'Clear',
+		selectAll: 'Select all',
+		selected: 'Selected',
+		selectedItemsOfTotalItems: '{0} of {1}',
+	},
+	onCheckboxChange = () => {},
```
bryceosterhaus: I'm pretty sure this needs to not be a no-op. Otherwise, if a user doesn't provide this prop, the component will show the checkbox, but won't function.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+	onClearSelectionClick = () => {},
```
bryceosterhaus: Same as above, if we allow for a conditional type and default to no-op, the component will show the button but it'll have no functionality.

packages/clay-management-toolbar/src/DefaultState.tsx
```diff
@@ -0,0 +1,258 @@
...
+							className="nav-link nav-link-monospaced"
+							displayType="unstyled"
+							onClick={() => setSearchMobile(true)}
+						>
+							<ClayIcon spritemap={spritemap} symbol="search" />
+						</ClayButton>
+					</Item>
+				)}
+
+				{showInfoButton && (
```
bryceosterhaus: This will always evaluate to true since we have a default no-op.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+														spritemap={spritemap}
+														symbol={item.symbolLeft}
+													/>
+												</ClayLink>
+											) : (
+												<ClayButton
+													className="nav-link nav-link-monospaced"
+													disabled={item.disabled}
+													displayType="unstyled"
+													onClick={onQuickActionClick}
```
bryceosterhaus: Won't each actionItem have a different onClick?

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+							</ClayButton>
+						)}
+					</Item>
+				)}
+			</ItemList>
+
+			{(actionItems.length > 0 || showInfoButton) && (
+				<>
+					<ItemList>
+						{showInfoButton && (
```
bryceosterhaus: This will always evaluate as true.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+	onCheckboxChange = () => {},
+	onClearSelectionClick = () => {},
+	onInfoButtonClick = () => {},
+	onQuickActionClick = () => {},
+	onSelectAllClick = () => {},
+	selectAllURL,
+	selectedItems = 0,
+	showInfoButton = false,
+	showSearch = false,
+	showSelectAllButton = false,
```
bryceosterhaus: Rather than have a flag like this, can we just infer this value if by checking for a truthy value of `onSelectAllClick || selectAllURL`

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,252 @@
...
+		selectedItemsOfTotalItems: '{0} of {1}',
+	},
+	onCheckboxChange = () => {},
+	onClearSelectionClick = () => {},
+	onInfoButtonClick = () => {},
+	onQuickActionClick = () => {},
+	onSelectAllClick = () => {},
+	selectAllURL,
+	selectedItems = 0,
+	showInfoButton = false,
```
bryceosterhaus: Do we require an onClick for the infoButton or do we ever use it just as a tooltip? If we require an onClick, we probably don't need this flag and we can infer if the `onInfoButtonClick` prop exists.

packages/clay-management-toolbar/src/DefaultState.tsx
```diff
@@ -0,0 +1,258 @@
...
+	creationMenu = false,
+	disabled = false,
+	filterItems,
+	labels = {
+		checkbox: 'Select Items',
+	},
+	onCheckboxChange = () => {},
+	onCreationMenuClick = () => {},
+	onInfoButtonClick = () => {},
+	onSearchValueChange = () => {},
```
bryceosterhaus: We can remove the no-op, and use this prop to determine if `showSearch` should be true. This means we can also remove `showSearch`.

packages/clay-shared/src/LinkOrButton.tsx
bryceosterhaus: Yeah I can see how the `displayType` may not overlap correctly. One option would be to create a sort of map, or we add `linkDisplayType` and `buttonDisplayType`. I'm inclined to use the two individual props, that way we can have some more flexibility to define how the button and link will look.

matuzalemsteles:
> @matuzalemsteles unless I am missing something here, I'm still not sure all of these no-op values for callbacks make sense. We also seem to use both callbacks with the addition of show*Button props. I think we can remove all no-ops, and then use the callbacks for inferring the show*Button values. Since we use TS, we are able to enforce callbacks via required props or conditionally render if the prop exists.
>
> One question I ask myself to help decide no-ops is "is this callback required for functionality." For example, in the case of a checkbox, we wouldn't ever want the end result to show a checkbox with a no-op for onChange. We either want the checkbox to be functional, or we don't show it at all. Same goes for most buttons and onChange type callbacks.

In this regard, no-ops callbacks are necessary for people to control actions, I would recommend you to see how it is used within the Portal...

In the particular case of checkbox `onChange` it is necessary since checkbox action is to select items from `table`, `card` and `list` for example.

About the `show *` props I like that since it is so much more intuitive, it makes sense to use only callbacks to render the elements since they are needed here but it does not seem so intuitive. This component can behave in many ways and has several use cases.

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: Because? I may only have `actionsItems` and not want to render the info button or the other way around.

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: Yes you can, I think I got it from Soy's use cases. There you have only one event that returns the element you clicked. I think we can do the same here:

```
onClick={event => onQuickActionClick(event, item)}
```

packages/clay-management-toolbar/src/DefaultState.tsx
matuzalemsteles: You say about that if we keep checking `onInfoButtonClick` right?

packages/clay-management-toolbar/src/ActiveState.tsx
matuzalemsteles: Oh yeah

packages/clay-drop-down/src/index.tsx
matuzalemsteles: Hmm, I'm not sure if this could help because I need the IItem interface and if I get from `ClayDropDownWithBasicItems` it will come with `Array<IItem>` I just tried to extract `IItem` but couldn't do anything to get the result.

bryceosterhaus:
> In this regard, no-ops callbacks are necessary for people to control actions, I would recommend you to see how it is used within the Portal

I'm still not totally sure what you mean by this and having trouble wrapping my head around this. Could you link me to a specific instance where this is needed? In particular, a place where the no-op is necessary for user interaction.

> in the particular case of checkbox onChange it is necessary

I understand that `onChange` is necessary for checkbox to work, but the question I am raising is why its necessary or helpful to provide `() => {}` to that component. If you add that callback, the checkbox doesn't actually function right?

> About the show * props I like that since it is so much more intuitive

I'd push back on this primarily due to its "tight coupling." If we use both props, those two props are tightly coupled and would never be used separately. You would never just use `showSelectAllButton` without `onSelectAllButton` or vice versa, `onSelectAllButton` without `showSelectAllButton`. Generally I would prefer any tightly coupled props, especially when one of them is a boolean.

matuzalemsteles:
> I'm still not totally sure what you mean by this and having trouble wrapping my head around this. Could you link me to a specific instance where this is needed? In particular, a place where the no-op is necessary for user interaction.

I don't have an exact Portal link or instance, you might want to look at Documents and Media, Forms... all main pages use the Management Toolbar. But especially I don't know if the code would be the best to show, because the use case there is different, they are all taglibs and it's metal, I think all of that makes more sense there. Since you have no callback props and only events.

> I understand that onChange is necessary for checkbox to work, but the question I am raising is why its necessary or helpful to provide () => {} to that component. If you add that callback, the checkbox doesn't actually function right?

All this becomes necessary since I followed the idea of ​​having props `show*` because I am not inferring the callbacks on conditionals to decide if I will render the element, the question is, are we going to remove the `show*` props? If yes we essentially don't need this in callbacks, the hesitation I'm having all about is using `show*` props or not? will this be more beneficial for the developer? Will he have to guess if using a callback is what defines rendering the component? callbacks with `() => {}` are just consequences of all this.

> I'd push back on this primarily due to its "tight coupling." If we use both props, those two props are tightly coupled and would never be used separately. You would never just use showSelectAllButton without onSelectAllButton or vice versa, onSelectAllButton without showSelectAllButton. Generally I would prefer any tightly coupled props, especially when one of them is a boolean.

Yes, I agree with you about this, it can be a redundancy but in the end I am wondering if this is all worth it in order to be the best for the developer.

To end our conversation and not extend so long to finish the deadline... I will remove the `show*` props and consequently callback no-ops and we can test how this will work, I will add in the conditionals all the elements that are needed to render the element so as not to be in charge of just the callback.

packages/clay-shared/src/LinkOrButton.tsx
matuzalemsteles: Hmm yeah, I think we can have two individual props for these cases. Sounds good.

bryceosterhaus:
Sounds good to me. We could further discuss this in an ADR to go more in depth on `*show` props, because if we go that route we will likely want to apply it across the whole codebase to keep consistency(for example, Alerts would likely need `onClose` and `showCloseButton`).

matuzalemsteles:
> Sounds good to me. We could further discuss this in an ADR to go more in depth on *show props, because if we go that route we will likely want to apply it across the whole codebase to keep consistency(for example, Alerts would likely need onClose and showCloseButton).

Yeah, I think in the end what will tell you what is the best is the devs, we could try to collect feedback to make that decision. Because if you follow some of the two we will have some exchange.

bryceosterhaus:
@matuzalemsteles yeah thats a great point. Maybe we could try to put together a few users as a case study and see how they use it and what they need from it. Or maybe even just post a feedback form on Loop and see if we can collect any data from users.

matuzalemsteles:
> @matuzalemsteles yeah thats a great point. Maybe we could try to put together a few users as a case study and see how they use it and what they need from it. Or maybe even just post a feedback form on Loop and see if we can collect any data from users.

Yeah that's a great idea!

matuzalemsteles:
hey @bryceosterhaus update. I removed the `show*` props and added `LinkOrButton`, let me know if you see anything else.

packages/clay-drop-down/src/index.tsx
bryceosterhaus: Sorry, I totally missed this comment. You should be able to get `IItem` from `Array<IItem>` by adding `[number]`. This isn't vitally important, but it might be helpful to have consistency.

packages/clay-management-toolbar/src/ActiveState.tsx
```diff
@@ -0,0 +1,229 @@
...
+				<Item>
+					<span className="navbar-text">
+						{selectedItems === totalItems ? (
+							labels.allSelected
+						) : (
+							<>
+								{sub(labels.selectedItemsOfTotalItems, [
+									selectedItems,
+									totalItems,
+								])}{' '}
```
bryceosterhaus: Is this empty space necessary or should this spacing be applied via css?

matuzalemsteles: It is necessary, not to merge the sub result with `labels.selected`, I think it would be strange to add a space via CSS should not be a problem with css. Maybe we can add backtick and add a space at the end but it's not very explicit.

```javascript
`${sub(labels.selectedItemsOfTotalItems, [selectedItems, totalItems])} `;
```

bryceosterhaus: No worries, the explicit `{' '}` is fine. I just wanted to make sure.

matuzalemsteles: 👍

matuzalemsteles:
@bryceosterhaus I will still add documents and tests for this component, I am working on it.

matuzalemsteles:
I'm just pushing formatted commits, tomorrow I'll be working on tests and docs.

drakonux:
> Hi everyone! We have received a [request](https://issues.liferay.com/browse/LEXI-663) from team-commerce to evaluate an improvement in the management toolbar based on the scenario they currently have.
> 
> So, I'm working on that now. I'll keep you posted.

Hi everyone, just for your information based on the [request](https://issues.liferay.com/browse/LEXI-663) team-commerce did a couple of weeks ago. I've created a new [issue](https://github.com/liferay/clay/issues/2292) requesting two extension points in the Management Toolbar.

matuzalemsteles:
@bryceosterhaus I added the docs and tests, let me know if you see anything.

PR #2233, opened 2019-07-30
refactor: add useFocusManagement to components and create FocusScope

### LINKED ISSUE #2227: Adding the focus management on components
The `useFocusManagement` hook has been added to some high-level components, we need to keep adopting to ensure a better experience, we need to add it to other components that are using DropDown's high-level.

- ClayInputWithAutocomplete
- ClayMultiStepNavWithBasicItems
- MultiSelect
- Breadcrumb
- ClayPagination

It's probably something we also want to see in `ColorPicker` and `DatePicker`, both use Portal and lose the trigger focus for the dropdown.

matuzalemsteles:
Fixes #2227

I ended up creating the `FocusScope` component that deals only with the logic of controlling focus with tab and arrow keys to minimize the logic we may have repeated between some components.

I also made some important improvements to `useFocusManagement` to cover cases that I just caught with our components (I added some comments to the code.).

I tried to force the use of the `ClayDropDownWithBasicItems` component in some cases which I found that it was not necessary to add a low-level component. Maybe we can force more elsewhere.

These changes are also in conjunction with #2224.

### DIFF
--- a/packages/clay-shared/src/FocusScope.tsx
+++ b/packages/clay-shared/src/FocusScope.tsx
@@ -1,0 +1,53 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+import {useFocusManagement} from './useFocusManagement';
+
+interface IProps {
+	arrowKeys?: boolean;
+	focusManager: ReturnType<typeof useFocusManagement>;
+}
+
+const ARROW_DOWN_KEY_CODE = 40;
+const ARROW_UP_KEY_CODE = 38;
+const TAB_KEY_CODE = 9;
+
+export const FocusScope: React.FunctionComponent<IProps> = ({
+	arrowKeys = true,
+	children,
+	focusManager,
+}) => {
+	const onKeyDown = (event: React.KeyboardEvent<any>) => {
+		const {keyCode, shiftKey} = event;
+
+		if (
+			(arrowKeys && keyCode === ARROW_DOWN_KEY_CODE) ||
+			(keyCode === TAB_KEY_CODE && !shiftKey)
+		) {
+			event.preventDefault();
+			focusManager.focusNext();
+		} else if (
+			(arrowKeys && keyCode === ARROW_UP_KEY_CODE) ||
+			(keyCode === TAB_KEY_CODE && shiftKey)
+		) {
+			event.preventDefault();
+			focusManager.focusPrevious();
+		}
+	};
+
+	return React.cloneElement(children as React.FunctionComponentElement<any>, {
+		onKeyDown: (event: React.KeyboardEvent) => {
+			onKeyDown(event);
+
+			// If the element already exists a `onKeyDown` event should invoke it as well.
+			// Any here is because React interfaces do not expose what lies under the hood.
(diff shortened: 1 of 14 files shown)


packages/clay-shared/src/FocusScope.tsx
```diff
@@ -0,0 +1,53 @@
...
+		}
+	};
+
+	return React.cloneElement(children as React.FunctionComponentElement<any>, {
+		onKeyDown: (event: React.KeyboardEvent) => {
+			onKeyDown(event);
+
+			// If the element already exists a `onKeyDown` event should invoke it as well.
+			// Any here is because React interfaces do not expose what lies under the hood.
+			if ((children as any).props.onKeyDown) {
```
bryceosterhaus: This `any` scares me a bit. We need to assert via types that `children` is only a single react component or node. Otherwise, accessing `props` on an array of children will error.

packages/clay-shared/src/FocusScope.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+import {useFocusManagement} from './useFocusManagement';
+
+interface IProps {
+	arrowKeys?: boolean;
```
bryceosterhaus: Docs here would be great.

packages/clay-shared/src/FocusScope.tsx
```diff
@@ -0,0 +1,53 @@
...
+	focusManager: ReturnType<typeof useFocusManagement>;
+}
+
+const ARROW_DOWN_KEY_CODE = 40;
+const ARROW_UP_KEY_CODE = 38;
+const TAB_KEY_CODE = 9;
+
+export const FocusScope: React.FunctionComponent<IProps> = ({
```
bryceosterhaus: Maybe some overall description of the purpose for this component would be helpful here.

packages/clay-shared/src/FocusScope.tsx
matuzalemsteles: Yeah, I agree, I also don't like to see an `any` here so I will go ahead and just set `React.ReactElement` for children so we can get rid of it.

packages/clay-shared/src/FocusScope.tsx
matuzalemsteles: Yeah, I'll write something, but just stressing that it's just for our internal use.

matuzalemsteles:
> Overall it looks pretty good. What do you think about making this a higher order component that we might call similarly to redux's connect(). That way we wouldnt be forced to always nest our components one step deeper inside of <FocusScope />

Hmm, I'm not so sure about that, in many cases I may want to wrap the whole component in `<FocusScope />` but there are some particular cases like MultiSelect, where I want to control just part of the focus, usually we will need it to cases where we are using Portal.

We also have another design issue if that were so, `<FocusScope />` needs to know `focusManager` to use its internal methods and` focusManager` needs to be in the component to create the scope so we might not be able to create a  `withFocusScope`, maybe we can get around that but I don't know if it's going to be so beneficial. Do you have any idea?

matuzalemsteles:
@bryceosterhaus update and rebase.

Issue #1567, opened 2019-02-18
RFC: Explore using a centralized script runner abstraction


wincent:
There is a lot of discussion in #1548 about how to set up packages moving forward, but there is related opportunity that we could explore soon. Creating this separate RFC issue so as not to delay or derail that PR.

# Objective

Explore using a centralized script runner abstraction so that we can easily run tasks at root or individual package level, and have painless updates as we evolve our task implementations.

# Current set-up

One of the things the generator does is create a `package.json` from a template with a set of hard-coded build scripts:

```
"scripts": {
	"build": "babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
	"build:types": "tsc --project ./tsconfig.declarations.json",
	"prepublish": "npm run build && npm run build:types",
	"start": "webpack-dev-server --mode development",
	"test": "jest --config ../../jest.config.js"
},
```

This is good because it means we don't have to stop and think about how to set up scripts for every package we create, nor do we risk copy-paste errors that could arise if we weren't using a generator to do the work for us.

However, just say we decide to change how we run the tests or do the build in the future, then every package's "package.json" needs to be updated.

# RFC set-up

If we had an abstraction that knew how to do those "script" tasks, then we could use that instead of the individual tools (`babel`, `tsc`, `npm`, `webpack-dev-server`, `jest` etc). If the underlying individual tool changed, then we would just update our abstraction in one place and it would take effect in all the packages immediately without any need to edit the "package.json" files. Additionally, with an abstraction in place designed to work in a monorepo, we have a single location that we can imbue with knowledge about the structure of the project, to give it the "intelligence" necessary to work identically at the root level or within individual packages (the latter is something that @bryceosterhaus brought up as being important in that other PR).

That's all very abstract, so what would this "abstraction" look like? I prototyped one way it could look like in [a side project](https://github.com/wincent/js):

* The abstraction is just another package called [workspace-scripts](https://github.com/wincent/js/tree/e62623354f786c7b306130a8a79c746ede461bc1/packages/workspace-scripts); it is basically an executable that knows how to run root-level and package-level tasks in the context of a monorepo that uses Yarn workspaces.
* All of the actual work is done by [this script](https://github.com/wincent/js/blob/master/packages/workspace-scripts/src/index.ts); if you look inside it you'll see it's calling `eslint`, `tsc` etc.
[...]

PR #2238, opened 2019-08-01
fix(tsconfig.json): consume clay packages from `node_modules` to correctly generate of the types

### LINKED ISSUE #2235: Typescript autocompletion is no longer working
Since we updated our components to `3.0.0-milestone.2` VS Code is no longer helping me to write components with props autocompletion. That was a productivity boost, so I'm missing it. Thanks!

matuzalemsteles:
Fixes #2235

I just changed the `@clayui/*` package path so that typescript looks inside `node_modules`, so it does not consider that it should generate types at build time but will still look at them to make sure everything is fine.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,7 @@
 		"allowJs": true,
 		"baseUrl": ".",
 		"paths": {
-			"@clayui/*": ["packages/clay-*/src"],
+			"@clayui/*": ["node_modules/@clayui/*/src"],
 			"*": ["custom-types/*", "node_modules/*", "packages/*"]
 		}
 	},


tsconfig.json
```diff
@@ -11,7 +11,7 @@
 		"allowJs": true,
 		"baseUrl": ".",
 		"paths": {
-			"@clayui/*": ["packages/clay-*/src"],
+			"@clayui/*": ["node_modules/@clayui/*/src"],
```
bryceosterhaus: Wont this defeat the purpose of the monorepo? I think if we do this, we won't be importing the local packages while developing, it would look for it in node_modules instead.

matuzalemsteles: I was worried about that too but had no problems testing locally, as we are still using the yarn workspace it generates a symbolic link from the packages in the project's `node_modules` root so we will still keep looking at the same files as before but when generating the build he will not add the types together.

bryceosterhaus: oh interesting, I didnt think about how yarn would impact that.

So how come this fixes the issue for generating types? It seems like either solution should work.

matuzalemsteles: Yeah, this is a bit curious but when we point to files outside `node_modules` as in the case for `packages/...` TypeScript was considering that it should compile its types also because it is outside `node_modules`, we could ignore those types adding in the `exclude` setting but that would break everything, now when we point inside `node_modules` TypeScript interprets it as a dependency even though we added it in the `exclude` setting will work fine and will not generate the types because it understands that it is a dependency.

bryceosterhaus: ah i see. hmm, this seems like a strange issue and something thats been solved before. Have you seen any other libraries do the same thing? I mostly just want to avoid config changes that "happen to work" without having followed a documented API approach.

matuzalemsteles: Sure, we haven't set it up yet to point to `node_modules`. This was broken before when we set TypeScript to generate types.

I've tried checking on other projects but never able to find a project similar to our style, as dependencies between packages and build in each package, usually only have one build that runs at the root.

I think it makes sense to point to `node_modules` since we are treating packages as dependencies on each other is as close as we can get to a Clay package usage environment.

bryceosterhaus: yeah I guess that makes sense to me as well. Im cool with this change, but we should just be aware in case there might be any strange behaviors with inter-package dependencies.

wincent: > I've tried checking on other projects but never able to find a project similar to our style, as dependencies between packages and build in each package, usually only have one build that runs at the root.

I think [my toy project](https://github.com/wincent/js) might have those attributes (eg. [a package](https://github.com/wincent/js/tree/master/packages/stable-stringify) with [a dependency](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/packages/stable-stringify/package.json#L33) on [another package](https://github.com/wincent/js/tree/master/packages/is-object) in the monorepo; and each package has [it's own build](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/packages/is-object/package.json#L19) where I build both [declarations](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/Makefile#L17) and [JS files](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/Makefile#L21)).

To see what it is actually doing when I run `yarn run build` in one of the packages, you can see the commands that are run here:

```
yarn run v1.17.3
$ workspace-scripts build stable-stringify
env BABEL_ENV=production babel --root-mode upward --source-root packages/stable-stringify/src --ignore '**/__tests__/**/*' --relative -d ../lib --extensions .js,.ts packages/stable-stringify/src
tsc --emitDeclarationOnly --declaration --project tsconfig-declaration.json
Successfully compiled 1 file with Babel.
env BABEL_ENV=es babel --keep-file-extension --root-mode upward --source-root packages/stable-stringify/src --ignore '**/__tests__/**/*' --relative -d ../lib --extensions .js,.ts packages/stable-stringify/src
Successfully compiled 1 file with Babel.
find packages/stable-stringify/lib -name '*.ts' -not -name '*.d.ts' -exec sh -c 'mv -f "$0" "${0%.ts}.mjs"' {} \;
find packages/stable-stringify/src -name '*.js.flow' -exec bash -c 'cp "$0" "${0/\/src\///lib//}"' {} \;
find packages -path 'packages/*/src/*' -and -name '*.d.ts' -and -not -path '*/node_modules/*' -exec bash -c 'mv "$0" "${0/\/src\///lib//}"' {} \;
✨  Done in 6.01s.
```

[This is the `tsconfig.json`](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/tsconfig.json) used for the source build, and [this one](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/tsconfig-declaration.json) extends it and is used to build the declarations. Notably, [the "paths" configuration](https://github.com/wincent/js/blob/4a035a69590643840644c69794a39322354091f5/tsconfig.json#L13) does not need "node_modules" in it:

```
    "paths": {
      "@wincent/*": ["*/src"]
    }
```

So I am not sure why this hack is required in Clay; it should be possible without it.

matuzalemsteles: Hmm this is interesting I tried changing the "path" to:
```
"@clayui/*": ["*/clay-*/src"]
"@clayui/*": ["clay-*/src"]
```

I made other settings in `tsconfig.json` also to approximate your but still still giving error, I don't know if it has to do with your `workspace-scripts` (probably)... but Storybook breaks and VSCode also.

wincent: > I don't know if it has to do with your `workspace-scripts` (probably)

`workspace-scripts` is just a "thin wrapper" that runs the commands I pasted above; ie:

```
# actual build
env BABEL_ENV=production babel --root-mode upward --source-root packages/stable-stringify/src --ignore '**/__tests__/**/*' --relative -d ../lib --extensions .js,.ts packages/stable-stringify/src

# emit declaration
tsc --emitDeclarationOnly --declaration --project tsconfig-declaration.json
```

(and now that I look at it, I see it is probably making `.d.ts` files for all packages in the monorepo — whoops — but the actual build is only running against one package)

matuzalemsteles: I think our build process could eventually be improved by adopting RFC #1567, but I think we can go about it safely. @bryceosterhaus If you're unsure about this, do some manual testing to make sure we're going the right way for now.

bryceosterhaus: I'm fine with doing something for now to alleviate the issue at hand. But maybe we could also open an additional issue to make sure we further address this in the future.

matuzalemsteles: ok, I'll go ahead with it. Issue created #2249.

wincent: Cross-referencing [this comment](https://github.com/liferay/clay/issues/2235#issuecomment-519809702) about unwanted "node_modules" prefix getting auto-inserted by VSCode, possibly related to this change.

bryceosterhaus:
I've been running into issues with this commit lately. If I am working in `Modal` I end up adding new props to `Portal` from `@clayui/shared`, the type definitions don't get picked up properly in the Modal directory. I fix it by reverting this commit.

matuzalemsteles:
hey @bryceosterhaus have you checked if it is a VSCode problem? I usually have problems with VSCode, I've had similar cases but haven't faced this problem yet.

tsconfig.json
matuzalemsteles: hey @wincent this is weird, we haven't released any of these changes yet and it wasn't supposed to happen since VSCode doesn't read the `tsconfig` file, it reads the `types` property from the package.json file from package to identify where the types are.

bryceosterhaus:
It doesn't seem to be from vscode, it seems more that the `node_modules` version of dependencies is not the same as the `/packages/*` version. Perhaps the node_modules version is looking at lib, and not src?

matuzalemsteles:
> It doesn't seem to be from vscode, it seems more that the node_modules version of dependencies is not the same as the /packages/* version. Perhaps the node_modules version is looking at lib, and not src?

I think this is not possible, because we are making it explicit that he should look at the `src` folder.

Can you look at your `node_modules` and check if the folders inside `@clayui` have symbolic links? running a `yarn` would solve this problem.

I had a similar problem but restarting and closing the file and opening again seemed to solve the problem, the same thing happened before this fix.

bryceosterhaus:
We can't force our config to look at `src` though since we don't want that when its published, only in our dev space.

Yeah the `node_modules` symlinked to the packages, but I believe its because when its in node_modules the package.json points to non-src, which means we need to rebuild the type definitions for lib if we make any changes.
```
"main": "lib/index.js",
"types": "lib/index.d.ts",`
```

matuzalemsteles:
> We can't force our config to look at src though since we don't want that when its published, only in our dev space.

Yes we do not force in production environment actually would have no vision because the `tsconfig` file is for dev environment or "pre-build ".

> Yeah the node_modules symlinked to the packages, but I believe its because when its in node_modules the package.json points to non-src, which means we need to rebuild the type definitions for lib if we make any changes.

This is strange because I'm pointing to `src` in theory he should look there and not lib, maybe this is happening because the packages have the `lib` folder with the build, in my test I don't have them, so in theory if he had done that he should not have found the types. You can try removing them and checking if VSCode is missing the types.

If VSCode can see the types but after updating the file it has not updated yet, try closing the file and opening again and waiting a while is the time to load it all over again.

```
"paths": {
	"@clayui/*": ["node_modules/@clayui/*/src"],
}
```

PR #2240, opened 2019-08-02
feat(@clayui/drop-down): create DropDownWithItems component

### LINKED ISSUE #2216: Create high-level DropDownWithAdvancedItems component
The idea of ​​this component is to be able to define through the API the use of components, checkbox, search, radio, grouping... in DropDown. Following the idea as we had in v2.

hey @bryceosterhaus let me know what you think about it.

### LINKED ISSUE #2221: ClayDataProvider doesn't work on IE11
Importing `@clayui/data-provider` fails on IE11.

Error happens in [yallist](https://www.npmjs.com/package/yallist) dependency, in this line:

```js
Yallist.prototype[Symbol.iterator] = function* ()
```

Error in IE11 console is: `Expected '('`.

Generators are not being correctly transpiled to something IE11 understands.

Dependency tree: `#frontend-taglib-clay#@clayui#data-provider#lru-cache#yallist`

matuzalemsteles:
Fixes #2216

I followed the same approach we had in v2, maybe something that does not have here is to render a custom item, I will think of doing something.

But this is just a draft to explore other ideas, there is still a lack of testing and documentation, I will work on it once we are fine with this component.

- [ ] Tests
- [ ] Docs


packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
```diff
@@ -0,0 +1,247 @@
...
+import warning from 'warning';
+
+type TType = 'checkbox' | 'group' | 'item' | 'radio' | 'radiogroup';
+
+interface IAdvancedItem extends Omit<IItem, 'type'> {
+	checked?: boolean;
+	items?: Array<IAdvancedItem>;
+	name?: string;
+	onChange?: Function;
+	separator?: boolean;
```
matuzalemsteles: I decided to remove the type `divider`, it felt weird to create a new object just to do it and ended up getting feedback that it looked better and easier to work with this, if we are fine with it I will do the same for` ClayDropDownWithBasicItems`.

packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
```diff
@@ -0,0 +1,247 @@
...
+	spritemap?: string;
+
+	/**
+	 * List of items to display in drop down menu
+	 */
+	items: Array<IAdvancedItem>;
+}
+
+interface IProps extends IDropDownContentProps {
+	action?: React.ReactElement;
```
bryceosterhaus: Since we are passing a react element here, I think its best if we label this as something like `footerContent`. `action` makes it seem like its a callback

packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
```diff
@@ -0,0 +1,247 @@
...
+	action,
+	caption,
+	helpText,
+	items,
+	onSearchValueChange = () => {},
+	searchable,
+	spritemap,
+	trigger,
+}) => {
+	const [searchValue, setSearchValue] = useState('');
```
bryceosterhaus: Rather than using internal state here, I think it might be best to make this a controlled component with both `onSearchValueChange` and `searchValue`.

packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
```diff
@@ -0,0 +1,247 @@
...
+	const [active, setActive] = useState(false);
+
+	const hasRightSymbols = !!items.find(item => item.symbolRight);
+	const hasLeftSymbols = !!items.find(item => item.symbolLeft);
+
+	const Wrap = action ? 'form' : React.Fragment;
```
bryceosterhaus: How come we wrap this in `form` if there is action content? Is this primarily for accessibility?

matuzalemsteles:
> In regards to the "seperator vs divider" only reason I prefer the divider is because its a bit more explicit and seems easier to create that list of items via pushing to an array vs modifying a certain element of an array.

Hmm I'm not so sure about that, I see that most of the time you will have a static value but I am seeing this change as another form of aesthetics but we can follow with the `divider` and see how it will work.

> Also, since this component seems to also work for the *WithBasicItems component, do you think we should just combine the two?

Hmm yeah, I think we can remove `*WithBasicItems` and rename this to `*WithItems` or Advanced get more explicit?

packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
matuzalemsteles: Yeah, I'm going with `footerContent`.

packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
matuzalemsteles: Hmm yeah, I think maybe it's even better this way, devs will be able to filter items when needed.

packages/clay-drop-down/src/DropDownWithAdvancedItems.tsx
matuzalemsteles: Yes we are considering that the `action` API can be treated as a button to be a form action for example in other cases.

bryceosterhaus:
`*WithItems` sounds good to me

matuzalemsteles:
hey @bryceosterhaus update. I've been thinking of adding in the `items` API the possibility of passing a custom type to render any content, it's something we had in v2 for extension points, so I'm thinking about it. What do you think?

bryceosterhaus:
@matuzalemsteles I would be hesitant to add that sort of extension point until its absolutely needed or can't be addressed in another matter. Mostly because these sort of open-ended extension points become hard to maintain and open the door for non-lexicon patterns.

I'm open to adding it in the future, but I generally prefer to start with a relatively closed API and slowly open it when needed. It's much easier to add new APIs than it is to remove existing API.

matuzalemsteles:
Hey @bryceosterhaus I understand your concern and it was the same thoughts we had when we were starting with Clay the counterpoint of this is that as we are a lib of components people expect them to be able to build things beyond or extend to streamline their processes we have to Think of it as a foundation for building great apps quickly.

Comparing the Portal from 7.1 to 7.3 (now) several other use cases and new products have emerged, and Clay's adoption is limited by not keeping up with the changes leaving people frustrated for not using them and not exceeding their expectations. We will always have to go back every year because products evolve fast and we have to try to predict some of them and stay ahead when possible. Following this the maintenance cost becomes super high.

So we have to be careful to think that starting small means it will be easier, actually it will be harder because we are starting with something closed and opening it to new things or paradigms becomes more difficult because we have to open new APIs and depreciate others and go opening. things over time. I think getting flexible (always putting on the scale) is easier going by closing the ends.

I believe we should start by opening the APIs a bit and not repeat what we did in Clay v2 to make people happier, I know we draw the low-levels for these cases but they won't always cover or work hard so I still prefer that We tried to make the high levels a bit flexible.

bryceosterhaus:
> Clay's adoption is limited by not keeping up with the changes leaving people frustrated for not using them and not exceeding their expectations

This seems like a much more complicated issue than what we provide from Clay. Only 2 or 3 modules are currently using react, we can't have a healthy judgement of how useful our components are until there is a greater adoption.

>  I know we draw the low-levels for these cases but they won't always cover or work hard so I still prefer that We tried to make the high levels a bit flexible.

The more flexible we make the high-level components, the more we will blur the line between high and low level. I fear we will blur the line so much that it becomes unclear to developers which to use. High-level components are easily susceptible _to death by a thousand ~cuts~ props_, each prop is fairly harmless in and of themselves but we need to be able to draw a line where high-level ends and low-level begins. IMO, this line is the lexicon definition + clay-css implementation. If we go _beyond_ those, we start to introduce complexity and overhead that should likely fall on specific apps and not the library the app consumes.


Overall, I think this conversation goes back to our [component offering ADR](https://github.com/liferay/clay/blob/master/adr/component_offerings.md). We likely need to expand on this in greater detail and discuss where we should draw the line for the clay library. I think the big question is how do we best provide a component library to teams here at Liferay. Do we intend to go deep in developing full functionality for each teams need, or do we focus on providing a foundation of building blocks for the most common use cases. I think it would be helpful to fully flesh out a feature request cycle and what our base requirements are. Base requirements are likely the core lexicon definitions, but how do we accomodate beyond those defintions. I'd be curious to hear other thoughts as well, and I'd be happy to move this conversation to another issue to keep it in one location. @jbalsas what do you think?

matuzalemsteles:
> This seems like a much more complicated issue than what we provide from Clay. Only 2 or 3 modules are currently using react, we can't have a healthy judgement of how useful our components are until there is a greater adoption.

This perception I have is more due to v2, so it is interesting to learn what we went through there so as not to go through it again here and that's what I realized when I was talking about Clay.

> The more flexible we make the high-level components, the more we will blur the line between high and low level. I fear we will blur the line so much that it becomes unclear to developers which to use.

Yes you are right, I am not saying that we should make the high levels totally flexible that we will not know what is high level and low level but I am saying that we do not need to create components that they become very distant from each other high level at one end and low-level at the other end, so I can think of this:

- High-level tail: The distance from high-level to low-level makes the component with a very short life cycle because the changes of this component will change with a certain frequency and the pain to replace them is greater and consequently we generate pain for those who use it or maybe not...
- Low-level tail: It can be very flexible and its components have a very high life cycle because they are the basis for building new high-level and extensions plus they are so flexible that when a high-level does not meet a team the pain to build the same can be very large.

I mean is that we shouldn't leave everything in the extreme, but in the middle and create a balance between them, high levels can be flexible so as not to kick out other teams who want to use the component because it covers the vast majority of its use cases minus a few.

Low-levels don't have to be so flexible otherwise devs will have to work hard to get to the same component we did at high-level just to fit your use case.

> Overall, I think this conversation goes back to our component offering ADR. We likely need to expand on this in greater detail and discuss where we should draw the line for the clay library. I think the big question is how do we best provide a component library to teams here at Liferay. Do we intend to go deep in developing full functionality for each teams need, or do we focus on providing a foundation of building blocks for the most common use cases. I think it would be helpful to fully flesh out a feature request cycle and what our base requirements are. Base requirements are likely the core lexicon definitions, but how do we accomodate beyond those defintions.

Yes I agree with you on that, we are based on Lexicon but I have an impression that over time Lexicon has come in a line of only predicting very specific use cases of Portal teams and that the new products that are emerging today. They are not using these core components (I mean the larger components) because they were not foreseen, I know this is very complex and complicated to deal with.

In the end the general idea is always to do all Lexicon use cases but still leaving escape doors, new products or new features will always do something new and different that we don't have and that way we won't be able to achieve the purpose of getting product teams to develop quickly and build complete prototypes.

matuzalemsteles:
hey @bryceosterhaus this sounds like a really long discussion, we can go without it for this component and we can exploit it at some point, I think we'll have demand for it at some point.

matuzalemsteles:
hey @bryceosterhaus update, Let me know if you see anything else before I get on with it.

PR #2243, opened 2019-08-05
refactor: remove another cast

wincent:
Noticed this while looking at: https://github.com/liferay/clay/pull/2223

This is perhaps a better fix than casting to `any`, which can hide legitimate errors.


wincent:
@matuzalemsteles, @bryceosterhaus: I don't know whether this is the best fix or not, but sending it for consideration anyway. I only "tested" it in a minimal way (by building the clay-drop-down package and confirming that it didn't report any type errors).

PR #2244, opened 2019-08-06
 fix(@clayui/data-provider): add LRUCache implementation

### LINKED ISSUE #2221: ClayDataProvider doesn't work on IE11

matuzalemsteles:
Fixes #2221

### DIFF
--- a/packages/clay-data-provider/src/LRUCache.ts
+++ b/packages/clay-data-provider/src/LRUCache.ts
@@ -1,0 +1,54 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const DEFAULT_MAX_SIZE = 20;
+
+export class LRUCache<K, V> {
+	private _maxSize: number;
+	private _storage: Map<K, V>;
+
+	constructor(maxSize: number = DEFAULT_MAX_SIZE) {
+		if (maxSize <= 0) {
+			throw new TypeError('LRUCache: max must be a non-negative number');
+		}
+
+		this._maxSize = maxSize;
+		this._storage = new Map();
+	}
+
+	reset() {
+		this._storage = new Map();
+		return this._storage;
+	}
+
+	has(key: K) {
+		return this._storage.has(key);
+	}
+
+	get(key: K) {
+		if (this._storage.has(key)) {
+			const value = this._storage.get(key);
+			this._storage.delete(key);
+			this._storage.set(key, value as V);
+
+			return value;
+		}
+	}
+
+	set(key: K, value: V) {
+		if (this._storage.size >= this._maxSize) {
+			const iterator = this._storage.keys();
+			let count = this._maxSize - (this._storage.size - 1);
+			while (count) {
+				const {value} = iterator.next();
+				this._storage.delete(value);
+				count--;
+			}
+		}
+
+		this._storage.set(key, value);
+	}
+}
--- a/packages/clay-data-provider/src/__tests__/LRUCache.ts
+++ b/packages/clay-data-provider/src/__tests__/LRUCache.ts
@@ -1,0 +1,48 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LRUCache} from '../LRUCache';
+
+let cache: LRUCache<string, {}>;
+
+describe('LRUCache', () => {
+	beforeEach(() => {
+		cache = new LRUCache(3);
+	});
+
+	it('throws an error if initialized with maxSize 0', () => {
+		expect(() => new LRUCache(0)).toThrow(/non-negative number/);
+	});
+
+	it('throws an error if initialized with a negative maxSize', () => {
+		expect(() => new LRUCache(-10)).toThrow(/non-negative number/);
+	});
+
+	it('returns new storage when cache is reset', () => {
+		expect(cache.reset().size).toBe(0);
+	});
+
+	it('returns undefined for undefined key', () => {
+		expect(cache.get('foo')).toBe(undefined);
+	});
+
+	it('returns value for present value', () => {
+		cache.set('foo', 'f');
+		expect(cache.get('foo')).toBe('f');
+	});
+
+	it('returns undefined for key less used beyond cache capacity', () => {
+		cache.set('foo', 'f');
+		cache.set('bar', 'b');
+		cache.set('baz', 'bz');
+		cache.get('foo');
+		cache.set('qux', 'q');
+		expect(cache.get('qux')).toBe('q');
+		expect(cache.get('baz')).toBe('bz');
+		expect(cache.get('bar')).toBe(undefined);
+		expect(cache.get('foo')).toBe('f');
+	});
+});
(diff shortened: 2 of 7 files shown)


packages/clay-data-provider/src/LRUCache.ts
```diff
@@ -0,0 +1,56 @@
...
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const DEFAULT_MAX_SIZE = 20;
+
+export class LRUCache<K, V> {
+	maxSize: number;
+	storage: Map<K, V>;
```
wincent: Shouldn't these be `private` or at least `_maxSize`/`_storage`? Otherwise I can break encapsulation.

packages/clay-data-provider/src/LRUCache.ts
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	has(key: K) {
+		return this.storage.has(key);
+	}
+
+	get(key: K) {
+		const value = this.storage.get(key);
+
+		if (value) {
```
wincent: This check is probably not correct, because I could create a cache that stores booleans or other `false`-y values, and those wouldn't get bumped in the cache. That's why in [my implementation](https://github.com/wincent/masochist/blob/8ee6fc3704b204adae6dc3ad9baca5108a741637/src/server/LRUCache.js) that I showed you it does a `storage.has()` check instead.

packages/clay-data-provider/src/LRUCache.ts
```diff
@@ -0,0 +1,56 @@
...
+			let count = this.maxSize - (this.storage.size - 1);
+			while (count) {
+				const {value} = iterator.next();
+				this.storage.delete(value);
+				count--;
+			}
+		}
+
+		this.storage.set(key, value);
+		return true;
```
wincent: Do you really want to return `true` here? It seems pointless to have a function that always returns `true` no matter what. Given that the cache is basically a `Map`, we could return `this` like `Map.prototype.set` does (that's what I did in [mine](https://github.com/wincent/masochist/blob/8ee6fc3704b204adae6dc3ad9baca5108a741637/src/server/LRUCache.js#L52) — this means you can chain multiple calls; eg. `new LRUCache().set('a', 1).set('b', 2)` — or just `undefined` (that's what [the other one](https://github.com/facebook/relay/commit/3f5d799fa8beb5f810f2bd37813674fc22598b74#diff-37a526caff576f11bac88a48c97391ebR49) I linked you to does).

Looking at how you are using this return value, I see it goes through many layers of indirection inside `useCache` and ultimately winds up discarded inside `useResource` when it does `cache.set(result)` and ignores the return value.

packages/clay-data-provider/src/__tests__/LRUCache.ts
```diff
@@ -0,0 +1,63 @@
...
+let cache: LRUCache<string, {}>;
+
+describe('LRUCache', () => {
+	beforeEach(() => {
+		cache = new LRUCache(3);
+	});
+
+	afterEach(() => {
+		cache.reset();
+	});
```
wincent: Delete this; you don't need to reset the object because you get a brand new object anyway each time.

packages/clay-data-provider/src/__tests__/LRUCache.ts
```diff
@@ -0,0 +1,63 @@
...
+
+	it('throw error if initialized with maxSize 0', () => {
```
wincent: "throw error if" isn't grammatical, so please change to: "throws if" or "throws an error if".

Same in the next example.

packages/clay-data-provider/src/__tests__/LRUCache.ts
```diff
@@ -0,0 +1,63 @@
...
+	it('returns undefined for undefined key', () => {
+		expect(cache.get('foo')).toBe(undefined);
+	});
+
+	it('returns value for present value', () => {
+		cache.set('foo', 'f');
+		expect(cache.get('foo')).toBe('f');
+	});
+
+	it('least recently set', () => {
```
wincent: Can you change this into a sentence that reads correct when read out loud? ie. "it something something something". I don't actually know what you are testing here, or in the next one.

Compare that with [these tests I wrote here](https://github.com/wincent/masochist/blob/8ee6fc3704b204adae6dc3ad9baca5108a741637/src/server/__tests__/LRUCache-test.js#L41-L58), where the `it` and the comments describe what is actually expected and happening:

```javascript
describe('set()', () => {
  it('evicts old items', () => {
    // Populate the cache.
    cache.set('a', '#a');
    cache.set('b', '#b');
    cache.set('c', '#c');
    expect(cache.has('a')).toBe(true);
    expect(cache.has('b')).toBe(true);
    expect(cache.has('c')).toBe(true);

    // Cause an eviction.
    cache.set('d', '#d');
    expect(cache.has('a')).toBe(false);
    expect(cache.has('b')).toBe(true);
    expect(cache.has('c')).toBe(true);
    expect(cache.has('d')).toBe(true);
  });
});
```

packages/clay-data-provider/src/LRUCache.ts
matuzalemsteles: Hmm, I had forgotten about these values ​​that may be here, thanks for this @wincent.

packages/clay-data-provider/src/LRUCache.ts
matuzalemsteles: Oh yeah, I'm hoping no one will touch the instance of `LRUCache`, I tried to use `Symbol` as a key in `useCache` to avoid these kinds of things. But I was thinking of going with `_maxSize`/`_storage` but I find the underline so horrible for that 😅, I will go with `private` for more security here.

packages/clay-data-provider/src/LRUCache.ts
matuzalemsteles: Hmm, good question, this would probably be a trade-off in returning `true` or `undefined` in the end would be the same as returning `true` even if you don't use it but the purpose is just another meaning of showing that everything will always work out and Assuming that the method actually did is `true` then `true`, `1`... it would make little sense to return `undefined`, `NuN`, `false` because they are treated as negative or nothing happened etc...

I think in the end I prefer to keep the chainable, it seems to be more useful but maybe not use here, so I prefer to go with the `undefined` return here to be in the middle ground 🙂.

packages/clay-data-provider/src/__tests__/LRUCache.ts
matuzalemsteles: Oh yeah, I guess I just put it in my mind that `new LRUCache` would only run once... even if it's `beforeEach` 😂

packages/clay-data-provider/src/LRUCache.ts
wincent: Yeah, seeing as this is a TypeScript project, `private` (and `#` private class fields once those get implemented) is probably preferable over `_` prefix. Even with `private` it is not literally private, it is only advisory. TypeScript will tell you if you do it wrong, but you can cast away:

```javascript
// Property 'x' is private and only accessible within class 'Thing'.
thing.x = 11;

// 😃 
(thing as any).x = 11;
```

But still, its good to put some measure in place to stop me going `cache.storage.clear()` from the outside.

matuzalemsteles:
Hey guys, I'm going ahead with this, I applied the requested changes.

PR #2245, opened 2019-08-06
feat(tabs): Creates Tabs component

diegonvs:


packages/clay-tabs/src/Item.tsx
```diff
@@ -40,9 +40,9 @@ interface IProps extends Omit<React.HTMLAttributes<HTMLLIElement>, 'onClick'> {
 	>;
 
 	/**
-	 * Registers the `tabKey` for the Tab Item.
+	 * Registers the `tabkey` for the Tab Item.
 	 */
-	tabKey: number;
+	tabkey: number;
```
wincent: Why the change from `tabKey` to `tabKey`? It is inconsistent with other props like `tabName` and `activeTabKey`?

packages/clay-tabs/src/Content.tsx
```diff
@@ -29,13 +34,9 @@ const Content: React.FunctionComponent<IProps> = ({
...
 		<div className={classNames('tab-content', className)} {...otherProps}>
 			{React.Children.map(children, (child, index) => {
-				if (!React.isValidElement(child)) {
-					return null;
-				}
-
 				return React.cloneElement(child, {
 					...child.props,
-					active: activeTabKey === child.props.tabKey,
+					active: activeTabKey === child.props.tabkey,
```
wincent: I feel a bit queasy about accessing `child.props` from the outside. The fact that the `TabPane` component takes `tabkey` as a prop but doesn't actually access it is another sign that this is a smell.

This feels like the kind of thing that the component should take care of instead of the user; ie. in the example below, having to provide a `tabkey` is just expanding with surface area of the API with another unique prop (like `tabName`), which is another thing for me to have to worry about and possibly get wrong. It feels a bit like a leaky abstraction; we're exposing what would ideally be internal-only plumbing:

```javascript
<ClayTabs>
	<ClayTabs.Item
		active={activeTabKeyValue == 0}
		onClick={onClick}
		tabName="Dummy1"
		tabkey={0}
	/>
	<ClayTabs.Item
		active={activeTabKeyValue == 1}
		data-testid="tabItem2"
		onClick={onClick}
		tabName="Dummy2"
		tabkey={1}
	/>
	<ClayTabs.Item
		active={activeTabKeyValue == 2}
		onClick={onClick}
		tabName="Dummy3"
		tabkey={2}
	/>
</ClayTabs>
```

it would be nice if we could take care of that. I have a feeling that could be made to work; did you explore that in the initial implementation?

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+					{...itemElementProps}
+				>
+					{tabName}
+					{dropdown && (
+						<ClayIcon spritemap={spritemap} symbol="caret-bottom" />
+					)}
+				</button>
+			</>
+		);
+	}
```
wincent: Should be no need to wrap in a fragment in either of these places because you're only returning a single element. Also, there is a lot of duplication between the `a` and `button` cases, can we factor that out a bit (for example, but spreading in the common props)?

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+			className={classNames(
+				'nav-item',
+				{
+					dropdown,
+				},
+				className
+			)}
+			{...otherProps}
+		>
+			{tabName && (
```
wincent: `tabName` is a required parameter, so why the guard here?

packages/clay-tabs/src/TabPane.tsx
```diff
@@ -0,0 +1,72 @@
...
+import React, {useEffect, useState} from 'react';
+
+interface ITabPaneProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Flag to indicate if `active` classname should be applied
+	 */
+	active?: boolean;
+
+	/**
+	 * Flag to indicate if `fade` classname that applies an fading animation should be applied.
```
wincent: "an fading" → "a fading"

packages/clay-tabs/src/TabPane.tsx
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * TabPane's key.
+	 */
+	tabkey: number;
+}
+
+const delay = (fn: Function) => {
+	return setTimeout(() => {
+		fn();
+	}, 100);
```
wincent: Where did this value come from? Could it be moved into a named constant?

packages/clay-tabs/src/__tests__/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+};
+
+describe('ClayTabs', () => {
+	afterEach(cleanup);
+
+	it('renders', () => {
+		const {container} = render(<ClayTabs />);
+
+		expect(container).toMatchSnapshot();
+	});
```
wincent: Don't we usually have blank lines between `it()` blocks? (Please ignore me if I'm mistaken about that)

packages/clay-tabs/src/Item.tsx
diegonvs: Yes, It's unnecessary. I was relying on Typescript to warn me about this. Removing 😅

packages/clay-tabs/src/TabPane.tsx
diegonvs: ✅

packages/clay-tabs/src/TabPane.tsx
diegonvs: Sure, I was minding about turn this delay a property. Makes sense?

/cc @bryceosterhaus @matuzalemsteles too

packages/clay-tabs/src/Item.tsx
diegonvs: Makes sense Greg, making changes.

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+							/>
+						)}
+					</a>
+				</>
+			);
+		}
+
+		return (
+			<>
+				<button
```
matuzalemsteles: Could we use `ClayButton` here?

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+			onClick,
+			spritemap,
+			tabName,
+		}: IProps,
+		IProps
+	) => {
+		if (component === 'anchor') {
+			return (
+				<>
+					<a
```
matuzalemsteles: Could we use `ClayLink` here?

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+						aria-controls={tabName.toLowerCase()}
+						aria-disabled={!active}
+						aria-selected={!active}
+						className={classNames('nav-link', {
+							active,
+							disabled,
+						})}
+						data-testid={`${tabName.trim().toLowerCase()}TabItem`}
+						href={`#${tabName.trim().toLowerCase()}`}
+						id={`${tabName}Tab`}
```
matuzalemsteles: Just `tabName` would be enough here.

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+					<a
+						aria-controls={tabName.toLowerCase()}
+						aria-disabled={!active}
+						aria-selected={!active}
+						className={classNames('nav-link', {
+							active,
+							disabled,
+						})}
+						data-testid={`${tabName.trim().toLowerCase()}TabItem`}
+						href={`#${tabName.trim().toLowerCase()}`}
```
matuzalemsteles: I think we should not impose that `href` is always an anchor, we should expose a `href` API instead of making it Implicit.

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+						id={`${tabName}Tab`}
+						onClick={() => onClick && onClick(tabkey)}
+						ref={forwardRef}
+						role="tab"
+						tabIndex={disabled ? -1 : undefined}
+						{...itemElementProps}
+					>
+						{tabName}
+						{dropdown && (
```
matuzalemsteles: I prefer this to be `showIcon` is more generic, `dropdown` brings another feeling.

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+					id={`${tabName}Tab`}
+					onClick={() => onClick && onClick(tabkey)}
+					ref={forwardRef}
+					role="tab"
+					tabIndex={disabled ? -1 : undefined}
+					type="button"
+					{...itemElementProps}
+				>
+					{tabName}
+					{dropdown && (
```
matuzalemsteles: The same here!

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,185 @@
...
+		);
+	}
+);
+
+const Item: React.FunctionComponent<IProps> = ({
+	active = false,
+	component,
+	className,
+	disabled,
+	dropdown,
```
matuzalemsteles: The same here!

packages/clay-tabs/src/TabPane.tsx
matuzalemsteles: I prefer to keep it away from the API, because this is just a fake time to cause the animation, developers don't need to know that.

packages/clay-tabs/src/__tests__/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+		fireEvent.click(getByTestId('oneTabItem'));
+		expect(onClick).toBeCalled();
+		expect(onClick).toBeCalledWith(1);
+
+		fireEvent.click(getByTestId('twoTabItem'));
+		expect(onClick).toBeCalled();
+		expect(onClick).toBeCalledWith(2);
+
+		expect(container).toMatchSnapshot();
```
matuzalemsteles: It is necessary to run a `toMatchSnapshot` here, we will try to minimize the use of snapshot in our tests.

packages/clay-tabs/src/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+			role="tablist"
+			{...otherProps}
+		>
+			{children}
+		</ul>
+	);
+};
+
+ClayTabs.Content = Content;
+
```
matuzalemsteles: No space required here.

packages/clay-tabs/src/Item.tsx
diegonvs: I was getting errors with this property using camel case. See the following stacktrace:
```log

packages/clay-tabs/src/Content.tsx
diegonvs: Yes, Initially I implemented using the order of the `Item`s that matches with the order of `Content` but I wasn't able to use Dropdown using this logic.

Personally, I prefer this version of the [API](https://github.com/diegonvs/clay/blob/47526867448545c7372466cb55e58a1e0cab4f0f/packages/clay-tabs/stories/index.tsx), however, I don't know what is the number of the items that I need for creating a [dropdown](https://issues.liferay.com/browse/LEXI-664) and I was thinking about this API could be high-level. 

Check out the versions that I tried before here: https://github.com/diegonvs/clay/commits/tabs

packages/clay-tabs/src/__tests__/index.tsx
diegonvs: Yes, I guess that is a pattern here. 😅

diegonvs: I gonna fix it. Would be awesome add an eslint rule for it

wincent: AFAIK there is not en existing ESLint rule that we could use for this; the closest I could find was [padding-line-between-statements](https://eslint.org/docs/rules/padding-line-between-statements), which has a zillion configuration possibilities but none which would apply here.

We might have to make a custom rule for it. In any case, created a ticket for it [https://github.com/liferay/eslint-config-liferay/issues/65](here).

packages/clay-tabs/src/Item.tsx
wincent: How do you know it would be globally unique?

packages/clay-tabs/src/Item.tsx
wincent: I see, because you are including it with `otherProps` and passing those to the `div` element rendered by `TabPane`.

packages/clay-tabs/src/Item.tsx
matuzalemsteles: We don't need to know but interactively who is using it will want to change that, since Diego is changing we can have an `innerProps` and dev can add the id as needed, I find it odd to create it implicitly.

packages/clay-tabs/src/Item.tsx
diegonvs: I don't think that would be good to remove `otherProps`. I can just extract the value of `tabKey` and don't use. What do you think about it?

packages/clay-tabs/src/Item.tsx
diegonvs: As discussed internally, we're waiting for a component to resolve when to use Button and when use Anchor. This code review is outdated. Currently is using an `ElementTag` like implementation.

packages/clay-tabs/src/Item.tsx
diegonvs: As discussed internally, we're waiting for a component to resolve when to use Button and when use Anchor. This code review is outdated. Currently is using an `ElementTag` like implementation.

packages/clay-tabs/src/Item.tsx
diegonvs: Okay

packages/clay-tabs/src/Item.tsx
diegonvs: `dropdown` isn't just adding an icon, is adding a `.dropdown` class too.

packages/clay-tabs/src/Item.tsx
diegonvs: `dropdown` isn't just adding an icon, is adding a `.dropdown` class too.

packages/clay-tabs/src/Item.tsx
diegonvs: `dropdown` isn't just adding an icon, is adding a `.dropdown` class too.

packages/clay-tabs/src/__tests__/index.tsx
diegonvs: Nope. Removing.

packages/clay-tabs/src/index.tsx
diegonvs: ✅

diegonvs:
Hey folks, I pushed some things after your comments ;)

packages/clay-tabs/src/Item.tsx
wincent: 👍 Yes, that's exactly what I meant by "you should remove it from `otherProps`".

matuzalemsteles:

matuzalemsteles:
Closing in favor of #2317

PR #2251, opened 2019-08-08
feat(navigation): add Navigation.Vertical component

bryceosterhaus:
Wanted to send this to get external opinions. As of now, I sort of see `Vertical` as a high-level component. I also created low-level nav components but I wasn't totally sure how we want to expose those within the navigation package.

Let me know your thoughts.

TODO:
- [x] exporting nav components
- [x] tests
- [x] docs

### DIFF
--- a/packages/clay-navigation/package.json
+++ b/packages/clay-navigation/package.json
@@ -26,6 +26,8 @@
 		"react"
 	],
 	"dependencies": {
+		"@clayui/icon": "^3.0.0-milestone.2",
+		"@clayui/link": "^3.0.0-milestone.2",
 		"@clayui/shared": "^3.0.0-milestone.2",
 		"classnames": "^2.2.6"
 	},
--- a/packages/clay-navigation/src/NavLink.tsx
+++ b/packages/clay-navigation/src/NavLink.tsx
@@ -1,0 +1,76 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+import classNames from 'classnames';
+import ClayIcon from '@clayui/icon';
+import ClayLink from '@clayui/link';
+import React from 'react';
+
+interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+	/**
+	 * Flag to indicate if `active` class should be applied.
+	 */
+	active?: boolean;
+
+	/**
+	 * Flag to indicate if `collapsed` class should be applied.
+	 */
+	collapsed?: boolean;
+
+	/**
+	 * Flag to indicate if icon should be shown.
+	 */
+	showIcon?: boolean;
+
+	/**
+	 * Path to the spritemap that Icon should use when referencing symbols.
+	 */
+	spritemap?: string;
+}
+
+export const NavLink: React.FunctionComponent<IProps> = ({
+	active,
+	children,
+	className,
+	collapsed,
+	showIcon,
+	spritemap,
+	...otherProps
+}) => {
+	return (
+		<ClayLink
+			{...otherProps}
+			className={classNames('nav-link', className, {
+				active,
+				['collapse-icon']: showIcon,
+				collapsed,
+			})}
+		>
+			{children}
+
+			{showIcon && (
+				<>
+					<span className="collapse-icon-closed">
+						<ClayIcon
+							focusable="false"
+							role="presentation"
+							spritemap={spritemap}
+							symbol="caret-right"
+						/>
+					</span>
+
+					<span className="collapse-icon-open">
+						<ClayIcon
+							focusable="false"
+							role="presentation"
+							spritemap={spritemap}
+							symbol="caret-bottom"
+						/>
+					</span>
+				</>
+			)}
+		</ClayLink>
+	);
+};
--- a/packages/clay-navigation/src/Vertical.tsx
+++ b/packages/clay-navigation/src/Vertical.tsx
@@ -1,0 +1,185 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import ClayIcon from '@clayui/icon';
+import Nav from './Nav';
+import React from 'react';
+import {useTransitionHeight} from '@clayui/shared';
+
+interface IItem {
+	/**
+	 * Flag to indicate if item is active.
+	 */
+	active?: boolean;
+
+	/**
+	 * Callback for when item is clicked.
+	 */
+	onClick?: () => void;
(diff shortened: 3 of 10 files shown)


packages/clay-navigation/src/Vertical.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+function renderItems(
+	items: Array<IItem>,
+	collapsedMap: any,
+	setCollapsedMap: any,
+	spritemap?: string,
+	level = 0
+) {
+	return items.map(({active, href, items: subItems, label}, i) => {
+		const key = `${level}${i}`;
```
wincent: Shouldn't we have a separator here? Absurd example: `level` 1/`i` 11 and `level` 11/ `i` 1 produce the same key (111). May not actually have a collision in practice but in general, if you are making a compound key out of two numbers, you need a separator to isolate the components.

bryceosterhaus: Good catch. just pushed an update.

packages/clay-navigation/src/NavLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+export const NavLink: React.FunctionComponent<IProps> = ({
+	active,
+	children,
+	collapsed,
+	showIcon,
+	spritemap,
+	...otherProps
+}) => {
+	return (
+		<a
```
matuzalemsteles: Could we use `ClayLink` here?

packages/clay-navigation/src/Vertical.tsx
```diff
@@ -0,0 +1,128 @@
...
+	spritemap?: string,
+	level = 0
+) {
+	return items.map(({active, href, items: subItems, label}, i) => {
+		const key = `${level}-${i}`;
+		const collapsed = collapsedMap[key];
+
+		return (
+			<NavItem key={i}>
+				<NavLink
```
matuzalemsteles: Chances are we also need to use a `ClayButton` here for when people want to control the component instead of creating link navigation.

bryceosterhaus:
@wincent @matuzalemsteles  updated.

I added ClayLink. We might want to look into using `ClayLinkOrButton` in the future, but as of not the css doesn't seem to work well for it.

I also removed the internal map entirely and rely on internal state instead. This allows for cleaner logic and gives easier use of animation. I don't see this being an issue, since we still allow users to set the initial expanded value in `items`.

packages/clay-navigation/package.json
```diff
@@ -26,6 +26,7 @@
 		"react"
 	],
 	"dependencies": {
```
matuzalemsteles: Remember to add `@clayui/icon` here

packages/clay-navigation/src/Vertical.tsx
```diff
@@ -0,0 +1,143 @@
...
+	const [active, setActive] = React.useState(false);
+
+	return (
+		<nav
+			className={classNames('menubar menubar-transparent', {
+				['menubar-vertical-expand-lg']: large,
+				['menubar-vertical-expand-md']: !large,
+			})}
+		>
+			<button
```
matuzalemsteles: Just to know, could we use a `ClayButtonWithIcon` here?

matuzalemsteles:
> I also created low-level nav components but I wasn't totally sure how we want to expose those within the navigation package.

I think it makes sense to expose as a low-level component too since people can compose `NavLink` with other elements or allow people to customize.

Can we expose packages with the same idea we did for `ClayForm` to shorten namespaces?

packages/clay-navigation/src/Vertical.tsx
bryceosterhaus: i forgot about that one. good catch

bryceosterhaus: Actually, I just realized we need to use it without `ClayButtonWithIcon` since this component requires `{activeLabel}` to be in the button and next to the icon. `ClayButtonWithIcon` is only for just an icon.

bryceosterhaus:
@matuzalemsteles  updated

packages/clay-navigation/src/index.tsx
```diff
@@ -6,13 +6,25 @@
...
+	Vertical: typeof Vertical;
 } = ({children}) => {
 	return children;
 };
 
 ClayNavigation.Breadcrumb = Breadcrumb;
+ClayNavigation.Nav = Nav;
+ClayNavigation.NavItem = NavItem;
+ClayNavigation.NavLink = NavLink;
+ClayNavigation.Vertical = Vertical;
```
matuzalemsteles: I was thinking of making it available the same way we did in `ClayForm`. It seems a little strange... what do you think?

```javascript
ClayVertical.Nav = Nav;
ClayVertical.NavItem = NavItem;
ClayVertical.NavLink = NavLink;

export {ClayVertical};
```

Now I've been wondering if all the components should be in the same namespace, maybe you get the impression that it's all part of `Navigation`, `Form` and you can compose them all anywhere, it looks like we can sell more...

bryceosterhaus: I didn't want to namespace it off of `ClayVertical` because `nav` is used outside of the context of `Vertical` in other types of navs.

We could possible change it from `@clayui/navigation` to `@clayui/nav` and the way we can do something like

```js
ClayNav = Nav;
ClayNav.Item = NavItem;
ClayNav.Link = NavLink;


// option #1
ClayNav.Vertical = Vertical;
ClayNav.BreadCumb = BreadCumb;

// option #2
export {
       ClayNavWithBreadCrumb,
       ClayNavWithVertical,
}

export default ClayNav
```

matuzalemsteles: Hmm, makes sense. I think we can keep this within `Navigation` even though we may also migrate `Navigation Bar` to it.

bryceosterhaus: Right on, so ship components like so? This means our root offering for Navigation is `'nav'`. 

```js
ClayNavigation = Nav;
ClayNavigation.Item = NavItem;
ClayNavigation.Link = NavLink;
ClayNavigation.Vertical = Vertical;
ClayNavigation.BreadCumb = BreadCumb;

export default ClayNavigation
```

This makes sense to me I think.

matuzalemsteles: Hmm, maybe so. Not sure about this because the breadcrumb base is not a nav but I think we can go with that.

bryceosterhaus:
@matuzalemsteles updated and added docs and tests.

The export structure I went with is

```js
import Nav from './Nav';
import {ClayBreadcrumbNav} from './Breadcrumb';
import {ClayVerticalNav} from './Vertical';

export {ClayBreadcrumbNav, ClayVerticalNav};

export default Nav;
```

I went with named exports because if I namespaces `Vertical` on `Nav`, it would create a circular dependency between the two.

Issue #2261, opened 2019-08-12
Can we add an option to pass in a custom fetch function to useRecource?


brunobasto:
In portal we have a custom `fetch` function that takes care of authentication and security configurations. If we use `useResource`, none of that will apply and there's no way to change it: https://github.com/liferay/clay/blob/master/packages/clay-data-provider/src/useResource.ts#L153

@brunofarache @matuzalemsteles @jbalsas

matuzalemsteles:
hey @brunobasto,you can pass to `link` an function but will have to deal with url construction and params if needed.

Example:
```javascript
const {} = useResource({
    link: () => Liferay.Util.fetch(...),
});
```

If you follow this path you will lose the benefit of using prop `variables` it will have no effect here.

Another option would be to do the same as `Liferay.Util.fetch`, for now it just adds `credentials: 'include'`, you can pass this through the `fetchOptions` API.

Example:
```javascript
const {} = useResource({
    link: '',
    fetchOptions: {
        credentials: 'include',
    }
});
```

wincent:
> Another option would be to do the same as `Liferay.Util.fetch`, for now it just adds `credentials: 'include'`, you can pass this through the `fetchOptions` API.

The problem with that is the "for now": there is *already* [one other thing it does](https://github.com/liferay/liferay-portal/blob/27800770b780195b90da7e420bd1ee6e67e3f095/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/fetch.es.js#L29) (setting a `x-csrf-token` header), and no guarantee that it won't end up doing more things in the future.

I think we should explore ways to make `Liferay.Util.fetch` usable inside `useResource`, which might be a `fetch` parameter alongside `fetchOptions`, or some other solution.

matuzalemsteles:
Oh thanks @wincent, don't you think passing `link` as a function is not enough? maybe we can pass the `variables` to the link function to maintain the benefit and also improve the caching system.

```javascript
const {} = useResource({
    link: (variables) => Liferay.Util.fetch(baseUrl + variables),
    variables: {...},
});
```

wincent:
Yeah, I think that would work.

For anybody else reading this, this is the code we're talking about:

https://github.com/liferay/clay/blob/91e59e98264511b6e3711b33049318b21267fb68/packages/clay-data-provider/src/useResource.ts#L153-L173

Now that I look at it though, I see the typing for the `link` prop is a bit weak:

[...]

PR #2267, opened 2019-08-13
feat(@clayui/data-provider): pass formatted variables when link is a function

matuzalemsteles:


packages/clay-data-provider/src/__tests__/index.tsx
```diff
@@ -267,6 +281,25 @@ describe('ClayDataProvider', () => {
 		);
 	});
 
+	it('call clay.data with search params and link being a function', async () => {
+		fetchMock.mockResponse(JSON.stringify({title: 'Foo'}));
+
+		render(
+			<DataProvider
+				link={variables => fetch(`https://clay.data/?${variables}`)}
```
bryceosterhaus: How come we pass in the variables via a prop, only to then consume it via link? Seems like the user should just form the link url themselves. I saw the comment below in `warning()` but still wasn't quite clear on it.

```js
link={variables => fetch(`https://clay.data/?name=bar`)}
```

matuzalemsteles: This is just to add support for using variables for when the `link` prop is a function, if you set the URL only in your fetch, you lose the benefits of `variables` functionality, for example: new fetch when `variables` change, improve cache match... This makes some improvements to #2261 use cases.

You can even set the URL in fetch method and use variables just to have that benefit and in your fetch URL, but you'd better declare it in one place and encourage people to use the API.

bryceosterhaus: ah i see. Just wanted to make sure I was understanding it correctly.

packages/clay-data-provider/src/types.ts
```diff
@@ -36,14 +36,16 @@ export enum FetchPolicy {
...
 
-export type TLink = Function | string;
-
 export interface IVariables {
 	[propName: string]: any;
 }
 
 export type TVariables = IVariables | null;
 
+export type TLinkFunction = (variables: string) => Promise<any>;
```
wincent: Thanks for improving the type here!

PR #2272, opened 2019-08-14
fix(clay-multi-select): HTML escape to prevent XSS attack for multiselect

matuzalemsteles:


packages/clay-multi-select/src/htmlEscape.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable no-control-regex */
+const AMP_RE_ = /&/g;
+const LT_RE_ = /</g;
+const GT_RE_ = />/g;
+const QUOT_RE_ = /"/g;
+const SINGLE_QUOTE_RE_ = /'/g;
+const NULL_RE_ = /\x00/g;
+const ALL_RE_ = /[\x00&<>"']/;
+
+// All credits from this code to https://github.com/google/closure-library/blob/master/closure/goog/string/internal.js#L173
```
wincent: This is probably insufficient from a legal perspective. That code is under the Apache License, which requires you to retain the copyright notices that came with it.

packages/clay-multi-select/src/htmlEscape.js
```diff
@@ -0,0 +1,37 @@
...
+	}
+	if (str.indexOf('\'') != -1) {
+		str = str.replace(SINGLE_QUOTE_RE_, '&#39;');
+	}
+	if (str.indexOf('\x00') != -1) {
+		str = str.replace(NULL_RE_, '&#0;');
+	}
+
+	return str;
+};
```
wincent: When I've done this in the past I did even less than this (ie. [just this](https://github.com/wincent/js/blob/master/packages/escape-html/src/index.ts)). I based that on [the top Stack Overflow answer](https://stackoverflow.com/a/7382028/2103996) on the subject; an old one but probably still valid? Having said that, doing a bit of extra escaping does no harm.

I'd be interested in testing whether those `indexOf()` probes make this any faster. I wouldn't expect them too because they involve a full scan over the string each time. Here's [a quick jsperf test](https://jsperf.com/escaping-blah) to compare them:

escaping-blah_·_jsPerf

jbalsas: could we maybe take this version then and avoid the "legal" issues? :D

wincent: Feel free.

matuzalemsteles: 👍

packages/clay-multi-select/src/htmlEscape.js
```diff
@@ -0,0 +1,13 @@
+const ampersandRegExp = /&/g;
+const lessThanRegExp = /</g;
+const greaterThanRegExp = />/g;
+const doubleQuoteRegExp = /"/g;
+const singleQuoteRegExp = /'/g;
+
+export const htmlEscape = string =>
```
matuzalemsteles: hey @wincent I didn't make many changes so in the end I keep the same thing.

jbalsas: Hey @matuzalemsteles, we're out for the day, so I think it should be okay to merge and release. We can go over this tomorrow if there are changes needed.

wincent: lg2m in any case. 👍

packages/clay-multi-select/src/ClayMultiSelect.js
```diff
@@ -75,7 +76,10 @@ class ClayMultiSelect extends ClayComponent {
 	_defaultAutocompleteItemSelected(event) {
 		const {label, value} = event.data.item;
 
-		const item = this._createItemObject(label, value);
+		const item = this._createItemObject(
+			htmlEscape(label),
+			htmlEscape(value)
```
samuelkong: @matuzalemsteles @jbalsas 
I'm not very familiar with Clay, but this doesn't seem like the right fix to me. 

1. You're basically escaping too early. While there might not be any problems right now, this is going to create problems later when `item` gets used slightly differently from how it's used now. I've seen this problem many times now. Instead of escaping here, you should delay the escape until right before the value is outputted to the DOM.

2. You cannot use the same escape method for label and value. From what I can tell, label is used between `<span>` and `</span>` where you need something like `HtmlUtil.escape()`. For value, this is used in a tag's attribute (e.g., `<input value="value">` where you need something like `HtmlUtil.escapeAttribute()`

packages/clay-multi-select/src/htmlEscape.js
```diff
@@ -0,0 +1,13 @@
...
+	string
+		.replace(ampersandRegExp, '&amp;')
+		.replace(lessThanRegExp, '&lt;')
+		.replace(greaterThanRegExp, '&gt;')
+		.replace(doubleQuoteRegExp, '&quot;')
+		.replace(singleQuoteRegExp, '&#39;');
```
samuelkong: @matuzalemsteles @jbalsas 

For escaping data between `<tag>` and `</tag>`, you need to also escape the forward slash `/`.

wincent: @samuelkong: How would a `/` be problematic in that position?

samuelkong: @wincent
This gets into very obscure and mostly unused parts of the HTML specs. 
`<div/data/` is actually interpreted as `<div>data</div>`. So an unescaped forward slash can be used to close the `<div>` tag.

Reference: https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-1---html-escape-before-inserting-untrusted-data-into-html-element-content

wincent: Thanks for the explanation @samuelkong, seeing as the cheatsheet doesn't really explain why (it just says, vaguely, "Forward slash is included as it helps end an HTML entity", which is pretty unsatisfying).

Although, seeing as we're escaping `<`, we'd never actually insert `<div/data/` data into the DOM but rather `&lt;/data/`. Would that still be a problem? At first glance, it doesn't seem it would.

samuelkong: @wincent
Off the top of my head, I can't think of anything. But we still have to implement the escape for forward slash. If not, you can be the one that goes to marketing and tell them to remove everything about following OWASP guidelines from our marketing material.

wincent: Oh, absolutely, there's no harm "over"-escaping; I'm just trying to *understand* as opposed to blindly cargo-culting. My googling didn't turn up anything that would explain it.

jbalsas: Hey @samuelkong, thanks for chiming in!

I share @wincent's desire for understanding what are the requirements we have to meet for security fixes so we can make sure we don't miss them next time. Do you have some specific links to where we say we follow OWASP guidelines and to the point in those guidelines where what you mention is specified for this case? That would be really formative for us!

I've also filed an internal ticket with the security team (https://issues.liferay.com/browse/PTR-1132) to get some guidance and clarification from them to make sure we don't miss anything.

Thanks!

samuelkong: @jbalsas 
I don't have anything off the top of my head, but a quick Google searched returned https://portal.liferay.dev/docs/7-0/deploy/-/knowledge_base/d/liferay-portal-security-overview. I couldn't find anything more recent. However, I know I've seen some recent RFP from sales where we says we following OWASP guidelines. As in this link, I think we usually only say we follow OWASP top ten and OWASP top ten includes protecting against injection vulnerability.

I general, we don't specifically mention that we follow OWASP XSS escaping rules to customers unless the customer ask. I also tell our security testing vendor that our escape methods complies with OWASP XSS Guidelines (specifically the OWASP XSS cheat sheet).

jbalsas: Thanks @samuelkong!

I had a brief conversation with @topolik and @csierra which is more or less reflected in https://issues.liferay.com/browse/PTR-1132.

The owasp recommendation can be found at https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#output-encoding-rules-summary. Unfortunately, it does not explain why either, but at least is a valid reference.

packages/clay-multi-select/src/ClayMultiSelect.js
jbalsas: About 

1. I think this is the last place where we can actually scape before this gets passed down to the template engine to be committed to the DOM, so I don't think there's other place we can do it.
2. That's correct, we should have a different scaping for `value` which goes to an attribute.

matuzalemsteles: > You cannot use the same escape method for label and value. From what I can tell, label is used between <span> and </span> where you need something like HtmlUtil.escape(). For value, this is used in a tag's attribute (e.g., <input value="value"> where you need something like HtmlUtil.escapeAttribute()

This is a particular case as `label` and `value` here are the same things at the end and we may not even need to add an escape here because these values ​​are from the autocomplete that comes from the backend data.

jbalsas: We commit `label` to the DOM inside html and `value` to the DOM inside an HTML attribute and SOY is failing to escape those for us. I'm not sure why that happens, but since we're migrating to the React components, this is meant to be a simple temporary mitigation.

topolik: Guys, please don't forget this is going to go into 7.2 patch and will stay with us for long, it's not that temporary as it seems. If you aim to make it more stable, please do it now. Thanks! :)

jbalsas: We're planning to update our component usage in 7.2, so it is temporary because we're looking to use more robust components there too.

topolik: I see, nice then. Thanks.

PR #2273, opened 2019-08-14
fix(clayui.com): stop refresh loop when accessing documentation

matuzalemsteles:


matuzalemsteles:
I am merging this to push a new version of the site.

PR #2276, opened 2019-08-14
feat(@clayui/modal): add ClayModalProvider

matuzalemsteles:
I'm just submitting this as a draft but I'm still not sure if this would be the right way for Modal or his high level. Leave your opinion about this API.

- [x] Tests
- [x] Docs
- [x] Define API


packages/clay-modal/stories/index.tsx
```diff
@@ -46,6 +46,30 @@ const ModalWithState: React.FunctionComponent<IProps> = ({
...
 };
 
+const MyApp: React.FunctionComponent<any> = ({}) => {
+	const [state, disptach] = useContext(Context);
+
+	return (
+		<ClayButton
+			displayType="primary"
+			onClick={() =>
+				disptach &&
```
matuzalemsteles: I had a problem with types so I'll have to expose the `TProvider` interface for people can add type to `useContext`.

packages/clay-modal/src/Provider.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+interface IProps {
+	spritemap?: string;
+}
+
+type TState = {
+	body: React.ReactElement | React.ReactText;
+	footerFirst?: React.ReactElement;
+	footerLast?: React.ReactElement;
```
bryceosterhaus: Instead of first and last, would it work to add an array for footer? It'd give a bit more flexibility in the future.

I saw we have `first` `middle` and `last` in our Footer component. So maybe an array would make sense, primarily thinking if in the future we add more areas to footer, we then have `footerNextToLast`. This may be a pre-mature optimization, but think it'd make it at least a bit more flexible.

packages/clay-modal/stories/index.tsx
```diff
@@ -46,6 +46,30 @@ const ModalWithState: React.FunctionComponent<IProps> = ({
 	);
 };
 
+const MyApp: React.FunctionComponent<any> = () => {
+	const [, disptach] = useContext(Context);
```
bryceosterhaus: Spelling here. should be `dispatch`

sidenote: I didn't realize you could do `[, dispatch]`. very cool.

packages/clay-modal/src/Provider.tsx
```diff
@@ -0,0 +1,91 @@
...
+							<ClayModal.Body url={url}>{body}</ClayModal.Body>
+							<ClayModal.Footer
+								first={footerFirst}
+								last={footerLast}
+							/>
+						</>
+					)}
+				</ClayModal>
+			)}
+			<Context.Provider value={[state, dispatch]}>
```
bryceosterhaus: Since we are using this Provider for modal now, can we use react children without a function and instead pass `onClose` via context? This would remove the need for renderprops.

packages/clay-modal/src/Provider.tsx
```diff
@@ -0,0 +1,91 @@
...
+	url?: string;
+};
+
+type TAction = {type: Action.Open; payload: TState} | {type: Action.Close};
+
+type TProvider = [TState, React.Dispatch<TAction>];
+
+const initialState = {
+	body: <div />,
+	visible: false,
```
bryceosterhaus: Not sure if I am fully understanding this right. But would this not allow for multiple modals? We might be able to leverage a Map for this and have unique ids for each modal open.

packages/clay-modal/src/Provider.tsx
```diff
@@ -0,0 +1,91 @@
...
+	return (
+		<>
+			{visible && (
+				<ClayModal
+					onClose={() => dispatch({type: Action.Close})}
+					size={size}
+					spritemap={spritemap}
+					status={status}
+				>
+					{onClose => (
```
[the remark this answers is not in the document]
bryceosterhaus: Woops, just seeing this now. I noted below about this idea as well. I think its a great idea to use onClose via hook or context instead of renderProps

packages/clay-modal/stories/index.tsx
bryceosterhaus: hmm, thats strange. What was the issue in particular?

packages/clay-modal/src/Provider.tsx
matuzalemsteles: It won't create multiple modals, we just want to have one Modal open, so calling a dispatch will change the existing context information. Because you think you would create multiple modals?

packages/clay-modal/src/Provider.tsx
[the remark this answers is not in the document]
matuzalemsteles: Oh great, We are aligned. 🙂

packages/clay-modal/stories/index.tsx
matuzalemsteles: Oh that's weird, now that I've done some testing seems to work fine, just changing it in `createContext<TProvider>([initialState, () => {}])`. This was breaking before, I don't know why ...

packages/clay-modal/src/Provider.tsx
matuzalemsteles: Hmm, I think that might make sense. But you claim to be an Array on the Provider side and pass it on to `ClayModal.Footer`? Considering the first element of the array is `first`, second `middle` and third `last`.

packages/clay-modal/src/Provider.tsx
bryceosterhaus: I am primarily thinking of a context where a developer might need nested models, they open one and then a button opens another one on top of that. I can imagine devs using it this way, even if its not necessarily intended, so I think we might want to think of having a global Map of open Models. That way we can close a specific modal, instead of closing everything at once.

packages/clay-modal/src/Provider.tsx
bryceosterhaus: Yeah, I would think `footer?: Array<React.ReactElement>` where the first index is the left most and the end is the right most element. Not super important, was just a thought of mine when thinking of "how does this sort of API scale."

packages/clay-modal/stories/index.tsx
bryceosterhaus: I wonder if it relates to the tsconfig issues a few people have run into. Or maybe vscode just needed to restart the ts server.

matuzalemsteles: > I wonder if it relates to the tsconfig issues a few people have run into. Or maybe vscode just needed to restart the ts server.

I think this was just VSCode, I don't think it's related to the types we were generating wrong for production. Why am I messing with the same package.

packages/clay-modal/src/Provider.tsx
matuzalemsteles: Oh yeah, sounds good.

packages/clay-modal/src/Provider.tsx
matuzalemsteles: Actually I wanted to discourage the use of nested modals is something we should not encourage, this is not correct.

matuzalemsteles:
hey @bryceosterhaus update.

packages/clay-modal/src/Provider.tsx
bryceosterhaus: I think we can still discourage it, but not necessarily outlaw it. Perhaps a normal use case would be a modal open and then an additional "Confirm" type modal/dialog on top.

It may not be needed quite yet, but perhaps we can add this down the line if its too much work to add in this PR

matuzalemsteles: I say this because I had a conversation implementing this PR about it https://github.com/liferay/clay/pull/1842#discussion_r278464379 and I also think calling modal within another modal is a wrong way to do it, Lexicon also seems not to recommend https://liferay.design/lexicon/core-components/modals.

packages/clay-modal/stories/index.tsx
```diff
@@ -47,17 +47,20 @@ const ModalWithState: React.FunctionComponent<IProps> = ({
...
 		<ClayButton
 			displayType="primary"
 			onClick={() =>
-				disptach &&
-				disptach({
+				dispatch({
 					payload: {
 						body: <h1>{'Hello world!'}</h1>,
-						footerLast: <ClayButton>{'Primary'}</ClayButton>,
+						footer: [
```
bryceosterhaus: Are you able to do the same thing as above instead of explicit `undefined`? not totally necessary, but since I just learned of this, I can't help but see it lol

```jsx
[,, <ClayButton>{'Primary'}</ClayButton>]
```

packages/clay-modal/src/Provider.tsx
jbalsas: As far as I know from every conversation I've had with Lexicon over the past 2 years, modals over modals are an anti-pattern to be avoided at all costs 😂 

Ideally, we should find some way to navigate within a modal back and forth rather than stacking them. This is mostly a UX issue that we can't really tackle at once because it means changing a lot of workflows and components, so we've always had to just take it as it is.

packages/clay-modal/stories/index.tsx
matuzalemsteles: Yeah it is possible.

matuzalemsteles:
hey @bryceosterhaus I added `useModal`, tests and docs.

matuzalemsteles:
Oh np, thanks @bryceosterhaus

Issue #2246, opened 2019-08-06
Dropdown created in a modal shows up behind the modal


oliv-yu:
A Clay Dropdown situated in a Clay Modal will show up behind the modal. It looks like it's not nested. Our temporary fix for our local case was to change the z-index, but let us know if there's a more targeted solution.

Sandbox for reference: https://codesandbox.io/s/nostalgic-proskuriakova-6cyv4

matuzalemsteles:
Hmm, I thought we already fixed this using Portals in DropDown, but it looks like really `z-index` is getting in the way. @marcoscv-work have a problem if we change DropDown's `z-index` to **1060**?

Thanks @oliv-yu for tracking this.

matuzalemsteles:
I had forgotten that we revert the use of `z-index` because it may look above Modal when the dropdown button is out of Modal. Maybe we have to think of some strategy for this in the components in React.

Maybe this can work:
- If the dropdown is inside Modal, we add `.dropdown-menu` as sibling of `modal-dialog` to prevent `modal-dialog` overflow from hiding part of the dropdown.

@bryceosterhaus do you have any idea about this?

bryceosterhaus:
Hmm, this is one of the hard parts about isolating dropdowns into a portal element on the html body. Ideally a dev wouldn't have to think about "is this in a modal" and then add an extra class.

I could see us going a few different routes:
1. Move dropdown html to be inline with where the trigger is, make portals only opt-in. This would ensure the proper html priority with z-index
2. Give dropdown highest z-index priority and make sure to close any dropdowns before opening a modal.
3. Make developers responsible for adding the appropriate class for when a dropdown is in a modal. One caveat is that we would have to make sure we account for nested modals and thus nested dropdowns in modals.

TBH, not really sure which route I'd prefer, I see issues with all of them. I feel like I am probably missing something here...

jbalsas:
Hey @bryceosterhaus, @matuzalemsteles, I haven't 100% checked this case, but based on our past experience, I think one possible approach would be to have multiple portals.

Modals should have their own portal at their body level similar to the one the root `body` has. With that, it should be possible to honour z-index depth while avoiding the inlining (`overflow: hidden` or unnecessary scrolls) issues. Would it be possible to pick the portal closest to your parent or your root automatically so developers wouldn't have to think about that either?

[...]

PR #2277, opened 2019-08-15
fix(portal): automatically nest portals if they are inside one another

bryceosterhaus:
Added two new APIs to our portals. One is a target for nested portals, and the other is to render the portal to a specific element. By default portals now nest in one another as well.

this fixes https://github.com/liferay/clay/issues/2246

### DIFF
--- a/packages/clay-shared/src/Portal.tsx
+++ b/packages/clay-shared/src/Portal.tsx
@@ -4,18 +4,44 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import React, {useEffect, useRef} from 'react';
+import React, {createContext, useContext, useEffect, useRef} from 'react';
 import {createPortal} from 'react-dom';
 
+const ClayPortalContext = createContext<React.RefObject<Element | null> | null>(
+	null
+);
+
+ClayPortalContext.displayName = 'ClayPortalContext';
+
 export const ClayPortal: React.FunctionComponent<
-	React.HTMLAttributes<HTMLDivElement>
-> = ({children}) => {
+	React.HTMLAttributes<HTMLDivElement> & {
+		/**
+		 * Ref of element to render portal into.
+		 */
+		containerRef?: React.RefObject<Element>;
+
+		/**
+		 * Ref of element to render nested portals into.
+		 */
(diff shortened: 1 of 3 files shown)


packages/clay-shared/src/Portal.tsx
```diff
@@ -4,18 +4,40 @@
...
+		containerRef?: React.RefObject<Element>;
+
+		/**
+		 * Ref of element to render nested portals into.
+		 */
+		subPortalRef?: React.RefObject<Element>;
+	}
+> = ({children, containerRef, subPortalRef}) => {
+	const parentPortalRef = useContext(ClayPortalContext);
+	const portalRef = useRef(document.createElement('div'));
```
matuzalemsteles: You can check if there is `document` here, I was having problems with it because it was being rendered on `next.clayui.com` and was breaking the site build (log https://app.netlify.com/sites/next-clayui/deploys/5d5595f8912d82502b05f511), because at build time there is no `document`.

```javascript
typeof document !== 'undefined' && document.createElement('div')
```

tsconfig.json
```diff
@@ -11,7 +11,7 @@
 		"allowJs": true,
 		"baseUrl": ".",
 		"paths": {
-			"@clayui/*": ["node_modules/@clayui/*/src"],
+			"@clayui/*": ["packages/clay-*/src"],
```
matuzalemsteles: Have you ever tested if this works well? by running `yarn build` on packages and checking if the `lib` folder is being generated correctly with types.

bryceosterhaus: oh my bad, I didn't mean to include this. This was the previous code before you updated it because I was running into issues.

bryceosterhaus:
@matuzalemsteles yeah I was confused how coverage dropped. Maybe because this is the first test inside of `shared` and it might have not been looking as those files previously.

matuzalemsteles:
> @matuzalemsteles yeah I was confused how coverage dropped. Maybe because this is the first test inside of shared and it might have not been looking as those files previously.

Oh yeah, you're right about that. We can work later on adding tests to the other files from `shared`.

bryceosterhaus:
@matuzalemsteles also, I just updated the PR. Let me know if there is anything else you notice.

Issue #2280, opened 2019-08-15
Forward References of ClayLink


diegonvs:
Some Routing libraries like `next/link` doesn't allow pass `ClayLink` due to a missing reference for it.

/cc @ygorcosta @matuzalemsteles

bryceosterhaus:
Duplicate of https://github.com/liferay/clay/issues/2250 ? also PR just sent upstream for this, let me know if I am missing something.

matuzalemsteles:
Fixed #2281

PR #2281, opened 2019-08-15
feat(link): add ability to supply component via context

bryceosterhaus:


packages/clay-link/src/index.tsx
```diff
@@ -37,19 +38,25 @@ const ClayLink: React.FunctionComponent<IProps> = ({
...
+				[`link-outline-${displayType}`]: displayType && outline,
+			})}
+			{...otherProps}
+		>
+			{children}
+		</TagOrComponent>
+	);
+};
+
+export {ClayLinkContext};
```
matuzalemsteles: hey @bryceosterhaus yet i'm not sure about that, i wish we could test this out, so i suggest we rename it to unstable (being experimental) at the moment and as soon as we feel and collect feedback that this is the right direction. We can make it stable.

```suggestion
export {ClayLinkContext as unstable_ClayLinkContext};
```

I think it's safer that way before it makes it stable along with the official release.

bryceosterhaus: You can test it out in the storybook and also see it working in the tests. Or are you thinking a different way to test it?

bryceosterhaus: Also, it looks like it was also brought up here by Diego yesterday, https://github.com/liferay/clay/issues/2280. Not sure if this overlaps. Ill go ahead and add a commit for  forwardRef as well

matuzalemsteles: > You can test it out in the storybook and also see it working in the tests. Or are you thinking a different way to test it?

Hey @bryceosterhaus wouldn't really be these tests, what I mean by that is we need to test with the teams and wait a while to know if this approach is the best way to get on with it.

Another concern that I would like to test in a real application is that we might want `ClayLink` components without being attached to a `react-router` for example, so we need to test how this will get inside those applications, add it as `unstable` gives us the ability to modify this at any time and test with teams even when we make the official release.

> Also, it looks like it was also brought up here by Diego yesterday, #2280. Not sure if this overlaps. Ill go ahead and add a commit for forwardRef as well

Yeah this is another use case we just found in DXP Cloud (In the new admin) using `next/router`, you can pass as children (`ClayLink`) on `next/router` for example that it will replace `ClayLink` markup but it adds a ref to the element.

bryceosterhaus:
@matuzalemsteles updated to be exported as `unstable`

bryceosterhaus:
> we might want ClayLink components without being attached to a react-router for example, so we need to test how this will get inside those applications.

This shouldn't be a problem with the current implementation and ClayLink isn't actually tied to any specific router. This implementation just allows us to pass a custom component instead of defaulting to a static anchor tag.

packages/clay-link/src/__tests__/index.tsx
```diff
@@ -5,7 +5,7 @@
  */
 
 import * as TestRenderer from 'react-test-renderer';
-import ClayLink from '..';
+import ClayLink, {unstable_ClayLinkContext} from '..';
```
matuzalemsteles: I think this was lost when sending review... just a tip for when we start doing this more often and have a lot of testing, we can add an alias here for `unstable_ClayLinkContext as ClayLinkContext` and when this is stable we can just change the alias.

matuzalemsteles:
> This shouldn't be a problem with the current implementation and ClayLink isn't actually tied to any specific router. This implementation just allows us to pass a custom component instead of defaulting to a static anchor tag.

Oh sorry, it wasn't very explanatory but what I meant is: In one application I can use `ClayLink` just to be an anchor and in the same hierarchy I can have `ClayLink` for a route system, so where the person putting the `context` can influence other `ClayLinks` that I do not want. This depends on the project hierarchy, so I would like to see this being tested in a real case and also if it makes sense to cover these use cases using `context` and if that is the best way.

bryceosterhaus:
Sounds good. Ill go ahead and merge with "unstable" then

PR #2291, opened 2019-08-21
feat(clay-css): Create slider component

pat270:
@matuzalemsteles @bryceosterhaus there are some quirks you need to account for with JS.
1) The progress bar won't be centered on the slider head if you take the value + steps and divide them up equally. I think I found the right ratio in my example, but let me know if you find a better way.
2) The progress bar width should be re-calculated on screen resize
3) IE11 doesn't support on `input` event
4) When disabling an input you need to add attribute `disabled` on `.form-control-range` and the class `disabled` on `.clay-range` and the `label`.


marcoscv-work:

PR #2301, opened 2019-08-21
fix(@clayui/pagination): imports ClayPagination from its origin component and avoids circular dependency

### LINKED ISSUE #2300: ClayPaginationWithBar with dependency circular
`ClayPaginationWithBar` is importing `ClayPagination` from the `index` file and this is breaking when it is being used inside the Portal, it seems that the Bundler has problems with circular dependencies.

The `default` import is something like `ClayPagination.default.default.ClayPagination` and in the transpiled code only `ClayPagination.default` this breaks React because it is not a valid component.

matuzalemsteles:
Fixes #2300

### DIFF
--- a/packages/clay-pagination/src/PaginationWithBar.tsx
+++ b/packages/clay-pagination/src/PaginationWithBar.tsx
@@ -7,7 +7,7 @@
 import classNames from 'classnames';
 import ClayButton from '@clayui/button';
 import ClayIcon from '@clayui/icon';
-import ClayPagination from './index';
+import ClayPagination from './Pagination';
 import React, {useState} from 'react';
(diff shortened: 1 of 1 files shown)


diegonvs:
We need to define a guideline for importing files when creating a high-level.
There are two different strategies:
```jsx
import Cell from '../src/Cell';

<Cell ...
```

and

```jsx
import ClayTable from '../src/index';

<ClayTable.Cell ...
```
Which approach is more adequated?

matuzalemsteles:
hey @diegonvs did you can move this question to an issue to discuss?

diegonvs:
Sure.

PR #2310, opened 2019-08-21
feat(@clayui/table): Adds `truncate` property to Table Cell

### LINKED ISSUE #2279: Adds a `truncate` property for ClayTable.Cell
The most part of the use cases of `ClayTable.Cell` component is using a `string` as a child. However, We currently can't turn this string truncated.
See the following [paragraph](https://github.com/liferay/clay/blob/master/packages/clay-table/src/Cell.tsx#L78) element that is being applied when passing a string as a child.

diegonvs:
Fixes #2279

### DIFF
--- a/packages/clay-modal/src/Modal.tsx
+++ b/packages/clay-modal/src/Modal.tsx
@@ -50,6 +50,9 @@ const ClayModal: FunctionComponent<IProps> & {
 
 	useEffect(() => observer.dispatch(ObserverType.Open), []);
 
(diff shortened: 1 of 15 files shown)


diegonvs:
> Can you update Storybook stories to use this correctly?

Added :)

matuzalemsteles:
hey @diegonvs you can take a look, I think there came some commits that shouldn't be here.

diegonvs:
Resent at #2360

PR #2312, opened 2019-08-22
feat(@clayui/slider): add ClaySlider component

### LINKED ISSUE #2157: Create Slider component
See https://liferay.design/lexicon/core-components/forms/slider/

matuzalemsteles:
Fixes #2157


packages/clay-slider/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * Set the current value of the slider.
+	 */
+	value: number;
+}
+
+const THUMB_WIDTH = 24;
+
+const calcProgressWidth = (
```
matuzalemsteles: hey @pat270 i ended up keeping your calculation, i tried to get some but without much success i am keeping it for now, if anyone has something better please leave your thoughts.

packages/clay-slider/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+	 */
+	tooltipPosition?: 'top' | 'bottom';
+
+	/**
+	 * Set the current value of the slider.
+	 */
+	value: number;
+}
+
+const THUMB_WIDTH = 24;
```
pat270: @matuzalemsteles This should be configurable. I was hoping to be able to grab the thumb width via js, but because it's in the shadow dom I wasn't able to quickly. Also the thumb selector is different for every browser which adds to the complexity.

Looking back now, I can probably add a thumb overlay (similar to the tooltip) to make this easier and forget about trying to style the browser default thumb. I can send an update to the slider markup. It should make it more customizable.

matuzalemsteles: Oh thanks @pat270, for the info, I think adding a thumb overlay is a better solution. Even if we make it configurable people will probably have to have a list of widths that the thumb can be, I think it's not good to add this complexity to the component.

jbalsas: hey guys, just keep in mind that the less we use of the native input the more we need to build back in. Native thumb will have builtin accessibility, keyboard navigation and so on... the more we can keep the better

packages/clay-slider/src/__tests__/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+describe('ClaySlider', () => {
+	afterEach(cleanup);
+
+	it('renders', () => {
+		const {container} = render(
+			<ClaySlider onValueChange={() => {}} value={10} />
+		);
+
+		expect(container).toMatchSnapshot();
```
bryceosterhaus: I mentioned this in another PR as well. But I think it would be best to be more explicit in our tests rather than just snapshots. We mention this a bit on the frontend guidelines as well, https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/testing/ui_components.md

packages/clay-slider/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+	value: number,
+	min: number,
+	max: number,
+	step: number
+) => {
+	const rangeWidth = element.clientWidth;
+	const currentStep = (value - min) / step;
+	const totalSteps = (max - min) / step;
+	const progressWidth = (currentStep / totalSteps) * 100;
+	const ratio =
```
bryceosterhaus: Make sure we have new lines separating non-sortable items. For example, since line `progressWidth ` is dependent on `totalSteps`, they should have a new line between.

packages/clay-slider/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+	tooltipPosition = 'top',
+	value,
+	...otherProps
+}) => {
+	const [offsetWidth, setOffsetWidth] = useState<number>(0);
+	const sliderRef = useRef<HTMLInputElement | null>(null);
+
+	const handleEvents = (event: React.ChangeEvent<HTMLInputElement>) => {
+		const value = Number(event.target.value);
+		onValueChange(value);
```
bryceosterhaus: new line above onValueChange

packages/clay-slider/src/index.tsx
pat270: @jbalsas it will still be using the native input and should work the same. The only thing we need to calculate correctly is the progress bar width.

packages/clay-slider/src/index.tsx
matuzalemsteles: Makes sense, thanks!

packages/clay-slider/src/__tests__/index.tsx
matuzalemsteles: Sure, I will decrease the use of snapshots here.

packages/clay-slider/src/index.tsx
matuzalemsteles: hey @pat270 with your changes will i need to change anything here? Let me know if I need to wait for your changes to come in first.

pat270: @matuzalemsteles yes, I'm building on top of this pr. The only thing that needs to change is `THUMB_WIDTH` should get the width of `.clay-range-thumb` and the tooltip needs to be nested inside `.clay-range-thumb`. I'll send my changes to you when I finish.

matuzalemsteles:
@bryceosterhaus updated, let me know if you see anything else.

packages/clay-slider/src/index.tsx
matuzalemsteles: @pat270 ok, I just made the small updates.

packages/clay-slider/src/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+	min: number,
+	max: number,
+	step: number
+) => {
+	const currentStep = (value - min) / step;
+	const totalSteps = (max - min) / step;
+
+	const progressWidth = (currentStep / totalSteps) * 100;
+	const rangeWidth = element.clientWidth;
+	const ratio =
```
bryceosterhaus: youd need one more line between rangeWidth and ratio

matuzalemsteles:
hey @pat270 is likely that we need to backport the range css to 2.x, Portal is not using the 3.x release to update css.

jbalsas:
Yeah, @matuzalemsteles, you're totally right :)

I did try to use `clay-css@3.x.x` in portal so we could drop `2.x.x` but ran into some breaking changes, so we might need to wait a bit longer, or figure out which ones so we can correct that.

packages/clay-slider/src/index.tsx
jbalsas: Awesome, @pat270! I thought we would need to control the thumb, but I see now that it's just a placeholder to figure out the thumb's native width?

PR #2317, opened 2019-08-22
feat(@clayui/tabs): add low-level Tabs component

### LINKED ISSUE #2154: High-level & low-level components for Tabs
Sample high-level API:
```jsx
<Tabs
	displayType // classic | modern
	items={[
		{
			active
			label
			href
			onClick
			items // for dropdown
		}
	]}
/>
```

matuzalemsteles:
Partial fix #2154

I ended up making some changes to minimize the Tabs APIs surface and a timely correction in the use of forwardRef.


packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface IProps extends Omit<React.HTMLAttributes<HTMLLIElement>, 'onClick'> {
+	/**
+	 * Flag to indicate if the component is active or not.
+	 */
+	active?: boolean;
+
+	/**
+	 * Tab Item component to render. Can be an 'anchor' or a 'button'.
+	 */
+	component?: ElementType;
```
matuzalemsteles: I intend to remove this API to use the new `LinkOrButton` component being introduced by #2225, so the component tag will be decided when passing `href` or `onClick`.

packages/clay-tabs/src/Content.tsx
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Receives a number that indicates the `tabkey` to be rendered.
+	 */
+	activeTabKey: number;
```
bryceosterhaus: `TabKey` is a bit strange. What about `activeIndex` since you are comparing it vs the array index.

packages/clay-tabs/src/Item.tsx
```diff
@@ -0,0 +1,92 @@
...
+	(
+		{
+			active = false,
+			children,
+			className,
+			component = 'button',
+			disabled = false,
+			href,
+			innerProps = {},
+			onClick = () => {},
```
bryceosterhaus: Is this no-op needed? If its not defined, the element just receives `undefined` which should be fine.

packages/clay-tabs/src/TabPane.tsx
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Flag to indicate if `fade` classname that applies a fading animation should be applied.
+	 */
+	fade?: boolean;
+}
+
+const delay = (fn: Function) => {
+	return setTimeout(() => {
+		fn();
+	}, 100);
```
bryceosterhaus: Should we be hardcoding a specific delay for animations, or should we be using `onAnimationEnd`? Typically I remember trying to avoid hardcoded `setTimeout` for animations.

packages/clay-tabs/src/__tests__/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+	it('renders', () => {
+		const {container} = render(<ClayTabs />);
+
+		expect(container).toMatchSnapshot();
+	});
+
+	it('renders justifying nav items', () => {
+		const {container} = render(<ClayTabs justified />);
+
+		expect(container).toMatchSnapshot();
```
bryceosterhaus: Instead of adding a bunch of snapshot tests, what do you think about testing explicit dom items? For example, in this case we could check `expect(container.querySelector('.nav.nav-justified')).toBeTruthy()`. This would at least help eliminate an accidental or blind update to all snapshots. And this way we also target specific truthy statements rather than the whole dom.

packages/clay-tabs/src/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+				`nav`,
+				{
+					'nav-justified': justified,
+					'nav-tabs': !modern,
+					'nav-underline': modern,
+				},
+				className
+			)}
+			role="tablist"
+			{...otherProps}
```
bryceosterhaus: Should this be declared first in the list of props? Otherwise, if someone passes `role` to this item, it'll overwrite `tablist`

matuzalemsteles: Oh yeah, thanks!

packages/clay-tabs/src/Content.tsx
matuzalemsteles: Makes sense, I'm going with `activeIndex`.

packages/clay-tabs/src/__tests__/index.tsx
matuzalemsteles: Oh yeah, I agree. I will take care of updating these tests to eliminate some of the snapshots.

packages/clay-tabs/src/TabPane.tsx
matuzalemsteles: I think this case is peculiar because in fact it is just a delay for as soon as the component render it adds the class and then it would cause an `onAnimationEnd`.

matuzalemsteles:
hey @bryceosterhaus updated, let me know if you see anything else.

PR #2318, opened 2019-08-22
fix(clay-css): SVG Icons reduce the size by remove inessential attrib…

### LINKED ISSUE #2309: Source Format SVG Icons
Reduce the size of SVG icons by removing inessential attributes and html comments.

pat270:
…utes and comments

Use closing tags instead of self closing (spams warnings in IE console)

fixes #2309

This saves ~53kb.


jbalsas:
Hey @matuzalemsteles, @pat270, could we have a view of the resulting `icons.svg`?

I'd be a bit worried about the generated id's for the symbols, that's all...

eduardoallegrini:
At first glance I don't see any problem, the only visual difference is about the default max size: now it's 100% and it fits any container, before it was 512px, but since these are icons using in the lexicon library it's not an issue. The `id`s are generated automatically by the Adobe Illustrator and it can be removed unless we are using it in some animation or particular customization (bad practice)

pat270:
@jbalsas 
The markup:
the-svg-markup
Icons rendered:
svg-icons

jbalsas:
Hey @pat270, I meant the `icons.svg` spritemap.

Just want to make sure we keep the symbol ids and are in no way tied to the original code and indeed autogenerated:

```
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <symbol id="add-cell" viewBox="0 0 512 512">
        ...
    </symbol>
    <symbol id="add-column" viewBox="0 0 512 512">
        ...
    </symbol>

    <symbol id="add-role" viewBox="0 0 512 512">
        ...
    </symbol>

    <symbol id="add-row" viewBox="0 0 512 512">
    </symbol>
    ...
</svg>

**Comment by pat270** [comment]
@jbalsas I can add them back in to keep it consistent if you like. Previously, the id's were inconsistent with the autogenerated version.

**Comment by jbalsas** [comment]
Hey @pat270, I don't mind the `id`s in the individual `.svg`s, but they need to be in the final SVG if I'm not mistaken.

We use:

```
<svg class="lexicon-icon">
	<use href="path/to/icons.svg#add-column" />
</svg>
```

And so, we need the final `icons.svg` to have symbol ids so that works.

I'm assuming that the id's in the spritemap are autogenerated, in which case it won't matter that we remove them from the individual svgs. I'm just asking to make sure we double-check.

pat270:
@jbalsas Nate's svg task autogenerates based on the file name. It's magic to me so I can't pinpoint exactly where.

marcoscv-work:
I think is working fine!

PR #2320, opened 2019-08-22
fix: stop closing DropDown or Modal when clicking inside element on a nested sub portal

### LINKED ISSUE #2288: Date picker component doesn't work well if rendered inside a dropdown
The date picker component gets rendered via portals and it loses the dom relationship with its parents. In particular scenarios it causes css and js problems like these:

Here we have a date rendered within a dropdown (the z-index is not correct):
image

If I click on a date, it closes the dropdown and changes the component position:
image

matuzalemsteles:
Fix #2288

- This prevents DropDown from closing when it has a portal nested to it, for example DatePicker.
- The same is done for Modal, it was closing when it has some action on these portals nested to it, the strategy was to wrap them all in a `fade modal` child element and simulate the element as if it were `modal-dialog` without an overflow for example to keep the modal closed when clicking outside the modal.
- There have been many updated snapshot tests due to the wraps on top of these components that affect many others.

Let me know if you think this is weird or have another strategy.

### DIFF
--- a/packages/clay-drop-down/src/Menu.tsx
+++ b/packages/clay-drop-down/src/Menu.tsx
@@ -5,7 +5,7 @@
  */
 
 import classNames from 'classnames';
-import React, {useLayoutEffect} from 'react';
+import React, {useLayoutEffect, useRef} from 'react';
 import {Align} from 'metal-position';
 import {ClayPortal} from '@clayui/shared';
 import {useDropdownCloseInteractions} from './hooks';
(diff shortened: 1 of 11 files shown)


packages/clay-drop-down/src/Menu.tsx
```diff
@@ -96,17 +95,19 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>((
...
-				className={classNames('dropdown-menu', className, {
-					'dropdown-menu-indicator-end': hasRightSymbols,
-					'dropdown-menu-indicator-start': hasLeftSymbols,
-					show: active,
-				})}
-				ref={ref}
-			>
-				{children}
+		<ClayPortal subPortalRef={subPortalRef}>
+			<div ref={subPortalRef}>
```
bryceosterhaus: Is this element necessary or can we just use the existing `ref`? If ref doesn't exist, we could have a default ref.

matuzalemsteles: Actually we can, but just I created another `ref` so that people do not have the `ref` of the div but the element... maybe it is something they do not expect.

bryceosterhaus: I think it might be best to use the existing `ref`, mostly so we don't have to have an extra div everywhere.

matuzalemsteles: Oh I guess I got it wrong, we need this extra div here because it will contain the nested portals inside if we use `dropdown-menu` as a sub portal it will break because the class has `overflow: hidden`.

PR #2323, opened 2019-08-23
feat(clay-css): Add usable mixins from bourbon

eduardoallegrini:


eduardoallegrini:
https://issues.liferay.com/browse/LPS-95490
In order to clean the portal from some old mixins library like bourbon, we are removing the usages from different parts of the portal but there are some still-usable mixins we should include in Clay

jbalsas:
@matuzalemsteles, @pat270, could you make sure this gets in for the `2.x` release so we can continue updating our mixin usage next week? 😘

pat270:
@jbalsas @eduardoallegrini  are we using autoprefixer only on the theme level or does it apply to module css also?

pat270:
New pulls:
3.x https://github.com/liferay/clay/pull/2326
2.x https://github.com/liferay/clay/pull/2325

PR #2346, opened 2019-08-28
chore(@clayui/navigation-bar): Adds a state for handling `triggerLabel` property on Navigation Bar stories

diegonvs:


matuzalemsteles:
hey @diegonvs you can take a look, I think there came some commits that shouldn't be here.

diegonvs:
Resent at #2359

PR #2354, opened 2019-08-29
chore(@clayui/modal): adds a warning when observer is not passed to ClayModal

matuzalemsteles:
Just adding a warning to help developers correctly use `ClayModal`.

### DIFF
--- a/packages/clay-modal/src/Modal.tsx
+++ b/packages/clay-modal/src/Modal.tsx
@@ -10,6 +10,7 @@ import Context, {IContext} from './Context';
 import Footer from './Footer';
 import Header from './Header';
 import React, {FunctionComponent, useEffect, useRef} from 'react';
+import warning from 'warning';
 import {ClayPortal} from '@clayui/shared';
 import {Observer, ObserverType, Size} from './types';
 import {useUserInteractions} from './Hook';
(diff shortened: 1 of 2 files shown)


packages/clay-modal/src/Modal.tsx
```diff
@@ -44,6 +45,20 @@ const ClayModal: FunctionComponent<IProps> & {
 }: IProps) => {
 	const modalBodyElementRef = useRef<HTMLDivElement | null>(null);
 
+	warning(
+		observer !== undefined,
+		`You need to pass the 'observer' prop to ClayModal for everything to work fine, use the 'useModal' hook that exposes the observer.
+
+> const {observer} = useModal({...});
```
matuzalemsteles: I had to push it aside to not appear with spaces in the console. 🙂

bryceosterhaus: Since its a bit awkward to have in the middle of the component. Could you declare this message as a variable above?

matuzalemsteles: sure, np

bryceosterhaus: feel free to merge when updated

PR #2363, opened 2019-08-30
fix(drop-down): automatically close dropdown when clicking on item

bryceosterhaus:
Two issues here:

1) `<ClayPaginationWithBar />` was an uncontrolled component. We should default our components to be controlled(both a value and onChange). This gives users a flexibility provide a specific value in reaction to some other change.

2) `<ClayDropDownWithItems />` isn't currently closing the dropdown after clicking an item. I added a context here so that it will automatically close when item is clicked.

Both of these issues arose from chatting with @thektan and @oliv-yu as they tried to use `<ClayPaginationWithBar />`

### DIFF
--- a/packages/clay-drop-down/src/DropDownWithItems.tsx
+++ b/packages/clay-drop-down/src/DropDownWithItems.tsx
@@ -25,7 +25,7 @@ interface IItem {
 	label?: string;
 	name?: string;
 	onChange?: Function;
-	onClick?: () => void;
+	onClick?: (e: React.MouseEvent<HTMLElement, MouseEvent>) => void;
 	symbolLeft?: string;
 	symbolRight?: string;
 	type?: TType;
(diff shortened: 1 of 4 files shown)


bryceosterhaus:
@matuzalemsteles is there a way to update docs at the same time as the PR? Would it just be in next.clayui.com?

matuzalemsteles:
In `next.clayui.com` uses development dependencies but we don't have any examples of `ClayPaginationWithBar` yet, it would be just to remember to add this in the release note for those coming from `milestone` releases but we also need to note these details in the documentation with DropDown automatic closing.

bryceosterhaus:
So add docs to DropDown in `next.clayui.com`?

matuzalemsteles:
Yeah, we can work on that on other PRs.

bryceosterhaus:
sounds good. ill go ahead and merge then

Issue #2275, opened 2019-08-14
[Discussion] High-level vs low-level component API


bryceosterhaus:
Opening this to continue discussion from this PR and doesn't get lost, https://github.com/liferay/clay/pull/2240#issuecomment-520930837.

Overall, I think we need to discuss what sort of guidelines or requirements we need to follow when constructing the APIs for high-level components. I think we are starting to run into issues where developers are required to make fairly subjective decisions on the API that a certain high-level component offers. For example, if we have component A and it has a list of basic lexicon interactions, how do we decide what high-level components get created? This is the sort of decision process that we might want to discuss further.

jbalsas:
Hey @bryceosterhaus, @matuzalemsteles, thanks for thinking this through and arriving to some compromises to avoid getting blocked.

As @matuzalemsteles says, we speak from experience here. We've failed to gain traction and usage in the past because we've failed to answer some of these questions. I want to think, as @wincent many times points out, that there's probably a sweet spot in between the different approaches where we could stand.

In Clay v2, we've been burned from every possible side:
> Your component is so high level that it doesn't allow me to do anything other than it was designed for

and 

> How come I have to reimplement everything on my own when I only want to tweak a little something on this overly complex component?

and

> I don't want small pieces to build my apps, this should be already provided so I can focus on my features!

and

> I don't want your components, I'd rather implement it all on my own so I have more control!

So, finding the sweet spot might definitely be harder than it seems 😂 

One key issue that I think @bryceosterhaus brought up is that **with little adoption and usage, it's hard for us to understand what's needed**. On top of that, I'll add that we need to go into the teams and work with them side by side to understand how is this being used so we can make better decisions. 

My suggestion would be something like:
- When in doubt, stick to low level components until a real usage is presented.
- Let's prioritize real usage issues so we can keep developers engaged and using Clay v3
- Try to allocate time (in my mind we could allocate up to 20%) to work with teams on new features/reviews/refactors to improve how Clay is used and our understanding of what's needed.

What do you think?

I know you had long conversations about this and I might have missed some key comments, so please, kindly bring them up again here and I promise I'll follow up 🙏 

Thanks!

bryceosterhaus:
Seems like an accurate summary to me and I would agree with all of it. A couple of other things come to mind as well, and some of this might overlap a bit.

### Library vs Product
[...]

PR #2373, opened 2019-09-03
feat(@clayui/form): add Select button in MultiSelect

### LINKED ISSUE #2366: MultiSelect does not have Select button
We should add the `Select` button to `ClayInputWithMultiSelect` or check if it is possible to compose.

Docs: https://docs.google.com/document/d/1cIx6RkWNbnho_fOSR__3Fp5alIBE_ObhO-jecfXlA9I/edit#heading=h.rbkugrj8xgm6

matuzalemsteles:
Fixes #2366

### DIFF
--- a/packages/clay-form/package.json
+++ b/packages/clay-form/package.json
@@ -27,6 +27,7 @@
 	],
 	"dependencies": {
 		"@clayui/autocomplete": "^3.0.0-milestone.3",
+		"@clayui/button": "^3.0.0-milestone.3",
 		"@clayui/drop-down": "^3.0.0-milestone.3",
 		"@clayui/icon": "^3.0.0-milestone.3",
 		"@clayui/label": "^3.0.0-milestone.3",
--- a/packages/clay-form/src/InputWithMultiSelect.tsx
+++ b/packages/clay-form/src/InputWithMultiSelect.tsx
@@ -6,6 +6,7 @@
 
 import classNames from 'classnames';
 import ClayAutocomplete from '@clayui/autocomplete';
+import ClayButton from '@clayui/button';
 import ClayDropDown from '@clayui/drop-down';
 import ClayForm from './Form';
 import ClayInput from './Input';
@@ -61,6 +62,19 @@ export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	onInputChange: (val: string) => void;
 
 	/**
+	 * Callback will always be called when the user clicks the select button,
+	 * passing a function will show the button.
+	 */
+	onSelectClick?: (
+		event: React.MouseEvent<HTMLButtonElement, MouseEvent>
+	) => void;
+
+	/**
+	 * Label for the button from Select.
+	 */
+	selectLabel?: string;
+
+	/**
 	 * List of pre-populated items that will show up in a dropdown menu
 	 */
 	sourceItems?: Array<string>;
@@ -86,6 +100,8 @@ const ClayMultiSelect: React.FunctionComponent<IProps> = ({
 	label,
 	onInputChange,
 	onItemsChange,
+	onSelectClick,
+	selectLabel = 'Select',
 	sourceItems,
 	spritemap,
 	...otherProps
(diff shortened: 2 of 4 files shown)


matuzalemsteles:
hey @bryceosterhaus you are totally right about that, i thought something similar and i think it would be much better to see it happening. We just got into one issue, since our high-level ceiling is Lexicon, we should ship a component that can be customized only by API following the specifications. But I agree to deliver it as you suggested, we can treat it as a low-level to make accession easier.

bryceosterhaus:
> since our high-level ceiling is Lexicon, we should ship a component that can be customized only by API following the specifications

Are you thinking we need to provide a _single_ component that meets lexicons specification via its props API? Maybe I am mistaken, but I thought our goal was to provide the _ability_ to build lexicon specifications with our low-level components and their API. For example, instead of having to deliver a single high-level component for MultiSelect that meets all specifications via its props, we would deliver the main blocks of components to compose that specification. Maybe it'd be helpful to have @jbalsas help clarify.

matuzalemsteles:
> Are you thinking we need to provide a single component that meets lexicons specification via its props API? Maybe I am mistaken, but I thought our goal was to provide the ability to build lexicon specifications with our low-level components and their API. For example, instead of having to deliver a single high-level component for MultiSelect that meets all specifications via its props, we would deliver the main blocks of components to compose that specification. Maybe it'd be helpful to have @jbalsas help clarify.

I think we're messing things up a bit here, I think I lost our idea of ​​high-level a bit. I see high-level in another way (as I said in #2275).

I think if we deliver high-level as part of the component's Lexicon specifications, it may override our low-level, in fact I see we went the other way here. We deliver most of our high levels with the entire component specification (or near) and the low levels as small parts that help achieve the specification.

bryceosterhaus:
> We deliver most of our high levels with the entire component specification (or near) and the low levels as small parts that help achieve the specification.

Yeah so what I am pushing for here is that we build low-level, which should be composable to meet lexicon specifications. Anything other than those low-level building blocks would be high-level. And instead of one high-level meeting every specification, we create one or more high-levels to meet a specific need that is used frequently, such as DropDownWithItems. It meets a basic need that is used frequently, but we don't try to adapt that high-level to meet _every_ need.

For this component specifically, I think the input portion of the component is technically a high-level component since it is comprised of other low-levels. However, the feedback and button portion should likely be composed together to create the full thing if needed. If we find that all three are used together really frequently, then we can create a high-level that meets that specific need. 

Does that make sense at all?

carloslancha:
@bryceosterhaus if the approach you're following for the componentes is the composition It makes sense to do this in that way, I guess.

I'm working now on implementing the multiselect un portal and for now I'm kind if strugling to realize how to do some things that in 2.x was made vía API. Maybe adding examples for all the cases (or make them configurable with knots) in the storybooks would ve helpful.

bryceosterhaus:
@carloslancha yeah thats a great idea. Would you mind linking to the one you are trying to implement? I can go through and try to add more demos to storybook for examples. Or even if there are a handful of components you want demos on, I can build out a handful, Im just not sure which examples to build out.

matuzalemsteles:
> For this component specifically, I think the input portion of the component is technically a high-level component since it is comprised of other low-levels. However, the feedback and button portion should likely be composed together to create the full thing if needed. If we find that all three are used together really frequently, then we can create a high-level that meets that specific need.
>
> Does that make sense at all?

Yeah, that makes sense. I think we can change `ClayInputWithMultiSelect` to a low-level component.

I agree with @carloslancha, even though we have these low-level components we should add more demos to our storybook showing how you can reach Lexicon specifications using the low-level components by listing the use cases predicted by the documents.

bryceosterhaus:
Technically I think `ClayInputWithMultiSelect` would still remain a high level component since it is high-level of input. But then we can also compose this high level with feedback and the button for the full range of multi-select specification. Does that sound about right to you? We may be saying the same thing in practicality but just using different language for it.

and im all for the extra demos. no reason we shouldn't add more.

matuzalemsteles:
> Technically I think ClayInputWithMultiSelect would still remain a high level component since it is high-level of input.

Hmm yeah, I think I'm thinking that MultiSelect is a different component from `Input`... you're right about that.

> But then we can also compose this high level with feedback and the button for the full range of multi-select specification. Does that sound about right to you? 

Sounds good for me.

I think we can go with that and start revamp `ClayInputWithMultiSelect` we have a lot of changes to it in other PR's.

bryceosterhaus:
@matuzalemsteles sounds good. Do you want to update that in this PR, or do you want me to go ahead and work on deconstructing it a bit? I'm fine either way.

matuzalemsteles:
@bryceosterhaus I can close this PR, and you can work on deconstructing it, then we can add to the other changes we're talking about in the other open PR's.

bryceosterhaus:
Sounds good. I'll add some examples for this exact case as well.

Thanks!

matuzalemsteles:
Thanks @bryceosterhaus

PR #2374, opened 2019-09-03
feat(@clayui/form): call `onItemsChange` callback only if value is unique

### LINKED ISSUE #2370: Multiselect is not controlling duplicated items
Actual result:
You can add the same item multiple times 

Expected result:
You can't add the same item multiple times or, at least, make it configurable.

image

matuzalemsteles:
Fixes #2370

I am adding a new API to `InputWithMultiSelect`, this decreases the work people have to do to follow the specification and also makes it possible to add their own rules.

The API works like a `compose`, so you can pass your rules with sole responsibility, the functions are called from right to left.

### DIFF
--- a/packages/clay-form/src/InputWithMultiSelect.tsx
+++ b/packages/clay-form/src/InputWithMultiSelect.tsx
@@ -19,7 +19,22 @@ const ENTER_KEY = 13;
 
 const DELIMITER_KEYS = [ENTER_KEY, COMMA_KEY];
 
+type TCompare = (
+	items: Array<string>,
+	newValues: Array<string>
+) => Array<string>;
+
 export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
+	/**
+	 * Function that compares the selected values with the new values that can be added,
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
Hmm, this seems like overkill of logic to provide for multi-select since this same thing can be achieved via is onChange and controlled aspect. By providing this parallel API, I think we are starting to oversaturate the API with complexity.

matuzalemsteles:
> Hmm, this seems like overkill of logic to provide for multi-select since this same thing can be achieved via is onChange and controlled aspect. By providing this parallel API, I think we are starting to oversaturate the API with complexity.

This is a good point but we have to try to add some things on our side so as not to let people do it every time we implement a MultiSelect, this is common to the component and probably everyone will have to. So this support introduces the default behavior that should work for most cases and allows people to add their own rule at any time.

If we don't do that, we have to start guiding people how to use Clay to get the specification result, people understand that just adding this component will work as expected.

bryceosterhaus:
Hey @matuzalemsteles, lets continue the conversation over on the issue, just so it doesn't get lost.

Issue #2376, opened 2019-09-03
ClayLabels focusing has unexpected results in InputWithMultiselect


oliv-yu:
QA on Search team pointed out a few bugs regarding the focus of the labels on InputWithMultiselect. Is the current behavior expected? Curious about the standards for accessibility, and I thought I would bring them up here, in case they were not discussed before. 

https://issues.liferay.com/browse/LPS-100506
- On removing an item in the input, the focus shifts directly to the close button of the item next to it.

https://issues.liferay.com/browse/LPS-100507
- The focus should be on the close button instead of on both the button and the entire label.

matuzalemsteles:
Hey @oliv-yu thanks for bringing up these questions!

> On removing an item in the input, the focus shifts directly to the close button of the item next to it.

This is a bug we have already tracked and we are correcting, the expected case is to remove the Label and focus back to the input.

> The focus should be on the close button instead of on both the button and the entire label.

The expected behavior would be to focus only on the label.

Hey @drakonux could you confirm that what I am saying here is correct or if we need to review some things.

marcoscv-work:
Hey guys, I suppose we changed the native behavior for some reason but I really think we should go back and not try to make a text element get the focus and let buttons be buttons and receive the focus from the beginning, same behavior than by using the mouse.

The only thing I think we should add is a correct name for the button which removes the label: "`remove " + label-name + " label`"

I've created an example here:
https://codepen.io/marcoscv/pen/BaBJBaa

You can use it with a screen reader

drakonux:
I agree with @marcoscv-work . I was talking about this with him this morning and what he says it makes total sense. 

Thus now tab order must jump from remove button to remove button instead from label to label as the old documentation says.

matuzalemsteles:
Thanks guys for the explanations, you can confirm how will continue the removal of labels in the behavior of when the input is empty and when pressing the backspace it focuses the label and when pressing again removes the label.

Just to keep it more centralized, we can add all these documented changes, we can lose it at any time, and our source for validating our components is tied to documents in Drive.

drakonux:
When the input is empty and the backspace key is pressed, remove the last label and the previous remove button gets the focus. Then:

- If the user presses **enter key** or **backspace** the label is removed. 
- If the user types. The input starts to work as it usually does.  

Additional behavior. When the label is removed, the screen reader should say: 

`"label-name removed"`

[...]

PR #2377, opened 2019-09-04
fix(@clayui/form): pressing enter on label with focus should remove

matuzalemsteles:
This is related to #2376

### DIFF
--- a/packages/clay-form/src/InputWithMultiSelect.tsx
+++ b/packages/clay-form/src/InputWithMultiSelect.tsx
@@ -187,13 +187,17 @@ const ClayMultiSelect: React.FunctionComponent<IProps> = ({
 										}}
 										key={i}
 										onKeyDown={({keyCode}) => {
-											if (keyCode !== BACKSPACE_KEY) {
-												return;
+											switch (keyCode) {
+												case ENTER_KEY:
+												case BACKSPACE_KEY:
+													if (inputRef.current) {
+														inputRef.current.focus();
+													}
+													removeItem();
+													break;
+												default:
+													break;
 											}
-											if (inputRef.current) {
-												inputRef.current.focus();
-											}
-											removeItem();
 										}}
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
Yeah, I think it follows the same idea that when pressing backspace the focus will go to the whole label and when pressing again removes it. @drakonux can confirm this for us.

bryceosterhaus:
Yeah my concern was that if the focus is on the label, that doesn't communicate to screen readers that hitting backspace would cause it to delete. I was under the impression that interactive elements need to be buttons/links/inputs. I would think that the `enter` keypress functionality would need to focus the "close" button and not the label.

drakonux:
Hi guys!

Well talking offline with @marcoscv-work this morning about https://github.com/liferay/clay/pull/2376 . We agreed that we shouldn't make assumptions about how to manage the focus when a user removes a label. So when the user clicks or press remove button the focus is not located on the next label or the input. 

Simply, if the user press tab key **again**. The focus will be located on the next remove button or input (if the label removed was the last)

matuzalemsteles:
Thanks @drakonux, can we document this? I'm following old behaviors here so that when we release Clay we don't have behavioral regressions.

> Simply, if the user press tab key again. The focus will be located on the next remove button or input (if the label removed was the last)

I don't know if this goes against what you said above, not controlling behavior when removing, since we do that we will let the browser behavior decide.

drakonux:
Yeah no problem. As I said, If you need something create a request in Lexicon because this is an intensive work for me (remember that none of the designers that wrote the old documentation is still in Lexicon). 

> I don't know if this goes against what you said above, not controlling behavior when removing, since we do that we will let the browser behavior decide.

I'm sorry. Finally, if you read the last comment from @marcoscv-work in [2376](https://github.com/liferay/clay/issues/2376) we agreed the behavior in that ticket because talking with Chema is the best option to implement for now (and solve most of the questions/doubts we had).

Issue #2379, opened 2019-09-04
Missing ClayMultiSelect hidden inputs for selectedItems


carloslancha:

matuzalemsteles:
Fixed #2380

PR #2380, opened 2019-09-04
feat(@clayui/form): add missing hidden input element

julien:


packages/clay-form/src/InputWithMultiSelect.tsx
```diff
@@ -45,6 +45,11 @@ export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	 */
 	items: Array<string>;
 
+	/**
+	 * Value used for the hidden input's name attribute
```
carloslancha: Maybe is better "Value used for every selectedItem hidden input's name attribute"

packages/clay-form/src/InputWithMultiSelect.tsx
```diff
@@ -181,36 +187,39 @@ const ClayMultiSelect: React.FunctionComponent<IProps> = ({
...
+												if (i === items.length - 1) {
+													lastItemRef.current = ref;
+												}
+											}}
+											spritemap={spritemap}
+											tabIndex={0}
+										>
+											{item}
+										</ClayLabel>
+										<input name={inputName}
```
carloslancha: Add blank line

packages/clay-form/src/InputWithMultiSelect.tsx
```diff
@@ -181,36 +187,39 @@ const ClayMultiSelect: React.FunctionComponent<IProps> = ({
...
+											type="hidden"
+											value={item} />
```
carloslancha: `/>` in a new line

packages/clay-form/src/__tests__/__snapshots__/InputWithMultiSelect.tsx.snap
```diff
@@ -44,6 +44,11 @@ exports[`ClayInputWithMultiSelect renders as not valid 1`] = `
         className="form-control form-control-tag-group"
         onKeyDown={[Function]}
       >
+        <input
+          name=""
```
wincent: I wanted to double check what would happen when using the component's default (empty string) name:

> If no name is specified, or name is empty, the input's value is not submitted with the form.

([Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input))

So seems ok.

packages/clay-form/src/__tests__/InputWithMultiSelect.tsx
```diff
@@ -128,7 +128,7 @@ describe('Interactions', () => {
 		expect(document.activeElement!.tagName).toEqual('BODY');
 
 		fireEvent.keyDown(
-			container.querySelector('input') as HTMLInputElement,
+			container.querySelector('input.form-control-inset') as HTMLInputElement,
```
wincent: This smells a bit implementation-detailish. Might be better to target by type eg.

```javascript
container.querySelector('input[type=text]')
```

packages/clay-form/src/__tests__/__snapshots__/InputWithMultiSelect.tsx.snap
julien: Thanks for checking that out.

matuzalemsteles:
Fixes #2379

PR #2397, opened 2019-09-05
feat: add issue templated for github issues

bryceosterhaus:
Inspired by https://github.com/liferay/liferay-frontend-guidelines/pull/82

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug.md
+++ b/.github/ISSUE_TEMPLATE/Bug.md
@@ -1,0 +1,17 @@
+---
+name: 🐛 Bug
+about: Issues for filing a Bug
+labels: bug
+---
+
+<!--
+
+Before making a proposal, have you used the issue search functionality?
+
+-->
+
+## Is there an example you can provide via codesandbox.com?
+
+## What are the steps to reproduce?
+
+## What is the expected result?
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/Bug.md
```diff
@@ -0,0 +1,17 @@
+---
+name: 🐛 Bug
+about: Issues for filing a Bug
+labels: bug
```
matuzalemsteles: Just a question, should we add the `type: bug` label here to reference our existing labels? or he had just created a new label.

bryceosterhaus: This is the GH syntax for picking up existing labels. 

You can check them out actually working on my local repo at https://github.com/bryceosterhaus/clay/issues/new/choose

bryceosterhaus:
@matuzalemsteles instead of creating a Feature Request one, I think the Proposal issue can cover that. Its a bit more generic to allow for other types of proposals as well.

matuzalemsteles:
> @matuzalemsteles instead of creating a Feature Request one, I think the Proposal issue can cover that. Its a bit more generic to allow for other types of proposals as well.

Oh ok, I get the thought of creating a proposal, because I already have something defined and feature request I don't want to think of something, but it's just an idea... but feel free to merge.

bryceosterhaus:
@matuzalemsteles lets start with this for now and see how people use it. We can always add more as we see more use cases.

wincent:
💍  I love this one.

PR #2402, opened 2019-09-06
add files page demo

bryceosterhaus:


matuzalemsteles:
That looks very good. Great to see all components working together. I think we can improve the organization of our Storybook, something like this: https://storybooks-official.netlify.com, can make it more visible.

The demos may have more interactions, we can increment gradually, I think this is a good start.

bryceosterhaus:
@matuzalemsteles updated to add nesting for components. If this sort of thing looks good to you, I'll clean up the commits and code and send for one last review.

matuzalemsteles:

matuzalemsteles:
@bryceosterhaus looks good

bryceosterhaus:
@matuzalemsteles updated. I cleaned up commit history and structure of demos.

I can imagine this is a good starting place for future demos. Any time we might need a more fully fleshed out demo, we can put it here. The more demos, the better.

bryceosterhaus:
@wincent hmm, maybe I have a narrow view of editors. sublime -> atom -> vscode. All of those have had non-case sensitivity for me. 🤷‍♂ 

I'm happy to change it back if we think its a better default.

wincent:
Consistency is good and we value it because it is one less thing to think about. We have two possible scopes here:

1. Internal consistency within Clay (easy to acheive, and arbitrary: pick whatever order you want).
2. Universal consistency across other projects, including liferay-portal (also easy, but you lose the ability to choose an arbitrary ordering because changing liferay-portal to sort case-insensitively is probably infeasible; my understanding is that the sorting in Java is case-sensitive, and that's something we can't realistically change at a reasonable cost).

The reason we went with case-sensitive in eslint-config-liferay is ultimately for harmony with liferay-portal. As far as Clay goes, you can choose whether you value "internal" or "universal" consistency.

PR #2404, opened 2019-09-07
chore: remove instances of href='javascript;'

bryceosterhaus:


packages/clay-card/src/Card.tsx
```diff
@@ -45,12 +45,13 @@ const ClayCard: React.FunctionComponent<IProps> & {
...
 	displayType,
 	horizontal,
-	href = 'javascript:;',
-	interactive,
+	href,
 	onClick,
 	selectable = false,
 	...otherProps
 }: IProps) => {
+	const interactive = !!(href || onClick);
```
matuzalemsteles: I'm not sure if that would be the best way to tell Card to be a Navigation.

PR #2409, opened 2019-09-09
fix(clay-css): Use `selector-unify` to combine parent selectors with …

### LINKED ISSUE #2392: Clay CSS Mixins use selector-unify to target closest parent selector
We use the pattern below to combine elements with parent selectors which doesn't produce the intended outcome.
```
@mixin example-mixin {
    @at-root {
        button#{&} {
            color: red;
        }
    }
}

.a-component .some-link { @include example-mixin; }
```
outputs:
```
button.a-component .some-link { color: red; }
```
We need:
```
.a-component button.some-link { color: red; }
```

`selector-unify` lets us grab the immediate parent.

pat270:
…nested element selectors in mixins and components

fixes #2392

### DIFF
--- a/packages/clay-css/src/scss/components/_utilities.scss
+++ b/packages/clay-css/src/scss/components/_utilities.scss
@@ -40,7 +40,7 @@
 	width: 100%;
 
 	@at-root {
-		ul#{&} {
+		#{selector-unify(&, ul)} {
 			@include list-unstyled;
 		}
 	}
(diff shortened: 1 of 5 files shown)


matuzalemsteles:
hey @pat270 It seems that this recent change eventually broke the clay-css build. Maybe it's related to https://github.com/sass/node-sass/issues/2420

pat270:
@matuzalemsteles I was worried about that. The Sass version we are using might not support it.

PR #2411, opened 2019-09-09
feat(@clayui/table): Adds ClayTableWithQuickActions high-level

diegonvs:


bryceosterhaus:
Going to close this for now and will re-open if it becomes relevant again for us when we add new table components

PR #2412, opened 2019-09-09
chore(CONTRIBUTING.md): add release process steps

matuzalemsteles:


CONTRIBUTING.md
```diff
@@ -199,3 +200,19 @@ You may want to use [Github search](https://help.github.com/articles/searching-i
...
 | Duplicate       | [search](https://github.com/search?q=is%3Aopen+is%3Aissue+repo%3Aliferay%2Fclay+label%3A%22resolution%3A+duplicate%22&type=Issues) | Resolution | Duplicates an existing issue                                                                                                                                                                                                                                                        |
 | Wontfix         | [search](https://github.com/search?q=is%3Aopen+is%3Aissue+repo%3Aliferay%2Fclay+label%3A%22resolution%3A+wontfix%22&type=Issues)   | Resolution | Will no longer be worked on                                                                                                                                                                                                                                                         |
+
+## Release process
+
+To publish a new version to monorepo, follow these steps:
+
+1. `git checkout master`
+2. Run `lerna version` and choose the version to cut.
+3. Run `git push $REMOTE master` - no problem since the last commit was green.
```
bryceosterhaus: Does this need `--follow-tags`? Also, instead of pushing to `liferay/clay`'s master, it might be best to push to a local master and send a PR for stable. What do you think?

wincent: What do you mean by "push to local master"?

This line doesn't need `--follow-tags` because that happens further down in the list (when pushing `stable`). And there is a PR for stable listed as step 4.

matuzalemsteles: Do you mean push to the origin branch of fork?

By default `lerna version` already tags, if we want to delay creating them when we see the draft with the green CI, we can skip tagging in this step and create feathers in step 6.

bryceosterhaus: > push to local master

Sorry, bad wording on my half. I mean pushing to a origin fork and then sending PR, rather than pushing directly to the origin repo source.

matuzalemsteles: Hmm, I see no problem following this path. It can be safer and avoids having problems like we had yesterday. Following this will change the steps, maybe some are no longer needed.

wincent: The downside of pushing to a fork, then merging from it, is that you create a basically meaningless merge commit.

That is, instead of having all these three things pointing at exactly the same thing:

1. A commit that updates the version number in the package.json files.
2. A Git tag corresponding to that version number.
3. A set of released NPM packages for the new version.

```
* chore: prepare for Clay vX.Y.Z release (vX.Y.Z)
```

You now have:

1. A commit that updates the version number in the package.json files.
2. A merge commit that points at that other commit and its parent (itself usually a merge).
3. A Git tag that points at that merge commit.
3. A set of released NPM packages that corresponds to a merge.

```
* Merge pull request #1200 from fork/prep-X.Y.Z (vX.Y.Z)
|\
| * chore: prepare for Clay vX.Y.Z release
|/
* Merge pull request #1199 from fork/new-feature
```

It's true that by pushing directly to the main repo you might push something wrong, but we can mitigate that by adding some extra steps (and automate them, if we want) that check that your local "master" (the one with the commit that updates the version number) is totally up-to-date with the remote master with that one additional commit on top).

matuzalemsteles: Thanks @wincent for the explanations. I'm fine with any decision we make here.

wincent: > I'm fine with any decision we make here.

Yeah, me too. You've got to have a release process that you are comfortable with.

matuzalemsteles:
I'm closing this in favor of #2630

PR #2413, opened 2019-09-09
3.0.0-alpha.0 pre-release

bryceosterhaus:


bryceosterhaus:
@matuzalemsteles have you seen these issues on ci before? first time I have seen these exact ones

matuzalemsteles:
@bryceosterhaus That's weird, I'll run CI again to check if it's not a problem with CI.

matuzalemsteles:
This is strange because it is not failing locally.

PR #2429, opened 2019-09-12
fix(clay-css): SVG Icons remove `import-export` and `embed` in favor of `order-arrow` and `code` respectively

pat270:


jbalsas:
Hey @pat270, let's keep the same `2.x` approach for `3.x` for now. That will make it easier for us to get rid of `2.x` since we won't have to account for breaking changes.

Thanks!

PR #2435, opened 2019-09-13
fix(clay-css): SVG Icons deprecate `import-export` and `embed` in favor of `order-arrow` and `code` respectively

pat270:
@jbalsas I also reverted the commit where we renamed `announcement`, `sticky`, `urgent` and deprecated those icons.


jbalsas:
LGTM!

PR #2470, opened 2019-09-16
feat(link): remove 'unstable_' from ClayLinkContext

bryceosterhaus:


bryceosterhaus:
@matuzalemsteles I noticed that `next.clayui.com` seems to be pulling its dependencies from npm rather than our yarn workspace. Do you know if this is intended? I was hoping to update the docs and simultaneous make changes, but wasn't sure if that is the right workflow.

matuzalemsteles:
@bryceosterhaus we also need to change the `package.json` versions of the `next.clayui.com` package so that it can get it from local. I will send a PR changing and you can rebase.

matuzalemsteles:
Send #2471

bryceosterhaus:
@matuzalemsteles seems to work now. Thanks! Let me know if you are okay with me merging this.

PR #2471, opened 2019-09-16
chore(next.clayui.com): update clay packages

matuzalemsteles:


bryceosterhaus:
@matuzalemsteles does this grab the local versions of clay or just the alpha.1 version via npm?

matuzalemsteles:
@bryceosterhaus when the version is the same as the local packages it will get the local versions.

PR #2476, opened 2019-09-17
fix: issue templates not picking up the right labels

bryceosterhaus:


bryceosterhaus:
merging since its trivial

Issue #2458, opened 2019-09-16
Verify clay-pagination for 3.0.0 release


bryceosterhaus:
- [ ] Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
- [ ] Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
- [ ] Does component accept a ref? Is the ref value what you would expect?
- [ ] Does this component allow for different translations?
- [ ] If this is a form element, does it accept a `name` attribute?
- [ ] Were you able to find documentation for this component? Was it clear?

Additional Questions:
- Is there anything about the public API that you would change?
- Is the naming of the available props clear?

julien:
I'm currently using a local project created with `create-react-app` to try the components and ran into this issue 

![2019-09-17-172618_1600x900_scrot](https://user-images.githubusercontent.com/5572/65056164-b05d9b80-d970-11e9-832c-7fbea07b6eb1.png)

Here's the code I was using and I might have forgotten something, but just in case someone can double check

```javascript
import React, {useState} from 'react';
import ClayPagination from '@clayui/pagination';

import '@clayui/css/lib/css/atlas.css';

const spritemap = require('@clayui/css/lib/images/icons/icons.svg');

function App() {
	const [active, setActive] = useState(1);
	const totalPages = 100;

  return (
		<div id="app">
			<ClayPagination
				arial-label="pagination"
				activePage={active}
				onPageChange={setActive}
				spritemap={spritemap}
				totalPages={totalPages}
			/>
		</div>
  );
}

export default App;
```

This one also needs support for extra attributes (it seemed that `aria-label` didn't get passed in my example) so I can send a PR for that

carloslancha:
Yep, I guess all components should accept `...otherProps` and render'em in the main (or more logic) element.

julien:
After speaking with @marcoscv @bryceosterhaus @diegonvs and @eduardoallegrini we think it might be better analysing each component to find a general pattern or at least a good way to support accesibility related attributes. So I'll won't send a PR just right now.

matuzalemsteles:
Publish #2608

PR #2479, opened 2019-09-17
fix(@clayui/time-picker): add support for `aria-*` and name attributes

julien:


packages/clay-time-picker/src/index.tsx
```diff
@@ -303,7 +310,7 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
 
 	return (
 		<div className="clay-time">
-			<ClayInput.Group>
+			<ClayInput.Group {...otherProps}>
```
julien: @bryceosterhaus I wasn't sure where to add this.
The `otherProps` in this case are used for `aria-*` attributes but it might be better to pick a better name?

matuzalemsteles: This is a bit confusing to have access to `aria-*`, for the context outside the component I think it makes sense, but adding it inside the component is a bit odd, depending on the context would probably be in the `inputs`?

julien: Yes, that's why I said I wasn't really sure of how to put them :joy: 

I agree that having them in the `input`s would be a good idea, but we'd need a way to separate what goes into each one, do you have an idea how we could do that and keep the API simple?

matuzalemsteles: No 😅, this is pretty confusing now haha

julien: @matuzalemsteles I'll make a few changes to this PR. 
For the moment I won't focus on accesibility related attributes (see [this](https://github.com/liferay/clay/issues/2458#issuecomment-532570784) comment) but just the `name` as this component will probably be used in forms.

matuzalemsteles:

matuzalemsteles:
@julien I rebased and updated the commit messages for the "Semantic Pull Request" bot to accept. #2521

julien:
@matuzalemsteles thanks

PR #2480, opened 2019-09-17
docs(@clayui/multi-step-nav): Update React documentation (#2454)

diegonvs:


next.clayui.com/content/docs/components/forms/multi-step-forms.mdx
```diff
@@ -0,0 +1,70 @@
+---
```
matuzalemsteles: I think it's more valuable to add information from this component and how to use it, just adding examples might not be a good idea since Storybook does the same.

matuzalemsteles:
hey @diegonvs, can you check if these changes are still needed? It seems that we have already merged this, so we can see if we can complement with your changes.

diegonvs:
Discarded, there is existing documentation for it.

PR #2483, opened 2019-09-17
fix:(multi-select) items need to be objects and add onClearAllButtonClick callback

carloslancha:


packages/clay-form/src/MultiSelectInput.tsx
```diff
@@ -155,7 +185,16 @@ const ClayMultiSelect: React.FunctionComponent<IProps> = ({
 
 		const pastedItems = pastedText
 			.split(',')
-			.map(item => item.trim())
+			.map(itemLabel => {
+				itemLabel = itemLabel.trim();
+
+				const newItem: Item = getSourceItemByLabel(itemLabel) || {
```
matuzalemsteles: Just out of curiosity, why do we need to get the item from the array of `sourceItems`?

carloslancha: 'cause most of the times the value of an item is no the same as the label, so if is an item present in the autocomplete items we need to get the correct value from there. If not present we just make the value the same as the label.

matuzalemsteles: Hm yeah, well thought.

packages/clay-form/src/MultiSelectInput.tsx
```diff
@@ -115,17 +136,21 @@ const ClayMultiSelect: React.FunctionComponent<IProps> = ({
 	useLayoutEffect(() => {
 		if (sourceItems) {
```
matuzalemsteles: We probably don't even need this conditional anymore.

```suggestion
		if (sourceItems.length > 0) {
```

bryceosterhaus:
I moved multi-select to its own package for now, we can add these changes over there. https://github.com/liferay/clay/tree/master/packages/clay-multi-select

Since we still have some work to do and the deadline for release is today, Chema said to just put multi-select as an alpha still, so that we can make changes if needed

carloslancha:
resent https://github.com/liferay/clay/pull/2544

PR #2484, opened 2019-09-17
fix(@clayui/alert): Forward refs for ToastContainer and Alert component

diegonvs:


bryceosterhaus:
@diegonvs we might actually back track on the idea of adding refs to everything. @matuzalemsteles raised a good point that itll get messy quickly and might be overkill. Plus I am additionally thinking that its always easier to add than it is to take away. In the case of Alert, I don't think we necessarily need to provide a ref right away. If someone needs it down the road though, we could possibly add one in.

What do you think?

diegonvs:
Makes sense, we could add `forwardRef` on demand. Would be a good idea change the question from `Does component accept a ref? Is the ref value what you would expect?` to `Makes sense forward ref for this component right now? Is the ref value what you would expect?`

bryceosterhaus:
Yeah thats probably a fine change, however, since there are so many issues I think its easier if I just post it in the slack to tell everyone.

PR #2485, opened 2019-09-17
fix(@claui/alert)!: Removes `variant` property in order to use the only available variant, called `stripe`, as a property.

diegonvs:


matuzalemsteles:
I still vote in favor of keeping the `variant` API even if it has only one property, it is much more readable, anyone without the context will not know what `stripe` will look like or not, within a `variant` helps remove some of that feeling.

bryceosterhaus:
I think I am like 55/45. 55 to keep it variant, and 45 to rename it. I think we should just keep it as is for now, it gives us better scalability for the future as well in case more variants are possible

PR #2487, opened 2019-09-17
fix(@clayui/navigation): Was added some missing aria attributes for compliance with https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

diegonvs:


packages/clay-navigation/src/Breadcrumb.tsx
```diff
@@ -69,21 +69,23 @@ export const ClayBreadcrumbNav: React.FunctionComponent<IProps> = ({
...
-							active={(item as IBreadcrumbItem).active}
-							href={(item as IBreadcrumbItem).href}
-							key={`breadcrumbItem${i}`}
-							label={(item as IBreadcrumbItem).label}
-							onClick={(item as IBreadcrumbItem).onClick}
-						/>
-					)
-			)}
-		</ol>
+		<nav aria-label="Breadcrumb">
```
matuzalemsteles: I'm wondering if wrapping this in a `nav` can prevent it from composing with other components. And if you follow that, the `otherProps` and `className` should be in `nav`. I'm not sure about that but I prefer to keep it as it is for now because it doesn't stop people from adding...

bryceosterhaus:
I agree with @matuzalemsteles I think we should leave the nav off of this so that it can be composed. Additionally, when we do add any aria attributes, we need to make sure we don't hard code any english specific words, it should be translatable.

PR #2489, opened 2019-09-18
chore: adds yarnrc to each package

bryceosterhaus:
So I was looking into release processes and there isn't exactly a clear way forward in starting this process. For this first release, I think we may have to publish packages individually as they are ready by running the `yarn version` and `yarn publish` commands in the specified package. Lerna doesn't exactly help us yet, since not all of the components are ready at this point. Once the majority of packages are released, I _believe_ lerna should be beneficial at that point. Otherwise, if we start with lerna right off the bat, we end up having to skip every package in the `lerna version` command except the one we intend to publish, so its sort of messy.

I am still going to investigate this, but I wanted to send this first in case anyone has experience with this.

### DIFF
--- a/packages/browserslist-config-clay/.yarnrc
+++ b/packages/browserslist-config-clay/.yarnrc
@@ -1,0 +1,3 @@
+# Make `yarn version` produce the right commit message and tag for this package.
+version-tag-prefix "browserslist-config-clay@"
+version-git-message "browserslist-config-clay@%s"
--- a/packages/clay-alert/.yarnrc
+++ b/packages/clay-alert/.yarnrc
@@ -1,0 +1,3 @@
+# Make `yarn version` produce the right commit message and tag for this package.
+version-tag-prefix "@clayui/alert@"
+version-git-message "@clayui/alert@%s"
--- a/packages/clay-autocomplete/.yarnrc
+++ b/packages/clay-autocomplete/.yarnrc
@@ -1,0 +1,3 @@
+# Make `yarn version` produce the right commit message and tag for this package.
+version-tag-prefix "@clayui/autocomplete@"
+version-git-message "@clayui/autocomplete@%s"
--- a/packages/clay-badge/.yarnrc
+++ b/packages/clay-badge/.yarnrc
@@ -1,0 +1,3 @@
+# Make `yarn version` produce the right commit message and tag for this package.
+version-tag-prefix "@clayui/badge@"
+version-git-message "@clayui/badge@%s"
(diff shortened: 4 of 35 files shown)


bryceosterhaus:
Potentially thinking about this for initial release process

To publish a new version of a package, follow these steps:

1. `git checkout master`
2. Navigate to package you want to publish and verify it is no longer `"private": true`
3. Verify any needed dependencies are published first.
4. Run `yarn version` and choose the version to cut.
5. Send a draft PR(not intended to be merged) asking to merge "master"; the sole purpose of this is to see CI green one last time before pushing the tag.
6. Run `git push $REMOTE master --follow-tags` - no problem since the last commit and PR was green.
7. Run `yarn publish` - This will push the package to NPM.

### Caveats
- No longer have a `stable` branch and would instead rely primarily on tags for package versions and stability. If we had a stable branch and also released packages independently, we could not gaurauntee what a `stable` branch would even be.
- More overhead for this intitial 3.0.0 release by not using lerna. Each package is manually released.
- Re-enable using lerna for versioning and releasing once the bulk of packages have been released on 3.0.0. By using lerna in the future, it will help us in updating multiple packages, updating their dependencies, and publishing packages in the correct order.

wincent:
One thing we're doing in some other projects is letting Yarn create the tag when you run `yarn version`. This, combined with a `.yarnrc` file [like this one](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/.yarnrc) ensures that the tag is always in the right format.

bryceosterhaus:
@wincent i was looking at doing that, however I was hoping I could automatically take the package name to do something like `@clayui/css@3.0.0`. Do you know if thats possible? Or will I have to configure that in each package?

bryceosterhaus:
@matuzalemsteles @wincent I went ahead and added a .yarnrc to each package for consistency. Let me know what you think.

bryceosterhaus:
approved offline, aka via slack

wincent:
> Let me know what you think.

Nice! One thing I noticed was that the Semantic Pull Request bot doesn't like the Git commit messages because they don't have "type:" prefix. I added a prefix in the liferay-npm-scripts repo [over here](https://github.com/liferay/liferay-npm-tools/pull/264) but I haven't tested it yet, although I think it should work.

bryceosterhaus:
@wincent good catch, I didn't think about that. Ill send an update.

wincent:
BTW @bryceosterhaus, I tested that PR [by preparing a release](https://github.com/liferay/liferay-npm-tools/pull/265) and can confirm the bot is happy with it.

PR #2491, opened 2019-09-18
chore: move @clayui/css as a peerDependency

bryceosterhaus:
`@clayui/css` is function more as a peer dependency and not an explicit devDependency. By moving to peerDependencies and by also changing the version to "*" we are communicating that clay components are components based on the latest css release.

Let me know if anyone sees issue with this.


bryceosterhaus:
note, the 4 space tab change was to make sure out package.json's also followed our prettier config.

bryceosterhaus:
Oh good point. Maybe we should do `3.x` instead of `*`. what do you think?

matuzalemsteles:
> Oh good point. Maybe we should do 3.x instead of *. what do you think?

Yeah, I think this is safer.

bryceosterhaus:
updated to 3.x, going to merge

PR #2498, opened 2019-09-18
fix(@clayui/navigation): Adds missing `otherProps` for Vertical Navigation

diegonvs:


packages/clay-navigation/src/Vertical.tsx
```diff
@@ -21,6 +21,11 @@ interface IItem {
 	 */
 	onClick?: () => void;
 
+	/**
+	 * Object of props for an item.
+	 */
+	itemProps?: Object;
```
matuzalemsteles: Why an exclusive prop? where the element is the parent of the component.

packages/clay-navigation/src/Vertical.tsx
```diff
@@ -154,6 +160,7 @@ export const ClayVerticalNav: React.FunctionComponent<IProps> = ({
 	items,
 	large,
 	spritemap,
+	...otherProps
```
matuzalemsteles: The `IProps` interface needs to extend from the `HTMLElement` interface, it looks like we do not have an `HTMLNavElement`.

packages/clay-navigation/src/Vertical.tsx
diegonvs: Solved after some internal conversation!

packages/clay-navigation/src/Vertical.tsx
diegonvs: Solved after some internal conversation!

PR #2502, opened 2019-09-18
refactor(@clayui/label): wrap React.forwardRef in component

matuzalemsteles:


matuzalemsteles:
Since these are not critical changes, I'm going with this forward. I'm going to release `@clayui/label` today.

PR #2503, opened 2019-09-18
chore: remove remnants of package @clayui/select and @clayui/css package from devDeps

matuzalemsteles:


matuzalemsteles:
I will go through with this since it's not trivial

PR #2505, opened 2019-09-19
feat(popover): add prop for disabling scroll

bryceosterhaus:


packages/clay-popover/src/index.tsx
```diff
@@ -60,7 +66,7 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 			>
 				<div className="arrow" />
 
-				<div className="inline-scroller">
+				<div className={disableScroll ? '' : 'inline-scroller'}>
```
diegonvs:
```suggestion
				<div className={classNames({'inline-scroller' : !disableScroll})}>
```

bryceosterhaus:
@diegonvs updated

PR #2507, opened 2019-09-19
fix(@clayui/drop-down): Skips adequately a disabled dropdown item when navigating via keyboard

diegonvs:


bryceosterhaus:
@diegonvs i think there may be an extra commit here. Is the navigation one intended?

diegonvs:
> @diegonvs i think there may be an extra commit here. Is the navigation one intended?

Yes 😓

diegonvs:
Resent at: https://github.com/liferay/clay/pull/2508

PR #2508, opened 2019-09-19
docs(@clayui/drop-down): Adds a disabled item in a example

diegonvs:


packages/clay-drop-down/src/Item.tsx
```diff
@@ -75,6 +75,7 @@ const ClayDropDownItem: React.FunctionComponent<IProps> = ({
 				href={href}
 				onClick={onClick}
 				ref={innerRef}
+				tabIndex={disabled ? -1 : otherProps.tabIndex}
```
bryceosterhaus: Instead of pulling tabIndex from otherProps, lets just destructure tabIndex from above

diegonvs: Good!

Issue #2461, opened 2019-09-16
Verify clay-progress-bar for 3.0.0 release


bryceosterhaus:
- [x] Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
- [x] Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
- [ ] Does component accept a ref? Is the ref value what you would expect?
- [x] Does this component allow for different translations?
- [x] If this is a form element, does it accept a `name` attribute?
- [x] Were you able to find documentation for this component? Was it clear?

Additional Questions:
- Is there anything about the public API that you would change?
- Is the naming of the available props clear?

mateomustapic:
I can take this one, you can assign it to me :)

marcoscv-work:
- [x] **HTML**
- [x] **CSS**
- [x] **Accessibility basics** 
  - We need to add `sr-only` text, from: `<div class="progress-group-addon">70%</div>` to `<div class="progress-group-addon">70% <span class="sr-only">loaded</span></div>` and `<div class="progress-group-addon">50/100</div>` to something like `<div class="progress-group-addon">50/100 <span class="sr-only">items uploaded</span></div>`

mateomustapic:
> * [ ]  Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
https://codesandbox.io/s/hopeful-einstein-cv0fc
It was easily implemented after small issues with implementing dependencies to codesandbox.
> * [ ]  Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
Yes, tested with setting `title` attribute to progressbar.
> * [ ]  Does component accept a ref? Is the ref value what you would expect?
No, it does not because it is a function component
> * [ ]  Does this component allow for different translations?
Yes
> * [ ]  If this is a form element, does it accept a `name` attribute?
This is not a form element
> * [ ]  Were you able to find documentation for this component? Was it clear?
There were no issues wen using documentation.
> Additional Questions:
> * Is there anything about the public API that you would change?
No
> * Is the naming of the available props clear?
Yes

bryceosterhaus:
@marcoscv-work a user should be able to achieve that by doing. Granted, this is extra information to add, but I think we might want to start without it first and then seeing how people use it.

```jsx
<ClayProgressBar spritemap={spritemap} value={50}>
	{`50/100`}
[...]

PR #2510, opened 2019-09-19
feat(@clayui/progress-bar): Adds a property called `addonLabel` for improving the message of addons for screen-readers (#2461)

diegonvs:


bryceosterhaus:
I think we should hold off on this until we evaluate its uses and needs. Currently people can still add their own accessibility, so we at least allow it for now. Maybe down the road we can add it in as internal functionality.

Issue #2457, opened 2019-09-16
Verify clay-navigation-bar for 3.0.0 release


bryceosterhaus:
- [ ] Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
- [ ] Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
- [ ] Does component accept a ref? Is the ref value what you would expect?
- [ ] Does this component allow for different translations?
- [ ] If this is a form element, does it accept a `name` attribute?
- [ ] Were you able to find documentation for this component? Was it clear?

Additional Questions:
- Is there anything about the public API that you would change?
- Is the naming of the available props clear?

mateomustapic:
@bryceosterhaus Please assign this issue to me

mateomustapic:
> * [ ]  Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
> Yes, no problems implementing demo.
https://codesandbox.io/s/modest-morse-whtnt
> * [ ]  Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
> Yes, accepted additional title attribute.
> * [ ]  Does component accept a ref? Is the ref value what you would expect?
>  It accepts ref, but it is not expected to be used in component
> * [ ]  Does this component allow for different translations?
> It is navigation bar with <a> element, so I suppose it allows translations
> * [ ]  If this is a form element, does it accept a `name` attribute?
> Not a form element.
> * [ ]  Were you able to find documentation for this component? Was it clear?
> Yes, the documentation was clear
> Additional Questions:
> /
> * Is there anything about the public API that you would change?
> nope
> * Is the naming of the available props clear?
> All the props are pretty straight-forward and named appropriately

PR #2517, opened 2019-09-19
fix(@clayui/navigation-bar): replace ClayLink with ClayButton and add truncated text

matuzalemsteles:
Related #2457


matuzalemsteles:
I'm merging this to release this package today.

Issue #2456, opened 2019-09-16
Verify clay-navigation for 3.0.0 release


bryceosterhaus:
- [x] Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
- [x] Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
- [x] Does component accept a ref? Is the ref value what you would expect?
- [x] Does this component allow for different translations?
- [x] If this is a form element, does it accept a `name` attribute?
- [ ] Were you able to find documentation for this component? Was it clear?

Additional Questions:
- Is there anything about the public API that you would change?
- Is the naming of the available props clear?

diegonvs:
- [ ]  Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
> ClayVerticalNav component doesn't support `otherProps` and doesn't treat A11y.

- [ ]  Were you able to find documentation for this component? Was it clear?
🙅‍♂ VerticalNav 
🙅‍♂ Breadcrumb

Additional Questions:

Is there anything about the public API that you would change? 
Is the naming of the available props clear?

> I don't consider `ellipsisBuffer` property too clear. However, there is a pattern on Clay with this name.

bryceosterhaus:
missing dependency of drop-down

eduardoallegrini:
- [x] **HTML**
- [x] **CSS**
- [ ] **Accessibility basics** 
  - In the [w/ vertical example](https://storybook.clayui.com/?path=/story/components-claynavigation--w-vertical) the links with dropdown are not targetable from keyboard

PR #2518, opened 2019-09-19
fix(@clayui/navigation): removes unnecessary use of FocusScope

matuzalemsteles:
Related #2456


matuzalemsteles:
I'm going ahead with this,  this don't need `FocusScope` here because `ClayDropDown` already has it.

Issue #2466, opened 2019-09-16
Verify clay-tabs for 3.0.0 release


bryceosterhaus:
- [x] Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
- [x] Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
- [x] Does component accept a ref? Is the ref value what you would expect?
- [x] Does this component allow for different translations?
- [x] If this is a form element, does it accept a `name` attribute?
- [x] Were you able to find documentation for this component? Was it clear?

Additional Questions:
- Is there anything about the public API that you would change?
- Is the naming of the available props clear?

julien:
> Is there anything about the public API that you would change?

Not at the moment.

> Is the naming of the available props clear?

Yes.

marcoscv-work:
- [x] **HTML**
- [x] **CSS**
- [x] **Accessibility basics** 
  - We need to establish a relationship between the `tab` and the `tabpanel` by adding a label to the panel, we could use `aria-labelledby="tab-X"` or `aria-label`. and add to the tab `aria-controls="tabpanel-X"`

eduardoallegrini:
I tried to add `aria-controls="value"` but it's not possible to target the `<a>` tag

marcoscv-work:
https://getbootstrap.com/docs/4.3/components/navs/#javascript-behavior

matuzalemsteles:
I added this to our stories #2522

PR #2522, opened 2019-09-19
fix(@clayui/tabs): add the relationship between tab and tabpane in stories

matuzalemsteles:
#2466


matuzalemsteles:
I'm merging this to be able to release this package today, if you see something wrong we can change later this does not prevent the release.

PR #2524, opened 2019-09-20
fix(form): remove high-level composed components from form

bryceosterhaus:


bryceosterhaus:
Updated to move multi-select into its own component. Its private for now, primarily a placeholder until we decide what to do with it.

bryceosterhaus:
Talked with @jbalsas. Plan is to merge this pr and release forms 3.0.0 and then also release @clay/multi-select on an alpha version. This allows us to still decide what to do with multi-select but also gives portal the ability to use the alpha version for now.

PR #2525, opened 2019-09-20
chore(@clayui): Review monorepo's dependencies

diegonvs:

PR #2530, opened 2019-09-23
chore(clayui): Adds a script for check depedencies issues

### LINKED ISSUE #2523: CI check for missing dependencies?
I've noticed we have run into a handful of "missing dependency" issues due to using a workspace. The package will likely run fine in the workspace without declaring an explicit dependency in the package.json. I think we need to incorporate some sort of CI check to verify that all of our used dependencies are also located in our package.json

diegonvs:
The saga started last Friday when I and @bryceosterhaus noticed about some missing/unnecessary dependencies in some packages.(https://github.com/liferay/clay/pull/2525/) STOP

> BUT DIEGO, YOU GUYS MISSED SOME DEPENDENCIES WHEN RELEASING THE PACKAGES?

No, apparently there aren't any package breaking in production due to it. BUUUUT monorepos aren't too easy to maintain and sometimes we miss these things.

With these issues, we thought that a script on CI for checking these dependencies would be a good one.

Still, on Friday, I decided to start reviewing some solutions by the community, like https://github.com/depcheck/depcheck but I got no success because of the usage on monorepos is very weak. @wincent, on Monday, presented me about some studies that he was doing at the beginning of the year about monorepo scripts and he presented me for some experimental scripts.

Yes, On this PR I just copied and paste a script for checking monorepo's dependencies issues just adding few things for adding support for our monorepo.

Few checks as:

- Check for (obviously) undeclared package dependencies
- Check for mismatched dependency versions
- Keep devDeps on the root of monorepo

Currently, we're failing when checking for mismatched dependency versions in monorepo because we have different versions of a "workspaced dependency" in different packages. Talking with @matuzalemsteles and @bryceosterhaus in past we decided to keep the dependencies using this style due to we're releasing the packages in an `independent` way. Would be a good idea just rewrite this check for work in our way of release or Should we keep this conversation about our releasing process on the monorepo?

Fixes #2523


scripts/check-deps/check-dependencies.js
```diff
@@ -0,0 +1,319 @@
...
+			}
+		}
+	}
+
+	print();
+	return success;
+}
+
+/**
+ * Can be used if
```
wincent: If what?

scripts/check-deps/src/getPackageConfig.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {join, resolve} = require('path');
+
+function getPackageConfig(name) {
+	return require(resolve(join('packages', name, 'package.json')));
```
wincent: Might want to rewrite this to use `fs.readFile` because it will trigger the [no-dynamic-require](https://github.com/liferay/eslint-config-liferay/blob/master/plugins/eslint-plugin-liferay/docs/rules/no-dynamic-require.md) rule once https://github.com/liferay/clay/pull/2536 gets merged (I left that one there for @matuzalemsteles to merge when he considered appropriate, but if it sits there too much longer it is going to get conflicts and have to be redone).

scripts/check-deps/check-dependencies.js
diegonvs: I forgot it, I gonna update this description. 😅

scripts/check-deps/src/getPackageConfig.js
diegonvs: Solved!

packages/clay-autocomplete/package.json
```diff
@@ -31,10 +31,6 @@
 		"@clayui/loading-indicator": "^3.0.0",
 		"fuzzy": "^0.1.3"
 	},
-	"devDependencies": {
-		"@clayui/data-provider": "^3.0.0",
```
bryceosterhaus: I noticed that we are removing a lot of devDependencies that are needed for the components storybook. Is this intentional?

diegonvs: Yes, these declarations of devDependencies for each package inside our monorepo is unnecessary. We just need these deps to be top level :)

diegonvs:
Ready to go! @matuzalemsteles @bryceosterhaus @wincent

scripts/check-deps/check-dependencies.js
```diff
@@ -0,0 +1,324 @@
+/* eslint no-for-of-loops/no-for-of-loops: 0 */
```
bryceosterhaus: For all of these files that are straight copies, can we add something in the comment block above that says something like

> File copied directly from https://github.com/wincent/js/blob/master/packages/workspace-scripts/src/dependencies.ts

This way we know exactly where its from and then also not any deviation from that copy.

wincent: FWIW, I don't think you really need it anywhere in the file, just mentioning it in the commit message is enough. (Consider it a fork, but one where it doesn't matter if it ever deviates and it won't have to track the upstream.)

Normally when you copy code you'd have to worry about licensing and preserving headers and all that (this project is BSD licensed; that repo is MIT licensed), but in this case I'm the original author and I told @diegonvs to do whatever he wants with the code 😂.

*Signed, @wincent* 🖊 

About license headers though, I think this ESLint suppression should go below the SPDX license header; we always want that to be first.

bryceosterhaus: I'd be fine with just a commit message as well. Primarily I just want us to be able to know "oh this is forked from this existing file".

diegonvs: @bryceosterhaus @wincent see 4e00e86

diegonvs: It's enough?

diegonvs:
@bryceosterhaus @wincent @matuzalemsteles updated!

scripts/check-deps/check-dependencies.js
bryceosterhaus: Hmm, could we maybe point to the exact file instead of just the repo? just since they are named differently

wincent: Seeing as we're having a great time talking about this, can we also switch `0` to `'off'` (because it is a bit more explicit), or better yet:

```javascript
/* eslint-disable no-for-of-loops/no-for-of-loops */
```

For consistency with what we mostly seem to do:

- [Hits for "eslint-disable" in liferay-portal](https://github.com/liferay/liferay-portal/search?q=eslint-disable&unscoped_q=eslint-disable)
- [Hits for "eslint-disable" in this repo](https://github.com/liferay/clay/search?q=eslint-disable&unscoped_q=eslint-disable) (albeit with some inconsistency there in `IncrementalInteractions.tsx`).

bryceosterhaus:
I finally got around to testing this, and it doesn't actually seem to catch any missed dependencies. I removed some dependencies in a package, and ran the script, and said everything was fine.

@wincent does your latest version of this work? Would I be able to just add your package? I am just looking for something simple to verify on ci that dependencies are correct.

wincent:
> I finally got around to testing this, and it doesn't actually seem to catch any missed dependencies. I removed some dependencies in a package, and ran the script, and said everything was fine.
> 
> @wincent does your latest version of this work? Would I be able to just add your package? I am just looking for something simple to verify on ci that dependencies are correct.

My version has always worked for me, but I have no idea if it will work outside of that project — it might, but I never tested it. I'm going to pull this down and see if I can repro your result.

wincent:
@bryceosterhaus: ok, so I rebased and force-pushed to update the PR.

There were/are a few problems here:

1. The missing dependencies check must run on the *built* version of the files (in order to catch missing dependencies corresponding to imports add by Babel), and our `yarn ci` task doesn't do a build before running the check, which means there are no files to actually check. I fixed that in 9edb7bde161ec46eea416be4bafcece9df62065f. Note that I put the check at the end of `yarn ci` because it is the slowest one to run and we can fail faster and provide feedback sooner y running the others first instead.
2. Somehow the code in here has diverged from [the upstream code](https://github.com/wincent/js/blob/7477dc15b134bb72210b860efa4d0bbc9e118fb8/packages/workspace-scripts/src/dependencies.ts), and some `console.log` revealed that it wasn't actually checking any files, so I fixed that in https://github.com/liferay/clay/pull/2530/commits/04361f1bfbda7347dba63ddcbe7238399d2e5208
3. There are some legit errors for missing dependencies in master right now so I fixed those in https://github.com/liferay/clay/pull/2530/commits/c591e2fcf5a50fe2836854cc5aac5ea47679c0ab

Some samples:

- Initial run before I fixed the failure to actual walk the tree: falsely shows no errors: https://gist.github.com/wincent/ba78fd063a74424fa11ff94bb0ad5070
- Fake errors that I added for demo purposes: https://gist.github.com/wincent/0e50866606cb38fdaf54680e0e53f4da
- Error report: https://gist.github.com/wincent/82bc4c83e74b47889b4709bb4a60564e
- Remaining legit errors actually on master: https://gist.github.com/wincent/b09e88aa6dd813bd2a6e7958a4e2fe5d

wincent:
Ah, two other things I noticed why looking at this:

1.  I am not sure of the value of the `checkForNonCurrentInternalDependencies` check. In my original script, this is useful because I use exact version internal dependencies (ie. `x@1.0.0` depends on `y@1.1.0`) and it is possible that all my dependencies could match (checked by `checkForMismatchedDependencyVersions`) but all pointing at an older version. This doesn't seem like it could happen in Clay because we use `^x.y.z` ranges instead. In my testing in Clay, there were no errors that were caught by `checkForNonCurrentInternalDependencies` that weren't already caught by `checkForMismatchedDependencyVersions`. I suspect we can just delete `checkForNonCurrentInternalDependencies` and drop the import of the `semver` package.
2.   The message that is printed when you are missing a dependency advises you to run commands like this:
   
       (cd packages/clay-breadcrumb && yarn add warning)
       (cd packages/clay-list && yarn add @clayui/form)
    
We probably want to update that to tell you the *actual* command that you want to run, which in the case of Clay is going to (at least) be an inexact version range (`^`).

bryceosterhaus:
Awesome, thanks for looking into this @wincent. 

As for removing `checkForNonCurrentInternalDependencies `, that sounds good with me. Additionally, do you think we should add this check under a `prepublishOnly` script as well? Since publishing is what we ultimately want to guard from, I thought it might be helpful to add there. However, the flip side is that if we just leave it on CI, we likely should never be publishing a branch that hasn't already passed on CI.

package.json
```diff
@@ -4,7 +4,9 @@
 		"build": "lerna run build && lerna run build:types",
 		"build:csb": "lerna run --no-bail --ignore 'clayui.com' build && lerna run --no-bail --ignore 'clayui.com' build:types",
 		"compile": "lerna run compile",
-		"ci": "yarn lint && yarn format:check && yarn test",
+		"checkDeps": "node scripts/check-deps/check-dependencies.js",
```
bryceosterhaus: I think we might want to add `yarn build` here and remove it from ci. Otherwise we might forget to build if we are checking this locally.

wincent: Sounds reasonable, although the only bummer is that `build` is super slow (not sure if it gets any faster on 2nd, 3rd etc run either 😢).

wincent:
> Additionally, do you think we should add this check under a `prepublishOnly` script as well?

That would definitely be a logical place for it; again the only downside is how slow the darn `build` step is.

> However, the flip side is that if we just leave it on CI, we likely should never be publishing a branch that hasn't already passed on CI.

True, true. Independently of all this, might be nice to explore speeding up the build at some point.

bryceosterhaus:
@wincent lets go ahead add the `yarn build` to the `checkDeps` script, remove the `checkForNonCurrentInternalDependencies ` check,  and then lets not worry about adding it to prePublish yet.

We can then see how it goes from there. Odds are, people likely won't run checkDeps multiple times or may not notice it until it hits CI. So I dont think the build time will be too much of a hassle. If it becomes an issue we can re-evaluate though.

Does that sound good to you?

wincent:
> Does that sound good to you?

👍

wincent:
Sample CI run here: https://travis-ci.org/liferay/clay/builds/624041657 (whole thing took 12 minutes).

bryceosterhaus:
Looks good to me. Going to merge and then additionally work on speeding up build time. I have a few ideas.

spawned new issue https://github.com/liferay/clay/issues/2786

PR #2536, opened 2019-09-24
chore: update to latest eslint 6.4.0 and eslint-config-liferay 11.0.1

wincent:
~Some concerning ones in here, like `useCallback` being undefined, but mostly just trivial.~

~Here's the full list:~

~https://gist.github.com/wincent/bda712046d33c230ca0fabc4dd71f996~

~Note that this is going to fail CI because the it only fixes errors, not warnings:~

~https://gist.github.com/wincent/3afc7cacaf90699d61f7553a2884577d~

~`yarn lint --quiet` (which only shows errors, not warnings), is clean.~

~Putting this up to ask for opinions on what we want to do about the warnings: @matuzalemsteles, @diegonvs, @bryceosterhaus — do you want to suppress them or tackle them?~

**Update:** Ok, so I sidestepped most of the thorny bits by turning off "react-hooks/exhaustive-deps" (which wasn't enabled before); we can look at adding it back when/if we're ready to do the work of sorting through all the issues that it highlights (either with fixes, refactorings, or suppressions).


next.clayui.com/gatsby/createPages.js
```diff
@@ -6,13 +6,17 @@
...
+/* eslint-disable liferay/imports-first */
+
 require('dotenv').config({
 	path: `.env.${process.env.NODE_ENV}`,
 });
 
 const componentWithMDXScope = require('gatsby-mdx/component-with-mdx-scope');
 const path = require('path');
 
+/* eslint-enable liferay/imports-first */
```
wincent: This is because `dotenv` is a special beast; it is recommended to call `config()` on it as early as possible in the app boot-up sequence, but we have lints against putting imports after non-import statements (and the `config()` call is considered a non-import statement), so I added this suppression.

I don't know if we even need dotenv though because I don't think anybody is creating `.env` files, right?

next.clayui.com/plugins/gatsby-remark-api-table/index.js
```diff
@@ -68,10 +69,12 @@ module.exports = ({markdownAST}) => {
 						});
 						const propsKeys = Object.keys(AST.props);
 
+						// eslint-disable-next-line require-atomic-updates
 						node.value = `<ul class="list-api">${propsKeys
 							.map(key => generateTr(AST.props[key], key))
 							.join('')}</ul>`;
 					} catch (error) {
+						// eslint-disable-next-line require-atomic-updates
```
wincent: Will want to evaluate these suppressions to see if they are ok or not.

packages/clay-color-picker/src/GradientSelector.tsx
```diff
@@ -3,10 +3,12 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
-import React, {useEffect, useRef} from 'react';
+
+import React, {useCallback, useEffect, useRef} from 'react';
```
wincent: There are several examples like this where we were using `useCallback` without importing it. So there must be a gap in our test coverage somewhere if this kind of breakage is going undetected.

matuzalemsteles: Hmm, this is weird. We don't use `useCallback` much. Maybe it's after the rules that were applied by adding `useCallback` in some functions. The rule does not matter `useCallback` when this happens.

next.clayui.com/gatsby/createPages.js
matuzalemsteles: This will load the env: `.env.development` and `.env.production`.

packages/clay-color-picker/src/GradientSelector.tsx
wincent: Ah yes, I didn't notice that the `useCallback` calls were added automatically by the autofix. Interesting.

next.clayui.com/gatsby/createPages.js
wincent: Yeah, I know what "dotenv" does — I was just asking, does anyone on our team (or any tool that we're using) — actually use those files?

wincent: And looking on the filesystem, I know see that we do have those files in a couple places; when I first looked for them I must have been looking in the repo root instead of under "next.clayui.com/".

packages/clay-charts/src/BillboardWrapper.tsx
```diff
@@ -19,29 +19,32 @@ const BillboardWrapper: React.FunctionComponent<IProps> = ({
...
+					});
+				}
 
-			forwardRef.current.load(data);
-		}
-	}, []);
+				forwardRef.current.load(data);
+			}
+		},
+		[forwardRef]
```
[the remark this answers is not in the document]
wincent: Not sure if it is a good idea or not; this change (and others like it) was made by the lint autofix rule. I think that technically, it is conservative, so only applies changes that "should" be made (for some definition of "should" 😂).

If you don't want these added, we will have to suppress that rule, but the rule does exist for a reason, so there's risk in that doing that.

packages/clay-charts/src/GeoMap.tsx
```diff
@@ -255,7 +256,7 @@ const Geomap: React.FunctionComponent<IProps> = ({
 		return () => {
 			forwardRef.current.disposed();
 		};
-	}, []);
+	}, [data, forwardRef, otherProps]);
```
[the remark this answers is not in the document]
wincent: Well these are all automated fixes designed to help you follow the "[Rules of Hooks](https://reactjs.org/docs/hooks-rules.html)" — if you don't want to follow the rules, we either have to refactor to avoid triggering them, or suppress them, in which case we'd need a strong rationale for why we think the rules don't apply.

packages/clay-data-provider/src/useResource.ts
```diff
@@ -4,18 +4,19 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {useCallback, useDebounce} from '@clayui/shared';
```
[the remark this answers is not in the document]
wincent: Good catch!

packages/clay-date-picker/src/InputDate.tsx
```diff
@@ -48,12 +48,12 @@ const ClayDatePickerInputDate: FunctionComponent<IProps> = ({
 	};
 
 	const memoizedValue = useMemo(() => isValidValue(value), [
+		isValidValue,
 		value,
-		currentTime,
```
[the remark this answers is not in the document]
wincent: Given that there is so much conflict between the "safe" autofixes and the patterns we have here, I was wondering if I could turn down "react-hooks/exhaustive-deps" from "error" to "warning" — but then I looked and saw it already is a "warning", and of course that doesn't stop it from autofixing anyway.

next.clayui.com/gatsby/createPages.js
matuzalemsteles: We have these two files at `next.clayui.com/`, to define a "Nightly"environment that prevents creating pages for production and only being enabled in the dev environment, it was a feature that was doing to migrate all Patrick docs to the clayui.com website but without some being visible in the production. But unsuccessfully getting Patrick out of there 😅

wincent:
Ok @matuzalemsteles, here's what I did:

- Given that the react-hooks/exhaustive-deps rule is so problematic, turned it off and undid *all* the changes in https://github.com/liferay/clay/pull/2536/commits/e92341a5a371cd8a49ca5e1e69273dcfc121e77d
- An autofixes on that clean slate, producing https://github.com/liferay/clay/pull/2536/commits/34935c103003e66ba8ad1e41b5044a7d74ed6e17
- Did the few (27) remaining not-autofixable lint fixes in https://github.com/liferay/clay/pull/2536/commits/10149ddbff0a12c0beea43450051393e781fe3a6 — the problems manually fixed are listed here: https://gist.github.com/wincent/62be5797b9bc40eff964fc11d7007160

wincent:
Looking into the issue with TS not recognizing the types on the "d3" module now.

matuzalemsteles:
hey @wincent, thanks for those changes. In summary I agree that we should follow the guidelines of the `react-hooks/exhaustive-deps` rule, I will keep some modifications that I found necessary for this rule. But there are some that we would need to review more carefully... Thanks again 😁

wincent:
Looks like it's green now.

wincent:
@matuzalemsteles I'll leave it up to you to decide when is best to merge this, as it may conflict with other PRs in flight.

PR #2537, opened 2019-09-24
feat(@clayui/card): add indeterminate and disabled state for high-level components

### LINKED ISSUE #2442: Verify clay-card for 3.0.0 release
- [x] Are you able to easily implement a demo of this component? Are there any “gotchas” when implementing? Provide a link to codesandbox for a demo.
- [ ] Does the component accept additional HTML attributes? For example, adding an aria attribute to a button.
- [ ] Does component accept a ref? Is the ref value what you would expect?
- [x] Does this component allow for different translations?
- [ ] If this is a form element, does it accept a `name` attribute?
- [x] Were you able to find documentation for this component? Was it clear?

Additional Questions:
- Is there anything about the public API that you would change?
- Is the naming of the available props clear?

matuzalemsteles:
Fixes #2442

### DIFF
--- a/packages/clay-card/src/CardWithHorizontal.tsx
+++ b/packages/clay-card/src/CardWithHorizontal.tsx
@@ -14,6 +14,12 @@ import ClayCard from './Card';
 
 interface IProps {
 	actions?: React.ComponentProps<typeof ClayDropDownWithItems>['items'];
+
+	/**
+	 * Flag to indicate that all interactions on the card will be disabled.
+	 */
+	disabled?: boolean;
+
 	/**
 	 * Path or URL to item
 	 */
@@ -47,6 +53,7 @@ interface IProps {
 
 export const ClayCardWithHorizontal: React.FunctionComponent<IProps> = ({
 	actions,
+	disabled,
 	href,
 	onSelectChange,
 	selected = false,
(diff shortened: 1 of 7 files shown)


matuzalemsteles:
Hm yeah, I'm not sure about the cases for the indeterminate ClayCard, probably when a card comes pre-selected. I am adding this possibility if it appears but we can go without it if we think this is overkill.

bryceosterhaus:
Lets see what the lexicon team thinks, and we can go with whatever they decide. @drakonux what do you think about this? Is there a use case for having an indeterminate state for a single selectable item such as a card?

drakonux:
Ei guys! For us the indeterminate state should be used just when an element that can be checked/unchecked directs other components that can be checked/unchecked (e.g. a checkbox directing other checkboxes - the management toolbar checkbox, a tree view node). 

Thus I can't say an actual case where a card directs other cards 😕

matuzalemsteles:
Thanks @drakonux. I think we can remove this state from the Card component and release without it.

bryceosterhaus:
Thanks for the clarification @drakonux!

matuzalemsteles:
@bryceosterhaus updated, removed indeterminate state and cleaned commits.

Issue #2540, opened 2019-09-25
Use Button instead of anchor in Vertical component


matuzalemsteles:
Attempting to add the button when the element state in Vertical is not necessarily a link is just a trigger to expand the menu, breaks the element because it does not support the button.

The ideal would be to use a button for this case to allow interaction via keyboard, we could use the anchor with `href="javascript:;"` but we will have problems with the React warnings and we want to move away from this pattern.

<img width="259" alt="Screen Shot 2019-09-25 at 11 54 43" src="https://user-images.githubusercontent.com/13750819/65612802-4904a480-df8b-11e9-9ed6-a0aa3ad44acf.png">

This is a button with class `.btn.btn-unstyled`.

PR #2541, opened 2019-09-25
chore(@clayui/navigation): improve types

matuzalemsteles:
This is just a subtle improvement, since we allow `otherProps` in this component we must say what other props can be passed to it.


matuzalemsteles:
This is a small change to go ahead with release of this package. Depend on the changes from issue #2540 so we can do it.

Issue #2474, opened 2019-09-17
Provide real information in How to Use Clay page


jbalsas:
## What are the steps to reproduce?

Navigate to https://next.clayui.com/docs/get-started/how-to-use-clay.html

## What is the expected result?

See some real information about how to use Clay rather than our current placeholder TODOs

eduardoallegrini:
https://github.com/liferay/clay/issues/2209

Issue #2478, opened 2019-09-17
Basic guidelines to test Clay components


eduardoallegrini:
Hi guys, I was trying to take a look at the Clay components https://github.com/liferay/clay/milestone/12 and I think the whole process could be confusing if you are not a js-developer

I don’t know if we already have these guidelines somewhere or if these are even correct, I had to guess and read online. So I was wondering, should we add them to our documentation?

## Getting started
> install/update node/npm/react/yarn

I know it’s a basic step but still we should add it to the documentation
> `npx create-react-app [app-name]`

I think this is optional but if you just want to try the component it’s a good start
> `cd [app-name]`

basic, suggested by react but still a good tip
> `yarn add @clayui/[selected-module]`

if you don't have experience with yarn you won't think of this
`npm -i [selected-module]` installs an older version
> `yarn add @clayui/css`

this is a good practice I copy from [bryce’s example](https://codesandbox.io/s/sad-carson-thluj) but it’s not mentioned anywhere
> `yarn start`

basic, suggested by react but still a good tip

## In the App
If you are using the app created by the `npx create-react-app [app-name]`
> Add the example code provided by the Clay website into App.js or index.js

basic, suggested by react but still a good tip
> Add `import "@clayui/css/lib/css/atlas.css";` to your code, after `import React from 'react';`

this is a good practice I copy from bryce’s example but it’s not mentioned anywhere
> Add `import ClayIcon from "@clayui/icon";`to your code, after `import React from 'react';`

this is a good practice I copy from bryce’s example but it’s not mentioned anywhere

## Components
Is there any reason why we are not using a prefix for the API options?

Example: in the [alert component](https://next.clayui.com/docs/components/alerts.html) we are using `title`, `variant`, `spritemap`, etc. shouldn’t we be using `clayTitle`, `clayVariant` or `claySpritemap`?

marcoscv-work:
For the components part:

We have some confusion when we use component attributes what actually are for html like `title`, for example: 
- In alert component `title=""'` is the title of the alert, this part:
  - ![image](https://user-images.githubusercontent.com/7963804/65045099-7a172080-d95e-11e9-88f7-a4e25bac6cdf.png)
- In icon component title is the HTML attribute:

As @eduardoallegrini has mentioned, we could use `clayTitle` and it would be abstracted from the HTML attribute; `title`

carloslancha:
[...]

Issue #2209, opened 2019-07-23
[next.clayui.com] Improve the get-started content


eduardoallegrini:
A few weeks ago I started to change the get-started content but I didn't have time/knowledge to finish it. I wrote down some general points but I think we (someone) should change and improve it.

PR #2543, opened 2019-09-25
chore(next.clayui.com): add guide to install components and quick start

matuzalemsteles:
This will most likely fix issues #2474, #2209, and #2478

I have added leaner documentation to the "How to use clay" document, if we need more information we can add it elsewhere.

I just added the document I'll work on tomorrow for "Composing" to explain how composing works for Clay components and low-level and high-level philosophy.

### DIFF
--- a/next.clayui.com/content/docs/get-started/how-to-use-clay.md
+++ b/next.clayui.com/content/docs/get-started/how-to-use-clay.md
@@ -27,7 +27,7 @@ Clay follows some fundamentals and we recommend that you read more about this be
 ### Install with NPM or Yarn
 
 
-Clay makes the components and CSS available in its own `@clayui` scope package, in some packages we have component aggregation, something like the `@clayui/form` package that aggregates components `Checkbox`, `Radio`, `Input`, `Select`...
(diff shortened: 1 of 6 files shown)


next.clayui.com/content/docs/get-started/how-to-use-clay.md
```diff
@@ -1,47 +1,186 @@
...
+<div class="nav-toc">
+
+- [Composing](/foundations/composing.html)
+
+</div>
+
+### Install with NPM or Yarn
+
+
+Clay makes the components and CSS available in its own `@clayui` scope package, in some packages we have component aggregation, something like the `@clayui/form` package that aggregates components `Checkbox`, `Radio`, `Input`, `Select`...
```
bryceosterhaus:
```suggestion
Clay makes the components and CSS available in its own `@clayui` scope, for example the card package is available through `@clayui/card` and the css is available through `@clayui/css`. In some packages we expose multiple components, for example the `@clayui/form` package contains components `Checkbox`, `Radio`, `Input`, `Select`...
```

PR #2549, opened 2019-09-30
chore(next.clayui.com): add ClayCard docs, add code snippets imports and general improvements 

matuzalemsteles:


next.clayui.com/content/docs/components/card.mdx
```diff
@@ -31,10 +48,67 @@ import {Card} from '../../../src/components/clay/Card';
...
+
+For some low-level components it is necessary to pass some modifier classes in order for the component to reach the Lexicon result, this was necessary since we did not want to attach the component just for this case and the low-level components could be used more flexibly and allow you to create new possibilities.
 
 <Card />
 
+## High-level Components
+
+ClayCard's high-level components predict the major cases of the Lexicon specification:
+
+- Cards with Imagens
```
bryceosterhaus: should be "Images"

matuzalemsteles:
> One thought I had while looking through this, should we use prettier for our code examples? Right now they seem to break the 80character limit.

Hm yeah 🤔, I agree now I don't know if we can run the prettier on literal template strings, or are you saying it somewhere else?

bryceosterhaus:
@matuzalemsteles oh true, I forgot those were in template literals. Hmm, maybe we could investigate if there is some sort of plugin we could utilize to help for this.

wincent:
> I forgot those were in template literals. Hmm, maybe we could investigate if there is some sort of plugin we could utilize to help for this.

I don't know if there are any existing plugins, but that `<Editor />` API isn't the friendliest. Would you rather write this?

```javascript
import ClayAlert from '@clayui/alert';
import React, {useState} from 'react';

import Editor from './Editor';

const spritemap = '/images/icons/icons.svg';

const AlertCode = `const Component = () => {
	return (
		<>
			<ClayAlert displayType="info" spritemap={spritemap} title="Info">
				This is a default alert
			</ClayAlert>

			<ClayAlert displayType="warning" spritemap={spritemap} title="Warning" variant="stripe">
				This is a stripe alert
			</ClayAlert>
		</>
	);
}

render(<Component />);`;

const alertImportsCode = `import React from 'react';
import ClayAlert from '@clayui/alert';`;

export const Alert = () => {
	const scope = {ClayAlert, spritemap, useState};
	const code = AlertCode;

	return <Editor code={code} imports={alertImportsCode} scope={scope} />;
};
```
or this?

```javascript
import ClayAlert from '@clayui/alert';
import React from 'react';

export const Alert = () => {
    const scope = {ClayAlert, spritemap};
    const code = AlertCode;

    const spritemap = '/images/icons/icons.svg';

    return (
        <Editor scope={scope}>
            <Imports>import React from 'react'; import ClayAlert from '@clayui/alert';</Imports>
            <Code>
                {({ClayAlert, spritemap}) => {
                    const Component = () => {
                        return (
                            <>
                                <ClayAlert displayType="info" spritemap={spritemap} title="Info">
                                    This is a default alert
                                </ClayAlert>

                                <ClayAlert displayType="warning" spritemap={spritemap} title="Warning" variant="stripe">
                                    This is a stripe alert
                                </ClayAlert>
                            </>
                        );
                    };

                    render(<Component />);
                }}
            </Code>
        </Editor>
    );
};
```

Doesn't seem like it would be hard to change it to work like that, and because it doesn't use template literals, you don't need to do anything special to get Prettier to format it (and you actually get syntax highlighting and code completion while editing them.)

wincent:
The other thing I notice about these examples is that they hide what `spritemap` is and where it comes from, which means you can't copy-paste them and expect them to work. :-(

matuzalemsteles:
hey @wincent, Yeah, actually I'm thinking of removing some of it all and creating a plugin for gatsby-mdx so we can add code snippets to mdx and use the `<Editor />` underneath.

> The other thing I notice about these examples is that they hide what spritemap is and where it comes from, which means you can't copy-paste them and expect them to work. :-(

Yeah, I think it might be better to put in snippets even if it's a repeat.

bryceosterhaus:
Great ideas for both the API for `<Editor />` and including spritemap. Im all for both.

PR #2552, opened 2019-09-30
feat(@clayui/autocomplete): add high-level Autocomplete component

matuzalemsteles:
@bryceosterhaus I am adding the high-level Autocomplete back to the `@clayui/autocomplete` package, we just have to decide if we are going to release this package with `minor` or `patch`, which is likely to be a `minor` (v3.1.0) since It is a new component but maintains compatibility.

### DIFF
--- a/packages/clay-autocomplete/package.json
+++ b/packages/clay-autocomplete/package.json
@@ -29,11 +29,11 @@
 		"@clayui/drop-down": "^3.0.0",
 		"@clayui/form": "^3.0.0",
 		"@clayui/loading-indicator": "^3.0.0",
+		"@clayui/shared": "^3.0.1",
 		"fuzzy": "^0.1.3"
 	},
 	"devDependencies": {
-		"@clayui/data-provider": "^3.0.0",
-		"@clayui/shared": "^3.0.1"
+		"@clayui/data-provider": "^3.0.0"
 	},
 	"peerDependencies": {
 		"@clayui/css": "3.x",
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -1,0 +1,85 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ClayInput} from '@clayui/form';
+import React, {useRef, useState} from 'react';
+
+import Context from './Context';
+import DropDown from './DropDown';
+import Input from './Input';
+import Item from './Item';
+import LoadingIndicator from './LoadingIndicator';
+
+const AutocompleteMarkup = React.forwardRef<
+	HTMLDivElement,
+	React.HTMLAttributes<HTMLDivElement>
+>(({children, ...otherProps}, ref) => (
+	<ClayInput.Group {...otherProps} ref={ref}>
+		<ClayInput.GroupItem>{children}</ClayInput.GroupItem>
+	</ClayInput.Group>
+));
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Div component to render. It can be a one component that will replace the markup.
+	 */
+	component?: React.ForwardRefExoticComponent<any>;
+}
+
+type Autocomplete = React.ForwardRefExoticComponent<IProps> & {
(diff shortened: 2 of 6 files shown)


bryceosterhaus:
One thought, do we know of anyone who is using this or needs this right now at the moment? We may be able to get away with just showing an example in our storybook for now and then adapting to how people start to use it. This way we an keep a focus on low-level right now, and slowly add high-level when its needed rather than over provide high-level out of the gate.

If we do want to release it sooner than later though, minor would make sense to me as well, unless @jbalsas has any objections.

matuzalemsteles:
> One thought, do we know of anyone who is using this or needs this right now at the moment? We may be able to get away with just showing an example in our storybook for now and then adapting to how people start to use it. This way we an keep a focus on low-level right now, and slowly add high-level when its needed rather than over provide high-level out of the gate.

Hmm, In the portal there are some cases but I don't remember which modules, but I think this high-level is very simple, I think we can use it to model things than create new components when we have demand... we can think better about this...

bryceosterhaus:
I think I would err on the side of using storybook to model what is capable, however if we have concrete use cases or multiple users needing this exact functionality then I think we could ship as a high-level. However, to be safe, I'm thinking it might be best to first put high-level examples in storybook and then once it is used by teams, we can start to evaluate the best way to move it into high-level. Would that make sense to you?

This is primarily to help protect us from releasing too much at once before it is used. And plus they should already have the available low-level components to try it out. This might give is a good way to iterate on feedback before committing to a certain high-level implementation.

matuzalemsteles:
> I think I would err on the side of using storybook to model what is capable, however if we have concrete use cases or multiple users needing this exact functionality then I think we could ship as a high-level. However, to be safe, I'm thinking it might be best to first put high-level examples in storybook and then once it is used by teams, we can start to evaluate the best way to move it into high-level. Would that make sense to you?

Yes it makes sense, the concern I have is that we fail to launch things that can help teams be productive, we will have to take a little risk to know what will work and what will not work. I like the idea of ​​trying to get it wrong as soon as possible (be it API design decisions ...), so we can iterate faster, if we take too long to realize these changes and only include it later, we can risk making a mistake. But is too late.

wincent:
> I like the idea of ​​trying to get it wrong as soon as possible

The trouble with that is that we can't make breaking API changes once things are shipped to liferay-portal (because for marketing reasons, we can't do a major version bump in between yearly product cycles). I agree with @bryceosterhaus that modeling informally via Storybook is the most prudent approach here. Adding new API isn't just adding something potentially positive to the library; it's also adding a potential (negative) liability — the sooner you rush to deliver that positive value, the more likely you'll end up dragging some of the negative along with it.

matuzalemsteles:
hey @wincent, you're right, but allowing us to do just one major release a year can be bad for doing things, we need quick interactions, I'm worried about moving slowly to implement changes.

At any moment the Lexicon team can change something or suggest new changes, we can't get stuck in the main version once a year... I understand that would be ideal, but Clay ends up being governed by other things that we have no control over. I understand that we should not do major releases 3, 4 times a year 😂, but the idea is that we should not be stuck with this rule either.

I have in mind that we should make smoother major releases without necessarily spending a lot of effort developing months of a major release, it's a bit painful, it can happen again of course.

> I agree with @bryceosterhaus that modeling informally via Storybook is the most prudent approach here. Adding new API isn't just adding something potentially positive to the library; it's also adding a potential (negative) liability — the sooner you rush to deliver that positive value, the more likely you'll end up dragging some of the negative along with it.

Yeah, we would be making this switch when delivery is fast, if we are not sure about delivering an API the ideal would be to release a beta or RC on this package, I have in mind that we should not wait for others to implement, we should show to how to use or disseminate good practice. We are not going here with something relatively complex like the Management Toolbar... so I would be fine to deliver this component in a `3.1.0-beta` version for testing but if we are very unsure about it, we can go with this approach.

wincent:
> but Clay ends up being governed by other things that we have no control over

That's exactly the point: these are parameters *that we have no control over*. We have to operate within the constraints that we're given; where we don't like those constraints, we still have to abide by them, although we can obviously seek to influence what those constraints are via the appropriate channels (eg. @jbalsas).

jbalsas:
Hey everyone! I'm working on collecting my thoughts and creating a draft with some guiding principles for Clay and how to operate. I'll send it before EOD today

Issue #2554, opened 2019-10-01
Investigate tooling for better code snippets in out docs.


bryceosterhaus:
Conversation at https://github.com/liferay/clay/pull/2549#issuecomment-536912114

bryceosterhaus:
Temporary solution: https://github.com/liferay/clay/pull/2567

We should still investigate this further to avoid having to make template literals of our code examples.

bryceosterhaus:
Going to just continue with #2567

PR #2567, opened 2019-10-01
docs: add prettier for react-live editor

bryceosterhaus:
In regards to https://github.com/liferay/clay/issues/2554

This was the quickest way to keep formatting consistent in the docs. I think we can later iterate on this, but I think this provides a quick and helpful formatting for our examples that utilize react-live

### DIFF
--- a/next.clayui.com/package.json
+++ b/next.clayui.com/package.json
@@ -63,6 +63,7 @@
 		"metal-dom": "^2.16.5",
 		"metal-tabs": "^2.3.1",
 		"node-sass": "^4.11.0",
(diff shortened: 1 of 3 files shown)


bryceosterhaus:
@matuzalemsteles i had to add a try/catch, this will help us catch any templates that aren't formatted correctly.

This is obviously a temporary fix for now, ideally we would want something along the lines that greg outlined. I thought this was most appropriate since its the simplest for the most effect.

matuzalemsteles:
@bryceosterhaus np. We can go ahead with this and then we can see a better solution for that.

PR #2572, opened 2019-10-01
docs(popover/tooltip): add docs

bryceosterhaus:


bryceosterhaus:
@matuzalemsteles any idea how to get more helpful errors from gatsby? this seems to work fine for me locally

matuzalemsteles:

matuzalemsteles:
@bryceosterhaus at build time is pretty bad to know what the error was, so they usually always recommend checking out third party components, in this case it would be `Tooltip`, `TooltipProvider` and `Popover`. Probably some of these components try to directly access `window` or something related to the DOM.

I took a quick look and in `TooltipProvider` it uses `window.setTimeout`, we can just go with `setTimeout`, had you added this with any concerns? I did a local test, but this is not the root cause of the problem.

bryceosterhaus:
@matuzalemsteles i used window since TS was inferring setTimeout to be from `NodeJS.Timeout`. I removed it for now and added a note to the type definition, lets see if this works.

matuzalemsteles:
I see, it is strange this error:
```shell
Component AlertToastContainer was not imported, exported
```

bryceosterhaus:
@matuzalemsteles just found a workaround for this. let me know if you think this might be too hacky lol

next.clayui.com/src/components/clay/Tooltip.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+
+import Editor from './Editor';
+
+// Workaround for building gatsby's SSR with Tooltip
+const {ClayTooltipProvider, default: ClayTooltip} =
```
matuzalemsteles: Yeah, that sounds weird but if there is no other solution 🤷‍♂️. Maybe we can just ignore the `metal-position` in `onCreateWebpackConfig.js`, could that solve?

```js
{
	loader: 'null-loader',
	test: /clay-tooltip|metal-clipboard|metal-position/,
},
```

bryceosterhaus: I tried that and it didn't work, it just caused another error where a component was calling `Align`

PR #2575, opened 2019-10-02
feat(next.clayui.com): Automate version and status labels

diegonvs:


next.clayui.com/src/utils/getPackageConfig.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {readFileSync} = require('fs');
+const {join, resolve} = require('path');
+
+function getPackageConfig(name) {
```
matuzalemsteles: hey @diegonvs this may break with a recent change in master, there is the case of Popover and Tooltip that `packageNpm` has two packages, so as not to break this we can separate the documentation.

diegonvs: Nice, I broke `Popover and Tooltip` into separated ones.

next.clayui.com/content/docs/components/css-tooltips.md
```diff
@@ -0,0 +1,202 @@
+---
+title: "Tooltips"
+description: "Popovers and Tooltips are patterns used to show helpful information in two different ways."
```
bryceosterhaus: Remove reference to popovers here

next.clayui.com/content/docs/components/popover.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: 'Popover'
+description: 'Popovers are used together with the question mark icon to communicate to the user that there is useful information for them to read. Popovers contain helpful information such as an explanation of a context.'
```
bryceosterhaus: > are used together with the question mark icon

Since we don't actually include that functionality in the component, I think it might be best to leave that out. We can point to lexicon for best practices and use cases, but I dont think we should express what item to align it to.

next.clayui.com/src/utils/getPackageConfig.js
```diff
@@ -0,0 +1,23 @@
...
+	const internalPackagePrefix = '@clayui/';
```
bryceosterhaus: We can move this line outside of the function since its a constant.

next.clayui.com/src/utils/getPackageConfig.js
```diff
@@ -0,0 +1,23 @@
...
+	if (name && name.startsWith(internalPackagePrefix)) {
+		const suffix = name.slice(internalPackagePrefix.length);
+		return readFileSync(
```
bryceosterhaus: New line above this return. We need to make sure we have new lines between any non-sortable content.

next.clayui.com/content/docs/components/popover.mdx
diegonvs: Makes sense :)

PR #2584, opened 2019-10-03
feat(@clayui/form): Adds missing size properties on Form and Input component

diegonvs:


packages/clay-form/src/Input.tsx
```diff
@@ -143,6 +145,11 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	 * Flag to indicate if `input-group-inset-before` class should be applied
 	 */
 	insetBefore?: boolean;
+
+	/**
+	 * Selects the height of the input.
+	 */
+	sizing?: TSize;
```
bryceosterhaus: Since TSize is only used once, we can skip that type declaration and directly put lg and sm here

matuzalemsteles: I would also choose to rename it to `size` to maintain consistency with other APIs.

diegonvs: * `size` is an attribute of `input`: https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/input

diegonvs: > Since TSize is only used once, we can skip that type declaration and directly put lg and sm here

Fixed! I was getting a strange error during the development but I solved it!

PR #2587, opened 2019-10-03
fix(@clayui/form): remove fixed value from Checkbox

### LINKED ISSUE #2573: Checkboxes at the moment only handle true/false value
At the moment checkboxes don't handle correctly values.

As you can see here: https://codesandbox.io/s/green-moon-ctnek the `value` attribute changes based on the `checked` one, while it should not be related to it.

I should be able to pass an ID as a value and being able to read it directly on the rendered view like:
```<input type="checkbox" name="assets" value="images" checked>```

### PR DESCRIPTION
**Pull request by matuzalemsteles**
Fixes #2573

I'm removing the trapped value that I was giving to Checkbox and letting people handle it. It's strange why we left the value fixed, I don't remember why.

### DIFF
--- a/packages/clay-card/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-card/src/__tests__/__snapshots__/index.tsx.snap
@@ -172,7 +172,6 @@ exports[`ClayCard renders a ClayCard as a selectable file card 1`] = `
             <input
               class="custom-control-input"
               type="checkbox"
-              value="false"
             />
             <span
               class="custom-control-label"
@@ -275,7 +274,6 @@ exports[`ClayCard renders a ClayCard as a selectable folder 1`] = `
         <input
           class="custom-control-input"
           type="checkbox"
-          value="false"
         />
         <span
           class="custom-control-label"
@@ -354,7 +352,6 @@ exports[`ClayCard renders a ClayCard as a selectable image card 1`] = `
             <input
               class="custom-control-input"
               type="checkbox"
-              value="false"
             />
             <span
               class="custom-control-label"
@@ -844,7 +841,6 @@ exports[`ClayCard renders a ClayCard as user selectable card 1`] = `
             <input
               class="custom-control-input"
               type="checkbox"
-              value="false"
             />
             <span
               class="custom-control-label"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by bryceosterhaus** [comment]
And just for background, I hadn't even used a checkbox value for anything other than true/false for apps I have made, so I'm not totally sure how people expect it to work.

**Comment by FabioDiegoMastrorilli** [comment]
@bryceosterhaus `onChange` is perfect if we want to handle the state via js but, sometime, we might use react just to render a form that will be normally submitted without any async request.

In that case I would expect to find all the values in the html so, the server, will be aware about the form state. We could use `onChange` to change values inside of a hidden input but that would be more a workaround.

**Comment by jbalsas** [comment]
@bryceosterhaus, `value` in checkboxes is the standard in html forms. It is what gets sent to the server when you submit a form.

Take this piece of code:

```html
<form action="/action">
    <ul>
        <li class="item">
            <input name="items" type="checkbox" value="id_product1" />
            <img src="/images/product1.png" />
        </li>
        <li class="item">
            <input name="items" type="checkbox" value="id_product2" />
            <img src="/images/product2.png" />
        </li>
        <li class="item">
            <input name="items" type="checkbox" value="id_product3" />
            <img src="/images/product3.png" />
        </li>
    </ul>

    <input type="submit">Add to cart</input>
</form>
```

When the submit button is pressed, only the values in the checked inputs will be submit. Assuming products 1 and 3 are checked, the server would see a request such as:

```
items=product1
items=product2
```

Which, in server-code gets received as an `items` array so they know which items to process.

matuzalemsteles:
Thanks @jbalsas and @FabioDiegoMastrorilli for bringing in these use cases.

@bryceosterhaus I think the @jbalsas case ends up preventing us from adding a default behavior to the checkbox, it can break that behavior and prevent the backend from making changes to just support it...

bryceosterhaus:
Makes sense to me that we need to allow that functionality. My thought is if the internal logic should be something like this...

```jsx
<input
	value={
		isDefined(value) ? value : (checked ? 'true' : 'false')
	}
/>
```

I guess its not super necessary, but for some reason it feels _more right_ to me. But now im starting to doubt my basic html knowledge for this haha

matuzalemsteles:
haha yeah, I think this ends up being eliminated by the default browser behavior as it will only send the checked checkboxes to the server.

matuzalemsteles:
@bryceosterhaus are you ok to go with this? we can go ahead.

bryceosterhaus:
@matuzalemsteles  lgtm

Issue #2553, opened 2019-10-01
ClayModal, how to close it properly


julien:
I've got a question regarding the new `@clayui/modal` component.

I've set up an example in codesandbox [here](https://codesandbox.io/embed/claymodalexample-koiiy?fontsize=15) and I'd like to know what I doing wrong.

If you look at the example and click on the "Close Me" button, you'll notice that a warning is displayed in the console. 

Note that, this warning only happens **once**, even if you have multiple modal components in your "app".

This wasn't happening before `3.0.0`.

 I also commented [this](https://github.com/liferay/clay/blob/d0a2e3659a8b0f56f02bc3a2ca9fc2c35ccfe6ad/packages/clay-modal/src/useModal.tsx#L39) line (in my local `node_modules` directory) and the warning disapears.

Any help is appreciated, thanks.

diegonvs:
Yes, this is weird. Needs more investigation but We found a temporary walkaround. If you move `useModal` from the Modal component to your App component this error doesn't appear more. https://www.diffchecker.com/3AXpWjOT

bryceosterhaus:
First off, thanks for the codesandbox, always a huge help.

Hmm I think it might have to do with the `useModal` hook being in a component that is unmounted at the same time as the modal itself. I refactored it slightly [here](https://codesandbox.io/embed/claymodalexample-l6pjj?fontsize=14) so that `useModal` exists outside of that mounted component, and it doesn't seem to error.

However, this does seem like a bug. I would have thought that line wouldn't even run since it should be cleared when the hook is unmounted here at this line. https://github.com/liferay/clay/blob/d0a2e3659a8b0f56f02bc3a2ca9fc2c35ccfe6ad/packages/clay-modal/src/useModal.tsx#L64

I'll label this as an issue and we can look into it.

bryceosterhaus:
looks like me and @diegonvs were both looking at the same time lol

matuzalemsteles:
hey @bryceosterhaus, @diegonvs yeah I think we have to also analyze that even correcting this, if this case would be the most recommending to close the modal. The hook may be lost every time the component is unmounted at the same time as Modal, due to the use of the above ternary in the example. For me it makes more sense for the hook to be out of ternary since it is Modal's controller, it needs to be listening to Modal's changes since it is `observer`.

julien:
Thanks to all for the quick reply.

[...]

PR #2591, opened 2019-10-07
chore(next.clayui.com): add more docs on using Modal

### LINKED ISSUE #2560: Review Modal documentation

matuzalemsteles:
Fixes #2560

I didn't even add good practice descriptions to create tests with Modal. As I said here #2553. I'll take care of adding later as soon as I have more use cases.

### DIFF
--- a/next.clayui.com/content/docs/components/modal.mdx
+++ b/next.clayui.com/content/docs/components/modal.mdx
@@ -38,7 +38,7 @@ The `useModal` [hook](https://reactjs.org/docs/hooks-intro.html) is required to 
 
 `useModal` returns a signature with `observer` and `onClose`:
 
-- **observer** - Observer is the communication channel that connects ClayModal with `useModal`, you can call it a tunnel.
+- **observer** - Observer is the communication channel that connects ClayModal with `useModal`.
 - **onClose** - Callback is called after Modal's close animation.
 
 ```jsx
@@ -51,7 +51,7 @@ const { observer, onClose } = useModal({
     To avoid future problems, do not use the information or change the values ​​of `observer`, it is considered a bad practice.
 </div>
 
-To prevent Modal from closing before animation never set its state directly in button events to close the modal instead call the `onClose` method returned by `useModal`.
+To prevent Modal from closing before animation always use the `onClose` method returned by `useModal`.
 
 ```diff
 <ClayModal.Footer
@@ -62,10 +62,9 @@ To prevent Modal from closing before animation never set its state directly in b
 
 #### Usage
 
-`UseModal` is a hook so it implies that you cannot use ClayModal in a class component, to use you can adopt some strategies:
+`useModal` is a hook so it implies that you cannot use ClayModal directly in a class component. If you'd like to use it in a class component, try the recommended strategy below.
 
-- Wrap ClayModal and `useModal` into one component.
-- Create your modal expert on a functional component.
+- Wrap ClayModal and `useModal` in a function component.
 
 ```jsx
 const ModalWithDialog = ({visible, onClose, title, body}) => {
@@ -112,7 +111,7 @@ class Page extends Component {
 
 ### Provider
 
-When your application uses a lot of modals, it may be bad to be creating multiple modes for your application, you can use `<ClayModalProvider />` to have only one open Modal on the screen and the component centered on your application, you can invoke the modal from the context through the API.
+When your application uses a lot of modals, you can use `<ClayModalProvider />` to have only one open Modal on the screen and the component centered on your application, you can invoke the modal from the context through the API. It is not recommended to have multiple modals open at the same time.
 
 Add in the project root:
 
--- a/next.clayui.com/content/docs/components/modal.mdx
+++ b/next.clayui.com/content/docs/components/modal.mdx
@@ -12,6 +12,9 @@ import {Modal} from '../../../src/components/clay/Modal';
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
+- [useModal hook](#usemodal-hook)
+    - [Usage](#usage)
+- [Provider](#provider)
 - [API](#api)
   - [ClayModal](#claymodal)
   - [ClayModal.Body](#claymodal.body)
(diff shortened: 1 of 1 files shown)


next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
 
 <Modal />
 
+### useModal hook
+
+The `useModal` [hook](https://reactjs.org/docs/hooks-intro.html) is required to create a Modal component in Clay, it is responsible for communicating with ClayModal and controlling states to make animations, this is necessary because Modal needs animation when opening and closing so we need to delay calls of the `onClose` so you can safely control Modal's visibility or do other things.
+
+`useModal` returns a signature with `observer` and `onClose`:
+
+- **observer** - Observer is the communication channel that connects ClayModal with `useModal`, you can call it a tunnel.
```
bryceosterhaus: Not sure the "you can call it a tunnel." part is helpful, it might add to the confusion.

next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
...
+- **onClose** - Callback is called after Modal's close animation.
+
+```jsx
+const { observer, onClose } = useModal({
+    onClose: () => setVisible(false)
+});
+```
+
+<div class="clay-site-alert alert alert-warning">
+    To avoid future problems, do not use the information or change the values ​​of `observer`, it is considered a bad practice.
```
bryceosterhaus: > "do not use the information"

Is this accurate? Makes sense to discourage the changing of the values, but reading the values doesn't seem like a bad pattern.

next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
...
+</div>
+
+To prevent Modal from closing before animation never set its state directly in button events to close the modal instead call the `onClose` method returned by `useModal`.
```
bryceosterhaus:
```suggestion
To prevent Modal from closing before animation always use the `onClose` method returned by `useModal`.
```

I think this should be clear and info info to convey that we want them to use the hook.

next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
...
+```diff
+<ClayModal.Footer
+-   last={<ClayButton onClick={() => setVisible(false)}>{"Primary"}</ClayButton>}
++   last={<ClayButton onClick={onClick}>{"Primary"}</ClayButton>}
+/>
+```
+
+#### Usage
+
+`UseModal` is a hook so it implies that you cannot use ClayModal in a class component, to use you can adopt some strategies:
```
bryceosterhaus:
```suggestion
`UseModal` is a hook so it implies that you cannot use ClayModal directly in a class component. If you'd like to use it in a class component, try the recommended strategy below.
```

next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
...
+
+- Wrap ClayModal and `useModal` into one component.
```
bryceosterhaus:
```suggestion
- Wrap ClayModal and `useModal` in a function component.`
```

next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
...
+- Create your modal expert on a functional component.
```
bryceosterhaus: Not sure this makes sense, we can probably remove it and just include the one strategy with the code example.

next.clayui.com/content/docs/components/modal.mdx
```diff
@@ -29,6 +32,147 @@ You may want to compose your content and customize with your use cases or add yo
...
+        );
+    }
+}
+```
+
+`useModal` may be lost if you raise the visible condition above the hook, if Modal happens to be unmounted at the same time as the component. Ideally the hook is not unmounted every time Modal is opened and closed.
+
+### Provider
+
+When your application uses a lot of modals, it may be bad to be creating multiple modes for your application, you can use `<ClayModalProvider />` to have only one open Modal on the screen and the component centered on your application, you can invoke the modal from the context through the API.
```
bryceosterhaus:
```suggestion
When your application uses a lot of modals, you can use `<ClayModalProvider />` to have only one open Modal on the screen and the component centered on your application, you can invoke the modal from the context through the API. It is not recommended to have multiple modals open at the same time.
```

next.clayui.com/content/docs/components/modal.mdx
matuzalemsteles: Hmm yeah, I will remove it.

next.clayui.com/content/docs/components/modal.mdx
matuzalemsteles: I didn't want to encourage reading these values, because we can change the information contained in `observer`, so I've been thinking of providing an API to` useModal` to address the need to know when Modal is really visible after the animation is done.

next.clayui.com/content/docs/components/modal.mdx
matuzalemsteles: I will remove this and just leave the example below if we see more cases we can add them.

next.clayui.com/content/docs/components/modal.mdx
bryceosterhaus: yeah thats a good idea to add an API, I don't think we should dissuade people from reading the values unless we have another means of reading that information

matuzalemsteles: Yeah, we can add this and wait for the release of Clay 3.0 officially, I'll send a PR later adding this API. But still I would like to keep this information.

bryceosterhaus: sounds good. feel free to merge

julien:
Hey @matuzalemsteles

I personally think that as a `ClayModal` consumer,  I don't want to have to use the `useModal` hook each time I use the component. 
Ideally, I'd like to be able to pass an `onClose` prop (and eventually a `onShow`/`onVisible` prop).

Another important point about that issue was that if you declare the `useModal` hook inside the component that contains the modal and not it's parent you get the undesired warning that I mentioned previously.

I know that this is a lot of work (and easier said than done) and that @bryceosterhaus and you have a lot on your plates right now, but I think getting this component's API right will save a lot of time and make our users happier.

wincent:
About that warning, I don't know if we have an open ticket for it somewhere, but I suspect we still need to deal with it.

https://github.com/liferay/clay/blob/ca3b1cde6866a43bdfe73074ac6c1f84391944f0/packages/clay-modal/src/useModal.tsx#L61-L67

There may be a race condition in there, because it may possible to have two timers in flight (ie. the open *and* close one) but we only `clearTimeout` on the last-scheduled one, so if a dismount occurs (in the parent that uses `useModal`) then you'll see the warning.

matuzalemsteles:
hey @julien and @wincent we will still work on the bug reported in #2553.

> I personally think that as a ClayModal consumer, I don't want to have to use the useModal hook each time I use the component.

This is interesting, could you describe a little more than you wanted to do? I'm thinking we can give different `ClayModal` scattered and just one `useModal` hook that would control... would that be?

julien:
Hey @matuzalemsteles,

Concerning your question 
> This is interesting, could you describe a little more than you wanted to do?

I'd like to be able to do this for example (if it's technically possible of couse)

```jsx
<ClayModal onClose={handleClose} onShow={handleShow}>
   <ClayModal.Header>
      { /* header's children */ }
   </ClayModal.Header>
   <ClayModal.Body>
     { /* body's children */ }
   </ClayModal.Body>
   <ClayModal.Footer>
      { /* footer's children */ }
   </ClayModal.Footer>
</ClayModal>
``` 

I think the "consumers" of this component shouldn't have to worry about the `useModal` hook, if it's not possible it's not a problem, it's just a suggestion based on the feedback I've received from other teams and myself included. 

> Ideally each time your application grows with the use cases of having many Modals we recommend using Provider as it eliminates some of these things.

In our case, our application has already grown a lot and there are lots of modal windows being used, we even have modals that open other modals (:rofl:). I'll check the storybook to figure out how to use a provider, but will using a provider eliminate the need of using `useModal`?

matuzalemsteles:
@julien Oh, I see. Thanks for bringing this feedback 🙂.

Maybe this is possible not necessarily with these Modal-bound APIs, we can create a Modal wrapper over it and the hook to make it possible, but we would have to evaluate... You could also handle this at the application side to suit your use, for example:

```jsx
const ModalWithHook = ({spritemap, children, size, status}) => {
	const [visible, setVisible] = useState(false);
	const {observer, onClose} = useModal({
		onClose: () => setVisible(false),
	});

	if (!visible) {
		return;
	}

	return (
		<ClayModal
			size={size}
			status={status}
			observer={observer}
			spritemap={spritemap}
		>
			{children(onClose)}
		</ClayModal>
	);
};

// Usage
<ModalWithHook spritemap={spritemap}>
	{onClose => (
		<>
			<ClayModal.Header>{"Title"}</ClayModal.Header>
			<ClayModal.Body>
				<h1>{"Hello world!"}</h1>
			</ClayModal.Body>
			<ClayModal.Footer
				first={
					<ClayButton.Group spaced>
						<ClayButton displayType="secondary">{"Secondary"}</ClayButton>
						<ClayButton displayType="secondary">{"Secondary"}</ClayButton>
					</ClayButton.Group>
				}
				last={<ClayButton onClick={onClose}>{"Primary"}</ClayButton>}
			/>
		</>
	)}
</ModalWithHook>
```

Your `children` would need to use render props due to the `onClose` method. But it would supply the need to always use the hook.

> In our case, our application has already grown a lot and there are lots of modal windows being used, we even have modals that open other modals (🤣). I'll check the storybook to figure out how to use a provider, but will using a provider eliminate the need of using useModal?

Yes, you do not have visibility of `useModal`. We just added documentation about using Provider, so you can check it out and give us feedback 😁. https://next.clayui.com/docs/components/modal.html#provider

PR #2604, opened 2019-10-08
chore: separate breadcrumb into its own package

bryceosterhaus:
Did a few things here.

1. Moved breadcrumb to its own package.
2. Renamed navigation to nav since that is the primary export and naming. This also include VerticalNav since it is a variation of nav.


bryceosterhaus:
moved prettier stuff to different PR. https://github.com/liferay/clay/pull/2605

yarn.lock
```diff
@@ -961,6 +961,16 @@
     lodash "^4.17.11"
     to-fast-properties "^2.0.0"
 
+"@clayui/navigation@^3.0.0-alpha.1":
```
matuzalemsteles: This probably shouldn't be here, where are we using it? Maybe it's at `next.clayui.com` that was downloading this dependency.

packages/clay-nav/.yarnrc
```diff
@@ -0,0 +1,3 @@
+# Make `yarn version` produce the right commit message and tag for this package.
+version-tag-prefix "@clayui/nav@"
+version-git-message "chore: prepare @clayui/nav@%s"
```
matuzalemsteles: We should also add this to the new `clay-breadcrumb` package

bryceosterhaus:
@matuzalemsteles updated. I also added the yarnrc change for our yeoman generator

bryceosterhaus:
@matuzalemsteles I always forget about the doc examples. Just updated

bryceosterhaus:
@matuzalemsteles updated. Does this look good to release both nav and breadcrumbs as well? I can release after merging if its good to you.

matuzalemsteles:
@bryceosterhaus yeah, feel free to go ahead with PR.

PR #2605, opened 2019-10-08
chore: remove prettier-eslint

bryceosterhaus:
Removed prettier-eslint dependency and opted to do prettier followed by eslint --fix. prettier-eslint was no longer supported by eslint@v6 and didnt look like much development was going into it. I discovered this because of inconsistencies and errors I would get in my editor. 
    - I also made sure our eslintignore and prettierignore were identical, not sure if we need differences between these two.


wincent:
Suggestion: one thing we did in `liferay-npm-scripts` was invert the order (ie. we made it ESLint then Prettier) and that would probably be a good idea here too. We probably had it running Prettier-then-ESLInt originally because that's what prettier-eslint does too, I think, but it's better to invert.

Rationale: An ESLint "fix" on occasions may produce changes that Prettier considers formatting errors (eg. it can make a line too long). If you make Prettier run second, you know that `format` will produce a good result. On the contrary, if you run Prettier first, then it is possible to hit issues where you run `format` and it "fixes" everything, but if you immediately run `checkFormat` after that it may report a problem.

wincent:
chore__remove_prettier-eslint_by_bryceosterhaus_·_Pull_Request__2605_·_liferay_clay

❤️

package.json
```diff
@@ -3,13 +3,15 @@
...
 		"coveralls": "node node_modules/coveralls/bin/coveralls.js < coverage/lcov.info",
-		"format": "npm run prettier -- --write",
+		"format": "npm run prettier:write && npm run lint:fix",
 		"lerna": "lerna bootstrap -- --no-optional --no-package-lock",
 		"link": "lerna run link",
 		"lint": "eslint '**/*.{js,ts,tsx}'",
-		"prettier": "prettier-eslint '**/*.{js,ts,tsx}'",
+		"lint:fix": "eslint '**/*.{js,ts,tsx}' --fix",
+		"prettier": "prettier '**/*.{js,ts,tsx}'",
+		"prettier:write": "prettier '**/*.{js,ts,tsx}' --write",
```
wincent: For consistency, you might want to change this to match some other projects, we have `prettier` (writes) and `prettier:check` (just reads/checks): rationale — as Prettier is just an pretty-printer (no AST changes) it is totally safe to run, so we make the shorter command (the default), do the most useful and safe thing. This is in contrast to ESLint, where the making "fix" the default wouldn't be safe because the changes it makes can be quite aggressive, so we make the shorter command (`eslint`) just check but not write.

Projects following this pattern are [alloy-editor](https://github.com/liferay/alloy-editor/blob/5795cf638525f1d37cb84f92dffcab7e5452b92c/package.json#L60-L66), [eslint-config-liferay](https://github.com/liferay/eslint-config-liferay/blob/b056a199ef1f239e43cd74c3980d8ac47ba5231e/package.json#L48-L51), [liferay-npm-tools](https://github.com/liferay/liferay-npm-tools/blob/40f5d3cd148ce007231a333bcb8f55cd28c47243/package.json#L15-L18), [liferay-js-themes-toolkit](https://github.com/liferay/liferay-js-themes-toolkit/blob/c95eca5fb6330f507f87266bd8486927cc893897/package.json#L15-L18), [liferay-js-toolkit](https://github.com/liferay/liferay-js-toolkit/blob/3004c05a99ff0252fd29759ae25efd922fd6e154/package.json#L14-L18), [liferay-amd-loader](https://github.com/liferay/liferay-amd-loader/blob/8b1f2a2ff563a6b279649e7121906f3ba4a69f21/package.json#L12-L15), which is to say, most of the ones that are under active development.

bryceosterhaus:
@wincent updated script names to follow our pattern, only one we might want to change in the future is `checkFormat` to be ``ci` but I dont have access to change that on the gh repo so I thought I would leave it.

Additionally, I added markdown and json formatting to prettier, so the latest commit introduces a lot of fixes for that.

wincent:
Looks great, although CI seems to think that next.clayui.com/content/docs/css/scss.md isn't correctly formatted.

> I dont have access to change that on the gh repo so I thought I would leave it.

You don't need special privs to change the `ci` one; you just rename it and then change this part of the .travis.yml to run that script:

https://github.com/liferay/clay/blob/e346a51db3d5e78347c9cae3fafd04780b0d7e62/.travis.yml#L24-L27

> Additionally, I added markdown and json formatting to prettier, so the latest commit introduces a lot of fixes for that.

Sweet!

The only other thing you could do here (or in another PR) would be to change those `npm run` to `yarn`.

bryceosterhaus:
@wincent @matuzalemsteles updated.

next.clayui.com/content/docs/components/forms/css-checkbox-radio.md
```diff
@@ -1,20 +1,13 @@
...
-- [Checkboxes and Radios](#checkboxes-and-radios)
-	- [Default (stacked)](#default-(stacked))
-	- [Inline](#inline)
-	- [Without labels](#without-labels)
-- [Custom Checkboxes and radios](#custom-checkboxes-and-radios)
-	- [Checkboxes](#checkboxes)
-	- [Radios](#radios)
-	- [Inline](#inline)
-	- [Disabled](#disabled)
+-   [Checkboxes and Radios](#checkboxes-and-radios) - [Default (stacked)](<#default-(stacked)>) - [Inline](#inline) - [Without labels](#without-labels)
```
matuzalemsteles: It would be nice to check the documents one by one that he modified in the documents, this breaks the ToC listing.

bryceosterhaus: Its probably gonna be difficult to go through each file one by one and running format. I think it might be best to run format first, and then fix any makdown we catch along the way or catch in this PR review. What do you think?

matuzalemsteles: Yeah, I can analyze all that ended up breaking if you wish and add in PR review. I just didn't want to pollute the PR with all the comments I might have.

bryceosterhaus: I can go through it first and I will ping you when I need anther set of eyes.

matuzalemsteles: ok, thanks Bryce

bryceosterhaus:
@matuzalemsteles looks like the primary thing broken was lists. I went through and think I addressed that issue. Let me know if you see anything else.

wincent:
>  looks like the primary thing broken was lists. I went through and think I addressed that issue.

Seems it couldn't handle tabs for indent. Something to bear in mind when writing docs in the future.

Issue #2592, opened 2019-10-07
Dropdown menu doesn't align properly


bryceosterhaus:
## Is there an example you can provide via codesandbox.com?
See https://storybook.clayui.com/?path=/story/components-claymultiselect--inputwithmultiselect-w-sourceitems

## What are the steps to reproduce?
Type "on" into the input

## What is the expected result?
Autocomplete menu should be directly below input.

This issue is related to https://github.com/liferay/clay/issues/2579, when we change `useEffect` to `useLayoutEffect` it caused this.

PR #2609, opened 2019-10-09
fix: ref callbacks are first calling with null. https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

bryceosterhaus:
Fixes: https://github.com/liferay/clay/issues/2592
So it turned out we were running into issues with callback refs for dropdown. It wasn't properly aligning due to the ref being null. See reacts doc for info on callback refs, https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

### DIFF
--- a/packages/clay-autocomplete/src/index.tsx
+++ b/packages/clay-autocomplete/src/index.tsx
@@ -54,11 +54,14 @@ const ClayAutocomplete = React.forwardRef(
 				{...otherProps}
 				className={className}
 				ref={(r: any) => {
-					containerElementRef.current = r;
-					if (typeof ref === 'function') {
-						ref(r);
-					} else if (ref !== null) {
-						(ref.current as React.MutableRefObject<any>) = r;
+					if (r) {
+						containerElementRef.current = r;
+
+						if (typeof ref === 'function') {
+							ref(r);
+						} else if (ref !== null) {
+							(ref.current as React.MutableRefObject<any>) = r;
+						}
 					}
 				}}
 			>
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -104,11 +104,13 @@ const ClayDropDown: React.FunctionComponent<IProps> & {
 					),
 					onClick: () => onActiveChange(!active),
 					ref: (node: HTMLButtonElement) => {
-						triggerElementRef.current = node;
-						// Call the original ref, if any.
-						const {ref} = trigger;
-						if (typeof ref === 'function') {
-							ref(node);
+						if (node) {
+							triggerElementRef.current = node;
+							// Call the original ref, if any.
+							const {ref} = trigger;
+							if (typeof ref === 'function') {
+								ref(node);
+							}
 						}
 					},
 				})}
--- a/packages/clay-multi-select/src/__tests__/index.tsx
+++ b/packages/clay-multi-select/src/__tests__/index.tsx
@@ -160,6 +160,37 @@ describe('Interactions', () => {
 		expect(onItemsChangeFn).toHaveBeenCalled();
 	});
 
+	it('autocomplete menu renders with left and top styles', () => {
+		const onItemsChangeFn = jest.fn();
+
+		render(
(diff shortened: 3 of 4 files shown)


packages/clay-drop-down/src/Menu.tsx
```diff
@@ -70,30 +70,54 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>((
...
+				).region;
+			} else {
+				newRegion = Align.getAlignRegion(
+					(ref as React.RefObject<HTMLElement>).current!,
+					alignElementRef.current,
+					alignmentPosition
+				);
+			}
+
+			setRegion(newRegion);
```
matuzalemsteles: It seems that this does not solve some things, take a look at the "List Page" demo and try to open most DropDown, some are randomly misaligned.

Maybe this part of align code would solve this: https://github.com/metal/metal-plugins/blob/master/packages/metal-position/src/Align.js#L42-L56. I'm not sure, I'm just making an assumption.

packages/clay-drop-down/src/Menu.tsx
```diff
@@ -70,30 +70,54 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>((
...
 		}
-	});
+	}, [autoBestAlign, active, alignElementRef.current, alignmentPosition]);
+
+	const positioning = region
+		? {
+				left: region.left,
+				top: region.top,
+		  }
```
matuzalemsteles: We can just do this:

```
region ? region : {};
```

Since the region has the same parameters.

bryceosterhaus: I didn't do that since we only want the top and left params, otherwise region returns 

```js
bottom
height
left
right
top
width
```

matuzalemsteles: Oh ok, np. Thanks for the information.

wincent: #golf

You could write this in two lines, although I wish there was a way to do it in one... (I can't think of one, although see below):

```javascript
const {left, top} = region || {};
const positioning = {left, top};
```

(Obviously, if there is no `region`, you'll get `{left: undefined, top: undefined}` rather than `{}` but I am guessing that is ok).

One-line version that uses [optional chaining](https://github.com/tc39/proposal-optional-chaining):

```
const positioning = {left: region?.left, top: region?.top};
```

We might want to turn on the transform for that soon, seeing as the proposal is now at stage 3.

matuzalemsteles: Oh yeah, optional chaining looks much better 🤩.

packages/clay-drop-down/src/Menu.tsx
bryceosterhaus: So I was debugging this and I think the ListPage demo is a particular edge case for the `autoBestAlign` feature from metal-position. If I go back to using the normal Align.align, it functions the same way. I think it has to do with the unique position it is next the the edge of the window. So I think this particular change should still be fine.

One other option is to entirely replace the metal-position library and use something that is more widely adopted. Maybe something like, https://www.npmjs.com/package/dom-align

What do you think?

bryceosterhaus: Actually I take that back... I was starting to see some other issues resulting from that change.

I decided to just remove that refactor and instead only fix the alignment issue.

bryceosterhaus:
@matuzalemsteles updated

packages/clay-drop-down/src/Menu.tsx
matuzalemsteles: Hmm, I didn't realize it was already broken in ListPage (Seems to work fine here https://storybook.clayui.com/?path=/story/demos-templates--list-page). 

Since @jbalsas wants to depreciate metal-position, it may be time to test `dom-align`.

bryceosterhaus:
@matuzalemsteles you okay with merging this first and adding a different align tool in a different PR?

matuzalemsteles:
@bryceosterhaus sounds good. we can go ahead with this PR.

Issue #2610, opened 2019-10-10
Investigate re-adopting lerna for releasing packages


bryceosterhaus:
Now that all of the clay packages have been officially released at at least 3.0.0, we should re-investigate using lerna for releases using `independent` mode.

PR #2630, opened 2019-10-11
docs(contributing): add release process steps

bryceosterhaus:
Spent some time investigating a release process that would allow us to do regular releases for all packages without having to manually do it like we did for 3.0.0. I updated our lerna config and added steps for publishing. Let me know if you think I missed anything or if anyone has any other ideas.

implements https://github.com/liferay/clay/issues/2610

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,7 @@ Here's a set of guidelines to contribute to Clay and its packages. Use your comm
 -   [Documentation Style Guide](#documentation-style-guide)
 -   [Additional Notes](#additional-notes)
 -   [Issue Labels](#issue-labels)
+-   [Release process](#release-process)
 
 ## What Should I Know Before I Get Started?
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -199,3 +200,34 @@ You may want to use [Github search](https://help.github.com/articles/searching-i
...
+
+To publish a new version for all packages that have updated, follow these steps:
+
+1.  `git checkout master`
+2.  Run `lerna version --conventional-commits` and verify each package's version change.
+
+    -   If you'd like to do a dry run first, you can also use the `--no-push` flag to see the commit and tags that lerna will apply. You can also use `lerna changed` to see an output of what packages will be updated
+    -   The `--conventional-commits` flag should automatically determine what version each package should be updated to. If this doesn't seem accurate, you can maually choose each version by just running `lerna version` without the flag.
+
+    > Note, if package A requires package B and package B recieves a minor update, package A will also recieve a minor update via lerna.
```
wincent: recieve -> receive

lerna.json
```diff
@@ -1,6 +1,13 @@
...
 	"lerna": "3.4.0",
-	"version": "3.0.0-alpha.1",
+	"version": "independent",
 	"npmClient": "yarn",
-	"useWorkspaces": true
+	"useWorkspaces": true,
+	"command": {
+		"publish": {
+			"ignoreChanges": ["**/stories/*", "**/__tests__/**", "**/*.md"],
+			"message": "chore(release): pubish updated packages"
```
wincent: pubish -> publish

😂

CONTRIBUTING.md
```diff
@@ -199,3 +200,34 @@ You may want to use [Github search](https://help.github.com/articles/searching-i
...
 | Duplicate       | [search](https://github.com/search?q=is%3Aopen+is%3Aissue+repo%3Aliferay%2Fclay+label%3A%22resolution%3A+duplicate%22&type=Issues) | Resolution | Duplicates an existing issue                                                                                                                                                                                                                                                        |
 | Wontfix         | [search](https://github.com/search?q=is%3Aopen+is%3Aissue+repo%3Aliferay%2Fclay+label%3A%22resolution%3A+wontfix%22&type=Issues)   | Resolution | Will no longer be worked on                                                                                                                                                                                                                                                         |
+
+## Release process
+
+To publish a new version for all packages that have updated, follow these steps:
+
+1.  `git checkout master`
+2.  Run `lerna version --conventional-commits` and verify each package's version change.
```
matuzalemsteles: @bryceosterhaus just out of curiosity, it will set the next version based on type, `feat`, `fix`... maybe we should try to use it more carefully and make sure a `feat` is attached to adding a new API. Just in case lerna gets confused and raises a very tall version.

CONTRIBUTING.md
bryceosterhaus: "I before E except after C or when sounded like A as in neighbor or weigh."

oh yeah checks out. good catch.

lerna.json
bryceosterhaus: yikes

CONTRIBUTING.md
bryceosterhaus: Yeah you're right, we will need to pay closer attention to our commit messages. The other option is to not use `--conventional-commits` but that might be a bit harder to manage all at once. Additionally, we lerna may say "bump to 3.1.0" since there is a "feat" message, but we can override and just say use a patch. So maybe for the first handful of releases we just pay extra close attention.

bryceosterhaus:
> I'm fine with that, maybe we should add here also about the release time and discuss patch releases if we will wait or release right away.

It might be best to start that convo in a separate issue and then we can add that to our docs if we find that necessary. That alright?

CONTRIBUTING.md
matuzalemsteles: Yeah, sounds good.

matuzalemsteles:
> It might be best to start that convo in a separate issue and then we can add that to our docs if we find that necessary. That alright?

Yeah, we can discuss this better later.

PR #2643, opened 2019-10-17
fix(@clayui/icon): forces React not to reuse node when symbol changes

### LINKED ISSUE #2642: Changing ClayIcon Symbol Inside Portal Causes Problems
DXP uses `svg4everybody` to avoid **cors** issues with external links, it makes an Ajax request and replaces `svg:use`.

Since we have ClayIcon in React, changing the `symbol` inside DXP causes it not to be changed because React reuses node to make changes. This only works the first time.

matuzalemsteles:
Fixes #2642

As suggested by @jbalsas, it was the same solution in Metal. Forces to change the `key` of the element so that React does not reuse the node, so `svg4everybody` will re-capture the node to make its changes.

I am adding this straight to `ClayIcon` because it is common for someone to do this and so we can support it directly, I don't see a real problem with that so please let me know if you see anything.

Once we get on with this, we will need to release some fix patches for the components that use the ClayIcon. I'm eyeing ColorPicker that symbol change...

### DIFF
--- a/packages/clay-icon/src/index.tsx
+++ b/packages/clay-icon/src/index.tsx
@@ -43,6 +43,7 @@ const ClayIcon = React.forwardRef<SVGSVGElement, IProps>(
 					`lexicon-icon lexicon-icon-${symbol}`,
 					className
 				)}
+				key={symbol}
 				ref={ref}
 				role="presentation"
 			>


packages/clay-icon/src/index.tsx
```diff
@@ -43,6 +43,7 @@ const ClayIcon = React.forwardRef<SVGSVGElement, IProps>(
 					`lexicon-icon lexicon-icon-${symbol}`,
 					className
 				)}
+				key={symbol}
```
bryceosterhaus: Should we also add a useEffect so that the component doesnt re-render unless symbol or one of the other props changes?

wincent: I don't think `useEffect` is really needed or desirable here.

matuzalemsteles: Hmm, not sure about `useEffect`... why would we use it?

bryceosterhaus: oh i mis-read your original post. I thought you were adding this to avoid re-renders, but i see now that its for not re-using the node.

PR #2651, opened 2019-10-21
fix(@clayui/drop-down): add `alignmentPosition` support in DropDownWithItems

### LINKED ISSUE #2647: `DropDownWithItems` does not support the `alignmentPosition` attribute

Create a `DropDownWithItems` and give it the attribute `alignmentPosition={Align.BottomRight}`.

The resulting dropdown will be aligned according to the attribute value.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.0.0  |
| React | v16.9.0 |

matuzalemsteles:
Fixes #2647

### DIFF
--- a/packages/clay-drop-down/src/DropDownWithItems.tsx
+++ b/packages/clay-drop-down/src/DropDownWithItems.tsx
@@ -47,6 +47,11 @@ interface IDropDownContentProps {
 
 interface IProps extends IDropDownContentProps {
 	/**
+	 * Default position of menu element. Values come from `metal-position`.
+	 */
+	alignmentPosition?: number;
+
+	/**
 	 * Informational text that appears at the end or above the `footerContent` prop.
 	 */
 	caption?: string;
@@ -246,6 +251,7 @@ const DropDownContent: React.FunctionComponent<IDropDownContentProps> = ({
 );
 
 export const ClayDropDownWithItems: React.FunctionComponent<IProps> = ({
+	alignmentPosition,
 	caption,
 	className,
 	containerElement,
@@ -268,6 +274,7 @@ export const ClayDropDownWithItems: React.FunctionComponent<IProps> = ({
 	return (
 		<ClayDropDown
 			active={active}
+			alignmentPosition={alignmentPosition}
 			className={className}
 			containerElement={containerElement}
 			hasLeftSymbols={hasLeftSymbols}


packages/clay-drop-down/src/DropDownWithItems.tsx
```diff
@@ -246,6 +251,7 @@ const DropDownContent: React.FunctionComponent<IDropDownContentProps> = ({
 );
 
 export const ClayDropDownWithItems: React.FunctionComponent<IProps> = ({
+	alignmentPosition,
```
bryceosterhaus: we might just want to add `dropDownProps` since we also might want to give the option for "autoBestAlign" and any others that come up. wdyt?

matuzalemsteles: Hmm, if we do that we will have to make changes to much of the current `DropDownWithItems` API, for example we have `containerElement`, `trigger` and `className`, the other APIs we are covering with internal implementations (`hasLeftSymbols`,` hasRightSymbols`, `onActiveChange`, `active`). So it can be a little misleading to have a `dropDownProps` and we will need to move all the APIs there.

bryceosterhaus: We could  omit any internal props we use from that interface. This would help give a straight forwardness to what props people can actually use and then make sure `dropDownProps` is the first item spread into its props, so that our internal uses are covered.

matuzalemsteles: Yeah I have no problem omitting it just a bit weird having a `dropDownProps` prop that refers to DropDown props on a component that is high-level DropDown, it seems simpler to see APIs like this as a component host without having to be passed to a specific prop...

Maybe we can extend from `ClayDropDown` and use `otherProps` to switch to `ClayDropDown` since it is the main component of `ClayDropDownWithItems`. And we can omit the internal uses of `ClayDropDown`.

bryceosterhaus: oh thats a good point, I didnt think of it that way. So actually, in that case I think I am fine with the implementation as is. If we find ourselves needing additional props in the future we can extend from dropdown and spread otherProps, this would keep the same compatibility as well.

So overall, this PR looks good to me.

matuzalemsteles: Yeah, sounds good. I will go ahead with this PR.

PR #2656, opened 2019-10-22
fix(clay-dropdown): trims whitespace

jbalsas:
Hey @matuzalemsteles, this is a fix for https://issues.liferay.com/browse/LPS-103325... I've looked at it with @wincent and we're not sure when or why this started happening, but we can clearly see that this message is not being properly compiled so the message is not correctly replaced.

Could you take a look and release a patch version for it?

Also, not sure if this is the branch I should be sending this to :D


matuzalemsteles:
Oh that's very strange. I will run CI again to see if it turns green so it can merge, so we can go ahead and release a version safely.

matuzalemsteles:
Just started reviewing :)

matuzalemsteles:
Resent #2657

Issue #2640, opened 2019-10-16
ClayTooltipProvider doesn't work with nested elements


thektan:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## Is there an example you can provide via codesandbox.com?

https://codesandbox.io/embed/clay-tooltip-with-icon-6pdzo

## What are the steps to reproduce?

ClayTooltipProvider seems to only work if you're wrapping a single-level element like `<span title="tooltip">test</span>`. 

## What is the expected result?

I'd expect the tooltip to appear even if there are more nested elements like
```
<div title="tooltip">
  <span>test</span>
</div>
```

My main use case is adding a tooltip to an icon.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.0.1 |
| React | 16.9.0 |

thektan:
Actually this works (which is my current use case for it):

```javascript
<ClayTooltipProvider>
	<ClayIcon
		spritemap={spritemap}
		symbol="question-circle-full"
		title="Tooltip Content"
	/>
</ClayTooltipProvider>
```

Not sure why locally, but I had multiple clay icons I was testing out and after hovering over a broken one (written like `<span title="test"><ClayIcon ... /></span>```), the icon I had in the snippet above wouldn't appear.

No use case to fix this bug yet, but I still believe it's a bug if you want a tooltip around a compound element. I guess a current workout would be extracting it into a separate component?

thektan:
Well this is strange...I can't seem to get spritemaps to load in the codesandbox, but here's the behavior I'm seeing with using `<ClayTooltipProvider><ClayIcon></ClayTooltipProvider>`.

![Kapture 2019-10-16 at 16 16 11](https://user-images.githubusercontent.com/4496722/66966005-8e6b3d80-f030-11e9-864b-07887a93238a.gif)

matuzalemsteles:
Hmm, thanks @thektan. This has to do with the internal use of `onMouseOut` and` onMouseOver` since we use it to show and hidden Tooltip over the active element, so it tracks any nested element.

@bryceosterhaus we can use an approach to cover these cases, one strategy would be to add `pointer-events: none` over the child elements of the `title` element, that would be a little difficult to do or we might suggest putting them.

bryceosterhaus:
[...]

PR #2661, opened 2019-10-23
fix(tooltip): find closest node with title attribute

bryceosterhaus:
Fixes: https://github.com/liferay/clay/issues/2640

@thektan can you verify this works as you would expect?  I added an additional nested element in the TooltipProvider story

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -66,6 +66,25 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
 	}
 };
 
+function closestAncestor(node: HTMLElement, s: string) {
+	const el = node;
+	let ancestor: HTMLElement | null = node;
+
+	if (!document.documentElement.contains(el)) {
+		return null;
+	}
+
+	do {
+		if (ancestor.matches(s)) {
+			return ancestor;
+		}
+
+		ancestor = ancestor.parentElement;
+	} while (ancestor !== null);
+
+	return null;
+}
+
 const TooltipBase: React.FunctionComponent<{
 	autoAlign?: boolean;
 	children: React.ReactElement;
@@ -79,17 +98,23 @@ const TooltipBase: React.FunctionComponent<{
 	// Using `any` type since TS incorrectly infers setTimeout to be from NodeJS
 	const timeoutIdRef = useRef<any>();
 	const targetRef = useRef<HTMLElement | null>(null);
+	const titleNodeRef = useRef<HTMLElement | null>(null);
 	const tooltipRef = useRef<HTMLElement | null>(null);
 
 	const handleHide = ({target}: any) => {
-		const dataTitle = target && target.getAttribute('data-title');
+		if (!titleNodeRef.current) {
+			return;
+		}
+
+		const dataTitle = titleNodeRef.current.getAttribute('data-title');
 
 		if (dataTitle) {
-			targetRef.current = null;
-
 			target.removeEventListener('click', handleHide);
-			target.setAttribute('title', dataTitle);
-			target.removeAttribute('data-title');
+			titleNodeRef.current.setAttribute('title', dataTitle);
+			titleNodeRef.current.removeAttribute('data-title');
+
+			titleNodeRef.current = null;
+			targetRef.current = null;
 
 			dispatch({type: 'hide'});
 			clearTimeout(timeoutIdRef.current);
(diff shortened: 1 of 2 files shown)


packages/clay-tooltip/src/TooltipProvider.tsx
```diff
@@ -66,6 +66,20 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
 	}
 };
 
+function closestParent(node: HTMLElement, s: string) {
+	const el = node;
+	let ancestor: HTMLElement | null = node;
+
+	if (!document.documentElement.contains(el)) return null;
```
matuzalemsteles: Just out of curiosity would this cover the case when the element disappeared from the screen or is it related to another use case?

packages/clay-tooltip/stories/index.tsx
```diff
@@ -64,6 +64,13 @@ storiesOf('Components|ClayTooltip', module)
 					>
 						{'Custom Delay'}
 					</button>
+
+					<div
```
matuzalemsteles: Maybe it is better to add another example? At first glance I thought the Tooltip position was broken, but it was only because the `div` is 100% 😅.

bryceosterhaus: I actually did that on purpose so that the user can see that it aligns to the node the title is on. Maybe making it a different color would help that?

packages/clay-tooltip/src/TooltipProvider.tsx
bryceosterhaus: hmm not totally sure, I copied this bit of code from an existing polyfil that I cound for `Element.closest()`

packages/clay-tooltip/stories/index.tsx
matuzalemsteles: Oh yes that's a great strategy. Yes, adding a soft color would help.

bryceosterhaus:
@matuzalemsteles updated

packages/clay-tooltip/src/TooltipProvider.tsx
```diff
@@ -66,6 +66,20 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
...
+
+	do {
+		if (ancestor.matches(s)) return ancestor;
```
wincent: I thought it was our convention to  always wrap statements in curlies. eg.

```javascript
if (ancestor.matches(s)) {
    return ancestor;
}

File: packages/clay-tooltip/src/TooltipProvider.tsx [hunk 4]
```diff
@@ -66,6 +66,20 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
 	}
 };
 
+function closestParent(node: HTMLElement, s: string) {
```
wincent: This should probably be called `closestAncestor` or even just `closest` (seeing as that is what you're polyfilling). The word "parent" implies an immediate parent-to-child relationship.

packages/clay-tooltip/src/TooltipProvider.tsx
```diff
@@ -79,35 +93,44 @@ const TooltipBase: React.FunctionComponent<{
...
+			titleNodeRef.current.getAttribute('data-title');
 
 		if (dataTitle) {
-			targetRef.current = null;
-
 			target.removeEventListener('click', handleHide);
-			target.setAttribute('title', dataTitle);
-			target.removeAttribute('data-title');
+			titleNodeRef.current!.setAttribute('title', dataTitle);
+			titleNodeRef.current!.removeAttribute('data-title');
```
wincent: It would be nice to get rid of these `!` assertions (because they reduce safety). I think you can do that with a minor tweak:

```diff
diff --git a/packages/clay-tooltip/src/TooltipProvider.tsx b/packages/clay-tooltip/src/TooltipProvider.tsx
index 38da5c4c9..4765d3549 100644
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -97,14 +97,16 @@ const TooltipBase: React.FunctionComponent<{
 	const tooltipRef = useRef<HTMLElement | null>(null);
 
 	const handleHide = ({target}: any) => {
-		const dataTitle =
-			titleNodeRef.current &&
-			titleNodeRef.current.getAttribute('data-title');
+		if (!titleNodeRef.current) {
+			return;
+		}
+
+		const dataTitle = titleNodeRef.current.getAttribute('data-title');
 
 		if (dataTitle) {
 			target.removeEventListener('click', handleHide);
-			titleNodeRef.current!.setAttribute('title', dataTitle);
-			titleNodeRef.current!.removeAttribute('data-title');
+			titleNodeRef.current.setAttribute('title', dataTitle);
+			titleNodeRef.current.removeAttribute('data-title');
 
 			titleNodeRef.current = null;
 			targetRef.current = null;
```

packages/clay-tooltip/src/TooltipProvider.tsx
bryceosterhaus: my bad, I had copied this code from a polyfil and forgot to format it.

packages/clay-tooltip/src/TooltipProvider.tsx
bryceosterhaus: Good catch. I was having issues with ts inferring that it wouldn't be possible to be null, however I can imagine that any refactoring in the future would cause this to be problematic.

bryceosterhaus:
@wincent updated commit

wincent:
👍

PR #2663, opened 2019-10-25
chore: update linting

wincent:
Moves to eslint-config-liferay v13.0.0, which requires the "notice/notice" plugin to be manually configured because of [changes in liferay-portal explained here](https://github.com/liferay/eslint-config-liferay/issues/115).


.eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+	plugins: ['@typescript-eslint'],
+	rules: {
+		'@typescript-eslint/array-type': ['error', 'generic'],
+		'@typescript-eslint/interface-name-prefix': ['error', 'always'],
+		'@typescript-eslint/no-unused-vars': 'error',
+		'no-unused-vars': 'off',
+		'notice/notice': [
+			'error',
+			{
+				templateFile: path.join(__dirname, 'copyright.js'),
```
wincent: And here's why we "needed" to change from JSON to JS for this config file: for consistency with other repos, we pass an absolute path here. This repo is a simple case, so it it would also work with plain old "copyright.js", but I think it's worth matching what we do in liferay-portal. As a bonus, that means we can now lint the config file itself, which means we get ordering enforcement etc.

packages/clay-alert/src/__tests__/ClayAlert.tsx
```diff
@@ -5,8 +5,8 @@
  */
 
 import ClayAlert from '..';
-import * as TestRenderer from 'react-test-renderer';
 import React from 'react';
+import * as TestRenderer from 'react-test-renderer';
```
wincent: These and all the other changes are just autofixes.

PR #2664, opened 2019-10-25
docs: add warning to point to 2.x on homepage

bryceosterhaus:
I added an alert to the homepage to point to v2. I think this will help anyone who might be confused at the design change and who didn't scroll to see the deprecation warning. This helps keep the alert above the scroll line and will help inform users of knowing where v2 docs are at, as well as knowing this is a v3 site.

We can eventually remove it, but I think itd be good to keep around for a bit.

Screen Shot 2019-10-25 at 12 01 16 PM

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,12 +2,12 @@
 bower_components
 build
 clay-css
-clayui.com
 copyright.js
 coverage
 lib
-next.clayui.com/public
-next.clayui.com/static
+clayui.com/.cache
+clayui.com/public
+clayui.com/static
 node_modules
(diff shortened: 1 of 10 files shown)


eduardoallegrini:
Maybe we should do the same for the v2 pages "find the newest version here [link-to-v3]"

matuzalemsteles:
Hey @eduardoallegrini might be a good idea, I haven't put it yet for the reason that people will probably go to clayui.com and then v2.clayui.com and the kind of inferred domain is a specific version, does it make sense?

PR #2672, opened 2019-10-30
feat(@clayui/data-provider): adds lazy call support for `useResource`

### LINKED ISSUE #2431: Question: ClayDataProvider - Prevent refetch on variable change
## What is your question?
**1. Is there a proper way of using [`variables`](https://next.clayui.com/docs/components/data-provider.html#variables-change) without always refetching?**

For example:
```
const {resource, refetch} = useResource({
	link: fetchDocumentsSearchUrl,
	variables: {
		keywords: searchQuery
	}
});
```
where `searchQuery` is an input field value.

I'd only want to perform a fetch when a user explicitly clicks a button or enter rather than on input change. 

Delays can be set by `fetchDelay`, but I was curious if there was an option for completely preventing the fetch or if I'd have to build the `searchParams` into the `link`.

**2. Is there a way to prevent the initial fetch when the component mounts?**

Thanks!

matuzalemsteles:
Fixes #2431

See the issue for more context. I am adding the functionality of making the first request manually.

### DIFF
--- a/packages/clay-data-provider/src/useResource.ts
+++ b/packages/clay-data-provider/src/useResource.ts
@@ -19,14 +19,28 @@ import {useCache} from './useCache';
 import {timeout} from './util';
 
 interface IResource extends IDataProvider {
+	/**
+	 * Flag that tells `useResource` not to make the first request as soon as the
+	 * component is mounted, you can do it manually by calling the `get()` method
+	 * returned by the hook.
+	 *
+	 * > const {get: getPosts} = useResource({lazy: true'});
+	 */
+	lazy?: boolean;
+
 	onNetworkStatusChange?: (status: NetworkStatus) => void;
 }
+
+const warningFirstFetchMessage = `DataProvider: The first fetch has already happened, if you want to make a new fetch, use the 'refetch()' method but if you want to make the first fetch enable the 'lazy' hook mode.
+
+> const {refetch, get} = useResource({lazy: true, link: 'https://...'});`;
 
 const useResource = ({
 	fetchDelay = 300,
 	fetchOptions,
 	fetchPolicy = FetchPolicy.NoCache,
 	fetchTimeout = 6000,
+	lazy,
 	link,
 	onNetworkStatusChange: dispatchNetworkStatus = () => {},
 	pollInterval = 0,
@@ -210,6 +224,17 @@ const useResource = ({
 		doFetch();
 	};
 
+	const doFirstFetch = () => {
+		warning(firstRenderRef.current, warningFirstFetchMessage);
+
+		if (!firstRenderRef.current) {
+			return;
+		}
+
+		maybeFetch(NetworkStatus.Loading);
+		firstRenderRef.current = false;
+	};
+
 	useEffect(() => {
 		if (!firstRenderRef.current) {
 			maybeFetch(NetworkStatus.Refetch);
@@ -217,8 +242,9 @@ const useResource = ({
 	}, [debouncedVariablesChange]);
 
 	useEffect(() => {
-		maybeFetch(NetworkStatus.Loading);
-		firstRenderRef.current = false;
+		if (!lazy) {
+			doFirstFetch();
+		}
 
 		return () => {
 			// Reset the cache only if the storage reference is
@@ -231,7 +257,7 @@ const useResource = ({
 		};
 	}, []);
 
-	return {refetch: handleRefetch, resource};
+	return {get: doFirstFetch, refetch: handleRefetch, resource};
 };
 
 export {useResource};
--- a/packages/clay-data-provider/src/useResource.ts
+++ b/packages/clay-data-provider/src/useResource.ts
@@ -145,21 +145,17 @@ const useResource = ({
 		return uri;
 	};
 
-	const getUrlFormat = (
-		link: string,
-		variables: TVariables,
-		fetchOptions?: RequestInit
-	) => {
+	const getUrlFormat = (link: string, variables: TVariables) => {
 		const uri = new URL(link);
-
-		if ((fetchOptions && fetchOptions.method !== 'GET') || !variables) {
-			return uri.toString();
-		}
 
 		warning(
 			uri.searchParams.toString() === '',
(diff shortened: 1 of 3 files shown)


packages/clay-data-provider/src/useResource.ts
```diff
@@ -19,14 +19,28 @@ import {useCache} from './useCache';
...
+	 * returned by the hook.
+	 *
+	 * > const {get: getPosts} = useResource({lazy: true'});
+	 */
+	lazy?: boolean;
+
 	onNetworkStatusChange?: (status: NetworkStatus) => void;
 }
 
+const warningFirstFetchMessage = `DataProvider: The first fetch has already happened, if you want to make a new fetch, use the 'refetch()' method but if you want to make the first fetch enable the 'lazy' hook mode.
```
bryceosterhaus: I think we might want to move this to our docs rather than emit a warning. Primarily because they aren't necessarily doing anything wrong, and if they intend to already use the first fetch, they will always see this warning. wdyt?

packages/clay-data-provider/stories/index.tsx
```diff
@@ -209,4 +209,79 @@ storiesOf('Components|ClayDataProvider', module)
...
 	.add('with variables change and storage', () => (
 		<ClayDataProviderWithVariablesAndStorage />
-	));
+	))
+	.add('with lazy fetch', () => {
+		const [
+			networkStatus,
+			setNetworkStatus,
+		] = useState<NetworkStatus | null>(null);
+		const {get: getNames, resource} = useResource({
```
bryceosterhaus: need new line above

matuzalemsteles: 👍

packages/clay-data-provider/src/useResource.ts
matuzalemsteles: I think it could also be added to the documentation, but this message will only appear in some cases, for example:

- If `lazy` is not enabled, the component once assembled will make the first request and if people calls `get` the message will appear.
- If `lazy` is enabled and people calls `get` twice the message will appear.

The message was not very clear about these cases?

wincent: This doesn't seem like a very developer-friendly API. As a user, I'd be asking, why should I have two separate methods for "first lazy fetch" versus "fetch again", when they are both methods that "fetch" the data and in both cases I really just mean "go get data now"?

matuzalemsteles: Yeah at first yes, but it's weird to have the same `refetch` method that makes the first request and then the subsequent ones when a `lazy` is activated. We are changing your responsibility...

The idea of ​​`lazy` is for cases when you want to execute the query in response to a different event other than the assembly of the component. So we just run it the first time. In essence `refetch` will only be used in other cases and it changes the network status to `Refetch`.

Another approach would be to create a `useLazyResource` that has only this behavior, something similar to React Apollo's `useLazyQuery` but would still have the same idea, a `get` and `refetch`.

matuzalemsteles:
Hey @bryceosterhaus and @wincent, I'm coming back to this again. Since we are not reaching API consensus for this functionality. I've been thinking of another alternative.

Instead of using a "`lazy`" we can add a `skip` API which if the condition is `false` the request will not be made and if `true` the next call by the `refetch()` method will be made but the state of `networkStatus` will be `NetworkStatus.Refetch`.

Dependent fetching:
```js
const {resource: contributors} = useResource({
    link: 'https://clay.dev/contributors'
});

const {resource: profile} = useResource({
    link: 'https://clay.dev/contributors',
    variables: {
        id: contributors.id
    },
    skip: !contributors.id
});
```

Skip first request:
```jsx
const isMounted = useIsMounted();

const {resource: contributors, refetch} = useResource({
    link: 'https://clay.dev/contributors',
    skip: !isMounted
});

return (
    <button onClick={() => refetch()}>{'Load data'}</button>
);
```

This may cover cases of issue #2723 but the call will still have to be manual to make a new request. Maybe we can always make a new request when `skip` changes the state to `true`. What you think?

bryceosterhaus:
@matuzalemsteles what if we called it `fetchOnMount` and defaulted it to `true`? This would give a bit clearer idea of what the prop does. If we just say `skip`, people may think toggling it later in the lifecycle might do something, which I don't think it would.

Also, I still think we should remove the warning and just keep it in the docs. Warning a user when their configuration is valid seems odd. We should keep warnings for dev mishaps and errors

matuzalemsteles:
> @matuzalemsteles what if we called it fetchOnMount and defaulted it to true? This would give a bit clearer idea of what the prop does. If we just say skip, people may think toggling it later in the lifecycle might do something, which I don't think it would.

Yeah, I think you might have the impression of skip later though I think it will be pretty unlikely to do that. Maybe `fetchOnMount` is the best direction as it is simpler for this case but` skip` would help us with the dependent request although I like the approach I mentioned in the issue.

> Also, I still think we should remove the warning and just keep it in the docs. Warning a user when their configuration is valid seems odd. We should keep warnings for dev mishaps and errors

Yeah, if we don't go in the direction of this PR we won't need to add a warning but I like the idea of ​​adding warnings in the development environment I think it is a good support to help the developer use the component. It may happen that they do not know what is happening, even though we have documented it.

matuzalemsteles:
@bryceosterhaus I was analyzing this again today and one thing that worries me about adding `fetchOnMount` to disable the initial request is that we will have to expose` refetch` for the developer to use to make the first request, I don’t like it very much because it doesn't seem semantic to me, because it will behave in two ways:

- One he will make the first request and call `dispatchNetworkStatus` with `NetworkStatus.Loading`
- The next calls to the `refetch` function with `NetworkStatus.Refetch`.

Another option is to always call `dispatchNetworkStatus` with `NetworkStatus.Refetch`, the status of `onNetworkStatus` will always be `NetworkStatus.Refetch`, we can have this as a side effect... I'm not sure yet, I want to evaluate other options. Any suggestion?

kresimir-coko:
What's the status of this PR @matuzalemsteles @bryceosterhaus?

bryceosterhaus:
@kresimir-coko I'm not positive... I know we talked about moving DataProvider to be a DXP-specific component, so this may fall in line with that process as well.

bryceosterhaus:

PR #2680, opened 2019-11-01
fix(@clayui/data-provider): replace setInterval in favor of setTimeout and polling engine improvements

### LINKED ISSUE #2679: DataProvider: enable and disable `pollInterval`

matuzalemsteles:
Fixes #2679

Replace the use of `setInterval` to `setTimeout` is safer for now, because we may have cases when `pollInterval` is set to 1000ms and the request may take longer than this time and we may have more than one request being made. and that number can accumulate. So we will always wait for the request to succeed in scheduling the next one.

Such a change means that `pollInterval` is not always executed every 1000ms for example, it also contains the time of the previous request. In a way, we already had this when we were using `setInterval`... but we are preventing more than one request from being made at the same time and possibly creating some kind of server overload.

A change to `pollInterval` also to enable you to make a mechanism to enable and disable `poll` by setting to some value above 0, enable and equal to 0/false disable.

### DIFF
--- a/packages/clay-data-provider/stories/index.tsx
+++ b/packages/clay-data-provider/stories/index.tsx
@@ -6,7 +6,7 @@
 
 import {number} from '@storybook/addon-knobs';
 import {storiesOf} from '@storybook/react';
-import React, {useContext, useState, useRef} from 'react';
+import React, {useContext, useState} from 'react';
 
 import ClayDataProvider, {useResource} from '../src';
 import {FetchPolicy} from '../src/types';
@@ -213,25 +213,18 @@ storiesOf('Components|ClayDataProvider', module)
 	.add('toggle poll', () => {
 		const [poll, setPoll] = useState(false);
 
-		const {resource} = useResource({
-			link:
-				'https://api-public.sandbox.pro.coinbase.com/products/BTC-USD/trades',
+		useResource({
+			link: 'https://rickandmortyapi.com/api/character',
 			pollInterval: poll ? 1000 : 0,
 			variables: {limit: 10},
 		});
 
-		console.log(resource);
-
 		return (
-			<div className="sheet pb-4">
-				<h3>Polling {poll ? 'enable' : 'disable'}</h3>
+			<div className="pb-4 sheet">
+				<h3>{`Polling  ${poll ? 'enable' : 'disable'}`}</h3>
 				<div className="row">
 					<div className="col-md-5">
-						<p>
-							{
-								'Open your console to see the data and see the network tab.'
-							}
-						</p>
+						<p>{'Open your console to see the network tab.'}</p>
 					</div>
 				</div>
 				<div className="row">
@@ -240,7 +233,7 @@ storiesOf('Components|ClayDataProvider', module)
 							className="btn btn-primary"
 							onClick={() => setPoll(!poll)}
 						>
-							Toggle Poll
+							{'Toggle Poll'}
 						</button>
 					</div>
 				</div>
--- a/packages/clay-data-provider/stories/index.tsx
+++ b/packages/clay-data-provider/stories/index.tsx
@@ -6,9 +6,9 @@
 
 import {number} from '@storybook/addon-knobs';
 import {storiesOf} from '@storybook/react';
-import React, {useContext, useState} from 'react';
+import React, {useContext, useState, useRef} from 'react';
 
-import ClayDataProvider from '../src';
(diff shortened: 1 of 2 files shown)


packages/clay-data-provider/stories/index.tsx
```diff
@@ -209,4 +209,42 @@ storiesOf('Components|ClayDataProvider', module)
...
 	.add('with variables change and storage', () => (
 		<ClayDataProviderWithVariablesAndStorage />
-	));
+	))
+	.add('toggle poll', () => {
+		const [poll, setPoll] = useState(false);
+
+		const {resource} = useResource({
+			link:
+				'https://api-public.sandbox.pro.coinbase.com/products/BTC-USD/trades',
```
bryceosterhaus: Could we maybe point to something else? Bitcoin doesn't always have a great reputation, and might throw some people off to see a link to coinbase.

packages/clay-data-provider/stories/index.tsx
```diff
@@ -209,4 +209,42 @@ storiesOf('Components|ClayDataProvider', module)
...
+			pollInterval: poll ? 1000 : 0,
+			variables: {limit: 10},
+		});
+
+		// eslint-disable-next-line no-console
+		console.log(resource);
```
bryceosterhaus: is this log necessary for our storybook?

packages/clay-data-provider/stories/index.tsx
matuzalemsteles: Oh yeah, np. Can i change for another 😁

packages/clay-data-provider/stories/index.tsx
matuzalemsteles: This is only for viewing polling in action when it is turned off and on.

julien:
Hey @matuzalemsteles and @bryceosterhaus I know I'm a bit late to the party, but I'm surprised we're not using `requestAnimationFrame` - I know the feature in this PR is not related with anything visual like animation or DOM manipulation, but it would still probably a good idea to experiment with `requestAnimationFrame` unless there are good reasons to use `setInterval/setTimeout`. What are your thoughts on this?

matuzalemsteles:
Hey @julien I'm not sure about using `requestAnimationFrame` since we don't have callback call interval control. Since its main function is to schedule a function that can be called before the next repaint, I would say it might not be necessary. Since the main function here is to schedule the next data call, once we have this data the developer can print this out in the React.js application, so underneath React.js will schedule and decide when it will update.

I have this line of reasoning now, but maybe I'm missing something in your question 😁, I'm sorry if I miss something 🙂.

julien:
> I'm not sure about using requestAnimationFrame since we don't have callback call interval control.

 I don't know what you mean by _callback call interval control_ but you actually do get passed an argument in the callback passed to raf which is typically named `deltaTime`, i.e.

```javascript
function loop(deltaTime) {
   /* Here you can do what you want with deltaTime
      (which is similar to what performance.now() returns) 
      so you could return out of the function if you needed to */

   
   requestAnimationFrame(loop);
}

requestAnimationFrame(loop);
```
The other advantage is that if the browser's tab is not active all this is paused so you don't drain the user's laptop or mobile phone battery

matuzalemsteles:
> I don't know what you mean by callback call interval control but you actually do get passed an argument in the callback passed to raf which is typically named deltaTime, i.e.

It would be to tell the scheduler to call the callback only after this set interval, perhaps with `deltaTime` information we can decide whether to execute now or wait for the next call. But there is a detail that maybe the callback may take a long time to call but I like this idea, we will be practically creating our own scheduler implementation 😁

> The other advantage is that if the browser's tab is not active all this is paused so you don't drain the user's laptop or mobile phone battery

Oh yeah, this is true. I think this is a great resource.

Edit:

Just on top of that, we would need to ensure compatibility with other browsers... one approach would be to use the very well designed React `scheduler`. But it still doesn't seem to be a public API but they intend to.
https://github.com/facebook/react/tree/master/packages/scheduler

OBS: We also have to see if it helps solve this problem, because they deal with priorities.

matuzalemsteles:
Hey @julien I will investigate this further in a separate issue so I will deal with it and then we can work on it better. @bryceosterhaus can you take a look again to see if you need anything else? I want to include this in today's release to send it all to Chema.

bryceosterhaus:
@matuzalemsteles only thing was the storybook example and switching away from a coinbase url. Once that is fixed, this looks good to me.

PR #2682, opened 2019-11-01
docs: update type structure so that react-docgen doesnt error

bryceosterhaus:
I've been hitting my head against a wall with our autogenerated docs. I believe this is an improvement on what we have so far.

Notes:
- I had to add some changes to react-docgen to work better with TS. Thus, I forked it into our project . The goal would be to go back to the upstream version whenever they release a stable 5.0.0.
- Instead of a list for our APIs, I switched to a table. Some of the default values look wacky, but I think its better than what we had before.
- Refactored types. react-docgen was choking on some of the ways we had defined our types, I re-worked it a bit so that it doesnt error.

Let me know what you think. I think this is a step in the right direction, but still have progress to be made.


bryceosterhaus:
or maybe its still broke.. I'll investigate and re-send whenever I figure it out

PR #2693, opened 2019-11-04
feat(@clayui/multi-select): adds support for DropDown.Menu content customization

### LINKED ISSUE #2685: MultiSelect should allow you to customize DropDown items.
ClayMultiSelect must allow you to customize DropDown content to adhere to Lexicon extension points.

matuzalemsteles:
Fixes #2685

See the Storybook with the use case, but this is a step to support the extension point, would still lack support to customize the content of Label, we still have to check what will be the best solution I will deal with in another PR.

### DIFF
--- a/packages/clay-multi-select/src/index.tsx
+++ b/packages/clay-multi-select/src/index.tsx
@@ -20,16 +20,32 @@ const ENTER_KEY = 13;
 const DELIMITER_KEYS = [ENTER_KEY, COMMA_KEY];
 
 type Item = {
+	[propName: string]: any;
+
 	/**
 	 * Label to show.
 	 */
-	label: string;
+	label?: string;
 
 	/**
 	 * Hidden value of the item.
 	 */
+	value?: string;
+};
+
+type Locator = {
+	label: string;
 	value: string;
 };
+
+interface IMenuRendererProps {
+	inputValue: string;
+	locator: Locator;
+	onItemClick?: (item: Item) => void;
+	sourceItems: Array<Item>;
+}
+
+type MenuRenderer = React.FunctionComponent<IMenuRendererProps>;
 
 export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	/**
@@ -41,6 +57,12 @@ export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	 * Aria label for the Close button of the labels.
 	 */
 	closeButtonAriaLabel?: string;
+
+	/**
+	 * Adds a component to replace the default component that renders
+	 * the content of the `<ClayDropDown />` component.
+	 */
+	menuRenderer?: MenuRenderer;
 
 	/**
 	 * Flag that indicates to disable all features of the component.
@@ -73,6 +95,11 @@ export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	items: Array<Item>;
 
 	/**
+	 * Sets the name of the field to map the value/label of the item
+	 */
+	locator?: Locator;
+
+	/**
 	 * Callback for when the clear all button is clicked
 	 */
 	onClearAllButtonClick?: () => void;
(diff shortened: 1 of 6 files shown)


packages/clay-multi-select/src/index.tsx
```diff
@@ -42,6 +58,12 @@ export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	 */
 	closeButtonAriaLabel?: string;
 
+	/**
+	 * Adds a component to replace the default component that renders
+	 * the content of the `<ClayDropDown />` component.
+	 */
+	menuContentComponent?: MenuContentComponent;
```
matuzalemsteles: Not sure about that name, maybe `menuComponent` just?

packages/clay-multi-select/src/index.tsx
```diff
@@ -77,6 +99,11 @@ export interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	 */
 	items: Array<Item>;
 
+	/**
+	 * Sets the name of the field to map the value/label of the item
+	 */
+	locator?: Locator;
```
matuzalemsteles: `locator` is something important that we should adopt not all multi selects adopt the same signature of `label, value` this may be different for each team as it will impact not only the data coming from the backend but also the data that goes to the backend.

packages/clay-multi-select/src/index.tsx
```diff
@@ -20,17 +20,33 @@ const ENTER_KEY = 13;
...
 	 */
+	value?: string;
+};
+
+type Locator = {
+	label: string;
 	value: string;
 };
 
+type MenuContentProps = {
```
bryceosterhaus: Should this be an interface?

packages/clay-multi-select/src/index.tsx
bryceosterhaus: I think we used to have something similar in multi-select when it was only high-level. I think it was called "identifier", which defaulted to "id"

packages/clay-multi-select/src/index.tsx
bryceosterhaus: Should we be allowing customization of the whole menu, or just the menu items? Typically I think we would do a `*Renderer` type prop, which is just a function that takes the typical props expected for an item.

I think I would expect this to be only for items as `autocompleteItemRenderer`, the whole menu seems strange to me.

matuzalemsteles: Yeah I thought about that if it was just the Item, but the filter rule is stuck in our implementation so I let all the Menu content change so that they can have filter control.

I thought about leaving `*Component` to follow the props pattern elsewhere in our components.

packages/clay-multi-select/src/index.tsx
matuzalemsteles: Hmm, do you think it's better to call it `locator` or `identifier`? I think I'm fine with both of them even though I lean towards `locator` 😁.

packages/clay-multi-select/src/index.tsx
matuzalemsteles: As it was just a few simple props, I had only thought to keep `type` but I will switch to `interface` so as not to get away.

packages/clay-multi-select/src/index.tsx
bryceosterhaus: locator is fine with me. i would have leaned toward identifier if it was how to decipher the uniqueness of the item. like `classPK` or something. But since this is for both label and value, I think locator is fine.

packages/clay-multi-select/src/index.tsx
bryceosterhaus: What if we did `autocompleteMenuRenderer` or just `menuRenderer` and it be a function? I lean towards using renderer functions rather than components since they are slightly more flexible but also allow for a component as well.

packages/clay-multi-select/src/index.tsx
bryceosterhaus: I think generally we should keep all props as interfaces for the sake of consistency.

packages/clay-multi-select/src/index.tsx
matuzalemsteles: Hmm, I'm not sure if I understand. But do you mean that going with `*Component` in the name of the prop can have a tone that can add Class components?

bryceosterhaus: I would go with `*Renderer` and have its type be something like...
```tsx
(props: any) => JSX.Element | React.ComponentType<any>
```

That make sense? This way users can pass either a component(function or class), or even just a function that returns jsx.

packages/clay-multi-select/src/index.tsx
bryceosterhaus: lol no worries, I actually just caught myself doing that the other day.

packages/clay-multi-select/src/index.tsx
matuzalemsteles: Yeah makes sense. I changed it to `menuRenderer` and kept the type `React.FunctionComponent<IMenuRendererProps>;` I think it covers these cases.

matuzalemsteles:
@bryceosterhaus updated.

PR #2699, opened 2019-11-05
feat(@clayui/form): add `forwardRef` support for `<ClayForm />`

### LINKED ISSUE #2694: ClayForm does not accept a ref
## What is your proposal?

Add a forwardRef to `ClayForm` so one can get a ref to the DOM form element

For better or for worse, using the form element (forcing a submit) is a common enough approach that could be supported

Use standard `<form>` when one needs to get a ref to it

matuzalemsteles:
Fixes #2694

### DIFF
--- a/packages/clay-form/src/Form.tsx
+++ b/packages/clay-form/src/Form.tsx
@@ -103,20 +103,19 @@ const FeedbackIndicator = React.forwardRef<
 	)
 );
 
-const ClayForm: React.FunctionComponent<
+const ClayForm = React.forwardRef<
+	HTMLFormElement,
 	React.HTMLAttributes<HTMLFormElement>
-> & {
-	FeedbackGroup: typeof FeedbackGroup;
-	FeedbackIndicator: typeof FeedbackIndicator;
-	FeedbackItem: typeof FeedbackItem;
-	Group: typeof Group;
-	Text: typeof Text;
-} = ({children, ...otherProps}) => <form {...otherProps}>{children}</form>;
+>(({children, ...otherProps}, ref) => (
+	<form {...otherProps} ref={ref}>
+		{children}
+	</form>
+));
 
-ClayForm.FeedbackGroup = FeedbackGroup;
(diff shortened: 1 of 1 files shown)


packages/clay-form/src/Form.tsx
```diff
@@ -103,15 +103,21 @@ const FeedbackIndicator = React.forwardRef<
...
 	Group: typeof Group;
 	Text: typeof Text;
-} = ({children, ...otherProps}) => <form {...otherProps}>{children}</form>;
+};
+
+const ClayForm = React.forwardRef(({children, ...otherProps}, ref) => (
+	<form {...otherProps} ref={ref}>
+		{children}
+	</form>
+)) as Form;
```
bryceosterhaus: Instead of using `as ...`, I think we could start using `Object.assign()` like i did here in this PR, https://github.com/liferay/clay/pull/2700/commits/5af7b9dd5163f7f7351c24fe57317b88028f2fd6#diff-c6dc8bae2d26a6f2f689a012da29b157.

This will help react-docgen, since they currently dont support `as`

matuzalemsteles:
@bryceosterhaus updated.

bryceosterhaus:
@matuzalemsteles can you rebase off of latest? I just want to see the docs preview with the latest changes I made.

matuzalemsteles:
@bryceosterhaus updated, I think we'll have to separate components inside `Form.tsx` in files. It seems that react-docgen gets lost when it has too many components in the same file.

bryceosterhaus:
@matuzalemsteles it will only look for the default export, and not the namespaced ones. The namespaces ones have to be imported via their own file. We could update our API plugin to allow for namespaced exports as well, maybe something like `<div>[APITable "clay-form/src/Form.tsx#FeedbackGroup"]</div>`

I'll put this in a separate issue though.

This PR LGTM

matuzalemsteles:
@bryceosterhaus  Oh thanks for the information, I think it would be nice to add `<div>[APITable "clay-form/src/Form.tsx#FeedbackGroup"]</div>`.

PR #2700, opened 2019-11-05
docs: update auto-generated docs

bryceosterhaus:
I've been hitting my head against a wall with our autogenerated docs. I believe this is an improvement on what we have so far.

Notes:

- I had to add some changes to react-docgen to work better with TS. Thus, I forked it into our project . The goal would be to go back to the upstream version whenever they release a stable 5.0.0.
- Instead of a list for our APIs, I switched to a table. Some of the default values look wacky, but I think its better than what we had before.
- Refactored types. react-docgen was choking on some of the ways we had defined our types, I re-worked it a bit so that it doesnt error. I took advice from [this guy](https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/issues/165) on using object.assign, it seems to look cleaner as well.

Let me know what you think. I think this is a step in the right direction, but still have progress to be made.

I think we can re-evaluate out docs generation whenever 5.0.0 is released for react-docgen. I created an issue asking them when it might be ready, so just waiting for a response on that. https://github.com/reactjs/react-docgen/issues/399


matuzalemsteles:
hey @bryceosterhaus i'm not sure about using tables to list the api's, it's a bit bad to see in a mobile view and also to the desktop the columns are squeezed.

Just one question, have you made any direct changes to the code or is it just using the react-docgen master branch? maybe we can just use git submodules?

bryceosterhaus:
This is the change I made for react-docgen. https://github.com/reactjs/react-docgen/pull/400

As far as the UI for mobile view and table vs list. My reasoning was primarily focused on desktop use since API tables probably arent used much on mobile, since dev is done on desktop. Also, I thought the table make it more concise and clear what the types are and what the values are. The list tended to be much longer for components with  a lot of props and was hard to read all of them.

Maybe we could get some outside eyes for table vs list. cc: @jbalsas @wincent @diegonvs  Any preference on this API table vs our current one?

jbalsas:
I would favour tables in this case... it certainly makes consuming the API easier...

wincent:
> As far as the UI for mobile view and table vs list. My reasoning was primarily focused on desktop use since API tables probably arent used much on mobile, since dev is done on desktop.
>
> Maybe we could get some outside eyes for table vs list.

Sounds like the table is a good trade-off.

Issue #2716, opened 2019-11-06
Docs API tables only work for default exports


bryceosterhaus:
Currently our API generation only works for default exports. We should update our plugin to allow for named exports, as well as namespaced components(`<Foo.Bar />`).

Possibly thinking we could do...

Default:
`[APITable "clay-foo/src/Foo.tsx"]`

Named:
`[APITable "clay-foo/src/Foo.tsx#Bar"]`

Namespaced:
`[APITable "clay-foo/src/Foo.tsx(Bar)"]`

Not sold on the examples, but something like that would be beneficial.

PR #2717, opened 2019-11-07
chore: scrape API table for a specific specific named component

bryceosterhaus:
Fixes https://github.com/liferay/clay/issues/2716

Note, we specify the component by using its name via `some/file.tsx#FooBar`

Also, `react-docgen` doesn't seem to accurately pick up displayName of some components. This may be because of using arrow functions, but im not quite sure. This is why I explicitly added displayName to the form components

### DIFF
--- a/clayui.com/content/docs/components/forms/form.mdx
+++ b/clayui.com/content/docs/components/forms/form.mdx
@@ -62,4 +62,26 @@ Use `ClayForm.Text` for creating a caption on Form Groups.
 
 ## API
 
+### Form
+
 <div>[APITable "clay-form/src/Form.tsx"]</div>
+
+### Form.FeedbackGroup
+
+<div>[APITable "clay-form/src/Form.tsx#FeedbackGroup"]</div>
+
+### Form.FeedbackIndicator
+
+<div>[APITable "clay-form/src/Form.tsx#FeedbackIndicator"]</div>
+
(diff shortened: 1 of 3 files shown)


clayui.com/content/docs/components/forms/form.mdx
```diff
@@ -62,4 +62,26 @@ Use `ClayForm.Text` for creating a caption on Form Groups.
 
 ## API
 
+### Form
+
 <div>[APITable "clay-form/src/Form.tsx"]</div>
+
+### Form.FeedbackGroup
+
+<div>[APITable "clay-form/src/Form.tsx(FeedbackGroup)"]</div>
```
matuzalemsteles: It seems a bit odd to see the component specification like this `"clay-form/src/Form.tsx(FeedbackGroup)"`, maybe because it looks like I see a function being called 😅. I think using `#` would be better, it would be almost the same as `svg:use` API for example.

```
<svg class="lexicon-icon lexicon-icon-drop" focusable="false" role="presentation">
	<use href="/images/icons/icons.svg#drop" />
</svg>
```

bryceosterhaus: Oh thats a good point. Ill change to `#`

bryceosterhaus:
@matuzalemsteles updated

PR #2721, opened 2019-11-08
feat(@clayui/table): WIP add DataTable component

### LINKED ISSUE #2212: Create high-level table component
likely should have:
- predefined markup for headers
- configurable row renderer via `rowRenderer` prop
- default row markup
- paginated

We may also want to consider a few different types of defaults for rows.

Also should consider inline-editable fields

matuzalemsteles:
Fixes #2212

This is in progress but I would like to have a different view for it, the idea is to have a component that can handle the data easily and allows little work to work with the data. I was wondering if the component name was a `ClayTableWithData` but for a moment I see that this component can handle a lot of customization too so I went with `DataTable`.

It's quite small but it has an API that can be mapped to print data very easily, based on a schema that is defined by `headers`, something like what we did in 2.x.

The developer has two customization entries, either cells using prop `cellRenderers` or using `render props` with `children` to handle all rendering. The effect we have is that the component will take care of normalizing the data, at the moment this is a big advantage but the idea is that we can implement more details that facilitate the manipulation with the data but at the same time allow easy customization. See the example in the Storybook.


packages/clay-table/stories/index.tsx
```diff
@@ -534,4 +534,71 @@ storiesOf('Components|ClayTable', module)
...
+		const Checkbox: React.FunctionComponent<any> = ({value, ...props}) => (
+			<ClayTable.Cell {...props}>
+				<ClayCheckboxWithState
+					aria-label="Select row"
+					checked={value}
+				/>
+			</ClayTable.Cell>
+		);
+
+		const cellRenderers = {
```
jbalsas: How are `cellRenderers` meant to be selected? Looking at code, it looks like `type` should do the trick, but I don't see any such field in the sample model here...

matuzalemsteles: The field above image and checkbox: https://github.com/liferay/clay/pull/2721/files#diff-3bf0ae6ed7a94d91023cf99cb7dc1ff2R540-R548

Is that what you are saying? `type` is in the prop of `header`.

jbalsas: Yes, I didn't see that... I was expecting to see the renderer in the cell... it makes sense that type is set per column... I wonder if we'll ever have the case where we want cells in different rows to have different renderers...

matuzalemsteles: Hmm, I'm not sure if we would get a case down, probably not.

jbalsas: Probably not.

One thing that will probably be needed is a `rowRenderer` concept and probably also the need for `cellRenderer` to control more than just the content. 

It's pretty common to either want a different `tr` with specific things (to customize a table) and a similar thing can be said for cells, where one might want to add/remove classes to the `td`, for example.

It's possible that this can already be achieved through `otherProps`, but we should validate the proper behaviour

matuzalemsteles: Yeah, essentially people can only control the whole cell with cellRenderer. I also thought about this a bit and people might want to add more things to the `tr` elements. The idea is that dev can override the default markup just by passing a new "layout" to children, for example:

```jsx
<ClayDataTable
	cellRenderers={cellRenderers}
	headers={headers}
	rows={rows}
>
	{(headers, rows, getCell) => (
		<Table>
			<Table.Head>
				<tr className="class-name">
					{headers.map(({header, key, ...props}) => (
						<Table.Cell
							{...Omit(props, ['type'])}
							headingCell
							key={key}
						>
							{header}
						</Table.Cell>
					))}
				</tr>
			</Table.Head>
			<Table.Body>
				{rows.map(row => (
					<Table.Row key={row.id}>
						{row.cells.map(({type, ...props}) =>
							getCell(type, props)
						)}
					</Table.Row>
				))}
			</Table.Body>
		</Table>
	)}
</ClayDataTable>
```

I think this approach is very valid, DataTable will deal with normalizing the data and we can also do more with the data in the future and dev will have full control of all the content, it seems very minimal and also decreases the amount of API's that we would have to create it so that we could change anything in the elements.

bryceosterhaus:
Gonna close this for now and can re-open in the future if/when it comes up again

PR #2735, opened 2019-11-18
chore: add codesandbox ci

bryceosterhaus:
I think we can still optimize on this so that codesandbox doesnt take 5 minutes for every build, but I think this is a good start for this.

For example, we need to exclude `clayui.com` from the `build:csb` script. This will help reduce csb build times.


bryceosterhaus:
nvm, codesandbox doesnt like me

PR #2736, opened 2019-11-19
chore: add codesandbox ci

bryceosterhaus:
I think we can still optimize on this so that codesandbox doesnt take 5 minutes for every build, but I think this is a good start for this. One possible next step would be limiting the built packages to only target packages that have changes.


matuzalemsteles:
hey @bryceosterhaus I tried to reset the Netlify cache and run the build again but it still seems to keep giving error in the build. Maybe it's the changes in `yarn.lock`.

bryceosterhaus:
@matuzalemsteles yeah I am waiting to hear back from someone from csb. The errors on the csb build doesn't explain why, so they may need to enable something. It seemed to work fine on my fork.

I'll follow up once I hear from them.

lbogdan:
Hello, there! Looks like it was an one-off issue, I re-ran the build manually, and it looks like it worked. Can you please test the built packages and generated sandboxes and confirm everything is OK?

bryceosterhaus:
Thanks @lbogdan for re-running it. As far as these one-off errors, is there a way for us as users to actually diagnose the issue or know when to re-run? I encountered this on my own fork but with a different PR, wasn't sure if there was some way we could contribute back to this or debug ourselves.

lbogdan:
For now, not really. But if it happens again, you can ping me either on this issue, or at hello_at_codesandbox.io.

PR #2737, opened 2019-11-19
chore(@clayui/popover): Adds a missing property knob on ClayPopover storybook example

diegonvs:
Related to #2738


bryceosterhaus:
We just need to verify that `show` is actually used first. See https://github.com/liferay/clay/issues/2738

bryceosterhaus:
I dont think this is needed anymore since we have stories with and without show now. https://github.com/liferay/clay/pull/2931

Issue #2758, opened 2019-12-02
MultiSelect suggestions rendered via a custom menu renderer are unnecessarily filtered


4lejandrito:
In Sharing we are [passing a custom menu renderer](https://github.com/liferay/liferay-portal/blob/7308c0145d85f26aac3d9f34c293d13d978639f3/modules/apps/sharing/sharing-web/src/main/resources/META-INF/resources/sharing/js/Sharing.es.js#L202) to the `multi-select` component. This [custom menu renderer](https://github.com/liferay/liferay-portal/blob/7308c0145d85f26aac3d9f34c293d13d978639f3/modules/apps/sharing/sharing-web/src/main/resources/META-INF/resources/sharing/js/Sharing.es.js#L311-L356) won't filter the results (see https://github.com/brianchandotcom/liferay-portal/pull/81519/commits/98346cb2ed30bde8f1daca350b8b3d8ecafd867c) since they already come filtered from the server. 

However the `multi-select` component [is still filtering the elements](https://github.com/liferay/clay/blob/95653bed8ec1784e387fe3ac4bd91cf00619fdd8/packages/clay-multi-select/src/index.tsx#L188) to set the active state which hides the suggestions when there are no elements passing the filter. 

We trust the server suggestions and don't need the frontend to do any matching.

bryceosterhaus:
Can you provide a codesandbox example to showcase the issue? From my understanding, whether the items are filtered on the client or the server, the component shouldn't behave any differently.

Say you have three items returned from the server `['Flintstones', 'Frozen 1', 'Frozen 2']` for the input value of `f`. The multi-select component should run the match on the name and return the same list of items.

The PR you sent breaks existing functionality since it will always display an active dropdown, whether there is an input value or not.

bryceosterhaus:
I am testing with https://codesandbox.io/s/crazy-edison-hdvu1

4lejandrito:
Hi @bryceosterhaus,

Thanks for looking into this.

I'll try to explain our use case a bit better: 

1. We set a custom renderer that **does not** filter the results.
2. The user types a query.
3. The server return results that we trust. They **may not** contain the query the user types (think about synonyms for example). For this we modified the component to [trigger a request to the server every time the query changes](https://github.com/brianchandotcom/liferay-portal/pull/81519/commits/28f806c3d30e853d48de78e0d3be59e8755b16dc) as part of https://github.com/brianchandotcom/liferay-portal/pull/81519. 
[...]

PR #2759, opened 2019-12-02
The component should not filter the list

4lejandrito:


4lejandrito:
Hi,

This is the bare minimum we need for our use case. However that filtering should somehow be moved to the default menu renderer instead of just removed. I leave it up to you guys.

Thanks!

4lejandrito:
#2758

Issue #2722, opened 2019-11-09
Clay Dropdown is misaligned in a modal after scrolling


thektan:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## Is there an example you can provide via codesandbox.com?
https://codesandbox.io/s/great-fog-qjhu7

## What are the steps to reproduce?
Scroll down, open the modal, and open the dropdown. 

You might have to adjust the filler to scroll less to see the misaligned dropdown, otherwise if you scroll down a lot, the dropdown will be completely misaligned off the page.

## What is the expected result?
The dropdown to be aligned regardless of scrolling down before opening the modal.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.1.0  |
| React | 16.9.0  |

PR #2764, opened 2019-12-03
fix(drop-down): use dom-align instead of metal-position

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/2722


yarn.lock
```diff
@@ -15046,7 +15051,7 @@ metal-keyboard-focus@^2.0.0:
     metal "^2.0.0"
     metal-events "^2.2.3"
 
-metal-position@^2.1.0, metal-position@^2.1.2:
+metal-position@^2.1.0:
```
bryceosterhaus: note that metal-position is no longer a dependency for clay v3, but it is still used on clayui.com via tooltip v2

bryceosterhaus:
cc @thektan check out the CSB link above `fragrant-rgb-dlskl` to verify your issue is fixed

wincent:
We might want to create a task to move off of metal-position in liferay-portal too, so that we don't have two very similar dependencies floating around in there (for example, in frontend-js-tooltip-support-web and possibly elsewhere).

thektan:
@bryceosterhaus Thanks for the fix! Looks good! 👌

bryceosterhaus:
@wincent whats the best way for us to additionally make this change in portal? Should I create a sibling issue somewhere, or should I just go ahead and make the change for portal? Thinking for future fixes like this in the future as well.

wincent:
@bryceosterhaus I don't think the changes need to go in atomically as a single change so it should be fine to create a sibling issue in Jira. Probably an IFI? If the change is trivial I would just do it; in this case it doesn't seem like it would be trivial though, so we can just create the IFI and then with the help of @esanzp and @jbalsas figure out the priority of it.

bryceosterhaus:
@matuzalemsteles good catch on those issues. I just added a fix for both, let me know if you see any issues.

wincent:
I created https://issues.liferay.com/browse/IFI-1234 to track the migration away from metal-position inside liferay-portal itself.

bryceosterhaus:
oh my bad, I created an issue but forgot to link it here. https://issues.liferay.com/browse/IFI-1226

I'll close my issue and just link to yours

wincent:
And my JSSR (Jira Search Success Rate) is close to 0% so I didn't even try searching for a ticket before creating mine. 😂

PR #2767, opened 2019-12-03
Extracting filter to a prop

4lejandrito:
#2758


4lejandrito:
See https://github.com/liferay/clay/pull/2770.

PR #2769, opened 2019-12-03
fix: prevent backspace from navigating backwards

### LINKED ISSUE #1448: ClayMultiSelect: When you press backspace in input to focus on the tag the browser returns to the previous page in Firefox

julien:
fixes #1448

### DIFF
--- a/packages/clay-multi-select/src/index.tsx
+++ b/packages/clay-multi-select/src/index.tsx
@@ -219,6 +219,10 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 
 			const {keyCode} = event;
 
+			if (keyCode === BACKSPACE_KEY && !sourceItems.length) {
+				event.preventDefault();
+			}
+
 			if (inputValue && DELIMITER_KEYS.includes(keyCode)) {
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
Just started reviewing :)

packages/clay-multi-select/src/index.tsx
```diff
@@ -219,6 +219,10 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 
 			const {keyCode} = event;
 
+			if (keyCode === BACKSPACE_KEY && !sourceItems.length) {
```
matuzalemsteles:
```suggestion
			if (keyCode === BACKSPACE_KEY && !inputValue) {
```

@julien I think we'd better check `inputValue`, so we will ignore it only when the input is empty. That seems to work better.

julien:
@matuzalemsteles thanks for the suggestion

matuzalemsteles:
@julien I will merge and release this package today and we can use LPS to send this new version to the portal.

julien:
Thanks @matuzalemsteles! /cc @4lejandrito

PR #2668, opened 2019-10-29
Fix ClayModal story

brunofarache:

PR #2777, opened 2019-12-06
chore: use fragments instead of empty args

bryceosterhaus:
After looking at https://github.com/liferay/clay/pull/2668, I think it's a bit more clear to use fragments instead of empty arguments. This should at least help readability.


bryceosterhaus:
fairly trivial. going to merge

PR #2778, opened 2019-12-09
feat(@clayui/data-provider): WIP Suspense Support

matuzalemsteles:
This is experimental is just a WIP, but I've been doing some testing to see what we need to change to add support for `<Suspense />`. See the demo in the Storybook, try turning off the network to see other states.

> 🚨: The tests are broken! Once we reach some conclusion about this I will have to refactor some of them. 

To work better with our cache engine, having a `link` that accepts a function greatly undermines our cache validator and prevents us from making further improvements. I've been thinking of depreciating this functionality and adding a new prop that passes the function that will replace the default `fetch`.

```js
import {fetch} from 'frontend-js-web';

useResource({
    link: 'https://clay.data/',
    fetcher: fetch
})
```

In addition it can help us get better support for `Suspense` as we need to catalog promises so we can retrieve when making the request. The `link` and `variables` are important for generating a key for this.

There are still some changes that need to be made to this code to work well with the other use cases. The vast majority is still supported here. Leave your thoughts.

### DIFF
--- a/packages/clay-data-provider/src/types.ts
+++ b/packages/clay-data-provider/src/types.ts
@@ -169,6 +169,11 @@ export interface IDataProvider {
 	storageMaxSize?: number;
 
 	/**
+	 * Flag to enable `useResource` integration with suspense.
+	 */
+	suspense?: boolean;
+
+	/**
 	 * Variables are analyzed and converted to be passed as parameters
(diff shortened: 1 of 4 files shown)


bryceosterhaus:

Issue #2577, opened 2019-10-03
[Feedback] Future for data visualization


matuzalemsteles:
Hey guys, this is an issue to start a thread about the future of data visualization in Clay (ClayCharts), we are wondering how you use Charts today, what your expectations were and what your pains were about the current Charts with Billboard.js.

An important point to note, dealing with charts is complex and requires a learning curve, we have many libraries out there that deliver different data visualization values, we want to help you build charts quickly according to Lexicon specifications and provide documentation so you can build more complex charts.

We're planning to make significant changes to the Charts, whether it's library shifting, improving APIs, writing new components... based on what we'll be seeing. These are some of the points we want to get with ClayCharts:

- Performance
- Flexibility
- Consistency with Lexicon
- Readable API (We will do our best to strike a balance between flexibility and readability)
- Best Documentations - This includes best practices, use cases, samples, and possible tutorials.

So let your thoughts on this, especially talk about the problems you had with him, the more details the better. To help you clear your mind, start by talking about:

- Pains with ClayCharts
- Expectations
- Use cases
- Tell us what your team is planning for the future about charts.

Spend a little time to devote a few words of what you have in mind that can help us build something better and make better decisions about it, please share this issue with everyone who worked with ClayCharts or mark them on here.

matuzalemsteles:
I'm tagging some people that I know worked with the charts. Please leave your thoughts: @interaminense @fcodias @FabioDiegoMastrorilli @oliveiraaraujo @phcp @kienD

@julien @jbalsas @bryceosterhaus please mark other people I may have forgotten and also leave your thoughts 🙂.

interaminense:
Do you guys think to change the chart lib? Are there other libs that I think are better than **Billboard.js**

Until today we face some issues with billboard.js that I can list here:

* Performance issue when we insert a large amount of data in a line chart for example;
* There is a bug when manipulating the axis values ​​(we don't allow the chart to control) that when updating the data, the axis isn't updated; So to solve the problem, we have to destroy the chart and build again, causing performance issues;
* If we want to override the default chart tooltip, we have to convert the tooltip (component) to a string and moving to the chart;
* We have an Analytics list that can display more than 10 charts on the screen and these 10 charts can have up to 90 points on each chart, and all these charts are customized using the billboard API, this is causing repaint and reflow on screen issues.

[...]

PR #2779, opened 2019-12-09
chore(demos): add recharts examples to storybook

bryceosterhaus:
I spent some time re-creating our existing charts using `recharts` and overall the experience was really nice. Recharts was super easy to use and highly configurable. I haven't approached the predictive or geomap charts yet, I am going to look into those now.

One question I have for people, is how would you expect us to offer the "clay" portion of charts? Is exporting colors and potentially custom components such as tooltips, legends, and axis enough? Or do you expect something else? I am hesitant to wrap rechart components and export them with our own default config. Check out the storybook for clay-recharts(name subject to change) and let me know what you all think.

### DIFF
--- a/packages/demos/package.json
+++ b/packages/demos/package.json
@@ -18,7 +18,11 @@
 		"@clayui/panel": "*",
 		"classnames": "^2.2.6",
 		"react-dnd": "^10.0.2",
-		"react-dnd-html5-backend": "^10.0.2"
+		"react-dnd-html5-backend": "^10.0.2",
+		"recharts": "1.8.5"
+	},
+	"devDependencies": {
+		"@types/recharts": "1.8.5"
 	},
 	"version": "0.0.1"
 }
--- a/packages/demos/src/recharts/Tooltip.tsx
+++ b/packages/demos/src/recharts/Tooltip.tsx
@@ -1,0 +1,59 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as React from 'react';
+import {TooltipPayload, TooltipProps} from 'recharts';
+
+const defaultRenderer = (item: TooltipPayload) => {
+	if (item.payload) {
+		item = item.payload;
+	}
+
+	return (
+		<div key={item.name}>
+			<span
+				style={{
+					backgroundColor: item.fill || (item as any).stroke,
+					display: 'inline-block',
+					height: 10,
+					marginRight: 6,
+					width: 10,
+				}}
+			/>
+
+			{`${item.name}: ${item.value}`}
+		</div>
+	);
+};
+
+interface IProps extends TooltipProps {
+	labelRenderer?: (val?: string | number) => React.ReactNode;
+	itemRenderer?: (val: any) => React.ReactNode;
+}
+
+const ClayRechartsTooltip: React.FunctionComponent<IProps> = ({
+	active,
+	label,
+	payload,
+	labelRenderer = val => val,
+	itemRenderer = defaultRenderer,
+}) => {
+	if (active) {
+		return (
+			<div className="popover" style={{position: 'static'}}>
+				<div className="popover-header">{labelRenderer(label)}</div>
+
+				<div className="popover-body">
+					{payload && payload.map(itemRenderer)}
+				</div>
+			</div>
+		);
+	}
+
+	return null;
+};
+
+export default ClayRechartsTooltip;
--- a/packages/demos/src/recharts/config.ts
+++ b/packages/demos/src/recharts/config.ts
@@ -1,0 +1,28 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const COLOR_MAP = {
+	Dothraki: '#AF78FF',
+	Kashtark: '#FFD76E',
+	Lannister: '#FF5F5F',
+	Martell: '#50D2A0',
+	Mormont: '#FFB46E',
+	Stark: '#4B9BFF',
+	Targaryen: '#9CE269',
+	Tarly: '#5FC8FF',
+	Tyrell: '#FF73C3',
+};
+export const COLORS = [
+	COLOR_MAP.Stark,
+	COLOR_MAP.Mormont,
+	COLOR_MAP.Lannister,
+	COLOR_MAP.Martell,
+	COLOR_MAP.Tyrell,
+	COLOR_MAP.Targaryen,
+	COLOR_MAP.Dothraki,
+	COLOR_MAP.Kashtark,
+	COLOR_MAP.Tarly,
+];
--- a/packages/demos/src/recharts/index.tsx
+++ b/packages/demos/src/recharts/index.tsx
@@ -1,0 +1,10 @@
+/**
+ * © 2018 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayRechartsTooltip from './Tooltip';
+import {COLORS} from './config';
+
+export {COLORS, ClayRechartsTooltip};
--- a/packages/demos/stories/Recharts.tsx
+++ b/packages/demos/stories/Recharts.tsx
@@ -1,0 +1,756 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import '@clayui/css/lib/css/atlas.css';
+import {ClayRadioGroup, ClayRadio} from '@clayui/form';
+import {storiesOf} from '@storybook/react';
+import moment from 'moment';
+import * as React from 'react';
+import {
+	BarChart,
+	Bar,
+	ReferenceArea as ReferenceAreaTest,
+	CartesianGrid,
+	ComposedChart,
+	Legend,
+	Line,
+	LineChart,
+	XAxis,
+	YAxis,
+	Tooltip,
+	Cell,
+	RadialBarChart,
+	RadialBar,
+	PieChart,
+	Pie,
+	PolarAngleAxis as PolarAngleAxisTest,
+	Scatter,
+	ScatterChart,
+	Area,
+	AreaChart,
+} from 'recharts';
+
+import {ClayRechartsTooltip, COLORS} from '../src/recharts';
+
+// Temporary hack until Recharts gets proper types exported
+const PolarAngleAxis = PolarAngleAxisTest as any;
+const ReferenceArea = ReferenceAreaTest as any;
+
+const DATA = [
+	{
+		amt: 2400,
+		name: 'Page A',
+		pv: 2400,
+		uv: 4000,
+	},
+	{
+		amt: 2210,
+		name: 'Page B',
+		pv: 1398,
+		uv: 3000,
+	},
+	{
+		amt: 2290,
+		name: 'Page C',
+		pv: 9800,
+		uv: 2000,
+	},
+	{
+		amt: 2000,
+		name: 'Page D',
+		pv: 3908,
+		uv: 2780,
+	},
+	{
+		amt: 2181,
+		name: 'Page E',
+		pv: 4800,
+		uv: 1890,
+	},
+	{
+		amt: 2500,
+		name: 'Page F',
+		pv: 3800,
+		uv: 2390,
+	},
+	{
+		amt: 2100,
+		name: 'Page G',
+		pv: 4300,
+		uv: 3490,
+	},
+];
+
+storiesOf('Demos|Recharts', module)
+	.add('line', () => (
+		<LineChart data={DATA} height={300} width={500}>
+			<CartesianGrid />
+			<XAxis dataKey="name" />
+			<YAxis />
+			<Tooltip />
+			<Legend />
+			<Line dataKey="pv" stroke={COLORS[0]} />
+			<Line dataKey="uv" stroke={COLORS[1]} />
+		</LineChart>
+	))
+	.add('bar', () => (
+		<BarChart data={DATA} height={300} width={500}>
+			<CartesianGrid />
+			<XAxis dataKey="name" />
+			<YAxis />
+			<Tooltip content={<ClayRechartsTooltip />} />
+			<Legend />
+			<Bar dataKey="pv" fill={COLORS[0]} />
+			<Bar dataKey="uv" fill={COLORS[1]} />
+		</BarChart>
+	))
+	.add('combo', () => (
+		<ComposedChart data={DATA} height={300} width={500}>
+			<CartesianGrid />
+			<XAxis dataKey="name" />
+			<YAxis />
+			<Tooltip content={<ClayRechartsTooltip />} />
+
+			<Legend
+				align="right"
+				layout="vertical"
+				verticalAlign="middle"
+				wrapperStyle={{paddingLeft: 24}}
+			/>
+
+			<Bar dataKey="uv" fill={COLORS[1]} />
+			<Line dataKey="pv" stroke={COLORS[0]} type="monotone" />
+		</ComposedChart>
+	))
+	.add('pie', () => {
+		const data = [{name: 'A', value: 30}, {name: 'B', value: 70}];
+
+		const customLabel = ({percent}: any) =>
+			`${(percent * 100).toFixed(0)}%`;
+
+		return (
+			<PieChart height={400} width={400}>
+				<Legend />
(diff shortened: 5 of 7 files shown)


julien:
Hey @bryceosterhaus, 

I think this is actually an improvement compared to what we had before in the sense that `recharts` seems much more "React friendly" than `billboard.js` - The only thing I was initially worried about is if we would be able to use the same features defined by Lexicon such as custom shapes, patterns and colors but this all seems possible. 

> One question I have for people, is how would you expect us to offer the "clay" portion of charts? Is exporting colors and potentially custom components such as tooltips, legends, and axis enough?

I also think that as a first step that would be enough.

bryceosterhaus:
I went ahead and updated some of the examples and built out a custom Clay tooltip.

@matuzalemsteles I tried to take the approach of providing Clay specific components such as `ClayLine`, however it quickly grew very complicated. For example, if we have multiple lines in a chart, we need to make sure they iterate through the list of default Clay colors. I investigated a few routes to do this, but they all became complicated quickly, so I almost want to lean on the side of giving examples instead of baked out components. If we wanted to build it out, it would require us to have some wrapper around Recharts that would contain a global store and keep track of the number of components and assign them proper colors and shapes. 

I'm thinking of just providing the colors and Tooltip for now, and then slowly iterate where teams need it. If we err on the side just creating a lot of examples, it gives us flexibility and we can help coach teams on best ways to implement a given chart. I think its going to be pretty common for teams to have drastically different use cases.

cc @jbalsas what are your thoughts on this?

jbalsas:
Hey @bryceosterhaus, 

> I tried to take the approach of providing Clay specific components such as ClayLine, however it quickly grew very complicated. For example, if we have multiple lines in a chart, we need to make sure they iterate through the list of default Clay colors. I investigated a few routes to do this, but they all became complicated quickly, so I almost want to lean on the side of giving examples instead of baked out components. 

I understand the feeling. My initial reaction is that as it is right now, with this change we're losing more than gaining... 

We should be 200% sure that at least recharts solves the problems that were brought up at https://github.com/liferay/clay/issues/2577 and also that we make it super simple to build consistent/accessible charts (see https://liferay.design/articles/2018/building-charts-for-multiple-products for some reference and context).

Off the top of my head we want to make sure that our new approach is:

Feature | Check
--- | ---
Easily composable from low level | ✅ 
Performant (ie: 20 charts with 2000 points together) | 
Themable |
Customizable |
Consistent |
Accessible |

> If we wanted to build it out, it would require us to have some wrapper around Recharts that would contain a global store and keep track of the number of components and assign them proper colors and shapes.

I would say this shouldn't be so bad... inevitably, whoever using this will end up making some wrappers, otherwise they'd need to make sure to pick the proper colors over and over, show axis and whatnot...

What if high-level mostly just worked with `DATA`?

```javascript
<LineChart data={DATA} height={300} width={500}>
    <CartesianGrid />
    <XAxis dataKey="name" />
    <YAxis />
    <Tooltip />
    <Legend />
    <Line dataKey="pv" stroke={COLORS[0]} />
    <Line dataKey="uv" stroke={COLORS[1]} />
</LineChart>
```

could be:

```javascript
<ClayLineChart data={DATA} height={300} dataKeys=["pv", "uv"] xDataKey="name" width={500} />

ClayLineChart = ({dataKeys, xAxisKey, ...otherProps}) => {
    return (
        <LineChart ...otherProps>
            <CartesianGrid />
            <XAxis dataKey="{xDataKey}" />
            <YAxis />
            <Tooltip />
            <Legend />
            {dataKeys.map(key, index) => <Line dataKey="{key}" stroke={COLORS[index]} />}
        </LineChart>
    );
}
```

We could try to model a bit more and map config to components and support children, although I wouldn't push for this right now.

```javascript
const chartConfig = {
    cartesianGrid: true,
    data: DATA
    height: 300,
    legend: {
        iconType: "wye"
        layout: "vertical"
    },
    tooltip: <CustomTooltip />,
    width: 500
};

<ClayLineChart config={chartConfig}>
    <ReferenceLine x="Page C" stroke="green" label="Min PAGE" />
</ClayLineChart>
```

Just to reiterate, to me key elements we're looking for (in order of importance) should be:
- Performance
- Clear OOTB offering 
- Composability

Similar to what we do with `billboard.js`, we could offer raw `recharts` plus our wrappers for simplicity.

What do you think?

bryceosterhaus:
@jbalsas at this point I think I would want to avoid any high-level components, if we start with just offering the low-level and giving some patterns, I think it will then help us establish patterns for future high-level components we might want. I think we struggled with billboard because it was too much config and really hard to customize small things. So far in my experience with just using recharts, its already been 10x easier and more fun to use than our implementation of billboard ever was.

I agree with you on the priorities but I would probably go about implementing it in phases. By doing this, I think we give ourselves flexibility and also helps us grow in the direction of need, rather than creating a bunch of wrapper before hand. I would imagine the phases looking like..

1. Give devs the thumbs up to use `recharts` and then expose an easily consumable configuration.
2. As devs find themselves running into common issues, we would help provide a solution.
3. As common solutions are solved, we start to provide high-level components for them to easily consume for their use case.

Would this make sense to you? I am curious how other teams would feel by just offering that first phase right now. From my experience as a user of Charts, I think just giving them recharts is already a huge headache removed.

cc: @kienD, I'm curious what your team would think about this.

jbalsas:
Hey @bryceosterhaus!

> I agree with you on the priorities but I would probably go about implementing it in phases.

I don't see anything wrong with that 😉 

> 1. Give devs the thumbs up to use recharts and then expose an easily consumable configuration.

I would want to do: **0a. Triple check recharts covers our performance requirements**
I would want to do: **0b. Check with legal if `recharts` use and redistribution is ok**

Then I would prefer if we swap the order in 1 to: -> **1. Expose an easily consumable configuration and give the devs the thumbs up to use recharts**. 

This last part worries me a little bit though... we'll want to lock down a specific version but that's also brought us some problems in the past, so we need to tread lightly there, but since `Analytics Cloud` is likely the only one to start with this, we can leave it open and up to them.

> 2. As devs find themselves running into common issues, we would help provide a solution.

I'm not particularly fond of putting the burden of proof on product teams. I know based on your experience `recharts` should be a much better, but still, exposing them to this raw library could endanger their progress on other features.

I would suggest **we take it on ourselves to apply this at least in one real project collaborating with the responsible team** so we find some of the issues and uncover the pain points (and see the patterns emerge).

> 3. As common solutions are solved, we start to provide high-level components for them to easily consume for their use case.

This should be fine, granted that we've developed real common solutions on our own.

How does that sound?

bryceosterhaus:
Yeah that sounds pretty good to me. My overall goal is to discover the needs first before creating anything high-level. I also don't anticipate being able to make a 1-for-1 copy of billboards styles, not sure how much of an issue this is though.

> 0a. Triple check recharts covers our performance requirements
> 0b. Check with legal if recharts use and redistribution is ok

I'll try to create a performance example today. As for as legal goes, who is the best person to ping about this?

> I'm not particularly fond of putting the burden of proof on Product Teams. I know based on your experience recharts will be much better, but still, exposing them to this raw library could endanger their progress on other features.

Ideally we would be able to iterate sooner than later. Meaning that hopefully with even AC moving to recharts, or at least slowly adopting them. We would start to see patterns. Right now, the hard part about creating high-level charts is that we are creating based off of "unknown" needs and really the only way to figure out what those needs are, is by having product teams bring those to us.

> I would suggest we take it on ourselves to apply this at least in one real project collaborating with the responsible team so we find some of the issues and uncover the pain points (and see the patterns emerge).

I'd be more than happy to take this on and work with a product team to revamp their charts. Is there a particular team that has a smaller charts foot print than AC that we could tackle?

jbalsas:
> I pinged Matija about the legal portion of this

Excellent... You should create an inbound licensing ticket in JIRA . See our [React-dnd Usage](https://issues.liferay.com/browse/FOSS-52) for reference.

> I'd be more than happy to take this on and work with a product team to revamp their charts. Is there a particular team that has a smaller charts foot print than AC that we could tackle?

We're [using LineChart in Workflow Metrics](https://github.com/liferay/liferay-portal/blob/master/modules/dxp/apps/portal-workflow/portal-workflow-metrics-web/src/main/resources/META-INF/resources/js/components/process-metrics/completion-velocity/VelocityChart.es.js#L12). This could be a good starting point, but I presume we'll still need to tackle AC for a bigger picture.

I think Commerce might be using some, but maybe through jsp tags... @FabioDiegoMastrorilli, @marco-leo, could you confirm and weigh in if necessary?

FabioDiegoMastrorilli:
> > I pinged Matija about the legal portion of this
> 
> Excellent... You should create an inbound licensing ticket in JIRA . See our [React-dnd Usage](https://issues.liferay.com/browse/FOSS-52) for reference.
> 
> > I'd be more than happy to take this on and work with a product team to revamp their charts. Is there a particular team that has a smaller charts foot print than AC that we could tackle?
> 
> We're [using LineChart in Workflow Metrics](https://github.com/liferay/liferay-portal/blob/master/modules/dxp/apps/portal-workflow/portal-workflow-metrics-web/src/main/resources/META-INF/resources/js/components/process-metrics/completion-velocity/VelocityChart.es.js#L12). This could be a good starting point, but I presume we'll still need to tackle AC for a bigger picture.
> 
> I think Commerce might be using some, but maybe through jsp tags... @FabioDiegoMastrorilli, @marco-leo, could you confirm and weigh in if necessary?

At the moment we don't use taglibs to wrap charts, but I'm pretty sure we've used them via js module in our machine learning related modules. @gcmznt surely knows more than me :)

kienD:
> @jbalsas at this point I think I would want to avoid any high-level components, if we start with just offering the low-level and giving some patterns, I think it will then help us establish patterns for future high-level components we might want. I think we struggled with billboard because it was too much config and really hard to customize small things. So far in my experience with just using recharts, its already been 10x easier and more fun to use than our implementation of billboard ever was.
> 
> I agree with you on the priorities but I would probably go about implementing it in phases. By doing this, I think we give ourselves flexibility and also helps us grow in the direction of need, rather than creating a bunch of wrapper before hand. I would imagine the phases looking like..
> 
>     1. Give devs the thumbs up to use `recharts` and then expose an easily consumable configuration.
> 
>     2. As devs find themselves running into common issues, we would help provide a solution.
> 
>     3. As common solutions are solved, we start to provide high-level components for them to easily consume for their use case.
> 
> 
> Would this make sense to you? I am curious how other teams would feel by just offering that first phase right now. From my experience as a user of Charts, I think just giving them recharts is already a huge headache removed.
> 
> cc: @kienD, I'm curious what your team would think about this.

@bryceosterhaus, we're open to using recharts as a standalone charting library for now.  I tried to do a quick conversion of a couple of our charts from billboardjs to recharts and had a relatively easy time doing so.  

It seems like it'll be easier to develop with recharts due to it's flexibility.

bryceosterhaus:
@jbalsas  Checkout this for a crude implementation for a performance test of multiple charts. I also reached out to AC team for data that they were having issues with. https://deploy-preview-2779--storybook-clayui.netlify.com/?path=/story/components-clayrecharts--performance-test

bryceosterhaus:
anyone else have any primary concerns? If so, ill continue to try and address examples within our storybook.

matuzalemsteles:
My view on this is that the biggest pain I saw and received complaints about using the charts was the work that had to be done to make the chart look like Lexicon, this seems to be one of our main issues. Before flexibility, even with recharts it seems easier, people will still have to work on each custom chart to look like Lexicon. I think our cases already exist https://liferay.design/lexicon/core-components/charts/charts/. Does it make sense to you?

bryceosterhaus:
@matuzalemsteles not quite sure what you mean. Are you saying it's easier with recharts or with billboard to make the chart look like lexicon?

matuzalemsteles:
Sorry @bryceosterhaus for the confusion. In fact, I'm saying that although rechart is easier than billboard, people still needed to customize it to get Lexicon cases for those who want to follow, so I think we need to provide a minimum of styles. so people don't have to have this work. I think that was one of the points that people complained about in the beginning.

bryceosterhaus:
@matuzalemsteles Ah I see. Right now the plan is for me to help AC refactor some of their charts using recharts, part of my goal in that is to try and figure out the common uses we can extract to clay itself. After going through AC I think I will have a better idea on which direction to go. Ill keep you updated.

matuzalemsteles:
@bryceosterhaus I see. I think the AC cases are good for this moment but take into consideration the small Liferay teams.

jbalsas:
Hey @matuzalemsteles, that's a great point!

One thing bothered me, though:

> I think we need to provide a minimum of styles. so people don't have to have this work. I think that was one of the points that people complained about in the beginning.

The old version of charts was a very close implementation of the Lexicon charts... do you have a particular example in which this was not the case and the teams struggled to make it look like it? Maybe they struggled with other type of charts outside of those defined by Lexicon and the library and designed elsewhere? Just curious here :)

matuzalemsteles:
Hey @jbalsas, the design problems involving our old graphics compared to Lexicon are the fine details that we didn't have. For example: The [Line Chart](https://liferay.design/lexicon/core-components/charts/chart-line/) graph compared to the [Line Chart in Clay](http://v2.clayui.com/docs/components/charts/basic/line-chart.html), some things that Design notes:

- The edges of the graph that surround are not rounded and do not exist at all ends
- The font size of the labels on the axes, apparently are not the same
- We have some small spacing problems between some elements

These are some things that are observable in what is expected and what we have, there are more in other graphs that we have that are as basic charts. From the point of view of who was using the graphics, they would not need to customize these details, when there was a CEREBRO team here in Brazil (now AC) they complained a lot about these details, especially the design (Rafael was at the time), he was very careful in how the data was presentable, so he found a lot of missing details in how the axes are presented, spacing between them and etc ... In some meetings I had with CEREBRO/Workflow it was that they expected that these graphics delivered by Clay already had these details that were predicted by the Lexicon designers, in their point of view, they wanted a high-level component with all these customizations.

jbalsas:
Hey @matuzalemsteles, thanks for the context! I totally agree with this and this is a good lesson for us when working in the future version of charts.

That being said, there's always some time where design meets reality that we need to get better at addressing. Past is past, so no need to go over this. We should make a good job at producing high-quality out-of-the-box charts... and if they don't 100% match the designs, maybe what we can do is work with Lexicon to adjust them to reality so designers and developers don't see their expectations frustrated.

bryceosterhaus:
@jbalsas @matuzalemsteles what do you guys think of me just moving these examples to the "demos" stories for now?

bryceosterhaus:
I went ahead and moved it to demos. Let me know what you think.

jbalsas:
LGTM @bryceosterhaus! Feel free to merge. 

I've pinged @andresfulla as they were attempting to use our current charts in DXP and finding some trouble with it. We can explore this as well for them rather than having to jump from one version to the other.

matuzalemsteles:
Sounds good for me! Can we release on top of a `v4.0.0-rc.1` version? since we are making a major change.

jbalsas:
As far as I can tell, changes here only involve the site, so we don't even need a new release. We should also be careful with major releases, as that will have a very big impact in our DXP integration and our maintenance costs. If we can roll this out in a semantically responsible way, that'd be ideal :)

We've attempted to use this in DXP and we've had some issues configuring the build including some not-so-pleasant surprises, like recharts dependencies on lodash and node.js native `events` package. This is to say, rolling this out might be a bit more tricky than it seems.

@izaera and @andresfulla might be able to share some more details

bryceosterhaus:
Yeah I don't think think we need to do any versioning for this, this is just an example to showcase for our demos and we won't actually be releasing anything.

> like recharts dependencies on lodash and node.js native events package

Hmm, this seems worrisome, I'll investigate a bit.

andresfulla:
Along with @izaera I included `recharts` in the portal, as I am using the library without Clay. We did it like this: https://github.com/andresfulla/liferay-portal/commit/85dda548e34dc41dc46df168ad4df947830b2bb7

We're using a different `loadash` and providing a node `events` polyfill`.

My idea is to get it on master in the following weeks.

wincent:
> My idea is to get it on master in the following weeks.

Tangential comment on that commit @andresfulla [here](https://github.com/andresfulla/liferay-portal/commit/85dda548e34dc41dc46df168ad4df947830b2bb7#r37172941).

andresfulla:
Just in case it might be relevant. I'm getting a warning about `componentWillReceiveProps` been renamed to `UNSAFE_componentWillReceiveProps` coming from `Line` and `LineChart`.

bryceosterhaus:
hmm, you see those errors on this page? https://storybook.clayui.com/?path=/story/demos-recharts--line

andresfulla:
no, I get it in the portal. On `development` and with `recharts@1.8.5`

bryceosterhaus:
@andresfulla hmm, thats odd we don't see it on the clay side. Regardless, it shouldn't be a huge issue unless they remove those lifecycles in future versions, which I think would be pretty unlikely at this point.

PR #2787, opened 2019-12-12
chore(ci): only build src and not types for ci

bryceosterhaus:
Going to keep this PR open to see how build time decreases with each change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
 	"scripts": {
 		"build": "lerna run build && lerna run build:types",
 		"build:csb": "lerna run --no-bail --ignore 'clayui.com' build && lerna run --no-bail --ignore 'clayui.com' build:types",
-		"checkDeps": "lerna run build && node scripts/check-deps/check-dependencies.js",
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
@wincent looks like just removing types from the build process speeds it up quite a bit.

wincent:
Yeah, that's a nice speed up: 7 minutes, down from 12 minutes, although I imagine there is probably a fair bit of variability between runs. The other checks seem to run in about 6 minutes (not sure, because only some of them show the actual time), so seeing as they run in parallel and not serially, 7 minutes is already pretty good.

Just one doubt: is there any conceivable way that building the types might fail even if the rest of the build succeeds?

bryceosterhaus:
> Just one doubt: is there any conceivable way that building the types might fail even if the rest of the build succeeds?

I don't think there is any chance this would happen since we are using the the same tsconfig. There might be a super edge case but I don't think we would need to worry about it. Additionally, we would still be verifying that it builds when we publish, so we would run into issues there if it did break.

bryceosterhaus:
reopening to trigger ci again

PR #2835, opened 2020-01-07
feat(popover): add align functionality

### LINKED ISSUE #2768: Popover positioning
## What is your proposal?
Implement positioning for ClayPopover so it can be CSS-independent of its parent.

I quickly implemented this with `React.createPortal` and `metal-align`.

```javascript
const Popover = props => {
  return ReactDOM.createPortal(
    <PopoverComponent {...props} />,
    document.body
  );
};

const PopoverComponent = ({anchor, children, ...rest}) => {
  const popRef = useRef(null);

  React.useLayoutEffect(() => {
    Align.align(popRef.current, anchor, Align.Right, false);
    // eslint-disable-next-line react-hooks/exhaustive-deps
}, []);

    return (
      <ClayPopover alignPosition="right" ref={popRef} {...rest}>
        {children}
      </ClayPopover>
    );
  };
```

I tried to use the ClayPopover as it is. But I had some trouble because the component that triggers the popover is inside of an element with `overflow: hidden`. So the Popover won't be entirely displayed.
I'm sure that if this comes from Clay out-of-the-box it will be better implemented and address corner cases.

Manual implementation per use case.

bryceosterhaus:
fixes #2768

### DIFF
--- a/packages/clay-popover/src/index.tsx
+++ b/packages/clay-popover/src/index.tsx
@@ -4,7 +4,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {ClayPortal} from '@clayui/shared';
 import classNames from 'classnames';
+import domAlign from 'dom-align';
 import React from 'react';
 
 export const ALIGN_POSITIONS = [
@@ -15,8 +17,27 @@ export const ALIGN_POSITIONS = [
 	'bottom-left',
 	'bottom-right',
 	'left',
+	'left-top',
+	'left-bottom',
 	'right',
+	'right-top',
+	'right-bottom',
 ] as const;
+
+const ALIGNMENTS_MAP = {
+	bottom: ['tc', 'bc'],
+	'bottom-left': ['tl', 'bl'],
+	'bottom-right': ['tr', 'br'],
+	left: ['cr', 'cl'],
+	'left-bottom': ['br', 'bl'],
+	'left-top': ['tr', 'tl'],
+	right: ['cl', 'cr'],
+	'right-bottom': ['bl', 'br'],
+	'right-top': ['tl', 'tr'],
+	top: ['bc', 'tc'],
+	'top-left': ['bl', 'tl'],
+	'top-right': ['br', 'tr'],
+} as const;
 
 interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
@@ -35,6 +56,13 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	show?: boolean;
 
 	/**
+	 * React element that the popover will align to when clicked.
+	 */
+	trigger?: React.ReactElement & {
+		ref?: (node: HTMLButtonElement | null) => void;
+	};
+
+	/**
 	 * Content to display in the header of the popover.
 	 */
 	header?: React.ReactNode;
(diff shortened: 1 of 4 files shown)


packages/clay-popover/src/index.tsx
```diff
@@ -15,15 +17,41 @@ export const ALIGN_POSITIONS = [
...
 	'right',
+	'right-top',
+	'right-bottom',
 ] as const;
 
+const ALIGNMENTS_MAP = {
+	bottom: ['tc', 'bc'],
+	'bottom-left': ['tl', 'bl'],
+	'bottom-right': ['tr', 'br'],
+	left: ['cr', 'cl'],
```
kresimir-coko: How come we mix and match the nomenclature of these attributes? Some have quotes and others don't. I understand that the ones that don't have them don't need them, but why not favor consistency? 

> Yeah, but your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should.

julien: I guess the idea is "use it when it's needed don't when it's not".

packages/clay-popover/src/index.tsx
```diff
@@ -43,26 +71,58 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
...
+		const [show, setShow] = React.useState(false);
+
+		const alignNodeRef = React.useRef<HTMLElement | null>(null);
+		const popoverRef = React.useRef<HTMLElement | null>(null);
+
+		if (!ref) {
+			ref = popoverRef as React.Ref<HTMLDivElement>;
+		}
+
+		React.useEffect(() => {
```
kresimir-coko: Is the preferred practice to use `React.useSomething` instead of importing `useSomething` from React, as I see examples of both when searching through the source code (mostly of `React.useSomething`)?

packages/clay-popover/src/index.tsx
bryceosterhaus: Yeah I think we would tend to follow juliens comment. Also we have an eslint rule in this project(`quote-props`) that auto formats it to not use quotes when it isn't necessary.

Here is an example eslint error
Screen Shot 2020-01-10 at 11 25 34 AM

packages/clay-popover/src/index.tsx
bryceosterhaus: TBH, it's not a huge issue if you go one way or the other, so long as its consistent for the component or file you are working in. Personally I would opt for `React.*` and I forget where this conversation had happened, but I think we decided to try an opt for `React.*` to avoid confusion with our other custom hooks `use*`.

packages/clay-popover/src/index.tsx
wincent: The only reason we haven't changed it to "consistent-as-needed" yet:

https://github.com/liferay/eslint-config-liferay/pull/39

is due to an open bug in Prettier:

https://github.com/prettier/prettier/issues/6064

Once that is resolved this will become somewhat nicer:

```js
const ALIGNMENTS_MAP = {
	'bottom': ['tc', 'bc'],
	'bottom-left': ['tl', 'bl'],
	'bottom-right': ['tr', 'br'],
	'left': ['cr', 'cl'],
        // etc...
};

File: packages/clay-popover/src/index.tsx (hunk 2 above)
Comment (wincent) [reply packages/clay-popover/src/index.tsx]: (Not this repo, but...) In liferay-portal we mostly just use `useState` and friends directly, because the number of built-in hooks is small and nobody actually gets confused about them coming from React. Less clutter in the code too, less likely to have to awkwardly wrap lines etc.

File: packages/clay-popover/src/index.tsx [hunk 3]
```diff
@@ -15,15 +17,41 @@ export const ALIGN_POSITIONS = [
...
 interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
 	 * Position in which the tooltip will be aligned to the element.
 	 */
 	alignPosition?: typeof ALIGN_POSITIONS[number];
 
+	/**
+	 * React element that the popover will align to when clicked.
+	 */
+	alignNode?: React.ReactElement & {
```
matuzalemsteles: It might be better to rename the prop to `trigger`, at first glance I thought I needed to just pass the element ref and will also create a consistency with DropDown that uses the same approach to alignment.

bryceosterhaus:
@matuzalemsteles updated

PR #2838, opened 2020-01-09
fix(css): allow newline characters to display properly for tooltips

### LINKED ISSUE #2836: ClayTooltipProvider doesn't support newline characters

- Create an element within `ClayTooltipProvider` and pass a string with `\n` newline character(s) in the `title` attribute
- Observe that newlines are not visible in the rendered content, though they show when inspecting the DOM

Newline characters should be honored 

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.0.5  |
| React |         |

bryceosterhaus:
fixes #2836

### DIFF
--- a/packages/clay-css/src/scss/components/_tooltip.scss
+++ b/packages/clay-css/src/scss/components/_tooltip.scss
@@ -1,5 +1,7 @@
 .tooltip-inner {
 	@include box-shadow($tooltip-box-shadow);
+
+	white-space: pre-line;
 }
 
 .tooltip-arrow {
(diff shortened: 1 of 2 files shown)


kresimir-coko:
Hey @bryceosterhaus Just a quick question related to this. When we make any functionality changes, basically `feats` related to components, even if they are in `css` files we need to add an entry in `stories` for them so they show up in the Storybook?

bryceosterhaus:
@kresimir-coko its not required to always make a new story, but I typically try to create stories to showcase requests that people have had. For this case, I thought it was helpful to add a story since the corresponding issue was related to the TooltipProvider component. So essentially its nice to add for people to see, but not required for every feat.

bryceosterhaus:
@pat270 I'll let you verify this before I merge. I always second guess myself when it comes to css.

pat270:
@bryceosterhaus it LGTM!

PR #2839, opened 2020-01-10
feat(clayui.com): make example codes collapsible

kresimir-coko:


clayui.com/src/components/clay/Editor.js
```diff
@@ -109,10 +114,27 @@ const Editor = ({code, disabled = false, imports, preview = true, scope}) => {
...
+
 			<div className="gatsby-highlight">
 				<div style={{padding: '10px'}}>
-					{imports && <LiveEditor disabled value={imports} />}
-					<LiveEditor />
+					<ClayButtonWithIcon
+						className="btn-collapse"
+						displayType="unstyled"
+						onClick={() =>
+							setCollapseCode(collapseCode => !collapseCode)
```
bryceosterhaus: Callback functions are great in `set*`, but in this case we can actually just do this

```suggestion
							setCollapseCode(!collapseCode)
```

clayui.com/src/components/clay/Editor.js
```diff
@@ -109,10 +114,27 @@ const Editor = ({code, disabled = false, imports, preview = true, scope}) => {
 					<LiveError />
 				</div>
 			)}
+
 			<div className="gatsby-highlight">
 				<div style={{padding: '10px'}}>
-					{imports && <LiveEditor disabled value={imports} />}
-					<LiveEditor />
+					<ClayButtonWithIcon
```
bryceosterhaus: We might as well also add the copy button here too to keep consistent with the css code snippets. What do you think?

clayui.com/src/components/clay/Editor.js
```diff
@@ -109,10 +114,27 @@ const Editor = ({code, disabled = false, imports, preview = true, scope}) => {
...
+						small
+						spritemap={spritemap}
+						symbol={collapseCode ? 'angle-right' : 'angle-down'}
+						title={collapseCode ? 'Expand' : 'Collapse'}
+					/>
+
+					{!collapseCode && (
+						<>
+							{imports && <LiveEditor disabled value={imports} />}
+							<LiveEditor />
```
bryceosterhaus: Typically we add new lines between sibling elements or components, so we would need one above this line.

clayui.com/static/js/code-collapse.js
```diff
@@ -0,0 +1,35 @@
+if (document.body) {
```
bryceosterhaus: The formatting seems a little off in this file. Can you run prettier again on it?

clayui.com/static/js/code-collapse.js
```diff
@@ -0,0 +1,35 @@
...
+		const target = event.target;
+        const targetClassList = target.classList;
+        
+		if (
+			targetClassList.contains('btn-collapse') ||
+			targetClassList.contains('icon-collapse') ||
+			(target.tagName === 'use' &&
+				target.parentNode.classList.contains('icon-collapse'))
+		) {
+			const codeContainer = target.closest('.code-container');
```
bryceosterhaus: `.closest` isn't supported in IE, so we should either add a polyfil or use something else.

reference: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

clayui.com/static/js/code-collapse.js
```diff
@@ -0,0 +1,35 @@
...
+			if (codeContainer) {
+				const gatsbyCode = codeContainer.querySelector(
+					'code[class^="gatsby-code-"]'
+				);
+
+                if (gatsbyCode) {
+                    const collapseBtn = codeContainer.querySelector('.btn-collapse');
+
+                    if (gatsbyCode.style.display === 'none') {
+                        collapseBtn.style.transform = 'rotate(0)';
```
bryceosterhaus: Rather than rotating the icon, we should be doing the same thing that we do on the React side and just change the icon reference to angle-right

clayui.com/plugins/gatsby-remark-use-clipboard/index.js
```diff
@@ -12,9 +12,22 @@ module.exports = ({markdownAST}) => {
 			return;
 		}
 
-		node.value = `<div class="code-container">
-            <button class="btn btn-sm btn-copy" title="Copy"><svg class="lexicon-icon"><use xlink:href="/images/icons/icons.svg#paste"></use></svg></button>
-            ${node.value}
-        </div>`;
+		node.value = `
+			<div class="code-container">
+				<button class="btn btn-sm btn-collapse" title="Collapse">
```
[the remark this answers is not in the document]
kresimir-coko: That makes sense, my other idea was to hide the `copy` button when the code is collapsed but that might not be necessary.

clayui.com/src/components/clay/Editor.js
kresimir-coko: That's how it was before I started working on it so I left it like that. But yea, I agree, empty lines between siblings is the way to go :)

clayui.com/src/components/clay/Editor.js
kresimir-coko: So, adding the `copy` button here would go along with the goal of having a single implementation of collapsing. I didn't know how all of these implementations were connected and there isn't a `copy` button on the React implementation of code examples.

What I'm trying to say is that, yes, I'll move the copy button here which will add the ability to copy code from all examples, not just markup/css ones.

clayui.com/static/js/code-collapse.js
kresimir-coko: Ah yes, I'll have to work around that then. How dirty are we `allowed` to get with chaining and DOM queries?

kresimir-coko:
I'm making progress on the requested changes, but I'm stuck at figuring out where/how the `Editor` component is rendered inside the docs. Here's what I've gathered so far:
- `<Editor>` is called in all of the individual component docs
- `docs.js` has DOM for an `advanced` and `simple` tab, which are the `React Component | CSS/Markup` tabs
- `advanced` tab is somehow defining that `<Editor>` should be rendered inside it
- `simple` tab is somehow defining that `<Editor>` isn't being rendered inside it

What I haven't figured out is how to decide the DOM structure of the `simple` tab, because I need to have `<Editor>` be called there in order to have a single implementation of collapsing, and the copy button being defined in one place

clayui.com/src/components/clay/Editor.js
bryceosterhaus: Its typically not strictly enforced and likely a handful of instances of gone into upstream, but if I notice it I'll likely recommend it. But overall, not a huge deal, its more of a portal SF rule that we try to also adopt here.

clayui.com/static/js/code-collapse.js
bryceosterhaus: What exactly do you mean by "dirty"? I'd probably err on the side of not a ton of chaining, but it really depends on the situation. We don't have a hard and fast rule.

This is an example of closest we use in tooltip, https://github.com/liferay/clay/blob/master/packages/clay-tooltip/src/TooltipProvider.tsx#L80

bryceosterhaus:
Hey @kresimir-coko, so I just took a look and I think the big difference is that the "css/markup" tabs are rendered with by `.md` files, whereas the "React" tabs are rendered via `.mdx`. In the `.mdx` files we are explicitly importing a component with the `<Editor />`. So for the "css/markup" tab, we have a few options...

1. Re-write the gatsby plugin(use-clipboard) to utilize the `Editor` component.
2.  We could rewrite those to use `.mdx` and import the editor components.
3. Not do anything and hold off on refactoring to a later date.

I think I am starting to lean towards #3 since the benefit of refactor isn't a huge payoff at the moment and the complexity of rewrite is a lot. So I am thinking we hold off on combining the two, and continue to just treat the css docs and react code examples separately.

kresimir-coko:
@bryceosterhaus Yea the refactor might be better to hold off, it's gonna take me a while to finish it with my current understanding of the project and Gatsby. We can go through with the initial PR and get the functionality in there and do the refactor at a later date?

bryceosterhaus:
@kresimir-coko yeah I agree. Go ahead and address the other comments I made and update the PR but don't worry about refactoring it to use a single means of rendering the `code-container`.

kresimir-coko:
@bryceosterhaus I've force pushed the commits.

kresimir-coko:
@bryceosterhaus  I've added the additional commit that adds the copy functionality to the button on React examples

Issue #2840, opened 2020-01-10
Working with option groups using ClaySelect


diegonvs:
## What is your proposal?

Use ClaySelectWithOption with [optgroups](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) for grouped options	

## Why would adopting this proposal be beneficial?

Avoiding the user to don't use our High Level

PR #2841, opened 2020-01-10
feat(select): add optgroups for select

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/2840

cc @diegonvs

### DIFF
--- a/packages/clay-form/src/SelectWithOption.tsx
+++ b/packages/clay-form/src/SelectWithOption.tsx
@@ -12,7 +12,14 @@ interface IProps extends React.ComponentProps<typeof Select> {
 	/**
 	 * Options of the select.
 	 */
-	options: Array<React.ComponentProps<typeof Select.Option>>;
+	options: Array<
+		(
+			| React.ComponentProps<typeof Select.Option>
+			| React.ComponentProps<typeof Select.OptGroup>) & {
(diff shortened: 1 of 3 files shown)


packages/clay-form/src/SelectWithOption.tsx
```diff
@@ -12,17 +12,35 @@ interface IProps extends React.ComponentProps<typeof Select> {
 	/**
 	 * Options of the select.
 	 */
-	options: Array<React.ComponentProps<typeof Select.Option>>;
+	options: Array<
+		(
+			| React.ComponentProps<typeof Select.Option>
+			| React.ComponentProps<typeof Select.OptGroup>) & {
+			options?: Array<React.ComponentProps<typeof Select.Option>>;
+			type?: 'group';
```
matuzalemsteles: @bryceosterhaus just a curiosity can you tell if it is possible to create nested groups?

bryceosterhaus: @matuzalemsteles its not possible with our implementation and it also doesnt seem to be in the HTML spec for `<optgroup>`

matuzalemsteles: Sounds good, np.

Issue #2422, opened 2019-09-11
Create drag and drop example


bryceosterhaus:
It might be helpful to include a dnd example inside of our demos folder.

jbalsas:
Added to https://github.com/liferay/clay/milestone/15 so we can prioritize it :)

bryceosterhaus:
Fixed by https://github.com/liferay/clay/pull/2847

PR #2847, opened 2020-01-14
feat(clay-css): Create drag and drop example

kresimir-coko:
I've made a drag and drop example here, but only a suggestion, a draft. 
It doesn't support reordering but only dragging between columns. 
I didn't manage to get Prettier to run on the `drag-drop.js` file so the formatting might be off.

The goal of this PR is to check if my approach is good, if the file structure I made is valid and to know of any groundbreaking mistakes before I proceed too far.

### DIFF
--- a/packages/demos/stories/DragDrop.tsx
+++ b/packages/demos/stories/DragDrop.tsx
@@ -1,0 +1,195 @@
+/**
+ * © 2020 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayList from '@clayui/list';
+import {XYCoord} from 'dnd-core';
+import React, {useState, useCallback, useRef} from 'react';
+import {useDrag, useDrop, DropTargetMonitor} from 'react-dnd';
+
+interface IDraggableListItemProps {
+	/**
+	 * List Item unique identifier
+	 */
+	id: any;
+
+	/**
+	 * The text displayed in the List Item
+	 */
+	text: string;
+
+	/**
+	 * Index of the List Item in order to keep track of it's positioning within the array
+	 */
+	index: number;
+
+	/**
+	 * Function that handles the moving of List Items
+	 */
+	onMove: (dragIndex: number, hoverIndex: number) => void;
+}
+
+interface IDragItem {
+	/**
+	 * List Item unique identifier
+	 */
+	id: string;
+
+	/**
+	 * Index of the List Item in order to keep track of it's positioning within the array
+	 */
+	index: number;
+
+	/**
+	 * Type of data that is allowed to be dragged
+	 */
+	type: string;
+}
+
+const DraggableListItem: React.FunctionComponent<IDraggableListItemProps> = ({
+	id,
+	index,
+	onMove,
+	text,
+}) => {
+	const ref = useRef<HTMLLIElement>(null);
+
+	const [, drop] = useDrop({
+		accept: 'listItem',
+		hover(item: IDragItem, monitor: DropTargetMonitor) {
+			const dragIndex = item.index;
+
+			const hoverIndex = index;
+
+			if (!ref.current || dragIndex === hoverIndex) {
+				return;
+			}
+
+			const hoverBoundingRect = ref.current!.getBoundingClientRect();
+
+			const verticalMiddle =
+				(hoverBoundingRect.bottom - hoverBoundingRect.top) / 2;
+
+			const mousePosition = monitor.getClientOffset();
+
+			const pixelsToTop =
+				(mousePosition as XYCoord).y - hoverBoundingRect.top;
+
+			const draggingUpwards =
+				dragIndex > hoverIndex && pixelsToTop > verticalMiddle * 1.5;
+
+			const draggingDownwards =
+				dragIndex < hoverIndex && pixelsToTop < verticalMiddle / 2;
+
+			if (draggingDownwards || draggingUpwards) {
+				return;
+			}
+
+			onMove(dragIndex, hoverIndex);
+
+			item.index = hoverIndex;
+		},
+	});
+
+	const [{isDragging, itemBeingDragged}, drag] = useDrag({
+		collect: (monitor: any) => ({
+			isDragging: monitor.isDragging(),
+			itemBeingDragged: monitor.getItem() || {id: 0},
+		}),
+		item: {id, index, type: 'listItem'},
+	});
+
+	const isItemBeingDragged = itemBeingDragged.id === id;
+
+	drag(drop(ref));
+
+	const style = {
+		backgroundColor: isItemBeingDragged ? '#EFEFEF' : '',
+		border: isItemBeingDragged ? '2px solid #555555' : '',
+		cursor: 'grab',
+		opacity: isDragging ? 0 : 1,
+	};
+
+	return (
+		<ClayList.Item
+			className="m-2"
+			flex
+			id={`listItem${id}`}
+			key={id}
+			ref={ref}
+			style={style}
+		>
+			<ClayList.ItemField>{`Item ${id}`}</ClayList.ItemField>
+
+			<ClayList.ItemField expand>
+				<ClayList.ItemTitle>{`Item ${id} Title`}</ClayList.ItemTitle>
+				<ClayList.ItemText>{text}</ClayList.ItemText>
+			</ClayList.ItemField>
+		</ClayList.Item>
+	);
+};
+
+const DraggableList: React.FunctionComponent = () => {
+	const [listItems, setListItems] = useState([
+		{
+			id: 1,
+			text: 'Write a cool JS library',
+		},
+		{
+			id: 2,
+			text: 'Make it generic enough',
+		},
+		{
+			id: 3,
+			text: 'Write README',
+		},
+		{
+			id: 4,
+			text: 'Create some examples',
+		},
+		{
+			id: 5,
+			text:
+				'Spam in Twitter and IRC to promote it (note that this element is taller than the others)',
+		},
+		{
+			id: 6,
+			text: '???',
+		},
+		{
+			id: 7,
+			text: 'PROFIT',
+		},
+	]);
+
+	const onMove = useCallback(
+		(dragIndex: number, hoverIndex: number) => {
+			const tempList = [...listItems];
+
+			tempList.splice(dragIndex, 1);
+
+			tempList.splice(hoverIndex, 0, listItems[dragIndex]);
+
+			setListItems(tempList);
+		},
+		[listItems]
+	);
+
+	return (
+		<ClayList className="col-4 m-2">
+			{listItems.map((listItem, index) => (
+				<DraggableListItem
+					id={listItem.id}
+					index={index}
+					key={listItem.id}
+					onMove={onMove}
+					text={listItem.text}
+				/>
+			))}
+		</ClayList>
+	);
+};
+
+export default DraggableList;
--- a/packages/clay-list/src/Item.tsx
+++ b/packages/clay-list/src/Item.tsx
@@ -34,42 +34,49 @@ interface IProps extends React.HTMLAttributes<HTMLLIElement> {
 	header?: boolean;
 }
 
-const ClayListItem: React.FunctionComponent<IProps> = ({
-	action = false,
-	active = false,
-	children,
-	className,
-	disabled = false,
-	flex = false,
-	header = false,
-	...otherProps
-}: IProps) => {
-	const [focus, setFocus] = React.useState(false);
+const ClayListItem = React.forwardRef<HTMLLIElement, IProps>(
+	(
+		{
+			action = false,
+			active = false,
+			children,
+			className,
+			disabled = false,
+			flex = false,
+			header = false,
+			...otherProps
+		},
+		ref
+	) => {
+		const [focus, setFocus] = React.useState(false);
 
-	return (
-		<li
-			{...otherProps}
-			className={classNames(className, {
-				active,
-				focus,
-				'list-group-header': header,
-				'list-group-item': !header,
-				'list-group-item-action': action && !disabled,
-				'list-group-item-disabled': disabled,
-				'list-group-item-flex': flex,
-			})}
-			onBlur={({currentTarget, relatedTarget}) => {
-				if (
-					relatedTarget &&
-					!currentTarget.contains(relatedTarget as Node)
-				) {
-					setFocus(false);
-				}
-			}}
-			onFocus={() => setFocus(true)}
-		>
-			{children}
-		</li>
-	);
-};
+		return (
+			<li
+				{...otherProps}
+				className={classNames(className, {
+					active,
+					focus,
+					'list-group-header': header,
+					'list-group-item': !header,
+					'list-group-item-action': action && !disabled,
+					'list-group-item-disabled': disabled,
+					'list-group-item-flex': flex,
+				})}
+				onBlur={({currentTarget, relatedTarget}) => {
+					if (
+						relatedTarget &&
+						!currentTarget.contains(relatedTarget as Node)
+					) {
+						setFocus(false);
+					}
+				}}
+				onFocus={() => setFocus(true)}
+				ref={ref}
+			>
+				{children}
+			</li>
+		);
+	}
+);
+
 export default ClayListItem;
--- a/packages/demos/package.json
+++ b/packages/demos/package.json
@@ -16,6 +16,9 @@
 		"@clayui/pagination": "*",
 		"@clayui/pagination-bar": "*",
 		"@clayui/panel": "*",
-		"classnames": "^2.2.6"
-	}
+		"classnames": "^2.2.6",
+		"react-dnd": "^10.0.2",
+		"react-dnd-html5-backend": "^10.0.2"
+	},
+	"version": "0.0.1"
 }
--- a/packages/demos/stories/index.tsx
+++ b/packages/demos/stories/index.tsx
@@ -7,7 +7,10 @@
 import '@clayui/css/lib/css/atlas.css';
 import {storiesOf} from '@storybook/react';
 import React from 'react';
+import {DndProvider} from 'react-dnd';
+import Backend from 'react-dnd-html5-backend';
 
+import DragDrop from './DragDrop';
 import FilesPage from './FilesPage';
 import FormPage from './FormPage';
 import ListPage from './ListPage';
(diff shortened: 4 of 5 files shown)


kresimir-coko:
> Hey @kresimir-coko for issue #2422, we actually want to create an example of dnd for the react components within our `clay/packages/demos/stories` folder and not on the `clay-css` side. This would mean that we would want to use `react-dnd` and then some existing clay component for the dragging and reordering, such as a list or cards. Does that make sense?

Alright, misunderstanding on my end, but I don't mind, gave me a chance to see how the clay-css project works. I think the misunderstanding came from the term `demos`

bryceosterhaus:
@kresimir-coko no worries at all! More than anything, just get familiar with clay.

kresimir-coko:
Hey @bryceosterhaus I've pushed new commits replacing the old ones. They're probably not final, especially design-wise, but I wanna see if the implementation fulfills the requirements and use this as an opportunity to ask how the example is supposed to look in the UI.

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+const DraggableListItem: React.FC<IDraggableListItemProps> = ({
+	index,
+	moveListItem,
+	text,
+}) => {
+	const ref = useRef<HTMLDivElement>(null);
+
+	const [, drop] = useDrop({
```
bryceosterhaus: Lets avoid using empty args, some people have had issues building with an empty arg. You might run into eslint complaining about an unused var, in which case we might want to add this rule https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md#args-after-used

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,135 @@
...
+		accept: 'listItem',
+		hover(item: IDragItem, monitor: DropTargetMonitor) {
+			if (!ref.current) {
+				return;
+			}
+
+			const dragIndex = item.index;
```
bryceosterhaus: you could also set a default name in the above arg as `{index: dragIndex}` to avoid having to reset it.

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+			const hoverIndex = index;
+
+			if (dragIndex === hoverIndex) {
```
bryceosterhaus: lets move this up to the ref check `if (!ref.current || dragIndex === hoverIndex)`

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+			const verticalMiddle =
+				(hoverBoundingRect.bottom - hoverBoundingRect.top) / 2;
+
+			const mousePosition = monitor.getClientOffset();
+
+			const pixelsToTop =
+				(mousePosition as XYCoord).y - hoverBoundingRect.top;
+
+			if (dragIndex < hoverIndex && pixelsToTop < verticalMiddle) {
```
bryceosterhaus: it might be helpful to set `dragIndex < hoverIndex && pixelsToTop < verticalMiddle` as a named variable so that reading through the code is a little clearer as to why you would do an early return.

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,135 @@
...
+				return;
+			}
+
+			if (dragIndex > hoverIndex && pixelsToTop > verticalMiddle) {
```
bryceosterhaus: same as above.

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,135 @@
...
+		item: {index, type: 'listItem'},
+	});
+
+	const opacity = isDragging ? 0 : 1;
+
+	drag(drop(ref));
+
+	return (
+		<div ref={ref} style={{opacity}}>
+			<ClayList.Item>{text}</ClayList.Item>
```
bryceosterhaus: Instead of just using this, lets make it a full list like we see in https://storybook.clayui.com/?path=/story/components-claylist--simple. no need for headers though.

package.json
```diff
@@ -94,5 +94,10 @@
 	"resolutions": {
 		"js-beautify": "1.7.5",
 		"event-stream": "3.3.4"
+	},
+	"dependencies": {
+		"immutability-helper": "^3.0.1",
+		"react-dnd": "^10.0.2",
+		"react-dnd-html5-backend": "^10.0.2"
```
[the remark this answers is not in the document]
kresimir-coko: When doing `yarn add react-dnd react-dnd-html5-backend` in the demos package I get this error:
`error An unexpected error occurred: "Cannot read property 'manifest' of undefined".`. That's why I added them to the root package.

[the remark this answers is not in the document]
wincent: I think the real reason is one line above that message:

```
warning Missing version in workspace at "/Users/greghurrell/code/clay/packages/demos", ignoring.
error An unexpected error occurred: "Cannot read property 'manifest' of undefined".
```

This is happening because there is no `version` field in `packages/demos/package.json`. Just add a bogus one (eg. "0.0.1") and it you will be able to add the dependencies locally.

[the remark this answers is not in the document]
kresimir-coko: @wincent interesting how the line that it ignores is the one that causes the error. I thought about doing that but assumed that the version was intentionally omitted. Works now, thanks!

kresimir-coko:
Hey @bryceosterhaus I've worked a bit on the requested changes, leaving the ones I haven't dealt with unresolved. I haven't sent any commits today because I invested a good chunk of time in the React testing workshop. Anyway, I've got a couple questions for you:

1. I couldn't replicate the chunk of code using the `update` method from `immutability-helper`. I did get the list to get spliced properly but `setListItems` wasn't changing the state. I made a temp variable to avoid mutating the `listItems` array to not bypass the restriction of setting state with mutated data.
2. Regarding the look of the demo, you linked `react-beautiful-dnd` as an example of how it should look, but said we should use `react-dnd` for the implementation, so that leaves me confused a bit. I wanted to push an implementation using `react-beautiful-dnd` so we can see both version and choose, but I didn't have enough time for that today. Should I make the implementation using `react-beautiful-dnd` or make `react-dnd` look like `react-beautiful-dnd`?

bryceosterhaus:
Hey, no worries about not getting to it yet, there is no rush at the moment and the testing workshop is beneficial.


For your comments you raised, I hope I can explain a bit more in depth here. Let me know if my responses make sense.
> 1. I couldn't replicate the chunk of code using the update method from immutability-helper. I did get the list to get spliced properly but setListItems wasn't changing the state. I made a temp variable to avoid mutating the listItems array to not bypass the restriction of setting state with mutated data.

For mutating, our main concern is to not mutate anything that has a reference to state or props, however we would be more than able to mutate a cloned item. For example, imagine we have a `user` object that we want to update and set back in state.

```js
// Imagine that `initialUserObject` has many fields.
const [user, setUser] = React.useState(initialUserObject) 

function updateUserName(name) {
	// This creates a new object with no references back to the initial object.
	const newUserInfo = {...user};

	// Mutating is fine since there is no reference to the state object
	// this means that newUserInfo !== user.
	newUserInfo.name = name;

	// Finally set the state of the new user.
	setUser(newUserInfo);
}
```

We can apply this same sort of thinking to the itemsList in our state and avoid having to use the 3rd party library.

> 2. Regarding the look of the demo, you linked react-beautiful-dnd as an example of how it should look, but said we should use react-dnd for the implementation, so that leaves me confused a bit. I wanted to push an implementation using react-beautiful-dnd so we can see both version and choose, but I didn't have enough time for that today. Should I make the implementation using react-beautiful-dnd or make react-dnd look like react-beautiful-dnd?

Sorry, thats my bad on not clarifying. I wasn't intending to communicate that we need to copy beautiful-dnd or use that as our dependency. I was trying to use that as an example for what the interaction would be like. Currently in your existing code, the demo is a little odd because when you release the mouse after dragging, the element looks  like it animates back to its original index. If you look at the example at [react-dnd](https://react-dnd.github.io/react-dnd/examples/sortable/simple), the animation after releasing the mouse looks more accurate. Also, in your demo, if you grab the first index and release, the 2nd card looks like it shoots up to the first spot. See the gif below for example.
image I would suspect this might have something to do with react `key`s in the list, but I'll let you debug that.

Again, just want to reiterate that the goal is not to copy or implement what `react-beautiful-dnd` already offers, but instead for us to showcase how to use `react-dnd` for a simple use case with ClayList and eventually also showcase it with a list of ClayCards in a grid format.

kresimir-coko:
Hey @bryceosterhaus I've fixed the dependencies, rewrote the splicing code that relied on immutability-helper, added some basic styles and made the list items resemble the basic ClayList.

What I didn't do is the animations that happen when the items get moved around, I'm unsure how I can even approach that. I don't think I can do it with CSS because the moving isn't being handled by styles but by positioning in JS, so the other option is to use one of the animation libraries. Being unsure of how to handle that I'm asking you here what do you think how I should approach animating?

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,200 @@
...
+		backgroundColor: '#FFF8C9',
+		border: '2px solid #4B9FFF',
+	};
+
+	drag(drop(ref));
+
+	const opacity = isDragging ? 0 : 1;
+
+	return (
+		<div
```
bryceosterhaus: Lets use `<ClayList>` instead of a div. This way we can make it look like a proper clay list.

packages/demos/stories/DragDrop.tsx
bryceosterhaus: Hey @kresimir-coko could you look into this as well?

packages/demos/stories/DragDrop.tsx
bryceosterhaus: Hey @kresimir-coko could you make this change as well?

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+			const mousePosition = monitor.getClientOffset();
+
+			const pixelsToTop =
+				(mousePosition as XYCoord).y - hoverBoundingRect.top;
+
+			const draggingDownwards =
+				dragIndex < hoverIndex && pixelsToTop < verticalMiddle / 2;
+
+			if (draggingDownwards) {
```
bryceosterhaus: Lets move this below `draggingUpwards ` and that way we can condense the logic to be

```js
if (draggingDownwards || draggingUpwards) {
    return;
} 
```

I think this might help readability a bit

packages/demos/stories/DragDrop.tsx
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			const hoverBoundingRect = ref.current!.getBoundingClientRect();
+
+			const verticalMiddle =
+				(hoverBoundingRect.bottom - hoverBoundingRect.top) / 2;
+
+			const mousePosition = monitor.getClientOffset();
+
+			const pixelsToTop =
```
bryceosterhaus: great names for the variables, helps make it nice and clear.

packages/demos/stories/DragDrop.tsx
bryceosterhaus: Or actually, `<ClayList>` should probably be the parent container to `{listItems.map` in `DraggableList` and then this component primary return should be `<ClayList.Item>`. This was we properly match the semantic markup from clay

packages/demos/stories/DragDrop.tsx
kresimir-coko: I don't understand what exactly you're trying to say here.
```
const dragIndex = item.index;
const hoverIndex = index;
```
Even if you got confused and wanted me to do `{index: hoverIndex}` I'm unsure where in the code to make the change. 
```
const DraggableListItem: React.FunctionComponent<IDraggableListItemProps> = ({
	id,
	index: hoverIndex,
	onMove,
	text,
}) => {
```
Adding `index: hoverIndex˙to the list of params throws an error:
`ERROR in [at-loader] ./packages/demos/stories/DragDrop.tsx:105:14 
    TS18004: No value exists in scope for the shorthand property 'index'. Either declare one or provide an initializer.`

packages/demos/stories/DragDrop.tsx
kresimir-coko:
```
<ClayList>
  <div ref={ref}>
    <ClayList.Item>
```
Is as semantic I can get it without adding `forwardRef` to the ClayList.Item

kresimir-coko:
Hey @bryceosterhaus I've got the latest of requested changes in. Except the `const dragIndex = item.index;` and the `const [, drop] = useDrop({` because I don't know what exactly I'm supposed to change there. Especially the `useDrop` one, perhaps if you clarify. I've left the comment [above](https://github.com/liferay/clay/pull/2847#discussion_r371119944)

packages/demos/stories/DragDrop.tsx
bryceosterhaus: Oh I was thinking of moving it to the functions arguments. at `hover(item: IDragItem, monitor: DropTargetMonitor) {`

But I realized you set the index value at the end of the function so its probably best to leave it as is for now.

packages/demos/stories/DragDrop.tsx
bryceosterhaus: Oh gotcha, this is a good chance for us to add a `fix: ` of adding forwardRef to `ClayList.Item`. I'm sure others will also have a reason to use forwardRefs with claylist.

bryceosterhaus:
For `const [, drop] = useDrop({`, look into adding [this rule](https://eslint.org/docs/rules/no-unused-vars#varsignorepattern) to our eslint config. We can add a pattern like `'^_` which would tell eslint to ignore any variable that starts with an underscore. This gives us a safe bailout to destructure arrays and not use some of the variables, for example `const [_ignored, used] = someArray;`.

wincent:
> look into adding [this rule](https://eslint.org/docs/rules/no-unused-vars#varsignorepattern) to our eslint config

Don't add the rule; it would be redundant — instead, remove [this line](https://github.com/liferay/clay/blob/ffd54cf45c97efb3892ba2aafe4d3ac2c29b633e/.eslintrc.js#L28) from the config. Because the config in this repo extends [eslint-config-liferay](https://github.com/liferay/eslint-config-liferay), you'd then get [the `argsIgnorePattern` set to `^_` by default](https://github.com/liferay/eslint-config-liferay/blob/5fdc0e04aee55ad75acf5d5daae6ef9704e09988/index.js#L46).

kresimir-coko:
@bryceosterhaus I've added the, probably final, commit, one removing the eslint rule and adding an unused variable

bryceosterhaus:
@kresimir-coko can you also add the forwardRef to `ClayList.Item` and then that way we can remove the div we are using for a wrapper.

Also, can you look into the CI failures? It looks like the eslint config still needs some work in order for it to pass. You can verify the ci locally by running `yarn ci`

bryceosterhaus:
@wincent thanks for pointing that out. We may need to adopt it into the specific ts rule in our eslint still, the CI seems to complain about it still.

wincent:
> We may need to adopt it into the specific ts rule in our eslint still, the CI seems to complain about it still.

Ah, I wasn't paying enough attention. Our lint rule uses `argsIgnorePattern` (which is for ignoring function arguments), but what would apply here is `varsIgnorePattern` (which would apply to local variables).

What was wrong with doing this instead?

```diff
diff --git a/packages/demos/stories/DragDrop.tsx b/packages/demos/stories/DragDrop.tsx
index 470f94c38..70e8f769f 100644
--- a/packages/demos/stories/DragDrop.tsx
+++ b/packages/demos/stories/DragDrop.tsx
@@ -56,7 +56,7 @@ const DraggableListItem: React.FunctionComponent<IDraggableListItemProps> = ({
 }) => {
 	const ref = useRef<HTMLDivElement>(null);
 
-	const [_ignored, drop] = useDrop({
+	const [, drop] = useDrop({
 		accept: 'listItem',
 		hover(item: IDragItem, monitor: DropTargetMonitor) {
 			const dragIndex = item.index;
```

Given that we use `@typescript-eslint/no-unused-vars` in addition to the ESlint-builtin `no-used-vars`, it would be kind of annoying to have to applying configuration overrides to both rules.

kresimir-coko:
> What was wrong with doing this instead?

Do you mean omitting the `_ignored` variable and just leaving it empty like it used to be: `const [, drop] = useDrop({`?

wincent:
Yeah. I know @bryceosterhaus said earlier on that "some people have had issues building with an empty arg", but I'd be curious to know what those were.

If destructuring works at all, I'd expect it to work for this case too (especially seeing as we use it so heavily all over Clay).

kresimir-coko:
@bryceosterhaus I've fixed the CI issue, kept the `_ignored` variable so you have the chance to think about [what Greg said](https://github.com/liferay/clay/pull/2847#issuecomment-579664405). I've also reordered the styles objects a bit because the old approach didn't work properly without the `div` wrapper. I've also added `forwardRef` to ClayList.Item.

bryceosterhaus:
@wincent I was referring back to this issue, https://github.com/liferay/clay/pull/2668. It may be slightly different since this case is declaring new variables and not creating a new array.

I actually don't think using or not using this syntax is a huge deal breaker, I'm sure the pr i linked above was potentially an edge case or something wrong on the user environment but I do actually think its a bit easier(entirely subjective) to read by not using `[, foo]` and having named ignores. That said, I still think its a helpful eslint rule to use for future areas of needing an escape hatch for ignored vars. I could be convinced otherwise though.

wincent:
Ah, I see. So the other issue was related to creating an array with "holes" in it. I see that as pretty different from destructuring, and I think we shouldn't shy away from Using The Language™ and expecting our developers to be familiar with what `const [, b] = ...` means just like we expect them to understand other forms of destructuring. Also, I am a bit allergic to config and I haven't personally actually run into a case where I legitimately *needed* to create a variable but ignore it ([this is a counterexample which Pavel asked about though](https://github.com/liferay/liferay-frontend-guidelines/issues/79)). Having said that, totally up to you guys to solve this in the way you prefer.

bryceosterhaus:
@wincent I think your points make sense to me and I'm happy to go that route. You're right in that we shouldn't shy away from using language capabilities that are available to use. As for the "ignore" rule, the only pattern I've encountered where I would use it is the destructuring example where I am intentionally omitting props of an object, although it hasn't been a big issue and I've found other ways around it.

TLDR; remove eslint rule I initially advocated for and go with @kresimir-coko's original design by just using `[, arg] = ...`.

kresimir-coko:
@bryceosterhaus I've removed the eslint rule and the `_ignored` var and squished everything

bryceosterhaus:
@kresimir-coko I did make 2 additional changes, I removed the margins from the list items and I also removed the `key` prop inside of the DraggableListItem component. `key`s are only needed when you are iterating through a list. See https://github.com/liferay/clay/pull/2899/files

Issue #2846, opened 2020-01-13
Running storybook locally results in error


bryceosterhaus:
Running storybook locally results in this error for any component using "warning"

![Screen Shot 2020-01-13 at 2 29 42 PM](https://user-images.githubusercontent.com/6843530/72297706-4e0c6300-3611-11ea-94ab-7b063fc7bfe9.png)

Issue #2775, opened 2019-12-06
TS editor inference and autocomplete doesn't seem to work.


bryceosterhaus:
Check out the example below. The `antd` library correctly allows users to autocomplete its value for `type`. The ClayButton does not work for autocompleting values. Note this is in the editor, not related to the UI.

https://codesandbox.io/s/sad-panini-o3oez

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.x  |

bryceosterhaus:
It seems to work correctly if you add `"esModuleInterop": true,` to your tsconfig

bryceosterhaus:
similar reported issue https://github.com/microsoft/TypeScript/issues/28009

bryceosterhaus:
I think I've hit a bit of a dead end here.

A few possible solutions:
1. Refactor our use of named default exports `React from 'react'` and instead use start syntax of `* as React from 'react'.`
2. Don't worry about it and just tell people to use `"esModuleInterop": true` for proper editor and autocomplete support.

Any other ideas for this or have anyone ran into this? cc @matuzalemsteles  @wincent

julien:
It seems to me @andresfulla reported something about autocomplete in #2235 not 100% sure if it's related though.

wincent:
Using `import * as React from 'react'` doesn't seem like the worst thing in the world, or rather, it seems better than forcing people to use a particular `esModuleInterop` setting (they may have good reasons for not wanting to). If you think that will fix it without any downsides other than requiring us to go back to something that we had before, then I'd say go for it.

FWIW, I think the appeal of avoiding importing the entire namespace may be more theoretical than practical: even if you do tree-shaking, real apps end up using just about all of React somewhere, so treating it as a monolithic bundle is probably not as bad as it sounds.

wincent:
Fun fact: I remember [talking about this back in February](https://github.com/liferay/clay/pull/1548#discussion_r259716545) so I went back to see if anything had changed about the way React declares its exports. It's still the same ([React master, still b87aabd](https://github.com/facebook/react/blob/0cf22a56a18790ef34c71bef14f64695c0498619/packages/react/index.js#L14-L16), [ReactDOM master, still b87aabd](https://github.com/facebook/react/blob/0cf22a56a18790ef34c71bef14f64695c0498619/packages/react-dom/index.js#L14-L16)) and I guess will be for the foreseeable future.

matuzalemsteles:
[...]

PR #2851, opened 2020-01-14
chore: go back to using esModuleInterop since it is causing issues for a developer environment for storybook

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/2846

reopens https://github.com/liferay/clay/issues/2775

Spent the last few days debugging this. There seems to be a conflict between how storybook is processing our imports and how jest and babel do. If we don't merge this, using storybook in development results in an error tying to import `warning`. Whereas if we change all warning imports to `import warning from 'warning';`, it results in test errors. We will need to continue working on #2775


bryceosterhaus:
merging since this is causing issues for local development.

PR #2859, opened 2020-01-16
fix(clay-dropdown): escape special characters in label

### LINKED ISSUE #2856: Dropdown component is not rendering options with escaped special characters

Pass an option with label containing escaped special character (ex.: '2 &gt 1')
It should interpret labels as html special characters. In case of the example above it should render '2 > 1'
## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v2.19.0  |

matuzalemsteles:
Fixes #2856

Just a hack to force the label to be treated as HTML so it can escape special characters.

### DIFF
--- a/packages/clay-dropdown/demos/index.html
+++ b/packages/clay-dropdown/demos/index.html
@@ -211,7 +211,7 @@
 				items: [
 					{
 						href: '#1',
-						label: 'Item 1',
+						label: '&lt;Item 1&gt;',
 						title: 'Item 1',
 					},
 					{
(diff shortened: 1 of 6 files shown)


matuzalemsteles:
@bryceosterhaus would have no problem, because the Metal + Soy deals with it under the hood. Here I am just forcing Metal to treat the `string` so it can escape. It would be equivalent to allowing HTML in the parameters but we would have problems with conditional checking on many of our components would end up rendering unnecessary markups because these parameters are like functions and on a conditional they are `true` and also another problem with `ClayRadio` and `ClayCheckbox` which uses this parameter as well but only accepts `string`.

bryceosterhaus:
@matuzalemsteles sounds good, LGTM then

PR #2870, opened 2020-01-22
fix(clay-dropdown): renders new items values ​​when they change

### LINKED ISSUE #2869: Dropdown's _originalItems is never reset when rendering the dropdown list with other values

## Description:
Liferay's product Forms has a type of field called Select. Select uses ClayDropdown to list their options, and this type of field can be re-rendered depending on other values of the form if the user chooses to. The problem with that is that the list `_originalItems`, only used to display the results of the dropdown's search box, is never reset after its first attribution, therefore when a second set of options come to be rendered, the search still works with the first set of options.

The steps to reproduce involve using Liferay since it has the use case where you can re-render the component without refreshing the page and reloading the entire JavaScript. Here's the original bug found in Liferay DXP:

https://issues.liferay.com/browse/LPS-105813

The search should be made with the most recent rendered set of options.

## Notes:
One possible solution discussed with @matuzalemsteles offline might be exchanging the purpose of the `_originalItems` list and the `items` list in the `ClayDropdownBase.js` file.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v2.19.0  |

matuzalemsteles:
Fixes #2869

### DIFF
--- a/packages/clay-dropdown/src/__tests__/ClayDropdown.js
+++ b/packages/clay-dropdown/src/__tests__/ClayDropdown.js
@@ -203,6 +203,13 @@ describe('ClayDropdown', function() {
 		expect(clayDropdown).toMatchSnapshot();
 	});
 
+	/**
+	 * ClayDropdownBase contains the internal state `_filteredItems` which is the source of the truth to
+	 * perform operations so that the public state `items` is not changed, so as soon as the component is
+	 * created `_filteredItems` is filled with `items` this will cause a new rendering, in the tests this
+	 * behavior causes the Jest to be lost making this testing Flask. Metal probably needs an `act`
+	 * equivalent to that of React so that it can control these renderings in tests.
+	 */
 	it.skip('should keep the expanded updated with that of the dropdown base', () => {
 		jest.useFakeTimers();
 		clayDropdown = new ClayDropdown({
--- a/packages/clay-dropdown/src/__tests__/ClayDropdown.js
+++ b/packages/clay-dropdown/src/__tests__/ClayDropdown.js
@@ -203,7 +203,7 @@ describe('ClayDropdown', function() {
 		expect(clayDropdown).toMatchSnapshot();
 	});
 
-	it('should keep the expanded updated with that of the dropdown base', () => {
+	it.skip('should keep the expanded updated with that of the dropdown base', () => {
 		jest.useFakeTimers();
 		clayDropdown = new ClayDropdown({
 			items: items,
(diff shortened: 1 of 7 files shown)


packages/clay-dropdown/src/__tests__/ClayDropdown.js
```diff
@@ -203,7 +203,7 @@ describe('ClayDropdown', function() {
 		expect(clayDropdown).toMatchSnapshot();
 	});
 
-	it('should keep the expanded updated with that of the dropdown base', () => {
+	it.skip('should keep the expanded updated with that of the dropdown base', () => {
```
bryceosterhaus: We probably shouldn't commit skipped tests right?

wincent: We have [a lint](https://github.com/liferay/eslint-config-liferay/blob/5fdc0e04aee55ad75acf5d5daae6ef9704e09988/index.js#L43) to prevent the committing of `it.only` tests, because those are pretty much always a mistake (akin to leaving in a `debugger`).

But `it.skip` can be legitimate for something that you want to disable *temporarily*, especially if paired with a comment explaining why (eg. if it needs a dependency to be updated or something else). Of course, if possible, better to fix it and make it pass (or remove it if it doesn't make sense).

matuzalemsteles: Hey guys, sorry for not informing why, I tried to circumvent some Metal problems so that these tests would pass but the real problem is that as I am using a new internal state `_filteredItems` to be updated with `items` in the start of component construction for some reason Metal does not do well with a rendering right after being built, this in the tests, makes these tests in specific look like Flask, despite being working very well manually. I think it would be a kind of 'act' for Metal, something similar to what exists in React to deal with these situations. At the moment I didn't want to work on it now and disable testing to move on. In a way we have a test with this same use case passing the ClayDropdownBase.

bryceosterhaus: Ah makes sense. And I agree, committing skips shouldn't be strictly  prohibited, but it might be nice to add a comment block above the test explaining why it is skipped. This would help both the review process and help when feature devs are reading through tests. Even if you just copy/paste your explanation as a comment above the test, I think that would suffice.

matuzalemsteles: Yeah, sure. I'll put the comment in the tests with `skip`

matuzalemsteles:
@bryceosterhaus updated

Issue #2805, opened 2019-12-12
Drop down menu from Nav Bar on Mobile is disruptive on wider pages.


calvinctvo:
## What is your proposal?
Nav Bar should be stationary on the top of the Mobile device
## Why would adopting this proposal be beneficial?
So drop down menu won't be disruptive for the user.
## What are the alternatives to this proposal?
Possibly could keep the code section condense so the page don't have to extend

![DropdownNavbar](https://user-images.githubusercontent.com/51280814/70757572-f4a0d380-1cf4-11ea-9f73-b3d617193e13.gif)

PR #2876, opened 2020-01-25
fix(clayui.com): Adds overflow hidden on body when mobile nav is open

pat270:
issue #2805

This prevents content under nav from scrolling. I'm not sure how to wrap the api table in `table-responsive`.

### DIFF
--- a/clayui.com/plugins/gatsby-remark-api-table/index.js
+++ b/clayui.com/plugins/gatsby-remark-api-table/index.js
@@ -15,13 +15,13 @@ const generateTr = (item, key) => `<tr>
 			${key}
 		</div>
 	</td>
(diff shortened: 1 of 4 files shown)


pat270:
@matuzalemsteles the only place where table overflow should occur is in Table CSS / Markup section. I didn't change the examples to include table-responsive because I wasn't sure if we wanted to show what tables looked like without table-responsive.

matuzalemsteles:
hey @pat270 maybe in the Table session (CSS/Markup) it is not better to put the `table-responsive`? it can get a little bad when you're on mobile navigation, but we can leave it at that if we want to show how it really looks on mobile.

pat270:
@matuzalemsteles I agree thanks!

PR #2889, opened 2020-01-30
feat(modal): add prop for scrollable body

### LINKED ISSUE #2887: ClayModal Body should accept custom className or a scroller

There is no way to provide custom classes to the <ClayModal.Body> component, i.e: 
`<ClayModal.Body className="inline-scroller">` is ignored

The best solution would be that ClayModal shows a scroll when the content in it's body is too large. This is not happening now, and a possible workaround could be to add a custom className to the modal body

### LINKED ISSUE #2888: ClayPaginationBarWithBasicItems doesn't actually implement initialSelectedDelta
`<ClayPaginationBarWithBasicItems />` [claims](https://github.com/liferay/clay/blob/f53e37b1d0776821c1a35d93049cb3b63306bfe4/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx#L99) to support a `initialSelectedDelta` prop, but if you try using it you'll see this in the console:

```
Warning: React does not recognize the `initialSelectedDelta` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `initialselecteddelta` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
          in div (created by ForwardRef)
          in ForwardRef (created by ClayPaginationBarWithBasicItems)
          in ClayPaginationBarWithBasicItems (created by TableWithPagination)
```

Grepping to see where `initialSelectedDelta` is used in Clay, it turns up no results apart from the interface declaration I linked to at the top. Looks like it was removed in fbb914dc2263f0dd9faeee39675d1ea395c765a6.

Noticed this while running `app-builder/app-builder-web` tests in liferay-portal, because they invoke a component that [tries to use this prop](https://github.com/liferay/liferay-portal/blob/d0fd9753a784b6858e59f2c2f059b9a480bb79c4/modules/apps/app-builder/app-builder-web/src/main/resources/META-INF/resources/js/components/table/TableWithPagination.es.js#L39). (It also uses another non-existent prop, `initialActivePage`, renamed to `activePage` in fbb914dc2263f0dd9faeee39675d1ea395c765a6.)

bryceosterhaus:
fixes #2888, fixes #2887

### DIFF
--- a/packages/clay-modal/src/Body.tsx
+++ b/packages/clay-modal/src/Body.tsx
@@ -9,6 +9,11 @@ import React from 'react';
 
 interface IBodyProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
+	 * Flag to indicate if body should be a fixed height with a scrollable overflow.
+	 */
+	scrollable?: boolean;
+
+	/**
 	 * Url to place an iframe in the body of the modal.
 	 */
 	url?: string;
@@ -16,10 +21,12 @@ interface IBodyProps extends React.HTMLAttributes<HTMLDivElement> {
 
 const ClayModalBody: React.FunctionComponent<IBodyProps> = ({
 	children,
+	scrollable,
 	url,
 }: IBodyProps) => (
 	<div
 		className={classNames('modal-body', {
+			'inline-scroller': scrollable,
 			'modal-body-iframe': url,
 		})}
 	>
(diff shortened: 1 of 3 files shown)


bryceosterhaus:
both changes are fairly trivial, going to merge.

Issue #2029, opened 2019-05-29
Create a searchbox for Icon's symbols on CSS > Components > Icons


diegonvs:
Would be nice filter all the available symbols on Icons documentation.
Current view:

<img width="857" alt="Screen Shot 2019-05-29 at 15 22 31" src="https://user-images.githubusercontent.com/7663513/58581299-9d8a5400-8225-11e9-88e8-38de63ca0159.png">

/cc @matuzalemsteles

diegonvs:
As discussed in https://github.com/liferay/clay/pull/2027#discussion_r288566450

bryceosterhaus:
Not sure a search here will solve our problem, unless we alias icons to other types of names. For example, the musical note icon, I would probably type in "music" or "note" to the search, and nothing would appear. I might conclude that it doesn't exist, but the problem was that I needed to search for the "audio" icon.

If we just add a search that filters the icon name, its still gonna be hard to use and no better than the default browser search. So I think if we add this, we need to make sure and add "aliases" for each icon.

matuzalemsteles:
@bryceosterhaus good point, I think if we add it we will have to add aliases that are easy to find for the icons. We'll have a little work, but I think it's worth it.

kresimir-coko:
I'm thinking of picking this one up next, but I couldn't find the `Icons` referenced in `https://github.com/liferay/clay/blob/master/clayui.com/content/docs/components/css-icons.md` as `[foreach Icons]`. I'm still figuring out how to properly search for things in the repo :smile:

matuzalemsteles:
hey @kresimir-coko, don't worry. When we have such syntax in `md` files, we usually create gatsby plugins to handle them. You can find this in particular at `gatsby-remark-foreach-icons`.

kresimir-coko:
Can you confirm @eduardoallegrini ? I was planning on starting work on this today.

kresimir-coko:
[Here](https://gist.github.com/kresimir-coko/011bd192f55529930feb23c9313a0cc6) I've gathered all the icon names and added empty values for aliases. In case someone else didn't wanna do the busywork of compiling them.

drakonux:
Hey guys! yeah @eduardoallegrini maybe this can help. 

[...]

PR #2892, opened 2020-02-03
feat(@clayui.com): Icon Filtering

kresimir-coko:
Hey @bryceosterhaus here's my draft of the Icon Filtering feature we talked about in [the issue](https://github.com/liferay/clay/issues/2029).

A couple notes:
- The filtering via the search input and the icon rendering is in the React component
- I couldn't figure out how to import that React component in the `css-icons.md` file, I did convert it into an `mdx` file but then it wasn't considered a sibling of `icon.mdx` as defined in [icon.mdx:5](https://github.com/liferay/clay/blob/master/clayui.com/content/docs/components/icon.mdx#L5)
- Gonna have to do a cleanup of the `icons.json` file because there are some icons that don't have a file associated with them

I doubt that putting the `<IconSearch />` in the `icon.mdx` is the best place to do it because it duplicates the icon list found in the `CSS/Markup` page. What do I have to do to get it in the proper place, I assume the `css-icons.md`?

### DIFF
--- a/clayui.com/content/docs/components/css-icons.mdx
+++ b/clayui.com/content/docs/components/css-icons.mdx
@@ -3,10 +3,15 @@ title: 'Icons'
 description: 'Icons are a visual representation of an idea and/or action.'
 ---
 
+import IconSearch from '../../../src/components/IconSearch';
+import icons from '../../../static/js/icons.json';
+import flags from '../../../static/js/flags.json';
+
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
 -   [Usage](#usage)
+-   [Icon List](#icon-list)
 -   [Language Flags](#language-flags)
 -   [SVG Icons as Background Images](#svg-icons-as-background-images)
 -   [Clay-SVG-url()](<#clay-svg-url()>)
@@ -16,11 +21,22 @@ description: 'Icons are a visual representation of an idea and/or action.'
 </div>
 
 <div class="clay-site-alert alert alert-info">
-	Check the <a href="https://liferay.design/lexicon">Lexicon</a> <a href="https://liferay.design/lexicon/core-components/icons/">Icons Pattern</a> for a more in-depth look at the motivations and proper usage of this component.
+	Check the <a href="https://liferay.design/lexicon">Lexicon</a>{' '}
+	<a href="https://liferay.design/lexicon/core-components/icons/">
+		Icons Pattern
+	</a>{' '}
+	for a more in-depth look at the motivations and proper usage of this
+	component.
 </div>
 
 <div class="clay-site-alert alert alert-warning">
-	See the implementation of the <a href="/docs/components/icons.html">Icons component in React</a> following the Lexicon <a href="https://liferay.design/lexicon/core-components/icons/">Icons Pattern</a>.
+	See the implementation of the{' '}
+	<a href="/docs/components/icons.html">Icons component in React</a> following
+	the Lexicon{' '}
+	<a href="https://liferay.design/lexicon/core-components/icons/">
+		Icons Pattern
+	</a>
+	.
 </div>
 
 ## Usage
@@ -39,9 +55,9 @@ We use SVG elements that link to an SVG sprite, like so:
 
 Note that the ID after the # symbol is the ID of the icon to use, so if you wanted to use plus icon, you would do change the href to `path/to/icons.svg#plus`.
 
-<ul class="lexicon-icon-list list-unstyled">
-	[foreach Icons]
-</ul>
+## Icon List
+
+<IconSearch source={icons} />
 
 ## Language Flags
 
@@ -53,9 +69,12 @@ Or, to paraphrase Winston Churchill, "Using flags is the worst system for indica
 
 To use the flags below, follow the same process as you would for a standard icon, and use the locale and country code indicated in parenthesis for the icon's name (e.g. to use the Japanese icon, you would use `ja-jp`)
 
-<ul class="lexicon-icon-list list-unstyled">
-	[foreach Flags]
-</ul>
+<IconSearch
+	label="Search Flags"
+	placeholder="Search Flags..."
+	source={flags}
+	type="flags"
+/>
 
 ### SVG Icons as Background Images
 
@@ -75,7 +94,8 @@ We have created a Sass function to turn Lexicon SVG icon's into data uri schemes
 You can turn your custom SVG into a data uri using the Sass function `clay-svg-url($svg)`, where `$svg` is the code for your inline SVG. The function returns `url(your_svg_as_data_uri)`.
 
 <div class="alert alert-info">
-	When using `clay-svg-url()`, wrap your inline SVG code with single quotes if your SVG's attributes are delimited with double quotes.
+	When using `clay-svg-url()`, wrap your inline SVG code with single quotes if
+	your SVG's attributes are delimited with double quotes.
 </div>
 
 <span class="clay-site-linux-tux clay-site-svg-bg">Linux Tux</span>
@@ -83,9 +103,11 @@ You can turn your custom SVG into a data uri using the Sass function `clay-svg-u
 ## Why do we use SVG
 
 <p id="lexicon-icon-explanation">
-  Font icons, while fairly simple, suffer drawbacks, such as sub-pixel aliasing. This results in a lower quality than we would like.
-	SVG gives us and you a greater amount of freedom in styling the icons, as well as a higher level of fidelity and clarity in the icons.
-	Also, SVG supports multi-color icons, as shown below:
+	Font icons, while fairly simple, suffer drawbacks, such as sub-pixel
+	aliasing. This results in a lower quality than we would like. SVG gives us
+	and you a greater amount of freedom in styling the icons, as well as a
+	higher level of fidelity and clarity in the icons. Also, SVG supports
+	multi-color icons, as shown below:
 </p>
 
 <div class="lexicon-icon-examples">
--- a/clayui.com/content/docs/components/css-icons.mdx
+++ b/clayui.com/content/docs/components/css-icons.mdx
@@ -69,12 +69,8 @@ Or, to paraphrase Winston Churchill, "Using flags is the worst system for indica
 
 To use the flags below, follow the same process as you would for a standard icon, and use the locale and country code indicated in parenthesis for the icon's name (e.g. to use the Japanese icon, you would use `ja-jp`)
 
-export function flagLabelFormatter(icon) {
-	return icon.aliases.join(' - ');
-}
-
 <IconSearch
-	iconLabelFormatter={flagLabelFormatter}
+	iconLabelFormatter={icon => icon.aliases.join(' - ')}
 	label="Search Flags"
 	placeholder="Search Flags..."
 	source={flags}
--- a/clayui.com/content/docs/components/css-icons.mdx
+++ b/clayui.com/content/docs/components/css-icons.mdx
@@ -69,11 +69,15 @@ Or, to paraphrase Winston Churchill, "Using flags is the worst system for indica
 
 To use the flags below, follow the same process as you would for a standard icon, and use the locale and country code indicated in parenthesis for the icon's name (e.g. to use the Japanese icon, you would use `ja-jp`)
 
+export function flagLabelFormatter(icon) {
+	return icon.aliases.join(' - ');
+}
+
 <IconSearch
+	iconLabelFormatter={flagLabelFormatter}
 	label="Search Flags"
 	placeholder="Search Flags..."
 	source={flags}
-	type="flags"
 />
 
 ### SVG Icons as Background Images
--- a/clayui.com/content/docs/components/icon.mdx
+++ b/clayui.com/content/docs/components/icon.mdx
@@ -6,6 +6,8 @@ sibling: 'docs/components/css-icons.html'
 ---
 
 import {Icon, IconWithContext} from '../../../src/components/clay/Icon';
+import IconSearch from '../../../src/components/IconSearch';
+import icons from '../../../static/js/icons.json';
 
 <div class="nav-toc-absolute">
 <div class="nav-toc">
@@ -16,6 +18,8 @@ import {Icon, IconWithContext} from '../../../src/components/clay/Icon';
 
 </div>
 </div>
+
+<IconSearch source={icons} />
 
 <div class="clay-site-alert alert alert-warning">
 	Take a look at our{' '}
--- a/clayui.com/src/components/IconSearch/index.js
+++ b/clayui.com/src/components/IconSearch/index.js
@@ -1,0 +1,84 @@
+/**
+ * © 2020 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayForm, {ClayInput} from '@clayui/form';
+import ClayIcon from '@clayui/icon';
+import React, {useMemo, useState} from 'react';
+
+const spritemap = '/images/icons/icons.svg';
+
+/**
+ * Component that allows searching through icons
+ * @param {string} label - Label of the search input field.
+ * @param {string} placeholder - Placeholder of the search input field.
+ * @param {Array[Object]} source - Path to the source JSON. Needs to follow pattern of [{name: String, aliases: Array<String>}]
+ * @param {string} type - Type of icons being provided, supports "icons" and "flags"
+ */
+const IconSearch = ({
+	label = 'Search Icons',
+	placeholder = 'Search Icons...',
+	source,
+	type = 'icons',
+}) => {
+	const [searchQuery, setSearchQuery] = useState('');
+	let list = [];
+
+	const filteredIcons = useMemo(() => {
+		const query = searchQuery.toLowerCase();
+
+		return source.filter(
+			({aliases, name}) =>
+				name.toLowerCase().includes(query) ||
+				aliases.some(alias => alias.toLowerCase().includes(query))
+		);
+	}, [searchQuery]);
+
+	if (filteredIcons.length) {
+		list = filteredIcons;
+	} else {
+		list = searchQuery ? [] : source;
+	}
+
+	return (
+		<>
+			<ClayForm.Group>
+				<label style={{width: '100%'}}>
+					{label}
+
+					<ClayInput
+						onChange={event => setSearchQuery(event.target.value)}
+						placeholder={placeholder}
+						type="text"
+						value={searchQuery}
+					/>
+				</label>
+
+				<ul className="lexicon-icon-list list-unstyled">
+					{list.map(icon => (
+						<li key={icon.name}>
+							<ClayIcon
+								spritemap={spritemap}
+								symbol={icon.name}
+							/>
+
+							<span>
+								{type === 'flags'
+									? icon.aliases.join(' - ')
+									: icon.name}
+							</span>
+						</li>
+					))}
+				</ul>
+
+				{!list.length && (
+					<span>{`No results found for ${searchQuery}`}</span>
+				)}
+			</ClayForm.Group>
+		</>
+	);
+};
+
+export default IconSearch;
--- a/clayui.com/src/components/IconSearch/index.js
+++ b/clayui.com/src/components/IconSearch/index.js
@@ -24,7 +24,6 @@ const IconSearch = ({
 	iconLabelFormatter = icon => icon.name,
 }) => {
 	const [searchQuery, setSearchQuery] = useState('');
-	let list = [];
 
 	const filteredIcons = useMemo(() => {
 		const query = searchQuery.toLowerCase();
@@ -36,11 +35,7 @@ const IconSearch = ({
 		);
 	}, [searchQuery, source]);
 
-	if (filteredIcons.length) {
-		list = filteredIcons;
-	} else {
-		list = searchQuery ? [] : source;
-	}
+	const list = searchQuery ? filteredIcons : source;
 
 	return (
 		<>
--- a/clayui.com/src/components/IconSearch/index.js
+++ b/clayui.com/src/components/IconSearch/index.js
@@ -21,7 +21,7 @@ const IconSearch = ({
 	label = 'Search Icons',
 	placeholder = 'Search Icons...',
 	source,
-	type = 'icons',
+	iconLabelFormatter = icon => icon.name,
 }) => {
 	const [searchQuery, setSearchQuery] = useState('');
 	let list = [];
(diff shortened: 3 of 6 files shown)


clayui.com/src/components/IconSearch/IconSearch.js
```diff
@@ -0,0 +1,76 @@
...
+
+import icons from './icons.json';
+
+const spritemap = '/images/icons/icons.svg';
+
+export default () => {
+	const [searchQuery, setSearchQuery] = useState('');
+	const [filteredIconNames, setFilteredIconNames] = useState([]);
+
+	const handleIconSearch = event => {
```
bryceosterhaus: No need to declare this separately from the `onChange` prop below. We can inline this function directly to the jsx.

clayui.com/src/components/IconSearch/IconSearch.js
```diff
@@ -0,0 +1,76 @@
...
+		setSearchQuery(event.target.value);
+	};
+
+	useEffect(() => {
+		const filteredIcons = icons.filter(icon => {
+			if (icon.aliases.includes(searchQuery)) {
```
bryceosterhaus: Instead of using an `if` block here, we can simplify this with an implicit return.

```js
const filteredIcons = icons.filter(
     icon => icon.aliases.includes(searchQuery)
);
```

clayui.com/src/components/IconSearch/IconSearch.js
```diff
@@ -0,0 +1,76 @@
...
+				return icon.name;
+			}
+		});
+
+		setFilteredIconNames(
+			filteredIcons.map(icon => {
```
bryceosterhaus: Lets use an implicit return here `filteredIcons.map(icon => icon.name)`

clayui.com/src/components/IconSearch/IconSearch.js
```diff
@@ -0,0 +1,76 @@
...
+					id="iconSearch"
+					onChange={handleIconSearch}
+					placeholder="Search icons"
+					type="text"
+					value={searchQuery}
+				/>
+			</ClayForm.Group>
+
+			<ul className="lexicon-icon-list list-unstyled">
+				{filteredIconNames.length
```
bryceosterhaus: We can simplify this block a bit by doing something like this...

```js
// This would live above the jsx
const list = filteredIconNames.length ? filteredIconNames : icons;

// This would live where the list is rendered
{list.map(icon => (
	<li key={icon.name}>
		<ClayIcon
			spritemap={spritemap}
			symbol={icon.name}
		/>

		<span>{icon.name}</span>
	</li>
))}
```

Also note that we should avoid using the index for the `key`. Sometimes its fine to use, but if we can use a unique identifier instead, it will re-render more accurately.

clayui.com/src/components/IconSearch/IconSearch.js
bryceosterhaus: Also, we should additionally filter based off of the icon name itself. For example, if someone wants to search for "add" the `add-cell` icon won't actually show up since we are only searching based off its aliases.

Screen Shot 2020-02-03 at 10 41 28 AM

clayui.com/src/components/IconSearch/IconSearch.js
```diff
@@ -0,0 +1,76 @@
...
+							<li key={i}>
+								<ClayIcon
+									spritemap={spritemap}
+									symbol={iconName}
+								/>
+
+								<span>{iconName}</span>
+							</li>
+					  ))
+					: icons.map((icon, i) => (
```
bryceosterhaus: Instead of rendering all of the icons if no results are found, we should show an empty state message like `No results found for "test"`

bryceosterhaus: For example, if you search for "zzzz" it shows all icons which is a little confusing for the end user.

Screen Shot 2020-02-03 at 10 41 17 AM

clayui.com/src/components/IconSearch/IconSearch.js
kresimir-coko: Ah yea, good catch. I didn't go deep into cleanup since I just wanted to get the filtering to work and for it to actually render somewhere. Thanks for the in-depth comments.

kresimir-coko:
Hey @bryceosterhaus, 
I've pushed 2 commits, 1 is related to the IconSearch.js file improving the interactions and basically resolving all the suggested changes.

The other is my progress made so far with the switch to the MDX file and playing around with GraphQL. 
I'm satisfied with this rough implementation but I did notice the fact that a new page appeared in the navigation called `Icons` which is probably because I renamed the file from html to mdx so whichever part of the code handles that will need to be updated too, I'll fix that if this approach proves to be worth pursuing.

Let me know if this approach is good, GraphQL + Gatsby surprised me with how powerful they are.

clayui.com/content/docs/components/icon.mdx
```diff
@@ -2,10 +2,11 @@
 title: 'Icon'
 description: 'Icons are a visual representation of an idea and/or action.'
 packageNpm: '@clayui/icon'
-sibling: 'docs/components/css-icons.html'
+sibling: 'docs/components/css-icons.mdx'
```
bryceosterhaus: I don't think we would want to change this to `.mdx` since we still want to point to the compiled page. Meaning, the page should still be `.html`

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,9 @@
+/**
```
bryceosterhaus: We don't need this file really, we could just move everything from `IconSearch` over to this file.

clayui.com/src/templates/docs.js
```diff
@@ -362,5 +400,22 @@ export const pageQuery = graphql`
 				}
 			}
 		}
+		icons: mdx(fields: {slug: {eq: "docs/components/css-icons.html"}}) {
```
bryceosterhaus: Rather than making a _specific_ case for the css-icons file, could we make it work for all cases that use `.mdx` as a sibling?

Meaning it would be something like this, where it would compile any mdx file that is declared as a sibling.
```js
mdxTab: mdx(fields: {slug: {eq: $sibling}}) {
	code {
		body
	}
}
```

Additionally, we would probably want to update the existing `tabs` one by doing something like
```js
mdTab: markdownRemark(fields: {slug: {eq: $sibling}}) {
	html
}
```

Notice that we don't really need to find _all_ the tabs since we only ever declare a single css tab.

bryceosterhaus: And after this sort of change, we will have to do something like `const tab = mdTab || mdxTab;` in the logic above to get the actual tab to render. One of those tabs will render via its `html` and the other will render via its `code.body`

clayui.com/src/components/IconSearch/index.js
kresimir-coko: Alright, I was following the pattern in it's siblings, `LayoutNav`, `Sidebar`, etc.

clayui.com/content/docs/components/css-icons.mdx
```diff
@@ -15,6 +16,8 @@ description: 'Icons are a visual representation of an idea and/or action.'
 </div>
 </div>
 
+<IconSearch />
+
```
matuzalemsteles: hey @kresimir-coko and @bryceosterhaus I would add this component below the warnings, it is important that they are visible as soon as the user enters the page.

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayForm, {ClayInput} from '@clayui/form';
+import ClayIcon from '@clayui/icon';
+import React, {useEffect, useState} from 'react';
+
+const spritemap = '/images/icons/icons.svg';
+
+const IconSearch = source => {
```
bryceosterhaus: Since this is a react component we typically either call this arg `props` or we will immediately destructure it `({source}) =>`

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+	const [searchQuery, setSearchQuery] = useState('');
+	const [filteredIconNames, setFilteredIconNames] = useState([]);
+	const list = filteredIconNames.length ? filteredIconNames : [];
+
+	source = [...source.source];
```
bryceosterhaus: Cloning the array here seems pretty odd... When we clone items for the purpose of mutation, I think we should avoid doing is on a large scale here(being in the main body of the component function), and instead we should only mutate for small intermediate steps like the example below.

```js
const pushItem = (items, newItem) => {
     const clonedItems = {...items};

     clonedItems.total++
     clonedItems.items = [...clonedItems.items, newItem];

     return clonedItems;
};
```

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+	useEffect(() => {
+		const filteredIcons = source.filter(icon => {
+			const iconPrimary = icon[Object.keys(icon)[0]].toUpperCase();
```
bryceosterhaus: Instead of using `Object.keys(icon)` and picking indexes, if we have a strict type for the icons object, we can just get the items by its key. This would require us to define the flags and icons json to be the same schema, which should be fine since we have control of that. This would allow us to clean up the filtering logic. For example...

```ts
// Imagine this is the type of the icons provided
type iconsSourceJSON = Array<{
	aliases: Array<string>;
	name: string;
}>;

// Logic for filtering
// `source` is the iconsSourceJSON type
const filteredIcons = source.filter(icon => {
	const allItems = [icon.name, ...icon.aliases].map(item =>
		item.toUpperCase()
	);

	return allItems.includes(searchQuery.toUpperCase());
});
```

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+		setFilteredIconNames(filteredIcons);
+	}, [searchQuery]);
+
+	return (
+		<>
+			<ClayForm.Group>
+				<label htmlFor="iconSearch">{'Search Icons'}</label>
+
+				<ClayInput
+					id="iconSearch"
```
bryceosterhaus: We shouldn't hard code this `id` here since we would break html semantics if we had more than one of these components on the page. Either we need to pass a unique id, or you may also be able to wrap the input in `<label>`.

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+					onChange={event => setSearchQuery(event.target.value)}
+					placeholder="Search icons"
```
bryceosterhaus: We should pass this from the parent. This way we could pass `Search Flags` for when we use flags.

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+					type="text"
+					value={searchQuery}
+				/>
+			</ClayForm.Group>
+
+			<ul className="lexicon-icon-list list-unstyled">
+				{list.map(icon => {
+					const iconPrimary = icon[Object.keys(icon)[0]];
```
bryceosterhaus: See previous comment about cleaning this up.

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,81 @@
...
+			);
+		});
+
+		setFilteredIconNames(filteredIcons);
+	}, [searchQuery]);
+
+	return (
+		<>
+			<ClayForm.Group>
+				<label htmlFor="iconSearch">{'Search Icons'}</label>
```
bryceosterhaus: We should also add this label content to come from the parents prop, that way it is can be customized per use, such as flags. You can have `Search Icons` as the default value though in case a label is not passed in.

clayui.com/src/templates/docs.js
```diff
@@ -242,12 +249,51 @@ export default props => {
 												{hasTabs && (
```
bryceosterhaus: Rather than using the `has*` variables, we could have something like this for the render logic

```jsx
{tab && (
	<>
		{tab.html && (
			<div
				dangerouslySetInnerHTML={{
					__html:
						tab.html,
				}}
			/>
		)}

		{tab.code && (
			// MDX render logic
		)}
	</>
)}
```

clayui.com/src/templates/docs.js
```diff
@@ -22,11 +22,21 @@ export default props => {
...
+		mdTab,
+		mdx,
+		mdxTab,
+	} = data;
 	const {code, excerpt, fields, frontmatter, html, timeToRead} =
 		mdx || markdownRemark;
 
-	const hasTabs = tabs.edges.length > 0;
+	const hasTabs = mdTab ? mdTab.html.length > 0 : false;
+	const hasMdxTab = mdxTab ? mdxTab.code.body.length > 0 : false;
```
bryceosterhaus: I don't think we need these `has*` variables here, we can simply just do `const tab = mdTab || mdxTab;` since the graphql query will return `null` if the tab doesn't exist. And then when we have any conditional for if a tab renders we can just do `{tab && ...}` and `{tab.html && ...}` or `{tab.code && ...}`

clayui.com/content/docs/components/css-icons.mdx
```diff
@@ -39,9 +54,7 @@ We use SVG elements that link to an SVG sprite, like so:
 
 Note that the ID after the # symbol is the ID of the icon to use, so if you wanted to use plus icon, you would do change the href to `path/to/icons.svg#plus`.
 
-<ul class="lexicon-icon-list list-unstyled">
-	[foreach Icons]
-</ul>
+<IconSearch source={icons} />
```
bryceosterhaus: Can we add an additional `## header` here and then add that link to the table of contents? I think it will help the flow and also will make this table linkable.

clayui.com/src/components/IconSearch/index.js
kresimir-coko: This and the destructuring from above is related. Destructuring when declaring the param crossed my mind but I disregarded it before even trying, idk why, must have spaced out.

The cloning here was just to get the `source` to be an array instead of an object, which I'll do via destructuring

clayui.com/src/components/IconSearch/index.js
kresimir-coko: Yeah that was the original implementation but I went with a more vague approach which made the code uglier. 
I prefer having them follow the same schema, not sure how to enforce or inform future devs of the component that their source JSON needs to have a specific schema.

clayui.com/src/templates/docs.js
kresimir-coko: I was getting some issues when I tried doing this but it might have been due to me debugging GraphQL at the time. But yea, this is a much cleaner approach.

clayui.com/src/components/IconSearch/index.js
bryceosterhaus: You could add a comment jsdoc block at the top of the component and define props and type accepted.

kresimir-coko:
@bryceosterhaus I've force pushed more changes, most notably I've changed the filtering logic from `useEffect` to `useMemo` per advice by Greg

clayui.com/src/components/IconSearch/index.js
```diff
@@ -6,74 +6,77 @@
...
-			);
-		});
+	const filteredIcons = useMemo(() => {
+		const query = searchQuery.toLowerCase();
 
-		setFilteredIconNames(filteredIcons);
+		return source.filter(
+			({aliases, name}) =>
+				aliases.some(alias => alias.toLowerCase().includes(query)) ||
+				name.toLowerCase().includes(query)
```
wincent: One little nit here @kresimir-coko: when you have a chain of conditional expressions like this, you generally want to put the cheapest one first, to maximize the chances of short-circuiting and bailing out of the computation early. In this example, the `name` check is cheaper (two function calls) than the `aliases` one (one function call plus two more for each alias in the list), which is why I put it first in my suggestion to you.

This kind of thing is microoptimization .... until it isn't. 😂 So even though it doesn't matter in this case, it's a good habit to get into.

kresimir-coko: Ah, I was going for alphabetical sorting, I'll force-push with them organized in the manner you proposed! Thanks for the heads up

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,84 @@
...
+				<ul className="lexicon-icon-list list-unstyled">
+					{list.map(icon => (
+						<li key={icon.name}>
+							<ClayIcon
+								spritemap={spritemap}
+								symbol={icon.name}
+							/>
+
+							<span>
+								{type === 'flags'
```
bryceosterhaus: Rather than creating a one off API for a specific case like flags, let's try to make the API as generic and configurable as possible.

For example, instead of a `type` prop, we could do something like `iconLabelFormatter` which is a function that renders the label for the icon. If we did this, we can pass it in as a prop for flags and then internally have a default function for it.

```js
const IconSearch = ({
	label = 'Search Icons',
	iconLabelFormatter = (item) => item.name,
	placeholder = 'Search Icons...',
	source,
}) => {
//...
<li key={icon.name}>
	<ClayIcon
		spritemap={spritemap}
		symbol={icon.name}
	/>

	<span>
		{iconLabelFormatter(icon)}
	</span>
</li>
//...
}
```



_Side note / soap box_
**While this may not actually be a huge issue for us on the clayui.com codebase or impact us greatly, the goal would be to get in the habit of designing more generic APIs.** Otherwise, we may start designing APIs in our components that require us to add additional "cases" and logic for different circumstances. For example, suppose we have the following data and an additional case where we need to render the Spanish word for each icon.

```jsx
// aliases: [english, spanish, french]
[
	{
		name: 'icon-apple',
		aliases: ['apple', 'manzana', 'pomme'],
	},
	{
		name: 'icon-banana',
		aliases: ['banana', 'plátano', 'banane'],
	},
];

// With the existing API, we would have to add a "spanish" type and add the logic to
// handle that case.
<IconSearch type="spanish" source={icons} />

// If we go with a more generic API, we won't have to add any internal logic or types, and
// it can be used as is.
<IconSearch iconLabelFormatter={(item) => item.aliased[1]} source={icons} />
```

By adopting a more generic API, we are making it easier for ourselves to maintain over time and avoiding having to create a new API for every case.

This whole idea sort of relates to [big-o notation](https://rob-bell.net/2009/06/a-beginners-guide-to-big-o-notation/) and complexity. That link may help give more of a theoretical idea of complexity issues, but if it creates confusion, feel free to ignore it. 
/end

I hope that example is helpful for clarity on API design. And as always, if you have any questions just let me know or I would be happy to get on a call and talk through it more.

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,84 @@
...
+
+	const filteredIcons = useMemo(() => {
+		const query = searchQuery.toLowerCase();
+
+		return source.filter(
+			({aliases, name}) =>
+				name.toLowerCase().includes(query) ||
+				aliases.some(alias => alias.toLowerCase().includes(query))
+		);
+	}, [searchQuery]);
```
bryceosterhaus: I think you would need to additionally add `source` to this array, since we would want this function to update if a new set of items was passed into the component.

clayui.com/src/components/IconSearch/index.js
```diff
@@ -0,0 +1,84 @@
...
+
+	if (filteredIcons.length) {
```
bryceosterhaus: I think you could simplify this if/else block slightly since the `.filter()` should always return an array.

```js
list = filteredIcons.length ? filteredIcons : source
```

clayui.com/static/js/flags.json
```diff
@@ -0,0 +1,394 @@
+[
+	{
+		"name": "ar-sa",
+		"aliases": [
+            "Arabic",
```
bryceosterhaus: whitespace seems to be off here in this file

wincent:
> You for sure need the colon and you may also need to lowercase the first word of the subject... I am less sure about the casing though.

The conventional commits spec doesn't have any special requirements for the case, although their examples happen to use lowercase, and if you look across our repos (including this one) we seem to all be using lowercase.

clayui.com/src/components/IconSearch/index.js
kresimir-coko: I don't think we can, since we want to have an empty array in the case where the `searchQuery` doesn't match anything. The logic being if `searchQuery` exists but doesn't match anything return an empty array instead of `source`, which is what your proposal would do. Resulting in no icons matching the filter but all icons being shown.

If `filteredIcons.length` is falsy, that means no icons match the filter, in that case we want to show the `no results` message, not the `source`.

clayui.com/src/components/IconSearch/index.js
kresimir-coko: You're right. I think that if we spent more time defining the requirements of this component before going into writing the implementation we would have avoided this problem.

This did cross my mind, and it did feel bad to hardcode the type, but I didn't go far in my thinking to get the idea of giving the user of the component the option to pass a formatting function. I'll have to think about that stuff in the future!

kresimir-coko:
Should the component name maybe change to `IconFilter` perhaps? Maybe even something more generic than `Icon`? It's not a simple search, it provides both the search functionality and the displaying of icons which it searches through.

clayui.com/content/docs/components/css-icons.mdx
```diff
@@ -53,9 +69,16 @@ Or, to paraphrase Winston Churchill, "Using flags is the worst system for indica
...
 
-<ul class="lexicon-icon-list list-unstyled">
-	[foreach Flags]
-</ul>
+export function flagLabelFormatter(icon) {
+	return icon.aliases.join(' - ');
+}
+
+<IconSearch
+	iconLabelFormatter={flagLabelFormatter}
```
bryceosterhaus: We can just pass this in directly.

```suggestion
	iconLabelFormatter={icon => icon.aliases.join(' - ')}
```

clayui.com/src/components/IconSearch/index.js
bryceosterhaus: Oh I see what you mean, then for that case we could do `list = searchQuery ? filteredIcons : source`

This should account for

`searchQuery` exists and matches no icons  === `[]` or `filteredIcons`(should be empty array if nothing matches)
`searchQuery` exists and matches icons  === `filteredIcons`
`searchQuery` does not exist  === source

unless I am going crazy and just can't think straight...

bryceosterhaus:
Thanks @kresimir-coko!

PR #2902, opened 2020-02-04
feat(clay-css): Adds pattern `.c-inner` for highjacking focus state on click

pat270:
This is a big change, but doesn't change the existing behavior even if this setting is enabled. This is disabled by default and can be enabled by setting `$enable-c-inner: true;` in your Clay CSS variable overwrites.

Once enabled the markup follows the pattern:
```
<button class="btn btn-primary" type="button">
    <span class="c-inner" tabindex="-1">Primary</span>
</button>
```

Let me know if this is a feature we want to enable by default.

### DIFF
--- a/packages/clay-css/src/content/utilities_c_inner.html
+++ b/packages/clay-css/src/content/utilities_c_inner.html
@@ -1,0 +1,5109 @@
+---
+title: Utilities (C Inner)
+section: Components
+---
+
+<div class="alert alert-danger">
+	Before using this pattern, make sure to enable it by setting <code>$enable-c-inner: true;</code>.
+</div>
+
+<blockquote class="blockquote">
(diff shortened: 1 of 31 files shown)


marcoscv-work:

marcoscv-work:
Hey @pat270 !
It increases by about 30kb. Could we even use a new type of button/tab or something to make it lighter?
Thoughts?

pat270:
@marcoscv-work 
I considered moving Button/Link sizing styles directly to `c-inner`, but decided not to because it still required resetting styles from Bootstrap and both types of focus patterns couldn't co-exist. The quickest way to reduce the size is to allow disabling by component so devs can disable what they don't want or vice versa.

A new type of Button/Tab/Link doesn't cover existing Clay CSS Components, which defeats the purpose of this issue imo. In the next major version we can create components with this pattern in mind which should reduce the additional CSS required.

The increase in size is ~30kb unminified/zipped. Minification + gzipping should increase the footprint ~2kb. If there is a cleaner way to do this I'm definitely open to it.

marcoscv-work:

marcoscv-work:
Pull request submitted to https://github.com/liferay/clay/pull/2921. [See changes here.](https://github.com/liferay/clay/compare/2609129a9~1...2609129a9) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #2908, opened 2020-02-06
feat(@clayui/form): create new toggle component

### LINKED ISSUE #2893: [New] ToggleSwitch Component
This is a formal request to **implement Lexicon's [Toggle Input Variation](https://liferay.design/lexicon/core-components/forms/radio-check-toggle/#toggle)** as a Clay Component.

> A toggle is an interface element that, by its activation or deactivation, provokes an immediate action on the screen.

switchlabelon

#### References
- [Lexicon Definition](https://liferay.design/lexicon/core-components/forms/radio-check-toggle/#toggle)
- [app-builder-web::ToggleSwitch](https://github.com/liferay/liferay-portal/blob/master/modules/apps/app-builder/app-builder-web/src/main/resources/META-INF/resources/js/components/toggle-switch/ToggleSwitch.es.js)
- [segments-web::ClayToggle](https://github.com/liferay/liferay-portal/blob/master/modules/apps/segments/segments-web/src/main/resources/META-INF/resources/js/components/shared/ClayToggle.es.js)
- Inner component at [portal-search-tuning-web::PageToolbar](https://github.com/liferay/liferay-portal/blob/master/modules/dxp/apps/portal-search-tuning/portal-search-tuning-rankings-web/src/main/resources/META-INF/resources/js/components/PageToolbar.es.js#L49-L70) 

/cc @bryceosterhaus, @kresimir-coko, feel free to ping @drakonux for any clarification

bryceosterhaus:
fixes #2893

Note, I still need to add and updates docs, just wanted to get approval on this API pattern before going with docs.

### DIFF
--- a/packages/clay-form/src/RadioGroup.tsx
+++ b/packages/clay-form/src/RadioGroup.tsx
@@ -8,9 +8,15 @@ import classNames from 'classnames';
 import React from 'react';
 
 import {IRadioProps} from './Radio';
+import {IToggleProps} from './Toggle';
 
 interface IGroupProps extends React.HTMLAttributes<HTMLDivElement> {
-	children: Array<React.ReactElement<IRadioProps>>;
+	/**
+	 * Takes either Radio or Toggle as a child.
+	 */
+	children: Array<
+		React.ReactElement<IRadioProps> | React.ReactElement<IToggleProps>
+	>;
 
 	/**
 	 * Flag to indicate if radio elements should display inline.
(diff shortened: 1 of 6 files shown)


bryceosterhaus:
@matuzalemsteles I was following the "kitchen sink" pattern seen here, https://pat270.github.io/lexicon/v2.16.2/content/toggle-switch/#toggle-switch-text

jbalsas:
🔧 let's try to avoid kitchensinks! 😂 

It should be enough if https://liferay.design/lexicon/core-components/forms/radio-check-toggle/ is covered by the component. We don't need to implement crazy stuff we came up with one day and no-one ever used :(

Basically, since Figma components won't be there, the chances of developers needing sth outside of there are minimal... and if they do need it, they might need to figure how to build it themselves.

I would try to avoid adding unnecessary features as much as possible (I know you want to avoid it too 😂). In any case, I haven't looked at the component in detail, so I'll leave that up to you :)

bryceosterhaus:
@jbalsas good call. I looked at the clay-css implementation first and didnt think of the lexicon site. I'll simplify it a bit.

bryceosterhaus:
@matuzalemsteles @jbalsas just simplified the API. let me know what you think

matuzalemsteles:
@bryceosterhaus looks clean now, LGTM!

PR #2911, opened 2020-02-07
 chore: update storybook dependencies and config

bryceosterhaus:
I've been wanting to do this for awhile now since `awesome-typescript-loader` is no longer worked on and also because it takes so long for storybook to start up. Here are a list of tradeoffs with this change...

## After Changes

**Startup:** ~120 seconds

**Terminal Display:** 
Screen Shot 2020-02-07 at 1 20 22 PM

**UI Display:** 
Screen Shot 2020-02-07 at 1 20 30 PM

## After Changes

**Startup:** ~16 seconds

**Terminal Display:** 
Screen Shot 2020-02-07 at 1 13 55 PM

**UI Display:** 
Screen Shot 2020-02-07 at 1 14 52 PM

Another big tradeoff is that before these changes, if you started up storybook if a type error, you would have to restart the whole storybook process, now that wouldn't be the case and it'll save you time. There is a tradeoff in that we dont get errors in the UI, but sometimes that can also be more annoying than it is helpful.

Any thoughts or concerns?


diegonvs:
> If you started up storybook if a type error, you would have to restart the whole storybook process

When catching a type error on the hot reload, I should restart whole storybook build again?

bryceosterhaus:
@diegonvs Currently in master, if you start up storybook with a ts error, it requires you to restart the whole thing. If it starts up cleanly the first time and then you get a ts error, the hot reload should be fine.

jbalsas:
The perf improvement sounds awesome!!

Not sure how @wincent will feel about these changes to the `yarn.lock` though 😅 

Screen Shot 2020-02-09 at 09 10 25

bryceosterhaus:
@jbalsas yeah I also noticed that and it gave me some worry. Ideally I think we need to update our current dependencies first and then work on reducing our dependency footprint as a whole.

wincent:
> Not sure how @wincent will feel about these changes to the `yarn.lock` though 😅

You are sure, though, aren't you? 😂

jbalsas:
> You are sure, though, aren't you? 😂

I might be... 😂

bryceosterhaus:
@wincent @jbalsas how do you guys feel about updating our current dependencies first and then working on reducing total footprint?

jbalsas:
I'm fine with it, @bryceosterhaus!

wincent:
> how do you guys feel about updating our current dependencies first and then working on reducing total footprint?

Makes sense: personally, I'd only go so far as updating the "easy" ones (that is, until I run into headaches, if any, such as things with painful breaking changes), and then I'd switch gears and see if anything could be stripped.

bryceosterhaus:
If we find any issues with this, we can revert. Going to merge

Issue #2916, opened 2020-02-11
Avoidable DOM thrash in ClayCard


wincent:
Not sure if this is a React bug or we were just doing it wrong, but using [the `<Content>` component inside the `ClayCard` implementation](https://github.com/liferay/clay/blob/1fbdfa21e64e623f9470fa50e5d242c0743a0809/packages/clay-card/src/Card.tsx#L56-L60) was causing React to remove one of the descendent elements from the DOM and re-insert it, even though there was no actual diff. When the descendent element is an `<img>`, this is bad and can cause a visual flicker (maybe something we haven't noticed before because we tend not to re-render cards much; but @boton noticed it now because he is rendering a progress bar inside the card that updates frequently).

To see this in action, see this demo:

https://codesandbox.io/s/blissful-shaw-xrekn

There is a fork of `ClayCard` in there called `ClayCardFork` so that you can edit it. When you turn off the browser cache, and use the `<Content>` component, you'll see a new image request going out over the network every second, even though nothing in the DOM is actually changing.

If we avoid the `<Content>` component and just inline the `Context.Provider` directly, no DOM mutations occur and no bogus network requests happen. Fortunately, the solution actually looks a bit less repetitive than what we had before, so there is no downside to this workaround.

PR #2917, opened 2020-02-11
fix(card): prevent unwanted DOM thrash in ClayCard (#2916)

wincent:
Not sure if this is a React bug or we were just doing it wrong, but using [the `<Content>` component inside the `ClayCard` implementation](https://github.com/liferay/clay/blob/1fbdfa21e64e623f9470fa50e5d242c0743a0809/packages/clay-card/src/Card.tsx#L56-L60) was causing React to remove one of the descendent elements from the DOM and re-insert it, even though there was no actual diff. When the descendent element is an ``, this is bad and can cause a visual flicker (maybe something we haven't noticed before because we tend not to re-render cards much; but @boton noticed it now because he is rendering a progress bar inside the card that updates frequently).

To see this in action, see this demo:

https://codesandbox.io/s/blissful-shaw-xrekn

There is a fork of `ClayCard` in there called `ClayCardFork` so that you can edit it. When you turn off the browser cache, and use the `<Content>` component, you'll see a new image request going out over the network every second, even though nothing in the DOM is actually changing.

If we avoid the `<Content>` component and just inline the `Context.Provider` directly, no DOM mutations occur and no bogus network requests happen. Fortunately, the solution actually looks a bit less repetitive than what we had before, so there is no downside to this workaround.

Closes: https://github.com/liferay/clay/issues/2916

### DIFF
--- a/packages/clay-card/src/Card.tsx
+++ b/packages/clay-card/src/Card.tsx
@@ -53,12 +53,6 @@ const ClayCard: React.FunctionComponent<IProps> & {
 	selectable = false,
 	...otherProps
 }: IProps) => {
-	const Content: React.FunctionComponent<IProps> = ({children}) => (
-		<Context.Provider value={{horizontal, interactive}}>
-			{children}
-		</Context.Provider>
-	);
-
 	const isCardType = {
 		file: displayType === 'file',
 		image: displayType === 'image',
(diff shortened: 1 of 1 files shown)


wincent:
To see what actually changed here, suggest turning off whitespace changes:

fix_card___prevent_unwanted_DOM_thrash_in_ClayCard___2916__by_wincent_·_Pull_Request__2917_·_liferay_clay

bryceosterhaus:
Fix looks good to me. Really odd behavior though, there may be some nuance to `Provider` that I am unaware of. I'll look into this.

As for the whitespace, how come it looks so odd without toggling that setting? I've never had to use that before. I'm assuming its still fine to merge?

bryceosterhaus:
Looks like its just a bad practice to declare a new component with a Provider in the function body, especially a provider that is passing a new object every time. [reference](https://reactjs.org/docs/context.html#caveats)

wincent:
> As for the whitespace, how come it looks so odd without toggling that setting? I've never had to use that before. I'm assuming its still fine to merge?

Yes, it's just because the change involved modifying the indentation of a bunch of lines (because we moved the context provider up and wrapped it around everything), which makes it look like all the lines changed dramatically, when really they didn't — they just moved one tabstop to the right.

Re: the caveat you linked, that is actually a separate issue that I didn't want to address in this pull in order to keep the scope small. This component *still* has that problem (creating a new context value on every render, because we create a new `{horizontal, interactive}` object on every render), which means that the children will (or at least may) re-render more often than they have to. To fix that, we'd move the context value into state like the page you linked suggests.

In this PR, I wanted to address the very specific and weird edge case that even though the component "renders" the same markup, React unhelpfully thrashes the actual DOM (re-inserts the `img` tag) and causes a visual flicker. In other words, not just a potential performance problem but an actual bug.

bryceosterhaus:
Ah gotcha, that makes sense to me and this PR looks good to me then. I will go ahead and merge.

As far as the caveat, initially when creating the components with context I wanted to avoid over optimizing things initially and rather wait till they its actually needed. There is still likely a lot of performance optimizations across clay that we can tackle, but since nothing has reared its head too badly, I think its fine to sort of wait and see before going overkill on context, memos, and optimal re-renders.

PR #2920, opened 2020-02-12
fix: keep it phresh w/a dynamic copyright year

plhnk:
2019 is so two-thousand-and-late!

### DIFF
--- a/clayui.com/src/pages/index.js
+++ b/clayui.com/src/pages/index.js
@@ -403,7 +403,7 @@ export default () => {
 
 							<div className="col-lg text-center text-lg-right">
 								<div>
-									{'Copyright © ' + `${year} `}
+									{`Copyright © ${year} `}
 									<a
 										className="font-weight-bold"
 										href="https://www.liferay.com/"
--- a/clayui.com/src/pages/index.js
+++ b/clayui.com/src/pages/index.js
@@ -15,6 +15,8 @@ export default () => {
 	const description =
 		'This is Clay. A web implementation of the Lexicon Experience Language; built by Liferay.';
 
+	const CopyrightYear = new Date().getFullYear()
+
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
@plhnk looks good to me. Can you just update the commit message and verify ci? Commit message should be prefixed with `fix(clayui.com): ` but the rest of the message is fine

clayui.com/src/pages/index.js
```diff
@@ -401,7 +403,7 @@ export default () => {
 
 							<div className="col-lg text-center text-lg-right">
 								<div>
-									{'Copyright © 2019 '}
+									{'Copyright © ' + `${CopyrightYear} `}
```
bryceosterhaus: ci might be complaining about the not using string templating for the "copyright" portion.
```suggestion
									{`Copyright © ${CopyrightYear}`}
```

julien:
Just adding my 2 cents here, if it can help.

- In general for real constants we use uppercase:

```javascript
const COPYRIGHT_YEAR = new Date().getFullYear();
```

although in this case we could probably be fine with:

```javascript
const year = new Date().getFullYear();
```

- We also use semicolons

- But In this case I also wanted to share [this](https://grow.liferay.com/share/Internal+Liferay+Copyright+_AMPERSAND_+Licensing+FAQ) document which has general guidelines on how to use copyright in our code, more specifically [this](https://grow.liferay.com/share/Internal+Liferay+Copyright+_AMPERSAND_+Licensing+FAQ#q-not-update-the-year-in-the-copyright-header-are-you-mad-surely-this-is-asking-for-trouble) part. You'll notice that in liferay-portal for example all source files have a license header, but the year is not updated. 

Let me know what you think
Thanks

PR #2921, opened 2020-02-12
feat(clay-css): enable-c-inner: true by default

marcoscv-work:


pat270:
@marcoscv-work fixed conflicts sent new pull at #2922

PR #2924, opened 2020-02-12
feat(@clayui/drop-down): add more options for alignment

### LINKED ISSUE #2909: ClayDropdown positions

No =(

The position TopLeft TopRight is not supported, passing it through props works, but without offset as a side effect.

It will be great an API to define offset, and also support to this position if possible.

## Environment

| Tech  | Version |
| ClayDropdown  | v3.1.3  |
| React |   v16.8.6      |

bryceosterhaus:
fixes #2909

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -55,6 +55,11 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement | HTMLLIElement> {
 	onActiveChange: (val: boolean) => void;
 
 	/**
+	 * Function for setting the offset of the menu from the trigger.
+	 */
+	offsetFn?: React.ComponentProps<typeof Menu>['offsetFn'];
+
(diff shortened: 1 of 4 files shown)


bryceosterhaus:
@fcodias can you verify this is what your expectations are? Check out the storybook to see the example, https://deploy-preview-2924--storybook-clayui.netlify.com/

fcodias:
Thanks @bryceosterhaus , it looks good!! I think it's ok for our case. 
About the offset, do you think it will be possible to define it?

matuzalemsteles:
Looks good to me! Before going with that maybe we could implement an API for `offset` (as mentioned by @fcodias).

bryceosterhaus:
@fcodias @matuzalemsteles added a new commit with custom offset functionality. Let me know if that looks like what you'd expect

fcodias:
Looks nice @bryceosterhaus! Thank you very much!!

PR #2931, opened 2020-02-13
fix(@clayui/popover): make 'show' prop actually usable

### LINKED ISSUE #2738: ClayPopover `show` property is not working
-   [ x ] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

https://github.com/liferay/clay/pull/2737

Set the property `show` of ClayPopover component to `false`.

ClayPopover don't be shown.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.0.0  |
| React | v16.9.0 |

bryceosterhaus:
fixes #2738

This fix may look a little odd, and here is my line of thinking...

1. We initially released and documented a `show` prop on the components interface.
2. However, we don't use it for anything and handle all show/hide internally.
3. Since we initially released it on the interface and because we typically release "controlled" components, we should provide a full controllable API.
4. In order to keep current functionality and not break the internal show/hide, make it so we can do both.
5. Conclusion: make it usable via both controlled and uncontrolled.

Any concerns with this?

The other option would be to just remove it from the interface since it doesnt actually do anything and act like nothing happened

### DIFF
--- a/packages/clay-popover/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-popover/src/__tests__/__snapshots__/index.tsx.snap
@@ -28,6 +28,34 @@ exports[`ClayPopover renders 1`] = `
 </div>
 `;
 
+exports[`ClayPopover renders with \`show\` 1`] = `
+<div>
+  <div
+    class="popover clay-popover-bottom show"
+  >
+    <div
+      class="arrow"
+    />
+    <div
+      class="inline-scroller"
+    >
+      <div
+        class="popover-header"
+      >
+        Popover
+      </div>
+      <div
+        class="popover-body"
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
Another option would be to mark this as a deprecated and remove it in future majors, as this API had no effect people will not feel changes 😁 but I think it makes sense to go with those changes that you are making. Perhaps the ideal would be to just support the use of the controlled component but in that case I'm fine with going with that.

bryceosterhaus:
@matuzalemsteles yeah I agree, I think keeping the "controlled" aspect was  the right initial API and most in line with our current designs.

Does this PR okay to merge?

PR #2936, opened 2020-02-14
feat(@clayui/alert): add hideCloseIcon prop

### LINKED ISSUE #2935: @clayui/alert Not possible to use autoClose and not display the close icon
https://codesandbox.io/s/inspiring-glade-t38n1

bryceosterhaus:
fixes #2935


bryceosterhaus:
@matuzalemsteles the case is where you want something to show on load but be hidden quickly. Sometimes users don't always want to display the close button. I think it's a more rare case, but I've definitely seen the pattern before outside of liferay.

matuzalemsteles:
@bryceosterhaus no problem, just a concern that I had. I think this is good to go.

PR #2937, opened 2020-02-17
feat(@clay-components): Empty State Component Draft

kresimir-coko:
Hey @bryceosterhaus here I have the draft for this component, stuff is gonna fail, CI is failing locally saying there's a problem in /clay-empty-state/package.json but without any leads. Also tests are failing due to svg characters. I just to get your thoughts on the approach and the initial API


packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayButton from '@clayui/button';
+import React from 'react';
+
+const emptyStateImg = require('../images/empty_state.gif');
+const searchStateImg = require('../images/search_state.gif');
+const successStateImg = require('../images/success_state.gif');
```
bryceosterhaus: So I am thinking that instead of actually supplying these images, we would just add a prop on the component for the path to the image.

```jsx
<ClayEmptyState imageSrc="/some/path/to.gif" si />
```

packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const DISPLAY_STATES = {
```
bryceosterhaus: If we move the imageSrc to be a prop, we won't need these states.

packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return (
+		<div
+			className={`empty-state-root ${!imageState &&
+				'empty-state-root--no-image'}`}
+		>
+			{imageState && <img alt="empty-state-image" src={imageState} />}
+
+			<h2>{title || 'No results found'}</h2>
```
bryceosterhaus: Rather than put the `||` here, we typically just set a default value to the prop. `title = 'No results found',`

packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+			<div>{description || 'Sorry, there are no results found'}</div>
```
bryceosterhaus: Same as above comment

packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+			{actionLabel && onClickAction && (
+				<div className="empty-state-action">
+					<ClayButton displayType="secondary" onClick={onClickAction}>
```
bryceosterhaus: Instead of hardcoding a button here, lets just have people pass the action element as either `{children}` or `actionRenderer` prop. I am leaning towards using `children` for this so that it gives users more flexibility in how they want to use the component. 

It would look like
```jsx
{children && <div className="empty-state-action">{children}</div>}
```

This would also allow us to remove `actionLabel` and `onClickAction` props.

packages/clay-css/src/scss/components/_empty-state.scss
```diff
@@ -0,0 +1,15 @@
+.empty-state-root {
```
bryceosterhaus: @pat270 should we be adding css for this or do you already have markup for the empty state?

packages/clay-empty-state/src/EmptyState.tsx
bryceosterhaus: @jbalsas what do you think about this, should the images be internal to the component itself?

kresimir-coko:
@bryceosterhaus I've force pushed the changes, including adding of tests and stories, adding one more style rule, and some SF stuff. If there are more or different tests/stories that need to be added, let me know!

packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const ClayEmptyState: React.FunctionComponent<IProps> = ({
+	children,
+	description = 'Sorry, there are no results found',
+	imageSrc,
+	title = 'No results found',
+}) => (
+	<div
+		className={`empty-state-root ${!imageSrc &&
```
bryceosterhaus: You can use our `classNames` dependency we use in a lot of other packages and convert this to.

```jsx
className={classNames('empty-state-root', {
	'empty-state-root--no-image': !imageSrc,
})}
```

packages/clay-empty-state/src/EmptyState.tsx
```diff
@@ -0,0 +1,46 @@
+/**
+ * © 2020 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
```
bryceosterhaus: Instead of extending all attributes from the div element here, lets move these extra html attribute props to its own prop, `containerProps`. This would be similar to Checkbox, Radio and Toggle. 

Example:
```ts
interface IProps {
	containerProps?: React.HTMLAttributes<HTMLDivElement>
	// ...
}

const ClayEmptyState: React.FunctionComponent<IProps> = ({
	containerProps = {},
	//...

	const {className, ...otherContainerProps} = containerProps;

	return <div className={classNames(className, ...)} {...otherContainerProps}
```

Reason I recommend `containerProps` instead of just extending is because we would have a conflict with `title` and `containerProps.title`. I think it'll be more common for people to use this component with only title, description, and imageSrc.

packages/clay-empty-state/stories/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import '@clayui/css/lib/css/atlas.css';
+import ClayButton from '@clayui/button';
+import {storiesOf} from '@storybook/react';
+import React from 'react';
+
+const emptyImage = require('../images/empty_state.gif');
```
bryceosterhaus: lets move these files into `clay-empty-state/stories/images` rather than leave at the root level.

kresimir-coko:
Hey @bryceosterhaus I'm unsure how I ended up closing the PR, and I'm unsure what's the proper way of reopening it again without making a worthless commit. Perhaps I pushed on upstream instead of origin, I'm unsure.

wincent:
@kresimir-coko: do you see the "Reopen" button? Gonna click it now...

feat__clay-components___Empty_State_Component_Draft_by_kresimir-coko_·_Pull_Request__2937_·_liferay_clay

kresimir-coko:
> @kresimir-coko: do you see the "Reopen" button? Gonna click it now...
> 
> feat__clay-components___Empty_State_Component_Draft_by_kresimir-coko_·Pull_Request__2937·_liferay_clay

It was greyed out for me, when hovering it said i needed to add a new commit.

wincent:
> It was greyed out for me, when hovering it said i needed to add a new commit.

Another black mark for my "GitHub is terrible" list. 💀

packages/clay-css/src/scss/components/_empty-state.scss
pat270: We don't have anything for empty state right now. Here are my suggestions:

Instead of relying on `h2` to size and space the text (e.g., No results found), we should scope it to a class like `empty-state-text`. This prevents styles from bleeing into the `empty-state` component if someone modifies `h2`.

Also the text for the `h2` element is limited to one line, you could wrap it in `<span class="text-truncate-inline"><span class="text-truncate"></span></span>` and make sure to add a `title` attribute with the full text.

We need to scope the message text inside `<div></div>` to a class `empty-state-message` and style it to match the design.

The image needs to be responsive just add class `img-fluid` (e.g., shrink to fit container size for narrow containers).

Empty State needs to have `max-width: 340px` and shrink to fit in narrow containers.

We don't delimit class names with double dash in Clay CSS (e.g., `.empty-state-root--no-image` should be `.empty-state-root-no-image`).

The way I would structure this component is:
```
<div class="empty-state empty-state-animation">
  <div class="empty-state-container">
    <div class="empty-state-image"></div>
    <div class="empty-state-text"></div>
    <div class="empty-state-message"></div>
    <div class="empty-state-actions"></div>
  </div>
</div>
```

bryceosterhaus: cc @kresimir-coko

kresimir-coko: @pat270 Why introduce so many `div`s to the component?
I personally dislike having all that obsolete DOM. Putting 2 `spans` over the `h2` feels painful, let alone introducing 5, or so, new `divs`.
If that's the way we do things I'll do it that way, but it just feels wrong

jbalsas: We already have this component in DXP, we shouldn't need to rewrite it entirely from scratch... can't we just port over what we have over there?

kresimir-coko: @jbalsas I'm not sure how the porting process would look exactly, but I did something similar I assume. I copied the implementation from DXP and then made everything specific for Clay on top of that (stories, tests...)

pat270: @kresimir-coko I wouldn't call it obsolete. They provide classes for devs to target if they need to modify it for some reason and prevent styles from bleeding into the component.

The spans for truncating text is the most failsafe way for truncating text I've found. It's pretty much a CSS hack. If there is a better way I'm always open to suggestions. Additional markup is a side effect of trying to be modular with our CSS components, it's a trade off. We can provide a class that truncates text in specific situations or we can provide two spans that always truncates text.

This PR doesn't fully implement the rules specified at https://liferay.design/lexicon/core-components/empty-states/. It also doesn't take into account the possibility of long text with no white space. For Empty Search Results, what if someone searches for ReallySuperInsanelyJustIncrediblyLongAndTotallyNotPossibleWordButWeAreReallyTryingToCoverAllOurBasesHereJustInCaseSomeoneIsNutsAsPerUsual? How will the text be displayed?

kresimir-coko: You make a valid point @pat270 , I agree on the `spans`, it's the tool we have to achieve something, I don't know how those 2 spans work because this is the first time I'm seeing them, but I see their value. But on the topic of `divs` while yes, providing flexibility is awesome, and I haven't thought about that aspect as much as you have, I assumed that giving the `p` a class would be enough.

When it comes to the long text with no space, I haven't seen that in the spec.

I'll add the changes you advised and push when I get the chance :+1:

pat270: @kresimir-coko long text no space is never in the spec, but always gets filed as a bug in every component. You can structure the markup how you want, the above was just a quick suggestion. We should have specific classes for most of elements though.

bryceosterhaus:
@kresimir-coko whats the status on this component?

kresimir-coko:
> @kresimir-coko whats the status on this component?

@bryceosterhaus  Since we were waiting for Patrick I started working on InputMoveBoxes, so I thought to leave the final changes to this one until after I finish InputMoveBoxes

pat270:
@kresimir-coko I didn't realize you were waiting for me. Did you want me to send a pr with markup and css?

bryceosterhaus:
@pat270 I think it would probably be best for you to send in the markup & css and then we can add this after. I trust your knowledge of the html/css structure much more than I trust myself 😆

kresimir-coko:
@pat270 the comment about waiting for you was related to week ago, not right now, so no worries. But yea as @bryceosterhaus said feel free to send the markup & css and then we can piggyback on that 🙃

kresimir-coko:
Hey @bryceosterhaus I've pushed the (probably) last commit for this PR. It adjusts the markup and image imports to fit @pat270's work. I've tested it with his commits on top of mine and it looks good.

bryceosterhaus:
thanks @kresimir-coko! Once the markup gets merged I will take a look here and get this merged.

bryceosterhaus:
@kresimir-coko I took your PR and squashed the commits and added some changes on top. Take a look at https://github.com/liferay/clay/pull/2989

PR #2942, opened 2020-02-19
chore(clayui.com): clean up clayui.com

bryceosterhaus:
handful of things on the clay site I have been wanting to clean up. I think it helps make things a bit more readable as the alerts were a bit distracting on every page.

**Before:**
Screen Shot 2020-02-19 at 11 27 40 AM

**After:**
Screen Shot 2020-02-19 at 2 34 33 PM

~Another option would be to create some sort of label or badge for our demos as well.~ added


jbalsas:
One tiny nit: there's no feedback when clicking on the "Copy" button. It's not clear if anything happened. One common pattern I've seen is to show a tooltip with the "Text copied to the clipboard" or something like that when the action is completed so users know they can use `cmd+v` to paste it.

jbalsas:
Another tiny nit: in our own demo, the alert icon and the text are way too close.

Screen Shot 2020-02-20 at 10 38 20

We had a related issue about it and I think we said this should be fixed in user-land... so maybe we should update our demo to show what that means :)

jbalsas:
Final nit for today 😂: All tables seem to have horizontal scroll (at least in my MB15" screen). The container is smaller than the table. Even though some columns are empty, we still give them `minw-150` which pushes the important content outside the viewport 🤔 

Screen Shot 2020-02-20 at 10 40 40

bryceosterhaus:
@jbalsas updated. I added some feedback for the "copy" action. Its not the best, but I think its good for now. Ideally we would have everything be an mdx file and both the css and react side would be the same, this would make it much easier to implement a proper tooltip for it.

I also updated alerts for having the space and then consulted with Patrick to make the tables a bit better. We ended up making the middle section wider, I think it should help.

packages/clay-alert/src/index.tsx
```diff
@@ -138,7 +138,7 @@ const ClayAlert: React.FunctionComponent<IClayAlertProps> & {
 		>
 			<ConditionalContainer>
 				<div className="alert-autofit-row autofit-row">
-					<div className="autofit-col">
+					<div className="autofit-col inline-item-before">
```
jbalsas: @dgarciasarai, not sure if I'm remembering correctly, but was this what we discussed in the past?

jbalsas:
LGTM!

packages/clay-alert/src/index.tsx
bryceosterhaus: @jbalsas this was the same fix we had in the other ticket. https://github.com/liferay/clay/pull/2762/files#diff-ba7147c142a149f7ae7a4b2c0874be71R108

jbalsas: Yeah, I was wondering why we didn't apply it in the first place... 🤔 . I thought we told @dgarciasarai she should add that class in her own code, not sure if it's the same case, though

PR #2944, opened 2020-02-21
fix(@clayui/tabs): adjust class name timing so that it the tab properly fades in

### LINKED ISSUE #2941: Tabs flash when fade is enabled
https://clayui.com/docs/components/tabs.html

Go here: https://clayui.com/docs/components/tabs.html

Click between tabs in the example on the page. The tab content flashes when switching.

Tab content should fade in smoothly.

## Environment

https://clayui.com/docs/components/tabs.html

bryceosterhaus:
fixes #2941

### DIFF
--- a/packages/clay-tabs/src/TabPane.tsx
+++ b/packages/clay-tabs/src/TabPane.tsx
@@ -19,11 +19,10 @@ interface ITabPaneProps extends React.HTMLAttributes<HTMLDivElement> {
 	fade?: boolean;
 }
 
-const delay = (fn: Function) => {
-	return setTimeout(() => {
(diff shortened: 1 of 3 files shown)


bryceosterhaus:
@ethib137 can you verify this is what you would expect? https://deploy-preview-2944--storybook-clayui.netlify.com/?path=/story/components-claytabs--w-state

ethib137:
@bryceosterhaus yeah, that's definitely much better.

bryceosterhaus:
@matuzalemsteles nope, just wanted to make sure another pair of eyes saw it. merging....

PR #2948, opened 2020-02-24
chore: update stale snapshots

wincent:
Looks like e393c07f3f9b21bb7f146037d7372e1b updated some but not all of the snapshots.


wincent:
@bryceosterhaus: I am just assuming that the snapshot is stale and there's no actual bug here, but I could be wrong — I trust you'll know either way! 😂

bryceosterhaus:
Thanks for sending this. I realized I had these updated on a different branch but forgot to push before I merged the the branch.

PR #2949, opened 2020-02-24
feat(@clay-components): Input Move Boxes - Draft implementation

kresimir-coko:
Here's the draft implementation of the Input Move Boxes component.
Note: 
The ordering and transferring works only for 1 selection at the moment, in order to see if my approach is valid before pursuing, and also to maybe get a piece of advice on how best to do it, since outside of some convoluted array manipulation I didn't get an idea of how to do it.

Tests are failing unrelated to my code, at least from what I can see locally, ClayAlert is the culprit.

### DIFF
--- a/packages/clay-css/src/scss/_components.scss
+++ b/packages/clay-css/src/scss/_components.scss
@@ -33,6 +33,8 @@
 @import "components/_form-validation";
 @import "components/_icons";
 @import "components/_input-groups";
+@import "components/_input-move-boxes";
+@import "components/_select-box";
 @import "components/_list-group";
 @import "components/_modals";
 @import "components/_multi-step-nav";
--- a/packages/clay-css/src/scss/components/_input-move-boxes.scss
+++ b/packages/clay-css/src/scss/components/_input-move-boxes.scss
@@ -1,0 +1,23 @@
+.input-move-boxes {
+    display: flex;
+    max-width: 768px;
+    width: 100%;
+
+    .select-box {
+        position: relative;
+        display: flex;
+        flex-direction: column;
+        width: 40%;
+    }
+}
+
+.transfer-buttons {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    margin: 0 24px;
+
+    .btn {
+        flex: 0 1 auto;
+    }
+}
--- a/packages/clay-css/src/scss/components/_select-box.scss
+++ b/packages/clay-css/src/scss/components/_select-box.scss
@@ -1,0 +1,6 @@
+.select-box-order-buttons {
+    bottom: 8px;
+    right: 8px;
+    position: absolute;
+    z-index: 9;
+}
--- a/packages/clay-empty-state/src/__tests__/index.tsx
+++ b/packages/clay-empty-state/src/__tests__/index.tsx
@@ -3,12 +3,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-/**
- * © 2019 Liferay, Inc. <https://liferay.com>
- *
- * SPDX-License-Identifier: BSD-3-Clause
- */
-
 import ClayEmptyState from '..';
 import {cleanup, render} from '@testing-library/react';
 import React from 'react';
--- a/packages/clay-form/package.json
+++ b/packages/clay-form/package.json
@@ -26,6 +26,7 @@
 		"react"
 	],
 	"dependencies": {
+		"@clayui/button": "^3.3.0",
 		"@clayui/icon": "^3.0.4",
 		"classnames": "^2.2.6"
 	},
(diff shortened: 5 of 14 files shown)


pat270:
You can probably piggy back off of `form-group-autofit`. For example:
```
<div class="clay-input-move-boxes form-group-autofit">
    <div class="form-group-item">
        <label for="a">Current</label>
        <select class="form-control" id="a" name="a" multiple></select>
    </div>
    <div class="form-group-item form-group-item-shrink">
        <button class="component-action" type="button">gt</button>
        <button class="component-action" type="button">lt</button>
    </div>
    <div class="form-group-item">
        <label for="b">Available</label>
        <select class="form-control" id="b" name="b" multiple></select>
    </div>
</div>
```
also the css component should be configurable via Sass variables.

kresimir-coko:
Hey @bryceosterhaus I've force pushed the changes I made today.
I have some questions:
1. Which tests and stories to write?
2. What do we do about the buttons, do we absolute position them inside the select?
3. What do we do about size relative to the buttons too? The buttons are supposed to be stuck to the bottom of the select box
4. InputMoveBoxes.es.js:107 -> should this line be moved to the story, is that what you meant by having a callback in there, like I changed with the Select Box in index.js:107?

pat270:
@kresimir-coko I'll go ahead and send a pr with markup and css for this component also

bryceosterhaus:
> 1. Which tests and stories to write?

Write tests to cover all interactions such as re-ordering the list, moving an item from left to right, selecting multiple, etc. And one basic story is probably sufficient enough with `knobs` for toggling the reordering and multiple flag

> 2. What do we do about the buttons, do we absolute position them inside the select?
> 3. What do we do about size relative to the buttons too? The buttons are supposed to be stuck to the bottom of the select box

Let's wait for @pat270 to send in the markup and css for this and we can just piggy back on whatever structure he sees as best.

> 4. InputMoveBoxes.es.js:107 -> should this line be moved to the story, is that what you meant by having a callback in there, like I changed with the Select Box in index.js:107?

Yeah for `InputMoveBoxes`, we actually don't need internal state for left/right items. We can make this a controlled component so that we get the items from the prop and then we provide an `onChange` callback.

For example, this is a super simple representation of how you might go from internal state, to making it a controlled component.
```js
// from this
(items) => {
	const [itemsLeft, setItemsLeft] = React.useState(left.items);
	const [itemsRight, setItemsRight] = React.useState(right.items);

}

// to this
(items, onChange) => {
     const [itemsLeft, itemsRight] = items;

     // these are just helper functions, but you would be fine to just use `onChange` by itself
     const setItemsLeft = (newItems) => onChange([newItems, itemsRight]);
     const setItemsRight = (newItems) => onChange([itemsLeft, newItems]);
}

// or if we split it into left/right props we would use the `onChange` and `items` from the props itself
({left, right}) => {
// ... 
}
```

And then if you did this, we would be able to write the story like
```jsx
const [items, setItems] = useState([[1,2,3], [4,5,6]]);

<ClayInputMoveBoxes
     items={items}
     onChange={setItems}
/>

// or if we split it into left/right props we would do...

<ClayInputMoveBoxes
	left={{
		items: items[0],
		onChange: (newLeftItems) => setItems([newLeftItems, items[1]]),
	}}
	right={{
		items: items[1],
		onChange: (newRightItems) => setItems([items[0], newRightItems]),
	}}
/>
```

kresimir-coko:
@bryceosterhaus Hey, I've made the component fully controlled and added some tests, got stuck on the transfer selected options test, in which the expect line is commented out so the tests pass. So, these commits are here so that you have something to leave comments on.

kresimir-coko:
Hey @bryceosterhaus I've modified the markup of this one too to fit Patrick's. The test for transferring items between lists is still failing, I haven't found a solution yet.

kresimir-coko:
Hey @bryceosterhaus I've updated the types, there's a bunch of them :) 
Tests are still incomplete, the one for transferring items between lists is commented out just so CI doesn't fail. I'm waiting on you to give me some feedback on them. Also, didn't get to write JSDoc for the props, ran out of time.

bryceosterhaus:
@kresimir-coko were you able to look at and apply the PR I sent you at https://github.com/kresimir-coko/clay/pull/1 ? I had made some changes to the API and also worked on the tests you were talking about. Let me know if these changes made sense to you. And if they do make sense, could you apply your changes? Feel free to squash my commits and author them yourself.

I changed the API to take a single `items` and `onChange` prop so that we don't have to call two separate `onChange` functions from left/right. This should help us avoid unnecessary re-renders or things being out of sync. I also added `internalLeftSelected` `internalRightSelected` so that the "selected" props are not required. This still gives us a controlled API when we want, but I think most of the time when people use `DualListBox`, they won't care about the selections, but rather just the end result of `items`.

Let me know what you think

kresimir-coko:
Hey @bryceosterhaus I've squashed all the commits into one and force-pushed it here. I've updated the types and written JSDoc on top of your commits.

bryceosterhaus:
Updated PR https://github.com/liferay/clay/pull/3033

PR #2947, opened 2020-02-24
chore: update for compliance with current Outbound Licensing Policy

wincent:
This change is analogous to the one I made in:

- https://github.com/liferay/eslint-config-liferay/pull/151
- https://github.com/liferay/liferay-npm-tools/pull/394
- https://github.com/liferay/liferay-js-themes-toolkit/pull/441

It updates the license header format to match the latest policy:

https://grow.liferay.com/excellence/Liferay+Outbound+Licensing+Policy

Did this by updating `copyright.js` and running `yarn lint:fix`. I used a Vim macro to strip out the original headers. Note some files required manual fixing because they had ESLint or other comments at the top of the file (this is bad; the header should always be at the top). Some of the comments I could just remove (eg. unnecessary lint suppressions), but the others I moved below the header.

Adds a copy of the license to `LICENSES/MIT.txt` as required by the policy too.

PR #2950, opened 2020-02-24
chore: update lints

wincent:
Updates to eslint-config-liferay v19.0.2 which brings in some new rules:

- curly (enforces use of curlies around conditional blocks).
- liferay/import-extensions (prohibits superflous extensions in imports).
- liferay/array-is-array (enforces use of `Array.isArray` instead of `instanceof Array`).
- liferay/sort-import-destructures (yet another in the family of "sorting" rules).
- padding-line-between-statements (specifically, requiring a blank line before `return` statements, at Brian's request).
- react/forbid-foreign-prop-types (so that we can be sure that stripping propTypes in production builds is safe; obviously won't be relevant in this repo).

and produces [these new lint errors](https://gist.github.com/wincent/84bbd941eec684218acb21ad34fdb156).

All of those have autofixes, so I applied them.

I also removed some cruft from the config file (values that are not needed because they are the defaults).

Note that in testing this, I discovered that the new "liferay/sort-import-destructures" rule and the `@typescript-eslint/parser` parser were not playing nicely together. I first tried updating some of the related NPM packages (ESLint, the parser etc) but that didn't work (although it did require me to update one config value), so I then dug deeper and [tweaked the rule implementation to work with both](https://github.com/liferay/eslint-config-liferay/pull/153) the `@typescript-eslint/parser` and "espree" (ie. default) parsers:


wincent:
Marking this as a draft @bryceosterhaus because it will conflict with https://github.com/liferay/clay/pull/2947 — once that one is in I will just re-run the autofixes on this branch and update it. Expect the tests to fail because of the stale snapshots ([fixed here](https://github.com/liferay/clay/pull/2948)).

wincent:
Rebased on top of https://github.com/liferay/clay/pull/2947 (although I was wrong about it conflicting — it didn't) and taking out of draft status.

wincent:
Two stalled jobs again:

- https://app.netlify.com/sites/storybook-clayui/deploys/5e54fcedb17a5e00081eb426

That seem to have completed just fine:

> Build time: 4m 12s. Total deploy time: 4m 12s
>
> Build started at 11:54:39 AM and ended at 11:58:51 AM. Learn more about build minutes.

bryceosterhaus:
Tests are passing so that is good for me. The changes don't look like they would cause issue either.

Thanks for adding this!

PR #2952, opened 2020-02-25
feat(clayui.com): add playground for css

### LINKED ISSUE #2867: Paste Sample Code Feature
Based on https://github.com/liferay/clay/issues/2843#issuecomment-574905300:

> My test markup is included at the bottom. You can copy paste it into clayui.com.

It might be a good idea to have some type of "Sandbox" page or feature where we could simply do what @pat270 suggests. That is:
- Go to a page
- A "Paste your HTML Sample" textarea appears
- When pasted, the HTML is injected somewhere to validate the result

Just an idea, feel free to discard it 😉

bryceosterhaus:
fixes #2867

Demo: https://deploy-preview-2952--next-clayui.netlify.com/docs/css/playground.html

**Page:**
Screen Shot 2020-02-25 at 3 20 12 PM

**Sidebar Location:**
Screen Shot 2020-02-25 at 3 20 18 PM

### DIFF
--- a/clayui.com/content/docs/css/playground.mdx
+++ b/clayui.com/content/docs/css/playground.mdx
@@ -1,0 +1,9 @@
+---
+title: 'Playground'
+description: 'Live editor for testing out @clayui/css'
+disableTOC: true
+---
+
+import Playground from '../../../src/components/Playground/index';
+
+<Playground />
--- a/clayui.com/gatsby/onCreateNode.js
+++ b/clayui.com/gatsby/onCreateNode.js
@@ -20,6 +20,7 @@ module.exports = exports.onCreateNode = ({actions, getNode, node}) => {
 		const {
 			alwaysActive,
 			description,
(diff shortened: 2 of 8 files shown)


bryceosterhaus:
@matuzalemsteles updated the size of the titles of the panes. Also, instead of adding a link to the playground page, I added a codesandbox link to our nav. We we linked to our playground, it would require a bit more maintenance I think. I think the more we can push people to use csb for demos, the better since we won't have to maintain our own editor as heavily. The playground still seems beneficial for just the css side though.

matuzalemsteles:
@bryceosterhaus no problem. makes sense. It seems that the site build is failing due to the `packageStatus` field not being present.

bryceosterhaus:
Not sure why it shows some of the netlify tasks hanging, they show completed if you click into the netlify link. So overall I think CI is good. @matuzalemsteles anything else you see before merging?

PR #2966, opened 2020-02-27
chore(release): publish updated packages

bryceosterhaus:
One thing I noticed is that `lerna version` seemed to pick up this commit(https://github.com/liferay/clay/commit/38e132b70904f44ed9e02b129e49672589a2d8b8#diff-ab3780567c21247fcd9946d4d3056e99) and updated every package since this touched everything. Do we see this being a major issue? I was trying to find a way to only look at the non-chore commits, but it looks like lerna is the one making this change and not the `--conventional-commits` flag.

Any ideas to fix this or concerns if we just go ahead and bump everything? Not sure if license updates would automatically require a patch regardless.


wincent:
If the code changes, even if only because a linter added a blank line, it's still new code, so that sounds like a "patch" level release to me. I don't think we want to get into the game of trying to decide whether a given "chore" PR (much less all "chore" PRs) is something that can be considered to have substantially changed the code and therefore "warrants" a release. For example, we might label a dependency update as a "chore" even though it might bring in a large change to the dependency graph. Somebody else making the same change might label it a "feat" if they think it unlocks new functionality, or "fix" if they think it fixes a bug — and the problem with that is that there is an element of human judgment in applying the right label, so there is potential for human error, or for not agreeing about what the "right" label is. In other words, I think it is simpler to just view these version bumps as part of periodic maintenance, especially given that this kind of blanket change probably won't happen *that* often.

Releases that *only* contains that kind of change are noise, but they don't really *hurt* anybody — if `@clayui/nav` bumps from 3.0.6 to 3.0.7, I should be able to update to it as part of routine maintenance without any fear.

bryceosterhaus:
Makes sense to me. So we can just assume that any package with changed files will _always_ result in a patch version.

PR #2967, opened 2020-02-28
feat(@clayui/drop-down): adds section API support to drop-down items

### LINKED ISSUE #2811: Dropdown Menu: Checkbox and Radio hover interaction styles inherited

1. Go to [Clay Storybook](https://storybook.clayui.com/)
2. Open ClayDropdown menu item
3. Select Checkbox or Radio 
4. Click "Click me" button

## Current behavior

checkbox-hover

Dropdown item hover or selected interaction styles mustn't be inherited to check boxes and radio buttons

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.0  |

matuzalemsteles:
Fixes #2811

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -12,11 +12,11 @@ import Caption from './Caption';
 import Divider from './Divider';
 import Group from './Group';
 import Help from './Help';
-import Section from './Section';
 import Item from './Item';
 import ItemList from './ItemList';
 import Menu, {Align} from './Menu';
 import Search from './Search';
+import Section from './Section';
 
 interface IProps extends React.HTMLAttributes<HTMLDivElement | HTMLLIElement> {
 	/**
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -12,6 +12,7 @@ import Caption from './Caption';
 import Divider from './Divider';
 import Group from './Group';
 import Help from './Help';
+import Section from './Section';
 import Item from './Item';
 import ItemList from './ItemList';
 import Menu, {Align} from './Menu';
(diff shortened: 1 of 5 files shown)


packages/clay-drop-down/src/Item.tsx
```diff
@@ -28,6 +28,13 @@ interface IProps
 
 	innerRef?: React.Ref<any>;
 
+	/**
+	 * Flag to indicate that the element is a session, are for cases when
+	 * Checkbox and Radio do not need the styles when hovering over the
+	 * drop-down item.
+	 */
+	isSection?: boolean;
```
matuzalemsteles: I'm not really sure about that name, I thought about maybe just leaving `section`. Suggestions?

bryceosterhaus: Yeah I think just `section` is a better fit for this, unless we want to create a new "Section" component for dropdown, but I think just a flag would be enough.

packages/clay-drop-down/src/Item.tsx
```diff
@@ -28,6 +28,13 @@ interface IProps
 
 	innerRef?: React.Ref<any>;
 
+	/**
+	 * Flag to indicate that the element is a session, are for cases when
+	 * Checkbox and Radio do not need the styles when hovering over the
+	 * drop-down item.
+	 */
```
bryceosterhaus:
```suggestion
	/**
	 * Flag to indicate that the element is a section. This is used for cases like
	 * Checkbox and Radio where they do not need hover styles.
	 */
```

packages/clay-drop-down/src/Item.tsx
matuzalemsteles: Hmm, I think a new component for "Section" makes sense, but in order not to duplicate it we would have to have an ItemBase to reuse the logic. What do you think? I think I like to have a new component for that even though what will only change is the className 😅

bryceosterhaus: @matuzalemsteles new component is fine with me. Any idea if we need to support the icon for a section? If not, we wont need to duplicate too much

matuzalemsteles: @bryceosterhaus I don't know but we can go without the icon support for now.

matuzalemsteles:
@bryceosterhaus updated.

PR #2978, opened 2020-03-04
feat(@clayui/label): Adds `truncated` option for ClayLabel component

### LINKED ISSUE #2977: ClayLabel don't have an option for truncating texts inside
-   [✅] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

There isn't a way to truncate the text inside the ClayLabel component just using a property like `trucated` on 2.x(https://github.com/liferay/clay/pull/2858)

An option for truncating text inside ClayLabel without passing `text-truncate-inline` to the component. Like:

```jsx
<ClayLabel truncate>{'my lovely label'}</ClayLabel>
```

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.1.0 |
| React | Doesn't matter |

diegonvs:
Fixes #2977﻿

### DIFF
--- a/packages/clay-label/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-label/src/__tests__/__snapshots__/index.tsx.snap
@@ -97,3 +97,15 @@ exports[`Rendering with a different displayType  1`] = `
   </span>
 </span>
 `;
+
+exports[`Rendering with truncated text 1`] = `
+<span
+  className="label label-secondary"
+>
+  <span
+    className="label-item label-item-expand text-truncate"
+  >
+    LaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLaLa
+  </span>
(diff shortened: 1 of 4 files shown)


bryceosterhaus:
I think rather than adding a `truncate`, what if we added either `labelItemClassName`,`labelItemProps`, or if `labelItem` is too confusing we could do `innerProps`. This would give us more flexibility and room for users to add various utility classes. Otherwise establish a pattern where we need to add all sort of text utility props for the inner portion.

Thoughts?

diegonvs:
I agree, but we could at least document it by adding a truncate knob to the storybook. What do you think?

bryceosterhaus:
@diegonvs instead of a knob, lets just add a new story for labels called `w/ long text` and then show it being used there.

bryceosterhaus:
@diegonvs I didn't realize at first that you were adding the truncate to the inner element. I went ahead and added a prop for that. https://github.com/liferay/clay/pull/3060

PR #2980, opened 2020-03-05
feat(clay-css): Adds Dual Listbox component (input move box replacement)

pat270:
There are some issues with `select[size]` and `select[multiple]` in Firefox and Safari. They render the scrollbar track even if there is no overflowing content. I couldn't work around this with native inputs in those browsers so there is a potential "bug" that could pop up. On the other hand they look nice on Chrome, IE11, and Edge.

@kresimir-coko you can find the markup for the component on my test site under Form Elements (Dual Listbox). @jonmak08 mentioned he put in a lot of work with keyboard shortcuts in the old Input Move Boxes component. ~I tried look for the ticket, but can't find it.~ Here it is https://issues.liferay.com/browse/LPS-77051. We should find out what shortcuts we need.

clay-dual-listbox

### DIFF
--- a/packages/clay-css/src/content/form_elements.html
+++ b/packages/clay-css/src/content/form_elements.html
@@ -195,6 +195,36 @@ section: Components
 		</select>
 	</div>
 	<div class="form-group">
+		<label>A Div Styled Like a Select Element</label>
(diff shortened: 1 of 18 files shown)


marcoscv-work:

PR #2981, opened 2020-03-06
chore: add gh action for checking package size

bryceosterhaus:
Quick demo of what I am imagining. I did end up adding `parcel-bundler` which adds some dependency bloat, but it seemed to be the simplist bundler to implement and compare against.

I want to iterate more on this and provide a cleaner UI, but I am curious to get more thoughts on this.


bryceosterhaus:
I'm not sure if the GH action will trigger on here, but here is what it looked like on my fork.

https://github.com/bryceosterhaus/clay/pull/28

.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
...
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: '10.x'
+            - run: npm install -g yarn
+            - run: yarn install
+            - run: yarn checkSize
+            - name: Comment on PR
+              uses: machine-learning-apps/pr-comment@master
```
wincent: From a security perspective would be slightly better to specify a tagged version instead of "master". I see they have [a "1.0.0" tag](https://github.com/machine-learning-apps/pr-comment/tree/1.0.0), for example. Just because the action receives `GITHUB_TOKEN` and I don't know what scope for naughty things it could potentially do with that in the future if their project got compromised. (Of course, an attacker who gained control of their repo could also just re-write the 1.0.0 tag, but using the tag is still a partial mitigation of risk, and in any case shields us from bugs that might creep into master)

Having said all that, I see there is also [unsplash/comment-on-pr](https://github.com/unsplash/comment-on-pr) which seems to be slightly more used (more stars), and [this Stack Overflow question](https://stackoverflow.com/questions/58066966/commenting-a-pull-request-in-a-github-action) mentions doing this [with just `curl`](https://stackoverflow.com/a/58704739/2103996) or [with `actions/github`](https://github.com/actions/github) (which seems to be a little more widely used again).

scripts/processMarkdownTable.js
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+function processMarkdownTable(dataJson) {
+	let table = `
+|Package|Size|
+|----|----|`;
+
+	Object.keys(dataJson).forEach(package => {
```
wincent: As you're using Node 10, I think you can safely use `Object.entries` here.

I'd avoid using `package` as an identifier though because it is [a reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords), at least in strict mode.

ie. `Object.entries(dataJson).forEach(([packageName, size]) => { ... }`

scripts/check-size.js
```diff
@@ -0,0 +1,82 @@
...
+const processMarkdownTable = require('./processMarkdownTable');
+
+const WORKSPACE_PACKAGES_WHITELIST = [
+	'browserslist-config-clay',
+	'demos',
+	'generator-clay-component',
+];
+
+const convertBytes = function(bytes) {
+	const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
```
wincent: `TB` 😂

wincent:
> I did end up adding parcel-bundler which adds some dependency bloat

One thought that did occur to me about this one: if you wanted to avoid the dependency, we don't actually need to make a "real" bundle — just concatenating the files would be enough. It would be faster too, although I am frankly pretty amazed at how fast it was in [your test run](https://github.com/bryceosterhaus/clay/runs/489018690) (55 seconds to build and bundle, which is less than the 1m29s that Yarn took to install). The fact that it's so fast gives me hope for the viability of my suggestion above (about doing two builds so that we can show a delta).

And I forgot to ask, any ideas for `clay-css`?

scripts/check-size.js
bryceosterhaus: you never know these days with npm lol

bryceosterhaus:
@wincent Thanks for the feedback. Only reason I didn't include clay-css was mostly unintentional, although I'm not sure we can used parcel for just a css package though. Ill check it out and see.

bryceosterhaus:
I'll resend in a new PR. Going to demo it first on my own fork.

PR #2987, opened 2020-03-09
chore: add gh action for checking package size

bryceosterhaus:
See demo of it on my fork, https://github.com/bryceosterhaus/clay/pull/34#issuecomment-596776206

This first step is just for the js packages, I am working on a proper one for clay-css now and will send a new PR once this looks good.


yarn.lock
```diff
@@ -116,6 +144,17 @@
     levenary "^1.1.1"
     semver "^5.5.0"
 
+"@babel/helper-compilation-targets@^7.8.7":
```
jbalsas: Maybe we can do something to deduplicate these changes? `npx yarn-deduplicate yarn.lock`?

jbalsas:
I just took a glance, but looks promising to me!

One thing strikes as odd from the report:


Package | Previous Size
-- | --
... | ...
@clayui/management-toolbar | 4.7 KB
... | ...
@clayui/nav | 46.9 KB

Why would a package as simple as `nav` weigh `47KB` against the `4.7KB` of a more complicated one such as `management-toolbar`? 

This seems to happen to a series of packages. It looks like we have 2 sets of packages. Those that weigh `~5KB` and those that weight `~50KB`. Might be good to look at the differences between those sets (or independently) to see if we're bundling unnecessary stuff.

scripts/check-size/convertBytes.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const sizes = ['B', 'KB', 'MB', 'GB', 'TB'];
```
wincent: Maybe do `SIZES` like you did for the constants elsewhere.

wincent:
Prior draft of this, for posterity: https://github.com/liferay/clay/pull/2981

bryceosterhaus:
> Why would a package as simple as nav weigh 47KB against the 4.7KB of a more complicated one such as management-toolbar?

One thing to note is that each package we are building includes any `react` and `react-dom` dependencies. I've noticed that packages that use `react-dom` somewhere(including its dependencies), are much bigger. Any time we use a  `Portal`, we have to add `react-dom`, so any package that uses `clay-shared` ends up looking bigger.

> For each package, its own weight (ignoring dependencies).

I included the dependencies and not just the package itself because I think it gives us a more holistic view of the package. If we just did the package, it may look like a small growth in the package but we could add a large dependency like `billboard.js`. We could look at removing react and react-dom from the build size though, those might be good dependencies to ignore.

I think we could improve this in the future by adding a column or row for each package's isolated weight(ignoring dependencies). However I think we might want to wait on that and just look at the whole size for now.

> At its simplest, even a single number — "all of Clay" — is useful once you have a delta

I tried to include something like this in the summary above the able itself.
"2 packages changed. ⬇️ -57.3 KB total changed."

Thoughts on getting this merged as a first step and then iterating on it more?

wincent:
> Thoughts on getting this merged as a first step and then iterating on it more?

I see no harm in merging it and iterating.

I do think, however, that a clue that the way we're dealing with dependencies is off is the fact that you can add up all the individual package weights and get a huge number (1.3 MB) that doesn't map onto the reality of how people actually use Clay (ie. they use it as a related set of packages with common dependencies deduplicated, not as a series of independent packages).

Your observation about needing to include the weight of things like `billboard.js` is a good one. But everything that's in `peerDepencies`, pretty obviously should not be included, and definitely not counted more than once, and same goes for "own" (ie. Clay) dependencies.

bryceosterhaus:
> they use it as a related set of packages with common dependencies deduplicated, not as a series of independent packages

This is a good point, I hadn't thought about that in a way. This conversation helps give me some ideas to iterate on though. I am going to merge this PR and then I'll look to refine it more next week as well as adding clay-css to the build stats.

bryceosterhaus:
more future ideas at https://github.com/liferay/clay/issues/3001

PR #2995, opened 2020-03-11
fix: use ClayLink over anchor tag

### LINKED ISSUE #2991: @clayui/breadcrumb Replace anchor tag in Breadcrumb Item with a ClayLink component
## What is your proposal?
Replace anchor tag in Breadcrumb `Item` with a `ClayLink` component

https://github.com/liferay/clay/blob/master/packages/clay-breadcrumb/src/Item.tsx#L40-L47

Since we now have `ClayLinkContext`, we can provide the context with a different rendering function for `ClayLink`. 

This would allow us to use the `Link` component from `react-router-dom` or anything else we would need to render.

N/A

bryceosterhaus:
fixes #2991


bryceosterhaus:
pretty trivial, going to merge

PR #3007, opened 2020-03-12
fix(@clayui/tooltip): add polyfill for element.matches

### LINKED ISSUE #2998: ClayTooltipProvider throws due to use of "matches" method
Description here:

- https://issues.liferay.com/browse/PTR-1576
- https://issues.liferay.com/browse/LPS-107924

We're using [`matches()` here](https://github.com/liferay/clay/blob/a17ee449be5dc95a173a3595e263a8e7fa6a84e9/packages/clay-tooltip/src/TooltipProvider.tsx#L88).

It's [suggested](https://caniuse.com/#feat=matchesselector) that via the non-standard `msMatchesSelector` alternative we can get at [least partial support on IE](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches).

bryceosterhaus:
fixes #2998

This approach seemed the most straight forward. I had looked into via MDN and stackoverflow and this was the primary way recommended.

If you have any concerns, let me know!

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -76,6 +76,15 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
 	}
 };
 
+// Polyfill to account for any browsers that might not support `element.matches`.
+// Need to make sure window exists first since this can be run server side.
+if (typeof window !== 'undefined' && !Element.prototype.matches) {
+	Element.prototype.matches =
+		// @ts-ignore
+		Element.prototype.msMatchesSelector ||
+		Element.prototype.webkitMatchesSelector;
+}
+
 function closestAncestor(node: HTMLElement, s: string) {
 	const el = node;
 	let ancestor: HTMLElement | null = node;


bryceosterhaus:
Tested on IE11, Chrome, and Firefox. Also passes SSR for gatsby build. Going to merge for todays release and if we make adjustments if needed.

wincent:
😢 I don't think Clay should be polyfilling anything — we shouldn't mutate global state, because we don't know if the environment we're going to be running in may provide or want to provide a polyfill of its own, and which order things will load in. In general, I think libraries are more broadly useful if they ask users to follow a BYOP (Bring Your Own Polyfill) pattern.

Instead of a polyfill:

```js
if (typeof window !== 'undefined' && !Element.prototype.matches) {
	Element.prototype.matches =
		// @ts-ignore
		Element.prototype.msMatchesSelector ||
		Element.prototype.webkitMatchesSelector;
}
```

[A "ponyfill"](https://ponyfoo.com/articles/polyfills-or-ponyfills) would be better:

```js
function matches(element, selectorString) {
  if (element.matches) {
    return element.matches(selectorString);
  } else if (element.msMatchesSelector) {
    return element.msMatchesSelector(selectorString);
  } else if (element.webkitMatchesSelector) {
    return element.webkitMatchesSelector(selectorString);
  } else {
    return false;
  }
}
```

There is exactly one place where we use `matches`, so you could just define it there — for me, needing it in only one place is even more reason to avoid mutating global state.

bryceosterhaus:
@wincent yeah you're totally right. I was thinking about it more last night after releasing and felt uneasy about it. I was concerned about mutating global state as well, but I figured that if we are already checking if it exists first, we should be relatively safe. But the caveat is that we can't dictate when this part of code is run, maybe someone mutates that state later. 

I like your solution though, I can add that and release again today to help mitigate any problems from my initial change.

PR #3012, opened 2020-03-13
fix(@clayui/tooltip): remove polyfill in favor of ponyfill(ponyfill.com)

bryceosterhaus:
Update in regards to https://github.com/liferay/clay/pull/3007#issuecomment-598643306

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -76,13 +76,21 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
 	}
 };
 
-// Polyfill to account for any browsers that might not support `element.matches`.
-// Need to make sure window exists first since this can be run server side.
-if (typeof window !== 'undefined' && !Element.prototype.matches) {
-	Element.prototype.matches =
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
tested in IE, safari, chrome, and firefox. All seem to work fine

bryceosterhaus:
@wincent I really loved that whole repo, so great

wincent:
One thing that I forgot to make explicit was that if this were ever in a hot code path we might want to cache the result of the whole `if`/`else if`/`else if`/`else` check, but in practice:

- It is mostly going to take the first `if` branch.
- Even if it doesn't, the property checks are going to be ridiculously fast.
- Most DOM trees aren't *that* deep that we'd end up calling this more than a few times to while looking for an ancestor.
- Premature optimization is the root of all evil and all that; it is exceedingly unlikely that this would ever turn up in a profiler as a hot spot.

At least, that's what I was thinking. Maybe our future selves will come back here and tell us I was wrong. 😂

bryceosterhaus:
@wincent good point. I don't think its super necessary at this point, but as we continue to _pony up_I think its good to remember. I'll keep it in mind.

PR #3018, opened 2020-03-16
set dom-align prop for rtl languages

bryceosterhaus:
So I still am not entirely sure why dom-align breaks when portal is set to right-to-left. I did discover that if we use the css transform flag on dom-align that is works fine. However, I don't feel great about just throwing this boolean on there and being okay with. Has anyone seen anything similar in portal? See the gif below for example. The top button is using Clay's drop-down without css transform, and the below example is using dom-align with that flag enabled.

Other options I also thought of were

1. Allowing users to pass args to domAlign from the component props.
2. Adding a very specific hack for portal where we only add this flag if dir="rtl" id enabled.

Any other thoughts on this? Really scratching my head as to what portal is adding or doing that ends up breaking dom-align.

Kapture 2020-03-16 at 16 25 29

### DIFF
--- a/packages/clay-drop-down/src/Menu.tsx
+++ b/packages/clay-drop-down/src/Menu.tsx
@@ -191,6 +191,8 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>((
 						adjustY: autoBestAlign,
 					},
 					points,
+					// @ts-ignore
+					useCssTransform: !!window.TransitionEvent,
 				}
 			);
 		}


pat270:
@bryceosterhaus Is `dom-align` a third-party plugin? If so, my guess is that they tested rtl for CSS `transform`, but not `position {top|right|bottom|left}` or `margin-{top|right|bottom|left}` or whatever they use.

bryceosterhaus:
@pat270 it works fine in a sandbox environment when I set the direction to `rtl`. Also I don't think `dom-align` would really even care about accounting for rtl since the user is the one picking the alignment. I also looked through their code and they don't seem to do any transforms.

PR #3022, opened 2020-03-17
fix(@clayui/css): override specific styles for rtl support

### LINKED ISSUE #2971: Date picker calendar position fixed on the right end of the page when using dir="rtl" in the liferay-portal

A: Unable to reproduce this problem outside the liferay-portal.

1 - Install @clayui/datepicker@3.0.7; 
2 - Import "ClayDatePicker" on my component file;
3 - Load the liferay-portal configured with a language that uses dir="rtl" (for example: "ar").
4 - Click on the calendar icon to open the date picker;
5 - Date picker calendar is displayed at the right end of the page;. [(View image)](https://user-images.githubusercontent.com/28923749/75676671-7eaf6700-5c68-11ea-950f-5917ffc25256.png)

- Date picker calendar should be displayed relative to the date picker field;

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  (@clayui/date-picker) |  v3.0.7  |

bryceosterhaus:
Fixes #2971

So it turns out this issue was because `left: 0;` on `.dropdown-menu` was getting flipped to `right: 0;` when using rtl in portal. This is an issue because dom-align is setting the `left` property and that is ignored if the `right` property exists.
rlt specification says
> If direction of the containing block is rtl, right wins and left is ignored.

It also looks like we encountered this in portal before and we added this file, https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-css/frontend-css-web/src/main/resources/META-INF/resources/portal/_dropdown.scss#L5. However that style was specifically scoped within `#clay_dropdown_portal` which is no longer used.

So ultimately the question here is, should we add this to `@clayui/css`(like ive done here), or do we just update liferay-portal.

I am inclined to change it here in clay since we are also the same ones giving them the dropdown component and dom-align dependency.

### DIFF
--- a/packages/clay-css/src/scss/components/_dropdowns.scss
+++ b/packages/clay-css/src/scss/components/_dropdowns.scss
@@ -93,6 +93,11 @@
 	}
 }
 
+html[dir="rtl"] .dropdown-menu {
+	left: auto;
+	right: 0;
+}
+
 .dropdown-menu {
 	border-style: $dropdown-border-style;
 	font-size: $dropdown-font-size;
--- a/clayui.com/content/docs/css/content/c-kbd.md
+++ b/clayui.com/content/docs/css/content/c-kbd.md
@@ -7,29 +7,29 @@ order: 4
 <div class="nav-toc">
 
 -   [Base](#base)
-    -    [Kbd](#c-kbd)
-    -    [Monospaced](#c-kbd-monospaced)
-    -    [Inline](#c-kbd-inline)
-    -    [Group](#c-kbd-group)
+    -   [Kbd](#c-kbd)
+    -   [Monospaced](#c-kbd-monospaced)
+    -   [Inline](#c-kbd-inline)
+    -   [Group](#c-kbd-group)
 -   [Color Variants](#color-variants)
-    -    [Light](#c-kbd-light)
-    -    [Dark](#c-kbd-dark)
-    -    [Group Light](#c-kbd-group-light)
-    -    [Group Dark](#c-kbd-group-dark)
+    -   [Light](#c-kbd-light)
+    -   [Dark](#c-kbd-dark)
+    -   [Group Light](#c-kbd-group-light)
+    -   [Group Dark](#c-kbd-group-dark)
 -   [Size Variants](#size-variants)
     -   [Small](#c-kbd-sm)
     -   [Large](#c-kbd-lg)
 -   [Unicode Character List](#unicode-character-list)
-    -    [Standard](#standard-0-255-(utf-8)-character-set)
-    -    [General Punctuation](#general-punctuation-(u+2000-to-u+206f))
-    -    [Superscripts and Subscripts](#superscripts-and-subscripts-(u+2070-to-u+209f))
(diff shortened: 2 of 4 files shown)


packages/clay-css/src/scss/components/_dropdowns.scss
```diff
@@ -93,6 +93,11 @@
 	}
 }
 
+html[dir="rtl"] .dropdown-menu {
+	left: auto;
+	right: 0;
+}
+
```
pat270: @bryceosterhaus I believe portal's rtl css will swap these styles to ` left: 0; right: auto;`. I'm not sure if the library we use checks to see if the selector is `rtl` specific.

jbalsas: Hey @pat270, I think that's exactly what @bryceosterhaus expects to override the current `right: 0` with a more specific `right: auto`.

This is pretty obscure, and I'm pretty ashamed of https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-css/frontend-css-web/src/main/resources/META-INF/resources/portal/_dropdown.scss#L5 myself. 

This seems to depend on a build process that happens in DXP outside of Clay... for a problem that might be self-inflicted by DXP... I would probably apply it in DXP first, but it's totally up to you.

pat270:
I looked into this more and is this issue only with Date Picker or does it happen with all `dropdown-menu`s? From the screenshot provided, it looks like `dropdown-menu` is working but I don't know if that is actually a `dropdown-menu`. 

I was wondering if we can scope this to Date Picker so I can fix the next month and previous month arrows pointing the wrong way.

isDropdownMenu

bryceosterhaus:
@pat270 it applied to anything that used `DropDown.Menu`, which included other components like  autocomplete and date-picker.

I am also fine just adding this to portal since like @jbalsas said, its self inflicted. @jbalsas  do you see any issue with me updating that scss file to remove the id scope so that it applies to all dropdowns? I can send to bchan once we decide if portal or clay is the best route.

@pat270 I think regardless if we add it in clay or portal we can send a fix for the arrows.

pat270:
ok sounds good to me. I'm thinking about adding an `rtl` option for some mixins. I can send those later.

jbalsas:
Hey @bryceosterhaus, wouldn't this cause issues in your sandbox environment when not flipped?

I'd expect that if you add `atlas.css` and have some markup like this:

```html
<div dir="rtl">
    <div class=".dropdown-menu">
        ...
    </div>
</div>
```

Then align should fail...

I'd just go to DXP and make the change you suggest, @bryceosterhaus. If you need some sanity checks, you could lean on @john-co before pushing it to Brian.

bryceosterhaus:
@jbalsas I don't think it should fail. Are you thinking outside of portal or within portal?

CSB: https://codesandbox.io/s/practical-napier-z239y
Full page: https://z239y.csb.app/

bryceosterhaus:
@pat270 lets actually revert this and then just add it to portal specifically since thats the one who is inverting the right/left.

bryceosterhaus:
https://github.com/liferay/clay/pull/3028

PR #3027, opened 2020-03-18
feat(@clayui/date-picker): use locale from moment

### LINKED ISSUE #2972: Implement internationalization with "moment.js" receiving only the locale as property in ClayDatePicker
## What is your proposal?
Implement internationalization with "moment.js" receiving only the locale as property in ClayDatePicker

As suggestion:

Component API: 

```javascript
<ClayDatePicker locale="en" ... />
```

By default ClayDatePicker get dateFormat, firstDayOfWeek, months, timeFormat and weekdaysShort:

```javascript
import moment from "moment/min/moment-with-locales"

const localeData = moment.localeData(locale);

//when not providing a dateFormat get a "L" format by locale
const dateFormat = localeData.longDateFormat("L") // MM/DD/YYYY
const firstDayOfWeek = localeData.firstDayOfWeek() // 0
const months = localeData.months() // ["January", "February", ...]
//when not providing a timeFormat get a "LT" format by locale
const timeFormat = localeData.longDateFormat("LT") // HH:mm a
const weekdaysShort = localeData.weekdaysShort() // ["Sun", "Mon", "Tue", ... ]

``` 

The solution prevents the use of a different version of the moment in user module and the need to retrieve several times: the list of months, days of the week and the starting day of the week for each location. Knowing that the moment library (a dependency on ClayDatePicker) implements a solution that provides this data and only requires locale as a parameter.

bryceosterhaus:
fixes #2972

Hey @matuzalemsteles, what do you think about this change? I now you did a lot of work on this, do you think its helpful to add our own locale functionality like this or do you think its best for users just to pass their own props for locale.

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -87,6 +87,11 @@ interface IProps {
 	spritemap?: string;
 
 	/**
+	 *
+	 */
+	timeFormat?: string;
+
+	/**
 	 * Flag to enable datetime selection.
 	 */
 	time?: boolean;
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
@matuzalemsteles I was actually thinking similarly to you so that gives me confidence. I think we could decouple from moment but also provide from OOTB helpers for if they want to use moment. I'll look further into this and see about how we can best go about it. Thanks for the feedback!

matuzalemsteles:
Great!! 🚀

PR #3033, opened 2020-03-19
chore(@clayui/form): capitalize the 'B' for consistency

bryceosterhaus:
Hey @kresimir-coko, I made some changes on top of your PR. Mostly refactored some functions and also removed some types since TS can infer most of them.

@pat270, could you check out the markup for this component and verify the css? If we need to change markup or if you need to add some styles, go ahead and send me a PR with the changes and I will add them on this PR

Only thing left to do after that is the docs for the clayui site. @kresimir-coko could you add this component on the docs for clayui.com? I'll ping you on slack some specifics


bryceosterhaus:
@kresimir-coko one more thing to note. For the commit messages can you add the component the work is in? It should be "@clayui/form" since its within the form package. I already update your commit in this PR, but just wanted to let you know in case you didnt notice it. I think before you had @clay-components

bryceosterhaus:
Looks like we will have to break it off into its own package to avoid circular dependencies.

`@clayui/drop-down -> @clayui/form -> @clayui/drop-down`

Anyone have any better ideas than just `@clayui/localized-input`?

The other option would be to move it to dropdown(seems a little odd to me).

bryceosterhaus:
Never realized how annoying git it is when renaming files based on case sensitivity...
 TIL: `git mv`

packages/clay-css/src/scss/_components.scss
```diff
@@ -33,6 +33,8 @@
 @import "components/_form-validation";
 @import "components/_icons";
 @import "components/_input-groups";
+@import "components/_input-move-boxes";
+@import "components/_select-box";
```
pat270: Don't need `@import "components/_input-move-boxes";` and `@import "components/_select-box";`

packages/clay-css/src/scss/components/_input-move-boxes.scss
```diff
@@ -0,0 +1,23 @@
+.input-move-boxes {
```
pat270: Don't need this file.

packages/clay-css/src/scss/components/_select-box.scss
```diff
@@ -0,0 +1,6 @@
+.select-box-order-buttons {
```
pat270: Don't need this file.

packages/clay-form/src/DualListBox.tsx
```diff
@@ -0,0 +1,182 @@
...
+						onItemsChange([newLeftItems, rightItems])
+					}
+					onSelectChange={handleLeftSelectedChange}
+					showArrows
+					size={size}
+					spritemap={spritemap}
+					value={leftSelected}
+				/>
+
+				<ClayButton.Group className="clay-dual-listbox-actions clay-dual-listbox-item">
```
pat270: Can we add at `type="vertical"` or `type="stacked` and have the container be `<div class="btn-group-vertical">` instead of `<div class="btn-group btn-group-vertical">`? If not I can add a css overwrite for this one.

packages/clay-form/src/DualListBox.tsx
```diff
@@ -0,0 +1,182 @@
...
+					<ClayButtonWithIcon
+						className="transfer-button-ltr"
```
pat270: needs `btn-sm`

packages/clay-form/src/DualListBox.tsx
```diff
@@ -0,0 +1,182 @@
...
+							);
+
+							onItemsChange([arrayLeft, arrayRight]);
+						}}
+						spritemap={spritemap}
+						symbol="caret-right"
+					/>
+
+					<ClayButtonWithIcon
+						className="transfer-button-rtl"
```
pat270: needs `btn-sm`

packages/clay-form/src/SelectBox.tsx
```diff
@@ -0,0 +1,215 @@
...
+		Array.isArray(value) ? value : [value]
+	);
+
+	return (
+		<div className={classNames(className, 'form-group')}>
+			{label && <label className="reorder-label">{label}</label>}
+
+			<div
+				className={classNames('clay-reorder', {
+					[`clay-reorder-footer-${buttonAlignment}`]: buttonAlignment,
```
pat270: Not sure what class this outputs I broke the storybook. It should be `<div class="clay-reorder clay-reorder-footer-end">`. The other option is to add `clay-reorder-footer-end` to `clay-reorder-footer` below.

packages/clay-form/src/SelectBox.tsx
```diff
@@ -0,0 +1,215 @@
...
+					))}
+				</select>
+
+				<div className="clay-reorder-underlay form-control" />
+
+				{showArrows && onItemsChange && (
+					<div className="clay-reorder-footer">
+						<ClayButton.Group className="reorder-order-buttons">
+							<ClayButtonWithIcon
+								className="reorder-button reorder-button-up"
```
pat270: needs `btn-sm`

packages/clay-form/src/SelectBox.tsx
```diff
@@ -0,0 +1,215 @@
...
+									onItemsChange(
+										reorderUp(items, selectedIndexes)
+									)
+								}
+								spritemap={spritemap}
+								symbol="caret-top"
+							/>
+
+							<ClayButtonWithIcon
+								className="reorder-button reorder-button-down"
```
pat270: needs `btn-sm`

pat270:
One thing I forgot is arrow keys, when you reach the bottom of `select[multiple]` (using down arrow on keyboard) causes the main page to scroll the same way as it did with Date Picker

packages/clay-form/src/DualListBox.tsx
jbalsas: Hey @pat270, not exactly sure what you mean by that suggestion, but it stroke me as odd... I'm not sure `type` is a valid `html` attribute for a `div` tag. If we need some kind of specifier, we can always use `data-type`, though.

kresimir-coko:
Hey @jbalsas I can't know for sure with the wording @bryceosterhaus used, am I supposed to review this and make changes that @pat270 suggested, or is he going to do that?

jbalsas:
I think you can go for it @kresimir-coko

packages/clay-form/src/DualListBox.tsx
kresimir-coko: `type` doesn't work on `ClayButton.Group` or `div`. Replacing it with `<div class="btn-group-vertical">` makes them squished.

kresimir-coko:
Updated PR [#3035](https://github.com/liferay/clay/pull/3035)

packages/clay-form/src/DualListBox.tsx
pat270: @jbalsas doesn't need to be type just something to make it vertical. I guess you could even get rid of `ClayButton.Group` here and just use a `<div  className="btn-group-vertical clay-dual-listbox-actions clay-dual-listbox-item">`.

kresimir-coko: @pat270 They are vertical now as it stands.

pat270: @kresimir-coko `btn-group-vertical` and `btn-sm` on `ClayButtonWithIcon` will make them the correct size and style.  They are vertical now as it stands is no good because the styles are off.

Regular sized `btn-monospaced` gets squished because Bootstraps `.btn-group-vertical > .btn` styles are overwriting `btn-monospaced`. I need to fix this but not in this PR.

kresimir-coko: Alright @pat270 , I'll add the proper container in the updated PR :)

PR #3035, opened 2020-03-20
chore(@clayui/form): Make buttons small, remove scroll on keyboard up/down

kresimir-coko:
Hey @bryceosterhaus I've made the changes suggested by @pat270 in [#3033](https://github.com/liferay/clay/pull/3033) except for the comment on [container classes](https://github.com/liferay/clay/pull/3033#discussion_r395366295)

@jbalsas should the clayui.com documentation be it's own separate issue, or at least it's own PR?


jbalsas:
Whatever you guys see fit, @markocikos. Just make sure however you're doing, you're helping @esanzp track down the progress.

packages/clay-form/src/SelectBox.tsx
```diff
@@ -7,11 +7,13 @@ import ClayButton, {ClayButtonWithIcon} from '@clayui/button';
 import classNames from 'classnames';
 import React from 'react';
 
+export type TItem = {
```
kresimir-coko: Interesting pattern, eliminates duplicate code, I like it. Why the name `TItem`, should I use this name in the future if I have a similar pattern?

packages/clay-form/src/SelectBox.tsx
```diff
@@ -129,6 +111,18 @@ interface IProps extends React.SelectHTMLAttributes<HTMLSelectElement> {
 	spritemap?: string;
 }
 
+export const getSelectedIndexes = (
```
kresimir-coko: Why wasn't this a `function` like `reorderDown` but a `const` instead? Just curious so I know for the future when writing these functions, also interesting to see that we do these exported functions from child to parent if we want

bryceosterhaus: This is still a function, but just using an arrow function rather than a regular `function()`. I typically I use opt to using arrow functions since they are often less lines of code and allow for implicit returns. 

I think we can typically follow these guidelines
- Use `function()` when you need to access `this` and for Object.prototype properties.
- Use => everywhere else.

packages/clay-form/src/SelectBox.tsx
bryceosterhaus: We have an eslint rule that enforces all `type` names begin with "T" and all `interface` names begin with "I". 

I would say follow this pattern when we have some sort of re-used type. Since this shape of item is used in quite a few places, it made sense to make a single type for it.

packages/clay-form/src/SelectBox.tsx
wincent: Another use case for `function()` is when you need to rely on hoisting eg.

```js
// I can use `fn()` before its definition, because it gets hoisted...
console.log(fn()); // "fn"

function fn() {
  return 'fn';
}

// Can't do that with arrow functions...
console.log(arrow()); // ReferenceError: arrow is not defined

const arrow = () => 'arrow';
```

PR #3043, opened 2020-03-20
feat(@clayui/date-picker): add helper for easily getting locale info from moment

### LINKED ISSUE #2972: Implement internationalization with "moment.js" receiving only the locale as property in ClayDatePicker

bryceosterhaus:
fixes #2972

### DIFF
--- a/packages/clay-date-picker/src/Helpers.ts
+++ b/packages/clay-date-picker/src/Helpers.ts
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import moment from 'moment';
+
 export interface IDay {
 	date: Date;
 	outside?: boolean;
@@ -99,4 +101,19 @@ export function range({end, start}: {end: number; start: number}) {
 		},
 		(v, k) => k + start
 	);
+}
+
+/**
+ * Helper function for getting certain props from `moment`.
+ * This allows users to not have to import and use `moment` themselves.
+ */
+export function getLocaleProps(locale: string) {
+	const localeData = moment.localeData(locale);
+
+	return {
+		dateFormat: localeData.longDateFormat('L'),
+		firstDayOfWeek: localeData.firstDayOfWeek(),
+		months: localeData.months(),
+		weekdaysShort: localeData.weekdaysShort(),
+	};
 }
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -13,6 +13,7 @@ import React from 'react';
 import DateNavigation from './DateNavigation';
 import DayNumber from './DayNumber';
 import DaysTable from './DaysTable';
+import {getLocaleProps} from './Helpers';
 import {useCurrentTime, useWeeks} from './Hooks';
 import InputDate from './InputDate';
 import TimePicker from './TimePicker';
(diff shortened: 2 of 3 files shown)


packages/clay-date-picker/src/index.tsx
```diff
@@ -164,6 +171,7 @@ const ClayDatePicker: React.FunctionComponent<IProps> = ({
 		start: DateNow.getFullYear(),
 	},
 }: IProps) => {
+	moment.locale('fr');
```
wincent: What's this?

packages/clay-date-picker/src/Helpers.ts
```diff
@@ -100,3 +102,19 @@ export function range({end, start}: {end: number; start: number}) {
 		(v, k) => k + start
 	);
 }
+
+/**
+ * Helper function for getting certain props from `moment`.
+ * This allows users to not have to import and use `moment` themselves.
+ */
+export function getLocaleProps(locale: string) {
```
matuzalemsteles: Just a thought, instead of using the `moment` instance directly, we can ask the user to pass that instance along with the `locale`, so we can have a better chance of removing the `moment` internally. What do you think?

packages/clay-date-picker/src/index.tsx
bryceosterhaus: this is.... very unintentional

packages/clay-date-picker/src/Helpers.ts
bryceosterhaus: I think removing moment internally is probably better suited for a major release, so instead of going that route I thought it might be more helpful to just add some utilities that make it a bit easier for users so that they don't have to use moment themselves.

matuzalemsteles: Hmm yeah, makes sense.

packages/clay-date-picker/src/index.tsx
```diff
@@ -86,6 +87,11 @@ interface IProps {
 	 */
 	spritemap?: string;
 
+	/**
+	 *
+	 */
+	timeFormat?: string;
```
matuzalemsteles: Just one caveat about adding this API: I had removed this implementation because our TimePicker only behaves with an `H:m` format, if the user escapes from that format it will not make any difference in the visualization in TimePicker maybe just in the DatePicker input here, but still in the end we’ll always be adding a `H:m` format. See `useCurrentTime`

matuzalemsteles: I don't know how it works with different formats...

bryceosterhaus: Oh good catch, I will just remove this API since we don't necessarily need it right now.

bryceosterhaus:
@matuzalemsteles updated and removed timeFormat API

bryceosterhaus:
Resent due to CI being finicky. https://github.com/liferay/clay/pull/3057

PR #3044, opened 2020-03-23
chore(@clayui/form): Initial documentation setup for DualListBox and SelectBox

kresimir-coko:
Hey @bryceosterhaus I've made a first version of the documentation to see if the file structure is correct and if this is the type of documentation we wanted in the first place.

If you're willing let me know give me some guidance regarding how to go forward, it seems kinda overwhelming with writing down paragraphs and deciding which examples to showcase.

Also the selecting functionality doesn't work in the mdx files, idk what I'm missing there to get it working.

### DIFF
--- a/clayui.com/content/docs/components/forms/css-dual-list-box.md
+++ b/clayui.com/content/docs/components/forms/css-dual-list-box.md
@@ -1,0 +1,90 @@
+---
+title: 'Dual List Box'
+description: 'This section demonstrates the different text input types, including usage and validation states.'
+---
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
+-   [Basic Usage](#basic-usage)
+
+</div>
+</div>
+
+<div class="clay-site-alert alert alert-warning">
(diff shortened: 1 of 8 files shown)


bryceosterhaus:
Hey @kresimir-coko this is great! We can start with this basic info and add more on demos and paragraphs as needed. I also noticed that the selecting of items didnt work in the code examples, I tried to debug it for a bit but couldn't seem to pin point it. You might try investigating `react-live` which is our how we render those inline react examples. I would even try without ClaySelectBox and see if you are able to reproduce at all. Very odd that doesnt work though.

kresimir-coko:
Vanilla `select` works fine, I couldn't figure out why the selecting doesn't work. react-live documentation is lacking in that regard.

I've pushed a commit containing some cleanup and additions of documentation that were missing, like the html example code and tidied up headings and descriptions.

bryceosterhaus:
I opened a new issue on react-live at https://github.com/FormidableLabs/react-live/issues/196.

Hopefully that will provide some direction

bryceosterhaus:
Updated and added the note about the select functionality not working at https://github.com/liferay/clay/pull/3076

PR #3046, opened 2020-03-23
fix(@clayui/date-picker): adds the API to disable date picker features

### LINKED ISSUE #3045: DatePicker: Make it possible to disable the DatePicker and listen to onFocus and onBlur events
I'm migrating from DDM to use the new DatePicker in Clay's React, but we need some features like disabling DatePicker and listening for input events.

matuzalemsteles:
Fixes #3045

### DIFF
--- a/packages/clay-date-picker/src/InputDate.tsx
+++ b/packages/clay-date-picker/src/InputDate.tsx
@@ -7,7 +7,7 @@ import {ClayInput} from '@clayui/form';
 import moment from 'moment';
 import React from 'react';
 
-interface IProps {
+interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	ariaLabel?: string;
 	currentTime: string;
 	dateFormat: string;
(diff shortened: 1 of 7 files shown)


packages/clay-date-picker/src/InputDate.tsx
```diff
@@ -55,7 +58,9 @@ const ClayDatePickerInputDate: React.FunctionComponent<IProps> = ({
 		<>
 			<input name={inputName} type="hidden" value={memoizedValue} />
 			<ClayInput
+				{...otherProps}
 				aria-label={ariaLabel}
+				disabled={disabled}
```
bryceosterhaus: Since we are already using `{...otherProps}`, we don't need to explicitly set `disabled` or use it off of the props anymore

bryceosterhaus: We can also probably remove other explicit props like `placeholder` and `onChange`. Unless they are used elsewhere.

matuzalemsteles: Oh yeah, I'll clean it up.

matuzalemsteles:
@bryceosterhaus updated.

Issue #3029, opened 2020-03-19
@clayui/css Date Picker RTL icons are backwards


pat270:
Related to #2971, I noticed next month and previous month button icons are pointing the wrong direction.

<img width="276" alt="date-picker-icons" src="https://user-images.githubusercontent.com/788266/77021508-1b018b00-6944-11ea-980e-c0cb08e97e9d.png">

jbalsas:
I'm closing this so we can discuss a more generic approach as part of [RTL Strategy](https://github.com/liferay/clay/issues/3073)

PR #3062, opened 2020-03-24
fix(clay-css): 2.x Date Picker icons should flip when dir="rtl"

pat270:
issue #3029

### DIFF
--- a/packages/clay-css/src/scss/_mixins.scss
+++ b/packages/clay-css/src/scss/_mixins.scss
@@ -1,5 +1,7 @@
 @import "mixins/_vendor-prefixes";
 
+@import "mixins/_globals";
+
 @import "mixins/_aspect-ratio";
 @import "mixins/_background";
 @import "mixins/_badges";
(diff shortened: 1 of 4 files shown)


jbalsas:
Hey @pat270, I'm not convinced about this approach... and I'm actually scared of its unintended consequences.

Specifically, in the icon case, this is at best a partial fix that would only address icons in DatePicker, but likely not elsewhere directional icons are used. 

While this might work in an isolated environment (at the cost of yet again increasing our bundle size...), I have doubts it won't cause any other issues downstream in DXP. In there, we rely on the mirroring process to serve a mirrored css file. We used to also map directional icons to flip them (that is, `chevron-left` and `chevron-right` are swapped and when you ask for one you'd get the other). I think this is probably what's missing right now and what would solve most of the issues at once.

pat270:
@jbalsas what you said makes sense. If we want to rely on mirroring in Portal we can. I just saw the issue in the screenshot and decided to fix it here. I can close these, just lmk.

jbalsas:
Closing as per the comments above. I think we can discuss a general approach in [Define RTL Strategy](https://github.com/liferay/clay/issues/3073)

PR #3065, opened 2020-03-25
feat(@clayui/label): Add support for truncating the Label

kresimir-coko:
This PR fixes [#2977](https://github.com/liferay/clay/issues/2977).

I'm unsure if any tests or stories need to be added for this new prop, if they do need to be added let me know.


bryceosterhaus:
Hey @kresimir-coko this is added in this PR, https://github.com/liferay/clay/pull/3060. Rather than adding a specific prop for truncating, I added the ability to add attrs to the inner element. This also gives more flexibility for users to add accessibility attributes to the href.

Issue #2939, opened 2020-02-18
Clay label doesn't support icons inside


4lejandrito:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## Is there an example you can provide via codesandbox.com?

https://codesandbox.io/s/nostalgic-diffie-yck1g

## What are the steps to reproduce?

Create a ClayLabel with a ClayIcon + text inside.

## What is the expected result?

The icon is on the left of the text.

We need this to support the labels in [this Figma design](https://www.figma.com/file/6A17MZaJL3T3JItVDlfKIO/lps-104256-regular-roles-add-possibility-to-select-repositories-as-scope.).

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3  |
| React |    16     |

bryceosterhaus:
@4lejandrito we can get this added for the next release. How urgent of a need is this? Our next scheduled release is Feb 27th.

4lejandrito:
Hi @bryceosterhaus,

It is not urgent. 

Thanks!

kresimir-coko:
@4lejandrito Hey, I've got a question about this issue. Currently the way ClayLabel works is that it allows the user to define the `children` of the component like. It also accepts a `className` prop that will allow the user to style the component however they want.
```
<ClayLabel className="custom-class">
    <ClayIcon spritemap="/icons.svg" symbol="repository" /> 
    
    Text...
</ClayLabel>
```
Why not leave the component as is and leave the responsibility on the user to add the icon and style it the way they want? 

Otherwise can you specify what the expected behaviour of the `ClayLabelWithIcon` should be, eg. will the icon only be on the left side of the ClayLabel?

4lejandrito:
Hi @kresimir-coko,

I agree with you that it is possible to achieve now. But I also think it is not expected to put an icon and a text inside and have them behave like "block" elements. 

I think it would be much more natural to have them one next to the other. Same with any other kind of elements you put inside. 

But it is up to you guys.

Thanks!

kresimir-coko:
Alright, that helps clarify it a bit, thanks @4lejandrito

PR #3066, opened 2020-03-25
feat(@clayui/label): Add contentBefore prop to ClayLabel

kresimir-coko:
Hey @pat270 I think this one's for you. It's a simple change as requested [here](https://github.com/liferay/clay/issues/2939).

Not specific to the task I'm wondering, when we take on these CSS tasks how do we know what CSS rules to apply, mixins to use, etc.? Do you have a list of rules/classes to use when you want to achieve some things, like is there a class for `display: inline-flex` for example?

### DIFF
--- a/packages/clay-label/src/index.tsx
+++ b/packages/clay-label/src/index.tsx
@@ -8,6 +8,25 @@ import ClayLink from '@clayui/link';
 import classNames from 'classnames';
 import React from 'react';
 
+const ClayLabelItemBefore: React.FunctionComponent<
+	React.HTMLAttributes<HTMLDivElement>
+> = ({children}) => (
+	<>
+		<div className="label-item label-item-before">
+			{React.Children.map(children, (child, i) =>
+				React.cloneElement(
+					<span className="label-item label-item-before">
+						{child}
+					</span>,
+					{
+						key: i,
+					}
+				)
+			)}
+		</div>
+	</>
+);
+
 type DisplayType =
 	| 'secondary'
 	| 'info'
@@ -24,6 +43,11 @@ interface IProps
 	closeButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement> & {
 		ref?: (instance: HTMLButtonElement | null) => void;
 	};
+
+	/**
+	 * Any content that you would like to have render inside the label before the children content
+	 */
+	contentBefore?: React.ReactFragment | typeof ClayLabelItemBefore;
 
 	/**
 	 * Determines the style of the label.
@@ -54,6 +78,7 @@ const ClayLabel = React.forwardRef<HTMLAnchorElement | HTMLSpanElement, IProps>(
 			children,
 			className,
 			closeButtonProps,
+			contentBefore,
 			displayType = 'secondary',
 			href,
 			innerElementProps = {},
@@ -75,6 +100,10 @@ const ClayLabel = React.forwardRef<HTMLAnchorElement | HTMLSpanElement, IProps>(
 				})}
 				ref={ref}
 			>
+				{contentBefore && (
+					<ClayLabelItemBefore>{contentBefore}</ClayLabelItemBefore>
+				)}
+
 				<TagName
 					{...innerElementProps}
 					className={classNames(
--- a/packages/clay-label/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-label/src/__tests__/__snapshots__/index.tsx.snap
@@ -25,7 +25,7 @@ exports[`Rendering default 1`] = `
 </span>
 `;
 
-exports[`Rendering renders as a link and closable  1`] = `
+exports[`Rendering renders as a link and closable 1`] = `
 <span
   className="label label-dismissible label-secondary"
 >
@@ -82,6 +82,23 @@ exports[`Rendering renders as closable 1`] = `
         />
       </svg>
     </button>
+  </span>
+</span>
+`;
+
+exports[`Rendering renders with an icon before content 1`] = `
+<span
+  className="label label-secondary"
+>
+  <span
+    className="label-item label-item-before"
+  >
+    Content before
+  </span>
+  <span
+    className="label-item label-item-expand"
+  >
+    Label
   </span>
 </span>
 `;
--- a/packages/clay-label/src/__tests__/index.tsx
+++ b/packages/clay-label/src/__tests__/index.tsx
@@ -50,7 +50,7 @@ describe('Rendering', () => {
 		expect(testRenderer.toJSON()).toMatchSnapshot();
 	});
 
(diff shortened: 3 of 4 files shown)


pat270:
@kresimir-coko I think @4lejandrito is looking to add
```
<span className="label-item label-item-before">
	<Icon spritemap={spritemap} symbol="{the_icon_symbol}" />
</span>
```
to https://github.com/liferay/clay/blob/master/packages/clay-label/src/index.tsx#L70

> Do you have a list of rules/classes to use when you want to achieve some things, like is there a class for display: inline-flex for example?

The functional CSS classes we have are provided by Bootstrap. You can see them at https://getbootstrap.com/docs/4.4/utilities/borders/.

bryceosterhaus:
@kresimir-coko I think we could address this by adding a `contentBefore` prop that we would render as the first child in the container `<span>` and before the `<TagName` element. This would allow the user to add anything they want before the main content of the label.

kresimir-coko:
Hey @bryceosterhaus I've updated the PR with the markup suggested by @pat270 and prop suggested by you. I've also written a barebones story for it and a simple rendering test.

packages/clay-label/src/index.tsx
```diff
@@ -67,6 +73,12 @@ const ClayLabel = React.forwardRef<HTMLAnchorElement | HTMLSpanElement, IProps>(
 				})}
 				ref={ref}
 			>
+				{contentBefore && (
+					<span className="label-item label-item-before">
+						{contentBefore}
+					</span>
+				)}
```
bryceosterhaus: I think we could get creative with this and add the ability to add more than one element with `label-item-before`. First we will need to extract this into another component, `<ClayLabel.ItemBefore>` and that can just be in this same file. Then we probably want to have some logic to handle multiple items so that both the examples work below when passed to `contentBefore`

```jsx
// Handles all non React.Fragment items by wrapping them with `<ClayLabel.ItemBefore />` internally
<ClayIcon />


// If user passes in a React.Fragment we expect every child to be <ClayLabel.ItemBefore />
<>
	<ClayLabel.ItemBefore />
	<ClayLabel.ItemBefore />
</>
```

I think you would set the type for contentBefore to be 
```ts
contentBefore?: React.ReactFragment | typeof ItemBefore;
```

And I think  the rendering logic would be something like

```jsx
{
	contentBefore && contentBefore.type === React.Fragment ? (
		contentBefore
	) : (
		<ClayLabel.ItemBefore>{children}</ClayLabel.ItemBefore>
	);
}
```

kresimir-coko:
Hey @bryceosterhaus I haven't managed to figure this out so it properly loops over `children`. 

Couple questions:
1. How to get the component name with a dot inside like `<ClayLabel.ItemBefore />` instead of `<ClayLabelItemBefore />`?
2. How did you think of defining `contentBefore.type `, if it's supposed to pick up the type automatically it didn't

matuzalemsteles:
hey @bryceosterhaus and @kresimir-coko, just a concern about this API: It is a little strange that we only have a `contentBefore` API, I think we should support `contentAfter` also because I would be wondering when I see this API, why can't we add an API after?

ClayLabel is becoming a bit of a weird component, for two reasons:

- We have a `closeButtonProps` on one side for an extremely exclusive case given the name of its API.
- Adding `contentBefore` differs from its essence, with just `closeButtonProps` it seems an exclusive component for a case, with `contentBefore` it seems confusing because it is giving flexibility and at the same time it has an API that limits on the other end. I would say that APIs do not communicate and do not have amornization.

What I am saying is that we should support these APIs but in another way, I would say that we made a little mistake at the beginning of this component in adding the `closeButtonProps` prop that has a feeling of limitation, maybe something that we can address in a major version of this component and correct design problems.

A thought on how we could go with this component in the future:

- ClayLabel become a more low-level component
- ClayLabel high-level for `closeButtonProps`

Just a sketch of design that I have in mind. Redesign of `<ClayLabel />`, in essence the API's would still be the same but without support for `href` directly and `closeButtonProps`, internally it would appear as something like this:
```jsx
const ClayLabel = ({...}) => (
	<span
		{...otherProps}
		className={classNames('label', className, {
			'label-dismissible': closeButtonProps,
			'label-lg': large,
			[`label-${displayType}`]: displayType,
		})}
		ref={ref}
	>
		{contentBefore && (...)}
		{React.cloneElement(children, {
			// We probably have to check if it exists
			...children.props
			className={classNames(
				'label-item label-item-expand',
				children.props.className
			)}
		})}
		{contentAfter && (...)}
	</span>
);

<ClayLabel>
	<ClayLink href="#">{'Label with Link'}</ClayLink>
</ClayLabel>

<ClayLabel
	contentAfter={
		<ClayButton>
			<ClayIcon symbol="times" />
		</ClayButton>
	}
	contentBefore={<ClayIcon symbol="angle-right" />}
>
	<span>{'Label with Link'}</span>
</ClayLabel>
```

From this we can extract high-level components, such as `ClayLabelWithLink`, `ClayLabelWithClose` for example. We can explore other designs to support `contentAfter` and `contentBefore`, for example, the first 3 elements of the children are the APIs in their order.

As a last word, if this is not urgent for the time being, we can keep the Label API as it is and make these component design changes later in a v4 version.

bryceosterhaus:
@matuzalemsteles yeah I found myself thinking something similar but wasn't quite positive, but I totally agree with the way you put it. I went ahead and refactored it a bit to preserve the current API but also adds the additional low-level API and then for the next breaking change we can condense it slightly. What do you think about my changes? https://github.com/liferay/clay/pull/3066

bryceosterhaus:
@kresimir-coko 

> 1. How to get the component name with a dot inside like <ClayLabel.ItemBefore /> instead of <ClayLabelItemBefore />?

For #1, internally we actually just use the component name that we create, but we export it on the same object as the default export, this allows it to be called like...

```jsx
import ClayLabel from '@clayui/label';

<ClayLabel.ItemBefore />
```

Check out https://github.com/liferay/clay/pull/3081/commits/0b79df3334ebfcbde90d7ef850b9e55f1c8f8e34#diff-6ba2b1599ae5d6eb64277e748ccaf8beR147-R152 to see how it was implemented.


> 2. How did you think of defining contentBefore.type , if it's supposed to pick up the type automatically it didn't

I don't think it actually makes sense to go this route anymore after hearing Matu's feedback. But for future reference, we actually wouldn't need to define a type for that since `type` exists on the `React.ReactElement` type and we could have just defined `contentBefore?: React.ReactElement`. However, since we aren't going this route, no need to worry about it lol

Let me know your thoughts on my updated changes!

wincent:
> What do you think about my changes? #3066

I think you actually meant: https://github.com/liferay/clay/pull/3081

bryceosterhaus:
merged at https://github.com/liferay/clay/pull/3034

PR #3069, opened 2020-03-25
fix(@clayui/css): Use prettier to format Clay CSS

### LINKED ISSUE #2037: Use Prettier to format SCSS
We're using Prettier in liferay-portal now to format JS and SCSS files. Because the clay-css SCSS files end up getting copied around inside portal (eg. in themes; random example: apps/frontend-theme/frontend-theme-unstyled/src/main/resources/META-INF/resources/_unstyled/css/clay/mixins/_position.scss ) it would be nice if they were formatted the same way.

For now, I am going to add an exclusion so that Prettier in liferay-portal doesn't try to check or format those Clay files, but it would be nice if we got Prettier to format SCSS in this repo too (currently it is only used to format JS, AFAIK). This might be a change that we want to make on both the "2.x-develop" branch and "develop".

@pat270, @carloslancha, @matuzalemsteles, @bryceosterhaus

pat270:
fixes #2037

@bryceosterhaus I updated `.prettierignore` to exclude files related to the test site, Bootstrap source, and `_global-functions.scss`.


.prettierignore
```diff
@@ -1,7 +1,15 @@
 *.d.ts
+.codesandbox
+.git
+.github
+.storybook
```
bryceosterhaus: How come all of these got added? I  think we should still include all of these  except for the .git folder.

.prettierignore
```diff
@@ -1,7 +1,15 @@
...
 bower_components
 build
-clay-css
+packages/clay-css/src/scss/bootstrap/**/**/**/*.scss
+packages/clay-css/**/**/**/**/**/*.js
+packages/clay-css/**/**/**/**/**/*.md
+packages/clay-css/**/**/**/**/**/*.json
```
bryceosterhaus: Also are we able to format the js md and json in clay? Or are those ignored for a specific reason?

.prettierignore
pat270: I didn't realize these were configurations we added. I assumed these were autogenerated similar to `.cache` on clayui.com

.prettierignore
pat270: The test site uses several third-party js plugins. I didn't want to ignore them separately since the test site isn't a critical part of the Clay repo. For example, we shouldn't be formatting `svg4everybody.js`, `clay-css/src/js/site/jquery.ba-outside-events.min.js`, `sassdoc-theme-clay-css/dist/index.js`, `sassdoc-theme-clay-css/dist/nunjucks.js`, etc.

If it's something you want I can add it.  I was thinking along the lines of why process files that don't really matter?

wincent: You don't need all those repeated `**` — one is equivalent to any number of intervening directories (full syntax [described here](https://git-scm.com/docs/gitignore#_pattern_format)).

For example:

- `packages/clay-css/**/*.js` is the same as `packages/clay-css/**/**/**/**/**/*.js`

If you really want to say "match *exactly* 5 levels down" you would use a single `*` at each level:

- `packages/clay-css/*/*/*/*/*/*.js`

If you want to ignore dist-like third-party stuff like the stuff you list you can either list it explicitly:

```
/packages/clay-css/dist
/packages/clay-css/src/js/site/jquery.*
# etc
```

or ignore everything then whitelist specific stuff:

```
/packages/clay-css/**/*.js
!packages/clay-css/specific/place/to/whitelist/**/*.js
```

pat270: Thanks @wincent, I'll make these changes and push it up.

pat270:
@bryceosterhaus I removed `.codesandbox`, `.github`, and `.storybook`. 
@wincent I couldn't get the whitelist pattern to work. Prettier isn't formatting the js files located in `clay-css/tasks/lib`. I tried `!/packages/clay-css/tasks/lib`, `!packages/clay-css/tasks/lib`, and several others.

bryceosterhaus:
@pat270 the js,json,md in clay-css should be fine, just wanted to verify.  And the reason its not formatting `tasks/lib` is probably because we have `lib` declared in the prettierignore already.

wincent:
> And the reason its not formatting `tasks/lib` is probably because we have `lib` declared in the prettierignore already.

That's right. Given patterns like these:

```
lib
packages/clay-css/src/js/popper.js
````

Any `lib` directory at any level will get ignored (which means all of its descendents get ignored too). Generally, it's good hygiene to be as specific as possible, so I would anchor a lot of these patterns to the root with a leading `/`; ie that second line above would be:

```
/packages/clay-css/src/js/popper.js
```

And if we want to ignore `lib` everywhere but unignore it in a specific place, that is totally doable:

```
lib
!/some/other/lib
```

pat270:
@bryceosterhaus @wincent here is the updated `.prettierignore`

pat270:
Fixed conflicts.

bryceosterhaus:
@pat270 I think we can go ahead and merge this whenever you feel comfortable with it since this is likely gonna keep conflicting with every new pr merged. If @wincent has anything to add we can add it in a new pr.

pat270:
@bryceosterhaus sounds good. Merging.

PR #3074, opened 2020-03-26
chore(@clayui/packages): Fix accessibility issues on Storybook

kresimir-coko:
Hey @bryceosterhaus I've updated the [accessibility doc](https://docs.google.com/spreadsheets/d/1rrzP_H9XBF9pW6omBLpI22NYLjB7JO-6ZvM2P260Z44/edit#gid=0) with more of my findings while removing these violations.

One of the commits contains simpler changes I could do in the component implementations, the other containing the changes in stories only.

There are some components that use `buttons` that would require us to add `aria-labels` to them either as I suggested in the commit or with an extensive change of adding a prop like it's done in some components like `DatePicker` I believe.

All in all, we have a lot more work to do if we want to make all our components follow the same pattern of receiving and consuming props specific to accessibility.


packages/clay-card/src/CardWithHorizontal.tsx
```diff
@@ -86,6 +86,7 @@ export const ClayCardWithHorizontal: React.FunctionComponent<IProps> = ({
 							spritemap={spritemap}
 							trigger={
 								<button
+									aria-label="Open Dropdown"
```
bryceosterhaus: Unfortunately we can't add these to the `src` since we need the ability to localize the labels. Check out the [colorpicker](https://github.com/liferay/clay/blob/master/packages/clay-color-picker/src/index.tsx#L61-L64) and see how we add configureable aria-labels. Any sort of english text added to our components src needs to have the ability to be configured via props.

kresimir-coko:
Hey @bryceosterhaus I've added `ariaLabels` props to the components as suggested, and a `tabIndex` for one of the components. Can you take a look and see if everything is in order, this should eliminate most of the accessibility issues

bryceosterhaus:
@kresimir-coko looks good! I made a few changes so that we get some more flexibility out of the API. Let me know if my changes make sense to you, I sent them at https://github.com/liferay/clay/pull/3095/commits/59019c72c030c15342b9ca3c7dbe3485e41440c4

PR #3081, opened 2020-03-27
feat(@clayui/label): use a hybrid API appraoch to expose low-level components

bryceosterhaus:
continued conversation from https://github.com/liferay/clay/pull/3066

### DIFF
--- a/clayui.com/content/docs/components/label.mdx
+++ b/clayui.com/content/docs/components/label.mdx
@@ -22,6 +22,12 @@ import {
 </div>
 </div>
 
+## Overview
+
+`ClayLabel` offers two different APIs for use by toggling the prop `withClose`. By default(`withClose={true}`), `ClayLabel` behaves like a high-level component. If you want to use the lower-level components and compose multiple parts to your label, all you need to do is set `withClose={false}`.
+
+Check out [this storybook example](storybook.clayui.com/?path=/story/components-claylabel--w-content-before) for a demo.
+
 ## Display Types
 
 Using [`displayType`](#api-displayType) property you can use these color variations on Label component:
--- a/packages/clay-label/src/__tests__/index.tsx
+++ b/packages/clay-label/src/__tests__/index.tsx
@@ -50,7 +50,7 @@ describe('Rendering', () => {
 		expect(testRenderer.toJSON()).toMatchSnapshot();
 	});
 
-	it('renders as a link and closable ', () => {
+	it('renders as a link and closable', () => {
 		const testRenderer = TestRenderer.create(
 			<ClayLabel
 				closeButtonProps={{
@@ -61,6 +61,14 @@ describe('Rendering', () => {
 			>
 				{'Label Closable'}
 			</ClayLabel>
+		);
+
+		expect(testRenderer.toJSON()).toMatchSnapshot();
+	});
+
+	it('renders with contentBefore prop', () => {
+		const testRenderer = TestRenderer.create(
+			<ClayLabel contentBefore="Content before">{'Label'}</ClayLabel>
 		);
 
 		expect(testRenderer.toJSON()).toMatchSnapshot();
--- a/packages/clay-label/src/__tests__/index.tsx
+++ b/packages/clay-label/src/__tests__/index.tsx
@@ -66,9 +66,23 @@ describe('Rendering', () => {
 		expect(testRenderer.toJSON()).toMatchSnapshot();
 	});
 
-	it('renders with contentBefore prop', () => {
+	it('renders with ItemBefore', () => {
 		const testRenderer = TestRenderer.create(
-			<ClayLabel contentBefore="Content before">{'Label'}</ClayLabel>
+			<ClayLabel withClose={false}>
+				<ClayLabel.ItemBefore>{'Content before'}</ClayLabel.ItemBefore>
+				<ClayLabel.ItemExpand>{'Label'}</ClayLabel.ItemExpand>
+			</ClayLabel>
+		);
+
+		expect(testRenderer.toJSON()).toMatchSnapshot();
+	});
+
+	it('renders with ItemAfter', () => {
+		const testRenderer = TestRenderer.create(
+			<ClayLabel withClose={false}>
+				<ClayLabel.ItemExpand>{'Label'}</ClayLabel.ItemExpand>
+				<ClayLabel.ItemAfter>{'Content after'}</ClayLabel.ItemAfter>
+			</ClayLabel>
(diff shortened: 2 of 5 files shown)


packages/clay-label/src/index.tsx
```diff
@@ -59,37 +78,25 @@ interface IProps
...
-	>;
-
 	/**
 	 * Path to the location of the spritemap resource used for Icon.
 	 */
 	spritemap?: string;
 }
 
-const ClayLabel = React.forwardRef<HTMLAnchorElement | HTMLSpanElement, IProps>(
+const ClayLabelBase = React.forwardRef<HTMLSpanElement, IBaseProps>(
```
matuzalemsteles: I think we went back to the name `Base` 😂, is a dead end!

packages/clay-label/src/index.tsx
```diff
@@ -100,35 +107,46 @@ const ClayLabel = React.forwardRef<HTMLAnchorElement | HTMLSpanElement, IProps>(
...
+								'close'
+							)}
 							type="button"
 						>
-							<Icon spritemap={spritemap} symbol="times" />
+							<ClayIcon spritemap={spritemap} symbol="times" />
 						</button>
-					</span>
+					</ClayLabelItemAfter>
 				)}
```
matuzalemsteles: Maybe move this to `ClayLabel` below?

bryceosterhaus: I was debating back and forth whether this should be an easier API, or if we should require the end user to wire up the appropriate close button. I sort of like having it internal to ClayLabelBase since it doesn't prohibit or get in the way at all, but I also could see moving it to just the high-level. Are you thinking we would name it `ClayLabelWithClose`? If that's the case, we could do what we did modal(withTitle) and add `withClose` as a prop that defaults to true, and then the withClose logic would ultimately get extracted to a high-level component

packages/clay-label/src/index.tsx
bryceosterhaus: haha yeah I wasn't sure what else to call it. I'll update to be similar to our other components so that we avoid confusion.

bryceosterhaus:
@matuzalemsteles check out the latest commit for an example of adding the pattern that we did with Modal

packages/clay-label/src/index.tsx
matuzalemsteles: Yeah, actually, I wasn't thinking of creating a `ClayLabelWithClose` now, just thinking of making `ClayLabelBase` cleaner and currently moving it to `ClayLabel` only, we will still keep supporting it. I don't think we need `withClose` here yet. We can extract it in the future when we better consolidate the ClayLabel API.

packages/clay-label/src/index.tsx
matuzalemsteles: haha no problem. I think we can keep the `Base`.

packages/clay-label/src/index.tsx
```diff
@@ -101,51 +93,93 @@ const ClayLabelBase = React.forwardRef<HTMLSpanElement, IBaseProps>(
...
+				{!withClose && children}
+
+				{withClose && (
+					<>
+						<ClayLabelItemExpand {...innerElementProps} href={href}>
+							{children}
+						</ClayLabelItemExpand>
+
+						{closeButtonProps && (
+							<ClayLabelItemAfter>
```
kresimir-coko: It seems like this implementation doesn't consume `ClayLabelItemBefore` or am I wrong?

packages/clay-label/src/__tests__/index.tsx
```diff
@@ -65,6 +65,28 @@ describe('Rendering', () => {
...
+		);
+
+		expect(testRenderer.toJSON()).toMatchSnapshot();
+	});
+
+	it('renders with ItemAfter', () => {
+		const testRenderer = TestRenderer.create(
+			<ClayLabel.Base>
+				<ClayLabel.ItemExpand>{'Label'}</ClayLabel.ItemExpand>
+				<ClayLabel.ItemAfter>{'Content before'}</ClayLabel.ItemAfter>
```
wincent: before → after

Obligatory copy-pasta image:

copy-pasta

packages/clay-label/src/index.tsx
bryceosterhaus: Correct. Rather than add a new prop to the API, I went the route of breaking up Label into its low-level components so that users have more flexibility in how they want to use it. You can see the example of how we would add a before item at https://github.com/liferay/clay/pull/3081/files#diff-d25f4565a23c4c3cc5475dedd79d9ad1R62-R67

matuzalemsteles:
hey @bryceosterhaus just to complement this change we can also add documents to clayui.com, perhaps to explain the addition of `withClose`.

bryceosterhaus:
@matuzalemsteles good call. I will add that in this PR as well

bryceosterhaus:
@matuzalemsteles update docs. let me know what you think

clayui.com/content/docs/components/label.mdx
```diff
@@ -22,6 +22,12 @@ import {
 </div>
 </div>
 
+## Overview
+
+`ClayLabel` offers two different APIs for use by toggling the prop `withClose`. By default(`withClose={true}`), `ClayLabel` behaves like a high-level component. If you want to use the lower-level components and compose multiple parts to your label, all you need to do is set `withClose={false}`.
+
+Check out [this storybook example](storybook.clayui.com/?path=/story/components-claylabel--w-content-before) for a demo.
+
```
matuzalemsteles: @bryceosterhaus as it is a typical feature that we don't want to support for a long time, it might be worth adding a warning about it and that we will remove this double behavior in a major version later. What do you think?

bryceosterhaus: @matuzalemsteles I'm thinking its probably best to not mention that since they are essentially toggling the intended behavior of what this component will be. The next major version will likely behave the same even if `withClose={false}` is set.

When we do ever release 4.x, we can then note that `with*` props are moved to high-level components. But overall, I think it'd be unnecessary to mention at the moment since 4.x isn't really even on the radar yet. Once we get 4.x on the radar and have a better idea of what it would be, we can always add the note then.

matuzalemsteles: Yeah, no problem. LGTM

packages/clay-label/src/index.tsx
kresimir-coko: Okay I see, good to know! Thanks

PR #3095, opened 2020-03-30
fix: Fix accessibility issues

bryceosterhaus:
Hey @kresimir-coko, here are some changes I made on top of your PR. Check out my PR comments for explanations of changes


packages/clay-card/src/CardWithHorizontal.tsx
```diff
@@ -15,16 +15,14 @@ interface IProps {
...
-		dropDownTrigger: string;
-	};
+	disabled?: boolean;
 
 	/**
-	 * Flag to indicate that all interactions on the card will be disabled.
+	 * Props to add to the dropdown trigger element
 	 */
-	disabled?: boolean;
+	dropDownTriggerProps?: React.HTMLAttributes<HTMLButtonElement>;
```
bryceosterhaus: Rather than have just aria labels available to the trigger element, I thought it would be better to provide a more robust API to allow more customization on that element. This allows them to add other accessibility attrs as needed.

packages/clay-card/src/CardWithInfo.tsx
```diff
@@ -36,6 +29,11 @@ interface IProps {
 	 */
 	disabled?: boolean;
 
+	/**
+	 * Props to add to the dropdown trigger element
+	 */
+	dropDownTriggerProps?: React.HTMLAttributes<HTMLButtonElement>;
```
bryceosterhaus: Same as above

> Rather than have just aria labels available to the trigger element, I thought it would be better to provide a more robust API to allow more customization on that element. This allows them to add other accessibility attrs as needed.

packages/clay-card/src/CardWithUser.tsx
```diff
@@ -19,12 +19,9 @@ interface IProps {
...
 
 	/**
-	 * Labels for the aria attributes
+	 * Props to add to the dropdown trigger element
 	 */
-	ariaLabels?: {
-		cardCheckbox: string;
-		dropDownTrigger: string;
-	};
+	checkboxProps?: React.HTMLAttributes<HTMLInputElement>;
```
bryceosterhaus: Same as above. This is even more important for form elements since they should be able to also add a `name` attr if needed.

packages/clay-card/src/CardWithUser.tsx
```diff
@@ -36,6 +33,11 @@ interface IProps {
 	 */
 	disabled?: boolean;
 
+	/**
+	 * Props to add to the dropdown trigger element
+	 */
+	dropDownTriggerProps?: React.HTMLAttributes<HTMLButtonElement>;
```
bryceosterhaus: same as above

packages/clay-form/src/DualListBox.tsx
```diff
@@ -56,8 +56,8 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	 * Labels for the aria attributes
 	 */
 	ariaLabels?: {
-		transferButtonLTR: string;
-		transferButtonRTL: string;
+		transferLTR: string;
```
bryceosterhaus: For all ariaLabel props, I think we should focus less on the element itself and more on the language the user wants to use. So by removing `Button` from the name, it's more clear that this prop is for a language map.

packages/clay-form/src/SelectBox.tsx
```diff
@@ -57,13 +57,13 @@ function reorderDown(array: Array<TItem>, selectedIndexes: Array<number>) {
...
 
-interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+interface IProps extends React.HTMLAttributes<HTMLSelectElement> {
 	/**
 	 * Labels for aria attributes
 	 */
 	ariaLabels?: {
-		reorderButtonUp: string;
-		reorderButtonDown: string;
+		reorderUp: string;
```
bryceosterhaus: Same as above

> For all ariaLabel props, I think we should focus less on the element itself and more on the language the user wants to use. So by removing Button from the name, it's more clear that this prop is for a language map.

packages/clay-form/src/SelectBox.tsx
```diff
@@ -96,11 +95,6 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	 */
 	onSelectChange: (val: Array<string>) => void;
 
-	/**
-	 * Id of the selectbox to poperly link it to the label
-	 */
-	selectBoxId?: string;
```
bryceosterhaus: Since the component itself is called a `Select`, I think its important we extend from its html element so that it behaves similarly to how people would use the html select. By doing this, we only need the `id`

packages/clay-form/src/SelectBox.tsx
```diff
@@ -164,9 +156,9 @@ const ClaySelectBox: React.FunctionComponent<IProps> = ({
 	);
 
 	return (
-		<div {...otherProps} className={classNames(className, 'form-group')}>
```
bryceosterhaus: Since this is called a `Select` component, I would expect the ...otherProps to apply to the select element and not the container. If we need the ability to add props to the container, we can use `containerProps`. This is really only for the case when the component name is not consistent with the container element.

packages/clay-list/src/ListWithItems.tsx
```diff
@@ -33,6 +26,11 @@ interface IListItem {
 		typeof ClayDropDownWithItems
 	>['items'];
 
+	/**
+	 * Props to add to the dropdown trigger element
+	 */
+	dropDownTriggerProps?: React.HTMLAttributes<HTMLButtonElement>;
```
bryceosterhaus: See above

packages/clay-pagination/src/PaginationWithBasicItems.tsx
```diff
@@ -66,8 +66,8 @@ export const ClayPaginationWithBasicItems = React.forwardRef<
 		{
 			activePage,
 			ariaLabels = {
-				nextPageButton: 'Next',
-				previousPageButton: 'Previous',
+				next: 'Next',
```
bryceosterhaus: See above

packages/clay-time-picker/src/index.tsx
```diff
@@ -51,8 +51,8 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 		hours: string;
 		minutes: string;
 		clearButton: string;
-		timeDownButton: string;
-		timeUpButton: string;
+		timeDown: string;
```
bryceosterhaus: See above

packages/clay-time-picker/src/index.tsx
```diff
@@ -51,8 +51,8 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 		hours: string;
 		minutes: string;
 		clearButton: string;
```
bryceosterhaus: I noticed this didn't follow the "language map" pattern, so I updated this in the following commit

packages/clay-card/src/CardWithHorizontal.tsx
kresimir-coko: I see, you went a step further and added props for those triggers, nice!

packages/clay-form/src/DualListBox.tsx
kresimir-coko: Yea the naming was a bit tricky to decide on

packages/clay-form/src/SelectBox.tsx
kresimir-coko: Alright, good to know! Makes sense now that you put it like that but I kinda forgot about `containerProps`

kresimir-coko:
@bryceosterhaus all makes sense, the naming I was unsure of, as well as the `{...otherProps}` part in the select component, but yea your changes look great!

bryceosterhaus:
I squashed everything into one commit for the sake of versioning. Just need to add some updated docs for this

PR #3106, opened 2020-04-01
feat(@clayui/demos): Drag and drop demo for table columns and rows

kresimir-coko:
Hey @bryceosterhaus I've drafted a PR with 2 commits:

- Demos Commit is the one that I've tackled first thinking that's the requirement of the task, but then I figured out that your sentence "Feel free to get creative with it and make a "real" table that has drag and drop." might mean to give users of ClayTable an option to support DnD via a `draggable` prop. So in demos is an example of having a DnD Table but, at least IMO that doesn't really help our users as much as it could, it just gives them an example of how to use it.
- The @clayui/table commit is my proposition of how the API could look without investing too much time into making a working example as I'm unsure what're the exact requirements of the task.

Let me know which of the commits to pursue and polish. In case you wanna have a discussion about this hit me up on Slack when you see this.

### DIFF
--- a/packages/demos/stories/TableColumnsDragDrop.tsx
+++ b/packages/demos/stories/TableColumnsDragDrop.tsx
@@ -13,6 +13,7 @@ import {
 	useDragLayer,
 	useDrop,
 } from 'react-dnd';
+import {getEmptyImage} from 'react-dnd-html5-backend';
 
 const TABLE_COLUMN_DND_TYPE = 'tableColumn';
 
@@ -141,7 +142,7 @@ const ColumnDragPreview: React.FunctionComponent<IColumnDragPreviewProps> = ({
 	return (
 		<div style={dragLayerStyles}>
 			<div style={dragItemStyles}>
-				<ClayTable>
+				<ClayTable className="table-clone">
 					<ClayTable.Head>
 						<ClayTable.Row>
 							<ClayTable.Cell headingCell headingTitle>
@@ -229,12 +230,16 @@ const DraggableTableHeaderCell: React.FunctionComponent<
 		},
 	});
 
-	const [{itemBeingDragged}, drag] = useDrag({
+	const [{itemBeingDragged}, drag, preview] = useDrag({
 		collect: (monitor: any) => ({
 			itemBeingDragged: monitor.getItem() || {id: 0},
 		}),
 		item: {id, index, type: TABLE_COLUMN_DND_TYPE},
 	});
+
+	React.useEffect(() => {
+		preview(getEmptyImage(), {captureDraggingState: true});
+	}, [preview]);
 
 	drag(drop(ref));
 
@@ -273,12 +278,35 @@ const ClayTableWithDraggableColumns: React.FunctionComponent = () => {
 	const tableRows = new Array(tableColumns[0].data.length)
 		.fill(null)
 		.map((_item, index: number) =>
-			tableColumns.map((column: TColumn) => column.data[index])
+			tableColumns.map((column: TColumn) => ({
+				data: column.data[index],
+				id: column.id,
+			}))
 		);
+
+	const [{itemBeingDragged}] = useDrag({
+		collect: (monitor: any) => ({
+			itemBeingDragged: monitor.getItem(),
+		}),
+		item: {type: TABLE_COLUMN_DND_TYPE},
+	});
+
+	const [{canDrop}] = useDrop({
+		accept: TABLE_COLUMN_DND_TYPE,
+		collect: (monitor: DropTargetMonitor) => ({
+			canDrop: monitor.canDrop(),
+		}),
+	});
+
+	const isDragging = itemBeingDragged ? true : false;
 
 	return (
 		<>
-			<ClayTable className="table-drag">
+			<ClayTable
+				className={classNames('table-drag', {
+					'table-dragging': isDragging,
+				})}
+			>
 				<ClayTable.Head>
 					<ClayTable.Row>
 						{tableColumns.map((column: TColumn, index: number) => (
@@ -294,18 +322,26 @@ const ClayTableWithDraggableColumns: React.FunctionComponent = () => {
 				</ClayTable.Head>
 
 				<ClayTable.Body>
-					{tableRows.map((row, index) => (
-						<ClayTable.Row key={`${row[index]}_${index}`}>
+					{tableRows.map((row, index: number) => (
+						<ClayTable.Row
+							key={`${row[index].id}_${row[index].data}`}
+						>
 							{row.map(cell => (
 								<ClayTable.Cell
-									key={`${cell}_${Math.random().toString()}`}
+									className={classNames({
+										'c-dragging':
+											itemBeingDragged &&
+											itemBeingDragged.id === cell.id,
+										'c-droppable': canDrop,
+									})}
+									key={`${cell.id}_${cell.data}`}
 									ref={node =>
 										node &&
 										setDragPreviewWidth(node.offsetWidth)
 									}
 									truncate
 								>
-									{cell}
+									{cell.data}
 								</ClayTable.Cell>
 							))}
 						</ClayTable.Row>
--- a/packages/demos/stories/TableColumnsDragDrop.tsx
+++ b/packages/demos/stories/TableColumnsDragDrop.tsx
@@ -1,0 +1,321 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayTable from '@clayui/table';
+import classNames from 'classnames';
+import React from 'react';
+import {
+	DropTargetMonitor,
+	XYCoord,
+	useDrag,
+	useDragLayer,
+	useDrop,
+} from 'react-dnd';
+
+const TABLE_COLUMN_DND_TYPE = 'tableColumn';
+
+const initialTableColumns = [
+	{
+		data: ['Red, white & blue', 'Tricolore', 'Stars & stripes'],
+		id: 'flag',
+		title: 'Flag',
+	},
+	{
+		data: ['Europe', 'Europe', 'North America'],
+		id: 'region',
+		title: 'Region',
+	},
+	{
+		data: ['Croatia', 'Italy', 'USA'],
+		id: 'country',
+		title: 'Country',
+	},
+	{
+		data: [
+			'Lorem ipsum dolor, sit amet consectetur adipisicing elit.',
+			'Lorem ipsum dolor, sit amet consectetur adipisicing elit.',
+			'Lorem ipsum dolor, sit amet consectetur adipisicing elit.',
+		],
+		id: 'description',
+		title: 'Description',
+	},
+];
+
+export type TColumn = {
+	data: Array<string>;
+	id: string;
+	title: string;
+};
+
+export interface ICustomDragLayerProps {
+	/**
+	 * List of columns in which the Drag Layer will find the one currently being dragged
+	 */
+	columns: Array<TColumn>;
+
+	/**
+	 * Width of the Drag Layer as specified by the width of the heading cell
+	 */
+	width: number;
+}
+
+const CustomDragLayer: React.FunctionComponent<ICustomDragLayerProps> = ({
+	columns,
+	width,
+}) => {
+	const {currentOffset, initialOffset, isDragging, item} = useDragLayer(
+		monitor => ({
+			currentOffset: monitor.getSourceClientOffset(),
+			initialOffset: monitor.getInitialSourceClientOffset(),
+			isDragging: monitor.isDragging(),
+			item: monitor.getItem(),
+		})
+	);
+
+	if (!isDragging) {
+		return null;
+	}
+
+	const columnBeingDragged = columns.find(
+		(column: TColumn) => column.id === item.id
+	);
+
+	return (
+		<ColumnDragPreview
+			column={columnBeingDragged!}
+			currentOffset={currentOffset}
+			initialOffset={initialOffset}
+			width={width}
+		/>
+	);
+};
+
+export interface IColumnDragPreviewProps {
+	/**
+	 * Column to be used as a Drag Preview
+	 */
+	column: TColumn;
+
+	/**
+	 * Current offset of the Drag Preview from the initial position of the heading cell
+	 */
+	currentOffset: XYCoord | null;
+
+	/**
+	 * Initial offset of the Drag Preview
+	 */
+	initialOffset: XYCoord | null;
+
+	/**
+	 * Width of the Drag Layer as specified by the width of the heading cell
+	 */
+	width: number;
+}
+
+const ColumnDragPreview: React.FunctionComponent<IColumnDragPreviewProps> = ({
+	column,
+	currentOffset,
+	initialOffset,
+	width,
+}) => {
+	const dragLayerStyles: React.CSSProperties = {
+		height: '100%',
+		left: 0,
+		pointerEvents: 'none',
+		position: 'fixed',
+		top: 0,
+		width,
+		zIndex: 1060,
+	};
+
+	const dragItemStyles: React.CSSProperties = {
+		display: !initialOffset || !currentOffset ? 'none' : '',
+		transform:
+			initialOffset && currentOffset
+				? `translate(${currentOffset.x}px, ${currentOffset.y}px)`
+				: '',
+	};
+
+	return (
+		<div style={dragLayerStyles}>
+			<div style={dragItemStyles}>
+				<ClayTable>
+					<ClayTable.Head>
+						<ClayTable.Row>
+							<ClayTable.Cell headingCell headingTitle>
+								{column.title}
+							</ClayTable.Cell>
+						</ClayTable.Row>
+					</ClayTable.Head>
+
+					<ClayTable.Body>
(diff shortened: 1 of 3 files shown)


packages/demos/stories/TableDragDrop.tsx
```diff
@@ -0,0 +1,170 @@
...
+		}),
+		item: {id, index, type: 'tableRow'},
+	});
+
+	const isItemBeingDragged = itemBeingDragged.id === id;
+
+	drag(drop(ref));
+
+	const style = {
+		backgroundColor: isItemBeingDragged ? '#EFEFEF' : '',
```
bryceosterhaus, line 102: Are these styles taken from lexicon or clay? I think lexicon/clay has some

packages/demos/stories/TableDragDrop.tsx
```diff
@@ -0,0 +1,170 @@
...
+import ClayTable from '@clayui/table';
+import {XYCoord} from 'dnd-core';
+import React from 'react';
+import {DropTargetMonitor, useDrag, useDrop} from 'react-dnd';
+
+interface IDraggableTableRowProps {
+	/**
+	 * List Item unique identifier
+	 */
+	id: any;
```
bryceosterhaus, line 15: Shouldn't have an `any` type. `string | number` is probably best

packages/demos/stories/TableDragDrop.tsx
```diff
@@ -0,0 +1,170 @@
...
+		backgroundColor: isItemBeingDragged ? '#EFEFEF' : '',
+		border: isItemBeingDragged ? '2px solid #555555' : '',
+		cursor: 'grab',
+		opacity: isDragging ? 0 : 1,
+	};
+
+	return (
+		<ClayTable.Row id={`tableRow${id}`} ref={ref} style={style}>
+			<ClayTable.Cell headingTitle>{id}</ClayTable.Cell>
+			<ClayTable.Cell>{'South America'}</ClayTable.Cell>
```
bryceosterhaus, line 111: Instead of hardcoding these values, lets pass in unique values. Similar to how a table might look and behave in a real app.

packages/demos/stories/TableDragDrop.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+			tempList.splice(hoverIndex, 0, tableRows[dragIndex]);
+
+			setTableRows(tempList);
+		},
+		[tableRows]
+	);
+
+	return (
+		<ClayTable>
```
bryceosterhaus, line 139: Same for this component, lets not use hardcoded values and instead pass in values like we might use in an app.

bryceosterhaus:
Oh and can you also look into dnd for the columns of the table too? I noticed lexicon mentions that at https://liferay.design/lexicon/core-components/table/#columns-drag-and-drop

packages/demos/stories/TableDragDrop.tsx
kresimir-coko: These were just copy/pasted from the DraggableList example, just as a PoC

jbalsas:
In case you guys want to draw inspiration: https://docs.google.com/document/d/1PNudTqmS1JCSJwMt_h2CA9QobplxIDfwF7tZsjUpavw/edit#heading=h.9lnxnyfkrxsl

kresimir-coko:
@bryceosterhaus I "love" the 🔧 you threw into this by mentioning columns.

So, I made the drag and drop on Rows work nicely, and that's a separate commit. Then I did some work on making columns drag and droppable but I didn't finish it, so I included a commit of where I ended up today. 

I got to defining `columnBeingReplacedContent` and `columnReplacingContent`, with the future step being somehow switching those values in Rows, but I don't know how to really achieve that without significantly refactoring the demo.

One idea I have is to not use an array of objects to iterate over that represent Rows, but instead create an array of objects that represent Cells.

I don't mind refactoring the demo, just ran out of time today so it's a good time to ask you what you think the next steps should be when making columns draggable?

pat270:
@kresimir-coko I'm taking a look right now. I might have a demo for you tomorrow or the next day.

kresimir-coko:
Hey @bryceosterhaus I've polished up the demo of draggable Rows, and made a crude, but working version of draggable Columns.

Had some issues wrapping my head around the implementation/algorithm, whatever you'd call fiddling with arrays. But all in all it works right now, it's not close to how Lexicon envisioned, but we can get there with this as a foundation.

LMK what you think and if you think there are just some small fixes, as always, feel free to take over and handle those :)

pat270:
@kresimir-coko @bryceosterhaus I just fired up this demo and realize the way this drag and drop plugin works doesn't fit with the way the Lexicon team envisioned. Should we ask them to change it to better fit the way the plugin works or you think we can modify it enough to fit their vision?

My main concern is that the column shifts when you mouse over instead of having a placeholder and shifting after the mouse is released.

bryceosterhaus:
@kresimir-coko looks pretty good. Lets continue to iterate  more and get closer to the lexicon specification. There is no rush right now for anything to be committed so we can take some time on this can refine it so that we have a consistent example with the lexicon definition. But so far its looking pretty good! And if you need guidance and direction for the logic of manipulating the columns, I'm sure there are some examples out there of people using react-dnd with tables. You could try googling around or even searching on github for how other people are implementing it. Its a popular library so I would expect people to have already done it.

pat270:
See https://github.com/liferay/clay/pull/3114 on how to apply drag and drop styles

kresimir-coko:
Hey @bryceosterhaus I've spent the day trying to change the implementation so that when you drag the heading, the entire column gets dragged, was inspired by [this article](https://medium.com/@stides303/flexible-layouts-with-react-dnd-6049571ef325), but it didn't really do much, even the example couldn't get built 🤷‍♂so I abandoned it.

I kinda doubt it's even possible with the current implementation of the ClayTable component, where we are forced into using rows instead of columns for rendering. But it might just be that I'm not thinking outside the box enough. The closest I've come up while investigating was [Add a Drag Preview Image](https://react-dnd.github.io/react-dnd/docs/tutorial#add-a-drag-preview-image) example from their tutorial, which could maybe, somehow, if we stretch it, preview the entire column while dragging.

The `c-dragging` & `c-droppable` classes were easy to add, the problem came when I wanted to add classes to the table itself, as I'm unsure how I can expose react-dnd's state to the ClayTable component without, i guess, just moving the code that handles dragging and dropping into it's scope.

I'm unsure how to bring it closer to Lexicon's example apart from making sure proper styles are applied depending on the state. So if you've got any specific pointers I'm all ears. The examples returned by google search are kinda lacklustre, either too simple or too old. The only other thing I could work on here is adding an example of dragging by dragging the handle.

packages/demos/stories/TableRowsDragDrop.tsx
```diff
@@ -0,0 +1,206 @@
...
+			const draggingDownwards =
+				dragIndex < hoverIndex && pixelsToTop < verticalMiddle / 2;
+
+			if (draggingDownwards || draggingUpwards) {
+				return;
+			}
+
+			onMove(dragIndex, hoverIndex);
+
+			item.index = hoverIndex;
```
bryceosterhaus, line 91: This seems odd to be mutating the item itself. Is this the components props we are mutating? I would expect that the `onMove` callback should be sufficient for our logic.

packages/demos/stories/TableRowsDragDrop.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+	drag(drop(ref));
+
+	const style = {
+		cursor: isItemBeingDragged ? 'grabbing' : 'grab',
+		opacity: isDragging ? 0 : 1,
+	};
+
+	return (
+		<ClayTable.Row id={`tableRow${content.id}`} ref={ref} style={style}>
```
bryceosterhaus, line 113: The use of `id` here is a little suspect to me. We should avoid using `id` on elements since we can only have one unique id per page.

packages/demos/stories/TableColumnsDragDrop.tsx
```diff
@@ -0,0 +1,215 @@
...
+	type: string;
+}
+
+const DraggableTableHeaderCell: React.FunctionComponent<
+	IDraggableTableHeadingProps
+> = ({expanded, id, index, onMove, title}) => {
+	const ref = React.useRef<HTMLTableHeaderCellElement>(null);
+
+	const [, drop] = useDrop({
+		accept: 'tableColumn',
```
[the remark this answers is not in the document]
kresimir-coko: This is something that we define, it's purpose is to prevent elements being dropped in places they aren't supposed to be dropped. For example, dropping a table row where a heading is supposed to be dropped.

packages/demos/stories/TableRowsDragDrop.tsx
kresimir-coko: That line is taken straight from their examples: [Sortable Example](https://react-dnd.github.io/react-dnd/examples/sortable/simple), if you go into the `Card.tsx` file in the typescript example, on line 72 there's the following comment: 

> Note: we're mutating the monitor item here! Generally it's better to avoid mutations, but it's good here for the sake of performance to avoid expensive index searches.

@wincent what do you think, should I invest some time and try to remove the need for the mutation or keep it since it is recommended by the author of the package?

packages/demos/stories/TableColumnsDragDrop.tsx
```diff
@@ -0,0 +1,215 @@
...
+	tableColumns[0].data.forEach(() => {
+		tableRows.push([]);
+	});
+
+	tableColumns.forEach(column => {
+		for (let i = 0; i < tableRows.length; i++) {
+			console.log(tableRows[i]);
+			tableRows[i].push(column.data[i]);
+		}
+	});
```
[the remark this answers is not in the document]
wincent: Don't forget the `new` before `Array()`.

packages/demos/stories/TableRowsDragDrop.tsx
wincent: Unfortunately, GitHub pages is down so I can't even look at the example. 😂

https___react-dnd_github_io_react-dnd_examples_sortable_simple

I don't know what the `hover` property in `useDrop` actually works. It does seem like you could do the work inside `onMove`. As this code is written it looks like `onMove` takes the item from `dragIndex` and moves it to `hoverIndex`, and then this line updates the dragged item's index, which seems wrong.

eg: if I drag "E" to the "C"

```
item    A B C D E F G
index   1 2 3 4 5 6 7
```

the desired end state should be:

```
item    A B E C D F G
index   1 2 3 4 5 6 7
```

but it will be:

```
item    A B E C D F G
index   1 2 3 3 4 6 7
```

Because we're only changing the index of item "E", and not the indices of "C" and "D", which both got bumped further to the right.

I don't know whether that leads to any actual breakage though, but it might be worth doing it without mutation in `onMove()`. The simplest way to do that is probably with two passes: one to put the items in the right place, and one to adjust the index of any item that isn't right; something like (untested):

```js
	const onMove = React.useCallback(
		(dragIndex: number, hoverIndex: number) => {
			const list = [...tableColumns];

			list.splice(dragIndex, 1);

			list.splice(hoverIndex, 0, tableColumns[dragIndex]);

			setTableColumns(list.map((item, i) => {
				if (item === i) {
					return item;
				} else {
					return {
						...item,
						index: i,
					};
				}
			}));
		},
		[tableColumns]
	);
```

wincent: You could do this in one pass, but I think it would be fiddly to get right, and harder to read. We don't expect the number of columns to be very large (even in crazy scenarios, I can't see it being more than about a dozen or so), so I would be wary of prematurely optimizing.

kresimir-coko:
Hey @bryceosterhaus, it seems I didn't push my changes yesterday when I referred to simplifying some things. It's okay tho, because your review hit on the points present in the code I had anyway. So I tackled some of those as well as some others I noticed.

The latest commit is a herald of good news, it seems that using some of react-dnd's more advanced APIs we can display the entire column being dragged, instead of just the heading. I've commited code up to a point I reached today, with more work needing to be done, but you can see how it might end up looking when it's finished.

On a different note, @pat270, I implemented some of the class suggestions you made, but it seems that we do not yet have `c-dragging` & `c-droppable` classes defined, or am I mistaken? My code applies them but no styles are being added.

@bryceosterhaus in case you can't review this today, don't worry, I have more work on this tomorrow, but still wanted to give you an option to see where I got to.

pat270:
@kresimir-coko the CSS isn't available in master. It's in #3114.  I just wanted to make sure it could be done before making it a component.

kresimir-coko:
Hey @bryceosterhaus here's finally the code that you should spend some quality time reviewing. Drag preview works, doesn't look 100% like it's supposed to but that might be more on @pat270 to take a look. Here's some bullet points to keep in mind:
- `keys` are lacking
- JSDoc needs writing
- not sure if possible to get `table-dragging` class to the table
- drag preview styles aren't final
- trying to implement @wincent's suggestion on removing the mutation made the drop action lag immensely , I can give it more time tomorrow and try to come up with something different

wincent:
> trying to implement @wincent's suggestion on removing the mutation made the drop action lag immensely

ashamed

packages/demos/stories/TableColumnsDragDrop.tsx
```diff
@@ -0,0 +1,286 @@
...
+	},
+];
+
+const dragLayerStyles: React.CSSProperties = {
+	height: '100%',
+	left: 0,
+	pointerEvents: 'none',
+	position: 'fixed',
+	top: 0,
+	width: '100%',
```
bryceosterhaus, line 52: Any way we can get the width of the row itself? It looks a little odd to have the drag layer with 100% width.

packages/demos/stories/TableColumnsDragDrop.tsx
```diff
@@ -0,0 +1,286 @@
...
+	zIndex: 100,
```
bryceosterhaus, line 53: Where is this zIndex from? I think we should probably grab these from clay-css if possible

packages/demos/stories/TableColumnsDragDrop.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+export interface IColumnDragPreviewProps {
+	column: TColumn;
+	currentOffset: XYCoord | null;
+	initialOffset: XYCoord | null;
+}
+
+const ColumnDragPreview: React.FunctionComponent<
+	IColumnDragPreviewProps
+> = React.memo(({column, currentOffset, initialOffset}) => {
```
bryceosterhaus, line 103: Whats the thought behind adding `memo`? Does it make a notable difference if we don't use them? Primarily I am thinking of how we can simplify for the sake of others who will look at this demo for an example.

kresimir-coko: Removing the memo worked! It was taken from their examples.

packages/demos/stories/TableColumnsDragDrop.tsx
kresimir-coko: The `100`was in the example, now I grabbed `1060` from [_globals-z-index.scss#L6](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/variables/_globals-z-index.scss#L6). Not sure if we can somehow access SCSS variables from JSX.

kresimir-coko:
Hey @bryceosterhaus I've squashed the commits into 3 that make the most sense.

I haven't managed to get the `isDragging` state to the `ClayTable`, but I did get the height of the Drag Preview to be dynamically set. I also added the JSDoc. I think that style-wise we need to wait for @pat270 to add them to master, I added the classes that I could, it's just that they don't have any styles attached to them yet.

I ask you to try and get the `table-dragging` class sorted, and fix up the `key`s on some of the `map`s that are kinda iffy, I had to use indexes or `Math.random()` because I couldn't think of something better.

pat270:
If it's too much work to get a class on the `table` element, we can add `c-dragging` to all `th` and `td` instead.

packages/demos/stories/TableColumnsDragDrop.tsx
bryceosterhaus: No worries about grabbing a variable, as long as we are in sync with our css definition we should be good

pat270:
@kresimir-coko @bryceosterhaus I sent the CSS in https://github.com/liferay/clay/pull/3125. We might not need to add `table-dragging`. I scoped as much as I could to `table-drag`. The `table-dragging` class is only used to prevent table text from being highlighted and keeping the move cursor when dragging a column. I don't remember seeing that happening with the react plugin, but I didn't test it.

packages/demos/stories/TableColumnsDragDrop.tsx
```diff
@@ -0,0 +1,321 @@
...
+	 */
+	columns: Array<TColumn>;
+
+	/**
+	 * Width of the Drag Layer as specified by the width of the heading cell
+	 */
+	width: number;
+}
+
+const CustomDragLayer: React.FunctionComponent<ICustomDragLayerProps> = ({
```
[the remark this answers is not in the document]
kresimir-coko: Great find!

kresimir-coko:
Hey @bryceosterhaus I've pushed another commit here since you didn't get to reviewing it yet. 

I've managed to add `table-dragging` to the table, implemented the suggestion by @matuzalemsteles which fixed the double heading on drag preview, cleaned up one of the keys, although it's still not perfect I guess. I've also added, I think, all CSS thingies that @pat270 made and it seems to be matching Lexicon now.

Issue #3068, opened 2020-03-25
Move component docs from clayui.com to their respective packages 


bryceosterhaus:
Instead of having all of our component docs in the clayui.com directory, I think it would make more sense to co-locate the docs with each package. This way the docs get versioned in the same manner as the component itself. This also would provide a lower barrier of entry to updating the docs when we update components. And ideally this move would not change the site at all, but instead clayui.com would compile the docs from each package.

jbalsas:
![hellyeah](https://user-images.githubusercontent.com/905006/77624561-4981f600-6f42-11ea-8402-1f00c5f20fae.gif)

bryceosterhaus:
https://github.com/liferay/clay/pull/3092#issuecomment-606451143

pat270:
Since we are releasing each package individually, I was wondering how we were going to unify the docs to make it easy to consume by Portal users? I'm visioning having to comb through Portal's source to see what version of Clay Autocomplete it's using and matching it to the same Clay CSS version.

bryceosterhaus:
@pat270 Ideally portal should be on the latest release of clay and whatever is latest on clayui.com. The primary goal initially is for the site not to contain docs for features that aren't released yet. As of now, we can't update the site without having it go live. Also by co-locating the docs, we can at least see the history of the docs in the same context of the history of the source code.

PR #3117, opened 2020-04-08
chore(clayui.com): co-locate docs with components

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/3068

I thought it was best to get this first iteration in before adding on top. This PR moves all component mdx docs to their respective package. This first step helps the docs be updated at the same rate and versioned with its respective package. I also removed nesting of components on sidebar so that its more consistent and findable. I noticed we had nested some components and not others, I thought it would be better to keep everything flat.

Old: 
Screen Shot 2020-04-08 at 10 50 01 AM

New:
Screen Shot 2020-04-08 at 10 50 22 AM

## To do
- Eventually I would like to get the README of each package to be more readable as its own docs source. For example, see this [button component from ant](https://github.com/ant-design/ant-design/blob/master/components/button/index.en-US.md).
- Move inline examples to the mdx file or to the package itself. Right now the demos are still under `clayui.com/src`.

Curious to hear your thoughts on this. I ran into numerous issues with gatsby and mdx, this was the cleanest first step forward in my opinion.


bryceosterhaus:
Eventually I would love to be able to have the demos in-line in the mdx like https://mdxjs.com/guides/live-code. This would make it easier to read the README on github.

matuzalemsteles:
hey @bryceosterhaus this looks cool, I will leave some considerations:

- I think that instead of the README being a `.mdx` we can just have a simple `README.md` linking to the respective files of the documentation, I say this because in some packages they can have more than one component, in the case of Forms and currently these files viewed outside the site are not very readable.
- I think moving the documentation to a `/docs` folder within each package instead of being at the root of the package, I think it would be a good standard so the package root is not full of `.mdx` documents in the case of forms. I think this also follows a common pattern that we see in other repositories, not necessarily as a monorepo.
I don't see any problem in keeping the navigation flat, but keeping this some nested has to do with the Lexicon's navigation, so we maintain the same pattern that Lexicon follows so it makes navigation easier for those looking in the Lexicon site and here. I would say that for me this would not be a problem but that was one of the main arguments for keeping the sessions as they are today. Another reason is the grouping of components in common, in the case of components that deal with Forms.

> Eventually I would love to be able to have the demos in-line in the mdx like https://mdxjs.com/guides/live-code. This would make it easier to read the README on github.

Yeah, I think it would be a great idea, we can do that. I tried to do that at the time but I had a lot of problems because the integration with MDX with Gatsby was still being implemented.

bryceosterhaus:
@matuzalemsteles 
> I think that instead of the README being a .mdx we can just have a simple README.md
>
>I think moving the documentation to a /docs folder within each package instead of being at the root of the package, I think it would be a good standard so the package root is not full of .mdx documents in the case of forms

yeah I totally agree that this would be a better approach. However, I think we should do these as a later step as that would require re-writing a lot of markup and demos in the docs. So I would say thats a next step after this first move.

Would you be good with this as a first step and then spinning off another ticket to improve this to add MD files and more structured docs in each package?

matuzalemsteles:
@bryceosterhaus I have no problem going with that and then modifying it, I would just keep the sessions in their current state for now, I think some people still have strong opinions about it.

@drakonux @marcoscv-work @eduardoallegrini do you see any problems about making navigation flat?

I think @jbalsas had an opinion on that too, but as he's not around, we can do the same thing as @wincent 😁

eduardoallegrini:
I can picture some issues with the flat navigation in the case we need to add more and more documentation (as I would like to do 😄) But If I'm not wrong, we can change this behavior in the next step, right?

jbalsas:
> I think @jbalsas had an opinion on that too, but as he's not around, we can do the same thing as @wincent 😁

Double treason!!

beback

Just a consideration... we need to make it easier to navigate back and forth from Lexicon, so people know how to match patterns and components and get their job done. That's my opinion. How we make that happen is up to you. There's definitely more than one way to achieve it (better search, cheatsheet, cross-links...)

wincent:
> yeah I totally agree that this would be a better approach. However, I think we should do these as a later step as that would require re-writing a lot of markup and demos in the docs. So I would say thats a next step after this first move.

(Independently of how many steps you do this in) I really hope you can do this non-manually (with a script, or a Vim macro, or whatever hackery works)...

drakonux:
We agree with termichema  :)

From our side, we have plans to review the components taxonomy at some moment in Lexicon Site. But without testing, I can't say if creating more levels or increasing the breadth (flattening the hierarchies) will be better. In my experience, a good taxonomy has a balance between deep and breadth. 

For now, not having some hierarchies **in Lexicon site** would cause us to review some names such as Navigation card. 

But feel free to do it here because I have to admit that in components should be just components. i.e. **navigation** and **forms** are names created to relate the components within however a checkbox or an input doesn't have to be exclusively in a form.

bryceosterhaus:
> @matuzalemsteles 
> I would just keep the sessions in their current state for now, 

What do you mean by "sessions" here? 

> @jbalsas 
> we need to make it easier to navigate back and forth from Lexicon, so people know how to match patterns and components and get their job done. 

I'd be curious how people use the two docs. Do users typically go to clayui.com first to find what they need, or do they go to lexicon first? My thought is that a dev would see the design they need to implement for their app, head to clayui.com to find the right component, and then only occasionally go to lexicon for a specific definition. Would I be wrong to assume this? By flattening the nav, I was hoping for more visibility of components rather than clicking through nested structures.

> @wincent 
> (Independently of how many steps you do this in) I really hope you can do this non-manually (with a script, or a Vim macro, or whatever hackery works)...

I can automate a handful of it, but some of it requires manual manipulation unfortunately.

> @drakonux 
> In my experience, a good taxonomy has a balance between deep and breadth

Totally agree, which is why I think it's still helpful to have the first level of "Components" and "Get Started". If it starts to cause too much headache navigating the individual components, we can start to think about how we would structure that, but for now I think tis better being flat.

jbalsas:
> [...] My thought is that a dev would see the design they need to implement for their app, head to clayui.com to find the right component, and then only occasionally go to lexicon for a specific definition. Would I be wrong to assume this?

I think that's a valid hypothesis. We should then make it extremely easy to find the right component for a given design 😉

drakonux:
> I think that's a valid hypothesis. We should then make it extremely easy to find the right component for a given design 😉

Currently, we try to help users to connect with the technical doc in Clay (disclaimer: we know we have a lot of links missed). Maybe you can follow the same pattern in Clay to link us. 

Clay link

> Totally agree, which is why I think it's still helpful to have the first level of "Components" and "Get Started". If it starts to cause too much headache navigating the individual components, we can start to think about how we would structure that, but for now, I think it's better being flat.

Well, I have to admit that I understood the idea was to remove hierarchies just for the components within Components like **navigation** and **forms**. So, I totally agree with you, the first level is needed (Get Started, Foundations, Components, CSS Framework, Migration)

Just to give you more ideas to increase components visibility. A classic rule for structuring navigation on a web page is to create on top a global-navigation and use the left-side panel for local navigation. But you would need to rethink how the whole site is structured. 

Global-Local-Navigation

matuzalemsteles:
> What do you mean by "sessions" here?

@bryceosterhaus at the end I was referring to the hierarchy of structure of the documents in the sidebar for navigation, it got more confusing 😅?

> I'd be curious how people use the two docs. Do users typically go to clayui.com first to find what they need, or do they go to lexicon first?

Just to add, I have seen cases here in Brazil of designers always accessing the clayui.com website to find out if the component was being covered, so they always came to complain about the inconsistencies. Today I don't know anymore, because we only have 1 design in Brazil now 😬

> Currently, we try to help users to connect with the technical doc in Clay (disclaimer: we know we have a lot of links missed). Maybe you can follow the same pattern in Clay to link us.

@drakonux we do this in all of our documents using the alerts on the CSS pages, I thought we had the components in React too 🤔, do you think this is too erased?

> Just to give you more ideas to increase components visibility. A classic rule for structuring navigation on a web page is to create on top a global-navigation and use the left-side panel for local navigation. But you would need to rethink how the whole site is structured.

I really like that, the global navigation bar is a great way to link things quickly, clearly and directly in my opinion.

bryceosterhaus:
> @drakonux 
> Maybe you can follow the same pattern in Clay to link us.

I will re-add this in. I think before we had it at the top but that got removed somewhere... This is what it looked like.
Screen Shot 2020-04-16 at 3 54 54 PM



**Conclusion?**
So overall consensus seems like we move forward with this change and continue to iterate on it for readability in the README file of each component, co-locating demos with components, and continue to look into more improvements for navigation on the site. Anything I miss in that?

bryceosterhaus:
Going to merge this tomorrow unless there are any last concerns.

jbalsas:
moral_support

drakonux:
Sorry I didn't read your mentions on time. But I would like to answer anyway.

> @drakonux we do this in all of our documents using the alerts on the CSS pages, I thought we had the components in React too 🤔, do you think this is too erased?

image

**My personal opinion about the current usage of alerts in CSS pages is that maybe it is causing just the opposite effect I would like to (reading the alert and capture the user's attention). **A similar behavior happens when bold font-weight is used for long sentences in a paragraph. It causes the opposite effect. Bold text should be used for keywords, highlight or emphasize something. If all is relevant nothing it is.**** 

> I will re-add this in. I think before we had it at the top but that got removed somewhere... This is what it looked like.

For me, it's ok Bryce 👍. I don't know if the placement of the label within the React component Tab works well hierarchically but it works better than using Alerts. Also, I have doubts about the naming. We use "_View in Clay_", I don't know if "_View in Lexicon_" would be more appropriate, but I don't have a strong opinion about it.

PR #3126, opened 2020-04-14
chore(clayui.com): Remove titles from Table CSS/Markup examples from various elements

### LINKED ISSUE #2201: Inconsistent use of Tooltip on Table according to Lexicon
In [this](https://clayui.com/docs/components/table/table.html) section that has the examples of the Table component on clayui.com the "Title" field of each row shows a tooltip while hovering on it.

The "Dos and Don'ts" section of the "[Popovers and Tooltips](https://liferay.design/lexicon/core-components/popovers-tooltips/)" page in Lexicon's documentation reports "Use tooltips on elements that trigger an action, such as button icons." and the provided example discourages an use like the one that is present on clayui.com, since it's not an element that triggers any action.

In my opinion the guideline provided by Lexicon is incorrect because a tooltip should show a hint wherever the informations directly provided by the element aren't enough, like in the trimmed title in the table.

Anyway, since Clay should implement Lexicon's guidelines I'm reporting this here so you might raise this doubt to the Lexicon's team.

kresimir-coko:
Fixes #2201 

I believe this is what the requirements of the issue are, just removing all of the non-necessary titles from various table cells, like `title` which should never need it, but leaving titles for long values like `Wings eu, pumpkin spice robusta, kopi-luwak mocha caffeine froth grounds`. If I've misunderstood the issue feel free to clarify.

### DIFF
--- a/clayui.com/content/docs/components/css-tables.md
+++ b/clayui.com/content/docs/components/css-tables.md
@@ -58,7 +58,7 @@ A table is styled like a list. The active state can be invoked by adding class `
                             </svg>
                         </a>
                     </span><a class="inline-item text-truncate-inline" href="#1">
-                        <span class="text-truncate" title="ID">ID</span><span class="inline-item inline-item-after">
+                        <span class="text-truncate">ID</span><span class="inline-item inline-item-after">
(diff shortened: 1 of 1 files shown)


pat270:
@bryceosterhaus @kresimir-coko I'm good with this.

PR #3127, opened 2020-04-14
chore: update changelogs

### LINKED ISSUE #3118: Is there an easy way to find a list of releases for a Clay component?
At the time of this issue, there are 500 releases ordered by date released. https://github.com/liferay/clay/tags also is ordered by release date and we have a tag for each component@version. It would be nice to have a curated list of releases by component if it doesn't exist already.

bryceosterhaus:
Fixes #3118

In order for the changelog to properly generate I needed to point all the "repository" fields in our package.json to the root of the repo. I also went through and updated broken and stale links.

@wincent I was able to use your tool from https://github.com/liferay/liferay-frontend-guidelines/blob/06af3e3ec8e24760e7b85e2f1ab1a43fe54236ee/support/checkLinks.js and it seemed to work really well. It didn't properly infer a few `# header` links in markdown files, but it worked well other than that. Any plans to turn this into its own package to be used via cli or npm scripts? I could see it being helpful across multiple projects.


bryceosterhaus:
@wincent. adding mention here since I had to edit your mention into the original message and gh doesnt like notifications from edits.

wincent:
> @wincent I was able to use your tool from https://github.com/liferay/liferay-frontend-guidelines/blob/06af3e3ec8e24760e7b85e2f1ab1a43fe54236ee/support/checkLinks.js and it seemed to work really well. It didn't properly infer a few `# header` links in markdown files, but it worked well other than that. Any plans to turn this into its own package to be used via cli or npm scripts? I could see it being helpful across multiple projects.

Generally with these little tools I like to battle-test them in one place to iron out the bugs for a while before trying to generalize them, but if it (mostly) worked here too, then perhaps that is a sign that we should move it over to liferay-npm-tools as a new package soon. Created [an issue for that](https://github.com/liferay/liferay-frontend-guidelines/issues/152).

If you have any recollection or record of *which* header links it didn't infer correctly, can you share them over there? Otherwise we can just try running it on Clay again when we extract the package and deal with any issues that crop up.

bryceosterhaus:
@wincent I noted it at https://github.com/liferay/liferay-frontend-guidelines/issues/152#issuecomment-613594071

pat270:
@bryceosterhaus nice this is much better.

PR #3130, opened 2020-04-15
feat: enable build stats in the action log

### LINKED ISSUE #2965: Add CI analysis for package size
Stems from https://github.com/liferay/clay/issues/2918

bryceosterhaus:
fixes #2965

@wincent, I went with your idea and I think its a step in the right direction. I have both a [failing example](https://github.com/bryceosterhaus/clay/pull/47) and [passing example](https://github.com/bryceosterhaus/clay/pull/48) on my fork. I have it set so that it fails if the total size difference is greater than 5%. 

Curious to hear other opinions.


.github/workflows/main.yml
```diff
@@ -1,12 +1,11 @@
...
 on:
     pull_request:
-        branches-ignore:
-            - '**'
+        branches: ['master']
 
 jobs:
-    build_and_comment:
+    generate_stats:
+        name: Generate Build Stats
```
jbalsas, line 8: I would be a bit more specific with the name here. Right now, seeing the failure does not immediately point to the problem:

Screen Shot 2020-04-16 at 09 47 20

If we're simply using this job to `checkSize`, I would go with something like "Bundle Size" that clearly hints what's being tested there...

jbalsas:
awesome

scripts/check-size/generateTable.js
```diff
@@ -0,0 +1,89 @@
...
+	const icon = bytesDiff > 0 ? '⬆️' : '⬇';
+
+	return `${icon} ${convertBytes(bytesDiff)} (${diffPercent})`;
+};
+
+const padValue = (val, length) => {
+	const trailingSpaces = Math.max(length - 1 - val.length, 1);
+
+	return SPACE + val + SPACE.repeat(trailingSpaces);
+};
```
wincent, line 40: Can't you just use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd here?

scripts/check-size/generateTable.js
```diff
@@ -0,0 +1,89 @@
...
+		prevSize: prevStats[dirName],
+	}));
+
+	return [
+		packageStats,
+		Math.round(((newTotal - prevTotal) / prevTotal) * 100),
+	];
+};
+
+function processMarkdownTable(prevStats, newStats) {
```
wincent, line 72: Seems the function name and the file name should match, so I'd change one or the other.

scripts/check-size/index.js
```diff
@@ -69,12 +71,14 @@ function main() {
...
-				JSON.stringify({
-					body: processMarkdownTable(prevStats, newStats),
-				})
-			);
+			const [table, totalDiff] = generateTable(prevStats, newStats);
+
+			// eslint-disable-next-line
+			console.log(table);
+
+			if (Math.abs(totalDiff) > THRESHHOLD) {
```
wincent: Is the rationale of the `abs` here that any big change is suspicious, including when sizes go down? (Just checking.)

wincent:
This is great. The only thing I am not clear on is the `'../../base-stats/.parcel-ci-build.json'` path: if you run this locally, I imagine it blows up because that path won't exist, right? But you could stick it in a tmp directory and have it work easily enough in both CI and locally, I suspect (eg. using `require('os').tmpdir()`).

wincent:
> But you could stick it in a tmp directory and have it work easily enough in both CI and locally, I suspect

Ah, looking at it now, I see that you're doing the thing where you upload the file and then download it, so that suggestion won't work. 😢 

There might be some way to get `github.event.pull_request.head.sha` and `github.event.pull_request.base.sha` via environment variables and just do all this in a single invocation without the uploading/downloading (ie. you'd call once with `--compare` and the script would do the first build, then checkout the other rev, build again, then report), but that is probably an idea to consider another day.

You still might want to make this work locally though by using a different path in the case that `process.env.GITHUB_ACTIONS` isn't set.

scripts/check-size/index.js
bryceosterhaus: Yeah for now the idea is to fail on major changes, at least to grab the attention of the reviewer. We can hone this in more later as we evaluate how we use this. In the future I am thinking we can add per-package threshold or a user-defined limit.

bryceosterhaus:
note that this PR's action will break since when it goes to checkout master, the code isn't up to date.

.github/workflows/main.yml
```diff
@@ -20,10 +20,15 @@ jobs:
               run: |
                   yarn install
                   yarn checkSize
+            - name: Clean directory
+              run: |
+                  git clean -ffdx --exclude /tmp
```
wincent, line 25: Oh, didn't know you could do double `-f`

Niiiice....

bryceosterhaus: I copied this from the `actions/checkout` [docs](https://github.com/actions/checkout#usage). Since all we wanted to do was add the `--exclude`, i just disabled clean and copied their code.

bryceosterhaus:
@wincent added border at the bottom and I swapped out the emoji for a unicode arrow. I _think_ it should work in the raw logs now. Even if it doesnt work in the raw logs though, I think it should be fine since typically I think users will be looking at the embedded rich logs.

Going to merge once this passes tests and we can continue to add anything needed

Issue #3123, opened 2020-04-13
ClayPanel Header renderProps


diegonvs:
## What is your proposal?

There are at least two specific cases where ClayPanel is being discarded to be used due to a limitation of the component to just add a title and not append more items to the Panel's header

Here is the examples: 

#### On App Builder:
![App Builder](https://user-images.githubusercontent.com/7663513/79128731-67d95580-7d7a-11ea-9129-72a7d90f5e64.png)

#### On DataEngine:
<img width="572" alt="Screen Shot 2020-04-13 at 11 32 21" src="https://user-images.githubusercontent.com/7663513/79128798-7b84bc00-7d7a-11ea-8bd4-4869fc6aff55.png">

I would suggest to add a property on ClayPanel component for using renderProps pattern as:

```tsx
<ClayPanel
    collapsible
    renderHeader={
        ({isExpanded}) =>
            (<> {Rule Name 1} <MyDropdown /> <ClayIcon symbol={isExpanded? "angle-down" ? "angle-up")}
        }
/>
    <MyContent />
</ClayPanel>
```

## Why would adopting this proposal be beneficial?
Allow people to have more flexibility on ClayPanel :)

## What are the alternatives to this proposal?
None

/cc @kevenleone

kevenleone:
This feature will help a lot here on App Builder. 
This way we could remove the manual adaptation to the first image be possible directly on ClayPanel.

Great idea @diegonvs

bryceosterhaus:
Hmm yeah this makes sense. This seems like another component where we went sort of in-between high and low level. I think we can make this more low level to provide more flexibility.  We made a similar change in Label recently, https://github.com/liferay/clay/commit/deffd0e09a4a1a274eba4077c6458dd28e62e3ff.

kresimir-coko:
Hey @bryceosterhaus I've done a little bit of research on this task today and here's some questions I came up with:

1. Do we want both `ItemBefore` and `ItemAfter` like in Label?
2. Do we need to make a wrapper for [immediate children](https://github.com/liferay/clay/blob/master/packages/clay-panel/src/index.tsx#L106-L137) of the root component like `ItemExpand` in Label?
3. This prop is supposed to be consumed inside the parent component, not the `Header`, correct? I saw `Header` mentioned but in the 1st image the additional content is inside the element that handles collapsing. In the code example provided by @diegonvs it's getting passed to the root component confirming my theory.
[...]

PR #3133, opened 2020-04-16
feat(@clayui/panel): Add ClayPanelTitle

kresimir-coko:
Hey @bryceosterhaus here's the PR that addresses #3123.

Not much to note, after some brainstorming with myself it was very simple due to the answers you gave to my questions. Unless I made it too simple in a way it doesn't do what we wanted it to do haha.

I can open an issue for adding documentation for this and work on that tomorrow, alongside something else.

### DIFF
--- a/packages/clay-panel/src/index.tsx
+++ b/packages/clay-panel/src/index.tsx
@@ -93,7 +93,7 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 		>
 			{!collapsable && (
 				<>
-					{React.isValidElement(displayTitle) ? (
+					{displayTitle && React.isValidElement(displayTitle) ? (
 						displayTitle
 					) : (
 						<ClayPanelHeader>
@@ -120,7 +120,7 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 						onClick={handleClickToggler}
 						role="tab"
 					>
-						{React.isValidElement(displayTitle) ? (
+						{displayTitle && React.isValidElement(displayTitle) ? (
 							displayTitle
 						) : (
 							<span className="panel-title">{displayTitle}</span>
--- a/packages/clay-panel/src/index.tsx
+++ b/packages/clay-panel/src/index.tsx
@@ -93,13 +93,16 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 		>
 			{!collapsable && (
 				<>
-					{displayTitle && React.isValidElement(displayTitle) ? (
-						displayTitle
-					) : (
-						<ClayPanelHeader>
-							<span className="panel-title">{displayTitle}</span>
-						</ClayPanelHeader>
-					)}
+					{displayTitle &&
+						(React.isValidElement(displayTitle) ? (
+							displayTitle
+						) : (
+							<ClayPanelHeader>
+								<span className="panel-title">
+									{displayTitle}
+								</span>
+							</ClayPanelHeader>
+						))}
 
 					{children}
 				</>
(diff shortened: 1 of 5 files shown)


packages/clay-panel/src/index.tsx
```diff
@@ -30,9 +31,9 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	defaultExpanded?: boolean;
 
 	/**
-	 * Add text to display in Panel Title.
+	 * Content to display in Panel Title.
 	 */
-	displayTitle?: React.ReactText;
+	displayTitle?: any;
```
bryceosterhaus, line 36: Lets use `React.ReactNode` instead of `any`

packages/clay-panel/src/index.tsx
```diff
@@ -91,10 +93,13 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 		>
 			{!collapsable && (
 				<>
-					{displayTitle && (
+					{typeof displayTitle === 'string' ||
```
bryceosterhaus, line 96: I think react offers a helper function called `isReactElement`, so instead of checking for string/number you could do something like

```jsx
{React.isValidElement(displayTitle) ? (
	displayTitle
) : (
	<ClayPanelHeader>
		<span className="panel-title">{displayTitle}</span>
	</ClayPanelHeader>
)}
```

I _think_ this should work...

packages/clay-panel/src/index.tsx
```diff
@@ -116,7 +121,12 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 						onClick={handleClickToggler}
 						role="tab"
 					>
-						<span className="panel-title">{displayTitle}</span>
+						{typeof displayTitle === 'string' ||
```
bryceosterhaus, line 124: Same as above

packages/clay-panel/src/index.tsx
kresimir-coko: It's [React.isValidElement](https://reactjs.org/docs/react-api.html#isvalidelement)

kresimir-coko:
Hey @bryceosterhaus I've pushed a commit on top with the requested changes so you don't have to go through the whole thing again. Feel free to squash them for cleaner merging if you want.

packages/clay-panel/src/index.tsx
```diff
@@ -91,7 +93,9 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 		>
 			{!collapsable && (
 				<>
-					{displayTitle && (
```
bryceosterhaus, line 94: Only thing here is that we also need to check first if `displayTitle` even exists. Right now if it doesn't exist, we will still be rendering this...

```jsx
<ClayPanelHeader>
     <span className="panel-title"></span>
</ClayPanelHeader>
```

Previously if we didn't pass in displayTitle, it wouldn't render anything.

kresimir-coko:
Hey @bryceosterhaus I had some issues with git so the commit listing might be a bit different, but I added the 2 new ones (sanity check and test) as separate so in any case you don't have to check the first 2 commits (ones from last week).

> also ensure our previous API results in the same markup

This line was kinda hard to interpret, seeing as I didn't change any of the old API, but just added new stuff, so I made sure all the tests passed. I assume it meant _make sure all old snapshots match_.

packages/clay-panel/src/index.tsx
bryceosterhaus: @kresimir-coko we can't just add `displayTitle &&` to the existing expression since we are using a ternary operator. We need to wrap the whole expression and only evaluate if `displayTitle` exists. Otherwise, we break the existing API since the existing API without `displayTitle` doesn't render anything for the header. If merge this change, those components would now render an empty element

```jsx
<ClayPanelHeader>
	<span className="panel-title"></span>
</ClayPanelHeader>
```

kresimir-coko:
Hey @bryceosterhaus I was too hasty with that yesterday and completely breezed over that sanity check thing, I pushed a commit that wraps it and a test you requested

diegonvs:
Great job @kresimir-coko! Thanks! @kevenleone

PR #3137, opened 2020-04-17
GH Actions Test

bryceosterhaus:
disregard this. wanted to verify and test the actions that were just merged to master


bryceosterhaus:
One thing to note is that this isn't actually an entirely accurate representation of what adding a dependency would do. Its essentially adding the dependency multiple times to the `size` since we bundle each package individually. We might want to consider adding an "all" bundle which would be all clay packages together in a single bundle, this would give a more accurate picture of the total size than adding all the bundles together.

Issue #3111, opened 2020-04-01
Add iframe onload handler to ClayModal API


markocikos:
Proposal:
```js
<ClayModal
	onIframeLoad={doSomething}
>
</ClayModal>
```
or 
```js
const {observer} = useModal({
	onIframeLoad={doSomething},
})
```

I'm working on [Update Modal Patterns to match Lexicon Definition](https://issues.liferay.com/browse/LPS-101984). I have now encountered two cases where this kind of a handler would be useful:
1. We need to [add a loading indicator to modal](https://issues.liferay.com/browse/LPS-111118). This loading indicator would start when ClayModal containing iframe is rendered, and stop when iframe content is loaded.
1. Currently, ClayModal hardcodes iframe height to 160px, see [here](https://github.com/liferay/clay/blob/2f5b60bb897cf1c6d4530af8167921a58d563a1d/packages/clay-css/src/scss/components/_modals.scss#L169). This cuts off modal content, for example in `fragments-web`. Ideally, we would want height to have the same size as content. I don't think there is a CSS way to achieve this. It can be done with JS, but we need to hook onto this load event.

Alternatives:
1. Maybe there is a way to reference iframe from outside component? A DOM query after render to bind a listener would be a race condition and ugly overall. ClayModal exposing iframe `ref` also seems bad. 
1. Maybe there is a CSS way to achieve autoheight?

/cc @jbalsas

jbalsas:
Hey @laugardie, @drakonux, could you take a look at this from a UX perspective? What would be the desired default behaviour?

I think @marcoscv-work and @eduardoallegrini worked on some CSS solution for the Modals heigh/scroll... can you guys dig that up?

IMHO... a height-changing modal is a terrible experience. Unless done with a "cool" transition (which soon gets repetitive and unnecessary...), flickering elements give systems a clunky and slow feeling. One of the "best" improvements we added in `7.2` was to make sure that css classes in iframes were properly set up at the beginning to avoid FOUC. Modals will take the same amount of time to load, but by avoiding the flickering, it seemed way faster.

So... I'd suggest to stick with some fixed heights (might be configurable) and make the content scrollable.

Just my 2 cents :)

jbalsas:
Ah, another thing... it feels like modals with iframe should always have the loading indicator... so that might be something worth baking in even if we also provide an external API to hook into the `load` event.

markocikos:
[...]

PR #3139, opened 2020-04-20
feat(@clayui/modal): Add iFrameProps to ModalBody

kresimir-coko:
Hey @bryceosterhaus & @matuzalemsteles is this what we want from the [issue](https://github.com/liferay/clay/issues/3111)? I included a usage example to showcase how I think it will get used. I picked up the pattern from other places that we have on Clay, like ClayTabs etc.


matuzalemsteles:
hey @kresimir-coko and @bryceosterhaus no problem for me! I think we can go ahead with this!

Issue #3142, opened 2020-04-20
Create storybook example for product card


bryceosterhaus:
We need to add a storybook example of this implementation, https://github.com/liferay/clay/pull/2914.

PR #3146, opened 2020-04-21
feat(@clayui/card): Add Product Card story to ClayCard

kresimir-coko:
PR for #3142 

LMK if anything can be improved. There are some parts where I had to use a lot of HTML instead of Clay components, not sure if we can fix that somehow, maybe by adding more small, highly reusable components. For example this excerpt, specifically the `c-mb-3` container, the `card-subtitle`, and some others used in the demo:
```jsx
<div className="c-mb-3">
	<div className="card-subtitle" title="AR385672">
		<span className="text-truncate-inline">
			<span className="text-truncate">{'AR385672'}</span>
		</span>
	</div>

	...
</div>
```

### DIFF
--- a/packages/clay-card/stories/index.tsx
+++ b/packages/clay-card/stories/index.tsx
@@ -38,21 +38,6 @@ const ClayCheckboxWithState = (props: any) => {
 	);
 };
 
-const DropDownWithState = (props: any) => {
-	const [active, setActive] = React.useState(false);
-
-	return (
-		<ClayDropDown
-			active={active}
-			alignmentPosition={5}
-			onActiveChange={newVal => setActive(newVal)}
-			trigger={props.trigger}
-		>
-			{props.children}
-		</ClayDropDown>
-	);
-};
-
 storiesOf('Components|ClayCard', module)
 	.add('ClayCardWithInfo', () => {
 		const [value, setValue] = React.useState<boolean>(false);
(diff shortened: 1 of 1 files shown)


packages/clay-card/stories/index.tsx
```diff
@@ -440,4 +458,159 @@ storiesOf('Components|ClayCard', module)
...
+											'ReallySuperInsanelyJustIncrediblyLongAndTotallyNotPossibleWordButWeAreReallyTryingToCoverAllOurBasesHereJustInCaseSomeoneIsNutsAsPerUsual'
+										}
+									</a>
+								</span>
+							</div>
+
+							<div className="card-text">
+								<s>{'$99.00'}</s>
+								<span>{'$70.00'}</span>
+							</div>
```
bryceosterhaus, line 507: I think you can use `<ClayCard.Description />` for this sections and using its different `displayType`'s available.

packages/clay-card/stories/index.tsx
```diff
@@ -28,13 +30,29 @@ const ClayCheckboxWithState = (props: any) => {
...
 			disabled={false}
+			label={props.label}
 			onChange={() => setValue(val => !val)}
 		>
 			{props.children}
 		</ClayCheckbox>
 	);
 };
 
+const DropDownWithState = (props: any) => {
```
bryceosterhaus, line 41: No need to declare this as a separate component. We can just include this in the story itself.

pat270:
@bryceosterhaus @kresimir-coko Mostly looks good to me. The only thing I noticed is that the Order By: Newest dropdown toggle should have a `caret-bottom` icon

packages/clay-card/stories/index.tsx
kresimir-coko: Aah, yea I missed this because I was searching by `card-title`. I see now that it's implemented as `card-${displayType}`

bryceosterhaus:
@kresimir-coko this looks good to me. Was there anything else you were wanting to add before merging?

kresimir-coko:
> Was there anything else you were wanting to add before merging?

If this was a question for me @bryceosterhaus then the answer is no I can't think of anything that's missing.

PR #3153, opened 2020-04-22
chore: update prettier

### LINKED ISSUE #3048: Update to Prettier v2.0.0
copied from https://github.com/liferay/liferay-npm-tools/issues/418
> Low-ish priority on this one because there may be some bugs that need to get sorted out in the next few patch releases, and this could generate a fair bit of churn:
> 
> https://prettier.io/blog/2020/03/21/2.0.0.html
> 
> But still, something we'll want to do at some point in order to benefit from the bug fixes.

bryceosterhaus:
fixes #3048

Since https://github.com/liferay/liferay-npm-tools/issues/418 is looking to be merged soon I thought we might as well try it here. Looks like no major issues from what I see


.storybook/main.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
 		'@storybook/addon-a11y/register',
 		'@storybook/addon-knobs/register',
 	],
-	webpackFinal: config => {
+	webpackFinal: (config) => {
```
matuzalemsteles, line 20: 😱 Oh no 💔! I loved it when didn't have brackets for those cases.

bryceosterhaus: I know... I also liked not having it but I guess they thought it was more useful to put them mandatory. On the brightside, you should be able to write them without and then auto-format it to having them

matuzalemsteles: Haha yeah, it's a good way to ignore 😅!

wincent: FWIW, they had good reasons for [changing the default](https://prettier.io/blog/2020/03/21/2.0.0.html):

> At first glance, avoiding parentheses in the isolated example above may look like a better choice because it results in less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments, default values, or a variety of other things. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the change.

(You could override the setting, but in liferay-portal and the other repos we're going to accept the new default.)

PR #3166, opened 2020-04-23
chore(release): publish updated packages

bryceosterhaus:
So I realized as I was releasing this next version of clay that we accidentally added a breaking change and version in a [previous release](https://github.com/liferay/clay/commit/3fc84c9320fa652a452b7c2881a35fdc9f4cc17c#diff-bdff5ecdc3ad0325256c5d796068e18fR25-R27). It looks like this might've gotten merged in our initial push to get clay released. 

Now I don't think we should still have this version released, since 4.x isn't intended to be a release yet. So my idea for fixing this is to simply create a new 3.1.0 version that contains this change rather than opting for a 4.x version. Now I know this isn't perfect semver, but I think we can consider this a mistake for the initial API. Since the property change is simple, I don't imagine it being a huge issue. You can see my commit history of how I managed to get lerna to handle the versioning. And then finally, we would remove the 4.0.0 version of table, since we don't have any intention of that existing.

Any thoughts on this approach? Also note that the current 4.0.0 version is specified in liferay-portal, so we would also bring that back to 3.1.0

cc: @jbalsas, @wincent, @matuzalemsteles


bryceosterhaus:
If this sounds good to you guys, I will go ahead and publish this

wincent:
It is not entirely explicit in [SemVer](https://semver.org/), but I don't think it's right to publish 3.0.0-alpha.1 of `@clayui/localized-input` *after* 3.0.0 has been published. It's kind of like time travel back in time. The alpha is "prerelease" and the "pre" means "before" in the temporal sense.

Contrast that with releasing 3-dot-anything of `@clayui/table` *after* 4.0.0. That is fine because conceptually they correspond to two "series" of development (even if they are not literal Git branches, at least not yet). We do this all the time when we release Clay 2.x and 3.x packages independently of one another, while developing both series concurrently; no violation of The Laws of Time Travel.

(Subtle point about this: while major version "series" usually correspond to Git branches, they don't have to. And while the temporal relationship between releases and the shape of the development graph ideally should map sensibly onto one another, they don't always *have* to: case in point, it's "ok" for a "version" field in a `package.json` file to go up and down in a series of commits in order to fix a problem like the one we have here.)

All of which is to say I think the general procedure when we make releases by mistake should be to:

- **Either:** [`npm unpublish`](https://docs.npmjs.com/cli/unpublish) them if they are literally broken/dangerous (not the case here) — disclaimer: this can only happen within 72 hours of releasing, which is another reason not to do it here; also, in my experience, NPM won't let you re-publish a different version immediately after running `npm unpublish`, so you have to be careful to do things in the right order.
- **Or:** Fast-follow with a fixed package with a different version number, and [`npm deprecate`](https://docs.npmjs.com/cli/deprecate) the bad release (if it is broken enough that you want to be sure nobody uses it by mistake).

So, in this concrete case, here's what I would do:

1. Bump `@clayui/localized-input` to 3.0.1-alpha.1 and release.
2. I would say mark `@clayui/localized-input` 3.0.0 as deprecated, but it [looks like you never actually published it](https://www.npmjs.com/package/@clayui/localized-input), so seems there's nothing to do there.
3. Mark `@clayui/table` 4.0.0 as deprecated because it is too late to `unpublish` it.
4. Make sure that `@clayui/table` 3.0.6 has the `latest` tag using `npm dist-tag add @clayui/table@3.0.6 latest` (check this on [versions page](https://www.npmjs.com/package/@clayui/table?activeTab=versions)).
5. Next maintenance update to `@clay/table` can be 3.0.7 or 3.1.0 without any problems.
6. When we eventually release 4.0, because `@clayui/table` 4.0.0 was already released, the first release will actually be 4.0.1 (alpha first, if necessary).

Does that make sense?

matuzalemsteles:
I agree with @wincent, just a question the version 4.0 of `@clayui/table` arrived to enter the master of the Portal? If not, can we revert this publication? if so, we probably have to follow what @wincent recommended and deprecated this version 4.0.0 specifically in NPM, if possible add a message of the reason for not causing confusion that version 4.x will be deprecated.

wincent:
> can we revert this publication

You can't unpublish, because `npm unpublish` only works for 72 hours after publishing, and 4.0.0 was [published to NPM 2 months ago](https://www.npmjs.com/package/@clayui/table?activeTab=versions).

You can of course undo [the change in liferay-portal though](https://github.com/liferay/liferay-portal/commit/04ff649ed63d0c533b9005f57bd8676c4e46d798), if that's what you meant.

bryceosterhaus:
Okay great thats what I was thinking as well. I will go ahead and make that change for `@clayui/table`.

As for localized-input, that package hasn't been tagged or published yet. The reason I moved its version back to alpha.1 is so that `lerna version` would bring it to 3.0.0 for its first publish since we haven't actually released anything for it. Otherwise, if we don't make this commit it will make the initial version as 3.1.0 So I don't think that will be an issue, that commit was just more of a hack for a better experience from running `lerna version`. In the future, we should likely get in the habit of publishing new components once they are merged, or label as an alpha.

bryceosterhaus:
updated and sent at https://github.com/liferay/clay/pull/3173

I was able to keep the commit history simple by making changes to the file without commiting and then running lerna

Issue #2918, opened 2020-02-11
Performance Budget Strategy


jbalsas:
Just an RFC for now to explore what we want to do regarding Perf. Budgets. As we keep adding more and more components to the mix, it's important that we look at how that affects our final output so we keep the impact of using our library is contained.

We currently have 2 main deliverables in Clay:
- **@clayui/css**: It's "statically" added through a css link but accounts to most of the css we load nowadays. Need to keep it in check and try to trim it down if possible
- **@clayui components**: Mostly loaded on demand, but need to check how it performs on aggregate when most of the components are used.

Some random literature bits:
- [Performance Budgets 101](https://web.dev/performance-budgets-101/)
- [Incorporate performance budgets into your build process](https://web.dev/incorporate-performance-budgets-into-your-build-tools/)
- [Performance Budgets](https://developer.mozilla.org/en-US/docs/Web/Performance/Performance_budgets)
- [Performance Budgets (Keep Request Counts Low And File Sizes Small)](https://developers.google.com/web/tools/lighthouse/audits/budgets)
- [The Cost Of JavaScript In 2018](https://medium.com/@addyosmani/the-cost-of-javascript-in-2018-7d8950fbb5d4)
- [The cost of JavaScript in 2019](https://v8.dev/blog/cost-of-javascript-2019)
- [Can You Afford It?: Real-world Web Performance Budgets](https://infrequently.org/2017/10/can-you-afford-it-real-world-web-performance-budgets/)

bryceosterhaus:
I read through the articles you mentioned and they all seemed to speak of a performance budget in regards to a specific webapp or site. Since Clay is a library, I am wondering what the best way to measure performance is since we can't really use something like lighthouse. Perhaps we could do something similar to material-ui, where they have a CI that analyzes the bundle size on each PR. [example](https://github.com/mui-org/material-ui/pull/19794#issuecomment-589586464)

pat270:
@jbalsas is there a specific size for @clayui/css we are trying to target?

matuzalemsteles:
I think trying to impose a performance budget inside Clay can be a bit much for a library, it would be difficult as Bryce said to measure the size of our packages. We do not deliver our processed packages with minification, dead code removal or more refined optimizations. I think this part is really up to the application to add these steps, since they can apply it in the best way.

[...]

PR #3183, opened 2020-04-28
chore: set hard limits for total size

bryceosterhaus:
In regards to #2918, here is an idea of how to set a hard limit for overall size of Clay. I'm not super stoked with this implementation, but I wanted to get an idea out there first and then see wheat everyone else thought.


jbalsas:
Hey @bryceosterhaus! 

As I've mentioned in the past, we used to do this on Clay2 as well... and other than being able to control the overall bundle size, it provided a [Quick Start](https://v2.clayui.com/docs/get-started/introduction.html) entry point to the library:

```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/clay/lib/css/atlas.css">
<script src="https://cdn.jsdelivr.net/npm/clay/lib/js/clay.js"></script>
```

It also makes it super easy to integrate with some "less-advanced" sandbox tools like [JSFiddle](https://jsfiddle.net/) or [JS Bin](https://jsbin.com/?html,output).

All this to say, I wouldn't necessarily just keep this on the tooling side, but maybe make it part of our official release process and make it available through our [How to Use Clay](https://clayui.com/docs/get-started/how-to-use-clay.html) instructions.

all.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import * as alert from '@clayui/alert';
```
wincent, line 6: Pro-Tip™: These can all be:

```js
export * as alert from '@clayui/alert';
```

(May depend on project settings and dependency versions, but I have been able to do this in other TypeScript projects.)

bryceosterhaus:
> @jbalsas 
> As I've mentioned in the past, we used to do this on Clay2 as well... and other than being able to control the overall bundle size, it provided a Quick Start entry point to the library:

I'm hesitant to do this since I don't want people to just pull it all in at once, especially since `@clayui/charts` nearly doubles the size(billboard & d3). We would also need to provide some sort of wrapper around the components so that they can be instantiated without React. Unless you are thinking people would also pull in the `React` global.

bryceosterhaus:
Made some updates to utilize gh actions rather than tacking on to our unit tests.

See this example of a failing PR https://github.com/bryceosterhaus/clay/pull/50

jbalsas:
> I'm hesitant to do this since I don't want people to just pull it all in at once, especially since @clayui/charts nearly doubles the size(billboard & d3). We would also need to provide some sort of wrapper around the components so that they can be instantiated without React. Unless you are thinking people would also pull in the React global.

That's fine, we don't have a clear demand for this yet, so we can table this for when/if it comes up as a hard requirement.

bryceosterhaus:
@jbalsas sounds good. How do you feel about this current implementation? Is this would you would expect, a hard limit of a certain size and trigger a failure if it exceeds that?

all.js
bryceosterhaus: For some reason it wouldn't let me name them, so I changed them all to 
```js
export * from '@clayui/package';
```

wincent: That's quite a bit different though. That won't export the `default` export if you do that; it's just mashing in all the named exports. Note the difference in [how Babel transpiles the two forms](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAKjgQwM5xHAZlCBbOAcghgAtgpCBuAKBtElkW1wMLIEsA7Ac2qA&debug=false&forceAllTransforms=true&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Ces2016%2Ces2017%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose%2Cenv&prettier=false&targets=&version=7.9.6&externalPlugins=).

> For some reason

I have a feeling it depends on recent enough versions of TypeScript and/or Babel, and possibly also on how they are configured. In liferay-portal, for example, we have Babel [configured](https://github.com/liferay/liferay-npm-tools/blob/b50878235a3bfd187430dea9e12cdca4e03dcfb9/packages/liferay-npm-scripts/src/config/babel.json#L10) to use the `@babel/proposal-export-namespace-from` plugin. (Note: Despite the "proposal" in the name, [this feature made it to Stage 4 and is in the ECMAScript spec at this point](https://github.com/tc39/proposal-export-ns-from)).

bryceosterhaus: Strange, the bundle size after switching from `* as` to `* from` didn't change at all. I'll look into adding that babel config and see if I can get it working.

bryceosterhaus:
going to merge and see how it goes, if we find any issues with it we can continue to iterate.

Issue #3056, opened 2020-03-24
Add parent pages for sidebar


bryceosterhaus:
Currently on the sidebar, we don't have an overview page for each of the sections. I think it would probably be helpful to have short descriptive pages for each of these categories. This would also allow us to have better links, rather than having to go to a specific page right away. For example, if you want to link to the components as a whole, you would likely link straight to Alerts

![Screen Shot 2020-03-24 at 11 57 22 AM](https://user-images.githubusercontent.com/6843530/77465894-a0fb5700-6dc6-11ea-908c-7feab1382622.png)

pat270:
I think this would fix https://github.com/liferay/clay/issues/2793.

matuzalemsteles:
I support 😁, I had done this in the beginning. But they will remove 😞I had an idea at the beginning of using these pages of the session to try to "sell" Clay, explain how it works...

kresimir-coko:
I did some digging around this issue and I couldn't wrap my head around it entirely. I'd like to tackle it but I need some information before I do that.

1. Is the behaviour expected to be the same as [Lexicon's](https://liferay.design/lexicon/foundations/) where the link to the parent page is the same that has the dropdown functionality (eg. clicking on  `Foundation` opens the parent page)?
2. The `order` keyword in Markdown didn't want to cooperate in any of the Markdown files that live in the `components` folder. In case the answer to the question above is negative, it means we want to have a separate page, like `components.md` to serve as the descriptive page of components, but I couldn't order it so it's the first one in the navigation under the `Components` tab
3. If the answer to the 1st question is positive, where is the code that builds the navigation contained? I tried scouring around our codebase but couldn't find it
4. Do these links need to lead to the newly created pages? 
![image](https://user-images.githubusercontent.com/24564752/78018765-bf290000-734e-11ea-99f1-e0ade165a4d7.png)

matuzalemsteles:
hey @kresimir-coko  if I understood Bryce's idea, if we keep the same behavior that I had thinking at the beginning too, we will probably continue like this:

1. Yes, we will follow this behavior.
2. To create a section page we can use the `index.md` files, you can access them via `index.html`, they are already being created but the navigation menu is not linking to it. For example https://clayui.com/docs/components/index.html
3. If you are looking for the navigation menu it is at https://github.com/liferay/clay/tree/master/clayui.com/src/components/Sidebar.
4. Yeah, once we have those sessions, we should aim there.

bryceosterhaus:
[...]

PR #3186, opened 2020-04-29
feat(clayui.com): Add ability to click on parent pages in Side Navigation

kresimir-coko:
Hey @bryceosterhaus this is my attempt to tackle #3056.

A couple things to note:
- I have issues with Prettier, i guess. It keeps removing the parentheses `(prop)` and I saw that you added new formatting rules which added those in. So not sure how to fix that one.
- I'm unsure if it's the problem with DOM structure, or something in JavaScript, but I couldn't find a way to separate the link from the collapse icon, `Compontens     >` is a clickable link, but when trying to split them the icon `>` would become a mostly empty 200x200 square element and it would lose the icon rotation feature. _So please let me know if you have any thoughts on this?_
- I removed the `alwaysActive` property on all docs pages to clean up the Navigation so it behaves like Lexicon's, I'm not married to the idea so if you liked it better before we can ditch my proposition easily.
- I added some copy to the parent pages, with the `Get Started` one being different, I removed the `what-is-clay` page and moved that content to the index page for `Get Started`, as I didn't figure out what would be better to put into the landing page of that section

I'm marking it as a Draft PR because it is not a full-fledged solution to the problem but a milestone with some possible talking points and a question in there 🚧


wincent:
> * I have issues with Prettier, i guess. It keeps removing the parentheses `(prop)` and I saw that you added new formatting rules which added those in. So not sure how to fix that one.

I see two versions of Prettier in the yarn.lock:

```
prettier@^1.16.4:
  version "1.17.1"
  resolved "https://registry.yarnpkg.com/prettier/-/prettier-1.17.1.tgz#ed64b4e93e370cb8a25b9ef7fef3e4fd1c0995db"
  integrity sha512-TzGRNvuUSmPgwivDqkZ9tM/qTGW9hqDKWOE9YHiyQdixlKbv7kvEqsmDPrcHJTKwthU774TQwZXVtaQ/mMsvjg==

prettier@^2.0.5:
  version "2.0.5"
  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.0.5.tgz#d6d56282455243f2f92cc1716692c08aa31522d4"
  integrity sha512-7PtVymN48hGcO4fGjybyBSIWDsLU4H4XlvOHfq91pz9kkGlonzwTfYkaIEwiRg/dAJF9YlbsduBAgtYLi+8cFg==
```

Looks like [the update](https://github.com/liferay/clay/commit/f3e9914d91d5aad7c5dcc6338598b507436be8ff) was bad.

wincent:
Output of `yarn why prettier`:

```
=> Found "prettier@2.0.5"
info Has been hoisted to "prettier"
info Reasons this module exists
   - "workspace-aggregator-d134ba26-417f-4e0f-af0c-33622e7081ee" depends on it
   - Specified in "devDependencies"
   - Hoisted from "_project_#prettier"
info Disk size without dependencies: "10.24MB"
info Disk size with unique dependencies: "10.24MB"
info Disk size with transitive dependencies: "10.24MB"
info Number of shared dependencies: 0
=> Found "clayui.com#prettier@1.17.1"
info This module exists because "_project_#clayui.com" depends on it.
info Disk size without dependencies: "8.91MB"
info Disk size with unique dependencies: "8.91MB"
info Disk size with transitive dependencies: "8.91MB"
info Number of shared dependencies: 0
=> Found "@storybook/addon-storysource#prettier@1.17.1"
info This module exists because "_project_#@storybook#addon-storysource" depends on it.
info Disk size without dependencies: "8.91MB"
info Disk size with unique dependencies: "8.91MB"
info Disk size with transitive dependencies: "8.91MB"
info Number of shared dependencies: 0
=> Found "@storybook/source-loader#prettier@1.17.1"
info This module exists because "_project_#@storybook#addon-storysource#@storybook#source-loader" depends on it.
info Disk size without dependencies: "8.91MB"
info Disk size with unique dependencies: "8.91MB"
info Disk size with transitive dependencies: "8.91MB"
info Number of shared dependencies: 0
```

It's possible that if you do an `rm -r node_modules` and then re-install with `yarn`, it may choose to put 2.0.5 at the top level for you.

kresimir-coko:
> It's possible that if you do an rm -r node_modules and then re-install with yarn, it may choose to put 2.0.5 at the top level for you.

@wincent I did `git clean -dfx` followed by the regular Clay build setup which should get rid of `node_modules` if I understood the command correctly, but to no avail.

wincent:
I am sorry for your loss... for me, `yarn run prettier --version` in Clay's `master` prints 2.0.5... But the only robust fix will be to eliminate the duplicate dependency. (Worst case scenario, with a `resolutions` hack in the package.json.)

bryceosterhaus:
> - I have issues with Prettier, i guess. It keeps removing the parentheses `(prop)` and I saw that you added new formatting rules which added those in. So not sure how to fix that one.

Hmm this seems odd. My guess is that your editor might be configured for a certain version of prettier rather than picking it up from your node_modules. But you also likely have to follow what Greg had said and remove the old version from the yarn.lock.

> - I'm unsure if it's the problem with DOM structure, or something in JavaScript, but I couldn't find a way to separate the link from the collapse icon, `Compontens     >` is a clickable link, but when trying to split them the icon `>` would become a mostly empty 200x200 square element and it would lose the icon rotation feature. _So please let me know if you have any thoughts on this?_

Yeah I think we should separate the two so that you are able to click the dropdown without having to select that page. As of now, if you expand/collapse a category, it also selects that page would could be odd. Let me look into this a bit and ill get back to you on this PR

> - I removed the `alwaysActive` property on all docs pages to clean up the Navigation so it behaves like Lexicon's, I'm not married to the idea so if you liked it better before we can ditch my proposition easily.

What exactly did this do when it was enabled?

I also noticed that when a parent page was selected there was no indicator in the nav showing it was selected. I think we could add in the same active color that we used for the child pages.

bryceosterhaus:
@kresimir-coko check out this commit, I was playing around with trying to make the collapse portion clickable. https://github.com/bryceosterhaus/clay/commit/d82822f228b880fb5666061796669124910007df

kresimir-coko:
Damn @bryceosterhaus I was so close to your solution, which is working like a charm. I think all that I was missing was the change you made with `event.preventDefault()` moving it to the SVG click event. And I see you removed the `#openNav` which I was trying to keep.

> What exactly did this (`alwaysActive`) do when it was enabled?

It made some of the collapsable navigation groups stay always open, like Components for example. It made for some wacky interactions, or I just started having hallucinations since I spent like 5 hours clicking on the navigation.

### Style concerns
Hey @pat270 I added styles for `nav-link` inside an `active nav-heading` with the goal of signaling the user when they're on the parent page, but that also applies styles to the parent page (e.g. Components) when one of it's children (e.g. Breadcrumbs) is active because it retains the `active` class. If we dislike that behaviour we can figure out how to change it later.

### Final thoughts
I think this is a complete solution with just the content of the new pages being a bit on the weaker side, but I guess we can get someone to provide us with some nice copy there. I've also condensed the commits and resent everything because I had to do some more changes in the files I sent yesterday.
Local CI complained about the following but passed without errors:
```
ts-jest[jest-transformer] (WARN) Got a unknown file type to compile 
(file: /Users/kresimircoko/Documents/Work/Projects/clay/packages/clay-charts/stories/static/map.json). 
To fix this, in your Jest config change the 'transform' key which value is 'ts-jest' so that it does not match this kind of files anymore.
```

bryceosterhaus:
One UI issue I noticed is that we have multiple pages shown as active. For this case, only the page that the user is on should have the active color and not the parent page nav. Once we get that addressed I think its ready to merge.

Screen Shot 2020-04-30 at 10 22 56 AM


I can also follow up this PR and update some of the copy as needed.

bryceosterhaus:
Also, I would expect this page to still show as active since its the page shown, even though it's collapsed.
Screen Shot 2020-04-30 at 10 33 33 AM

kresimir-coko:
Hey @bryceosterhaus so after 2 full days of messing around with this, almost rewriting it into a Functional Component at some point, it turns out it was a 3-line change. :shrug: But here it is, the latest commit is the one that fixes the changes you requested.

Let me know if this can be improved in any other way.

matuzalemsteles:
hey guys, just one thing that i think would be interesting to keep, at least the Components session "always open", I think it is much better when the user enters the document directly and sees the list of components right away, it seems like a more experience friendly to find things. 

This shows what we want to deliver to them right away and then the user can move smoothly through the sessions.

kresimir-coko:
@bryceosterhaus & @matuzalemsteles I've added another commit on top that puts `alwaysActive` back on the Components parent page.

Issue #3149, opened 2020-04-21
Removing Bootstrap 4 Dependency


pat270:
Integrating Bootstrap in DXP has caused some issues for developers wanting to create custom sites based on Bootstrap. DXP's admin controls are tightly coupled with Bootstrap and presents a problem for devs that want to style Bootstrap's components directly without affecting admin control styles. See https://github.com/liferay/clay/issues/1104.

There is no solution to this issue because of the global nature of CSS. We can help alleviate some of the CSS bleed by namespacing all of our components and removing our dependency on Bootstrap.

Now that Bootstrap 4 is mature, we can begin this process by merging Clay CSS components with Bootstrap's. Clay CSS duplicates a lot of properties declared in Bootstrap because there is no way to customize them through Sass variables. Merging will give us more control over the component and will reduce the number of duplicate properties. A preliminary list of components we should take over:
- [x] import Bootstrap components separately #3261
- [x] move components `aspect-ratio`, `icons`, `buttons`, and `type` earlier in import chain #3263
- [x] root #3273
- [x] reboot #3271
- [x] type #3276
- [x] images #3277
- [x] code #3284
- [x] grid #3285
- [x] tables #3291
- [x] forms #3297
- [x] buttons #3312
- [x] button-group #3438
- [x] transitions #3318
- [x] dropdown #3322
- [x] input-group #3379
- [x] custom-forms #3438
- [x] nav #3381
- [x] navbar #3438
- [x] card #3382
- [x] breadcrumb #3438
- [x] pagination #3384
- [x] badge #3438
- [x] jumbotron #3386
- [x] alert #3438
- [x] progress #3388
- [x] media #3438
- [x] list-group #3390
- [x] close #3438
- [x] toasts #3392
- [x] modal #3438
- [x] tooltip #3394
- [x] popover #3438
- [x] carousel #3396
- [x] spinners #3438
- [x] utilities #3438
- [x] print #3398
- [x] bootstrap mixins #3562
- [x] bootstrap variables #3902

Once a component has been fully adopted, we can theoretically begin namespacing the components with Sass `@extend`. Take `.btn` for example:

```scss
$enable-old-classnames: true;

%c-btn {
  ...css declarations
}

.c-btn {
	@extend %c-btn;
}

@if ($enable-old-classnames) {
  .btn {
    @extend %c-btn;
  }
}
```

[...]

PR #3198, opened 2020-05-04
Remove Bootstrap 4 CSS dependency

pat270:
This is a partial draft of my plan for absorbing Bootstrap 4 into Clay CSS. I've absorbed Bootstrap's `_reboot`, `_type`, `_images`, `_code`, `_grid`, and `_buttons` into Clay CSS here. From my preliminary calculations (not zipped and unminified) this saves 23kb and 11kb in Atlas and Base respectively. Zipped and unminified saves 3kb and 1kb from Atlas and Base respectively.

I've split up `_reboot` and placed the resets in each component. I wasn't sure about this. Is it better to have a browser reset file like Bootstrap or apply the reset if you import the component? I chose the latter because I didn't need to repeat selectors.

I created new variables with old variables mapped to it. The new variables are prefixed with `c-` in the Buttons commit. Prefixing seems the best way to go forward so we don't cause collisions for people that have used or created variables.

I want to use mixins to generate all styles but ran into a problem with the best way to describe a CSS rule-set with a variable. For example the variable to modifiy the selector `.btn-block + .btn-block` is:
```scss
// .btn-block + .btn-block
$c-btn-block-c-btn-block: ();
```
I've toyed with `$c-btn-block\+c-btn-block` but this doesn't sit right with me.

Base element rule-sets will have no `c-` prefix (e.g., `html`, `kbd`, `button`). They will be named `$html`, `$kbd`, `$button` respectively.


marcoscv-work:
@pat270 ! Really big and great change!

From my point of view Import Bootstrap 4 components independently is something really good, It's something I'm always doing this in my personal projects and we even recommended to do this on some customer project focused on very light landings pages

I think it also can help to make more visible our dependencies with bootstrap and reduce weight as you already did by absorbing files

If we have time I would like to save time to take a deep look into it.

What about the failed tests?

pat270:
Thanks @marcoscv-work, this was just a test to show what needs to be done and discuss semantics. I'm going to split these up into smaller pulls so it's easier to review and catch bugs.

packages/clay-css/src/scss/components/_buttons.scss
```diff
@@ -1,3 +1,32 @@
+// Remove the default `border-radius` that macOS Chrome adds.
```
jbalsas: I'm worried about the legality of such changes... we're doing:

- Copy paste `_reboot.scss` which is in fact `Normalize.css` keeping its LICENSE
- Strip our copy for parts and remove all traces of its LICENSE

While both `Bootstrap` and `Normalize.css` are MIT, and of course ianal, this doesn't feel right.

I was initially going to comment just on the first commit and ask why we needed to "absorb" this rather than just adding a direct dependency on `Normalize.css`, then this commit further pushed my concerns...

I know @wincent is partial (me too) to having no-dependencies, specially when they don't pull their own weight, so I'm curious to hear his opinion (and everyone else's) on this...

wincent: You should never ask me to talk about the costs and perils of dependencies... I may go on for several hours!

On the subject of the licensing question, IANALE (I am not a lawyer either), but the MIT is quite clear about the need to preserve the license text in any derivative work.

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

The "substantial" bit is where an expert opinion could weigh in — and of course, whatever that opinion might be, I bet you could find another expert to offer an equally hefty counter-opinion 😂. If in doubt, I just keep the original license.

You might argue that if we fork something hard and then change every single thing about it, then it is no longer a derivative of the original work (see: [https://en.wikipedia.org/wiki/Ship_of_Theseus](Ship of Theseus), but I tend to err on the side of caution. A "clean room" dependency-free from-scratch reset would obviously liberate of this question, but I don't know if we'd bother with that. `Normalize.css` seems bloat-free enough that it's not really going to swell our dependency footprint by much.

pat270: I didn't think about attribution too much. I'm good with using normalize or reboot. I do think Bootstrap's _reboot.scss would be easier since it's already in use.

marcoscv-work:
I think if we finally import BS by parts we should show this improvement and philosophy to the final customer by documenting and adding this information to the official frontend training.

Some quick possibilities it will offer them:

- Restore the imports like previous versions if they find out some incompatibility (I don't think this will happen).
- Easy mode to remove parts they won't need and create light themes, for example creating two versions of themes, a light one for a public site or landing site, and a full one when an admin is logged.
- Possibility to create short pieces of CSS for fragments collections (it could be even automatized in order to in a future remove themes). I have some ideas about this.

And what can this offer for us?:

- Good start to getting away from bootstrap :-P
- Facilities to drop off old components support (we could create maps of dependencies between components/variables/files). in order to create rules, for example: if you remove bootstrap buttons you will need also to remove "xxx files/components"

Last questions!
Do we have an Epic for all this work? should we have some other place to discuss all this?

pat270:
@marcoscv-work no epic, but there is https://github.com/liferay/clay/issues/3149

pat270:
Closing this we can continue discussion on issue above.

PR #3202, opened 2020-05-06
chore(clayui.com): restructure pages and structure

bryceosterhaus:
I noticed that our sidebar structure was a bit odd. I moved a few pages into the "Get Started", added links for contributing and storybook page, and I also removed pages that we weren't using.


bryceosterhaus:
going to merge since there aren't any major concerns.

Issue #2421, opened 2019-09-11
Layout util components


bryceosterhaus:
We might want to look into creating some helper components for people that are creating layouts. I am thinking something like containers, rows, columns, sheets, etc.

jbalsas:
I've added this to the https://github.com/liferay/clay/milestone/19

matuzalemsteles:
hey @jbalsas as comment from https://github.com/liferay/clay/issues/3082#issuecomment-606457343. I will leave some ideas that I have about this.

I follow a design system from Seek that Mark Dalgleish talks a lot about where he works, he presented some of his characteristics and it is very interesting how they approached the creation of primitive components only. They did a great job with the harmonization of API's for the Layout components, in case you are curious https://seek-oss.github.io/braid-design-system/foundations/layout.

I think the main idea that we should take with the creation of these Layout components is also the primitive ones, we can offer `Rows`, `Columns` with APIs that allow people to achieve different spacing and layout flexibility.

A component that caught my attention is the `<Stack />` (https://seek-oss.github.io/braid-design-system/components/Stack), it only handles components vertically and offers API's to handle with the blank spaces between them, this is very interesting because it replaces the use of `Row` and offers layout flexibility. I like this design style, instead of describing it a lot to assemble the layout, the component does a bit of magic underneath but still adding a lot of value and covering cases of layouts that have different spacing.

```jsx
<Stack space="gutter">
  <Placeholder height={40} />
  <Placeholder height={40} />
  <Placeholder height={40} />
</Stack>
```

I think this can help to solve a lot of problems in some layouts, the spacing between components in vertical and horizontal (if I am not mistaken I saw @drakonux saying that they have a space rule every 8px, correct if I am wrong.). These are some shallow thoughts that I have now, I know that there are some other challenges, but in the components in React we can do something to make it all simple and configurable, this may seem more like a high level component.

Ping @eduardoallegrini and @marcoscv-work if that helps.

eduardoallegrini:
I took a look at the design system and I think it's great, and even if it's too soon to decide what we need (since we don't know for sure what we have yet) I will link your comment in the results of the spike we are working on. 

My only concern is that it seems to have some limits, for example: what happens if we need different spaces between the elements inside the container? (but again, I didn't read the whole documentation :) )

matuzalemsteles:
> My only concern is that it seems to have some limits, for example: what happens if we need different spaces between the elements inside the container? (but again, I didn't read the whole documentation :))

[...]

Issue #803, opened 2018-04-17
Autofit Row add more utilities to help ease alignment


pat270:
- `.autofit-padded` allows to add space between columns, we need a way to remove space as well.
- clean up `.autofit-float` pattern

PR #3204, opened 2020-05-07
feat(@clayui/layout): create low-level layout utilities

bryceosterhaus:
Here is my initial idea for the low-level components of row, col, and container.

cc @pat270 can you checkout the demos on storybook and verify these make sense?

### DIFF
--- a/packages/clay-layout/package.json
+++ b/packages/clay-layout/package.json
@@ -1,0 +1,39 @@
+{
+	"name": "@clayui/layout",
+	"version": "3.0.0-alpha.1",
+	"description": "ClayLayout component",
+	"license": "BSD-3-Clause",
+	"repository": "https://github.com/liferay/clay",
+	"engines": {
+		"node": ">=0.12.0",
+		"npm": ">=3.0.0"
+	},
+	"main": "lib/index.js",
+	"types": "lib/index.d.ts",
+	"ts:main": "src/index.tsx",
+	"files": [
+		"lib",
+		"src"
+	],
+	"scripts": {
+		"build": "cross-env NODE_ENV=production babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
+		"build:types": "cross-env NODE_ENV=production tsc --project ./tsconfig.declarations.json",
+		"prepublishOnly": "yarn build && yarn build:types",
+		"test": "jest --config ../../jest.config.js"
+	},
+	"keywords": [
+		"clay",
+		"react"
+	],
+	"dependencies": {
+		"classnames": "^2.2.6"
+	},
+	"peerDependencies": {
+		"@clayui/css": "3.x",
+		"react": "^16.8.1",
+		"react-dom": "^16.8.1"
+	},
+	"browserslist": [
+		"extends browserslist-config-clay"
+	]
+}
--- a/packages/clay-layout/src/Col.tsx
+++ b/packages/clay-layout/src/Col.tsx
@@ -1,0 +1,81 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+type TColSize = boolean | number | 'auto';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+
+	/**
+	 * The number of columns to span on large devices
+	 */
+	lg?: TColSize;
+
+	/**
+	 * The number of columns to span on medium devices
+	 */
+	md?: TColSize;
+
+	/**
+	 * The number of columns to span on all devices
+	 */
+	size?: TColSize;
+
+	/**
+	 * The number of columns to span on small devices
+	 */
+	sm?: TColSize;
+
+	/**
+	 * The number of columns to span on extra-small devices
+	 */
+	xs?: TColSize;
+}
+
+const ClayCol: React.FunctionComponent<IProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	lg,
+	md,
+	size,
+	sm,
+	xs,
+	...otherProps
+}) => {
+	const noBreakPoints = !lg && !md && !sm && !xs && !size;
+
+	xs = xs || size;
+
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, {
+				col: noBreakPoints,
+				'col-lg': lg === true,
+				[`col-lg-${lg}`]: lg && typeof lg !== 'boolean',
+				'col-md': md === true,
+				[`col-md-${md}`]: md && typeof md !== 'boolean',
+				'col-sm': sm === true,
+				[`col-sm-${sm}`]: sm && typeof sm !== 'boolean',
+				'col-xs': xs === true,
+				[`col-xs-${xs}`]: xs && typeof xs !== 'boolean',
+			})}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+export default ClayCol;
--- a/packages/clay-layout/src/Container.tsx
+++ b/packages/clay-layout/src/Container.tsx
@@ -1,0 +1,70 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+
+	/**
+	 * Adds `.container-fluid` class to create a fluid container that
+	 * doesn't expand beyond a set width
+	 */
+	fluid?: boolean;
+
+	/**
+	 * Adds `.container-fluid-${size}` class to set max width on container.
+	 */
+	fluidSize?: 'sm' | 'md' | 'lg' | 'xl';
+
+	/**
+	 * Adds the `.container-form-${formSize}` class to properly space
+	 * between application controls and the form. This class only modifies
+	 * the padding on the container.
+	 */
+	formSize?: 'sm' | 'md' | 'lg' | 'xl';
+
+	/**
+	 * Adds the `.container-view` class to properly space between application
+	 * controls and view pages (e.g., Card View, Table View, or List View).
+	 * This class only modifies the padding on the container.
+	 */
+	view?: boolean;
+}
+
+const ClayContainer: React.FunctionComponent<IProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	fluid,
+	fluidSize,
+	formSize,
+	view,
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, {
+				container: !fluid,
+				'container-fluid': fluid,
+				[`container-form-${formSize}`]: formSize,
+				'container-view': view,
+				[`container-fluid-max-${fluidSize}`]: fluid && fluidSize,
+			})}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+export default ClayContainer;
--- a/packages/clay-layout/src/ContainerFluid.tsx
+++ b/packages/clay-layout/src/ContainerFluid.tsx
@@ -1,0 +1,30 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import React from 'react';
+
+import Container from './Container';
+
+interface IProps
+	extends Omit<
+		React.ComponentProps<typeof Container>,
+		'fluid' | 'fluidSize'
+	> {
+	size?: React.ComponentProps<typeof Container>['fluidSize'];
+}
+
+const ClayContainerFluid: React.FunctionComponent<IProps> = ({
+	children,
+	size = 'xl',
+	...otherProps
+}) => {
+	return (
+		<Container {...otherProps} fluid fluidSize={size}>
+			{children}
+		</Container>
+	);
+};
+
+export default ClayContainerFluid;
--- a/packages/clay-layout/src/Content.tsx
+++ b/packages/clay-layout/src/Content.tsx
@@ -1,0 +1,155 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IRowProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+
+	/**
+	 * Provides the benefit of aligning content via flexbox without losing the behavior
+	 * of floated elements at the expense of extra markup.
+	 */
+	float?:
+		| boolean
+		| 'sm-down'
+		| 'md-down'
+		| 'end'
+		| 'end-sm-down'
+		| 'end-md-down';
+
+	/*
+	 * Give negative margins on the top, right, bottom, and left to offset the padding
+	 */
+	noGutters?: 'x' | 'y' | true;
+
+	/*
+	 * Gives padding to all autofit-cols that are direct children of autofit-row.
+	 */
+	padded?: boolean;
+
+	/**
+	 * Adds class for aligning items within the row.
+	 */
+	verticalAlign?: 'center' | 'end';
+}
+
+const ContentRow: React.FunctionComponent<IRowProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	float,
+	noGutters,
+	padded,
+	verticalAlign,
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'autofit-row', {
+				'autofit-float': float === true,
+				[`autofit-float-${float}`]: typeof float === 'string',
+				'autofit-padded': padded,
+				'autofit-padded-no-gutters': noGutters === true,
+				[`autofit-padded-no-gutters-${noGutters}`]:
+					typeof noGutters === 'string',
+				[`autofit-row-${verticalAlign}`]: verticalAlign,
+			})}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+ContentRow.displayName = 'ClayContentRow';
+
+interface IColProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+
+	/*
+	 * Makes column expand and fill available space in row.
+	 */
+	expand?: boolean;
+
+	/**
+	 * Provides the benefit of aligning content via flexbox without losing the behavior
+	 * of floated elements at the expense of extra markup.
+	 */
+	float?: 'end';
+
+	/**
+	 * Applies the `autofit-col-gutters` class
+	 */
+	gutters?: boolean;
+}
+
+const ContentCol: React.FunctionComponent<IColProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	expand,
+	float,
+	gutters,
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'autofit-col', {
+				'autofit-col-expand': expand,
+				'autofit-col-gutters': gutters,
+				[`autofit-col-float-${float}`]: float,
+			})}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+ContentCol.displayName = 'ClayContentCol';
+
+interface IColProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+}
+
+const ContentSection: React.FunctionComponent<IColProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'autofit-section')}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+ContentSection.displayName = 'ClayContentSection';
+
+export {ContentCol, ContentRow, ContentSection};
--- a/packages/clay-layout/src/Row.tsx
+++ b/packages/clay-layout/src/Row.tsx
@@ -1,0 +1,51 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+
+	/**
+	 * This removes the negative margins from .row and the
+	 * horizontal padding from all immediate children columns
+	 */
+	gutters?: boolean;
+
+	/**
+	 * Horizontal positioning of row contents
+	 */
+	justify?: 'start' | 'center' | 'end' | 'around' | 'between';
+}
+
+const ClayRow: React.FunctionComponent<IProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	gutters = true,
+	justify,
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames('row', className, {
+				'no-gutters': !gutters,
+				[`justify-content-${justify}`]: justify,
+			})}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+export default ClayRow;
--- a/packages/clay-layout/src/Sheet.tsx
+++ b/packages/clay-layout/src/Sheet.tsx
@@ -1,0 +1,101 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IContainerProps extends React.HTMLAttributes<HTMLElement> {
+	/**
+	 * Element or component to render for container
+	 */
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+}
+
+const SheetHeader: React.FunctionComponent<IContainerProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'sheet-header')}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+SheetHeader.displayName = 'ClaySheetHeader';
+
+const SheetFooter: React.FunctionComponent<IContainerProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'sheet-footer')}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+SheetFooter.displayName = 'ClaySheetFooter';
+
+const SheetSection: React.FunctionComponent<IContainerProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'sheet-section')}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+SheetSection.displayName = 'ClaySheetSection';
+
+interface IProps extends IContainerProps {
+	/**
+	 * Setting this to sets a max-width on the sheet
+	 */
+	size?: 'lg';
+}
+
+const Sheet: React.FunctionComponent<IProps> = ({
+	children,
+	className,
+	containerElement: ContainerElement = 'div',
+	size,
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'sheet', {
+				[`sheet-${size}`]: size,
+			})}
+		>
+			{children}
+		</ContainerElement>
+	);
+};
+
+Sheet.displayName = 'ClaySheet';
+
+export {Sheet, SheetFooter, SheetHeader, SheetSection};
--- a/packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
@@ -1,0 +1,146 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ClayLayout renders 1`] = `
+<div>
+  <div
+    class="container"
+  >
+    <div
+      class="row justify-content-start"
+    >
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+    </div>
+    <div
+      class="row justify-content-center"
+    >
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+    </div>
+    <div
+      class="row justify-content-end"
+    >
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+    </div>
+    <div
+      class="row justify-content-around"
+    >
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+    </div>
+    <div
+      class="row justify-content-between"
+    >
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+      <div
+        class="col-xs-4"
+      >
+        One of two columns
+      </div>
+    </div>
+  </div>
+  <div
+    class="row"
+  >
+    <div
+      class="col-md-8 col-xs-12"
+    >
+      .col-12 .col-md-8
+    </div>
+    <div
+      class="col-md-4 col-xs-6"
+    >
+      .col-6 .col-md-4
+    </div>
+  </div>
+  <div
+    class="row"
+  >
+    <div
+      class="col-md-4 col-xs-6"
+    >
+      .col-6 .col-md-4
+    </div>
+    <div
+      class="col-md-4 col-xs-6"
+    >
+      .col-6 .col-md-4
+    </div>
+    <div
+      class="col-md-4 col-xs-6"
+    >
+      .col-6 .col-md-4
+    </div>
+  </div>
+  <div
+    class="row"
+  >
+    <div
+      class="col-xs-6"
+    >
+      .col-6
+    </div>
+    <div
+      class="col-xs-6"
+    >
+      .col-6
+    </div>
+  </div>
+  <div
+    class="row"
+  >
+    <div
+      class="col"
+    >
+      First, but unordered
+    </div>
+    <div
+      class="order-12 col"
+    >
+      Second, but last
+    </div>
+    <div
+      class="order-1 col"
+    >
+      Third, but first
+    </div>
+  </div>
+</div>
+`;
--- a/packages/clay-layout/src/index.tsx
+++ b/packages/clay-layout/src/index.tsx
@@ -1,0 +1,39 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import Col from './Col';
+import Container from './Container';
+import ContainerFluid from './ContainerFluid';
+import {ContentCol, ContentRow, ContentSection} from './Content';
+import Row from './Row';
+import {Sheet, SheetFooter, SheetHeader, SheetSection} from './Sheet';
+
+export {
+	Col,
+	Container,
+	ContainerFluid,
+	ContentCol,
+	ContentRow,
+	ContentSection,
+	Row,
+	Sheet,
+	SheetFooter,
+	SheetHeader,
+	SheetSection,
+};
+
+export default {
+	Col,
+	Container,
+	ContainerFluid,
+	ContentCol,
+	ContentRow,
+	ContentSection,
+	Row,
+	Sheet,
+	SheetFooter,
+	SheetHeader,
+	SheetSection,
+};
--- a/packages/clay-layout/stories/index.tsx
+++ b/packages/clay-layout/stories/index.tsx
@@ -1,0 +1,165 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import '@clayui/css/lib/css/atlas.css';
+import {storiesOf} from '@storybook/react';
+import React from 'react';
+
+import ClayLayout from '../src';
+
+const exampleStyles = {
+	backgroundColor: '#E7E7ED',
+	border: '1px solid #CDCED9',
+	paddingBottom: '.75rem',
+	paddingTop: '.75rem',
+};
+
+const ClayDemoColumn: typeof ClayLayout.Col = (props) => (
+	<ClayLayout.Col style={exampleStyles} {...props} />
+);
+
+storiesOf('Components|ClayLayout', module)
+	.add('row positioning', () => (
+		<ClayLayout.ContainerFluid view>
+			<ClayLayout.Row justify="start">
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+			</ClayLayout.Row>
+			<ClayLayout.Row justify="center">
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+			</ClayLayout.Row>
+			<ClayLayout.Row justify="end">
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+			</ClayLayout.Row>
+			<ClayLayout.Row justify="around">
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+			</ClayLayout.Row>
+			<ClayLayout.Row justify="between">
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+				<ClayDemoColumn size={4}>{'One of two columns'}</ClayDemoColumn>
+			</ClayLayout.Row>
+		</ClayLayout.ContainerFluid>
+	))
+	.add('stacking at breakpoints', () => (
+		<ClayLayout.ContainerFluid view>
+			<ClayLayout.Row>
+				<ClayDemoColumn sm={8}>{'col-sm-8'}</ClayDemoColumn>
+				<ClayDemoColumn sm={4}>{'col-sm-4'}</ClayDemoColumn>
+			</ClayLayout.Row>
+			<ClayLayout.Row>
+				<ClayDemoColumn sm>{'col-sm'}</ClayDemoColumn>
+				<ClayDemoColumn sm>{'col-sm'}</ClayDemoColumn>
+				<ClayDemoColumn sm>{'col-sm'}</ClayDemoColumn>
+			</ClayLayout.Row>
+		</ClayLayout.ContainerFluid>
+	))
+	.add('ordering', () => (
+		<ClayLayout.ContainerFluid view>
+			<ClayLayout.Row>
+				<ClayDemoColumn>{'First, but unordered'}</ClayDemoColumn>
+				<ClayDemoColumn className="order-12">
+					{'Second, but last'}
+				</ClayDemoColumn>
+				<ClayDemoColumn className="order-1">
+					{'Third, but first'}
+				</ClayDemoColumn>
+			</ClayLayout.Row>
+		</ClayLayout.ContainerFluid>
+	))
+	.add('mix and match', () => (
+		<ClayLayout.ContainerFluid view>
+			<ClayLayout.Row>
+				<ClayDemoColumn md={8} size={12}>
+					{'.col-12 .col-md-8'}
+				</ClayDemoColumn>
+				<ClayDemoColumn md={4} size={6}>
+					{'.col-6 .col-md-4'}
+				</ClayDemoColumn>
+			</ClayLayout.Row>
+
+			<ClayLayout.Row>
+				<ClayDemoColumn md={4} size={6}>
+					{'.col-6 .col-md-4'}
+				</ClayDemoColumn>
+				<ClayDemoColumn md={4} size={6}>
+					{'.col-6 .col-md-4'}
+				</ClayDemoColumn>
+				<ClayDemoColumn md={4} size={6}>
+					{'.col-6 .col-md-4'}
+				</ClayDemoColumn>
+			</ClayLayout.Row>
+
+			<ClayLayout.Row>
+				<ClayDemoColumn size={6}>{'.col-6'}</ClayDemoColumn>
+				<ClayDemoColumn size={6}>{'.col-6'}</ClayDemoColumn>
+			</ClayLayout.Row>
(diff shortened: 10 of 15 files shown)


packages/clay-layout/src/Col.tsx
```diff
@@ -0,0 +1,85 @@
...
+	children,
+	className,
+	lg,
+	md,
+	order,
+	size,
+	sm,
+	xs,
+	...otherProps
+}) => {
```
jbalsas, line 59: Nice! 

We didn't add `align` and `order` to our JSP implementation... might be worth doing based on this! 🤔

packages/clay-layout/src/Container.tsx
```diff
@@ -0,0 +1,61 @@
...
+	 * Adds the `.container-form-${formSize}` class to properly space
+	 * between application controls and the form. This class only modifies
+	 * the padding on the container.
+	 */
+	formSize?: 'sm' | 'md' | 'lg' | 'xl';
+
+	/**
+	 * Adds `.container-fluid-${size}` class to set max width on container.
+	 */
+	fluidSize?: 'sm' | 'md' | 'lg' | 'xl';
```
jbalsas, line 26: We just [called this size](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/java/com/liferay/frontend/taglib/clay/servlet/taglib/ContainerTag.java#L160) in our JSP tag since it seems to be the basic function of the Container, so why the prefix?

We also default to `xl` since it's by far the main usage we saw across the board... might want to consider it here too

packages/clay-layout/src/Container.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Adds `.container-fluid` class to create a fluid container that
+	 * doesn't expand beyond a set width
+	 */
+	fluid?: boolean;
```
jbalsas, line 14: I lost track of where, but we went through a back and forth over `To Fluid or Not To Fluid` and concluded that we didn't want To Fluid 😂 

So, by default, we'd have `<Container>...</Container>` output:

```html
<div class="container-fluid container-fluid-xl">
    ...
</div>
```

This is somewhat opinionated... but so is our design language and as long as this can be overridden we should be fine. Not sure we have a valid case for just `container`, at the very least is not widespread enough to grant a first-level API

packages/clay-layout/src/Container.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * Adds the `.container-form-${formSize}` class to properly space
+	 * between application controls and the form. This class only modifies
+	 * the padding on the container.
+	 */
+	formSize?: 'sm' | 'md' | 'lg' | 'xl';
```
jbalsas, line 21: This is actually an interesting take... we treated this as `className` but it might make sense as an API...

@eduardoallegrini, thoughts?

packages/clay-layout/src/Container.tsx
```diff
@@ -0,0 +1,61 @@
...
+	 * Adds `.container-fluid-${size}` class to set max width on container.
+	 */
+	fluidSize?: 'sm' | 'md' | 'lg' | 'xl';
+
+	/**
+	 * Adds the `.container-view` class to properly space between application
+	 * controls and view pages (e.g., Card View, Table View, or List View).
+	 * This class only modifies the padding on the container.
+	 */
+	view?: boolean;
```
jbalsas, line 33: Same as with `.formSize`...

@eduardoallegrini, thoughts?

packages/clay-layout/src/Row.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const ClayRow: React.FunctionComponent<IProps> = ({
+	align,
+	children,
+	className,
+	gutters = true,
+	justify,
+	...otherProps
+}) => {
```
jbalsas, line 34: We didn't add any of these to `<clay:col>` since we couldn't find real usages, so it might be premature API, but LGTM

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -0,0 +1,146 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ClayLayout renders 1`] = `
+<div>
+  <div
+    class="container"
```
jbalsas, line 6: As I said above, if we want to somewhat match expectations, we might want to be a tad bit opinionated here and have this be `container-fluid container-fluid-xl` which is by far our most beloved container 😂

wincent:
@jbalsas already beat me to it on the subject of cross-implementation consistency (which I think is pretty desirable), but I was going to provide a couple of links that could be useful for comparing against the taglib API:

- [`<clay:container>` attributes and class names they map to](https://github.com/john-co/liferay-portal/blob/2179cd9aba2cd1b5d726a6d3c767a0065cd880d9/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/java/com/liferay/frontend/taglib/clay/servlet/taglib/ContainerTag.java#L130-L150).
- [`<clay:col>` attributes and class names they map to](https://github.com/john-co/liferay-portal/blob/2179cd9aba2cd1b5d726a6d3c767a0065cd880d9/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/java/com/liferay/frontend/taglib/clay/servlet/taglib/ColTag.java#L166-L208).

packages/clay-layout/src/Container.tsx
bryceosterhaus: I used the prefix since I didn't want people to think `size` could correlate to `container-${size}`. I thin by making it explicitly named `fluid` its apparent that its for the `container-fluid` class

packages/clay-layout/src/Container.tsx
bryceosterhaus: Hmm, I don't think I was privy to that convo.

IMO, I think its really odd to default to `container-fluid` for a component called `Container`, especially when the class `container` is also a thing. 

I think it'd be fine to keep `container-fluid` as a default for the a taglib in portal, but for the Clay library I don't think it would make sense to default to fluid for the name `Container`

packages/clay-layout/src/Container.tsx
bryceosterhaus: I went through our documentation and tried to expose API that we explicitly talk about in the css or in our layouts definition. This is why I added it.

packages/clay-layout/src/Col.tsx
bryceosterhaus: My goal was to mimic the docs examples as much as possible without having the user to have to add custom classes.

packages/clay-layout/src/Row.tsx
bryceosterhaus: I think its hard for devs to remember the util classes for alignment, my thought was by adding the align and justify props, it would be more accessible to devs via ts autocomplete.

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
bryceosterhaus: I think I would push back on that for the library level. Unless you can think of a good naming convention for the component to indicate it will include those. For a user to use `<Container>` without props and then get those classes seems odd to me. On the portal side of things, we can default to that since it makes sense for that to be portal's default, but not sure it makes sense to default at the library level.

I would also potentially push back and add that if we expect the default behavior of container to be `container-fluid container-fluid-xl` we might consider making that at the css level.

bryceosterhaus:
> aligning some APIs and outputs with what we found based on real usage.

I think we should probably clarify expectations on this a bit. Are we thinking the taglibs and the components should be 1-for-1 API match? IMO I think they should be similar but not exact. If we make them exact, I think it forces clay to export odd defaults to the API that would be unexpected for a user. I made a comment about this on the container component with the container-fluid class above.

On the portal side, I think it makes sense to default values though..

```jsx
// via jsx
<Clay.Container />

// renders
<div class="container" />




// jsp
<clay:container />

// clay:container internal use
<Clay.Container fluid fluidSize="xl" />

// renders
<div class="container-fluid container-fluid-max-xl">
```


Otherwise, if we are pushing that the default Container of clay to be `container-fluid container-fluid-max-xl`, then maybe that should be the default on the css side for `.container`

Does my line of thinking make sense here or am I wildly off?

bryceosterhaus:
added an additional commit. Went ahead and added the `row-cols-${size}-${value}`  API to row. This will make it easy to configure a full grid. I also updated some demos and added `containerElement` prop which allows users to provide their own component or a different html element.

packages/clay-layout/src/Container.tsx
kresimir-coko: I'm unsure as to the importance of these "proprietary" classes if we're going with the approach of separating the library from Portal. If I were making an app using Clay, I wouldn't really bother too much with props that have these very specific use cases.

Perhaps the question to ask is: _"How often will the user use this component with pre-existing knowledge of which props they'll need?"_
If it were me, I'd use `<ClayContainer flexOrEquivalent />` and then build out the rest, and then when it came to styling I'd probably not go back and check every component to see if I can configure the layout using props but instead just work in CSS.

packages/clay-layout/src/Row.tsx
```diff
@@ -0,0 +1,96 @@
...
+import classNames from 'classnames';
+import React from 'react';
+
+type TNumberOfColumns = 1 | 2 | 3 | 4 | 5 | 6;
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Vertical positioning of row contents
+	 */
+	align?: 'start' | 'center' | 'end';
```
kresimir-coko, line 15: `align-items` has more than these 3 options. [MDN on align-items](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items)

packages/clay-layout/src/Row.tsx
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * This removes the negative margins from .row and the
+	 * horizontal padding from all immediate children columns
+	 */
+	gutters?: boolean;
+
+	/**
+	 * Horizontal positioning of row contents
+	 */
+	justify?: 'start' | 'center' | 'end' | 'around' | 'between';
```
kresimir-coko, line 34: Same as `align-items`, has more options than the ones listed here.

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
kresimir-coko: I agree with @bryceosterhaus here. Default values can be provided in the implementation on Portal, thus preserving abstraction for the end-user while still providing users of Clay as a library the flexibility.

packages/clay-layout/stories/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+import '@clayui/css/lib/css/atlas.css';
+import {number} from '@storybook/addon-knobs';
+import {storiesOf} from '@storybook/react';
+import React from 'react';
+
+import ClayLayout from '../src';
+
+const exampleStyles = {
+	backgroundColor: 'rgba(86,61,124,.15)',
```
kresimir-coko, line 14: Not sure if it's too much, maybe you can provide some insight @pat270, but what do you think about using some brand colors here? Adding some branding to storybook might make it look more "ours" if you know what I mean.

packages/clay-layout/src/Container.tsx
jbalsas: > I were making an app using Clay, I wouldn't really bother too much with props that have these very specific use cases [...]
> 
> [...] I'd probably not go back and check every component to see if I can configure the layout using props but instead just work in CSS [...]

That begs the question... why would you bring in a library dependency such as Clay to substitute `<div>` by `<ClayContainer>` and keep it all the same?

If you won't bother checking the props and defaults of the library of your choosing, why'd you choose it in the first place? 

We've discussed at length our vetting process for picking dependencies that are actually worth their value. If we were to have that conversation around this library, maybe @wincent, or yourself would argue that we don't need Clay since we already have `flexOrEquivalent`.

That might mean, you wouldn't be a user of Clay in a normal situation, so we shouldn't model it around this preference :)

packages/demos/stories/ListPage.tsx
```diff
@@ -296,7 +297,7 @@ export default () => {
 				)}
 			</ClayManagementToolbar>
 
-			<div className="container" style={{paddingTop: 8}}>
+			<ClayLayout.Container style={{paddingTop: 8}}>
```
kresimir-coko, line 300: In the words of my toxic teacher from elementary school: "8 what? Potatoes?". Shouldn't we explicitly say pixels here?

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
jbalsas: Hey @bryceosterhaus, @kresimir-coko, I'll try to go at some more length in the general comments, but:

Once again, I think we're failing to see that right now, 100% (unless Faro moved to Clay) of our users are DXP. Also, **90% of our usage when implementing lexicon is `container-fluid`.** This means that by not providing this we're refusing to provide **real value to real users** of our library based on some not-so-real user/library expectation. Since 90% of the times someone in Liferay looking to build a UI had to use `container-fluid`, it is our goal to make this easier for them.

Please, go through:
- https://github.com/jbalsas/liferay-portal/pull/2160/commits/e72f30c231e196b72dcf15bbff7d94ddda6b2655
- https://github.com/eduardoallegrini/liferay-portal/pull/71/commits
- https://github.com/eduardoallegrini/liferay-portal/pull/72/commits/4c9f07b95116fd6f27dba4f59e6e0557299bd29d

The main goal of Clay is to **simplify implementing Lexicon UIs**. The main implementor of Lexicon UIs is DXP. Based on the above changes, I'd say those are better defaults for implementing Lexicon UIs. Thus, those defaults belong in Clay, not in DXP. Otherwise, we're not making it any easier. There are literally thousands of occurrences of `container-fluid` and around tens of `container`. It doesn't make any sense that we'd favour the latter here.

If you want to be real purist here, **let's maybe create `FluidContainer` that composes `Container` and provides the right semantics and ergonomics to our devs**. We can do the same in the Portal side and mostly simply use `<clay:fluid-container>` and `<ClayLayout.FluidContainer>`.

jbalsas: @wincent, @marcoscv-work, thoughts?

jbalsas:
Hey @bryceosterhaus!

> Are we thinking the taglibs and the components should be 1-for-1 API match? IMO I think they should be similar but not exact.

They can't really be 1-for-1 since Java-JS doesn't translate all that well. That being said, the closer the better.

> If we make them exact, I think it forces clay to export odd defaults to the API that would be unexpected for a user.

This is a different topic. I also answered in the [container-fluid](https://github.com/liferay/clay/pull/3204#discussion_r422835440) discussion. To summarize, I believe Clay should provide the defaults that make sense to achieve its main goal. Clay's goal is making it easier to build Lexicon UIs. Based on real experience building Lexicon UIs, those defaults are not in this PR.

```
// jsp
<clay:container />

// clay:container internal use
<Clay.Container fluid fluidSize="xl" />
```

One source of confusion I see here is that you seem to think `JSP` implementations actually use React components...  which will be true and not true at the same time. We currently don't have SSR, so JSPs do not defer on React components to render. They do to hydrate for interactive components, though.

> Otherwise, if we are pushing that the default Container of clay to be container-fluid container-fluid-max-xl, then maybe that should be the default on the css side for .container
>
> Does my line of thinking make sense here or am I wildly off?

I think there's a mixup with 2 concerns.
- What are the right defaults for Clay as a Library? -> Those that simplify building Lexicon UIs. Forcing a developer to always write `<Clay.Container fluid fluidSize="xl">` means the defaults we provide are not valuable since they're of no use 90% of the time. What good are defaults that only work 10% of the time?
- Defaulting `<Container>` to `.container-fluid` seems odd to you, and on that I can agree. Maybe a middle ground would be to create a highish lower level component like **FluidContainer** that composes **Container** and provides the right semantics and ergonomics to our devs.

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
kresimir-coko: @jbalsas When you put it like that yea, I agree. I guess I didn't understand the exact goal of Clay, I thought it was a low-level library first, DXP-specific second. 

I figured people using DXP used Clay to build their apps, but I assume it's not like that after all. I don't personally care about being a purist here, so I have no problem with having the default be `container-fluid container-fluid-xl`. 

Something to maybe think about is that we can have a prop that strips all custom classes from the container, like `<ClayContainer basic/vanilla/lite />`, this will provide the defaults for 90% of the users in DXP like you said, but also give the dev that's using Clay for a personal project a single prop to enable flexibility in styling via CSS with just a single prop.

jbalsas: > I didn't understand the exact goal of Clay, I thought it was a low-level library first, DXP-specific second

Screen Shot 2020-05-11 at 09 44 02

That 👆 😂 

> `<ClayContainer basic/vanilla/lite />`

You were earlier arguing that you wouldn't bother checking props and go about it your way with CSS, so why the change of heart? 

I can understand @bryceosterhaus's struggle here, so I'm not pushing hard to make those the defaults for `<ClayContainer>`.

We need to:
- Fulfil the library's goal which is making Lexicon UIs easier
- We need to analyze what it means to build Lexicon UIs. Our biggest source of data here is DXP since that's the main Lexicon UI implementor. 
- We need to provide defaults and components based on how real developers are developing real UIs. That's the fastest way to provide real value to the users of our library.

packages/clay-layout/src/Container.tsx
kresimir-coko: The fact that the primary goal of Clay is to serve DXP makes my concern kinda obsolete. And that's fine, I was just looking out for the little man here 😄, all of us using Clay while building DXP will look into all the props and use them accordingly. So having the classes that are omnipresent throughout DXP  in this component as props is a time saver for us.

jbalsas: > primary goal of Clay is to serve DXP

I didn't say this!!

**Primary goal of Clay is to make building Lexicon UIs easy, consistent, robust and safe.**

DXP is the main consumer right now, so we shouldn't fall into the common trap of serving imaginary users and failing to deliver to those teams. We've been down that path before. It ain't pretty 😉 

That being said, we also aim for other departments and teams to jump on board as Design practices converge. AC (Faro), IS, DXP Cloud... they will all benefit in the long run if we do a good job here. That's why looking out for those "general" users is still valid and important, just not at the expense of DXP.

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
kresimir-coko: You got me there @jbalsas with that screenshot 😆 

### Why the change of heart? 
Not really a change of heart I think, if it's one prop that will disable all custom classes except `.container` I think is a good compromise. If that prop shows up on autocomplete as one of the first ones that's even better. But then again, my PoV was different when we started this discussion, as I was looking at it from a different use-case than I am now. Now I agree with you @jbalsas.

### Fluid Container
I think that having a FluidContainer on top of Container might muddy things up unnecessarily, as you said, 90% of usages are Fluid, and we can maybe add a `notFluid` prop that will omit Fluidity from the container. I would really dislike going into DXP code a couple years from now and seeing `<FluidContainer />` everywhere

packages/clay-layout/src/Container.tsx
kresimir-coko: Yea, I was worried my wording was too strong. I get what you're saying 👍 . I was focusing on the imaginary users more than on DXP. That's what makes Lexicon great, as long as we make components based on their design it should be great for both DXP and the other projects Liferay has, right?

jbalsas: > That's what makes Lexicon great, as long as we make components based on their design it should be great for both DXP and the other projects Liferay has, right?

That's the dream 😂

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
wincent: > @wincent, @marcoscv-work, thoughts?

This is one of those ones where I don't have anything new to add beyond what I've already said (here, [and on the issue](https://github.com/liferay/clay/issues/2421))  — and in other places about [what/why I think Clay even is](https://github.com/liferay/clay/pull/3034#issuecomment-603125574) — , and the discussion here is already very difficult to follow so I won't help much by adding more comments.

bryceosterhaus: > This means that by not providing this we're refusing to provide real value to real users of our library based on some not-so-real user/library expectation. 

Yeah this totally makes sense, my thought is based more css vs react component discrepancies. **My main concern is that we should have consist naming in our components that stem from @clayui/css.** By defaulting `ClayLayout.Container` to be fluid, seemed to be at odds with @clayui/css. Also, if Lexicon is recommending the styles of `.container-fluid` to be used 90% of the time, and `.container` just 10% of the time, maybe we should just have one class and not use the non-fluid styles.

> If you want to be real purist here, let's maybe create FluidContainer that composes Container and provides the right semantics and ergonomics to our devs. 

I'm not sure I would label myself as a _purist_ 😂, I just want to make sure we are consistent from @clayui/css to our react components. I like this idea for naming though. Maybe even going with `ClayLayout.ContainerFluid` for more consistency. I think this would be inline with both @clayui/css and the need in DXP.

bryceosterhaus: > @kresimir-coko 
> I would really dislike going into DXP code a couple years from now and seeing <FluidContainer /> everywhere


What would be the issue with this? I don't think it'lI be an issue really and I think it will actually be better for readability in the future and knowing exactly what the containers purpose is. I also don't mind it because it'll be easy to find/replace for any future refactors

packages/clay-layout/src/Row.tsx
bryceosterhaus: Yeah I am following bootstraps docs on their alignment classes. https://getbootstrap.com/docs/4.0/layout/grid/#vertical-alignment

I think we can start with these 3 for now and then evaluate later if people need more.

packages/clay-layout/src/Row.tsx
bryceosterhaus: Same as above comment, I am trying to keep it minimal and follow bootstraps grid docs for what should be available.

packages/demos/stories/ListPage.tsx
bryceosterhaus: I don't think its a huge issue for us to specify since in React we know that any style that is applied via integer becomes a px. I would agree with you in the case of a string, then we would need to specify.

https://reactjs.org/docs/dom-elements.html#style
> React will automatically append a “px” suffix to certain numeric inline style properties. If you want to use units other than “px”, specify the value as a string with the desired unit.

bryceosterhaus:
@jbalsas I commented on the container-fluid comment as well, but I'll respond to 

> create a "highish" lower level component like FluidContainer that composes Container and provides the right semantics and ergonomics to our devs.

Yeah I think this is the best means forward for this since it keeps consistency with @clayui/css and also provides a helpful OOTB solution for dxp.


On similar note, I created an issue to further decide how to avoid these reoccurring discussions around clay + dxp. I think the confusion comes up because we have so many different layers(dxp, taglibs, react components, css) that are intertwined and don't necessarily have clear definitions to guide them. https://github.com/liferay/clay/issues/3214 **Please use this issue for continued high-level comments.**

bryceosterhaus:
updated with `ContainerFluid` component

packages/clay-layout/src/Col.tsx
marcoscv-work: We even don't recommend to use `order` since it causes accessibility problems of tab perception, I would not add it to avoid encouraging its use

packages/clay-layout/src/Container.tsx
marcoscv-work: I think the correct name properties should be `fluidMaxSize` since we only use classes `container-fluid-max-*breakpoint*` to create all Lexicon grids they need. On the other hand, I would recommend forgetting all rest of ways to create container size limit because there are a lot which probably a final customer could use but we do not need

packages/clay-layout/src/Container.tsx
marcoscv-work: Initially, my concern about the using of `container` when we already have a class called `container` was the same @bryceosterhaus already has wrote, after some days taking a look to what lexicon want and what we finally use, the uses cases are only a few, majority of them are the classes `container-fluid container-fluid-max-xl` and we don't use the `standard` bootstrap container.
So for us and right now the default container is the class combination: `container-fluid container-fluid-max-xl`. It still sounds bad to me but I have to admit our container is a no very semantic combination of classes,  maybe it's something we can study and try to improve in a future more semantic CSS.

IMHO, the `<Container>` output should be our standard and more used container

packages/clay-layout/src/Container.tsx
marcoscv-work: I think it can be confusing with form sizes, if we speak about 'padding or margin' this probably should be 'spacers'. We have 8 unit sizes defined:
https://github.com/liferay/clay/blob/0273c22d1f6a9dfbd22e39f9490d7a9da05e53b7/packages/clay-css/src/scss/atlas/variables/_globals.scss#L182-L211
It generates all famous helpers like: `m-1`, `pt-5`, and our won utilities helpers: `c-m-5`, `c-pt-5`, etc.

Info about forms size:

For forms we have two groups sizes, and it can be consider a container: the default one `form-group` and `form-group form-group-sm`

for individual forms (form-controls) we have three sizes:
image

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
marcoscv-work: > @wincent, @marcoscv-work, thoughts?

We could create `FluidContainer` and `Container` to fit semantic but finally, we wouldn't use `Container` for anything, so I'm closer to fix the semantic in future versions of CSS than creating components or variations we don't really need

jbalsas: Actually, when we weigh in our `autofit` rows and cols this might play a part since those do more or less need an inert container... so `Container` (without the "container"  class actually... 😂) could help bridge the gave between the different grids.

Or maybe something else, so we'd have

Layout (Stack)  -> Container/FluidContainer -> Row/Col
                          |-> Row autofit -> Col autofit

Adding one additional layer feels a bit over-engineered... so maybe we could simplify with:

```jsx
<Container> -> <div class="container">...</div>
<Container autofit> -> <div>...</div>
<FluidContainer> -> <div class="container-fluid container-fluid-max-xl">...</div>
```

packages/clay-layout/src/ContainerFluid.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface IProps
+	extends Omit<
+		React.ComponentProps<typeof Container>,
+		'fluid' | 'fluidSize'
+	> {
+	size?: React.ComponentProps<typeof Container>['fluidSize'];
+}
+
+const ClayContainerFluid: React.FunctionComponent<IProps> = ({
```
jbalsas, line 18: LGTM!

nit: Should it `FluidContainer` instead of `ContainerFluid`?

jbalsas:
@bryceosterhaus, thanks for the update, that makes sense to me! 

At the risk of dragging this over for an additional day, do you mind weighing over at https://docs.google.com/document/d/13K-jishFtNLOoWCsWffjsCc6xsfaiH9_-ofXwoGdx3Y? (and anyone else interested, of course... looking at you, @kresimir-coko, @matuzalemsteles, @pat270...). 

I think we might be close to getting something more or less coherent that covers our most used cases that involve what we have here plus the `autofit-row` + `autofit-col` cases... plus some additional higher levels like `<Sheet>` and `<Card>` that we might or might not need to add high-level components for...

Looking for everyone to have a bit of a bigger picture to avoid too much back and forth in the implementation during the next few days, but certainly looking to wrap this up no later than this week (if possible) and continue iterating.

pat270:
@bryceosterhaus The only thing that caught my eye on the storybook demos was:
```
<ClayLayout.Row>
	<ClayDemoColumn md={8} size={12}>
		{'.col-12 .col-md-8'}
	</ClayDemoColumn>
</ClayLayout.Row>
```

having the size attribute output `.col-#` seems a little odd. Bootstrap groups all the `.col-#` classes under XS, see https://getbootstrap.com/docs/4.4/layout/grid/#grid-options.

packages/clay-layout/src/ContainerFluid.tsx
bryceosterhaus: Since its `.container-fluid`, it made more sense to me in keeping it consistent with that. I also prefer `NounDescriber`(if that makes sense)

packages/clay-layout/src/__tests__/__snapshots__/index.tsx.snap
marcoscv-work: Container 'type' autofit sounds good to me, if it makes clear this is not the BS container is enough.

It is out of time but I was thinking about something more semantic which wraps everything, something like:

Using **<Grid>**
<Grid container>
<Grid row>
<Grid col>
<Grid autofit-row>
<Grid autofit-col>

bryceosterhaus:
Updated with a few additional ideas. `<ClayLayout.Content*` is a component that uses "autofit-*" classes. See https://github.com/liferay/clay/pull/3204/commits/6244ab8aa35059cc681ead2c1530e6fe320eed2d#diff-58014a94aedc67ab5431dd06b8a5bf63 for what it would look like in action.

jbalsas:
> Updated with a few additional ideas. <ClayLayout.Content* is a component that uses "autofit-*" classes. See 6244ab8#diff-58014a94aedc67ab5431dd06b8a5bf63 for what it would look like in action.

Interesting @bryceosterhaus! I was playing around with the idea of an `autofit` attribute yesterday at https://github.com/jbalsas/liferay-portal/pull/2166.

I think this is more aligned with what @marcoscv-work envisioned... as we discussed, the problem here is that now you need to know 2 different set of APIs... but you did have to know about `autofit` anyways, so it doesn't look so different. I can be convinced and follow this to match our jsp tags...

Also, if the desire is to move away from lowish-level container-row-col to more content-aware components this seems like a good move...

@marcoscv-work, @eduardoallegrini, @wincent... thoughts?

wincent:
> @marcoscv-work, @eduardoallegrini, @wincent... thoughts?

I'd much rather hear from the first two people you mention, @jbalsas; I don't want to contribute any noise to an already very dense thread.

marcoscv-work:
> > Updated with a few additional ideas. <ClayLayout.Content* is a component that uses "autofit-*" classes. See [6244ab8](https://github.com/liferay/clay/commit/6244ab8aa35059cc681ead2c1530e6fe320eed2d)#diff-58014a94aedc67ab5431dd06b8a5bf63 for what it would look like in action.
> 
> Interesting @bryceosterhaus! I was playing around with the idea of an `autofit` attribute yesterday at [jbalsas/liferay-portal#2166](https://github.com/jbalsas/liferay-portal/pull/2166).
> 
> I think this is more aligned with what @marcoscv-work envisioned... as we discussed, the problem here is that now you need to know 2 different set of APIs... but you did have to know about `autofit` anyways, so it doesn't look so different. I can be convinced and follow this to match our jsp tags...
> 
> Also, if the desire is to move away from lowish-level container-row-col to more content-aware components this seems like a good move...
> 
> @marcoscv-work, @eduardoallegrini, @wincent... thoughts?

That's closer to what I tried to explain, it looks more semantic and I really like it.
If we agree we can use the pattern in clay and Liferay portal :-D

packages/clay-layout/src/ContainerFluid.tsx
jbalsas: I think it's weird... because in the other components we did the opposite?

I'm thinking now at `ContentSection`, `ContentRow` and `ContentCol`...  it might just be my lack of familiarity with the grammar, but those feel like `DescriberNoun`. Basically i can tell because describer is common and noun is the thing that changes at the end.

It's true that we don't have many such cases in Clay, but others I found that also seem to contradict this are `ItemField`, `ItemText` and `ItemTitle`...

jbalsas: Ah, yeah, never mind... I think I see it now 😂 👍

jbalsas: Ah, yeah, never mind, I think I see it now... 😂 👍

jbalsas: Or maybe not! 😂 @wincent, @pat270, some native english speakers here, please? I'm fine either way 👍

jbalsas: Just to state my confusion.

`ContentSection`, `ContentRow` and `ContentTitle`... one would assume as follows that we might have a `ContentContainer`.

`ItemField`, `ItemText`, `ItemTitle`... similarly, I'd think an `ItemContainer` seems logical...

From the above, `FluidContainer` seems to follow the sequence... but I was never good at this 😂

jbalsas:
> That's closer to what I tried to explain, it looks more semantic and I really like it.
If we agree we can use the pattern in clay and Liferay portal :-D

I'll start moving in that direction, then

packages/clay-layout/src/ContainerFluid.tsx
wincent: Native English speaker here... English laughs at your pathetic request to have any kind of consistent rules.

fart-in-your-general-direction

(I know the gif is "French", but it accurately summarizes the English attitude to consistency.)

The general grammatical pattern in English is to put modifiers (adjectives) before nouns (eg. "red car"). Things get complicated because English is quite happy to use nouns as modifers too (ie. noun like "money" being used like an adjective in the term "money bag") — but still, the modifier goes first. Unlike in Spanish, which is pretty much the opposite (eg. "coche rojo").

But programming is another layer above English, so there are other concerns beyond mere grammar. For example, we like to group related things together using a prefix. That's why `ItemField`, `ItemText`, `ItemTitle` start with a common prefix: the "thing" here is "Field", "Text", "Title", but the common relationship is that they are all the "thing" of an "Item", so we extract the common bit out and stick it at the front. And even more obviously, everything in Clay starts with "Clay", not because "Clay" is a modifier or an adjective or a noun or anything — it's the common root that it all relates to.

With respect to `ContainerFluid` vs `FluidContainer`, I have no idea. The "thing" it is is a container, and "Fluid" is the modifier that describes the container, so I'd expect it to be `FluidContainer`. But as @bryceosterhaus  said, he was being consistent with `.container-fluid`.

packages/clay-layout/src/Content.tsx
```diff
@@ -0,0 +1,129 @@
...
+	containerElement: ContainerElement = 'div',
+	expand,
+	gutters,
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'autofit-col', {
+				'autofit-col-expand': expand,
```
jbalsas, line 90: Looking at usages in DXP, `autofit-col-end` is also possible and very extended. It seems to be we usually have either nothing, `expand` or `end`. The `gutters` value is compatible with the above. 

We might want to add `end` as another prop... or maybe there's a better more semantic name for this...

@pat270, @marcoscv-work?  What's the difference between nothing, `expand` and `end`? How would we need to name an enumeration of this?

packages/clay-layout/stories/index.tsx
pat270: This sounds good to me. Maybe #E7E7ED or #cdced9?

packages/clay-layout/src/Content.tsx
pat270: @jbalsas `expand` is a good name.

We shouldn't include `autofit-col-end`. It was deprecated at https://github.com/liferay/clay/issues/803. The replacement for it is using `autofit-float-end` on `autofit-row`.

---

```
<div class="autofit-float-end autofit-row">
    <div class="autofit-section">
      The quick brown fox jumped over the lazy dog.
    </div>
  </div>
  <div class="autofit-col">
    <div class="autofit-section">
      <button>Add</button>
    </div>
  </div>
</div>
```

If the `autofit-col`s don't fit on one line, the last item will break to new line and everything (`autofit-col`) will be aligned right.

---

```
<div class="autofit-float-end-{sm|md}-down autofit-row"></div>
```
At the breakpoint `{sm|md}`, if `autofit-col`s don't fit on one line, last item will break to new line and aligned right.

---

```
<div class="autofit-float autofit-row"></div>
```

If doesn't fit on one line, breaks to new line and aligned left.

---

```
<div class="autofit-float-{sm|md}-down autofit-row"></div>
```

At breakpoint `{sm|md}`, if doesn't fit on one line, breaks to new line and aligned left.

---

- `autofit-col` sizes the column to be as wide as its content.

- `autofit-col autofit-col-expand` forces the column to fill the remaining space. If there are multiple `autofit-col-expand`s then the column widths are supposed to be divided equally. This doesn't work in all situations right now. I think we can fix this.

We also have:

```
<ul class="card-page card-page-equal-height">
  <li class="card-page-item col-{breakpoint}-#"></li>
  <li class="card-page-item col-{breakpoint}-#"></li>
  <li class="card-page-item col-{breakpoint}-#"></li>
</ul>
```

I don't think we should include this use case here, but just getting it out. We also have a grid for cards `card-page` and `card-page card-page-equal-height`. You can see a demo at https://pat270.github.io/clay3-test-site/vclay-table-dd/content/test-card-type-template/.

jbalsas: > It was deprecated at #803

That was like 2 years ago! 😂 We will need to discuss also how to deprecate things in claycss in the future :D 

@bryceosterhaus, that means if I understand correctly that `expand` on col is okay, but we're missing `float` or something like that on [ContentRow](https://github.com/liferay/clay/pull/3204/files#diff-3d9a224ed0f982f449f890b93b0b359aR47)?

packages/clay-layout/src/Sheet.tsx
```diff
@@ -0,0 +1,106 @@
...
+	containerElement: ContainerElement = 'div',
+	...otherProps
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(
+				className,
+				'sheet-footer',
+				'sheet-footer-btn-block-sm-down'
```
jbalsas, line 49: Is this always needed? This actually feels oddly specific. I can't find any usage of this in DXP... but @pat270 can call me out on this 😉

pat270: It was a specific request for https://projects.invisionapp.com/d/main#/projects/prototypes/12432242. It hasn't been deprecated. If we don't need it, I'm good with removing it if Lexicon approves.
sheet-block-btn

jbalsas: I meant if this class was really a default one, and always ncessary.

We might need to keep it, I’m just not sure if it’s a default... I’ll take a closer look at DXP next week, but we might just pass it through ‘className’ for now if in doubt...

pat270: OH I see. I would say don't default to `sheet-footer sheet-footer-btn-block-sm-down`. It's very specific to that form.

packages/clay-layout/src/ContainerFluid.tsx
bryceosterhaus: yeah the more I think about it I really have no idea why somethings sounds "more right" than the other way around. So I think I'll just rest my decision based on the class name itself.

bryceosterhaus:
Updated to add float

packages/clay-layout/src/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default {
+	Col,
+	Container,
+	ContainerFluid,
+	ContentCol,
+	ContentRow,
+	ContentSection,
+	Row,
+	Sheet,
```
jbalsas, line 32: What's the reason behind grouping all `Sheet` components but not all `Content` ones?

bryceosterhaus: Thats my bad, I will update it. The thought behind it is that we don't want multiple namespaces, so instead of `<ClayLayout.Sheet.Header>` it should be `<ClayLayout.SheetHeader`. So generally, the only namespace should be the top level component, in this case `ClayLayout`

bryceosterhaus: Just updated

packages/clay-layout/src/Content.tsx
```diff
@@ -0,0 +1,147 @@
...
+}) => {
+	return (
+		<ContainerElement
+			{...otherProps}
+			className={classNames(className, 'autofit-row', {
+				'autofit-float': float === true,
+				[`autofit-float-${float}`]: typeof float === 'string',
+				'autofit-padded': padded,
+				'autofit-padded-no-gutters': noGutters === true,
+				[`autofit-padded-no-gutters-${noGutters}`]:
```
jbalsas, line 64: Hey @pat270, one doubt here... is it `autofit-padded` always needed with `autofit-padded-no-gutters`? Is there any logic here that needs to be coded?

Meaning, is just `autofit-padded-no-gutters` correct? or setting `-no-gutters` implies `padded`?

@bryceosterhaus, just in case those 2 props are in fact related...

packages/clay-layout/src/Content.tsx
```diff
@@ -0,0 +1,147 @@
...
+	containerElement?: React.JSXElementConstructor<{
+		className: string;
+		[key: string]: any;
+	}>;
+
+	/**
+	 * Provides the benefit of aligning content via flexbox without losing the behavior
+	 * of floated elements at the expense of extra markup.
+	 */
+	float?:
```
jbalsas, line 22: Not sure what I'll name this in Java since `float` is a reserved keyword 😂

packages/clay-layout/src/Content.tsx
wincent: > You can see a demo at https://pat270.github.io/clay3-test-site/vclay-table-dd/content/test-card-type-template/.

(Off topic, but) there are some funny flickering on that page in Chrome 81:

Screenshot 2020-05-18 -115816-2fAZXhEY

Takes a while to trigger it, but you see it when I move the mouse off of the rightmost card.

packages/clay-layout/src/Container.tsx
jbalsas: So... now that we have `container-fluid`... maybe this could be:

`fluid: boolean | ['sm', 'md', 'lg', 'xl']`?

packages/clay-layout/src/Container.tsx
jbalsas: I would remove this one, just on the basis that it's easier to add them later than to remove them.

We can implement for now using `className`, and can bring it as an API once the pattern is more clear?

packages/clay-layout/src/Container.tsx
jbalsas: Similar to `formSize`... I know this is being used, si this has indeed real backing, but I wouldn't commit to it too early.

jbalsas:
Hey @bryceosterhaus, this looks almost good to go to me...

I just added a question about the [`padded` and `noGutters` relation](https://github.com/liferay/clay/pull/3204#discussion_r426158603) and a couple more comments here and there about APIs we might not want to commit to without more data.

My suggestion would be to remove those props for now, and go through [LPS-111257 Analyze Layout Strategies used in React Components in DXP - Report](https://docs.google.com/spreadsheets/d/1040_UfBNNj2wzoBjhsoec872Ns2Uvf4MiFbQuUoMRqg) based on an initial release (or local) of this. Keep in mind that I think that's missing `autofit` and `sheet` usage, so those should be added too.

With that, we should have first-hand visibility on these APIs and their ergonomics. We can create some LPSs to send updated Apps to their teams using the new components.

What do you think?

packages/clay-layout/src/Content.tsx
pat270: @jbalsas @bryceosterhaus 
No `autofit-padded` isn't always needed with `autofit-padded-no-gutters`. I added the same padding to both because it gets confusing. You can use `autofit-padded autofit-padded-no-gutters` or just `autofit-padded-no-gutters`.

I think the most common use case would be `autofit-row autofit-padded-no-gutters-x` which uses negative margins to align the left and right side of `autofit-row` (similar to Bootstrap's `row`), but provides a little white space top and bottom. 

---

`autofit-padded` adds `padding: 0.25rem 0.5rem` on `autofit-col`

---

`autofit-padded-no-gutters-x` adds `padding: 0.25rem 0.5rem` on `autofit-col` and `margin-left: -0.5rem; margin-right: -0.5rem` on `autofit-row`

---

`autofit-padded-no-gutters` adds `padding: 0.25rem 0.5rem` on `autofit-col` and `margin: -0.25rem -0.5rem` on `autofit-row`

packages/clay-layout/src/Content.tsx
pat270: Thanks @wincent, not sure when this started happening (I'm blaming a Chrome update on this). Any `transition-timing-function` other than `ease-in` or `ease-in-out` fixes this artifact.

bryceosterhaus:
> My suggestion would be to remove those props for now, 

Are we sure we want to remove them? `*-padded` and `*-padded-no-*` are used at least 20 times in DXP from my count. Also, it seems from the conversation that there isn't really any coding needed to link the two classes together since they behave independently.

bryceosterhaus:
> We can create some LPSs to send updated Apps to their teams using the new components.

Yeah I am up for this. I think it'd be wise to get an initial release of `@clayui/layout` merged into portal first and then we can go ahead and make these updates and LPSs.

jbalsas:
> Are we sure we want to remove them? *-padded and *-padded-no-* are used at least 20 times in DXP from my count. Also, it seems from the conversation that there isn't really any coding needed to link the two classes together since they behave independently.

No, I meant the `container`, `formSize` and `order`. The `padded` ones make sense to me and are pretty straightforward.

jbalsas:
> Yeah I am up for this. I think it'd be wise to get an initial release of @clayui/layout merged into portal first and then we can go ahead and make these updates and LPSs.

SGTM

bryceosterhaus:
> No, I meant the view, formSize and order. The padded ones make sense to me and are pretty straightforward.

Ah gotcha, I just did a quick search and it looks like view would be used 28 times and formSIze would also be used 28. `order` looks to be used much less.

Does it make sense to keep the props for those 28 cases of each? I would lean towards yes, unless you are thinking they will be removed.

jbalsas:
> Does it make sense to keep the props for those 28 cases of each? I would lean towards yes, unless you are thinking they will be removed.

Yeah, for 28 cases it definitely makes sense... I was just trying to err on the side of caution and add them later once we saw how a minimal API looked like. Just trying to avoid a "breaking change" if at some point we realized one of those was unused or weird. If it's clear to you, go for it, I don't have a particular concern about this.

bryceosterhaus:
> I was just trying to err on the side of caution and add them later once we saw how a minimal API looked like.

Totally makes sense. I think these APIs won't be too problematic for us and given the number of uses, I think it makes sense to add.

I went ahead and updated the PR:
- removed `order` prop since it wasn't being used much
- removed the sample uses of @clayui/layout in other components. Once the package is released I will go through and update our existing packages.
- squashed into a single commit

Not sure If I missed anything else here in this PR, I think I covered everything though. I will mark this as ready to review.

packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -74,41 +74,56 @@ export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
...
+										symbol={horizontalSymbol}
+									/>
+								</ClaySticker>
 							</span>
-						</span>
-					</ClayCard.Row>
-				)}
-			</ClayCard.Body>
+							{title && (
+								<span className="autofit-col autofit-col-expand">
```
jbalsas, line 112: This would be a good place to use `ContentCol` 😂

packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -74,41 +74,56 @@ export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
...
-							<ClaySticker inline>
-								<ClayIcon
-									spritemap={spritemap}
-									symbol={horizontalSymbol}
-								/>
-							</ClaySticker>
-						</span>
-						<span className="autofit-col autofit-col-expand">
-							<span className="autofit-section">
+			{(horizontal || title || description) && (
```
jbalsas, line 77: I'm looking at this in GitHub, so this might just be @wincent's most beloved UI playing tricks, but how did this change make it into this PR? Was it always here?

jbalsas: Yeah... probably just appears in the diff because we merged upstream... :sigh:

packages/clay-layout/package.json
```diff
@@ -0,0 +1,39 @@
+{
+	"name": "@clayui/layout",
+	"version": "3.0.0-alpha.1",
+	"description": "ClayLayout component",
```
jbalsas, line 4: nit, for the future: go over the descriptions in our packages. Doesn't feel like this says much about the package, what it does or why you'd want to download. Maybe we can get some copy from the Lexicon site and/or team

packages/clay-layout/src/Col.tsx
```diff
@@ -0,0 +1,88 @@
...
+import classNames from 'classnames';
+import React from 'react';
+
+type TColSize = boolean | number | 'auto';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Position in which to align column
+	 */
+	align?: 'start' | 'center' | 'end';
```
jbalsas, line 15: I'm still unsure about `align`... do we also have usages for this? I'm fine to keep it if you think it's safe and useful.

packages/clay-layout/src/Row.tsx
```diff
@@ -0,0 +1,96 @@
...
+				lg?: TNumberOfColumns;
+				md?: TNumberOfColumns;
+				sm?: TNumberOfColumns;
+				xs?: TNumberOfColumns;
+				xl?: TNumberOfColumns;
+		  };
+}
+
+const ClayRow: React.FunctionComponent<IProps> = ({
+	align,
```
jbalsas, line 52: Same thoughts I posted about `Col`

packages/clay-layout/src/Row.tsx
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * Horizontal positioning of row contents
+	 */
+	justify?: 'start' | 'center' | 'end' | 'around' | 'between';
+
+	/**
+	 * Number of columns the row should contain before breaking
+	 * on to a new line
+	 */
+	numOfCol?:
```
jbalsas, line 40: Do we want to commit to this API?

packages/clay-card/src/CardWithNavigation.tsx
wincent: E5R-Rwjp_400x400

packages/clay-layout/package.json
wincent: One of the consequences of dividing a library into 45 packages is that you have to come up with 45 descriptions. 😀

packages/clay-card/src/CardWithNavigation.tsx
bryceosterhaus: Yeah I will go through and update our other components once a cut is released. I think if we update it before releasing @clayui/layout then CI might break for other PRs

jbalsas: Sounds good! 👍

bryceosterhaus:
@jbalsas I'll verify some of these props in portal just to make sure. And yeah, I plan on adding the docs once this PR gets merged and before release tomorrow.

Issue #3209, opened 2020-05-11
API missing from several components


kresimir-coko:
Some of our components are missing or have incomplete API tables as taken from #3197 .

Here's a list of the components that fall under this umbrella:

- Form
- Input
- Link

kresimir-coko:
Fixed in #3221

PR #3221, opened 2020-05-12
feat(clayui.com): Improve API docs of Form, Input & Link components

kresimir-coko:
Issue #3209 

In this PR I've improved/fixed API tables for the components that need it:

- Form
- Input
- Link

Took me a while to figure out that `componentName.displayName = 'newName'` is a thing that makes these things work.


kresimir-coko:
Hey @bryceosterhaus I've updated the `displayName`'s across the board and tested if it still works, so we should be good to go 😄

PR #3223, opened 2020-05-12
chore: lower size limit boundary and remove bundle step

bryceosterhaus:
I realized that there was no need to add a bundle step since size-limit can already test that with a vanilla webpack bundle. I think using size-limit alone will give a more realistic idea of sizes.

I also realized you can run `yarn sizeLimit --why` to see a graph of all our dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,14 +3,17 @@
 	"repository": "https://github.com/liferay/clay",
 	"size-limit": [
 		{
-			"path": "lib/all.js",
-			"limit": "150 kb"
+			"path": "all.js",
+			"limit": "140 kb",
+			"ignore": [
+				"react",
+				"react-dom"
+			]
 		}
 	],
 	"scripts": {
 		"build:csb": "lerna run --no-bail --ignore \"clayui.com\" build && lerna run --no-bail --ignore \"clayui.com\" build:types",
 		"build": "lerna run build && lerna run build:types",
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
minimal change, going to merge.

package.json
```diff
@@ -3,14 +3,17 @@
 	"repository": "https://github.com/liferay/clay",
 	"size-limit": [
 		{
-			"path": "lib/all.js",
-			"limit": "150 kb"
+			"path": "all.js",
+			"limit": "140 kb",
+			"ignore": [
```
jbalsas, line 8: What does "ignore" here means exactly in the context of `size-limit`?

bryceosterhaus: @jbalsas 

It removes the dependency from the generated size. For example, if we don't have these two ignored we get this for node_modules
Screen Shot 2020-05-13 at 10 33 13 AM


Whereas with them ignored we get
Screen Shot 2020-05-13 at 10 32 36 AM

Reason behind ignoring them is because react and react-dom are peer dependencies of these packages.

jbalsas: Ok, that works for now... but we need to be careful about marketing if it ever comes to that, since it might feel like we're hiding needed deps. 

If you're bringing in Clay to a React project, then by all means, this is the library footprint. But to start from scratch, you'd need to throw in some additional bytes.

Let's focus on that tiny `moment` problem, then :D

PR #3224, opened 2020-05-13
feat(clayui.com): Improve headings & content structure of components

### LINKED ISSUE #3211: Several components need better content structure i.e. headings
Here's a list of components that need better content structure as taken from #3197.

- Checkbox
- Management Toolbar
- Select Box
- Sticker
- Table
- Tabs

kresimir-coko:
Fixes #3211 

Hey @bryceosterhaus, I also added another commit in here. Components in question:
- Checkbox
- Management Toolbar
- Select Box
- Sticker
- Table
- Tabs

### API Table
It adds `false` in the API Table under `required` for optional props. This is a proposition and feel free to not merge it. I think it will make the table look less empty at the cost of visual clarity where the required ones are clearly distinct from the optional ones. 
Another idea is to show both true/false for required/optional props but add some color to differentiate between them.

### Tabs
I've reduced the inside content of the tabs here in order to streamline the example so it focuses on the usage instead of having a lot of content which muddied up the example

### DIFF
--- a/clayui.com/content/docs/components/css-checkbox-radio.md
+++ b/clayui.com/content/docs/components/css-checkbox-radio.md
@@ -23,11 +23,11 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/forms/radio-c
 Default checkboxes and radios are improved upon with the help of `.form-check`, **a single class for both input types that improves the layout and behavior of their HTML elements**. Checkboxes are for selecting one or several options in a list, while radios are for selecting one option from many.
 By [Bootstrap](https://getbootstrap.com/docs/4.1/components/forms/#checkboxes-and-radios)
 
-# Default
+## Default
 
 By default, any number of checkboxes and radios that are immediate sibling will be vertically stacked and appropriately spaced with `.form-check`.
 
-## Default Checkboxes
+### Checkboxes
 
 <div class="sheet-example">
 	<div class="form-check">
@@ -70,7 +70,7 @@ By default, any number of checkboxes and radios that are immediate sibling will 
 </div>
 ```
 
-## Default Radios
+### Radios
 
 <div class="sheet-example">
 	<div class="form-check">
@@ -134,11 +134,11 @@ By default, any number of checkboxes and radios that are immediate sibling will 
 </div>
 ```
 
-# Inline(#css-inline)
+## Inline(#css-inline)
 
 Group checkboxes or radios on the same horizontal row by adding `.form-check-inline` to any `.form-check`.
 
-## Checkbox
+### Checkbox
 
 <div class="sheet-example">
 	<div class="form-check form-check-inline">
@@ -181,7 +181,7 @@ Group checkboxes or radios on the same horizontal row by adding `.form-check-inl
 </div>
 ```
 
-## Radio
+### Radio
 
 <div class="sheet-example">
 	<div class="form-check form-check-inline">
@@ -226,7 +226,7 @@ Group checkboxes or radios on the same horizontal row by adding `.form-check-inl
 </div>
 ```
 
-# Disabled
+## Disabled
 
 Disable checkboxes or radios by adding a `disabled` prop.
 
@@ -283,7 +283,7 @@ Disable checkboxes or radios by adding a `disabled` prop.
 </div>
 ```
 
-# Without labels
+## Without labels
 
 Remember to still provide some form of label for assistive technologies (for instance, using `aria-label`).
 
@@ -323,7 +323,7 @@ Remember to still provide some form of label for assistive technologies (for ins
 </div>
 ```
 
-# Custom
+## Custom
 
 The two ways for you to structure the marking of a Checkbox and Radio:
 
@@ -355,7 +355,7 @@ Using the `id` binding engine with `<label />`and `<input />`.
 </div>
 ```
 
-## Custom Checkboxes
+### Checkboxes
 
 <div class="sheet-example">
 	<div class="custom-control custom-checkbox">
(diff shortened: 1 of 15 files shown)


packages/clay-form/CHECKBOX.mdx
```diff
@@ -32,22 +41,36 @@ Use [`checked`](#api-checked) property to check or not the checkbox.
...
 
 <CheckboxContainerProps />
 
+### Checkbox with a label
+
 Use [`label`](#api-label) property to describe what the checkbox is for.
 
 <CheckboxLabel />
 
+### Inline Checkboxes
```
bryceosterhaus, line 72: For all of these titles, I think adding in "Checkbox" is redundant since they are on the checkbox page. They could simply be `Container Props... Label... Inline...`

packages/clay-form/CHECKBOX.mdx
```diff
@@ -32,22 +41,36 @@ Use [`checked`](#api-checked) property to check or not the checkbox.
...
+
+Here you can see some options or ideas on how to extend our Checkbox component to fit your use.
+
+### Indeterminate Checkbox
+
 Use [`indeterminate`](#api-indeterminate) property for making the checkbox indeterminate.
 
 <CheckboxIndeterminate />
 
+### Checkbox with a callback
```
bryceosterhaus, line 54: This should likely be our first example since we expect users to always use these components in a controlled way

kresimir-coko:
Great comments @bryceosterhaus I was looking into what was already there and just improved it a bit. I like removing Clay prefixes and component names from the headings and navigation, makes it look more clear.

kresimir-coko:
I've made the requested changes @bryceosterhaus plus a bit more, still contained within the confines of the issue 😄 

One thing to note, the links to rows in the API table don't really work. For example `the [`position`](#api-position) property.` doesn't scroll down to the API table. We can open an issue to fix all of those, but before that what would be the correct sytax to get the links to work?

clayui.com/content/docs/components/css-checkbox-radio.md
```diff
@@ -24,15 +20,15 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/forms/radio-c
...
-
 Default checkboxes and radios are improved upon with the help of `.form-check`, **a single class for both input types that improves the layout and behavior of their HTML elements**. Checkboxes are for selecting one or several options in a list, while radios are for selecting one option from many.
 By [Bootstrap](https://getbootstrap.com/docs/4.1/components/forms/#checkboxes-and-radios)
 
-### Default (stacked)
+# Default
 
 By default, any number of checkboxes and radios that are immediate sibling will be vertically stacked and appropriately spaced with `.form-check`.
 
+## Default Checkboxes
```
bryceosterhaus, line 30: Since this is nested under "Default", we can avoid redundancy by just calling it "Checkboxes".

clayui.com/content/docs/components/css-checkbox-radio.md
```diff
@@ -74,6 +70,8 @@ By default, any number of checkboxes and radios that are immediate sibling will
 </div>
 ```
 
+## Default Radios
```
bryceosterhaus, line 73: Redundancy

clayui.com/content/docs/components/css-checkbox-radio.md
```diff
@@ -24,15 +20,15 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/forms/radio-c
 	Don't forget to check <a href="https://www.w3.org/TR/wai-aria-practices/#checkbox">WAI-ARIA</a> accessibility pratices for checkboxes when writting your markup.
 </div>
 
-## Checkboxes and Radios
-
 Default checkboxes and radios are improved upon with the help of `.form-check`, **a single class for both input types that improves the layout and behavior of their HTML elements**. Checkboxes are for selecting one or several options in a list, while radios are for selecting one option from many.
 By [Bootstrap](https://getbootstrap.com/docs/4.1/components/forms/#checkboxes-and-radios)
 
-### Default (stacked)
+# Default
```
bryceosterhaus, line 26: We shouldn't have multiple `<h1>` tags on a page. I think the main categories should be an `h2` via `## some heading` and then nested items would be an h3

packages/clay-sticker/README.mdx
```diff
@@ -14,30 +14,34 @@ import {
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
--   [User Icon](#user-icon)
+-   [Composing](#composing)
+    -   [Using Color](#using-color)
```
bryceosterhaus, line 18: I don't think these would be nested under "Composing". I would also rename it from "Using Color" to "Display Type"

bryceosterhaus: I think I would also drop the "Composing" name and use "Positioning". I would also place this part last since its likely the least consulted. Display type would be first

packages/clay-tabs/README.mdx
```diff
@@ -19,7 +21,7 @@ import {Tabs, TabsDropdown} from '$packages/clay-tabs/docs/index';
 </div>
 </div>
 
-## Introduction
+# Composing
```
bryceosterhaus, line 24: I wouldn't call this "Composing", its more of an "Introduction"

kresimir-coko:
Hey @bryceosterhaus my bad with the headings, didn't realize why we had them starting at h2 until you pointed it out.

I've removed the h1's and changed the wording like you suggested in this latest commit.

I'll open the issue regarding the links to the API table 👍

Issue #3219, opened 2020-05-12
Upper Toolbar is missing it's Markup/CSS page


kresimir-coko:
Upper Toolbar doesn't have it's Markup/CSS page.

@pat270 Do we have markup for the Upper Toolbar component?

bryceosterhaus:
I believe the css content would be similar to https://pat270.github.io/lexicon/v2.16.2/content/tbar/

kresimir-coko:
Decided not to do it, see #3248

PR #3248, opened 2020-05-20
feat(clayui.com): Add Markup/CSS page for Toolbar component

kresimir-coko:
Hey @bryceosterhaus @pat270 I think that both of you can take a look at this and let me know if I've included the examples I was supposed to, taking inspiration from [here](https://github.com/liferay/clay/issues/3219#issuecomment-627505325).

It was tricky sifting through some close-to-duplicate examples, but please do let me know if I missed something or included something that wasn't necessary.

### DIFF
--- a/clayui.com/content/docs/components/css-toolbar.md
+++ b/clayui.com/content/docs/components/css-toolbar.md
@@ -1,0 +1,640 @@
+---
+title: 'Toolbar'
+description: 'Toolbar used as part of some Clay components like Management Toolbar and Modal'
+lexiconDefinition: 'https://liferay.design/lexicon/core-components/toolbars/'
+---
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
+-   [Subnav](#subnav)
+    -   [Subnav Light](#subnav-light)
+    -   [Subnav Primary](#subnav-primary)
+    -   [Subnav Primary Disabled](#subnav-primary-disabled)
+-   [Helper Classes](#helper-classes)
+
+</div>
+</div>
+
+<div class="sheet-example">
+	<nav class="component-tbar tbar">
+		<div class="container-fluid container-fluid-max-xl">
+			<ul class="tbar-nav">
+				<li class="tbar-item">
+					<a class="component-action disabled" href="#1" role="button" tabindex="-1">
+						<svg class="lexicon-icon lexicon-icon-angle-left" focusable="false" role="presentation">
+							<use xlink:href="/images/icons/icons.svg#angle-left" />
+						</svg>
+					</a>
+				</li>
+				<li class="tbar-item tbar-item-expand">
+					<div class="tbar-section">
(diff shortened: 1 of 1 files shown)


pat270:
@kresimir-coko The CSS docs probably should go under Toolbar instead of Upper Toolbar. Upper Toolbar is specifically https://liferay.design/lexicon/core-components/toolbars/toolbar-upper/. I don't have any markup for Upper Toolbar. I remember something along the lines of component teams will handle the markup.

We have `subnav-tbar-primary` docs under https://clayui.com/docs/components/management-toolbar.html#css-markup-search and `subnav-tbar-light` under https://clayui.com/docs/components/multi-step-nav.html#simplified. I think it would be ok to duplicate docs with more detail and links to where they're used if/when we create a Toolbar page.

jbalsas:
> Upper Toolbar is specifically https://liferay.design/lexicon/core-components/toolbars/toolbar-upper/. I don't have any markup for Upper Toolbar. I remember something along the lines of component teams will handle the markup.

I think that's `info_bar` in DXP: https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib/src/main/resources/META-INF/resources/info_bar/start.jsp

Might be a good component to bring over. It's heavily used across the board...

kresimir-coko:
Alright, I'll transfer this one into markup for a (as of now) non-existant page Toolbar.

@jbalsas as for bringing over the Info Bar component. 
1. Would that be the Toolbar, Upper Toolbar, or Info Bar component? 
2. If Toolbar or Info Bar, i couldn't find it in Lexicon, so we would need them to make the design for it.
3. If Upper Toolbar, we already have that one

jbalsas:
> If Upper Toolbar, we already have that one

I think this is the case, but maybe @drakonux or @laugardie can confirm...

kresimir-coko:
I've renamed the page from `Upper Toolbar` to just `Toolbar`, to go along with @pat270's comment. 

So if this gets merged as is now:
- the `Upper Toolbar` still doesn't have it's Markup/CSS page, and 
- we'll have a new "component" `Toolbar` that doesn't have a React Component page.

drakonux:
I agree with @pat270 a generic name like [toolbar ](https://liferay.design/lexicon/core-components/toolbars/) fits better about what we are looking for. The Product teams should be able to customize it to show information or whatever they need in the product context.

PS. The upper toolbar is just a DXP convention name for a toolbar they use. We should improve these distinctions in Lexicon Site with clearer descriptions.

kresimir-coko:
Awesome @drakonux let us know when you do that so we can reflect it on Clay!

kresimir-coko:
Hey @bryceosterhaus I think you overlooked this PR or thought it wasn't ready for review because of my latest comment above, but it is ready! :)

bryceosterhaus:
Whoops! My bad. Thanks for reminding me!

PR #3256, opened 2020-05-20
fix: apply layout components throughout

### LINKED ISSUE #3255: Apply new @clayui/layout component

bryceosterhaus:
fixes #3255


packages/clay-card/src/CardWithUser.tsx
```diff
@@ -7,6 +7,7 @@ import {ClayDropDownWithItems} from '@clayui/drop-down';
 import {ClayCheckbox} from '@clayui/form';
 import ClayIcon from '@clayui/icon';
 import ClayLabel from '@clayui/label';
+import * as ClayLayout from '@clayui/layout';
```
kresimir-coko, line 10: This one is taken as an example, I've seen similar cases throughout the commit, but we seem to be importing all and then just using one subcomponent. In this case `<ClayLayout.ContentCol` on lines 140 & 163

packages/clay-management-toolbar/src/ManagementToolbar.tsx
```diff
@@ -33,7 +34,9 @@ const ClayManagementToolbar: React.FunctionComponent<IProps> & {
 			}
 		)}
 	>
-		<div className="container-fluid container-fluid-max-xl">{children}</div>
+		<ClayLayout.ContainerFluid size="xl">
```
jbalsas, line 37: Is `size` really necessary as [it's the default](https://github.com/liferay/clay/blob/master/packages/clay-layout/src/ContainerFluid.tsx#L20) already?

packages/demos/stories/FormPage.tsx
```diff
@@ -18,7 +19,7 @@ export default () => {
 	const [selectedItems, setSelectedItems] = React.useState<any>([]);
 
 	return (
-		<div className="container-fluid">
+		<ClayLayout.ContainerFluid>
```
jbalsas, line 22: Interestingly, this is would actually be `xl` because of the [defaults](https://github.com/liferay/clay/blob/master/packages/clay-layout/src/ContainerFluid.tsx#L20)

packages/demos/stories/Recharts.tsx
```diff
@@ -537,7 +538,7 @@ storiesOf('Demos|Recharts', module)
 					</div>
 				</div>
 
-				<div className="container-fluid container-view">
+				<ClayLayout.ContainerFluid view>
```
jbalsas, line 541: Same, this would actually now be `xl` because of [defaults](https://github.com/liferay/clay/blob/master/packages/clay-layout/src/ContainerFluid.tsx#L20) which might not be so bad after all 😉

packages/clay-management-toolbar/src/ManagementToolbar.tsx
bryceosterhaus: it is not necessary... i just forgot there was a default

packages/clay-card/src/CardWithUser.tsx
bryceosterhaus: Just updated to have a default export so that we don't have to do this

bryceosterhaus:
Going to merge for release

PR #3263, opened 2020-05-21
feat(@clayui/css): Move components `type`, `icons`, `aspect-ratio`, a…

### LINKED ISSUE #3191: Clay CSS move components aspect-ratio, buttons, icons, and type earlier in the import chain
See https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/_components.scss.

These components are reused in other components and should be easier to overwrite. For example, we should be able to overwrite `.btn` styles from Alerts with one class (`.alert-btn`) instead of two (`.alert .alert-btn`). The order should be:
```
@import 'components/type';

@import 'components/icons';

@import 'components/aspect-ratio';
@import 'components/buttons';
@import 'components/grid';
...
```

pat270:
…nd `buttons` earlier in the import chain so they are easier to overwrite

fixes #3191

### DIFF
--- a/packages/clay-css/src/scss/_components.scss
+++ b/packages/clay-css/src/scss/_components.scss
@@ -2,14 +2,18 @@
 	@import url($font-import-url);
 }
 
+@import 'components/_type';
+
+@import 'components/_icons';
+
+@import 'components/_aspect-ratio';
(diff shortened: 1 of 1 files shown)


pat270:
Merging @marcoscv-work

PR #3264, opened 2020-05-21
feat(@clayui/date-picker): add props to control expand functionality

### LINKED ISSUE #3262: Hide date picker after selecting date
## What is your proposal?

Add the possibility of hiding the picker automatically after selecting the date

It could reduce the number of clicks to select a date, and through that, the use of the app would become more organic to the user.

Maybe provide a property to allow this behavior

bryceosterhaus:
fixes #3262

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -58,7 +58,8 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	id?: string;
 
 	/**
-	 * Flag to indicate if date is expanded.
+	 * Flag to indicate if date is initially expanded when
+	 * `expand` and `onExpandChange` are not being used.
 	 */
 	initialExpanded?: boolean;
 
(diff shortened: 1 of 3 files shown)


packages/clay-date-picker/src/index.tsx
```diff
@@ -226,7 +245,23 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+			}
+
+			setInternalExpanded(val);
+		};
+
+		React.useEffect(() => {
+			if (expanded !== undefined) {
+				setInternalExpanded(expanded);
+			}
+		}, [expanded]);
```
matuzalemsteles, line 264: Perhaps we can decrease the use here of `useEffect` which could happen a double rendering due to the `useState` that the developer will have to use to control. I'm thinking of something like this:

```js
const expanded = externalExpanded !== undefined ? externalExpanded : internalExpanded;
const setExpanded = onExpandedChange ? onExpandedChange : setInternalExpanded;
```
`externalExpanded` can only be an alias of `expanded` prop. What do you think? it's makes sense for you?

bryceosterhaus: @matuzalemsteles let me try that out. That was my goal but my first iteration was having issues. Let me look again to see if this is possible.

Issue #2219, opened 2019-07-24
[ClayCSS] Table title horizontal alignment behavior differs from Chrome to Safari


diegonvs:
The left page is running on Chrome and right, on Safari.
<img width="1261" alt="Screen Shot 2019-07-24 at 17 21 39" src="https://user-images.githubusercontent.com/7663513/61825891-a23f4300-ae37-11e9-9656-9fd6f04edbc3.png">

Same comparison on https://next.clayui.com/docs/components/table.html

<img width="1249" alt="Screen Shot 2019-07-24 at 17 23 47" src="https://user-images.githubusercontent.com/7663513/61825961-c733b600-ae37-11e9-8e5d-1342cf5a5d41.png">

Debugging, we are computing `text-align: start;` on Safari and `text-align: inherit;` on Chrome. Disabling this property on Safari's debugger, apparently fixed it. 
https://recordit.co/MxOMjKswBy

/cc @pat270 @brunofarache

brunofarache:
Thanks for creating this @diegonvs!

Fix should be merged to DXP as it affects all tables.

pat270:
This is caused by https://github.com/liferay/clay/issues/1743

zhangyan0701:
Create https://issues.liferay.com/browse/LPS-100174 for this issue.

PR #3271, opened 2020-05-21
feat(@clayui/css): Absorb Bootstrap 4 _reboot.scss

pat270:
issue #3149

List of changes to Bootstrap's Reboot:
- We don't need to set `font-family`, `line-height` on the `html` element already set on `body`
- Don't set `text-align: left` on `body` due to RTL issues. See #1743
- Remove `:focus-visible` future-proof rule (very experimental https://caniuse.com/#search=focus-visible)
- Remove uses of Bootstrap's `font-size` and `hover` mixin, we don't support it in Clay CSS
- Remove Bootstrap's `text-align` reset on `th`. See #2219
- Remove Bootstrap's "hand" cursor to non-disabled button elements. We don't support `$enable-pointer-cursor-for-buttons`. We can change the cursor to whatever we want in Clay CSS.
- Use selector `[href]` for styling anchor links instead of `a` so we don't need to overwrite placeholder anchor styles with `a:not([href])`. Styling `<a tabindex="0">a link</a>` should be done with a class like `link-primary`.
- Combine `body` and `a` styles from Type

### DIFF
--- a/packages/clay-css/src/scss/components/_reboot.scss
+++ b/packages/clay-css/src/scss/components/_reboot.scss
@@ -1,0 +1,540 @@
+// Reboot
+//
+// Normalization of HTML elements, manually forked from Normalize.css to remove
+// styles targeting irrelevant browsers while applying new styles.
+//
+// Normalize is licensed MIT. https://github.com/necolas/normalize.css
+
+*,
+*::before,
+*::after {
+	box-sizing: border-box;
+}
+
+html {
+	// Change the default font family in all browsers
+
+	font-family: sans-serif;
+
+	// Correct the line height in all browsers
+
+	line-height: 1.15;
+
+	// Prevent adjustments of font size after orientation changes in IE on Windows Phone and in iOS
+
+	-webkit-text-size-adjust: 100%;
+
+	// Change the default tap highlight to be completely transparent in iOS
+
+	-webkit-tap-highlight-color: rgba($black, 0);
+}
+
+// Shim for "new" HTML5 structural elements to display correctly (IE10, older browsers)
+
+article,
+aside,
+figcaption,
+figure,
+footer,
+header,
+hgroup,
+main,
+nav,
+section {
+	display: block;
+}
+
+body {
+	// As a best practice, apply a default `background-color`
+
+	background-color: $body-bg;
+	color: $body-color;
+	font-family: $font-family-base;
+
+	@include font-size($font-size-base);
+
+	font-weight: $font-weight-base;
+	line-height: $line-height-base;
+
+	// Remove the margin in all browsers
+
+	margin: 0;
+
+	// Set an explicit initial text-align value so that we can later use the `inherit` value on things like `<th>` elements
+
+	text-align: left;
+}
+
+// Future-proof rule: in browsers that support :focus-visible, suppress the focus outline
+// on elements that programmatically receive focus but wouldn't normally show a visible
+// focus outline. In general, this would mean that the outline is only applied if the
+// interaction that led to the element receiving programmatic focus was a keyboard interaction,
+// or the browser has somehow determined that the user is primarily a keyboard user and/or
+// wants focus outlines to always be presented.
+//
+// See https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
+// and https://developer.paciellogroup.com/blog/2018/03/focus-visible-and-backwards-compatibility/
+
+[tabindex='-1']:focus:not(:focus-visible) {
+	outline: 0 !important;
+}
+
+hr {
+	// Add the correct box sizing in Firefox
+
+	box-sizing: content-box;
+	height: 0;
+
+	// Show the overflow in Edge and IE
+
+	overflow: visible; // 2
+}
+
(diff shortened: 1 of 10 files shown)


packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -0,0 +1,477 @@
...
+	// Prevent adjustments of font size after orientation changes in IE on Windows Phone and in iOS
+
+	-webkit-text-size-adjust: 100%;
+
+	// Change the default tap highlight to be completely transparent in iOS
+
+	-webkit-tap-highlight-color: rgba($black, 0);
+}
+
+// Shim for "new" HTML5 structural elements to display correctly (IE10, older browsers)
```
jbalsas, line 24: Based on 

> [...] manually forked from Normalize.css to remove styles targeting irrelevant browsers while applying new styles

Maybe we can drop this?

packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -0,0 +1,477 @@
...
+// bottom margin to use `rem` units instead of `em`.
+
+p {
+	margin-bottom: $paragraph-margin-bottom;
+	margin-top: 0;
+}
+
+abbr[title],
+abbr[data-original-title] {
+	// Remove the bottom border in Firefox 39-
```
jbalsas, line 100: Firefox 39-? 🤔

packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -0,0 +1,477 @@
...
+sub {
+	bottom: -0.25em;
+}
+
+sup {
+	top: -0.5em;
+}
+
+[href] {
+	// Remove the gray background on active links in IE 10
```
jbalsas, line 184: IE10?

packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -0,0 +1,477 @@
...
+// Figures
+
+figure {
+	margin: 0 0 1rem;
+}
+
+// Images and content
+
+img {
+	// Remove the border on images inside links in IE 10-
```
jbalsas, line 233: IE10

packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -0,0 +1,477 @@
...
+[type='button']::-moz-focus-inner,
+[type='reset']::-moz-focus-inner,
+[type='submit']::-moz-focus-inner {
+	padding: 0;
+	border-style: none;
+}
+
+input[type='radio'],
+input[type='checkbox'] {
+	// Add the correct box sizing in IE 10-
```
jbalsas, line 346: More IE10-

packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -0,0 +1,477 @@
...
+}
+
+template {
+	// Add the correct display in IE
+
+	display: none;
+}
+
+// Always hide an element with the `hidden` HTML attribute (from PureCSS).
+// Needed for proper display in IE 10-.
```
jbalsas, line 473: This seems dangerous... but IE10- 🤷

packages/clay-css/src/scss/components/_reboot.scss
pat270: I wanted to remove these, but I don't have a working IE11 browser to test right now. I don't fully trust these comments.

pat270: I just got my vm to work. I'll test and remove.

pat270:
@jbalsas I went ahead and removed as much as I could. Here is a list:
- don't need `display: block` for `article`, `aside`, `figcaption`, `figure`, `footer`, `header`, `hgroup`, `nav`, `section`
- remove resets for `hr` behaves the same in all our supported browsers
- `abbr[title]` removes `border-bottom` CSS reset
- `address` removes `line-height: inherit`
- `small` removes `font-size` reset Chrome, FF, Safari (13.333px), IE (13.26px). Bootstrap also sets `font-size` in _type.scss.
- `sub` and `sup` removes `font-size: 75%` reset. Chrome, FF, Safari, IE all have similar font sizes
- links removes `background-color: transparent` reset. IE11 doesn't set a gray background on active links like IE10
- `pre`, `code`, `kbd`, `samp` removes `font-size` reset, Chrome, FF, Safari, and IE inherit from parent element
- `img` removes `border-style` reset on image inside a link. IE11 doesn't add border around image like IE10
- `table` removes opinionated `border-collapse: collapse` reset
- `button` removes `border-radius` reset and removes `button:focus` focus reimplementation for FF and IE. These don't normalize anything. The UX without is better.
- `button` and `select` remove `text-transform: none`. Firefox now sets this by default.
- `checkbox` and `radio` remove resets for IE10 and below
- `textarea` move `resize: vertical` to `textarea.form-control` this is specific to Bootstrap's flavor of textarea which is `width: 100%` by default.
- `[hidden]` removes `display: none` reset for IE10 and below

marcoscv-work:
did we make the required changes?

PR #3275, opened 2020-05-22
fix(clayui.com): Fix links that pointed to APITable but didn't work

### LINKED ISSUE #3228: Links to API Table rows on clayui.com don't lead anywhere

No

Go to https://clayui.com/docs/components/button-group.html and click on `spaced` in `Use spaced property to create spacing between buttons.`

It's supposed to scroll down to the API table where the `spaced` prop is mentioned

kresimir-coko:
Fixes #3228 

Hey @bryceosterhaus can you please review this.

I've done the following as part of this PR:

- Add `id="api-${key}"` to the `tr` that's being constructed by the `gatsby-remark-api-table` plugin
- Add all the DatePicker subcomponents to the APITable of DatePicker because some of them were being referenced so I assumed that was the right course
    - Note: most of these are missing descriptions/JSDoc
- Reword a couple sentences in Button and Date Picker to match these, now fixed, links to API props

### DIFF
--- a/packages/clay-date-picker/README.mdx
+++ b/packages/clay-date-picker/README.mdx
@@ -41,7 +41,7 @@ The component is treated as controlled, use the [`onValueChange`](#api-onValueCh
 <div class="clay-site-alert alert alert-warning">
 	<strong class="lead">Warning</strong>
 	You can set a range of years using the API <code class="language-text">
-		years
+		<a href="#api-years">years</a>
 	</code>, which can be displayed in the Date Picker, if the user enters a year
 	that is not within the range will be treated as an invalid date.
 </div>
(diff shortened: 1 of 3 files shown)


packages/clay-date-picker/README.mdx
```diff
@@ -89,3 +97,35 @@ If you want to expand or close the picker from outside of the component, use the
...
+
+<div>[APITable "clay-date-picker/src/DateNavigation.tsx"]</div>
+
+### DatePicker.DayNumber
+
+<div>[APITable "clay-date-picker/src/DayNumber.tsx"]</div>
+
+### DatePicker.DaysTable
+
+<div>[APITable "clay-date-picker/src/DaysTable.tsx"]</div>
```
matuzalemsteles, line 111: hey @kresimir-coko, the ideal would be not to expose these APIs, because they are internal in this case we are giving the impression that we have the components `DatePicker.DateNavigation` ... available, which is not true, all the APIs available in this case are on the `index.tsx` the internal APIs, we were avoiding adding JSDoc.

I commented on this in some other issue. Many APIs that reference other interfaces are not shown because `react-docgen` cannot do this, there is a PR that has been implementing this for a long time https://github.com/reactjs/react-docgen/pull/352, I'm following but I don't think it will be merged apparently 🤷‍♂️

matuzalemsteles, line 111: hey @kresimir-coko, the ideal would be not to expose these APIs, because they are internal in this case we are giving the impression that we have the components `DatePicker.DateNavigation` ... available, which is not true, all the APIs available in this case are on the `index.tsx` the internal APIs, we were avoiding adding JSDoc.

I commented on this in some other issue. Many APIs that reference other interfaces are not shown because `react-docgen` cannot do this, there is a PR that has been implementing this for a long time https://github.com/reactjs/react-docgen/pull/352, I'm following but I don't think it will be merged apparently 🤷‍♂️

kresimir-coko: Okay @matuzalemsteles, good to know! I'll remove the private API tables and also remove links to them.

kresimir-coko:
Hey @matuzalemsteles I've removed the DatePicker APITables and a couple links that aren't pointing anywhere because they were referencing props from the APITables we don't wanna show.

kresimir-coko:
@bryceosterhaus I think this is ready to be merged, in case you forgot ❤️

PR #3280, opened 2020-05-26
feat(clayui.com): Add tabular display for different code examples depending on language

kresimir-coko:
Here's my initial implementation for tabs in code examples. The goal is to offer both React and JSP code examples in a single view, by switching tabs.

### For those that just want the visual example
JSP tab example

### The approach
I utilized `ClayTabs` inside of ClayAlert's docs file, meaning that if we decided to take this approach we would have to change a lot of files in order to get the tabbing functionality to all our component docs. The alternative might be wrapping it at a lower level, perhaps in [templates/docs.js](https://github.com/liferay/clay/blob/master/clayui.com/src/templates/docs.js#L296)

### The alternative
In case you haven't noticed, this implementation swaps both the display(`.sheet-example`) and the code, we might want to just change the code but leave the display same for both React and JSP implementation

### DIFF
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -38,6 +38,7 @@ const Editor = ({
 }) => {
 	let reactSnippet = {
 		code: '',
+		format: true,
 		name: 'React',
 	};
 
@@ -45,14 +46,14 @@ const Editor = ({
 		const codeToFormat = [...code];
 
 		codeToFormat.forEach((snippet) => {
-			if (snippet.name !== 'JSP') {
+			if (snippet.format) {
 				snippet.code = formatCode(snippet.code);
 			}
 		});
 
 		code = codeToFormat;
 
-		reactSnippet = code.find((snippet) => snippet.name === 'React');
+		reactSnippet = code[0];
 	} else {
 		code = formatCode(code);
 
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -5,6 +5,7 @@
 
 import ClayAlert from '@clayui/alert';
 import {ClayButtonWithIcon} from '@clayui/button';
+import ClayTabs from '@clayui/tabs';
 import parserBabylon from 'prettier/parser-babylon';
 import prettier from 'prettier/standalone';
 import React, {useContext, useState} from 'react';
@@ -14,6 +15,20 @@ import theme from '../../utils/react-live-theme';
 
 const spritemap = '/images/icons/icons.svg';
 
+function formatCode(code) {
+	try {
+		return prettier.format(code, {
+			parser: 'babel',
+			plugins: [parserBabylon],
+		});
+	} catch (e) {
+		// eslint-disable-next-line
+		console.log(e);
+
+		return;
+	}
+}
+
 const Editor = ({
 	code,
 	disabled = false,
@@ -21,21 +36,35 @@ const Editor = ({
 	preview = true,
 	scope = {},
 }) => {
-	try {
-		code = prettier.format(code, {
-			parser: 'babel',
-			plugins: [parserBabylon],
+	let reactSnippet = {
+		code: '',
+		name: 'React',
+	};
+
+	if (Array.isArray(code)) {
+		const codeToFormat = [...code];
+
+		codeToFormat.forEach((snippet) => {
+			if (snippet.name !== 'JSP') {
+				snippet.code = formatCode(snippet.code);
+			}
 		});
-	} catch (e) {
-		// eslint-disable-next-line
-		console.log(e);
+
+		code = codeToFormat;
+
+		reactSnippet = code.find((snippet) => snippet.name === 'React');
+	} else {
+		code = formatCode(code);
+
+		reactSnippet.code = code;
 	}
 
 	const [collapseCode, setCollapseCode] = useState(false);
+	const [activeTabKeyValue, setActiveTabKeyValue] = React.useState(0);
 
 	return (
 		<LiveProvider
-			code={code}
+			code={reactSnippet.code}
 			disabled={disabled}
 			noInline
 			scope={{spritemap, useContext, useState, ...scope}}
@@ -78,12 +107,64 @@ const Editor = ({
 						title={collapseCode ? 'Expand' : 'Collapse'}
 					/>
 
-					{!collapseCode && (
+					{Array.isArray(code) ? (
 						<>
-							{imports && <LiveEditor disabled value={imports} />}
+							<ClayTabs modern>
+								{code.map((snippet, index) => (
+									<ClayTabs.Item
+										active={activeTabKeyValue === index}
+										innerProps={{
+											'aria-controls': `tabpanel-${snippet.name}`,
+										}}
+										key={snippet.name}
+										onClick={() =>
+											setActiveTabKeyValue(index)
+										}
+									>
+										{snippet.name}
+									</ClayTabs.Item>
+								))}
+							</ClayTabs>
 
-							<LiveEditor />
+							{!collapseCode && (
+								<ClayTabs.Content
+									activeIndex={activeTabKeyValue}
+									fade
+								>
+									{code.map((snippet) => (
+										<ClayTabs.TabPane
+											aria-labelledby={`tab-${snippet.name}`}
+											key={snippet.name}
+										>
+											{snippet === reactSnippet &&
+												imports && (
+													<LiveEditor
+														disabled
+														value={imports}
+													/>
+												)}
+
+											<LiveEditor
+												disabled={
+													snippet !== reactSnippet
+												}
+												value={snippet.code}
+											/>
+										</ClayTabs.TabPane>
+									))}
+								</ClayTabs.Content>
+							)}
 						</>
+					) : (
+						!collapseCode && (
+							<>
+								{imports && (
+									<LiveEditor disabled value={imports} />
+								)}
+
+								<LiveEditor />
+							</>
+						)
 					)}
 				</div>
 			</div>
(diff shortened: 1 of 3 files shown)


kresimir-coko:
I've made drastic changes to my initial approach, here they are:

- Add infrastructure for tabular display in one location, so that we don't have to change a billion files
- Add an example of code to one of the Alert examples
- Make sure that the tabs get rendered only when JSP code example is provided

Overall I think this is the way to go, barring any changes to the way I used the Tabs component.

### Here's how it looks in action
JSP tab example v2

clayui.com/src/components/Editor/index.js
```diff
@@ -78,12 +82,62 @@ const Editor = ({
 						title={collapseCode ? 'Expand' : 'Collapse'}
 					/>
 
-					{!collapseCode && (
+					{jspCode ? (
 						<>
-							{imports && <LiveEditor disabled value={imports} />}
+							<ClayTabs modern>
+								<ClayTabs.Item
+									active={activeTabKeyValue == 0}
```
bryceosterhaus, line 89:
```suggestion
									active={activeTabKeyValue === 0}
```

lets make sure to always use triple equals `===`, otherwise this can lead to hard to diagnose bugs.

clayui.com/src/components/Editor/index.js
```diff
@@ -78,12 +82,62 @@ const Editor = ({
...
+										'aria-controls': 'tabpanel-react',
+									}}
+									onClick={() => setActiveTabKeyValue(0)}
+								>
+									{'React'}
+								</ClayTabs.Item>
 
-							<LiveEditor />
+								<ClayTabs.Item
+									active={activeTabKeyValue == 1}
```
bryceosterhaus, line 99:
```suggestion
									active={activeTabKeyValue === 1}
```

packages/clay-alert/docs/index.js
```diff
@@ -27,11 +27,25 @@ render(<Component />);`;
 const alertImportsCode = `import React from 'react';
 import ClayAlert from '@clayui/alert';`;
 
+const AlertJSPCode = `<clay:alert
```
bryceosterhaus, line 30: Do we need to also provide a snippet of how to import the clay taglibs? Or would that be too redundant? We could possibly put that into on a different JSP specific page on how to get started with the taglibs.

clayui.com/src/components/Editor/index.js
kresimir-coko: Thought so, the [Tabs docs](https://clayui.com/docs/components/tabs.html) has double equals, I'll change that too then

packages/clay-alert/docs/index.js
kresimir-coko: I think that we should have a JSP specific page, a central area instead of polluting the examples

jbalsas: Should go in the snippets, basically:

```
<%@ taglib uri="http://liferay.com/tld/clay" prefix="clay" %>
```

kresimir-coko: @jbalsas  Is our pattern in DXP is to include the taglib inside `init.jsp` once and then use `clay:*` in whichever JSP imports `init.jsp`?

If so, maybe we don't want to leave the impression upon the user they need to import the taglib whenever they use a new `clay:*` tag. Probably even, the module they're working on already has this import.

jbalsas: I don't think this is an issue. DPX features some heavy SF rules and such trivial issue will be caught instantly and never forgotten. 

Let's not mud things here based on DXP conventions. I'm sure engineers will know how to adapt an import line in one jsp.

kresimir-coko: Alright, sure! It is one line of code after all so I'm sure it will be fine 👍

kresimir-coko:
@bryceosterhaus I've added triple equals, the disable prop to the JSP editor, and the import taglib line that @jbalsas recommended.

I haven't done anything regarding the design of the `disabled` editor, I looked briefly into it and I didn't come up with a quick solution so I'll leave that for its own issue, I think that's better than to increase the scope of this one.

jbalsas:
That's really cool @kresimir-coko!! 

One thing I realized is that the API might have slight differences between React and JSP components. As much as we try, there are some reserved words we can't use such as `float` or `className`... I don't think we need to add this right now, but we might need to add an API tab as well in the future. Let's just keep this in the backburner for now!

clayui.com/src/components/Editor/index.js
```diff
@@ -18,6 +19,7 @@ const Editor = ({
 	code,
 	disabled = false,
 	imports,
+	jspCode,
```
bryceosterhaus, line 22: I was thinking, instead of having a jsp specific prop, what if we made the `code` prop more flexible, meaning it could handle either a single block of code or an array of objects that have `{name, code}`. This would give us flexibility to add as many named tabs as we want. We could end up having

```js
[{
    name: 'React',
    code: '<SomeComponent'  
},
{
    name: 'JSP',
    code: '<clay-taglib />'  
},
{
    name: 'CSS',
    code: '<div class="container></div>  
}]
```

kresimir-coko:
Hey @bryceosterhaus your suggestion for a more flexible implementation was great, so I made it work! 💪 

I've had to force push due to the amount of changes, but the code is still split into 3 commits for easier reviewing.

🧪  I tested it locally with a 3rd tab just to see if it's at least somewhat future-proof.
🎨   I left the design of the disabled editor for it's own issue down the road, to not clutter this one. 
💭   I'm excited about this feature, it also warmed up my JS muscles, which is always good, so let me know if you have any more ideas for this.
👍   @jbalsas this new implementation will also take care of your comment (at least the part where we can add as many tabs as we want): 
> I don't think we need to add this right now, but we might need to add an API tab as well in the future. Let's just keep this in the backburner for now!

❓ Do you have any opinions on how we should approach adding all these JSP snippets? Splitting them into a billion issues doesn't seem that feasible, maybe something like I did with the docs table, one parent issue with separate PRs for each component?

clayui.com/src/components/Editor/index.js
```diff
@@ -14,28 +15,56 @@ import theme from '../../utils/react-live-theme';
...
+	let reactSnippet = {
+		code: '',
+		name: 'React',
+	};
+
+	if (Array.isArray(code)) {
+		const codeToFormat = [...code];
+
+		codeToFormat.forEach((snippet) => {
+			if (snippet.name !== 'JSP') {
```
bryceosterhaus, line 48: Instead of a strict check here, we can have include a `format` boolean in our code objects that default to true. And then for jsp or other code we don't want formatted, we just would have to pass `format: false`

clayui.com/src/components/Editor/index.js
```diff
@@ -14,28 +15,56 @@ import theme from '../../utils/react-live-theme';
...
+				snippet.code = formatCode(snippet.code);
+			}
 		});
-	} catch (e) {
-		// eslint-disable-next-line
-		console.log(e);
+
+		code = codeToFormat;
+
+		reactSnippet = code.find((snippet) => snippet.name === 'React');
```
bryceosterhaus, line 55: Its also odd that we are checking for the name of the code here specifically against react. I would expect this component to just use the first item in the array of code objects.

packages/clay-alert/docs/index.js
```diff
@@ -27,11 +27,30 @@ render(<Component />);`;
 const alertImportsCode = `import React from 'react';
 import ClayAlert from '@clayui/alert';`;
 
+const AlertJSPCode = `<%@ taglib uri="http://liferay.com/tld/clay" prefix="clay" %>
```
bryceosterhaus, line 30: Lets remove the jsp examples for now and we can figure out how we want to add those later

clayui.com/src/components/Editor/index.js
kresimir-coko: Alright, that was my initial approach too, but then I assumed that checking for the name would be more robust so I changed it, in case someone comes in and doesn't know that the React snippet must be at index `0`. @matuzalemsteles got any thoughts on this one?

kresimir-coko:
Hey @bryceosterhaus I've made the requested changes and put them in the 2 most recent commits.

clayui.com/src/components/Editor/index.js
matuzalemsteles: Yeah, I would agree that the use of the name would be better here, just in case someone puts the JSP snippet at index `0` 🤷‍♂️, but just getting the index zero is simpler too but doing this I would create a constant with `const REACT = 0;` for when we look at this stretch in the future, it will be clearer.

clayui.com/src/components/Editor/index.js
```diff
@@ -78,12 +108,64 @@ const Editor = ({
...
+								<ClayTabs.Content
+									activeIndex={activeTabKeyValue}
+									fade
+								>
+									{code.map((snippet) => (
+										<ClayTabs.TabPane
+											aria-labelledby={`tab-${snippet.name}`}
+											key={snippet.name}
+										>
+											{snippet === reactSnippet &&
```
bryceosterhaus, line 140: Note that we shouldn't do equality values with objects since they can easily be flakey if they are mutated

kresimir-coko:
> @bryceosterhaus: One thing to note is that the LiveEditor doesn't work if there are multiple snippets

I just tested it out and you're right, got any ideas on how we can make it work? I can work on fixing this, but atm I don't have any direction in my head as to what might fix it.

bryceosterhaus:
@kresimir-coko I think you'll have to add internal state to track the changes so that the component is properly controlled. See the API I linked above.

kresimir-coko:
Hey @bryceosterhaus I've added a new commit here, in which I've made changes to the `defaultSnippet` object, ran `yarn format` this time, and added state handling to the `LiveEditor` that needs it.

bryceosterhaus:
Looks good! I added some changes on top of this. check them out at https://github.com/liferay/clay/pull/3308

I'll leave some comments explaining as well

Issue #3281, opened 2020-05-26
Change ClayAlert default autoClose interval to 10000


wincent:
[Lexicon says](https://liferay.design/lexicon/core-components/alerts/) that toasts should stay open for 10 seconds if there is an associated action, or 5 seconds otherwise. But for some reason, our default interval [is 8 seconds](https://github.com/liferay/clay/blob/8b8776107c1858f3517d4508342fa88860d9cde2/packages/clay-alert/src/index.tsx#L17) and has been [ever since we added it](https://github.com/liferay/clay/commit/95342d26307f72a1078afd3d60f61e307bd3d95a). This makes me think that maybe the Lexicon interval changed at some point, or maybe we didn't know what to use as a default — whatever the explanation, what do you think about bumping the default to 10000?

Noticed while reviewing [a PR in liferay-portal](https://github.com/wincent/liferay-portal/pull/297/files) that we have a bunch of components that specify 10000 as an interval, for compliance with Lexicon, and we could simplify all those if we change the default here.

matuzalemsteles:
Hey @wincent, I remember that when we started to migrate to React, we rely on Lexicon documents https://docs.google.com/document/d/1mR8k7TE_5s6wzn94HtM7kB3NYVItDMoi4RkKZ0yRezw/edit (Maybe it became obsolete). Apparently that must have changed and we didn't know! But if this is the new guideline, I don't see why not follow it.

kresimir-coko:
Fixed in #3287

PR #3287, opened 2020-05-27
fix(@clayui/alert): Make the default autoClose interval 10 seconds instead of 8

kresimir-coko:
As per @wincent's report: #3281 I'm making the default `autoClose` interval 10 seconds instead of 8. Pretty self-explanatory, I've checked the docs, stories and tests for mention of the interval but found nothing, so this should all be self-contained within this one file.

### DIFF
--- a/packages/clay-alert/src/index.tsx
+++ b/packages/clay-alert/src/index.tsx
@@ -14,7 +14,7 @@ import ToastContainer from './ToastContainer';
 const useAutoClose = (autoClose?: boolean | number, onClose = () => {}) => {
 	const startedTime = React.useRef<number>(0);
 	const timer = React.useRef<number | undefined>(undefined);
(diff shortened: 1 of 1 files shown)


packages/clay-alert/src/index.tsx
```diff
@@ -61,7 +61,7 @@ export interface IClayAlertProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
 	 * Flag to indicate alert should automatically call `onClose`. It also
 	 * accepts a duration(in ms) which indicates how long to wait. If `true`
-	 * is passed in, the timeout will be 8000ms.
+	 * is passed in, the timeout will be 10000ms.
```
wincent, line 64: While you're here, you may as well fix the typo on the line above (missing space before the `(`).

bryceosterhaus:
lgtm! ill follow up with a commit adding the space

Issue #3259, opened 2020-05-20
Simplify automation to reduce build time


bryceosterhaus:
Currently we are relying on netlify to build our sites. I think we can minimize our build times if we use gh actions instead. This would also help give us earlier options to bail on the builds and making resource heavy tasks dependent on CI checks.

Example idea of workflow 
![Screen Shot 2020-05-20 at 3 54 18 PM](https://user-images.githubusercontent.com/6843530/82505290-2b53f480-9ab2-11ea-8c76-bcebc45a239c.png)

bryceosterhaus:
possibly even thinking of putting tests dependent on build and then the other 5 tasks dependent on unit tests. This ensures we don't waste build times.

wincent:
Makes sense to defer running expensive parts of the build until the cheaper and more fundamental bits have run (ie. fail fast). At the same time, it is still handy to get feedback from multiple things without having to do multiple pushes (eg. push once, formatting fails, fix formatting, push again, check dependencies fails, fix, push again, unit tests fail, fix etc...). So I'd be inclined to run all the things that might detect obvious human errors in batch up front (eg. formatting, unit tests, and run the build to prove it completes) and then only if all that is green do the follow-up tasks like clayui.com, storybook etc.

One possible gotcha with relying heavily on actions is that there [are cuotas](https://github.com/features/actions) and they may change in the future. It's supposedly totally free and unlimited on open source repos; the limits kick in on private repos — for now! Hopefully it stays that way.

matuzalemsteles:
That would be a good approach! Just adding: there is a newer latest version of Gatsby that implements a new Incremental Build feature, apparently, it seems to be a very powerful new feature to decrease the build time I saw some projects build for production in just 2s 👀. This has support added on Netlify now, maybe we can try https://www.netlify.com/blog/2020/04/23/enable-gatsby-incremental-builds-on-netlify/.

bryceosterhaus:
> At the same time, it is still handy to get feedback from multiple things without having to do multiple pushes

yeah I totally agree. The main goal is reducing so many netlify builds since its easy to push a few times and rack up a lot of build minutes that don't serve any purpose. 

Once I get the general idea down, I'll tweak it to push the more important things up first.

bryceosterhaus:
[...]

PR #3296, opened 2020-05-28
chore: only run stats if packages have changed

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/3259

I wasn't able to fully achieve the netlify builds, or at least they weren't as helpful via github actions as they were via netlify. I did add some improvements to the CI though.


.github/workflows/main.yml
```diff
@@ -12,14 +12,10 @@ jobs:
...
-              run: |
-                  npm install -g yarn
-                  yarn install
+            - name: Install Dependencies
+              run: yarn install
             - name: Check Formatting
-              run: |
-                  yarn lint
-                  yarn format:check
+              run: yarn lint && yarn format:check
```
wincent, line 18: I don't know whether these `steps` abort on first failure, but if they don't, it wouldn't hurt for these to be two separate steps (otherwise a lint failure will hide formatting errors).

.github/workflows/main.yml
```diff
@@ -37,29 +33,42 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/setup-node@v1
+            - uses: actions/checkout@v2
+            - name: Fetch and Diff
+              id: diff
+              run: |
+                  git fetch --no-tags --depth=1 origin ${{ github.event.pull_request.base.sha }}
+                  echo "##[set-output name=packages;]$(git diff --quiet ${{ github.event.pull_request.base.sha }} '**/packages*')"
```
wincent, line 41: Clever trick! Didn't know about this syntax.

.github/workflows/main.yml
bryceosterhaus: Oh good call, I didn't think about that

.github/workflows/main.yml
bryceosterhaus: Took me way too long to figure out. Some of the hokey work arounds for gh actions are so odd

PR #3298, opened 2020-05-29
fix(@clayui/list): add dev warning to ensure child elements

### LINKED ISSUE #3295: Make sure direct descendants of ClayList are li elements
stemming from https://github.com/liferay/clay/pull/3278#discussion_r431653472

Any direct child of ClayList should be an `li` element. We can do this via typescript and also a dev warning.

bryceosterhaus:
fixes #3295

I wasn't able to find a great way to force child elements to render an `li` element. I read a handful of places that people recommended setting the type to ReactElement with specific props, and then I also added a warning for if the child is not a Item or Header.

I'm open to other ideas here as well.

### DIFF
--- a/packages/clay-list/src/List.tsx
+++ b/packages/clay-list/src/List.tsx
@@ -6,6 +6,7 @@
 import ClayLayout from '@clayui/layout';
 import classNames from 'classnames';
 import React from 'react';
+import warning from 'warning';
 
 import Header from './Header';
 import Item from './Item';
@@ -13,13 +14,19 @@ import ItemText from './ItemText';
 import ItemTitle from './ItemTitle';
 import QuickActionMenu from './QuickActionMenu';
 
+type TLIAttributes = React.ReactElement<React.HTMLAttributes<HTMLLIElement>>;
+
 interface IProps extends React.HTMLAttributes<HTMLUListElement> {
-	/**
+	children?: TLIAttributes | Array<TLIAttributes>;
+
+	/*
 	 * Flag to indicate if action items should be shown on hover.
 	 * Defaults to `true`
 	 */
 	showQuickActionsOnHover?: boolean;
 }
+
+const CLAY_REGEX = /Clay(?!ListItem|ListHeader).+/;
 
 const ClayList: React.FunctionComponent<IProps> & {
 	Header: typeof Header;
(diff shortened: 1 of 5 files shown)


packages/clay-list/src/List.tsx
```diff
@@ -41,7 +49,19 @@ const ClayList: React.FunctionComponent<IProps> & {
 				'show-quick-actions-on-hover': showQuickActionsOnHover,
 			})}
 		>
-			{children}
+			{children &&
+				React.Children.map(children, (child) => {
+					warning(
```
wincent, line 54: So `warning` has an `if (__DEV__)` check inside it (great), but it would be nice to skip the entire `React.Children.map` call when `process.env.NODE_ENV !== 'development`.

packages/clay-list/src/List.tsx
```diff
@@ -41,7 +49,19 @@ const ClayList: React.FunctionComponent<IProps> & {
...
+						typeof child.type === 'string' ||
+							// @ts-ignore
+							(child.type.displayName !== 'ClayListItem' &&
+								// @ts-ignore
+								child.type.displayName !== 'ClayListHeader'),
+						'ClayList: Direct descendant of ClayList must be either ClayList.Item or ClayList.Header.'
```
matuzalemsteles, line 60: I'm just worried about one thing, it may happen that some people pass the primitive `li` here, do we want to ignore it from the warning? in theory, we should but this could be a way to force the use of `Item` and `Header` although both cover the edge cases. What do you think, does it make sense?

bryceosterhaus: Thats probably fine to check for `li` and allow it since that is semantic. Although, now this makes me re-think even adding this because its possible they could create their own custom component that renders an `li` and then that would fail this check since the displayName would be something other than the clay components.

Unless someone has another idea of how to enforce this via ts, I'm thinking we scrape this and add a note in the docs instead. cc @wincent

wincent: This is why [I said](https://github.com/liferay/clay/pull/3278#discussion_r431653472):

> Makes me wonder whether there should be a check (only when `process.env.NODE_ENV === 'development'`) that looks at the children and makes sure there are no non-list children... (or **at least no children that are definitely the wrong type of Clay component**).

(Emphasis added). That is to say, we can warn if you use the wrong Clay component (ie. `displayName` is `ClaySomething` but not `ClayListItem` or `ClayListHeader`), so it would catch obvious errors, and remain silent about non-obvious ones. ie. some protection being better than none.

> add a note in the docs instead

Never a bad idea.

I'm fine with doing the check or not doing it.

bryceosterhaus: @wincent ah I see, so you are thinking checking against `Clay*` but allowing ListItem and ListHeader? I think that makes sense to me, especially on a library level.

Ill add that plus the docs and update here

bryceosterhaus:
@wincent @matuzalemsteles updated

PR #3302, opened 2020-06-01
feat(@clayui/toolbar): Add Toolbar component

kresimir-coko:
Hey @bryceosterhaus here's my initial draft of how I imagine the Toolbar component should be structured.

### What I've done:
- Removed mentions of Upper Toolbar as a package
- Essentially renamed upper-toolbar.tsx to toolbar.tsx
  - I actually made a new file, and then in the existing upper-toolbar.tsx I just return Toolbar
  - This doesn't do anything different than the regular Toolbar but it serves as a placeholder for me to get my point across
- Added a new `default` story for Toolbar, which is just an empty `nav` right now
- Separated `Item` and `Input` into their own files and added them to the `ClayToolbar` namespace.

### Questions
1. Did I have to remove mentions of Upper Toolbar anywhere else? Both clayui.com and storybook build successfully with the current state of the PR
2. What are the differences between `Toolbar` and `Upper Toolbar`? Some classes, if so, which ones?
  2.1. Is Upper Toolbar a very specific combination of `Toolbar.Item`'s and `Toolbar.Input`'s? If so, which combination, the docs are vague.
3. Am I on the right track? The Lexicon docs aren't prescriptive on the same level as for other components, but that's probably on purpose.
4. Is the `info-bar` mentioned by @jbalsas [here](https://github.com/liferay/clay/pull/3248#issuecomment-631588453) basically a variation of this `Toolbar` I'm working on but with the `info-bar` class? It's hard to know for sure with the fact that the jsp file you linked is just some HTML
5. Should Management Toolbar also go under this component as a high level component?

It's all kinda muddy in information, @pat270 can you, as an extension of this [comment](https://github.com/liferay/clay/pull/3248#issuecomment-631586212) you made, give me some clarification as to when we use which classes in relation to Toolbar vs Management Toolbar vs Upper Toolbar vs Info Bar. Maybe you can make it a table or something that's easy for my tiny brain to understand.

### DIFF
--- a/packages/clay-toolbar/src/Action.tsx
+++ b/packages/clay-toolbar/src/Action.tsx
@@ -1,0 +1,41 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import ClayIcon from '@clayui/icon';
+import classNames from 'classnames';
+import React from 'react';
+
+export interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+	/**
+	 * Flag that determines if the Action will have a `disabled` class, disabling interactions.
+	 */
+	disabled?: boolean;
+
+	/**
+	 * Path to clay icon spritemap
+	 */
+	spritemap: string;
+	symbol: string;
+}
+
+const Action: React.FunctionComponent<IProps> = ({
+	className,
+	disabled,
+	spritemap,
+	symbol,
+	...otherProps
+}: IProps) => (
+	<a
+		className={classNames(className, 'component-action', {disabled})}
+		role="button"
+		{...otherProps}
+	>
+		<ClayIcon spritemap={spritemap} symbol={symbol} />
+	</a>
+);
+
+Action.displayName = 'ClayToolbarAction';
+
+export default Action;
--- a/packages/clay-toolbar/src/Action.tsx
+++ b/packages/clay-toolbar/src/Action.tsx
@@ -17,6 +17,10 @@ export interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
 	 * Path to clay icon spritemap
 	 */
 	spritemap: string;
+
+	/**
+	 * Symbol of the icon used inside the Action. You can find available symbols here: https://clayui.com/docs/components/icon.html
+	 */
 	symbol: string;
 }
 
--- a/packages/clay-toolbar/src/Item.tsx
+++ b/packages/clay-toolbar/src/Item.tsx
@@ -1,0 +1,36 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+export interface IItemProps extends React.HTMLAttributes<HTMLLIElement> {
+	/**
+	 * Flag to indicate if Item should auto expand to fill the remaining width.
+	 */
+	expand?: boolean;
+}
+
+const Item: React.FunctionComponent<IItemProps> = ({
+	children,
+	className,
+	expand,
+	...otherProps
+}: IItemProps) => {
+	return (
+		<li
+			className={classNames(className, 'tbar-item', {
+				'tbar-item-expand': expand,
+			})}
+			{...otherProps}
+		>
+			<div className="tbar-section">{children}</div>
+		</li>
+	);
+};
+
+Item.displayName = 'ClayToolbarItem';
+
+export default Item;
--- a/packages/clay-toolbar/src/Item.tsx
+++ b/packages/clay-toolbar/src/Item.tsx
@@ -6,19 +6,19 @@
 import classNames from 'classnames';
 import React from 'react';
 
-export interface IItemProps extends React.HTMLAttributes<HTMLLIElement> {
+export interface IProps extends React.HTMLAttributes<HTMLLIElement> {
 	/**
 	 * Flag to indicate if Item should auto expand to fill the remaining width.
 	 */
 	expand?: boolean;
 }
 
-const Item: React.FunctionComponent<IItemProps> = ({
+const Item: React.FunctionComponent<IProps> = ({
 	children,
 	className,
 	expand,
 	...otherProps
-}: IItemProps) => {
+}: IProps) => {
 	return (
 		<li
 			className={classNames(className, 'tbar-item', {
@@ -26,7 +26,7 @@ const Item: React.FunctionComponent<IItemProps> = ({
 			})}
 			{...otherProps}
 		>
-			<div className="tbar-section">{children}</div>
+			{children}
 		</li>
 	);
 };
--- a/packages/clay-toolbar/src/Label.tsx
+++ b/packages/clay-toolbar/src/Label.tsx
@@ -7,10 +7,25 @@ import ClayLabel from '@clayui/label';
 import classNames from 'classnames';
 import React from 'react';
 
-const Label: React.FunctionComponent<React.ComponentProps<
-	typeof ClayLabel
->> = ({children, className, ...otherProps}) => (
-	<ClayLabel className={classNames(className, 'tbar-label')} {...otherProps}>
+export interface IProps extends React.ComponentProps<typeof ClayLabel> {
+	/**
+	 * FLag that determines if the Link will have a `component-label` class, allowing CSS overrides.
+	 */
+	componentLabel?: boolean;
+}
+
+const Label: React.FunctionComponent<IProps> = ({
+	children,
+	className,
+	componentLabel,
+	...otherProps
+}) => (
+	<ClayLabel
+		className={classNames(className, 'tbar-label', {
+			'component-label': componentLabel,
+		})}
+		{...otherProps}
+	>
 		{children}
 	</ClayLabel>
 );
--- a/packages/clay-toolbar/src/Label.tsx
+++ b/packages/clay-toolbar/src/Label.tsx
@@ -7,23 +7,11 @@ import ClayLabel from '@clayui/label';
 import classNames from 'classnames';
 import React from 'react';
 
-export interface IProps extends React.ComponentProps<typeof ClayLabel> {
-	/**
-	 * FLag that determines if the Link will have a `component-label` class, allowing CSS overrides.
-	 */
-	componentLabel?: boolean;
-}
-
-const Label: React.FunctionComponent<IProps> = ({
-	children,
-	className,
-	componentLabel,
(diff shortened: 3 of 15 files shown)


kresimir-coko:
Hey @bryceosterhaus I've pushed a new commit that has no connection to the initial one from a couple days ago.

## Notes
### Low level components
Link and Label low level components are basically ClayLink and ClayLabel with a class specific to Toolbar so I've just made the low-level components be small wrappers around them, not sure if that's good because it means we have a dependancy, but then again, probably 90%+ of the Toolbars will end up using them so it might save on some boilerplate
#### Link and Button
Both of these have a `tbar-link` class assigned to them, I haven't checked the CSS that get's applied, but maybe that's something that we can take a 2nd glance at and make it more semantic by adding a `tbar-btn` class. @pat270 

### Error
Currently, I hadn't had the time to properly investigate, but I get an eror on build 
`src/index.tsx(70,1): error TS4082: Default export of the module has or is using private name 'IProps'.`. I cross-referenced ClayAutocomplete's way of importing low-level components and I can't see what I'm doing different.

### Component Action
I noticed there are some elements with a `component-action` class but I didn't figure out why a ClayLink, ClayButton or some other Clay component isn't used in that case. Maybe that can also be a low level component


### Stories and tests
I haven't included much in this department yet, the basic test that comes with the Yeoman generator and a custom basic story just to include all the low level components. I didn't want to spend time on that since the implementation isn't final yet.

Let me know if I'm on good track and what can be improved.

wincent:
Just want to put a reminder here that when this goes in and a release gets cut and we want to start using it in liferay-portal, will need to update the `liferay-npm-bundler-preset-dev` like ([example of this](https://github.com/liferay/liferay-npm-tools/pull/459) that I just made for `@clayui/layout`).

kresimir-coko:
Hey @bryceosterhaus  I've cleaned up the props and made the stories slightly better. Took me an entire day messing around with TypeScript and ESLint.

### Upper Toolbar
```
<nav
	className={classNames(
		className,
		'component-tbar',
		'subnav-tbar-light',
		'tbar'
	)}
	{...otherProps}
>
	<ClayLayout.ContainerFluid>
		<ul className="tbar-nav">{children}</ul>
	</ClayLayout.ContainerFluid>
</nav>
```
Since the implementation of Upper Toolbar is basically `Toolbar` > `ContainerFluid` > `Toolbar.Nav` with the `component-tbar` & `subnav-tbar-light` classes (@pat270 can you check if these are the exact classes that need to be in Upper Toolbar) we can probably make like a high level component with those specific classes and the container. 

I didn't wanna work on that yet until we confirm the Toolbar implementation is solid enough, plus I ran out of time.

packages/clay-toolbar/src/Item.tsx
```diff
@@ -0,0 +1,36 @@
...
+	...otherProps
+}: IProps) => {
+	return (
+		<li
+			className={classNames(className, 'tbar-item', {
+				'tbar-item-expand': expand,
+			})}
+			{...otherProps}
+		>
+			<div className="tbar-section">{children}</div>
```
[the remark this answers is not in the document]
pat270: I would say no. The purpose of `tbar-section` is to force the content inside to return to block / inline layout from a flex layout.

pat270:
> What are the differences between Toolbar and Upper Toolbar? Some classes, if so, which ones?
2.1. Is Upper Toolbar a very specific combination of Toolbar.Item's and Toolbar.Input's? If so, which combination, the docs are vague.

Upper Toolbar is the Lexicon name for [this](https://liferay.design/lexicon/core-components/toolbars/toolbar-upper/) component. It looks exactly like the Management Bar with different content inside it. Using `tbar` to create this component will require some CSS updates that I talk about below.

> Is the info-bar mentioned by @jbalsas here basically a variation of this Toolbar I'm working on but with the info-bar class? It's hard to know for sure with the fact that the jsp file you linked is just some HTML

I think we can ignore `info-bar` markup since it was implemented inside Portal. I think @jbalsas just mentioned it as a reference so we don't miss use cases.

> Should Management Toolbar also go under this component as a high level component?

Management Toolbar actually uses the Navbar component in Clay CSS. Don't think we can unify Toolbar and Management Bar markup without rewriting the markup inside the component.

### Component Action

`component-*` is a generic class that is reserved for items inside a component that can be styled any way you like. Component Action is `btn btn-sm btn-monospaced` without borders. It gives an out if designers want a one off button/link/text/color/etc.

### Clay CSS Tbar

This just provides the Basic structure for a Tbar. No colors, no sizing, just basic padding and alignment.

### Upper Toolbar

> Both of these have a tbar-link class assigned to them, I haven't checked the CSS that get's applied, but maybe that's something that we can take a 2nd glance at and make it more semantic by adding a tbar-btn class.

If we want to use Clay CSS `tbar` markup for this, I will need to add a modifier class that styles the `tbar` properly. It will probably be named `upper-tbar`. The styles for `tbar-link`, `tbar-link-monospaced`, `tbar-btn`, and `tbar-btn-monospaced` will come with `upper-tbar`. Just as a note, there is no real semantic meaning to `tbar-link` and `tbar-btn` since they can be used with `a`, `button`, and both can be made to look like each other.

packages/clay-toolbar/stories/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+import '@clayui/css/lib/css/atlas.css';
+import {storiesOf} from '@storybook/react';
+import React from 'react';
+
+import ClayToolbar from '../src';
+
+storiesOf('Components|ClayToolbar', module)
+	.add('default', () => (
```
pat270, line 13: Maybe for the default example https://pat270.github.io/clay3-test-site/vclay-table-dd/content/tbar/#tbar might be better?

packages/clay-toolbar/stories/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+				</ClayToolbar.Item>
+				<ClayToolbar.Item expand>
+					<ClayToolbar.Section>{'Section #1'}</ClayToolbar.Section>
+					<ClayToolbar.Section>{'Section #2'}</ClayToolbar.Section>
+				</ClayToolbar.Item>
+			</ClayToolbar.Nav>
+		</ClayToolbar>
+	))
+	.add('subnav', () => (
+		<ClayToolbar subnav={{disabled: true, displayType: 'primary'}}>
```
pat270, line 32: `subnav-tbar-primary subnav-tbar-disabled` example https://pat270.github.io/clay3-test-site/vclay-table-dd/content/tbar/#subnav-tbar-primary-subnav-tbar-disabled

kresimir-coko:
Hey @bryceosterhaus I've squashed the old commits and added 3 new commits on top, which contain:
- Improvements to low level components, like props, and markup
- New Toolbar.Action component, for those `a.component-action` elements with icons
- Stories improvements to match examples in Markup

pat270:
This LGTM. I didn't see anything related to the ClayUpperToolbar Storybook in this pull, but is it possible to get that demo removed so people don't get confused by it?

packages/clay-toolbar/src/Action.tsx
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * Flag that determines if the Action will have a `disabled` class, disabling interactions.
+	 */
+	disabled?: boolean;
+
+	/**
+	 * Path to clay icon spritemap
+	 */
+	spritemap: string;
+	symbol: string;
```
bryceosterhaus, line 20: need jsdoc here

packages/clay-toolbar/src/Label.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Label: React.FunctionComponent<IProps> = ({
+	children,
+	className,
+	componentLabel,
+	...otherProps
+}) => (
+	<ClayLabel
+		className={classNames(className, 'tbar-label', {
+			'component-label': componentLabel,
```
bryceosterhaus, line 25: I think we can just always leave this class as active, seems like itll always be there for labels in a tbar. This means we can remove the prop as well.

packages/clay-toolbar/src/Link.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import ClayLink from '@clayui/link';
+import classNames from 'classnames';
+import React from 'react';
+
+export interface IProps extends React.ComponentProps<typeof ClayLink> {
+	/**
+	 * FLag that determines if the Link will have a `component-link` class, allowing CSS overrides.
+	 */
+	componentLink?: boolean;
```
bryceosterhaus, line 14: Same for this, I think we can just remove these as configurable and always have this class on the component. If it becomes and issue later we can always add the API in.

Its easier to add an API than remove one.

packages/clay-toolbar/src/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+import Label from './Label';
+import Link from './Link';
+import Nav from './Nav';
+import Section from './Section';
+
+interface IProps extends React.HTMLAttributes<HTMLElement> {
+	/**
+	 * Specifies whether the Toolbar will have a `component-tbar` class.
+	 */
+	component?: boolean;
```
bryceosterhaus, line 20: Same as above, we can remove this

kresimir-coko:
> This LGTM. I didn't see anything related to the ClayUpperToolbar Storybook in this pull, but is it possible to get that demo removed so people don't get confused by it?

I don't think we want to remove it, yet. I didn't remove or do anything to Upper Toolbar.
I think the plan was when this gets merged to absorb Upper Toolbar into Toolbar. Then I was thinking of removing Upper Toolbar demo, as it would be a section under Toolbar.

Did I get that right @bryceosterhaus ?

I've pushed the requested changes, the JSDoc and adding `component-*` classes by default. 

Also let me know how I can `push` changes so that I don't have to `force push` them, I don't want to cause a problem one day because I keep `force-pushing`. When I do a regular `push` it says something like `"not all refs have been pushed..."`, is it because I always do a `pull` before pushing?

bryceosterhaus:
Sent update at https://github.com/liferay/clay/pull/3325

@kresimir-coko, for push/force-push, it depends if you are rebasing off of latest after updating the PR. You should be able to just `push` to your branch if you haven't rebased or changed anything in the history. If you do rebase, I think you will have to force push, I could be wrong though...

wincent:
> If you do rebase, I think you will have to force push, I could be wrong though...

Rebasing will create new commit objects (ie. because the commit date will change, and because the parent commit hash will be different — and even a single bit of difference in a commit object will cause it to have a new hash), which means the only way to update is to force-push. The only time this isn't true is when the upstream hasn't moved forward — in that case `git rebase` will say something like `Current branch master is up to date` and won't actually do anything.

Git will let you add commits on top no problems without force pushing (that is what is called a "fast-forward"); but any other kind of push (eg. pushing rebased commits, or pushing after resetting to a prior point etc) will need to be a force push.

force-push

PR #3305, opened 2020-06-01
fix: bad diff args

bryceosterhaus:
:facepalm:


wincent:
> 🤦

It's not that bad, is it? From `man git-diff`:

```
       --quiet
           Disable all output of the program. Implies --exit-code.
```

I would have thought that `--quiet` was fine (and actually better) for this.

bryceosterhaus:
@wincent thats what I thought too, but it didn't set the anything to the output of the task. So I ended up just using --exit-code and it worked fine.   🤷‍♂️

PR #3306, opened 2020-06-01
fix(@clayui/date-picker): remove moment.js

bryceosterhaus:
I went with a POC in removing moment by replacing it with ~luxon~ date-fns. We still have a dependency, but its about 50kb smaller than when we used moment.

One caveat is that the "format" strings are different, so that would be a breaking change, although we could create some sort of map.

Additionally, I want to see if it'd be possible to just use native APIs, so I will continue on that and see what it requires.

Curious to hear any thoughts on this

~**1st commit is using luxon**~

~**2nd commit is using date-fns**~

### DIFF
--- a/packages/clay-date-picker/src/Helpers.ts
+++ b/packages/clay-date-picker/src/Helpers.ts
@@ -3,7 +3,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import moment from 'moment';
+import {default as formatDate} from 'date-fns/format';
+import {default as parseDate} from 'date-fns/parse';
+
+export {formatDate, parseDate};
 
 export interface IDay {
 	date: Date;
@@ -17,8 +20,8 @@ export type Month = Array<WeekDays>;
 /**
  * Clone a date object.
  */
-export function clone(d: Date) {
-	return new Date(d.getTime());
+export function clone(date: number | Date) {
+	return new Date(date instanceof Date ? date.getTime() : date);
 }
 
 export function getDaysInMonth(d: Date) {
@@ -103,17 +106,36 @@ export function range({end, start}: {end: number; start: number}) {
 	);
 }
 
-/**
- * Helper function for getting certain props from `moment`.
- * This allows users to not have to import and use `moment` themselves.
- */
-export function getLocaleProps(locale: string) {
-	const localeData = moment.localeData(locale);
+export function addMonths(date: number | Date, months: number) {
+	date = clone(date);
 
-	return {
-		dateFormat: localeData.longDateFormat('L'),
-		firstDayOfWeek: localeData.firstDayOfWeek(),
-		months: localeData.months(),
-		weekdaysShort: localeData.weekdaysShort(),
-	};
+	date.setMonth(date.getMonth() + months);
+
+	return date;
+}
+
+export function setDate(
+	date: Date,
+	obj: {
+		date?: number | string;
+		seconds?: number | string;
+		milliseconds?: number | string;
+		hours?: number | string;
+		minutes?: number | string;
+		year?: number | string;
+	}
+) {
+	date = clone(date);
+
+	return Object.keys(obj).reduce((acc, key) => {
+		const method = `set${key.charAt(0).toUpperCase() + key.slice(1)}`;
+		// @ts-ignore
+		acc[method](obj[key]);
+
+		return acc;
+	}, date);
+}
+
+export function isValid(date: Date) {
+	return date instanceof Date && !isNaN(date.getTime());
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -7701,6 +7701,11 @@ dataloader@^1.4.0:
   resolved "https://registry.yarnpkg.com/dataloader/-/dataloader-1.4.0.tgz#bca11d867f5d3f1b9ed9f737bd15970c65dff5c8"
   integrity sha512-68s5jYdlvasItOJnCuI2Q9s4q98g0pCyL3HrcKJu8KNugUl8ahgmZYg38ysLTgQjjXX3H8CJLkAvWrclWfcalw==
 
+date-fns@^2.14.0:
+  version "2.14.0"
+  resolved "https://registry.yarnpkg.com/date-fns/-/date-fns-2.14.0.tgz#359a87a265bb34ef2e38f93ecf63ac453f9bc7ba"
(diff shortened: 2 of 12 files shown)


bryceosterhaus:
Only issue with date-fns is the lack of localization support, but that only really effects our helper function `getLocaleProps` which I don't think is actually used anywhere...

bryceosterhaus:
Going off of size-limit builds, with all dependencies, minified and gzipped


| Clay using...   |      Total SIze      |
|----------|:-------------:|
| moment.js |  120.51 KB |
| luxon |    73.69 KB   |
| date-fns | 71.18 KB |

wincent:
> Additionally, I want to see if it'd be possible to just use native APIs instead of luxon, so I will continue on that and see what it requires.

Glad you mention this as a possibility. If it could be made to work, that would clearly be a win in terms of size, isolation from the vagaries of third party stuff, and extracting maximum value from the platform.

jbalsas:
> Additionally, I want to see if it'd be possible to just use native APIs instead of luxon, so I will continue on that and see what it requires.

I would much rather prefer this.

If not possible, I would also prefer to see `date-fns` in here since it's the library we've been heavily using in DXP.

If, after all, we think `luxon` is the right choice, we should at least compare one with the other, pick based on merits and properly document our recommendation and support.

bryceosterhaus:
Alright so I don't think we can get away with strictly native APIs since we allow for the custom date formatting... However, I was able to use just the format and parse functions from `date-fns` instead of the whole package.

Total size with `date-fns`: **62.67 KB** (9.2KB from date-fns)

Knowing that we only need parse and format, I stumbled across `dayjs`, which is smaller again and has the same format patterns as moment. So in the [second commit](https://github.com/liferay/clay/pull/3306/commits/dce6950d4741df33bb0a748cb6e4221da40934b2), I tried that out and it seemed to work great.

Total size with `dayjs`: **55.94 KB** 🔥🔥 (2.6KB from dayjs)

Since we really don't rely on heavy date manipulation, I think going with `dayjs` would be appropriate for our use case. The only reason I see it appropriate to use `date-fns` instead is because its already in dxp.

Both libraries score well on ClearlyDefined(licensing) and both seem to be highly supported and popular in the community.

Overall, this would take clays total size from 120KB to 56KB. 😱

jbalsas:
> The only reason I see it appropriate to use date-fns instead is because its already in dxp.

I would then stick with `date-fns` for now. This is already a huge improvement and also reduces the changes we need to apply across the board... once settled we can go for `dayjs` or native once that's possible...

bryceosterhaus:
Removed dayjs and going with `date-fns`. 

Any final issues before merging other than docs?

jbalsas:
I haven't checked code-wise, but I agree on principle! ;)

packages/clay-date-picker/src/Helpers.ts
```diff
@@ -3,7 +3,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import moment from 'moment';
+import formatDate from 'date-fns/format';
+import parseDate from 'date-fns/parse';
+
+export {formatDate, parseDate};
```
wincent, line 9: If you write it like this, it becomes obvious that these are just re-exports and not used in this file:

```js
export {default as formatDate} from 'date-fns/format';
export {default as parseDate} from 'date-fns/parse';
```

packages/clay-date-picker/src/Helpers.ts
```diff
@@ -103,17 +106,36 @@ export function range({end, start}: {end: number; start: number}) {
...
-		weekdaysShort: localeData.weekdaysShort(),
-	};
-}
+export const addMonths = (date: number | Date, months: number) => {
+	date = clone(date);
+
+	date.setMonth(date.getMonth() + months);
+
+	return date;
+};
```
wincent, line 115: I don't think we should always export functions in the same way. You have both:

```js
export function a() {
  // ...
};
```

and:

```js
export const b() => {
  // ...
};
```

(FWIW, I prefer the former because of hosting.)

wincent: Edit (sigh): of course I meant "I do think"...

yarn.lock
```diff
@@ -15984,7 +15989,7 @@ modify-values@^1.0.0:
   resolved "https://registry.yarnpkg.com/modify-values/-/modify-values-1.0.1.tgz#b3939fa605546474e3e3e3c63d64bd43b4ee6022"
   integrity sha512-xV2bxeN6F7oYjZWTe/YPAy6MN2M+sL4u/Rlm2AHCIVGfo2p1yGmBHQ6vHehl4bRTZBdHu3TSkWdYgkwpYzAGSw==
 
-moment@^2.21.0, moment@^2.22.2, moment@^2.5.1:
+moment@^2.21.0, moment@^2.5.1:
```
wincent, line 15992: So `moment` is still in here; is it coming in transitively? What does `yarn why moment` say?

packages/clay-date-picker/src/Helpers.ts
wincent: I created https://github.com/liferay/eslint-config-liferay/issues/184 to track enforcing this, with the recommendation that we go with `export function` — but feel free to comment there if you want to advocate the opposite (either way, we should enforce it).

yarn.lock
bryceosterhaus: It comes from [gatsby](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/package.json#L108).

```
   - "_project_#clayui.com#gatsby" depends on it
   - Hoisted from "_project_#clayui.com#gatsby#moment"
   - Hoisted from "_project_#@clayui#css#metalsmith-permalinks#moment"
```

bryceosterhaus:
@wincent for some reason when I update the syntax to be `export {default as formatDate} from 'date-fns';`, [jest fails](https://github.com/liferay/clay/pull/3306/checks?check_run_id=739640258#step:8:14). I tried to update both jest and ts but no luck. Have you seen this before?

wincent:
> Have you seen this before?

Might be a Babel thing? If you look at the distributed source of `data-fns/format/index.js`, it looks like it is supposed to work:

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = format;
// ... etc
```

Maybe something that will magically start working in the future at some point when some dep gets updated.

bryceosterhaus:
@wincent I was hoping it wouldve resolved after the babel dependency updates but seems to still break. I'm just gonna import and then export separately for now and then look more into how our dependencies might be causing this.

PR #3320, opened 2020-06-04
chore: update dependencies

### LINKED ISSUE #3319: Update typescript, babel, and eslint
Some of these have been outdated for awhile, so wanted to get them up to date and not lag behind too far.

bryceosterhaus:
fixes #3319


packages/clay-time-picker/src/index.tsx
```diff
@@ -163,8 +163,8 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
 	onInputChange = () => {},
 }: IProps) => {
 	const useConfig: Config = config[use12Hours ? 'use12Hours' : 'use24Hours'];
-	const [actionVisible, setActionVisible] = React.useState(false);
-	const [isFocused, setIsFocused] = React.useState();
+	const [actionVisible, setActionVisible] = React.useState<boolean>(false);
+	const [isFocused, setIsFocused] = React.useState<boolean>(false);
```
wincent, line 167: I get why you added the `false` here, but why do you need the `<boolean>`? I would have expected TS to infer those, and it's easy enough for a human to infer it too.

bryceosterhaus: there was a complaint in TS, but it probably complained because there was no initial value to infer from. I can probably remove the boolean portion

PR #3324, opened 2020-06-05
fix(@clayui/layout): add ability to disable fluid-max-* class

### LINKED ISSUE #3313: New ClayLayout: `<ClayLayout.ContainerFluid>` cannot display only the class `container-fluid`
### New ClayLayout: `<ClayLayout.ContainerFluid>` cannot display only the class `container-fluid`
https://storybook.clayui.com/iframe.html?id=components-claylayout--row-positioning

We tried by adding:
`<ClayLayout.ContainerFluid fluidSize={false}>``
it does not work

Be able to remove the default class:
`container-fluid-max-xl`

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | master  |
| React |         |

bryceosterhaus:
fixes #3313

### DIFF
--- a/packages/clay-layout/src/__tests__/index.tsx
+++ b/packages/clay-layout/src/__tests__/index.tsx
@@ -108,6 +108,6 @@ describe('ClayLayout', () => {
 
 		expect(
 			container.querySelector('.container-fluid')!.classList
-		).toMatchSnapshot();
+		).not.toContain('container-fluid-max-xl');
 	});
 });
--- a/packages/clay-layout/src/__tests__/index.tsx
+++ b/packages/clay-layout/src/__tests__/index.tsx
@@ -96,4 +96,18 @@ describe('ClayLayout', () => {
 
 		expect(container).toMatchSnapshot();
 	});
+
+	it('removes size class when prop is set to false', () => {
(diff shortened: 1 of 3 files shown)


packages/clay-layout/src/__tests__/index.tsx
```diff
@@ -96,4 +96,18 @@ describe('ClayLayout', () => {
...
+			<>
+				<ClayLayout.ContainerFluid size={false}>
+					{'test'}
+				</ClayLayout.ContainerFluid>
+			</>
+		);
+
+		expect(
+			container.querySelector('.container-fluid')!.classList
+		).toMatchSnapshot();
```
wincent, line 111: It seems a shame to have to look at a snapshot file just to see what we expect the class list to look like (ie. a single class, `.container-fluid`). The trouble with snapshots is generally that they say too much (which means it can be hard to know whether an update to them is legitimate or evidence of breakage); but this one suffers from the opposite problem — that it doesn't say enough to justify being a snapshot. Can we inline this and avoid the snapshot?

bryceosterhaus: Good call, just updated

PR #3325, opened 2020-06-08
refactor(@clayui/toolbar): clean up logic

### LINKED ISSUE #3300: @clayui/toolbar package?
It might make more sense to create just a basic toolbar package and merge upper-toolbar into that package. This way we could put all the `tbar` classes in toolbar. upper-toolbar could be a high-level component from toolbar.

Stems from this convo, https://github.com/liferay/clay/pull/3248

bryceosterhaus:
fixes #3300

follow up from https://github.com/liferay/clay/pull/3302

@kresimir-coko, check out the comments on my updates

### DIFF
--- a/packages/clay-toolbar/src/index.tsx
+++ b/packages/clay-toolbar/src/index.tsx
@@ -1,0 +1,81 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+import Item from './Item';
+import Label from './Label';
+import Link from './Link';
+import Nav from './Nav';
+import Section from './Section';
+
+interface IProps extends React.HTMLAttributes<HTMLElement> {
+	/**
+	 * Specifies whether the Toolbar will have a `component-tbar` class.
+	 */
+	component?: boolean;
+	/**
+	 * Adds a helper class that turns the Toolbar inline at a specified breakpoint.
+	 */
+	inlineBreakpoint?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
+	/**
+	 * Defines if the toolbar should have the `subnav-tbar` class.
+	 */
+	subnav?:
+		| boolean
+		| {
+				disabled?: boolean;
+				displayType?: 'light' | 'primary';
+		  };
+}
+
+const ClayToolbar: React.FunctionComponent<IProps> & {
+	Item: typeof Item;
+	Label: typeof Label;
+	Link: typeof Link;
+	Nav: typeof Nav;
+	Section: typeof Section;
+} = ({
+	children,
+	className,
+	component,
+	inlineBreakpoint,
+	subnav,
+	...otherProps
+}: IProps) => {
+	const classes = classNames(
+		className,
+		'tbar',
+		{
+			'component-tbar': component,
+			'subnav-tbar': !!subnav,
+			'subnav-tbar-disabled': !!(
+				subnav &&
+				typeof subnav !== 'boolean' &&
+				subnav.disabled
+			),
+			[`tbar-inline-${inlineBreakpoint}-down`]: inlineBreakpoint,
+		},
+		subnav &&
+			typeof subnav !== 'boolean' &&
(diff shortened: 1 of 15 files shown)


packages/clay-toolbar/src/index.tsx
```diff
@@ -43,29 +40,24 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
...
-	);
+	subnav =
+		typeof subnav === 'boolean'
+			? subnav === true
+				? {}
+				: undefined
+			: subnav;
+
+	const classes = classNames(className, 'component-tbar tbar', {
+		'component-tbar': !subnav,
```
bryceosterhaus, line 55: Since this class always exists if its not a subnav, I removed the prop and did this

packages/clay-toolbar/src/index.tsx
```diff
@@ -43,29 +40,24 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
...
-			typeof subnav !== 'boolean' &&
-			subnav.displayType &&
-			`subnav-tbar-${subnav.displayType}`
-	);
+	subnav =
+		typeof subnav === 'boolean'
+			? subnav === true
+				? {}
+				: undefined
+			: subnav;
```
bryceosterhaus, line 52: This is for typescript and so that our class logic doesnt have to have so many steps. By setting it to an empty object if its true, this allows us to use optional chaining below. I think this cleans it up quite a bit.

wincent: I know that optional chaining makes things *look* neat, but I actually think it might be hiding a smell in some places (don't know about this one).

Like the similar pattern I commented on [over here](https://github.com/liferay/clay/pull/3303/files/916784ec784ae88a129392f147480a6b54062be1#r436503877), ideally TS would infer as much information as possible about what type a value has at any given point — and on that PR, we use `?.` to make up for TS's failure to infer what we would know to be true just by looking at the code. In that case, the "smell" was that we were using `?.` to say "this value might be undefined but you shouldn't blow up if it is", when the truth was "at this point we know it can't be undefined and we're just using `?.` to shut up the type checker". The alternative to using `?.` all over the place, is tweaking the code so that the inference is better and you don't need the optional chaining anymore. In that PR, you can see that assigning to a variable is enough to help TS narrow the type information. (I general like it when the type system and I agree with each other; it's not so fun when we're in conflict.)

packages/clay-toolbar/src/index.tsx
```diff
@@ -43,28 +40,24 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
 } = ({
 	children,
 	className,
-	component,
 	inlineBreakpoint,
 	subnav,
 	...otherProps
 }: IProps) => {
+	subnav = subnav === true ? {} : subnav;
```
bryceosterhaus, line 47: By setting this, there is no need to check for boolean types or `subnav && ...` multiple times below

packages/clay-toolbar/src/index.tsx
```diff
@@ -43,28 +40,24 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
...
-				typeof subnav !== 'boolean' &&
-				subnav.disabled
-			),
-			[`tbar-inline-${inlineBreakpoint}-down`]: inlineBreakpoint,
 		},
-		subnav &&
-			typeof subnav !== 'boolean' &&
-			subnav.displayType &&
-			`subnav-tbar-${subnav.displayType}`
+		subnav && {
```
bryceosterhaus, line 56: Separated these objects so that these are only added if subnav is an object

packages/clay-toolbar/src/index.tsx
bryceosterhaus: Thats a good point. I had another pass through that logic and think I found a better way to clean it up. Check out https://github.com/liferay/clay/pull/3325/commits/1e0fb14bd617cf13bedbc7ea936d1b39c0b6eca9#diff-60699a060cdb1fe818af4b8473b3240fR53-R60

kresimir-coko:
LGTM, thanks for writing the explanations, they help!

packages/clay-toolbar/src/index.tsx
wincent: Looks like a nice improvement!

PR #3328, opened 2020-06-09
feat(@clayui/layout): Make ContentRow and ContentCol forwardRef

### LINKED ISSUE #3327: ClayLayout low level components need forwardRef
While working on [Use new Layout Components in Components in layout-admin-web](https://issues.liferay.com/browse/LPS-114072) which has drag & drop functionality on `ClayLayout.ContentRow` I encountered that `ref` is undefined which causes d&d to not function.

kresimir-coko:
Fixes #3327 

Hey @bryceosterhaus here's my attempt at making ContentRow and ContentCol be `forwardRefs` instead of `FunctionComponents`. This is my first time using `forwardRef` so if I missed something let me know. Also feel free to take this and run away with it so we can merge it faster because an LPS linked in the issue depends on it.

I've also added a quick story to make sure it works and to provide people with the example of it working.

### DIFF
--- a/packages/clay-layout/src/Content.tsx
+++ b/packages/clay-layout/src/Content.tsx
@@ -45,17 +45,20 @@ interface IContentRowProps extends React.HTMLAttributes<HTMLDivElement> {
 	verticalAlign?: 'center' | 'end';
 }
 
-const ContentRow: React.FunctionComponent<IContentRowProps> = ({
-	children,
-	className,
-	containerElement: ContainerElement = 'div',
-	float,
-	noGutters,
-	padded,
-	verticalAlign,
-	...otherProps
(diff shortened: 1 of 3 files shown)


wincent:
Your title says "Make ContentRow and ContentCol forwardRef" but it also adds the `DraggableLayoutRow` stuff. You could split, or retitle.

bryceosterhaus:
@kresimir-coko, sent an update https://github.com/liferay/clay/pull/3329

I went ahead and just added forwardRef to all exposed components from layout since people may need them. I also removed the story you added since it was really similar to the other dnd example and I don't want users to think you have to do soemthing different when you use Layout components.

PR #3330, opened 2020-06-11
fix(@clayui/css): Update Clay CSS and Bootstrap 4 Licenses to Use SPDX Naming Scheme

pat270:
The license text:
```
/**
* Clay 3.12.0
*
* SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
* SPDX-FileCopyrightText: © 2020 Contributors to the project Clay <https://github.com/liferay/clay/graphs/contributors>
*
* SPDX-License-Identifier: BSD-3-Clause
*/

/**
* Bootstrap v4.4.1
*
* SPDX-FileCopyrightText: © 2019 Twitter, Inc. <https://twitter.com>
* SPDX-FileCopyrightText: © 2019 The Bootstrap Authors <https://getbootstrap.com/>
*
* SPDX-License-Identifier: LicenseRef-MIT-Bootstrap
*/
```

This updates the licenses in:
- All single SVG Icons
- icons.svg
- atlas.scss
- atlas-variables.scss
- base.scss
- base-variables.scss
- bootstrap.scss (is left alone and uses Bootstrap's license text verbatim)

@bryceosterhaus I think we need to change up the release process a little to:
1) copy over the `LICENSES` directory from Clay root when releasing @clayui/css.
2) run the task `gulp version` in `packages/clay-css` when running `lerna version --conventional-commits --no-push`. This task just updates the Clay CSS version in files that have it.

/cc @jbalsas @wincent @silverhook

fixes #3282

PR #3333, opened 2020-06-15
fix(clay-css): copy LICENSES to lib before publish

bryceosterhaus:
@pat270 is this what you were thinking for copying the licenses? It'll copy it straight to the lib folder before release

I also added the `gulp version` command. This should run after `lerna version` but before we commit.

### DIFF
--- a/packages/clay-css/package.json
+++ b/packages/clay-css/package.json
@@ -23,8 +23,9 @@
 	"scripts": {
 		"build": "yarn compile",
 		"compile": "gulp compile",
+		"link": "npm link",
 		"prepublish": "yarn compile && ncp ../../LICENSES ./lib",
-		"link": "npm link"
+		"version": "gulp version"
 	},
 	"repository": {
 		"type": "git",
--- a/packages/clay-css/package.json
+++ b/packages/clay-css/package.json
@@ -23,7 +23,7 @@
 	"scripts": {
 		"build": "yarn compile",
 		"compile": "gulp compile",
-		"prepublish": "yarn compile",
+		"prepublish": "yarn compile && ncp ../../LICENSES ./lib",
 		"link": "npm link"
 	},
 	"repository": {


pat270:
@bryceosterhaus it should copy to clay-css root. https://unpkg.com/browse/@clayui/css@3.12.0/ where `LICENSES.txt` is.

`gulp version` command is perfect.

pat270:
I forgot to mention we need https://github.com/liferay/clay/pull/3330 merged before `gulp version` will work.

bryceosterhaus:
@pat270 instead of copying them over to that directory, can the licenses just live in the directory of clay-css?

pat270:
I think we're supposed to have one License directory containing all our licenses at the root of the project according to https://reuse.software/spec/#license-files.

pat270:
I'm not sure about the technical details if we consider `clay-css` its own project? @silverhook might be able provide more details.

pat270:
new pull: #3480

silverhook:
Yes, the idea is to have all license texts that apply in a `LICENSES/` folder, located in the root of the project/repo/package/tarball.

If there are any technical limitations, let me know and I’ll try to think about it.

> I'm not sure about the technical details if we consider `clay-css` its own project? @silverhook might be able provide more details.

I’m not entirely sure I understand this together with the initial message and commits. Is this about packaging Clay’s CSS into Clay repo, or what?

wincent:
> I'm not sure about the technical details if we consider `clay-css` its own project?

You *could* make the argument that it's a separate project, but I don't think I would. The fact that we break Clay up into little folders and publish them to NPM as packages is an implementation decision that doesn't mean that each one is a separate project; I think it's a layer within the overall project that is "Clay" (an implementation of Lexicon).

silverhook:
Perhaps it might help if I explain the basic idea for the demand of license texts being in a root `LICENSES/` folder.

The basic gist (and reason for why this requirement is in REUSE) is two fold:
- whenever you distribute someone else’s code, you need to have a license – so you need to keep the license texts _somewhere_
- if you need to store it somewhere, might as well make it easy to find (i.e. in `LICENSES/` folder at the root) and easy to identify the license (hence SPDX ID filename)

Together this enables anyone who gets the code to just take a quick glance at the  `LICENSES/` folder and right away see which all licenses apply to that code.

pat270:
This makes sense, thanks for the clarification. A hypothetical question, let's say we have a 100 licenses in Clay, would it be wise to separate those licenses by component? For example, Clay CSS licenses would be under `LICENSES/clay-css/the-license.txt`.

silverhook:
Great question.

No, that might not help much. Ideally the files in those components should carry the license SPDX IDs in their headers, so it’s easy to match with the license text in the folder.

But I would be honestly surprised if we managed to pull this many different licenses, even hypothetically. More realistically a big project might include one or two dozen different licenses that are used by hundreds of different components.

PR #3334, opened 2020-06-15
fix(@clayui/shared): only preventDefault if next/previous node exists in the focusManager

### LINKED ISSUE #3332: Keyboard Trap in FocusScope

See https://codesandbox.io/s/heuristic-leaf-1oz21?file=/public/index.html

Case **1a**: Using **only your keyboard**, press `TAB` until you reach the button `NEXT`.
Case **1b**: Using **only your keyboard**, press `TAB` until you reach the `Dropdown` button. Click 
`ENTER` to expand the dropdown, then use `TAB` to reach `NEXT`.

Case **2a** and **2b**. Try the same steps starting from the `NEXT` button pressing `SHIT+TAB` instead to tab in reverse order until reaching the `PREV` button.

All buttons are always reachable using only the keyboard

## What is the current result?

The dropdown component (trigger or menu) trap the keyboard focus once it enters the component.

---

On a quick exploration, `event` is [always prevented regardless of the result of focus management](https://github.com/liferay/clay/blob/master/packages/clay-shared/src/FocusScope.tsx#L50-L64). This should likely be something like:

```javascript
if (...) {
	if (focusManager.focusNext()) {
	    event.preventDefault();
        }
} else if (...) {
	if (focusManager.focusPrevious()) {
	    event.preventDefault();
        }
}
```

This would solve the focus trap, but would not achieve proper tab order when leaving the dropdown menu. When tabbing out of the last element in the menu, focus should go to the next "tabbable" element after the trigger.

bryceosterhaus:
fixes #3332

I ran this through a handful of manual tests and seems to fix the problem while also maintaining previous functionality.

@matuzalemsteles, can you think of any edge cases this would break?

### DIFF
--- a/packages/clay-shared/src/FocusScope.tsx
+++ b/packages/clay-shared/src/FocusScope.tsx
@@ -52,15 +52,17 @@ export const FocusScope: React.FunctionComponent<IProps> = ({
 			(arrowKeysLeftRight && keyCode === ARROW_RIGHT_KEY_CODE) ||
 			(keyCode === TAB_KEY_CODE && !shiftKey)
 		) {
-			event.preventDefault();
-			focusManager.focusNext();
+			if (focusManager.focusNext()) {
+				event.preventDefault();
+			}
 		} else if (
 			(arrowKeysUpDown && keyCode === ARROW_UP_KEY_CODE) ||
 			(arrowKeysLeftRight && keyCode === ARROW_LEFT_KEY_CODE) ||
 			(keyCode === TAB_KEY_CODE && shiftKey)
 		) {
-			event.preventDefault();
-			focusManager.focusPrevious();
+			if (focusManager.focusPrevious()) {
+				event.preventDefault();
+			}
 		}
 	};
 


bryceosterhaus:
@jbalsas your demo with the current changes https://codesandbox.io/s/sleepy-buck-3z82f

One area that is broken is you can't open the drop down and then focus on the next button. You would have to go to "shift + tab" to the button and close the menu first

matuzalemsteles:
Hey @bryceosterhaus I think this is good, it shouldn't break anything. `useFocusManagement` will always return the element if it exists.

jbalsas:
> One area that is broken is you can't open the drop down and then focus on the next button. You would have to go to "shift + tab" to the button and close the menu first

Hey @bryceosterhaus, yeah, that's what I said [here](https://github.com/liferay/clay/issues/3332)

> This would solve the focus trap, **but would not achieve proper tab order when leaving the dropdown menu. When tabbing out of the last element in the menu, focus should go to the next "tabbable" element after the trigger**.

If we don't have a quick fix (I couldn't think of one off the top of my head), I'd merge this and create an additional issue for it. 

This is an important accessibility issue we have here, so it's important to address it as soon as we can.

matuzalemsteles:
Yeah, we can go with that to prevent FocusScope from getting stuck in the focus and we can work on another issue to create a fallback in `useFocusManagement` so when you don't find a focusable element in the React structure, try searching the DOM. Perhaps the "shift + tab" feature will be a problem when it exists inside a React.Portal, it will become more complicated.

bryceosterhaus:
going to merge this and then I created a new issue, https://github.com/liferay/clay/issues/3340

Issue #3337, opened 2020-06-16
Add Drilldown Component


kresimir-coko:
Using [this](https://docs.google.com/document/d/1PVQwknqVIOQhf2JL0kEv0k6WHOExKOskiX-CxxWYRqs/edit) documentation by the Lexicon team we should make a Drilldown component.

bryceosterhaus:
I spent some time on this last week and got decently far until the animations. Here is my work so far, https://github.com/bryceosterhaus/clay/tree/drilldown

I'm on vacation this week, but I will pick it up once I am back if there hasn't been any other progress.

kresimir-coko:
Alright @bryceosterhaus I'll take a look at your implementation and see if I can find some cleaning up and refactoring to do.

kresimir-coko:
Hey @bryceosterhaus I've done some cleanup of the Drilldown, and after a [short exchange](https://liferay.slack.com/archives/G3JBR21HA/p1593608032287000?thread_ts=1593608032.286900&cid=G3JBR21HA) with @jbalsas on Slack decided to let you finish it. Here's the couple commits I added after rebasing https://github.com/kresimir-coko/clay/tree/drilldown

bryceosterhaus:
Sounds good! I will pick it back up tomorrow and hopefully have something solid submitted this week

kresimir-coko:
Fixed in #3338

PR #3338, opened 2020-06-16
feat(@clayui/drilldown): API proposition for ClayDrilldown

kresimir-coko:
Hey @bryceosterhaus @matuzalemsteles this is an API proposition for the Drilldown component #3337.

- I've separated the component in 4 low-level components, **Header**, **Body**, **Item** & **Footer**, following the design guideline submitted by Lexicon.
- I've tried to use as many Clay components as possible instead of random HTML elements, if I've missed any let me know.
- With this API the expectation of the user is to provide the component with the following props:
  - `confirmLabel` - Renders a Footer Button with the user's provided label.
  - `items` - The meat of this component, contains a list of objects representing the List Items, with support for `childItems` as well for some of that sweet drilldown action.
  - `title` - This is the string rendered inside the Header.
  - `undo` - Used the undo keyword as it's the one used in the design guideline. It renders an icon  in the header to go up a layer of navigation.

### Some issues I ran into
- Getting the correct heights of Items without a right icon. ClayButtonWithIcon sets the correct height of the Items that have it, but otherwise I couldn't figure out a clean solution using Bootstraps classes.
- When an Item doesn't have a left icon, the label should be distanced from the left the same amount as when it does have an icon. I couldn't figure out a clean solution to this.
- Jest has an issue with parsing the test file:
```
/Users/kresimircoko/Documents/Work/Projects/clay/node_modules/@clayui/css/lib/images/icons/icons.svg:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><symbol id="add-cell" viewBox="0 0 512 512">
SyntaxError: Unexpected token <
```


wincent:
> * Jest has an issue with parsing the test file:

It's choking on this line:

```
const spritemap = require('@clayui/css/lib/images/icons/icons.svg');
```

in the implementation. You shouldn't be hardcoding the spritemap like that (it's ok to do so in under `stories/`, because Jest isn't going to see those and try to evaulate the SVG as though it were JS.

pat270:
@kresimir-coko I made a demo for drilldown a while back. I think it needs to be updated since the documentation looks like it changed. See https://pat270.github.io/clay3-test-site/vclay-table-dd/content/dropdowns-drill-down/

bryceosterhaus:
So I think we might be slightly off here in thinking that DrillDown is its own independent package. I think if we treat it as a variant of DropDown, we can make things a little more consistent and not have to worry about designing from the ground up. From a birds eye view, I would expect a couple of things.

- `DrillDown` should be a high-level component from the `@clayui/drop-down` component. Perhaps `ClayDropDownWithDrilldown`, but seems sort of odd for naming... This also allows for us to utilize and share the markup between dropdown and drilldown.
- I think DrillDown should be high-level since we don't want to require users to wire up all the back/forth logic of the menus.
- I would expect a similar API that we have for `DropDownWithItems` but we would also have a `children` prop for the next level of items.  https://github.com/liferay/clay/blob/d5b484fe1889347880b074b795f19593f4501dbf/packages/clay-drop-down/src/DropDownWithItems.tsx#L20-L34
- We can still expose any needed low-level components in case users need an escape hatch.

Overall, I think by using drop-down as a guide, we shouldn't have to make too many API decisions for this component other than handling the Header component and the back and forth interaction.

Hope that helps give some direction, especially since I think the css is already in place for this.

pat270:
@kresimir-coko I spent a good chunk of today updating the markup for drilldown. I noticed there were several updates to the pattern. I should have something soon.

kresimir-coko:
Hey @bryceosterhaus I've removed the drilldown package and started work on the implementation as a high level Dropdown. Take a look and let me know if I've taken a wrong turn somewhere. The Dropdown component is pretty unwieldy so it might be a mess. There's still some work to be done before I fully understand how I can properly mirror the API within the Drilldown component.

packages/clay-drop-down/src/DropDownWithDrilldown.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+	/**
+	 * Element that is used as the trigger which will activate the dropdown on click.
+	 */
+	trigger: React.ReactElement;
+
+	/**
+	 * Text to be displayed in the Dropdown's header.
+	 */
+	headerText?: string;
```
bryceosterhaus, line 85: I don't think I would expect users to have to manually handle providing the headerText and the undo callback. Ideally we would handle the back/forth logic internally for the users and all they need to add is the "children" attribute to the item they want to have drilldown abilities.

So if we have an item like this
```js
[{
    name: 'Documents and Media',
    children: [
        {
            name: 'Pictures',
            href: '#'
        }
    ]
}]
```

The item should automatically render a simple dropdown list with one item with an arrow and no header. If the user clicks the "Documents and Media" item, it should then render its children list of items, and the header would read "Documents and Media" and a back arrow. When the back arrow is clicked, it should go back to the original list.

Ideally we can handle all of that logic internally so that the user doesn't have to wire anything up themselves other than providing the items.

You may even try just updating `DropDownWithItems` first to see if you can add the logic needed there and not have to start entirely from scratch. This way you can focus just on the back/forth logic and interactions, rather than the menu as a whole.

pat270:
Just a heads up, @jbalsas wanted to accommodate the use case outlined at https://issues.liferay.com/browse/LPS-114110. I don't understand where the drilldown pattern comes into play. I'm guessing the [4 column layout](https://issues.liferay.com/secure/attachment/360159/Highlighted%20Box.png) collapses into a drilldown in smaller screens?

|level 1|level 2 (Content)|level 2 (Workflow)|
|---|---|---|
|Content >|Content Dashboard|Process Builder|
|Workflow >|Asset Libraries|Submissions|
|Change Lists >|
|Sites >|

We will need to render a drilldown outside of a dropdown-menu.

kresimir-coko:
A good point you brought up @pat270. It's not mentioned in the [design guidelines](https://docs.google.com/document/d/1PVQwknqVIOQhf2JL0kEv0k6WHOExKOskiX-CxxWYRqs/edit).

@bryceosterhaus  It looks to me like there are plenty of usages of Drilldown outside of the context of Dropdown, which makes me think it shouldn't be a high-level variant of a Dropdown. 

I can see users using a Drilldown component within a Sidebar, for example, but I can't see it making sense for them to use a DropdownWithDrilldown within a Sidebar, or a 4 column layout, since in those cases the Drilldown is always visible, not a pop-up.

I'll keep working on the implementation within a Dropdown for now, focusing on the parts that aren't dependant on it being a Dropdown first, in case we wanna switch it up and make it it's own package.

jbalsas:
Hey @pat270, @kresimir-coko, yeah, all the info collapses into a drilldown as is seen in the ticket:

image-2020-06-09-14-07-38-271

You can see it in the [figma designs](https://www.figma.com/file/Ln1fkqNTqRX4nniHXFfLiy/LPS-114110-global-apps-responsive) for tablet and mobile

jbalsas:
> I'll keep working on the implementation within a Dropdown for now, focusing on the parts that aren't dependant on it being a Dropdown first, in case we wanna switch it up and make it it's own package.

Sounds like a good plan 👍

matuzalemsteles:
Arriving a little late here, After looking at the documents I'm a little inclined so that we start with a low-level component with the back and forth mechanism instead of a high-level for DropDown, apparently, we have a lot of use cases in DrillDown variations, the developer must be able to customize the content of each item, in the case of rendered inputs it also seems to be a legitimate case for the Page Editor and the case that Chema showed above shows a little that DrillDown can be coupled in other contexts without necessarily being in a DropDown.

I prefer that we start from the bottom up, it seems that this is where we have the most use cases here within the Portal, we will be able to support and help them first. DrillDown seems to be a variant for DropDown also for a simpler case. I also would not discard this work.

kresimir-coko:
Hey @matuzalemsteles I'm following the latest piece of advice that @bryceosterhaus suggested. I'm altering ClayDropDownWithItems as a POC, then we can see what we'll do afterwards.

> I prefer that we start from the bottom up, it seems that this is where we have the most use cases here within the Portal, we will be able to support and help them first.

I agree with the above quote, the way I've initially understood the component was like a vertical nav, with drilldown interactions, a header and a footer, the DropDown aspect of it is secondary I think.

matuzalemsteles:
> Hey @matuzalemsteles I'm following the latest piece of advice that @bryceosterhaus suggested. I'm altering ClayDropDownWithItems as a POC, then we can see what we'll do afterwards.

@kresimir-coko I see no problem with continuing this and merging, but looking more closely at the documentation, it at low-level would be more effective for the variations that seem to be the most real cases here. Perhaps this high-level implementation should just be a simple mechanism that is supported for DropDown.

kresimir-coko:
Yea, we really should make a low-level one @matuzalemsteles

kresimir-coko:
I think this is gonna take a bit of time to develop @bryceosterhaus. It has plenty of different variants as far as I've seen. Anyhow, I've played around with DropDownWithItems today, trying to use it as a base for some Drilldown interactions, as you advised.

I got stuck trying to show the header, for whatever reason on runtime I get `Warning: Received 'true' for a non-boolean attribute 'parent'.` IDK what's happening there but anyway, think about if we should make a low-level component as well. 

Just an FYI, I'll be taking next week off, so if you feel like taking over (even tomorrow) feel free to, if not, I'll just play around a bit more with it and report back my findings.

jbalsas:
> Just an FYI, I'll be taking next week off, so if you feel like taking over (even tomorrow) feel free to, if not, I'll just play around a bit more with it and report back my findings.

Who authorized that!!! 😱

bryceosterhaus:
Sorry @kresimir-coko, I think I gave bad advice as I misunderstand how much drilldown is actually used, my fault for not fully understanding that. I was going off of old non-official docs from drilldown, which made it seem like it was just used in dropdown. I can see how its used outside of that though, so I think you were right in creating its own package.

Now that I see more clearly, I agree with @matuzalemsteles  in that we should start from the bottom up. Once we get the low-level established, my guess is that this component might be used to create other high-levels in the future, like using it in dropdown. I think your work so far is still applicable though, so don't just throw it out. If I get a chance next week, I'll play around with it a bit, but I'll let you know if I get to it.

kresimir-coko:
> Who authorized that!!! 😱

@jbalsas uhm, remember when we almost got confused about our last 1:1 being cancelled, and me clarifying I won't be here for the next one? I put it into the calendar a couple weeks ago. Thought that's all the authorization needed 😕

jbalsas:
> @jbalsas uhm, remember when we almost got confused about our last 1:1 being cancelled, and me clarifying I won't be here for the next one? I put it into the calendar a couple weeks ago. Thought that's all the authorization needed 😕

I was just kidding, @kresimir-coko!

kresimir-coko:
Hey @bryceosterhaus I've sent another incomplete and scuffed batch of code playing around with DropDownWithItems. I did make progress but got stuck in the end when I realized that the entire Item should be clickable, instead of just the right icon (going back and forward works but the dropdown closes upon clicking on the item). It has an inkling of what a possible API could look like, so there's at least something.

kresimir-coko:
I'll continue work on this PR then, but I'll start from the bottom up to make a low-level component which we can then utilize for a high-level DropdownWithDrilldown.

 @jbalsas I assume there hasn't been any development regarding this like we decided to pause it or something?

wincent:
Any updates on this PR? Just asking because at the Teams Integration meeting last week I told them we were going to resume work on it, but I don't see any activity.

kresimir-coko:
@bryceosterhaus will pick it back up I think, I did some cleanup and passed it over to him, he confirmed [here](https://github.com/liferay/clay/issues/3337#issuecomment-654477512)

wincent:
Thanks @kresimir-coko.

bryceosterhaus:
@wincent picking it back up today.

bryceosterhaus:
closing in favor of https://github.com/liferay/clay/pull/3489

PR #3362, opened 2020-06-23
feat(clayui.com): make it easier to work with clay-css on clayui.com

### LINKED ISSUE #2891: Size of dependency graph
Just planting this one as a seed for discussion and thought, although it may not lead to any action. The question:

> **What can we (or should we) do about the size of the Clay dependency graph?**

---

The Clay dependency graph is very large.

The `yarn.lock` has 23k lines.

`yarn audit` reports:

```
13427 vulnerabilities found - Packages audited: 960275
Severity: 11 Low | 11 Moderate | 13405 High
```
(Note that it is obviously some double-counting there due to some kind of Yarn bug; there are 80k files in the `node_modules` directory and it must be counting the same packages repeatedly.)

Running `yarn outdated` 108 outdated dependencies:

```
Package                          Current      Wanted  Latest     Workspace                Package Type           URL                                                                                          
@babel/cli                       7.4.4        7.8.4   7.8.4                               devDependencies        https://babeljs.io/                                                                          
@babel/core                      7.4.5        7.8.4   7.8.4                               devDependencies        https://babeljs.io/                                                                          
... 104 lines omitted ...
yeoman-generator                 0.19.2       0.19.2  4.5.0      generator-clay-component dependencies           http://yeoman.io                                                                             
yosay                            1.2.1        1.2.1   2.0.2      generator-clay-component dependencies           http://yeoman.io 
```

[Full output in this gist](https://gist.github.com/wincent/356415c845f4502fb0a39d101564b5b8)

Screenshot showing the output in all of its glorious red/yellow:

clay___yarn

Obviously, all of these dependencies bring some value, otherwise we wouldn't have added them. Most — but not all — of them are `devDependencies`. But I wanted to bring up the size of the graph because these dependencies are not free:

- The larger the graph, the higher the cost of keeping it up to date (something sadly necessitated by security problems, which mean that we don't upgrade because we need access to new features or bug fixes, but because we're forced to by an audit).
- Of the 108 outdated dependencies, 50 of them are semver-breaking major updates. The cost of updating these is hard to evaluate but probably pretty large. At some point, the cost may start to exceed the associated benefits.

Just to put things in perspective, liferay-portal itself "only" has an 18k-line lockfile (about 82% of the size of Clay's).

As I said above, I don't know if any specific action needs to or will be taken as the result of this ticket, but I just wanted to raise awareness about the cost-benefit trade-off involved in maintaining this dependency graph. We might want to check to see if there are any non-essential dependencies, or dependencies that aren't pulling their weight. Or we might merely keep this discussion in mind the next time a PR proposes to add another dependency to the graph.

Sighed, your grumpy, dependency-allergic curmudgeon, @wincent.

clint-eastwood

bryceosterhaus:
fixes #2891


clayui.com/gatsby-browser.js
```diff
@@ -3,8 +3,28 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-require('./src/styles/main.scss');
-
 // A stub function is needed because gatsby won't load this file otherwise
 // (https://github.com/gatsbyjs/gatsby/issues/6759)
-exports.onClientEntry = () => {};
+export const onClientEntry = () => {
+	if (process.env.GATSBY_CLAY_NIGHTLY === 'true') {
```
bryceosterhaus, line 9: Had to pull this out on its own in order for gatsby + webpack to handle these requires correctly.

pat270:
@bryceosterhaus Good work, I can work with this. I had to use `yarn develop -H 192.168.1.xxx -p 8000` to make it work with my IE11 vm. There's an issue on my end where clayui.com doesn't render in IE11 in dev mode, but we can figure that out later.

clayui.com/gatsby-browser.js
matuzalemsteles: Maybe just adding one more conditional `isNullOrTrue` can solve it? 🤔

```javascript
const isNullOrTrue = (val) => val === 'true' || val === null || val === undefined;
// or
const isNullOrTrue = (val) => val === 'true' || val == null;
```

I don't know if this is related.

bryceosterhaus: @matuzalemsteles I think its related to the pre-processing of find/replace environment variables. I remember dealing with this pack on AC with webpack.

matuzalemsteles: Hmm, no problem. I think maintaining that will not be a problem. It might even help because it could happen that you saved that key in localStorage and when you access the site it will look different.

bryceosterhaus:
@pat270 let me know if you run into any issues while using this!

PR #3369, opened 2020-06-26
fix(@clayui/modal): removes the duplication of modal- {size} and the size of the forced width to the default value

### LINKED ISSUE #3107: ClayModal size CSS class is applied twice
See `['modal-${size}']: size`, [here](https://github.com/liferay/clay/blob/2f5b60bb897cf1c6d4530af8167921a58d563a1d/packages/clay-modal/src/Modal.tsx#L92) and [here](https://github.com/liferay/clay/blob/2f5b60bb897cf1c6d4530af8167921a58d563a1d/packages/clay-modal/src/Modal.tsx#L99).

This is probably not intended. The effect is that is causes double padding, and awkward overriding in implementations. It it was intended, it's a bad practice that we should avoid.

| Tech  | Version |
| ----- | ------- |
| Clay  | v3  |

matuzalemsteles:
Fixes #3107

With `display: table;` we can make the content fit children and we don’t need to force the size of the element, we could achieve the same with `width: max-content` but it doesn’t support IE 11 😞

### DIFF
--- a/packages/clay-modal/src/Modal.tsx
+++ b/packages/clay-modal/src/Modal.tsx
@@ -68,9 +68,6 @@ const ClayModal: React.FunctionComponent<IProps> = ({
 
 	React.useEffect(() => observer.dispatch(ObserverType.Open), []);
 
-	// Defines a default Modal size when size is not set.
-	const maxWidth = size ? {} : {maxWidth: '600px'};
-
 	const [show, content] =
 		observer && observer.mutation ? observer.mutation : [false, false];
 
@@ -88,11 +85,11 @@ const ClayModal: React.FunctionComponent<IProps> = ({
 				})}
 			>
 				<div
-					className={classNames({
-						[`modal-${size}`]: size,
-					})}
 					ref={modalBodyElementRef}
-					style={{margin: 'auto', ...maxWidth}}
+					style={{
+						display: 'table',
+						margin: 'auto',
+					}}
 				>
 					<div
 						className={classNames('modal-dialog', {
--- a/packages/clay-modal/src/Modal.tsx
+++ b/packages/clay-modal/src/Modal.tsx
@@ -5,7 +5,7 @@
 
 import {ClayPortal} from '@clayui/shared';
 import classNames from 'classnames';
-import React from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 import warning from 'warning';
 
 import Body from './Body';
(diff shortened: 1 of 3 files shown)


matuzalemsteles:
Hey @pat270, do you see any side effects that this can cause?

pat270:
@matuzalemsteles `display: table;` seems to make the modal smaller in the first example at https://clayui.com/docs/components/modal.html.

matuzalemsteles:
oh yeah, thanks @pat270. Apparently this is going to break a lot, I can’t understand why it’s not breaking in the Storybook, see https://deploy-preview-3369--next-storybook-clayui.netlify.app/?path=/story/components-claymodal--default.

I'm thinking of going with another approach, taking the width of Modal, and adding the `max-width` to the style.

pat270:
I think it's because of the lorem ipsum text is long enough to force display table to hit the max width.

matuzalemsteles:
This is true, I go with another approach.

pat270:
@matuzalemsteles I thought about it more, if you need an outer container I can add a class `modal-dialog-outer` that just sets `max-width` and `margin: auto` for each modal size. This way the `max-width` can be changed via CSS only.

matuzalemsteles:
hey @pat270, that would be nice, but I think it would get into the problem reported by @markocikos in #3107. As this is an edge case implementation to correct some problems with the use of Portals, I don't think we need to force this into CSS. And it seems that we also have some problems with the sizes with the old Modals in DXP 🤷‍♂️.

matuzalemsteles:
Hey @pat270, After the entry of #3418, the implementation of the click outside Modal was broken, because now `.modal-dialog` reaches the screen size with `.modal-dialog-centered`, the click on the side works well but below and over the modal it ends up breaking. I'm wondering if we can use css, maybe change the implementation of `.modal-dialog-centered` to:

```css
.modal-dialog-centered {
    position: absolute;
    margin: auto !important;
    top: 50%;
    -ms-transform: translateY(-50%);
    transform: translateY(-50%) !important;
}
```

What do you think? do you see any problem?

Another approach I'm thinking about for this container problem, instead of adding the style to the container and checking the size of the Modal to be dynamic, I'm wondering if we could make `.modal-dialog` with `position: absolute;` this makes the container not be able to have visual effect and has no problems with the events, so we don't need to force the style of the container or change the `.modal-dialog-centered`, what do you think? do you see any problems about that too?

wincent:
Any update here? @liferay-echo is waiting on this fix.

matuzalemsteles:
hey @pat270 just pinging you again, if you missed the notification.

matuzalemsteles:
Resent #3499

Issue #3340, opened 2020-06-16
Unable to move focus from dropdown menu item to next button


bryceosterhaus:
Spawned from https://github.com/liferay/clay/pull/3334#issuecomment-644602233

https://codesandbox.io/s/sleepy-buck-3z82f?file=/public/index.html

PR #3370, opened 2020-06-26
fix(@clayui/date-picker): Adds FocusScope for an appropriated tab navigation on DatePicker's calendar

### LINKED ISSUE #3368: Date cannot be selected using the keyboard
-   [ ] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

Go to https://clayui.com/docs/components/date-picker.html and try to select a date using only the keyboard. After the date-picker expands when pressing _tab_ the focus goes to the next field and the date cannot be selected.
Note: The date-picker HTML is being appended at the end of the page body.
Users should be able to focus on the date-picker using the keyboard.
## Environment
Tested on Chrome and Firefox.
| Tech  | Version |
| ----- | ------- |
| Clay  | v3.4.0  |
| React |         |

diegonvs:
### Test Case:

> Add a ClayDatePicker to the Storybook Form Page Demo

Note that when opening the Date Picker calendar using SPACE, we were unable to focus any elements within the DatePicker using TABS.

---

To solve this problem, I used FocusScope and fixed the Focus problem when opening the Date Picker Calendar.

### Questions/Issues:

When going through the entire focus, the Focus scope does not close the component and focuses on the next component in the DOM tree (which is still right 🤔)

When closing the Date Picker Calendar by pressing the ESC key, the next element to focus according to the focus scope is the closest element to the document.body (where the ClayDropDown.Menu element portal was opened) making the focus stay on the submit.

### Commits:

- docs(@clayui/date-picker): Adds a DatePicker on FormPage story
- fix(@clayui/date-picker): Adds FocusScope for navigating int o DatePicker calendar component using TABS

Closes #3368

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -345,6 +345,29 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 		 */
 		const handleCalendarButtonClicked = () => setExpanded(!expanded);
 
+		/**
+		 * Handle with the focus when it's outside of the component
+		 * In this case, forces the state of expanded to be false
+		 */
+		const handleFocus = (event: FocusEvent) => {
+			if (
+				dropdownContainerRef.current &&
+				!dropdownContainerRef.current.contains(event.target as Node) &&
+				triggerElementRef.current &&
+				!triggerElementRef.current.contains(event.target as Node)
+			) {
+				setExpanded(false);
+			}
+		};
+
+		React.useEffect(() => {
+			document.addEventListener('focus', handleFocus, true);
+
+			return () => {
+				document.removeEventListener('focus', handleFocus, true);
+			};
+		}, [handleFocus]);
+
 		return (
 			<FocusScope arrowKeysLeftRight>
 				<div className="date-picker">
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -7,6 +7,7 @@ import Button from '@clayui/button';
 import DropDown from '@clayui/drop-down';
 import {ClayInput} from '@clayui/form';
 import Icon from '@clayui/icon';
+import {FocusScope} from '@clayui/shared';
 import React from 'react';
 
 import DateNavigation from './DateNavigation';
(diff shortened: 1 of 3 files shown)


jbalsas:
> To solve this problem, I used FocusScope and fixed the Focus problem when opening the Date Picker Calendar.

Nice!! I haven't tested this myself, but sounds like a great improvement! 👏 

> When going through the entire focus, the Focus scope does not close the component and focuses on the next component in the DOM tree (which is still right 🤔)
>
> When closing the Date Picker Calendar by pressing the ESC key, the next element to focus according to the focus scope is the closest element to the document.body (where the ClayDropDown.Menu element portal was opened) making the focus stay on the submit.

This is probably the same issue as [Unable to move focus from dropdown menu item to next button](https://github.com/liferay/clay/issues/3340)

kresimir-coko:
Hey @diegonvs I've tested your Form Page story and I'm unsure if I did something wrong but here's the behaviour I'm getting:
- [x] Click on Country input and Tab to Date
- [x] 2. Tab again to focus the Calendar icon
- [x] 3. Press Space which opens the Date Picker
- [x] 4. Tab a few times, find a date, press Enter to select the date, press Esc to close the Date Picker
- [ ] 5. Tab is nowhere to be found, after tabbing a few times it returns me back to the top left of the Storybook page. It doesn't focus on anything related to the DOM around the Date Picker

From what I understand, the goal of this PR was the make sure the last checkbox above is ticked, that the Tab focuses to the next input field, or whatever comes next in the DOM.

matuzalemsteles:
> When going through the entire focus, the Focus scope does not close the component and focuses on the next component in the DOM tree (which is still right 🤔)

I'm not sure if it should close, FocusScope does not handle it, so it should be a mechanism in Date Picker if it's the right behavior. If you think of it as the same behavior as DropDown, it must be an implementation similar to the one we have in DropDown https://github.com/liferay/clay/blob/master/packages/clay-drop-down/src/DropDown.tsx#L106-L123

> When closing the Date Picker Calendar by pressing the ESC key, the next element to focus according to the focus scope is the closest element to the document.body (where the ClayDropDown.Menu element portal was opened) making the focus stay on the submit.

Remembering that the use of Portal here makes no difference if it is in the body because we use the React tree, where the tree is preserved when using the Portal, so if the component is inside a React application it will go to the next focusable element (This is not true when there are two different React trees on the page).

If it is just a single component loose in the DOM, what else happens in DXP, we do nothing, we just let the browser handle the rest but as it is in the body, in this case, it will search from the body, Chema mentioned the issue with this problem. Perhaps a solution for this would be to return the focus to the first element before the Portal and without a `preventDefault` for the browser to look for the next focusable element from that.

The same happens in Storybook cases when the component is unique in the example but the focus is preserved because the component is rendered inside an iframe

packages/clay-date-picker/src/index.tsx
```diff
@@ -345,105 +346,110 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
-						className="date-picker-dropdown-menu"
-						data-testid="dropdown"
-						onSetActive={setExpanded}
-						ref={dropdownContainerRef}
-					>
-						<div className="date-picker-calendar">
-							<DateNavigation
-								ariaLabels={ariaLabels}
-								currentMonth={currentMonth}
+			<FocusScope arrowKeysLeftRight>
```
matuzalemsteles, line 349: I think that here, it doesn't make much sense to enable the `arrowKeysLeftRight`, the behavior is very strange. I think keeping the focus is enough to use Tab and Shift+Tab.

I think the use of arrow keys here must be its own implementation because the behavior can be different, but we have to evaluate with Lexicon, as this is based on the native... (And it looks like the new native Date Picker looks very different now, you get stuck in a loop with navigation 🤷‍♂️)

diegonvs:
> From what I understand, the goal of this PR was the make sure the last checkbox above is ticked, that the Tab focuses to the next input field, or whatever comes next in the DOM.

Yes, the main goal it is but during fixing it I found this bug :p

packages/clay-date-picker/src/index.tsx
diegonvs: I was wondering navigating using left and right navigation due to https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html, However I noticed We aren't reaching the same result of this example w3c example.

Hey @draconux, Do you have a design/a11y perspective on this?

diegonvs:
> I'm not sure if it should close, FocusScope does not handle it, so it should be a mechanism in Date Picker if it's the right behavior. If you think of it as the same behavior as DropDown, it must be an implementation similar to the one we have in DropDown https://github.com/liferay/clay/blob/master/packages/clay-drop-down/src/DropDown.tsx#L106-L123

Updated here: https://github.com/liferay/clay/pull/3370/commits/90d22d4edf54d5e876b079a227adcf17dc5a06a2

Thanks Matu!

packages/clay-date-picker/src/index.tsx
drakonux: I've created [LEXI-975](https://issues.liferay.com/browse/LEXI-975) to work on a new definition but there is no priority to do it (as far as I know). 

About the focus management, I think we should follow the W3 approach. 

- Thus, TAB and Shift + TAB to move the focus between focusable elements (orange numbers).
- Once the focus enters in the date zone (Red):
- If the user press TAB again the focus moves to the next area (Blue)
- ↑↓ are enabled to move the focus within the blue area

focus-date-picker

But I need time to work on it. My advice is that we are following this approach as we defined in other components such as the Vertical Bar Tabs.

matuzalemsteles: Thanks @drakonux for the information. @diegonvs I think we can deal with that later, we disable `arrowKeysLeftRight` for DatePicker. And we hope for the implementation so that @drakonux has more time to think about it.

PR #3371, opened 2020-06-27
fix(@clayui/card): When pressing `enter` or `space`, triggers a click event on CardWithNavigation

### LINKED ISSUE #3365: @clayui/card is not clickable with keyboard

[This example](https://codesandbox.io/s/interesting-driscoll-g66ot) uses `ClayCardWithNavigation` with an onClick handler.

I've seen that ClayCard uses a `div` with `role=button` internally when the card should be interactive, and using this markup instead of a regular `button` tag doesn't have the same browser behavior.

1. Use `ClayCardWithNavigation` with an `onClick` prop.
2. Navigate to the component using keyboard `tabbing`
3. Press `enter`/`space` key

onClick handler is executed.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay Card  | v3.2.1  |
| React | 16.10.2 |

diegonvs:
## Test Case

> Open Storybook on CardWithNavigation story: https://storybook.clayui.com/?path=/story/components-claycard--cardwithnavigation, try to focus the onClick card and then press SPACE or ENTER key.

The original behaviour is when a button is focused, this element be triggered by SPACE or a ENTER key but when using a div with a role="button" this isn't possible.

## Description

I needed to hack the onKeyDown event for detecting a SPACE or ENTER keydown event passing the event of keydown for these keys to onClick callback.

## Questions/Issues:

Is there a way for avoiding using a div with a role="button" attribute? /cc @pat270

See https://twitter.com/scottohara/status/1183020913509502978

Closes #3365

### DIFF
--- a/packages/clay-card/package.json
+++ b/packages/clay-card/package.json
@@ -32,7 +32,6 @@
 		"@clayui/label": "^3.3.1",
 		"@clayui/layout": "^3.1.0",
 		"@clayui/link": "^3.2.0",
-		"@clayui/shared": "^3.1.2",
 		"@clayui/sticker": "^3.1.1",
 		"classnames": "^2.2.6"
 	},
--- a/packages/clay-card/package.json
+++ b/packages/clay-card/package.json
@@ -32,6 +32,7 @@
 		"@clayui/label": "^3.3.1",
 		"@clayui/layout": "^3.1.0",
 		"@clayui/link": "^3.2.0",
+		"@clayui/shared": "^3.1.2",
 		"@clayui/sticker": "^3.1.1",
 		"classnames": "^2.2.6"
 	},
--- a/packages/clay-card/src/CardWithNavigation.tsx
+++ b/packages/clay-card/src/CardWithNavigation.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {noop} from '@clayui/shared';
 import ClayIcon from '@clayui/icon';
 import ClayLayout from '@clayui/layout';
 import ClaySticker from '@clayui/sticker';
@@ -66,6 +67,7 @@ export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
 	horizontalSymbol = '',
 	href,
 	onClick,
+	onKeyDown = noop,
 	spritemap,
 	title,
 }: IProps) => {
@@ -86,6 +88,8 @@ export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
 						onClick(event);
 					}
 				}
+
+				onKeyDown(event);
 			}}
 			tabIndex={0}
 		>
--- a/packages/clay-card/src/CardWithNavigation.tsx
+++ b/packages/clay-card/src/CardWithNavigation.tsx
@@ -3,9 +3,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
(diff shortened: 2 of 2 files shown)


packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -64,7 +74,19 @@ export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
...
+					(event && event.keyCode === KEYCODES.ENTER) ||
+					(event && event.keyCode === KEYCODES.SPACE)
+				) {
+					event.preventDefault();
+
+					if (onClickCallback) {
+						onClickCallback(event);
+					}
+				}
+			}}
```
matuzalemsteles, line 89: I think we should always call `onKeyDown` callback here.

packages/clay-card/src/CardWithNavigation.tsx
```diff
@@ -49,13 +54,18 @@ interface IProps {
...
+};
+
 export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
 	children,
 	description,
 	horizontal = false,
 	horizontalSymbol = '',
 	href,
-	onClick,
+	onClick: onClickCallback,
```
matuzalemsteles, line 68: Why not just keep `onClick`?

diegonvs: I forgot to remove this alias, I was using in a older implementation 😅

packages/clay-card/src/CardWithNavigation.tsx
diegonvs: If We change the element to a button, yes.

> The original behaviour is when a button is focused, this element be triggered by SPACE or a ENTER key but when using a div with a role="button" this isn't possible.

matuzalemsteles: What I'm saying is, regardless of not going to `button`, we should allow people to add `onKeyDown` listeners here.

packages/clay-card/src/CardWithNavigation.tsx
matuzalemsteles: haha np

packages/clay-card/src/CardWithNavigation.tsx
diegonvs: Oh, gotcha.

pat270:
@diegonvs @matuzalemsteles sorry I missed this. The problem of wrapping the card in `button` or `a` is that form elements, buttons, or other links aren't valid inside those tags. We should use `div` because we stuff so much in our cards.

The way I think it should work is `<div tabindex="0" role="button">` should dispatch a click event when keydown is enter.

matuzalemsteles:
Thanks @pat270, for the info.

packages/clay-card/package.json
```diff
@@ -32,6 +32,7 @@
 		"@clayui/label": "^3.3.1",
 		"@clayui/layout": "^3.1.0",
 		"@clayui/link": "^3.2.0",
+		"@clayui/shared": "^3.1.2",
```
matuzalemsteles, line 35: Looking at this better, I think it would not be very useful to have `shared` as a dependency only for noop, I think we can remove this.

matuzalemsteles: cc @diegonvs

diegonvs: Sure, I'll remove

PR #3382, opened 2020-06-29
feat(@clayui/css): Absorb Bootstrap's _card.scss into Clay CSS _cards…

pat270:
….scss

issue #3149


wincent:
Again, ci/codesandbox never finished, but I believe it's just flaky, so going to merge.

bryceosterhaus:
@wincent codesandbox ci can be a bit flakey, and its sort of a blackbox so its hard to know where the issue falls. So long as we can build via `yarn build`, I think we should be okay.

PR #3383, opened 2020-06-29
feat(@clayui/css): Absorb Bootstrap's _breadcrumb.scss into Clay CSS _breadcrumbs.scss

pat270:
issue #3149

### DIFF
--- a/packages/clay-css/src/scss/_bs4.scss
+++ b/packages/clay-css/src/scss/_bs4.scss
@@ -17,7 +17,7 @@
 @import 'bootstrap/nav';
 @import 'bootstrap/navbar';
 @import 'bootstrap/card';
-@import 'bootstrap/breadcrumb';
+// @import 'bootstrap/breadcrumb';
 @import 'bootstrap/pagination';
 @import 'bootstrap/badge';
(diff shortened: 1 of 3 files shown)


packages/clay-css/src/scss/_bs4.scss
```diff
@@ -17,7 +17,7 @@
 @import 'bootstrap/nav';
 @import 'bootstrap/navbar';
 @import 'bootstrap/card';
-@import 'bootstrap/breadcrumb';
+// @import 'bootstrap/breadcrumb';
```
wincent, line 20: Out of curiosity, why are we commenting these as opposed to deleting the lines?

pat270: @wincent That is a good question. It was to help me keep track of what components have been converted. Looking at it now, deleting the lines would accomplish the same haha. At the end of the conversion this, this file is going away.

wincent: kill-it-with-fire

wincent:
Skipping this one due to conflict.

PR #3384, opened 2020-06-29
feat(@clayui/css): Absorb Bootstrap's _pagination.scss into Clay CSS …

pat270:
…_pagination.scss

issue #3149


wincent:
Again ci/codesandbox showing as never finished, but I can't see how this PR would have caused that, so going to merge.

PR #3398, opened 2020-06-29
feat(@clayui/css): Absorb Bootstrap's _print.scss into Clay CSS _prin…

pat270:
…t.scss

issue #3149


packages/clay-css/src/scss/components/_print.scss
```diff
@@ -0,0 +1,124 @@
+// Source: https://github.com/h5bp/main.css/blob/master/src/_print.css
```
wincent, line 1: @pat270: I'm going to merge this, to thin the queue a bit, but we should make sure that we include the license info for this one too...

https://github.com/h5bp/main.css/blob/master/LICENSE.txt

PR #3403, opened 2020-06-30
feat(@clayui/css): Absorb Bootstrap's _btn-group.scss into Clay CSS _btn-group.scss

pat270:
feat(@clayui/css): Button adds placeholders `%clay-btn-monospaced-lg`, `%clay-btn-monospaced-sm` so we can extend with Sass `@extend` without unnecessary selectors

issue #3149


wincent:
Skipping due to conflict.

PR #3409, opened 2020-07-01
feat(@clayui/button): Add label prop to ClayButtonWithIcon

### LINKED ISSUE #3407: ClayButton should have an example of a ClayButtonWithIcon with a label
In the Markup page of the Button we have a markup example of a button with a label. 

ClayButtonWithIcon's API table doesn't even mention a possible label.

Here's the section about the ClayButtonWithIcon that has text inside as well.
image

Edit: after a quick investigation, the ClayButtonWithIcon doesn't even support a label, so that's something that should be added to the implementation.

kresimir-coko:
Fixes #3407 

While adding the `label` prop I came to the conclusion that we could also allow the users to change the positioning of the Icon, which is why I added an `iconPosition` prop that wraps the icon in a `span.inline-item`.

I also updated the related story, added a new test, and added to the documentation.

### DIFF
--- a/packages/clay-button/README.mdx
+++ b/packages/clay-button/README.mdx
@@ -9,6 +9,7 @@ sibling: 'docs/components/css-buttons.html'
 import {
 	ButtonDisplayTypes,
 	ButtonGroup,
+	ButtonWithIcon,
 } from '$packages/clay-button/docs/index';
 
 <div class="nav-toc-absolute">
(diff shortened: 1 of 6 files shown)


bryceosterhaus:
Hey @kresimir-coko, I think we should keep the API as is and not add a label option. See the issue for details

kresimir-coko:
@bryceosterhaus alrighty, closing this then. The issue can stay open for someone to pick up and make an example in the docs. 👍

PR #3413, opened 2020-07-01
fix(@clayui/date-picker): Fixes an edge case when the dot button does not select the current date

### LINKED ISSUE #3354: What is the expected behavior when the user clicks on the "Select current date" icon on date picker fields?
ActualResult_Jun22_LPS-TBD_CurrentDate-Picker (1)

## What is the expected behavior when the user clicks on the "Select current date" icon on date picker fields?

Actual Result 1: **When the input is empty** the current date is selected and the page about the current date is shown.

Actual Result 2: **When there is value in the input** the page about the current date is shown.

matuzalemsteles:
Instead of relying on `initialMonth` to set the day, we rely on a `new Date()` created when the file is loaded. The `initialMonth` API can be modified to a date unrelated to the current day, as is the case with DDM.

Fixes #3354


matuzalemsteles:
I will go ahead with this, it is not very complex but if someone sees a problem we can send changes later.

PR #3417, opened 2020-07-01
feat(@clayui/data-provider): Adds the new fetcher API and deprecated the link behavior in accepting async function

matuzalemsteles:
This PR introduces two changes and at the same time improvements for cache validation for custom fetch cases, this is explained better in the documentation.

- Deprecates the behavior of the `link` API to accept an async function (Throws a warning if used)
- Adds the new API `fetcher`

The `fetcher` API facilitates the custom fetch use cases and improves the cache validation when used, mainly for the Portal use cases that we need to use the `fetch` from ` frontend-js-web`.

```javascript
import {fetch} from 'frontend-js-web';

useResource({
    fetcher: fetch,
    link: 'https://clay.data/',
}); 
```

### DIFF
--- a/packages/clay-data-provider/README.mdx
+++ b/packages/clay-data-provider/README.mdx
@@ -110,7 +110,7 @@ The cache is guided by a policy, use the `fetchPolicy` prop to enable and config
 ### Data Fetching
 
 ```javascript
-const {resource} = useResource({fetcher, link});
+const {resource} = useResource({fetch, link});
 ```
 
 This is an API that replaces the `link` behavior of receiving an async function, this did not allow us to validate the cache correctly because we do not have the URL view. This API is more friendly and has a unique responsibility, you may be able to pass an async function that accepts the `link` and `options`, and return the data. `useResource` will use its async function instead of the `fetch` default.
@@ -120,10 +120,8 @@ This is an API that replaces the `link` behavior of receiving an async function,
 ```javascript
 import fetch from 'unfetch';
 
-const fetcher = (url, options) => fetch(url, options);
-
 const App = () => {
-	const {resource} = useResource({fetcher, link: 'https://clay.dev'});
+	const {resource} = useResource({fetch, link: 'https://clay.dev'});
 	// ...
 };
 ```
--- a/packages/clay-data-provider/README.mdx
+++ b/packages/clay-data-provider/README.mdx
@@ -26,6 +26,8 @@ import {
     -   [Network Status](#network-status)
     -   [Variables change](#variables-change)
     -   [Caching data](#caching-data)
+    -   [Data Fetching](#data-fetching)
+        -   [Fetch](#fetch)
 -   [Advanced](#advanced)
     -   [Avoiding thundering herd problem](#avoiding-thundering-herd-problem)
     -   [Caching data at root level](#caching-data-at-root-level)
@@ -104,6 +106,27 @@ The cache is guided by a policy, use the `fetchPolicy` prop to enable and config
 		storageMaxSize
 	</code> API. <b>Each new query is equivalent to 1 size</b>.
 </div>
+
+### Data Fetching
+
+```javascript
+const {resource} = useResource({fetcher, link});
+```
+
+This is an API that replaces the `link` behavior of receiving an async function, this did not allow us to validate the cache correctly because we do not have the URL view. This API is more friendly and has a unique responsibility, you may be able to pass an async function that accepts the `link` and `options`, and return the data. `useResource` will use its async function instead of the `fetch` default.
+
+#### Fetch
+
+```javascript
+import fetch from 'unfetch';
+
+const fetcher = (url, options) => fetch(url, options);
+
+const App = () => {
+	const {resource} = useResource({fetcher, link: 'https://clay.dev'});
+	// ...
+};
+```
 
 ## Advanced
 
--- a/packages/clay-data-provider/src/useResource.ts
+++ b/packages/clay-data-provider/src/useResource.ts
@@ -23,6 +23,7 @@ interface IResource extends IDataProvider {
 
 const useResource = ({
 	fetchDelay = 300,
+	fetcher,
 	fetchOptions,
 	fetchPolicy = FetchPolicy.NoCache,
 	fetchTimeout = 6000,
@@ -155,7 +156,7 @@ const useResource = ({
 
 		warning(
 			uri.searchParams.toString() === '',
-			'DataProvider: We recommend that instead of passing parameters over the link, use the variables API. \n More details: https://next.clayui.com/docs/components/data-provider.html'
+			'DataProvider: We recommend that instead of passing parameters over the link, use the variables API. \n More details: https://clayui.com/docs/components/data-provider.html'
 		);
 
 		if (!variables) {
@@ -170,6 +171,11 @@ const useResource = ({
 	const doFetch = (retryAttempts = 0) => {
 		let promise: Promise<any>;
 
+		warning(
+			typeof link === 'string',
+			'DataProvider: The behavior of the `link` accepting a function has been deprecated in favor of the `fetcher` API. \n More details: https://clayui.com/docs/components/data-provider.html#data-fetching'
+		);
+
 		switch (typeof link) {
 			case 'function':
 				promise = link(
@@ -182,12 +188,20 @@ const useResource = ({
 					).searchParams.toString()
 				);
 				break;
-			case 'string':
-				promise = fetch(
-					getUrlFormat(link, variables),
-					fetchOptions
-				).then((res) => res.json());
+			case 'string': {
+				const fn = fetcher ?? fetch;
+
+				promise = fn(getUrlFormat(link, variables), fetchOptions).then(
+					(res) => {
+						if (fetcher) {
+							return res;
+						}
+
+						return res.json();
+					}
+				);
 				break;
+			}
 			default:
 				return null;
 		}
--- a/packages/clay-data-provider/src/useResource.ts
+++ b/packages/clay-data-provider/src/useResource.ts
@@ -192,12 +192,12 @@ const useResource = ({
 				const fn = fetcher ?? fetch;
 
 				promise = fn(getUrlFormat(link, variables), fetchOptions).then(
-					(res) => {
-						if (fetcher) {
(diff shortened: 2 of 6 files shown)


packages/clay-data-provider/README.mdx
```diff
@@ -105,6 +107,27 @@ The cache is guided by a policy, use the `fetchPolicy` prop to enable and config
...
+```
+
+This is an API that replaces the `link` behavior of receiving an async function, this did not allow us to validate the cache correctly because we do not have the URL view. This API is more friendly and has a unique responsibility, you may be able to pass an async function that accepts the `link` and `options`, and return the data. `useResource` will use its async function instead of the `fetch` default.
+
+#### Fetch
+
+```javascript
+import fetch from 'unfetch';
+
+const fetcher = (url, options) => fetch(url, options);
```
wincent, line 123: This is a do-nothing pass-through wrapper, which seems like unnecessary indirection; ie. functionally equivalent to:

```js
const fetcher = fetch;
```

I would expect that if people actually wanted to use `unfetch`, they would probably write this:

```js
import fetcher from 'unfetch';

const App = () => {
  const {resource} = useResource({fetcher, link: 'https://clay.dev'});
};
```

Which begs the question, why isn't the property just called `fetch`, so that people can write?

```js
import fetch from 'unfetch';

const App = () => {
  const {resource} = useResource({fetch, link: 'https://clay.dev'});
};
```

packages/clay-data-provider/src/useResource.ts
```diff
@@ -170,6 +171,11 @@ const useResource = ({
 	const doFetch = (retryAttempts = 0) => {
 		let promise: Promise<any>;
 
+		warning(
+			typeof link === 'string',
+			'DataProvider: The behavior of the `link` accepting a function has been deprecated in favor of the `fetcher` API. \n More details: https://clayui.com/docs/components/data-provider.html#data-fetching'
+		);
+
```
wincent, line 178: Do we really need to deprecate this? The cost of maintaining it seems to be very close to zero, and if we spew a warning into the console we are really just creating churn for little benefit.

packages/clay-data-provider/README.mdx
matuzalemsteles: Yeah, this wrapper is just for a demonstration that people can add additional manipulations to their own implementation if necessary, as an example in the Storybook, this example is actually incomplete, I forgot to update. The dev must handle the return of the data, in this case:
```javascript
import unfetch from 'unfetch';

const fetcher = (url, options) => unfetch(url, options).then((res) => res.json());

const {resource} = useResource({fetcher, link: 'https://clay.dev'});
```

I think that is up to whoever is using it, in deciding to pass the fetcher without a wrapper, no problem, but he has to respect the signature:
```typescript
fetcher?: (
	link: string,
	init?: RequestInit | undefined
) => Promise<Response>;
```

matuzalemsteles:
> Do we have an issue written up somewhere to get more context on what is driving this?

I failed to create an issue explaining it better 😞, sorry. But the biggest reason is to try to remove this option from pass a `fetcher` using the` link` API because it is very bad for DataProvider to try to validate the cache without having the visibility of `URL`, in the end, it ends up that don’t cache. We use the URL to fortify the cache key generation.

I saw some cases inside DXP that are using `useResource` and not DXP `fetch` but are having to pass all the common configuration that DXP `fetch` is doing. The reverse is true, too.

> I get the sense that a pretty common use case in DXP is going to be using the fetch wrapper from frontend-js-web. But it may be that setting that up via React context once at the top of the app could be more useful that passing it everywhere that we use useResource. Understanding the use cases that are driving this could help us make the right decision here (note: that might mean offering control via both context and props; it doesn't necessarily have to be an either/or thing).

Yeah, that makes a lot of sense. We can probably create a Context for the Data Provider to handle global configurations. I think we can use this Context in the `render` of the `frontend-js-react-web` so that people do not pass manually.

packages/clay-data-provider/src/useResource.ts
matuzalemsteles: As I said in the comment below is that using the link to pass an async function brings some problems to the Data Provider, so the idea is to add warnings so that we can remove this behavior in version 4.0 and just stay with the `fetcher` API .

packages/clay-data-provider/README.mdx
wincent: The trouble is, whatever you use as an example is what people will blindly copy-paste into the code. So, that's why I'm commenting on a useless wrapper *in documentation*, not because I am pedantic (I am), but because I don't want to see useless wrappers *in PRs* (which I'll then have to comment on there instead)...

Also, please don't forget to answer my question:

> Which begs the question, why isn't the property just called fetch

Remember that a key use case in DXP will be:

```js
import {fetch} from 'frontend-js-web';

const App = () => {
  const {resource} = useResource({fetch, link: 'https://clay.dev'});
};
```

ie. no `fetcher = fetch` wrapper; just wanting to use our `fetch`.

Related to this; it's unfortunate that the behavior changes from the old version (which called `.json()`), to the new version (which returns the response directly), because that stops the above kind of drop-in substitution from working as-is (you *would* need a wrapper, so we wind up with two wrappers: the wrapper around `fetch` so we can add our standard headers, and the wrapper just to adapt us to `useResource`).

matuzalemsteles: > Which begs the question, why isn't the property just called fetch

Oh sorry, I ended up forgetting. The word `fetcher` seems more friendly to the context in my view because we want the "fetcher" that we will use it at some point and the `fetch` sounds more that we want to get the data now, in the context of `useResource` we just want the fetcher to use it at some point and replace the standard fetcher, does that make sense to you? I think it got more complicated to understand haha ​​😅, I may be fighting with English here, in my understanding:

- `fetch` immediate action
- `fetcher` a noun, which refers to something, the function itself.

Makes sense? Fixes me if I'm wrong 😬

> The trouble is, whatever you use as an example is what people will blindly copy-paste into the code. So, that's why I'm commenting on a useless wrapper in documentation, not because I am pedantic (I am), but because I don't want to see useless wrappers in PRs (which I'll then have to comment on there instead)...

Yeah yeah, sure!! Sorry for that 🙂

> Related to this; it's unfortunate that the behavior changes from the old version (which called .json()), to the new version (which returns the response directly), because that stops the above kind of drop-in substitution from working as-is (you would need a wrapper, so we wind up with two wrappers: our existing wrapper around fetch which adds our standard headers, and the wrapper just to adapt us to useResource).

At first glance I was doing this, I went this way of internally call `.json()` for the dev, regardless of whether he is passing a custom fetch but this can get complicated if the wrapper already has some `then`, probably dev will handle this so it can do any manipulation before it reaches `useResource`, so to avoid conflicts and errors, I let dev handle it. Maybe we can validate if the return is already the given otherwise we can call `.json()` what do you think?

wincent: > Makes sense? Fixes me if I'm wrong 😬

Yeah, I get the logic. I'm just wanting to optimize for the common case, which I *suspect* may be:

```js
useResource({fetcher: fetch, link});
```

At which point it would be easier if you could just write:

```js
useResource({fetch, link});
```

FWIW, I think JS developers are going to be pretty comfortable with functions as values, which means that they can tell the difference between `fetch` (the value) and `fetch()` (the call); they are not going to mistake the name `fetch` for the action-of-calling-`fetch`.

But this is a subtle distinction and obviously not the most important aspect of this API so I don't feel that strongly about it.

> Maybe we can validate if the return is already the given otherwise we can call `.json()` what do you think?

My main concern is that I would want the API to be as consistent and predictable as possible. The presence of the `if` in there means that we're currently doing something different based on something very subtle; eg:

```js
useResource({fetcher: window.fetch, link}) // will return the raw response
useResource({link}) // will return the JSON, even though it's also using window.fetch
```

It feels like a leaky abstraction that I have to know about the internals of how `useResource` works in order to understand that the response will be processed in two different ways depending on whether the *exact same* `fetch` implementation gets used due to being passed explicitly or used as a fallback.

matuzalemsteles: > FWIW, I think JS developers are going to be pretty comfortable with functions as values, which means that they can tell the difference between fetch (the value) and fetch() (the call); they are not going to mistake the name fetch for the action-of-calling-fetch.

Yeah, I think this is a valid argument. I'll go ahead and change the API name to `fetch`.

> My main concern is that I would want the API to be as consistent and predictable as possible. The presence of the if in there means that we're currently doing something different based on something very subtle; eg...

Yes, it makes a lot of sense. Now I'm thinking about how to make this more consistent with behavior without a custom fetcher, apparently, I wouldn't be able to circumvent the use case of having a `then` before, which can call `.json()`, so I'm thinking of ignoring it and be a limitation for that case. What do you think? I will think more about solutions before I go with this.

Writing this now, I wonder if this conditional of checking whether the data is raw or not to call `.json()` would be so bad or inconsistent 🤔.

wincent: > Writing this now, I wonder if this conditional of checking whether the data is raw or not to call `.json()` would be so bad or inconsistent 🤔.

What would that check look like?

matuzalemsteles: What I'm thinking is:
```typescript
(res: Response) => {
	if (res.ok && res.body && !res.body.locked) {
		return res.json();
	}

	return res;
}
```

If the dev returns the raw data, we can check if the body is locked, since `res.json()` can only be called once, when called, the body is locked.

matuzalemsteles:
These final commits will be merged with squash.

packages/clay-data-provider/README.mdx
wincent: Sorry to be so slow to reply to this one, but that seems reasonable to me, because it is consistent with the previous behavior. (On the other hand, if the old version *didn't* call `json()` and we started calling it, then that would feel like a breaking change.)

matuzalemsteles: No problem. Yeah, sure. I'll get on with that.

matuzalemsteles:
hey @wincent updated, let me know if you see anything else before I go with that.

matuzalemsteles:
Going ahead with this, @wincent if you have anything else to add I will send another PR.

PR #3419, opened 2020-07-01
fix(@clayui/date-picker): When selecting a date using space key, removes properly the `:active` state of the element

### LINKED ISSUE #3374: When navigating on DatePicker via TABS and selecting dates with SPACE, we can multi select days
-   [✅ ] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

Just use the master's storybook: https://storybook-clayui.netlify.app/?path=/story/components-claydatepicker--default

Open ClayDatePicker using space, then navigate thru dates using TABS and select using SPACE.

You'll notice is possible select multiple dates(at least the UI tells it 😂 )

Screen Shot 2020-06-29 at 14 58 45

Since We don't have a design specification for multi select dates, don't let user select multiple dates

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.4.0  |
| React | `undefined` |

diegonvs:
## Test Case

1. Open our Date Picker component on storybook: https://storybook.clayui.com/?path=/story/components-claydatepicker--default
2. Navigate thru days using tabs and select a day using SPACE key
3. Navigate faster thru days and try to select multiple days using SPACE key
4. You will able to select multiple days when using SPACE key

----

For solving this problem, the keyDown event when the space key is pressed was prevented and when keyUp, the click callback was called.

## Commits

- chore(@clayui/date-picker): Prefer destructuring
- fix(@clayui/date-picker): When selecting a date using space key, removes properly the `:active` state of the element. This issue can't be reproduced on IE/FF but it can be reproduced on Chrome/Safari. Apparently this is a WebKit issue

Closes #3374

### DIFF
--- a/packages/clay-date-picker/src/DayNumber.tsx
+++ b/packages/clay-date-picker/src/DayNumber.tsx
@@ -29,18 +29,20 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 	onKeyDown = noop,
 	onKeyUp = noop,
 }) => {
+	const {date, outside} = day;
+
 	const classNames = classnames(
 		'date-picker-date date-picker-calendar-item',
 		{
-			active: day.date.toDateString() === daySelected.toDateString(),
-			disabled: day.outside || disabled,
+			active: date.toDateString() === daySelected.toDateString(),
+			disabled: outside || disabled,
 		}
 	);
 
 	return (
 		<button
 			aria-label={formatDate(
-				setDate(day.date, {
+				setDate(date, {
 					hours: 12,
 					milliseconds: 0,
 					minutes: 0,
@@ -49,8 +51,8 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 				'yyyy MM dd'
 			)}
 			className={classNames}
-			disabled={day.outside}
-			onClick={() => onClick(day.date)}
+			disabled={outside}
+			onClick={() => onClick(date)}
 			onKeyDown={(event) => {
 				// When tabbing and selecting a DayNumber using
 				// SPACE key the active state it's not being removed.
@@ -63,14 +65,14 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 			}}
 			onKeyUp={(event) => {
 				if (event.keyCode === KEY_CODE_SPACE) {
-					onClick(day.date);
+					onClick(date);
 				}
 
 				onKeyUp(event);
 			}}
 			type="button"
 		>
-			{day.date.getDate()}
+			{date.getDate()}
 		</button>
 	);
 };
--- a/packages/clay-date-picker/src/DayNumber.tsx
+++ b/packages/clay-date-picker/src/DayNumber.tsx
@@ -13,13 +13,21 @@ interface IProps {
 	daySelected: Date;
 	disabled?: boolean;
 	onClick: (date: Date) => void;
+	onKeyUp?: (event: React.KeyboardEvent) => void;
+	onKeyDown?: (event: React.KeyboardEvent) => void;
 }
+
+const noop = () => {};
+
+const KEY_CODE_SPACE = 32;
 
 const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 	day,
 	daySelected,
 	disabled,
-	onClick,
+	onClick = noop,
+	onKeyDown = noop,
+	onKeyUp = noop,
 }) => {
 	const classNames = classnames(
 		'date-picker-date date-picker-calendar-item',
@@ -43,6 +51,23 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 			className={classNames}
 			disabled={day.outside}
 			onClick={() => onClick(day.date)}
+			onKeyDown={(event) => {
+				// When tabbing and selecting a DayNumber using
+				// SPACE key the active state it's not being removed.
+				// See https://github.com/liferay/clay/issues/3374 for more details.
+				if (event.keyCode === KEY_CODE_SPACE) {
+					event.preventDefault();
+				}
+
+				onKeyDown(event);
+			}}
+			onKeyUp={(event) => {
+				if (event.keyCode === KEY_CODE_SPACE) {
+					onClick(day.date);
+				}
+
+				onKeyUp(event);
+			}}
 			type="button"
 		>
 			{day.date.getDate()}
--- a/packages/clay-date-picker/src/DayNumber.tsx
+++ b/packages/clay-date-picker/src/DayNumber.tsx
@@ -15,10 +15,6 @@ interface IProps {
 	onClick: (date: Date) => void;
 }
 
-const KEYS = {
-	SPACE: ('Spacebar' || ' '),
-};
-
 const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 	day,
 	daySelected,
@@ -53,12 +49,12 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 				// When tabbing and selecting a DayNumber using
 				// SPACE key the active state it's not being removed.
 				// See https://github.com/liferay/clay/issues/3374 for more details.
-				if (event.key === KEYS.SPACE) {
+				if (event.key === 'Spacebar' || event.key === ' ') {
 					event.preventDefault();
 				}
 			}}
 			onKeyUp={(event) => {
-				if (event.key === KEYS.SPACE) {
+				if (event.key === 'Spacebar' || event.key === ' ') {
 					onClick(date);
 				}
 			}}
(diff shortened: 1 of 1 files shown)


packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -41,11 +51,28 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
...
 			className={classNames}
-			disabled={day.outside}
-			onClick={() => onClick(day.date)}
+			disabled={outside}
+			onClick={() => onClick(date)}
+			onKeyDown={(event) => {
+				// When tabbing and selecting a DayNumber using
+				// SPACE key the active state it's not being removed.
+				// See https://github.com/liferay/clay/issues/3374 for more details.
+				if (event.keyCode === KEY_CODE_SPACE) {
```
pat270, line 60: When I was looking up the keycode stuff, I saw a big red alert on MDN I didn't pay attention to. It looks like `event.keyCode` is gone. They recommend using `event.code` (not supported in IE) and `event.key`. I think the right way going forward is:
```
if (event.key === ' ' || event.key === 'Spacebar') {}
```

kresimir-coko:
Hey @diegonvs I've fired up the storybook with your changes and they do have the desired effect, I couldn't select 2 dates at the same time when going fast. LGTM

packages/clay-date-picker/src/DayNumber.tsx
matuzalemsteles: 😱 my +1 for that

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -13,26 +13,36 @@ interface IProps {
...
+const KEY_CODE_SPACE = 32;
+
 const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 	day,
 	daySelected,
 	disabled,
-	onClick,
+	onClick = noop,
+	onKeyDown = noop,
+	onKeyUp = noop,
```
matuzalemsteles, line 30: I think we don't need to keep adding `noop` ​​all the time here, I would avoid that. Since this is a component that is only used in the DatePicker, we don't need noop. We may not even need `onKeyDown` and `onKeyUp` for this case since we are not using it. If this were a public component it would make more sense.

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -41,11 +51,28 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
...
+
+				onKeyDown(event);
+			}}
+			onKeyUp={(event) => {
+				if (event.keyCode === KEY_CODE_SPACE) {
+					onClick(date);
+				}
+
+				onKeyUp(event);
+			}}
```
matuzalemsteles, line 72: One question: this just to retrieve the `space` event that we gave `preventDefault` before?

packages/clay-date-picker/src/DayNumber.tsx
diegonvs: Thanks for pointing up this Pat. I didn't noticed keyCode isn't recommended more. 

Apparently We're using on Clay in these files:

https://gist.github.com/diegonvs/89654431236967272f4ff3bfc43b8311

Ignoring clay-css/jquery related usages, We're using on the Modal, MultiSelect and on our FocusScope component.

I consider replacing this usages with `.key` the best option but I think We must pay attention in 

> `Has non-standard key identifiers and incorrect behaviour with AltGraph.`

from: https://caniuse.com/#feat=keyboardevent-key

I'll fill an issue for replacing this usages and I'll use `.key` on this PR

matuzalemsteles: Yeah, I didn't know that either. This is also used a lot on the Portal. cc @wincent maybe we should create a recommendation in the guidelines to use `key` and `code` instead of `keyCode`?

diegonvs: Bringing eslint-unicorn-plugin cool insights, https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-event-key.md. Can we enforce this on our eslint config?

jbalsas: > Bringing eslint-unicorn-plugin ™️ cool insights, https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-event-key.md.

nogodno

> Can we enforce this on our eslint config?

Probably yes 😉

diegonvs: https://github.com/liferay/clay/issues/3430

packages/clay-date-picker/src/DayNumber.tsx
diegonvs: Yes, makes sense Matu. Since We're not exposing directly this component, I'll remove these noops. :)

packages/clay-date-picker/src/DayNumber.tsx
diegonvs: Yes, when keyDown, We're preventing and when SPACE key is released, onClick cb will be called.

packages/clay-date-picker/src/DayNumber.tsx
diegonvs: > We may not even need onKeyDown and onKeyUp for this case since we are not using it. If this were a public component it would make more sense.

So, Should I remove from props interface?

diegonvs: Removed after some internal conversations

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -41,11 +47,24 @@ const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
...
 			className={classNames}
-			disabled={day.outside}
-			onClick={() => onClick(day.date)}
+			disabled={outside}
+			onClick={() => onClick(date)}
+			onKeyDown={(event) => {
+				// When tabbing and selecting a DayNumber using
+				// SPACE key the active state it's not being removed.
+				// See https://github.com/liferay/clay/issues/3374 for more details.
+				if (event.key === KEYS.SPACE) {
```
matuzalemsteles, line 56: @diegonvs we should probably also check out "Spacebar" here.

> [1] Older browsers may return "Spacebar" instead of " " for the Space Bar key. Firefox did so until version 37, as did Internet Explorer 9, 10, and 11.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

jbalsas: spacebar

https://www.threadless.com/product/2574/spacebar

packages/clay-date-picker/src/DayNumber.tsx
wincent: > Yeah, I didn't know that either. This is also used a lot on the Portal. cc @wincent maybe we should create a recommendation in the guidelines to use `key` and `code` instead of `keyCode`?

Guidelines are cool and all, but I'd much rather have a code abstraction that guides people to use "the right tool". Lint enforcing the use of the tool is a bonus that may be necessary if discoverability of the abstraction isn't good enough. Guidelines are useful support, but let's not kid ourselves that a document in a GitHub repo will stop "Random Product Team X" from shipping something we don't want into "Secret DXP Directory Z".

About the eslint-plugin-unicorn rule, I would steer clear of it. Apart from the undesirability of [the dependency footprint](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6c02ce4e32c84702bcd06e90aa91696366036294/package.json#L35-L49), I don't think it's going to help us avoid the issues with IE compat noted on #3430 (see [the implementation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6c02ce4e32c84702bcd06e90aa91696366036294/rules/prefer-event-key.js#L13-L60), which assumes non-IE browser).

So I'd provisionally say:

scooby-no

And suggest that if we care about linting, we might even want to make our own.

packages/clay-date-picker/src/DayNumber.tsx
diegonvs: See 64af3da

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -15,24 +15,30 @@ interface IProps {
 	onClick: (date: Date) => void;
 }
 
+const KEYS = {
+	SPACE: ('Spacebar' || ' '),
```
matuzalemsteles, line 19: @diegonvs I think that this will not work to check both, we have to make explicit in the `if` conditional below. What we are doing here is conditional if `Spacebar` does not exist which will never be true.

wincent: TS39 proposal: New binary operator, `││` (2 x Unicode U+2502, "Box Drawings Light Vertical"), which represents the quantum superposition of the left and right side operands.

matuzalemsteles: 🤔 Hmm

diegonvs: I was minding https://en.wikipedia.org/wiki/Distributive_property 🤔

PR #3427, opened 2020-07-02
feat(clayui.com): Add JSP code example to Alert

### LINKED ISSUE #3425: Alert should have a JSP code example
Part of #3424. Code example can be found [here](https://issues.liferay.com/browse/LPS-112456).

kresimir-coko:
Fixes #3425 

I used simple values within the JSP code example instead of `Liferay.Language.get(...)`.
I also added `clay:stripe` in here, and made the JSP code example match the 2 alerts in the React example.

### DIFF
--- a/packages/clay-alert/docs/index.js
+++ b/packages/clay-alert/docs/index.js
@@ -27,11 +27,38 @@ render(<Component />);`;
 const alertImportsCode = `import React from 'react';
 import ClayAlert from '@clayui/alert';`;
 
+const alertJSPImportsCode = `<%@ taglib uri="http://liferay.com/tld/clay" prefix="clay" %>`;
+
+const AlertJSPCode = `<clay:alert
+	message="This is a default alert"
+	displayType="info"
+	title="Info"
+/>
+
+<clay:stripe
+	message="This is a stripe alert"
+	displayType="warning"
+	title="Warning"
+/>`;
+
 export const Alert = () => {
 	const scope = {ClayAlert};
-	const code = AlertCode;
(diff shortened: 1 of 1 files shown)


jbalsas:
Looks good @kresimir-coko!!

I just have one doubt with this screenshot:

Screen Shot 2020-07-02 at 15 24 35

I was expecting to see JSP code at the same level as `React Component` and `CSS / Markup`.

I'm not agains it being inside, but then maybe we'd need to rename the `React Component` tab to simply `Component`... I think that makes both the React and JSP version slightly less discoverable, so sorry if this was discussed and I missed it :(

kresimir-coko:
@jbalsas 
> I was expecting to see JSP code at the same level as React Component and CSS / Markup.

Uhm, the whole [tabular code thingy a few months ago](https://github.com/liferay/clay/pull/3280) was basically this, just now I started adding content to the tabs. But yea, renaming the higher level tab from React Component to Component would be an easy fix. I'll open a separate issue for that if @bryceosterhaus @pat270 @matuzalemsteles agree with it, and to give time for them to share their thoughts.

jbalsas:
> I've made drastic changes to my initial approach, here they are:

You ninja-tricked me there and change the approach completely 😱 

ninja

kresimir-coko:
Pulled a sneaky one on you!

jbalsas:
yourefired

wincent:
May as well get this in. We can address structural issues later on; for now the important thing is having the JSP documented *in some form*

PR #3449, opened 2020-07-03
feat(clayui.com): Add JSP code example to Icon

### LINKED ISSUE #3447: Icon should have a JSP code example
Part of #3424. Code example can be found [here](https://issues.liferay.com/browse/LPS-112464).

kresimir-coko:
Fixes #3447.

Apart from the title, I removed an empty line from React code import because it was creating a double empty line, making the spacing between imports and code bigger than it's supposed to be.

image

### DIFF
--- a/packages/clay-icon/docs/index.js
+++ b/packages/clay-icon/docs/index.js
@@ -7,8 +7,7 @@ import Editor from '$clayui.com/src/components/Editor';
 import ClayIcon, {ClayIconSpriteContext} from '@clayui/icon';
 import React from 'react';
 
-const iconImportsCode = `import ClayIcon from '@clayui/icon';
-`;
+const iconImportsCode = `import ClayIcon from '@clayui/icon';`;
 
 const IconCode = `const Component = () => {
 	return (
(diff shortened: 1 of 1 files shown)


packages/clay-icon/docs/index.js
```diff
@@ -18,11 +17,28 @@ const IconCode = `const Component = () => {
 
 render(<Component />)`;
 
+const iconJSPImportsCode = `<%@ taglib uri="http://liferay.com/tld/clay" prefix="clay" %>`;
+
+const IconJSPCode = `<clay:icon symbol="heart" />`;
```
wincent, line 22: I know you've been following a pattern in this and the other examples, but it's non-obvious to me why we've gone with an initial capital letter for this variable. We usually use leading capitals for things like classes (or factory functions), and components etc. But here we're using it for a local variable that holds a string containing a code snippet (note the inconsistency with `iconJSPImportsCode`, which *also* is a local variable holding a string containing a code snippet).

kresimir-coko: 100% it's because that's how other variable names are. It does feel weird having them capitalized, but I went for being consistently inconsistent instead of changing lines that aren't a part of the issue.

PR #3457, opened 2020-07-06
feat(clayui.com): Add JSP code example to Navigation Bar

### LINKED ISSUE #3456: Navigation Bar should have a JSP code example
Part of #3424. Code example can be found [here](https://issues.liferay.com/browse/LPS-112470).

kresimir-coko:
Fixes #3456 

This one is a bit different than the ones before it because it has some JSP specific syntax inside the example, apart from that I did some formatting in the fle too, as it was indented using spaces before.

### DIFF
--- a/packages/clay-navigation-bar/docs/index.js
+++ b/packages/clay-navigation-bar/docs/index.js
@@ -9,28 +9,27 @@ import ClayNavigationBar from '@clayui/navigation-bar';
 import React from 'react';
 
 const navigationBarImportsCode = `import ClayLink from '@clayui/link';
-import ClayNavigationBar from '@clayui/navigation-bar';
-`;
+import ClayNavigationBar from '@clayui/navigation-bar';`;
 
 const NavigationBarCode = `const Component = () => {
 
 	return (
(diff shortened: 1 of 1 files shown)


packages/clay-navigation-bar/docs/index.js
```diff
@@ -39,51 +38,88 @@ render(<Component />);`;
...
+				add(
+					navigationItem -> {
+						navigationItem.setActive(true);
+						navigationItem.setHref(StringPool.BLANK);
+						navigationItem.setLabel("Item 1");
+					}
+				);
+				add(
+					navigationItem -> {
+						navigationItem.setHref(StringPool.BLANK);
```
jbalsas, line 90: We don't need this no-more. We were forced before by `Soy` since it was a mandatory field, but it generates the wrong markup, so better remove the empty `setHref`.

kresimir-coko:
I've pushed the change that removes the obsolete `href` attribute, should be ready for final review now.

bryceosterhaus:
Looks good!

PR #3471, opened 2020-07-07
fix(@clayui/pagination-bar): automatically go to page 1 if active page exceeds delta

### LINKED ISSUE #3453: ClayPaginationBarWithBasicItems: Incorrect results on delta change

https://codesandbox.io/s/condescending-fog-uw6zr?file=/src/App.js

- Create a react component and leverage ClayPaginationBarWithBasicItems from '@clayui/pagination-bar' module.
- When creating a delta array, set one or more labels to be larger than the total items.
- On init, make sure you have more than 1 page.
- Click on any page other than page 1.
- Change delta to be larger than total items.

You'll notice that the results do not display the correct data. ex: Showing 41 to 21 of 21 or Showing 61 to 21 of 21.

Currently, the active page value remains the same even if the value is larger than the total number of pages

- Active page should be temporarily set to the total number of pages.

The reason I say temporarily is not to lose the original active page when the delta changes back to the initial value or a value less than the total number of pages.

bryceosterhaus:
fixes #3453

### DIFF
--- a/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
+++ b/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
@@ -152,6 +152,14 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> = 
 		return item;
 	});
 
+	const totalPages = Math.ceil(totalItems / activeDelta);
+
+	React.useEffect(() => {
+		if (onPageChange && activePage > totalPages) {
+			onPageChange(1);
+		}
+	}, [totalPages]);
+
 	return (
 		<PaginationBar {...otherProps}>
 			<PaginationBar.DropDown
@@ -192,7 +200,7 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> = 
 					}
 				}}
 				spritemap={spritemap}
-				totalPages={Math.ceil(totalItems / activeDelta)}
+				totalPages={totalPages}
 			/>
 		</PaginationBar>
 	);
--- a/packages/clay-pagination-bar/src/__tests__/index.tsx
+++ b/packages/clay-pagination-bar/src/__tests__/index.tsx
@@ -78,4 +78,31 @@ describe('ClayPaginationBar', () => {
 			document.body.querySelector('.dropdown-menu')!.classList
 		).toContain('show');
 	});
+
+	it('automatically goes to page 1 if active page exceeds delta', () => {
+		const Comp = () => {
+			const [activePage, setActivePage] = React.useState(2);
+			const [delta, setDelta] = React.useState(5);
+
+			return (
+				<ClayPaginationBarWithBasicItems
+					activeDelta={delta}
+					activePage={activePage}
+					onDeltaChange={setDelta}
+					onPageChange={setActivePage}
+					spritemap={spritemap}
+					totalItems={15}
+				/>
+			);
+		};
+		const {container, getByTestId} = render(<Comp />);
+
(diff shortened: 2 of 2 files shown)


packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -152,6 +152,14 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> =
 		return item;
 	});
 
+	const totalPages = Math.ceil(totalItems / activeDelta);
+
+	React.useEffect(() => {
+		if (onPageChange && activePage > totalPages) {
+			onPageChange(1);
+		}
+	}, [totalPages]);
```
wincent, line 161: I don't know this component well enough to know if this is The Best Fix™, but based on my superficial understanding it looks like it will fix it.

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -152,6 +152,14 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> =
 		return item;
 	});
 
+	const totalPages = Math.ceil(totalItems / activeDelta);
+
+	React.useEffect(() => {
+		if (onPageChange && activePage > totalPages) {
+			onPageChange(1);
+		}
```
matuzalemsteles, line 160: @bryceosterhaus maybe not better to check this in the `onClick` of the delta above? so we don't need an extra `useEffect`.

bryceosterhaus: I went with the useEffect since its possible that the user could change the total items or active delta via the props and the same bug would result. By using `useEffect`, we cover all those cases as well as onClick.

packages/clay-pagination-bar/src/__tests__/index.tsx
```diff
@@ -78,4 +78,31 @@ describe('ClayPaginationBar', () => {
...
+				/>
+			);
+		};
+		const {container, getByTestId} = render(<Comp />);
+
+		fireEvent.click(getByTestId('selectPaginationBar'), {});
+
+		fireEvent.click(getByText(document.body, '20 items'), {});
+
+		expect(getByText(container, '1').parentElement?.classList).toContain(
```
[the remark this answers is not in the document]
bryceosterhaus: Hmm, that is odd. I was going off of ts complaints which is why I added it. I'll see if I can remove it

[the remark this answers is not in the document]
bryceosterhaus: Types:

`getByText()` returns `HTMLElement ` (ts doesn't account for the throw)

`HTMLElement.parentElement` is `HTMLElement | null`

So it makes sense why TS would complain. Do you think its worth trying to avoid the `?.`?

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
matuzalemsteles: Yeah, great! I had forgotten that.

bryceosterhaus:
Updated the test description but left the `?.` in.

packages/clay-pagination-bar/src/__tests__/index.tsx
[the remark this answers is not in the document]
wincent: > Do you think its worth trying to avoid the `?.`

Nope. My bad. Too much time in liferay-portal that I forgot this was TS.

Even though *we* know it's never going to try accessing undefined, TS doesn't, so it's fair.

PR #3472, opened 2020-07-07
chore(@clayui/css): Use Reuse Spec Snippets for licensing code snippets from Normalize.css, Suitcss, and HTML5 Boilerplate

### LINKED ISSUE #3437: Update Licensing for Normalize and HTML5 Boilerplate used in _reboot.scss and _print.scss
Bootstrap 4 uses:
- Normalize in https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_reboot.scss
- HTML5 Boilerplate in https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_print.scss

We should update our license text to attribute using the SPDX naming scheme.

pat270:
- moves license text to its own file so we don't overwrite commit history every time we release a new version
- removes Clay version from individual svg source files for same reason

fixes #3437

@bryceosterhaus @wincent

### DIFF
--- a/packages/clay-css/gulpfile.js
+++ b/packages/clay-css/gulpfile.js
@@ -205,13 +205,13 @@ gulp.task('version', function() {
 		path.join(scssDir, '_license-text.scss'),
 	];
 
-	function changeVersion(filePath) {
+	function changeVersion(filePath, regex = /\*\sClay\s(.*)\n/g, value = `* Clay ${VERSION}\n`) {
 		fs.readFile(filePath, 'utf8', function(err, data) {
 			if (err) {
 				return console.log(err);
 			}
 
-			var result = data.replace(/\*\sClay\s(.*)\n/g, `* Clay ${VERSION}\n`);
+			var result = data.replace(regex, value);
 
 			fs.writeFile(filePath, result, 'utf8', function(err) {
 				if (err) return console.log(err);
@@ -226,6 +226,23 @@ gulp.task('version', function() {
 			changeVersion(filePath);
 		}
 	});
+
+	var libIconsSvg = path.join('.', 'lib', 'images', 'icons', 'icons.svg');
+
+	var regex = /<\?xml version=\"1.0\" encoding=\"UTF-8\"\?([\s\S]+-->|>)/g;
+	var license = `<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* Clay ${VERSION}
+*
+* SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+* SPDX-FileCopyrightText: © 2020 Contributors to the project Clay <https://github.com/liferay/clay/graphs/contributor
+*
+* SPDX-License-Identifier: BSD-3-Clause
+-->`;
+
+	changeVersion(libIconsSvg, regex, license);
+
+
 });
 
 gulp.task('serve', ['serve:start', 'watch']);
--- a/packages/clay-css/gulpfile.js
+++ b/packages/clay-css/gulpfile.js
@@ -202,10 +202,7 @@ gulp.task('version', function() {
 	var scssDir = path.join('.', 'src', 'scss');
 
 	var clayFiles = [
-		path.join(scssDir, 'atlas-variables.scss'),
-		path.join(scssDir, 'atlas.scss'),
-		path.join(scssDir, 'base-variables.scss'),
-		path.join(scssDir, 'base.scss'),
+		path.join(scssDir, '_license-text.scss'),
 	];
 
 	function changeVersion(filePath) {
--- a/packages/clay-css/gulpfile.js
+++ b/packages/clay-css/gulpfile.js
@@ -199,7 +199,6 @@ gulp.task('version', function() {
 		fs.readFileSync(path.join('.', 'package.json'))
 	).version;
 
-	var svgsDir = path.join('.', 'src', 'images', 'icons');
 	var scssDir = path.join('.', 'src', 'scss');
 
 	var clayFiles = [
@@ -229,12 +228,6 @@ gulp.task('version', function() {
 		if (clayFiles.includes(filePath)) {
 			changeVersion(filePath);
 		}
-	});
-
-	fs.readdirSync(svgsDir).forEach(function(file) {
-		var filePath = path.join(svgsDir, file);
-
-		changeVersion(filePath);
 	});
 });
 
--- a/packages/clay-css/src/scss/_license-text.scss
+++ b/packages/clay-css/src/scss/_license-text.scss
@@ -1,0 +1,17 @@
+/**
+* Clay 3.12.0
+*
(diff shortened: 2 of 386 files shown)


packages/clay-css/src/scss/components/_print.scss
```diff
@@ -1,10 +1,7 @@
...
-
-// ==========================================================================
-// Print styles.
-// Inlined to avoid the additional HTTP request:
-// https://www.phpied.com/delay-loading-your-print-css/
-// ==========================================================================
+/* REUSE-Snippet-Begin
+* SPDX-License-Identifier: MIT
+* SPDX-FileCopyrightText: © 2018 HTML5 Boilerplate <https://github.com/h5bp/main.css>
+*/
```
wincent, line 4: This looks whitespace damaged; would expect:

```
/* REUSE-Snippet-Begin
 * SPDX-License-Identifier: MIT
 * etc
```

packages/clay-css/src/scss/components/_reboot.scss
```diff
@@ -1,9 +1,7 @@
-// Reboot
-//
-// Normalization of HTML elements, manually forked from Normalize.css to remove
-// styles targeting irrelevant browsers while applying new styles.
-//
-// Normalize is licensed MIT. https://github.com/necolas/normalize.css
+/* REUSE-Snippet-Begin
+* SPDX-License-Identifier: MIT
+* SPDX-FileCopyrightText: © 2016 Nicolas Gallagher and Jonathan Neal <https://github.com/necolas/normalize.css>
+*/
```
wincent, line 4: Same here with leading whitespace.

packages/clay-css/src/scss/components/_utilities-functional-important.scss
```diff
@@ -188,8 +188,10 @@
 	}
 }
 
-// Embed
-// Credit: Nicolas Gallagher and SUIT CSS.
+/* REUSE-Snippet-Begin
+* SPDX-License-Identifier: MIT
+* SPDX-Copyright: © 2012 Nicolas Gallagher <https://github.com/suitcss/components-flex-embed>
```
wincent, line 193: And here (leading whitespace).

packages/clay-css/gulpfile.js
```diff
@@ -231,11 +227,22 @@ gulp.task('version', function() {
 		}
 	});
 
-	fs.readdirSync(svgsDir).forEach(function(file) {
-		var filePath = path.join(svgsDir, file);
+	var libIconsSvg = path.join('.', 'lib', 'images', 'icons', 'icons.svg');
+
+	var regex = /<\?xml version=\"1.0\" encoding=\"UTF-8\"\?([\s\S]+-->|>)/g;
```
wincent, line 232: Pro-Tip™: you can use [the `s` flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) to make `.` match everything, including newlines, which means you can get rid of the `[\s\S]` construct:


```js
var regex = /<\?xml version=\"1.0\" encoding=\"UTF-8\"\?(.+-->|>)/gs;
```

Also, you have some excess escaping in there:

```js
var regex = /<\?xml version="1.0" encoding="UTF-8"\?(.+-->|>)/gs;
```

packages/clay-css/src/scss/_license-text.scss
```diff
@@ -0,0 +1,17 @@
...
+*/
+
+/**
+* Bootstrap v4.4.1
+*
+* SPDX-FileCopyrightText: © 2019 Twitter, Inc. <https://twitter.com>
+* SPDX-FileCopyrightText: © 2019 The Bootstrap Authors <https://getbootstrap.com/>
+*
+* SPDX-License-Identifier: LicenseRef-MIT-Bootstrap
+*/
```
wincent, line 17: This all looks whitespace damaged to me:

```scss
/**
 * Clay 3.13.0
 *
```

etc.

packages/clay-css/gulpfile.js
```diff
@@ -199,23 +199,19 @@ gulp.task('version', function() {
...
 	var clayFiles = [
-		path.join(scssDir, 'atlas-variables.scss'),
-		path.join(scssDir, 'atlas.scss'),
-		path.join(scssDir, 'base-variables.scss'),
-		path.join(scssDir, 'base.scss'),
+		path.join(scssDir, '_license-text.scss'),
 	];
 
-	function changeVersion(filePath) {
+	function changeVersion(filePath, regex = /\*\sClay\s(.*)\n/g, value = `* Clay ${VERSION}\n`) {
```
wincent, line 208: Slightly more robust regex would be:

```
/\*\s+Clay\s(.+)\n/g
```

packages/clay-css/src/scss/components/_print.scss
pat270: sent a SF pull at https://github.com/liferay/clay/pull/3485

packages/clay-css/gulpfile.js
pat270: Thanks for the tips! Regex is a mystery to me. I think I understand `\s+` below, but I don't understand the difference between `.*` and `.+`. `.*` matches 0 or more of the preceding token and `.+` matches 1 or more of the preceding token. Where would `.*` match something I wouldn't want in this example vs `.+`?

wincent: Original: `/\*\sClay\s(.*)\n/g`

Would match: `* Clay 4.0.0\n` but also `* Clay \n`

Proposed: `/\*\s+Clay\s(.+)\n/g`

Would only match: `* Clay 4.0.0\n`

In other words, the second one is more robust because it is less likely to run into a false positive if it sees a line like the second one here:

```
 * see also the documentation for
 * Clay 
```

Of course, we're talking about an unlikely edge case, but the art of this is being pragmatic: make stuff as robust as possible *at a reasonable cost*, but no more.

There's [that saying](https://en.wikipedia.org/wiki/Robustness_principle) about "be liberal in what you accept and conservative in what you provide", which, in the world of regexps, translates to something like this:

- Be "loose" so that you don't falsely reject inputs that could be valid. An example of that would be changing the `\s` to `\s*` or `\s+` in `/\*\s\Clay/`; ie. don't reject something on the basis of the amount of whitespace in an insignificant position.
- But be "strict" so that you don't accept obviously incorrect input. An example of that would be making something that matches version numbers instead of just `(.+)`. That could be something as simlpe as `/\d\.d\.\d/` or as complicated as `/(\d{1,2}\.\d{1,2}(?:\.\d{1,2})?(?:-.+)?)/`, or more...

Basically, it's all very subjective, and I think over time you develop a gut instinct for when it's time to take a step in the direction of strictness or of looseness, because you spider sense starts tingling about false positives or false negatives. There's also the trade-off on the axis of readability too (like those version number examples above); here again it's a largely subjective judgement about what is "optimal" — amusingly "optimal" doesn't mean "perfect", but rather "good enough". 😂

Issue #3451, opened 2020-07-03
Improve Documentation Architecture


kresimir-coko:
With the introduction of JSP code examples that will live alongside the React code examples under the React Component tab it doesn't make sense to keep the name of the parent tab `React Component`.

This has been brought up in 2 places so far: [here](https://github.com/liferay/clay/pull/3427#issuecomment-653006762), and [here](https://github.com/liferay/clay/pull/3446#pullrequestreview-442317436)

matuzalemsteles:
I don't see a real problem with putting the same JSP examples on the same page. I'm just concerned that depending on some documents we may have more component-oriented descriptions in React than the JSP, and I don't know if that will help or confuse. In v2 we had the same for SOY, WebComponents, and HTML, and for some people, they missed a lot more information about the components because the document was more the basis of HTML just with the examples in SOY.

I know that most JSP APIs must be consistent, but I think there are some different exceptions or use cases. I think if we can keep this under control it should be a good thing to go ahead with it.

jbalsas:
The other option is to have different tabs as it was in the original proposal... I'd lean towards that option for the reasons @matuzalemsteles mentions and to make them more discoverable... but I missed that conversation, so...

kresimir-coko:
How much documentation relating to JSP do we have? Is it enough to warrant an entire parent tab? From the beginning of the conversation [this type of example](https://github.com/liferay/clay/issues/3229#issuecomment-630487859) was being referenced. Nowhere in the discussion, as far as I understood, has it been mentioned to add a big tab, e.g `React Component | JSP | Markup & CSS`

jbalsas:
> How much documentation relating to JSP do we have?

It's not about how much we have, but about how much we SHOULD have ;)

kresimir-coko:
> how much we SHOULD have

I agree, what kind of information should there be, like an API Table like React? Some tips and tricks on how to use the JSP component? Perhaps the infrastructure needs to be setup before we do that, for example, what is the API Table gonna reference, like the React ones reference the `IProps` of their matching component.

jbalsas:
Yeah, the API could be interesting... that info is in the `.tld`, so we'd need to manually duplicate it, though 🤔

wincent:
> React Component tab should be renamed to Component

I don't think renaming the tab alone is enough. We have a structural problem where the code sample is embedded in a page that is crammed full of React-specific syntax (and in some cases, React-specific props).

[...]

PR #3477, opened 2020-07-08
feat(clayui.com): New documentation architecture

kresimir-coko:
So, here's my first proposal for a new documentation architecture (#3451). It's very barebones at the moment as I'm struggling to get it working, but the changes to the code might speak more than my words do currently.

The idea is for each component's `README.mdx` to have 2 siblings, `css-component.md` and `api-component.md` (or `API.mdx` if it needs to be mdx to support the `APITable`).

### The idea

In practice each component's docs should look something like this:

Overview | Usage | API
------------ | ------------- | -------------
General information about the component with some examples | Some specific examples and explanation detailing how they work | API table for React and JSP

### Current situation

The part where I'm struggling atm is with GraphQL and the way it's being consumed to create the sibling pages. I created a new field called `siblings` that could be the way of specifying 2 siblings, but I still haven't figured out how to consume them, the API of `createPages` and `onCreateNode` is still a bit foreign to me.

I started with a Badge here just to try and showcase an example but it doesn't work yet.


clayui.com/src/templates/docs.js
```diff
@@ -465,5 +492,8 @@ export const pageQuery = graphql`
 		mdxTab: mdx(fields: {slug: {ne: "", eq: $sibling}}) {
 			body
 		}
+		apiTab: mdx(fields: {slug: {ne: "", eq: $sibling}}) {
```
bryceosterhaus, line 495: Should this be `$siblings`? (plural)

packages/clay-badge/README.mdx
```diff
@@ -3,7 +3,7 @@ title: 'Badge'
 description: 'Badges help highlight important information, such as notifications or new and unread messages.'
 lexiconDefinition: 'https://liferay.design/lexicon/core-components/badges/'
 packageNpm: '@clayui/badge'
-sibling: 'docs/components/css-badges.html'
+siblings: ['docs/components/css-badges.html', 'docs/components/api-badges.html']
```
bryceosterhaus, line 6: are arrays permitted in frontmatter? I haven't seen it before, so I wasn't sure.

clayui.com/gatsby/onCreateNode.js
```diff
@@ -220,6 +221,12 @@ module.exports = exports.onCreateNode = ({actions, getNode, node}) => {
 			value: sibling || '',
 		});
 
+		createNodeField({
+			name: 'siblings',
+			node,
+			value: siblings || '',
+		});
```
bryceosterhaus, line 228: Default value should probably be an array since that is what we expect from siblings

clayui.com/src/templates/docs.js
bryceosterhaus: also since siblings is an array, you'll have to loop up the graphql query for that since `eq` expects a string I think

clayui.com/gatsby/createPages.js
```diff
@@ -33,15 +33,27 @@ const createDocs = (actions, edges, mdx, blacklist = []) => {
...
 						node: {
 							fields: {sibling},
 						},
-					}) => sibling === slug
+					}) => {
+						console.log('sibling: ', sibling);
+						console.log('slug: ', slug);
+
+						return sibling === slug;
+					}
```
matuzalemsteles, line 56: hey @kresimir-coko if I'm not mistaken we should treat `siblings` as a blacklist too, then we should add it in https://github.com/liferay/clay/blob/a50d9b724eedfab8fff5585d4a0cf783a7cb8740/clayui.com/gatsby/createPages.js#L183-L192

The logic of sibling is not to create a page for him here, because we only want his content that is taken from the docs template.

clayui.com/src/templates/docs.js
```diff
@@ -465,5 +492,8 @@ export const pageQuery = graphql`
...
+			body
+		}
```
matuzalemsteles, line 497: One thing to keep in mind, this can help you better understand how it works. `createPages.js` calls this template in each document (.md, .mdx) that it is going through and we receive the values ​​that were passed in that line: https://github.com/liferay/clay/blob/a50d9b724eedfab8fff5585d4a0cf783a7cb8740/clayui.com/src/templates/docs.js#L427-L428

Which are the values ​​coming from the call from `createPage`, which is the `context` here: https://github.com/liferay/clay/blob/a50d9b724eedfab8fff5585d4a0cf783a7cb8740/clayui.com/gatsby/createPages.js#L115-L125

Once we have the `siblings` declared we can get all the documents, in this case we need to get it using `allMdx`, I think we can also kill the above query that gets `mdxTab` and stop using `sibling`, we can only use `siblings` and to differentiate which document belongs to each tab, we add a new value in the frontmatter for this something like:
```markdown
---
tabName: 'API'
---
```

Or we can use the `title` for this case too, I think it makes more sense too. If the document becomes a sibling of another.

And we dynamically created the tabs. The query to get the content of the siblings, I think it would look like this (not tested):
```graphql
allMarkdownRemark(filter: {fields: {slug: {in: $siblings}}}) {
	edges {
		node {
			frontmatter {
				tabName
			}
			body
		}
	}
}
```

kresimir-coko:
So here's where I've got so far after spending the day playing around GraphiQL.

Here's the 2 queries I ended up with:
```jsx
{
  {
    APITab: allMdx(filter: {fields: {slug: {ne: "", in: ["docs/components/api-badges.html", "docs/components/css-badges.html"]}}}) {
    edges {
      node {
      	frontmatter {
      	  order
      	  description
      	  disableTOC
      	  redirectFrom
      	  lexiconDefinition
      	  packageNpm
      	  sibling
      	  tabName
      	}  
      }
    }
  }
  overviewTab: allMarkdownRemark(filter: {fields: {slug: {in: ["docs/components/css-badges.html", "packages/clay-badge/api.html"]}}}) {
    edges {
      node {
        frontmatter {
          tabName
        }
        html
      }
    }
  }
}
```

The `APITab` query returns the API tab, and the `overviewTab` query returns the Overview tab (the old Markup/CSS) tab.
I'm not sure if this is how they're supposed to look but that's where I got so far.

clayui.com doesn't get the data for any pages with these queries so my **next step** is to experiment adding `siblings` to frontmatter of all pages to test it out.

Also, I'm getting this error on build time:  `GraphQLError: Cannot query field "siblings" on type "MdxFields". Did you mean "sibling"?` Did I forget to remove `sibling` from somewhere since it's obviously expecting it?

Apart from adding siblings to all other pages I'm unsure how to properly consume all these queries. I need to use these (or similar) queries to render the tabs and display their proper information, perhaps I need to look into more docs about that.

matuzalemsteles:
hey @kresimir-coko sorry for that, I ended up taking it now after last Thursday's holiday and got PTO on Friday.

> clayui.com doesn't get the data for any pages with these queries so my next step is to experiment adding siblings to frontmatter of all pages to test it out.
> Also, I'm getting this error on build time: GraphQLError: Cannot query field "siblings" on type "MdxFields". Did you mean "sibling"? Did I forget to remove sibling from somewhere since it's obviously expecting it?

If this has already been removed, Gatsby has a strong use of caching, so you can try to delete the `.cache` folder on clayui.com and try to run it again. There are some things you need to keep in mind, if I am not mistaken to be a frontmatter or get it through the frontmatter, the siblings needs to be visible in all files, I think, but as we are already creating it in `onCreateNode`, we must deal with it in the `fields`, your queries should look like this:

```diff
- frontmatter {
+ fields {
   siblings
}
```

> For some reason it is adding these empty arrays which might be the reason why I'm getting the error above.

The empty arrays are probably because of the creation of the `siblings` field that its default value is a `[]` (https://github.com/liferay/clay/pull/3477/files#diff-1ed3aafc4aadd6656fe082aa8048f293R220). Strange is the case that sometimes it is coming `undefined`.

> Conclusion is that I'm unsure if I have the required skills to tackle this on my own, I'm getting blocked by my lack of knowledge on every step. There are so many moving parts and as soon as I touch one, something somewhere else complains 😕

Ohh 😕 I'm sorry that this is being a bad experience. I will be more present in this PR to help you guide the next steps. I will leave some comments on the work you have done so far and see what we can go and indicate some articles from Gatsby to complement the reading. Don't be discouraged haha 😁

clayui.com/src/templates/docs.js
```diff
@@ -444,26 +437,15 @@ export const pageQuery = graphql`
...
+					html
 				}
 			}
 		}
-		mdTab: markdownRemark(fields: {slug: {ne: "", eq: $sibling}}) {
-			html
-		}
-		mdxTab: mdx(fields: {slug: {ne: "", eq: $sibling}}) {
-			body
-		}
```
matuzalemsteles, line 467: @kresimir-coko were you able to get some data with this query?

Some things to keep in mind.

- Queries for `.md` files are always for the `allMarkdownRemark` field.
- Queries for `.mdx` files are always for the `allMdx` field

So we must always take the two cases into consideration since we can have some tab in a `.mdx` file, we have to start from the thought that the main file or the main tab will always be the file that contains the `siblings` property and in each interaction of a new page we must search `allMdx` and `allMarkdownRemark` for these files contained in `siblings`.

Always keep in mind that this `docs.js` file will be called whenever you create a new page.

clayui.com/gatsby/createPages.js
```diff
@@ -33,7 +33,7 @@ const createDocs = (actions, edges, mdx, blacklist = []) => {
 		.forEach(
 			({
 				node: {
-					fields: {layout, redirect, redirectFrom, sibling, slug},
+					fields: {layout, redirect, redirectFrom, siblings, slug},
 				},
 			}) => {
 				const hasSibling = blacklist.find(
```
matuzalemsteles, line 39: We should improve this here too, we don't want to create pages for the paths that are in the `siblings`, because we will get your content later in the template.

matuzalemsteles: Maybe this is related to the `siblings` errors being undefined, you can check the result of `siblings` in the call to `createPage` just to make sure that:

- `siblings` will always be an Array
- `siblings` will always be a String Array

clayui.com/src/templates/docs.js
```diff
@@ -444,26 +437,15 @@ export const pageQuery = graphql`
...
-						order
-						indexVisible
-						draft
+					frontmatter {
+						tabName
 					}
+					html
 				}
 			}
 		}
```
matuzalemsteles, line 449: Consider keeping this query as is because we will use it for side navigation. You can check `<Sidebar />` in this file. Try to create a new query for the tabs.

```
tabs: allMarkdownRemark(filter: {fields: {slug: {in: $siblings}}}) {
	edges {
		node {
			fields {
				slug
				tabName
			}
			html
		}
	}
}
```

clayui.com/gatsby/createPages.js
```diff
@@ -95,17 +95,17 @@ const createDocs = (actions, edges, mdx, blacklist = []) => {
...
 
 					createPage({
 						component: template,
 						context: {
 							blacklist: blacklistSlugs,
 							markdownJsx: mdx,
-							sibling,
+							siblings,
 							slug,
 						},
```
matuzalemsteles, line 110: Try to normalize the `blacklistSlugs` variable to be a flat array, it should now return something like this:

```
[
  ['path', 'path/some/'],
  ['path']
]
```

As we are passing this on to the template, we want a `Array<String>`.

kresimir-coko:
Thanks @matuzalemsteles for the encouragement and for the help. I'm again investigating and digging through our queries and code consuming them, have some fresh ideas inspired by your comments so I'll push something tomorrow for you to take a look at. I'm getting a grasp on some aspects of the whole infrastructure, feeling good!

matuzalemsteles:
haha this looks good! 🚀

kresimir-coko:
Hey @matuzalemsteles I think we're getting somewhere 🍾 

I've managed to rename the tabs to `Overview`, `Usage` and `API`, keeping the content of the first 2 tabs as they are on production. The `API` tab is the new one and I managed to display it's temporary content for the purpose of testing.

Currently the API tab is being shown on all pages, even those that don't have a dedicated API markdown file yet (I made it only for Badge for now). So next step is to make sure the API tab shows only on pages that have the API markdown file, and then making sure to display the content of the appropriate file.

I haven't had the time today to attempt this but if you know from a quick glance how to best achieve it feel free to offer your thoughts. The way it's going we might have something representative by the end of tomorrow 😀

bryceosterhaus:
Hey @kresimir-coko can you see why the clayui build is failing and push a fix? It's really important to verify that the netlify builds are successful so that reviewers can verify changes via those builds, otherwise reviewers can only see the code and not the actual UI and interactions of the component or docs.

Screen Shot 2020-07-22 at 10 56 59 AM

clayui.com/src/templates/docs.js
```diff
@@ -459,10 +497,13 @@ export const pageQuery = graphql`
...
+		apiTab: mdx(frontmatter: {tabName: {eq: "API"}}) {
+			body
+		}
+		mdTab: markdownRemark(fields: {slug: {ne: "", in: $siblings}}) {
 			html
 		}
-		mdxTab: mdx(fields: {slug: {ne: "", eq: $sibling}}) {
+		mdxTab: mdx(fields: {slug: {ne: "", in: $siblings}}) {
 			body
 		}
```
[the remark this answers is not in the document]
wincent: GraphQL Nerd Nit: ☝️ those things are called "fields" not "queries". Specifically aliased fields.

A _query_ contains a _selection_ of _fields_, and in the example above `tabsMd` and `tabsMdx` are _aliases_ for the selected fields, `allMarkdownRemark` and `allMdx`).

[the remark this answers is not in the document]
matuzalemsteles: Thanks sir! 🙂

kresimir-coko:
Hey @matuzalemsteles I've pushed some more changes, these ones actually work! ✨ 

The API tab only gets rendered on the Badge page, where it's intended to, and no tabs get rendered in some pages where they aren't supposed to (like the parent Components page).

One thing of note is plenty of conditional chaining in effect due to Gatsby complaining on runtime and when running `yarn build`.

There are some errors whenever I'm making changes, or when running `yarn build`, which can be seen on the Production site  in the console too /cc @bryceosterhaus along the lines of: `TypeError: Cannot read property 'match' of undefined`.

I think next steps are basically improving some of the code I wrote, optimising it better, and making a representable API page, and maybe shuffling some Badges content around to better showcase the intent of this restructuring.

matuzalemsteles:
Hey @kresimir-coko, this looks good! I didn't have time to look at your recent changes today, but tomorrow morning I will take a look.

clayui.com/gatsby/createPages.js
```diff
@@ -95,17 +95,20 @@ const createDocs = (actions, edges, mdx, blacklist = []) => {
...
 						component: template,
 						context: {
-							blacklist: blacklistSlugs,
+							blacklist: blacklistSlugs.reduce(
+								(acc, val) => acc.concat(val),
+								[]
+							),
 							markdownJsx: mdx,
-							sibling,
+							siblings,
```
matuzalemsteles, line 111: It seems that this is coming `undefined` in some moments and we need it to be at least an empty array, apparently, it is still creating pages that are ignored for being siblings of other pages, you have to improve the `find` of `hasSibling`:

```js
const hasSibling = blacklist.find(
	({
		node: {
			fields: {siblings},
		},
	}) => siblings.includes(slug)
);
```

this seems to be the problem with errors on the console.

clayui.com/src/templates/docs.js
```diff
@@ -153,26 +154,39 @@ export default (props) => {
...
+
+	const tabs = [];
+
+	if (mdTabs?.edges.length) {
+		tabs.push(mdTabs.edges[0]?.node);
+	}
+
+	if (mdxTabs?.edges.length) {
+		tabs.push(mdxTabs.edges[0]?.node);
+	}
```
matuzalemsteles, line 180: I thought it could be done this way:

```
const tabs = mdTabs
	.concat(mdxTabs)
	.map(({node: {frontmatter: {tabName}, html, body}}) => ({tabName, html: html ?? body}));
```

and below we can use `map` to render the tabs.

clayui.com/gatsby/createPages.js
kresimir-coko: Oh, that's a good catch @matuzalemsteles , it's been a couple weeks, but I think that at the time I focused too much on the queries and forgot about this file.

clayui.com/src/templates/docs.js
kresimir-coko: I thought it could be too, but there are a couple of reasons why I didn't go in this direction:
1. At most we have 3 tabs, which is good enough for a `map` but the problem is that the 1st tab, `Overview`,  isn't in the tab list that we get from GraphQL, meaning it has to be manually added before the `map`.
2. The `TabContent` low-level component accepts either `html` or `jsx` prop (it can accept both, but with a caveat), which is a problem: the `Usages` tab works fine, but the `API` tab prioritizes the `html` prop over the `jsx` one, meaning it dangerously sets HTML, as per the instructions within `TabContent`

P.S. I've tested this now, to make sure I'm not going from memory

kresimir-coko:
Hey @matuzalemsteles I've improved the `hasSiblings` and answered your suggestion on mapping the tabs. There's still a problem when running `yarn build` where GraphQL complains that `mdTabs` & `mdxTabs` are undefined, this I think, is due to them not existing unless one of the component pages is active. `yarn ci` passes with issues only related to `DataProvider`, unrelated to this PR.

clayui.com/src/templates/docs.js
matuzalemsteles: Hmm yeah, makes sense. No problem, so we can continue with this and we can review it later if necessary

matuzalemsteles:
> Hey @matuzalemsteles I've improved the hasSiblings and answered your suggestion on mapping the tabs. There's still a problem when running yarn build where GraphQL complains that mdTabs & mdxTabs are undefined, this I think, is due to them not existing unless one of the component pages is active. yarn ci passes with issues only related to DataProvider, unrelated to this PR.

@kresimir-coko I think it could be that the `siblings` is coming undefined according to the log, for `/docs/css/index.html`, that could be the problem. Try to investigate if the field `siblings` come undefined it must have come from at least one empty array, try to test locally using the `yarn build` this can help to debug so you don't need to send it to the PR directly.

bryceosterhaus:
@kresimir-coko you can run the ci build locally by running `yarn build` from the `clayui.com` directory. Thats the best way to debug failing deploy previews

kresimir-coko:
Alrighty @matuzalemsteles I've fixed it! I added a default value to the `siblings` argument inside `createPages`.

A reminder that this PR was supposed to be just an example of how we can modify the documentation to present the information in a better, clearer way. **That's why I'll leave it as a Draft PR for now.**

@bryceosterhaus @pat270 @wincent @jbalsas @dsanz Feel free to take a look at the deploy preview and share your thoughts. There are 2 examples utilizing this new architecture, Badge and Autocomplete

kresimir-coko:
Hey @jbalsas, @dsanz can you check the deploy preview and let me know your thoughts on this switch in the documentation layout

jbalsas:
> Hey @jbalsas, @dsanz can you check the deploy preview and let me know your thoughts on this switch in the documentation layout

Looks quite good to me @kresimir-coko!! Let's wait for @dsanz to come back from PTO and gather some more feedback, but I like where this is going! 

I feel there's still too much info in the `Overview` section... maybe we could split API into another category so it doesn't bloat that... 🤔

kresimir-coko:
@jbalsas 
> maybe we could split API into another category

If you check out the [Badge](https://deploy-preview-3477--next-clayui.netlify.app/docs/components/badge.html) and [Autocomplete](https://deploy-preview-3477--next-clayui.netlify.app/docs/components/autocomplete.html) components they have the API in it's own tab.

Apart from moving the API to it's own tab I didn't mess with content much, just renamed the tabs, and added the API tab in those 2 examples.

jbalsas:
> If you check out the [Badge](https://deploy-preview-3477--next-clayui.netlify.app/docs/components/badge.html) and [Autocomplete](https://deploy-preview-3477--next-clayui.netlify.app/docs/components/autocomplete.html) components they have the API in it's own tab.

You're right! Since I saw the tabs in other components I assumed you'd changed them as well and missed those.

LGTM!

kresimir-coko:
I'll work on applying this new architecture for all our components, so that we can get it that much closer to merging.

kresimir-coko:
I've added a commit with a few more examples of the new architecture, so feel free to check them out, they're Alert, Breadcrumb, Button Group, Button and Card

bryceosterhaus:
One thing I noticed that seems a little weird is that we renamed the tabs but the content is still pretty much the same. `Overview` seems to always point to react specific docs and `Usages` points to css specific examples.

I'm not sure these line up with the initial desire to have


> Overview | Usage
> -- | --
> General information ... | Some specific examples ...

For someone who re-occuringly comes to consult the docs, I think the primary focus is to find API and examples up front.

@jbalsas, from talking with other teams and people using Clay, do you know if this structure would better serve their purpose?

bryceosterhaus:
Food for thought, maybe we might benefit from removing a lot of the descriptive text and have tabs of `Examples`, `Markup`, `API`. We could defer most of the descriptive content to the lexicon docs. Really the only thing we need to describe is what a certain prop or config does, not necessarily the "why" of the component(more lexicon definitions).

- `Examples`: Visual examples that include snippets for both react, jsp (similar to https://material-ui.com/components/progress/#circular where it just throws you right into a bunch of examples)
- `Markup`: css docs and explanation
- `API`: Table for props and explanation of prop

I really like how you separated the API table into its own tab, I think thats really helpful.

kresimir-coko:
@bryceosterhaus
> One thing I noticed that seems a little weird is that we renamed the tabs but the content is still pretty much the same.

Yes, that was the point initially at least, to not spend too much time reorganizing specific bits and pieces but mostly to figure out how to bootstrap Gatbsy to work with 3 tabs.

> We could defer most of the descriptive content to the lexicon docs.

I agree, the reason why I didn't go all-in with reorganizing all of the docs is that we haven't reached a consensus on IF and HOW we are going to be going about this whole thing.

That's why, for now, all I'm doing is basically adding the API tab (and moving API to it) to the components as I go through them, just so I don't over-commit and then we decide to ditch most of the text, or decide that all examples go under one tab, etc. like you suggested here. 

I did ping people in the comment above but not all of them have looked through it, so I think we're mainly missing approval from people to get 100% confirmation of the direction. In the meantime I'll be adding the API tab to components that miss it because that's something that's gonna be 100% beneficial.

@matuzalemsteles, @pat270, @dsanz, @wincent, @jbalsas what do you guys think about the architecture currently in play vs the one @bryceosterhaus suggested in the comment above?

### Current:
Overview | Usage | API
-- | -- | --
General information about the component with some examples | Some specific examples and explanation detailing how they work | API table for React and JSP

vs

### Proposed:
Examples | Markup | API
-- | -- | --
Visual examples that include snippets for both react & jsp | css docs and explanation | Table for props and explanation of prop

matuzalemsteles:
I'm a little inclined towards @bryceosterhaus' ideas, mainly because I prefer that we have a specific session only for markups/css so that things don't get confused or we lose information.

One of the main complaints I heard when we had the v2 version documentation is that we had confusing documentation, using the same session to put markup/css information, examples in Metal+Soy and Web Components.

So I like to keep the markup/css session. I also like to have the Examples section only, because for some components we don't need to clarify a lot of things, if necessary we can add it if necessary.

kresimir-coko:
As I'm adding these API tabs, I've encountered following problems:
- `TypeError: Cannot read property 'allMarkdownRemark' of null` on component pages that only have one tab (case might be more specific but I didn't get into that yet)
- The `tabName` frontmatter property isn't being used yet in `docs.js`, I tried using it shortly but it produced a bug which I'll sort when I transfer all the examples

So I sent the updates for all of our components in the latest commit, and squashed the old commits to clean it up a bit. On monday I'll be working on fixing the 2 bugs listed above.

kresimir-coko:
I've sent a few more commits:
- Added `tabName=Markup` to `css-*.md` files,
- Removed no longer necessary tabName from one of the files
- Changed extension of `css-icon.md` to `css-icon.mdx`to make sure that the `flags` is getting defined
- Adapted `docs.js` to work with the new architecture better

Primarily the `docs.js` should be the one to review at this stage, as it's the one containing changes that influence how the site works

kresimir-coko:
#3654 caused me to completely redo the PR, making it incredibly simple in complexity but have the same volume of files changed

So here's the updated notes, gonna write them as if I'm sending it for the first time:
1. Added `api-*.mdx` files for each of the components
2. Utilized the new API by adding `mainTabURL` attribute to the `api-*.mdx` files
3. I didn't add `siblings` attribute because that's essentially replaced `mainTabURL`
4. Renamed some `css-*.md` files, making them use singular variant of the noun, instead of plural, eg. `icon` vs `icons`
5. Moved API Tables and related markup (navigation) to their respective `api-*.mdx` files

One new thing that I haven't done before, and therefore might be wrong, is using `.reverse()` in `...tabs.reverse().map(...)` to get the correct order of tabs `Overview | Markup | API` instead of the default `Overview | API | Markup`. I couldn't figure out how to change the order before `tabs` get consumed at that point, so I went with this. I researched a bit making sure that I don't do something obviously bad, mainly didn't want to mutate the array, but I think that the spread operator `...tabs` copies the original `tabs` array for us to safely `reverse()`.

clayui.com/gatsby/createPages.js
```diff
@@ -39,9 +40,9 @@ const getTabs = (permalink, pathGroup) => {
 	return [
 		{
 			href: permalink,
-			name: 'React Component',
+			name: 'Overview',
 		},
-		...tabs.map(({node: {fields: {slug}}}) => ({
+		...tabs.reverse().map(({node: {fields: {slug}}}) => ({
```
matuzalemsteles, line 45: @kresimir-coko you can use `.sort` with specific rules, it seems more secure but I don’t care about keeping `reverse`, I think the only problem with this is that we rely on the reverse order of the elements and the loading of the files and GraphQL.

matuzalemsteles (commented):
This is great, it got simpler, it looks much better now!

clayui.com/gatsby/createPages.js
kresimir-coko: I'm not sure how I can properly debug this, as the only output I'm getting is in the server log and that's outputting GraphQL values.

When I do something like this:
```js
const sortTabs = (a, b) => {
	console.log('a: ', a);

	if (a.name > b.name) {
		return 1;
	}

	if (a.name > b.name) {
		return -1;
	}
};
```

This is the output I get:
```bash
a:  [Object: null prototype] {
  node:
   [Object: null prototype] {
     fields:
      [Object: null prototype] {
        layout: '',
        nightly: false,
        redirect: '',
        redirectFrom: '',
        slug: 'docs/components/modal/api.html',
        mainTabURL: 'docs/components/modal.html' },
     body: ...
   } 
}
```

Am I missing something when debugging inside `createPages.js`? @matuzalemsteles

wincent:
```js
	if (a.name > b.name) {
		return 1;
	}

	if (a.name > b.name) {
		return -1;
	}
```

Note the repeated `>` operators.

Should be one `>` and one `<`, and then a `return 0` for good measure at the bottom.

kresimir-coko: True true, that was a typo while trying to get proper output, just a placeholder, but thanks for noticing!

matuzalemsteles: @kresimir-coko to avoid this, try to `sort` after `map`, it will be easier to deal with.

kresimir-coko:
@matuzalemsteles I've pushed the commit that sorts tabs instead of reversing them!

kresimir-coko:
@matuzalemsteles I've pushed 2 small commits that get this closer (or even ready) to merge. The commits consist of:

- Squashed all of the old commits to declutter
- Renaming `css-*` files to `markup-*` so they match the new tab name Markup
- Change the first tab name from `Overview` to `Examples`, and update the tab object key to be `markup` instead of `css`

I think that this is ready to merge, the only thing missing is actually moving content around so that only examples are under the Examples tab, but I think that should be a separate issue to not bloat and extend this one any more than we already did.

matuzalemsteles (approved):
@kresimir-coko sounds good! LGTM cc @bryceosterhaus 

> I think that this is ready to merge, the only thing missing is actually moving content around so that only examples are under the Examples tab, but I think that should be a separate issue to not bloat and extend this one any more than we already did.

Yeah, we can move this little by little.

bryceosterhaus (commented):
I think this is good to merge as well. Great work on this @kresimir-coko!

One improvement I think we might want to make is to continue co-locating the docs with the components. I noticed we moved the API docs to the `clayui.com` folder, but I think it might be better co-locate it like `clay-badge/docs/api.mdx`. This can be done later though.

kresimir-coko:
> One improvement I think we might want to make is to continue co-locating the docs with the components. 

I agree @bryceosterhaus , I was thinking about this when renaming files. I assume in order to properly move files we would need to adjust `createPages.js` to properly read them?

bryceosterhaus:
@kresimir-coko yeah we would need to update `createPages.js`. I can look into that today and see what it might require

Issue #3282, opened 2020-05-26
Update Bootstrap 4 License using SPDX naming scheme


pat270:
https://github.com/liferay/clay/pull/3277#issuecomment-633474606
https://issues.liferay.com/browse/FOSS-77

PR #3480, opened 2020-07-08
chore(@clayui/css): Copy Licenses from root to clay-css/LICENSES

pat270:
issue #3282 

Hey @bryceosterhaus, I ended up adding a gulp task (`gulp copy:licenses`) that runs when `gulp compile` is run. It copies the licenses used by Clay CSS into `clay-css/LICENSES`. We should include this directory when we publish. I can add a commit to do that, I just don't know where.

I had a hard time trying to figure out `ncp dir1 dir2 --filter=""`.  My conclusion is that it's bugged.

### DIFF
--- a/packages/clay-css/gulpfile.js
+++ b/packages/clay-css/gulpfile.js
@@ -196,7 +196,7 @@ gulp.task(
 );
 
 gulp.task('copy:licenses', function() {
-	var newLicenseDir = path.join('.', 'LICENSES');
+	var newLicenseDir = 'LICENSES';
 	var originalLicenseDir = path.join('..', '..', 'LICENSES');
 
 	return gulp.src([
--- a/packages/clay-css/gulpfile.js
+++ b/packages/clay-css/gulpfile.js
@@ -52,6 +52,7 @@ gulp.task(
 			'compile:css',
 			'compile:clean-scss',
 			'compile:svg',
+			'copy:licenses',
 			function(err) {
 				cb(err);
 			}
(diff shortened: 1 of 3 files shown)


packages/clay-css/gulpfile.js
```diff
@@ -194,6 +195,19 @@ gulp.task(
 	}
 );
 
+gulp.task('copy:licenses', function() {
+	var newLicenseDir = path.join('.', 'LICENSES');
```
wincent, line 199: Wouldn't this work too?

```js
var newLicenseDir = 'LICENSES';
```

Because for most intents and purposes, `LICENSES` and `./LICENSES` is the same thing.

packages/clay-css/gulpfile.js
```diff
@@ -194,6 +195,19 @@ gulp.task(
...
+	var originalLicenseDir = path.join('..', '..', 'LICENSES');
+
+	return gulp.src([
+		path.join(originalLicenseDir, 'BSD-3-Clause.txt'),
+		path.join(originalLicenseDir, 'LicenseRef-MIT-Bootstrap.txt'),
+		path.join(originalLicenseDir, 'MIT.txt')
```
wincent, line 205: Would a glob work here? If so, it would be more future-proof (ie. if we ever add/remove licenses):

```js
path.join(originalLicenseDir, '*.txt')
```

pat270: I was also trying to future-proof since all of our licenses for Clay would live in this directory. I didn't want to include licenses for React or other module we might be using in our React components.

packages/clay-css/gulpfile.js
pat270: Absolutely, this makes total sense.

bryceosterhaus:
@pat270 I think this is good, I ran `yarn pack` from clay-css to see what the package would look like and I got 

Screen Shot 2020-07-09 at 10 53 56 AM


That looks correct, right?

pat270:
@bryceosterhaus yep that looks correct

bryceosterhaus:
@pat270 this looks good to me then!

PR #3485, opened 2020-07-09
fix(@clayui/css): Adds leading whitespace for License texts

pat270:
fix(@clayui/css): `gulp version` use the s flag to make `.` match everything including newlines instead of `[\s\S]`, remove excess escaping of double quote, and use more robust `+` instead of `*`

issue #3437 

@wincent


jbalsas:
> Thanks for the polish!

polish

PR #3487, opened 2020-07-09
feat(@clayui/form): Adds `leftMaxItems` and `rightMaxItems` properties for providing the user to limit the number of items that each select box can store

diegonvs:
…s for providing the user to limit the number of items that each select box can store

### Test Case:

1. Open ClayDualListbox component on the deploy Preview of our storybook of this PR
2. On story, the limit for left list is 3 and for the right is 5
3. Use the knob for modifying the values and test

This PR is a part of https://issues.liferay.com/browse/LPS-116801 just for studying if We can have maxItems here :)

### DIFF
--- a/packages/clay-form/src/DualListBox.tsx
+++ b/packages/clay-form/src/DualListBox.tsx
@@ -61,19 +61,19 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	};
 
 	/**
+	 * Disables the button responsible for moving items from right to left box.
+	 */
+	disableRTL: boolean;
+
+	/**
+	 * Disables the button responsible for moving items from left to right box.
+	 */
+	disableLTR: boolean;
+
+	/**
 	 * Items spread across two arrays that will be displayed in the two Select Boxes.
 	 */
 	items: TItems;
-
-	/**
-	 * The maximum number of items for left Select Box.
-	 */
-	leftMaxItems?: number;
-
-	/**
-	 * The maximum number of items for left Select Box.
-	 */
-	rightMaxItems?: number;
 
 	/**
 	 * Handler that triggers when the content of the items prop are changed. Caused by either reordering or transfering of items.
@@ -107,10 +107,10 @@ const ClayDualListBox: React.FunctionComponent<IProps> = ({
 		transferRTL: 'Transfer Item Right to Left',
 	},
 	className,
+	disableLTR,
+	disableRTL,
 	items,
 	left = {},
-	leftMaxItems = Infinity,
-	rightMaxItems = Infinity,
 	onItemsChange,
 	right = {},
 	size,
@@ -161,10 +161,7 @@ const ClayDualListBox: React.FunctionComponent<IProps> = ({
 						aria-label={ariaLabels.transferLTR}
 						className="transfer-button-ltr"
 						data-testid="ltr"
-						disabled={
-							leftSelected.length === 0 ||
-							rightItems.length >= rightMaxItems
-						}
+						disabled={leftSelected.length === 0 || disableLTR}
 						displayType="secondary"
 						onClick={() => {
 							const [arrayLeft, arrayRight] = swapArrayItems(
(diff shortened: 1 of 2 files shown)


packages/clay-form/src/DualListBox.tsx
```diff
@@ -65,6 +65,16 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	 */
 	items: TItems;
 
+	/**
+	 * The maximum number of items for left Select Box.
+	 */
+	leftMaxItems?: number;
```
pat270, line 71: Instead of using `left` and `right` can we use something friendlier for rtl devs like `first`/`last` or `start`/`end`?

bryceosterhaus:
Something like `disableRTL` `disableLTR`

packages/clay-form/src/DualListBox.tsx
diegonvs: Makes sense Pat! I'll change.

diegonvs:
> Rather than internally check for this, what if we just provided an api to disable the left/right buttons? This way the user can disable it when a certain number of items have been reached and also provides a greater ability for other types of interactions for disabling.

Makes sense, We can empower the developer when needing to provide more complex rules for disabling or not these buttons. Thanks Bryce! :)

diegonvs:
@bryceosterhaus, @pat270 

Considering last @bryceosterhaus comment,

What about adding a `button` suffix on `disableRTL/LTR` for preventing the user think about html dirs?
It will be like `disableRTLButton` and `disableLTRButton`.

Considering Pat's comment about avoiding `left` and `right` words and use `first/last` or `start/end` on properties it can be `disableMoveToLast` and `disableMoveToFirst`. What do you think about it?

kresimir-coko:
Hey @diegonvs this looks great!

> Considering Pat's comment about avoiding `left` and `right` words and use `first/last` or `start/end` on properties it can be `disableMoveToLast` and `disableMoveToFirst`. What do you think about it?

This sounds good to me! Doesn't need to have a `button` suffix because it has the `move` word in there, it's LTR friendly, yea, to me it sounds great!

packages/clay-form/src/DualListBox.tsx
wincent: It would have been nice to establish a pattern like this when the component was first made, but not now.

Like [I said](https://github.com/liferay-frontend/liferay-portal/pull/98#pullrequestreview-446252517) on the related DXP [PR](https://github.com/liferay-frontend/liferay-portal/pull/98), the component already has "left"/"right" terminology, so introducing a *mix* of "left"/"right" terminology with "first"/"last" or "start"/"end" terminology would cause confusion (and if we want to change to all, to be consistent, we have an undesirable breaking change).

bryceosterhaus:
@diegonvs since we already used the `ltr` and `rtl` language, I think its fine for us to use `disableRTL` and `disableLTR`

diegonvs:
Ok @bryceosterhaus, I'll update the PR :)

packages/clay-form/src/DualListBox.tsx
```diff
@@ -60,6 +60,16 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 		transferRTL: string;
 	};
 
+	/**
+	 * Disables the button responsible for moving items from right to left box.
+	 */
+	disableRTL: boolean;
```
bryceosterhaus, line 66: These should be optional params `disableRTL?:`

packages/clay-form/src/DualListBox.tsx
```diff
@@ -60,6 +60,16 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
...
+
+	/**
+	 * Disables the button responsible for moving items from left to right box.
+	 */
+	disableLTR: boolean;
```
bryceosterhaus, line 71: same here

diegonvs: Sure. I completely forgot. I'll update

packages/clay-form/src/DualListBox.tsx
diegonvs: https://github.com/liferay/clay/pull/3487#discussion_r453835692

PR #3356, opened 2020-06-23
feat(@clayui/css): Drilldown should work with Dropdown Menu, have dual click areas inside `dropdown-item`, and have `min-height`

pat270:
issue #3337 
@kresimir-coko You can view the markup and the pattern I used under Dropdowns Drilldown on the test site.

PR #3489, opened 2020-07-09
feat(@clayui/drilldown): high-level component

bryceosterhaus:
Finally got the animations working for this. I had to make sure to add the transition class after applying the transform.

Anyways, here is a rough idea of an initial high-level component for drilldown. This would likely be for dropdown only right now. I think before we break this into smaller low-level components, we need an idea of where people will use this.

Do we have any sort of docs outlined for this? I wasn't able to find them. cc @jbalsas

### DIFF
--- a/packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
+++ b/packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
@@ -1,0 +1,24 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ClayDropDownWithDrilldown renders 1`] = `
+Document {
+  "location": Location {
+    "assign": [Function],
+    "hash": "",
+    "host": "localhost",
+    "hostname": "localhost",
+    "href": "http://localhost/",
+    "origin": "http://localhost",
+    "pathname": "/",
+    "port": "",
+    "protocol": "http:",
+    "reload": [Function],
+    "replace": [Function],
+    "search": "",
+    "toString": [Function],
+  },
+  Symbol(SameObject caches): Object {
+    "styleSheets": StyleSheetList {},
+  },
+}
+`;
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
 		"@types/react": "^16.8.0",
 		"@types/react-dom": "^16.8.0",
 		"@types/react-test-renderer": "^16.8.0",
+		"@types/react-transition-group": "^4.4.0",
 		"@types/storybook__addon-a11y": "5.1.2",
 		"@types/storybook__addon-knobs": "^5.2.1",
 		"@types/storybook__react": "^5.2.1",
--- a/packages/clay-css/src/scss/components/_drilldown.scss
+++ b/packages/clay-css/src/scss/components/_drilldown.scss
@@ -25,6 +25,15 @@
 	height: auto;
 }
 
+.drilldown-next-active,
+.drilldown-prev-initial {
+	transform: translateX(-100%);
+}
+
+.drilldown-prev-active {
+	transform: translateX(0%);
+}
+
 // Drilldown Dropdown Menu
 
 .drilldown.dropdown-menu {
--- a/packages/clay-drop-down/package.json
+++ b/packages/clay-drop-down/package.json
@@ -33,6 +33,7 @@
 		"@clayui/shared": "^3.2.1",
 		"classnames": "^2.2.6",
 		"dom-align": "^1.10.2",
+		"react-transition-group": "^4.4.1",
 		"warning": "^4.0.3"
 	},
 	"peerDependencies": {
--- a/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
+++ b/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
@@ -1,0 +1,146 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+import ClayDropDown from './DropDown';
+import ClayMenu from './Menu';
+import Drilldown from './drilldown';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Default position of menu element. Values come from `./Menu`.
+	 */
+	alignmentPosition?: React.ComponentProps<
+		typeof ClayMenu
+	>['alignmentPosition'];
+
+	/**
+	 * HTML element tag that the container should render.
+	 */
+	containerElement?: React.ComponentProps<
+		typeof ClayDropDown
+	>['containerElement'];
+
+	/**
+	 * The unique identifier of the menu that should be active on mount.
+	 */
+	initialActiveMenu: string;
+
+	/**
+	 * Prop to pass DOM element attributes to <DropDown.Menu />.
+	 */
+	menuElementAttrs?: React.ComponentProps<
+		typeof ClayDropDown
+	>['menuElementAttrs'];
+
+	/**
+	 * Map of menus and items to be used in the drilldown. Each key should be a unique identifier for the menu.
+	 */
+	menus: {
+		[id: string]: React.ComponentProps<typeof Drilldown.Menu>['items'];
+	};
+
+	/**
+	 * Function for setting the offset of the menu from the trigger.
+	 */
+	offsetFn?: React.ComponentProps<typeof ClayDropDown>['offsetFn'];
+
+	/**
+	 * Path to spritemap
+	 */
+	spritemap?: string;
+
+	/**
+	 * Element that is used as the trigger which will activate the dropdown on click.
+	 */
+	trigger: React.ReactElement;
+}
+
+interface IHistory {
+	id: string;
+	title: string;
+}
+
+export const ClayDropDownWithDrilldown: React.FunctionComponent<IProps> = ({
+	alignmentPosition,
+	className,
+	containerElement,
+	initialActiveMenu,
+	menuElementAttrs,
+	menus,
+	offsetFn,
+	spritemap,
+	trigger,
+}: IProps) => {
+	const [active, setActive] = React.useState(false);
+	const [activeMenu, setActiveMenu] = React.useState(initialActiveMenu);
+	const [direction, setDirection] = React.useState<'prev' | 'next'>();
+	const [history, setHistory] = React.useState<Array<IHistory>>([]);
+
+	const menuIds = Object.keys(menus);
+
+	return (
+		<ClayDropDown
+			active={active}
+			alignmentPosition={alignmentPosition}
+			className={className}
+			containerElement={containerElement}
+			hasRightSymbols
+			menuElementAttrs={{
+				...menuElementAttrs,
+				className: classNames(menuElementAttrs?.className, 'drilldown'),
+			}}
+			offsetFn={offsetFn}
+			onActiveChange={setActive}
+			trigger={trigger}
+		>
+			<Drilldown.Inner>
+				{menuIds.map((menuKey) => {
+					return (
+						<Drilldown.Menu
+							active={activeMenu === menuKey}
+							direction={direction}
+							header={
+								activeMenu === menuKey && history.length
+									? history.slice(-1)[0].title
+									: undefined
+							}
+							items={menus[menuKey]}
+							key={menuKey}
+							onBack={() => {
+								const [parent] = history.slice(-1);
+
+								setHistory(
+									history.slice(0, history.length - 1)
+								);
+
+								setDirection('prev');
+
+								setActiveMenu(parent.id);
+							}}
+							onForward={(title, childId) => {
+								setHistory([
+									...history,
+									{id: activeMenu, title},
+								]);
+
+								setDirection('next');
+
+								setActiveMenu(childId);
+							}}
+							spritemap={spritemap}
+						/>
+					);
+				})}
+			</Drilldown.Inner>
+		</ClayDropDown>
+	);
+};
+
+ClayDropDownWithDrilldown.displayName = 'ClayDropDownWithDrilldown';
+
+export default ClayDropDownWithDrilldown;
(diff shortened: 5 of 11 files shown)


matuzalemsteles:
hey @bryceosterhaus I can only think of this document https://docs.google.com/document/d/1PVQwknqVIOQhf2JL0kEv0k6WHOExKOskiX-CxxWYRqs, see the Variants.

About the animations: I'm thinking if we should start having animation only in JS, it seems "easier" to deal with.

packages/clay-drilldown/stories/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+				items: [
+					{href: '#', title: 'Four'},
+					{
+						items: [{title: 'Six'}, {title: 'Seven'}],
+						title: 'Five',
+					},
+				],
+				title: 'Three',
+			},
+		]}
```
matuzalemsteles, line 35: I like this style of API, but I am concerned with the depth of the API and if it could be bad for use in the case of DXP, for example Some of the data is handled in the backend and then sent to the component.

Thinking if we can only have a flat list of items, something like this:

```js
const items = {
	'x0a3': [
		{href: '#', title: 'One'},
		{onClick: () => alert('test'), title: 'Two'},
		{title: 'Three', children: 'x0a4'},
	],
	'x0a4': [
		{href: '#', title: 'Four'},
		{title: 'Five', children: 'x0a5'}
	],
	'x0a5': [
		{title: 'Six'},
		{title: 'Seven'}
	]
};
```

Just adding a new API option! What do you think? I'm fine with going on like this

bryceosterhaus:
@matuzalemsteles I agree that animation in JS only would be easier, but we also have these animations in the css since clay-css is agnostic of react. Since that is the case, I think its best we use something like `react-transition-group` so that we can use those animations from clay-css. Its a little tricky to use at first, but I think we can get in the pattern of using it correctly. However, if we were tied to having the animations in clay-css, I would be all for pure js animations. But I don't think we can count on that unfortunately...


> I am concerned with the depth of the API

I'm not totally sure what you mean here. Right now we allow for an unlimited level of menus, so we should cover menus that are 10 or 100 deep. Are you thinking of the case where the use may not have all the menus upon initialization(async child menus)? If thats the case, I think it would still work, but I would just need to test that out.

matuzalemsteles:
> @matuzalemsteles I agree that animation in JS only would be easier, but we also have these animations in the css since clay-css is agnostic of react. Since that is the case, I think its best we use something like react-transition-group so that we can use those animations from clay-css. Its a little tricky to use at first, but I think we can get in the pattern of using it correctly. However, if we were tied to having the animations in clay-css, I would be all for pure js animations. But I don't think we can count on that unfortunately...

🥺 Sure, makes sense.

> I'm not totally sure what you mean here. Right now we allow for an unlimited level of menus, so we should cover menus that are 10 or 100 deep. Are you thinking of the case where the use may not have all the menus upon initialization(async child menus)? If thats the case, I think it would still work, but I would just need to test that out.

Oh I'm actually referring to why the API is a tree, and that might be a problem. Just wondering if we could make it easier to use a list instead of a tree in depth.

jbalsas:
> Oh I'm actually referring to why the API is a tree, and that might be a problem. Just wondering if we could make it easier to use a list instead of a tree in depth.

I think one thing we saw in the past is that deeply nested structures are hard to update. We've experienced this both on PageEditor and Forms. Having to update some prop in the model involves a deep navigation and update throughout the whole structure, so in some of those cases we went with an indexed flat structure instead to simplify lookups and updates.

@wincent and @carloslancha might remember better...

wincent:
> @wincent and @carloslancha might remember better...

Indeed, although I don't have anything brief/pithy to say. The answer is always "it depends"... specifically it depends on the access patterns: how you have to read/traverse the data structure (and how often), and how you have to update it (and how often).

For example, in a mostly read-only context where the data is rarely updated, having a tree structure often leads to cleaner code, because subtrees in the data structure map directly onto subtrees in the view hierarchy, and you end up just passing them as props.

If there are a lot of updates though, and you are working with an immutable data structure, working with a tree can feel less convenient (for example, in order to update a leaf node, you end up having to create a new node, and then new nodes for each of its ancestors as you walk up the tree all the way to the root). For simple cases you can avoid that with a "sidecar" datastructure that is easier to update (eg. if every node has an ID, and you want to be able to update node colors, you can have a separate map that goes from node ID to color and never have to update the tree itself).

The case of the page editor, IIRC, was an example that was tricky because nodes can be reparented (ie. dragged from one spot of the tree to another), and they were going to (and did) implement undo on top, which meant that it was a relatively write-heavy workflow. Not only that, but in order to do things like check whether a node could be dropped onto another, they had to walk the data structure many times to during drags to check for eligible drop zones, so it was read-heavy too. In their case they did end up flattening the representation of the tree structure somewhat in order to avoid expensive traversals and updates.

matuzalemsteles:
Thanks @wincent and @jbalsas for bringing this. I think the Forms case would be practically the same case faced by the PageEditor, but really I think it depends on this case it is very likely that we will only use it for reading in the case of Frontend and maybe we will not change this data anymore or it will be updated. I'm concerned that this would be the opposite on the backend, and then we get into those problems again. I imagine that the backend will prepare the data to pass to this component and that this is probably not static.

wincent:
I don't think it's a problem that the backend and frontend have different "impedances" when it comes to data structure format. If there is a schism, you just translate from one format to another exactly once at the boundary (technically, can be on the server on on the frontend before the data is passed "into" the app). In an ideal world, everything would be in perfect harmony, with the frontend (of course!) dictating what the shape of the data should be 😉  (this is why GraphQL is designed the way it is) but we don't live in that world, and translating is cheap anyway.

pat270:
The way it's marked up in Clay CSS fits better with a flat structure as @matuzalemsteles mentioned. An item in `x0a5` can link back to the beginning of the menu.

bryceosterhaus:
@matuzalemsteles the latest commit is a POC of using a flat structure. There were a few gotchas

1. There must be a top level menu with no parent menu.
2. A child menu cannot have two different parents (for the sake of the header title)
3. Every menu _must_ have an identifier

I'll work on it a bit more today and think I might be able to get rid of 1 & 2.

pat270:
> A child menu cannot have two different parents (for the sake of the header title)

If the next menu is marked up before the current menu and needs to slide right to left (next menu slides in from right), you might need `translateX(-200%)`.

bryceosterhaus:
@pat270 I think you should only need -200% if more than 2 menus have `display: block`. With the current implementation, I think only 2 should ever have that at the same time(during animation).

bryceosterhaus:
Just updated again. I was able to make it a bit cleaner, the only requirement is that each menu has an identifier and also that each clickable item in the menus has a title.

matuzalemsteles:
> the only requirement is that each menu has an identifier and also that each clickable item in the menus has a title.

I think this is not too bad, it would be perhaps the expected for this structure! Yeah, it seems cleaner now.

bryceosterhaus:
@matuzalemsteles updated with docs and tests.

My thought is that we release with this basic high-level component and then see what peoples needs are for the low-level components and decide how granular we want to abstract. Right now it seems like the non-dropdown use cases still aren't fully worked out.

bryceosterhaus:
cc @pat270 can you verify the classes and markup on the demo? My concern is the `dropdown-*` classes used

matuzalemsteles:
@bryceosterhaus looks good, I just looked quickly, tomorrow I'll look in more detail 😬.

matuzalemsteles:
hey @bryceosterhaus the PR of @pat270 https://github.com/liferay/clay/pull/3356 can influence anything here? there seems to be a lot of changing it, it's likely to break something here.

pat270:
@bryceosterhaus @matuzalemsteles Looking at it right now, I don't think that pull will break much.

The only thing I see right now is `drilldown-item-header` might look something like:
drilldown-item-header

pat270:
@bryceosterhaus For dropdown-menu as drilldown I think we need:
```
<ul class="nav nav-stacked">
```
to be:
```
<ul class="inline-scroller">
```
Lexicon wants dropdown menu drilldowns with lots of items to scroll at a certain height.

Not sure if you want to take care of these here or in another pr: 
- I noticed the endless example adds `< Menu 1` to the header when it goes back to the beginning. It should have no header and just `Menu 2`.

- If I drilldown through the menu several times, I can drillback several times to get to the first menu.

- Using the keyboard to drilldown is confusing to navigate

pat270:
If we add in the other pr @matuzalemsteles mentioned, Lexicon wants to have the option of having dual click areas where the caret icon is a `a`/`button` and text is also a `a`/`button`/`span` for both header and item.

dual-click

matuzalemsteles:
@pat270 and @bryceosterhaus so, if we are planning to merge the PR, then it would be better to merge it and make the necessary markup changes. What do you think?

pat270:
@matuzalemsteles yes ideally, 3356 should go in first.

bryceosterhaus:
@pat270 if I remove all the dropdown-* classes and also add the "inline-scroller" class instead of nav, I get this

Screen Shot 2020-07-23 at 1 34 28 PM

Screen Shot 2020-07-23 at 1 34 33 PM

Is that expected? Or do we expect to always use the dropdown classes with this component?

pat270:
@bryceosterhaus If the drilldown is supposed to look like a dropdown-menu we need to have the `dropdown-*` classes. I haven't seen a clear definition of where else we will use drilldown so maybe we can only have the option of `inline-scroller` + `dropdown-*` classes.

wincent:
> #3356 should go in first.

So that one went in. Any updates on this one? I am not sure if there are any teams waiting for this, but I have a doc that gets shared with other teams listing the stuff we're working on and I just want to update it with status.

bryceosterhaus:
@wincent I'm going to get this merged today, just need to tweak a few more things.

bryceosterhaus:
Alright, I ended up modifying this and taking a slightly different approach. 

Since we don't have super clear details on where the non dropdown version is going to be used or how it is going to be used, I ended up just primarily making this component used through `@clayui/drop-down`.

- `@clayui/drilldown` is a new package that exposes low-level components
- `@claui/drop-down` consumes drilldown and exports the `DropDownWithDrilldown` component

The main reason I went this route is because I realized if we put this all into the drilldown package, it would require us to re-create all of the dropdown logic with the trigger and show/hide. I think going this direction makes it a bit more clear and reduces the maintenance overhead of duplicating functionality across packages.

See dropdown stories for example

wincent:
> The main reason I went this route is because I realized if we put this all into the drilldown package, it would require us to re-create all of the dropdown logic with the trigger and show/hide. I think going this direction makes it a bit more clear and reduces the maintenance overhead of duplicating functionality across packages.

Good call, but another option is just to put everything into `@clayui/drop-down`, so as to avoid the overhead that comes with a separate package. I know that you can in theory use a drilldown without a drop-down, but in practice it seems you almost always will use them together. I know that modularity is all the rage, but I think it's overrated in this context. Up to you though how you slice it.

jbalsas:
> I know that you can in theory use a drilldown without a drop-down, but in practice it seems you almost always will use them together.

In this case, we have the opposite requirement. We know this can be used within a dropdown and is the most natural place since that's also where it was defined, but the initial request for this actually lives **outside** of it :) 

See https://www.figma.com/file/Ln1fkqNTqRX4nniHXFfLiy/LPS-114110-global-apps-responsive?node-id=1%3A60 in the medium and small screen sizes

wincent:
> In this case, we have the opposite requirement. We know this can be used within a dropdown and is the most natural place since that's also where it was defined, but the initial request for this actually lives **outside** of it :)

I'm not arguing about the _conceptual hierarchy_ (eg. drilldown subordinate to drop-down vs drop-down subordinate to drilldown), I'm making a _practical_ argument in favor of reducing the number of packages by grouping two oft-used-together things in a single package. Seeing as `drop-down` already exists, that would mean add `drilldown` to it, thus avoiding the creation of the `@clayui` package number 45.

I think the cumulative cost (in terms of adding friction to the development, maintenance, and even usage) of having such a large number of packages in Clay has added up to quite a bit over the lifetime of the project, and while modularity sounds great in theory (granularity — pay only for what you use), in practice I think it ends up providing no significant benefit.

jbalsas:
> I think the cumulative cost (in terms of adding friction to the development, maintenance, and even usage) of having such a large number of packages in Clay has added up to quite a bit over the lifetime of the project, and while modularity sounds great in theory (granularity — pay only for what you use), in practice I think it ends up providing no significant benefit.

I totally agree with this. Just wanted to point out that we DO have evidence and designs that specifically require us to use this outside of a dropdown contrary to what was stated throughout the conversation.

That being said, I'm more than fine with it solely living inside `@clayui/dropdown` if that simplifies maintenance and overhead.

jbalsas:
> @wincent @jbalsas how about we put it all in dropdown for now to handle the dropdown use case and then whenever the other use case and markup get fleshed out, we can evaluate the best way forward to provide that? Regardless whether we end up with another package or not, I think its appropriate to be able to use Drilldown via @clayui/drop-down.

I'm fine either way. Keeping it under wraps until we see the real generic use case for it seems more in line with our current general approach, so it sounds reasonable.

bryceosterhaus:
Updated. Removed drilldown package and put it all into dropdown.

packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
```diff
@@ -0,0 +1,24 @@
...
+    "protocol": "http:",
+    "reload": [Function],
+    "replace": [Function],
+    "search": "",
+    "toString": [Function],
+  },
+  Symbol(SameObject caches): Object {
+    "styleSheets": StyleSheetList {},
+  },
+}
```
matuzalemsteles, line 23: 🤔 Curious why it didn't generate a DropDown snapshot with Drilldown

bryceosterhaus: or weird, I didn't notice this. Ill take a look

pat270:
@matuzalemsteles yeah it looks off by 1px. We can tweak that in Clay CSS.

bryceosterhaus:
Going to merge this and if we need to update anything else, just let me know

PR #3491, opened 2020-07-10
feat(clayui.com): Add documentation for @clayui/toolbar

### LINKED ISSUE #3486: Toolbar should have documentation that matches it's new implementation
When I [made the new Toolbar component](https://github.com/liferay/clay/issues/3300) I didn't add any documentation about it. 

Was looking through our issues and noticed this one [Move upper-toolbar to toolbar](https://github.com/liferay/clay/issues/3336) and I noticed that we don't have any documentation, so here we go, it might help with the issue above.

@bryceosterhaus not sure which milestone to pick here, both have 1 same date and 1 different one. (screenshot was too big and not relevant to the issue so here's a link to it for context https://prnt.sc/temzqs)

kresimir-coko:
Fixes #3486 

Let me know if my grammar is correct and if the heading `Composing` is correct in this context, I think it is because the content under it is describing how to mix and match low-level components to get the desired result.

@pat270 let me know if the Markup/CSS page needs work for this component.


pat270:
@kresimir-coko this looks good to me. I think adding more examples would help, but that can come later. Also maybe we can keep the code collapsed by default? I find myself having to scroll a lot through the code to find the component I'm looking for, only to get lose my place because all the code looks the same.

kresimir-coko:
I've pushed a commit with an additional example, like the one we have in Upper Toolbar, that showcases some specific examples and nesting as recommended in the paragraph above it.

packages/clay-toolbar/README.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Introduction
+
+A toolbar organizes actions into a horizontal bar that is responsive. Its height changes to accommodate the height of the elements it contains. The toolbar always maintains the vertical alignment.
+
+<Toolbar />
+
+## Composing
+
+**Nav** is the highest level wrapper to be used inside the Toolbar to wrap all the content inside.
```
bryceosterhaus, line 35: For these bolded component names, I would prefix them, `ClayToolbar.*`, since that is the way they will be used.

packages/clay-toolbar/README.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: 'Toolbar'
+description: 'A toolbar is a set of actions related to a specific context that are grouped into a horizontal bar.'
+lexiconDefinition: 'https://liferay.design/lexicon/core-components/toolbars/'
+packageNpm: '@clayui/toolbar'
+sibling: 'docs/components/css-toolbar.html'
```
bryceosterhaus, line 6: Since this component isn't published yet, lets add "draft" so that its not published in the docs yet.

kresimir-coko:
@bryceosterhaus I've pushed the requested changes! 😄

bryceosterhaus:
Looks good!

PR #3492, opened 2020-07-10
fix(@clayui/css): Reboot removes focus outline on negative `tabindex` elements. Chrome adds outline on focus to any element with a `tabindex` attribute.

### LINKED ISSUE #3436: @clayui/css Chrome adds focus outline on elements with [tabindex="-1"] when clicked
## What is your proposal?
 From MDN:
> A negative value (usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation, but could be focused with Javascript or visually by clicking with the mouse. It's mostly useful to create accessible widgets with JavaScript.

We should disable focus outline on `tabindex="-1"` elements by default by adding a rule-set that removes focus outline on any element with a negative tabindex `[tabindex^='-'] { outline: 0; }`. We can redo the focus outline if we need that behavior.

This causes focus outline on click for `a.disabled` elements that shouldn't have them. It's also an issue in places like https://clayui.com/docs/components/modal.html, where clicking the modal dialog shows a focus border.

pat270:
fixes #3436

### DIFF
--- a/clayui.com/content/docs/components/css-links.md
+++ b/clayui.com/content/docs/components/css-links.md
@@ -90,7 +90,7 @@ Use these patterns for actions in components.
 			<use href="/images/icons/icons.svg#times" />
 		</svg>
 	</a>
-	<a class="component-action disabled" href="#disabled" role="button" tabindex="-1">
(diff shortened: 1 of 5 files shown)


marcoscv-work:

PR #3499, opened 2020-07-10
fix(@clayui/modal): simplifies subportal grouping implementation and removes the extra container

### LINKED ISSUE #3107: ClayModal size CSS class is applied twice

matuzalemsteles:
Fixes #3107

This removes the extra container that we were adding before. I did some tests with a DropDown being rendered and other bugs that we got related to that and it works well. The strange thing is that I have the feeling that I’ve done this before and I don't remember the reason why I didn't do it 🤷‍♂️.

Thanks @pat270 for helping me with this!

### DIFF
--- a/packages/clay-modal/src/Hook.ts
+++ b/packages/clay-modal/src/Hook.ts
@@ -27,12 +27,13 @@ const FOCUSABLE_ELEMENTS = [
  * modal by pressing the ESC key and control the focus within the Modal.
  */
 const useUserInteractions = (
-	elementRef: React.MutableRefObject<any>,
+	modalElementRef: React.MutableRefObject<any>,
+	modalBodyElementRef: React.MutableRefObject<any>,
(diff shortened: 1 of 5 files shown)


bryceosterhaus:
@matuzalemsteles looks good to me!

PR #3505, opened 2020-07-13
fix(@clayui/layout): use col- in favor of col-xs-

### LINKED ISSUE #3363: ClayLayout remove `col-xs` from the Col component
In the new [Clay Layout Component Col](https://github.com/liferay/clay/blob/master/packages/clay-layout/src/Col.tsx#L65) we are setting `col-xs` as default when `size` is defined but I think we should remove that because when I declare `size` I expect to get `col-{size}` instead of `col-xs-{size}`

You can see it in [this PR](https://github.com/liferay-frontend/liferay-portal/pull/53#issuecomment-648859955)

At this point I'm not sure if we need `col-xs` at all https://getbootstrap.com/docs/4.5/layout/grid/#grid-options

bryceosterhaus:
fixes #3363


eduardoallegrini:
I was wondering... do we need to keep the `xs` in the Component? [(lines: 41-45, 58, 63-65, 78-79)](https://github.com/liferay/clay/blob/master/packages/clay-layout/src/Col.tsx)

bryceosterhaus:
@eduardoallegrini we should keep the prop in the component since it would technically be a breaking change if we removed it. With this fix though, you shouldn't encounter any more `col-xs-${size}` classes anymore. Both the `size` and `xs` prop will do the same thing.

bryceosterhaus:
merging for todays release

Issue #3430, opened 2020-07-02
Replace event.keyCode usages with event.key


diegonvs:
## What is your proposal?

While reviewing a [PR](https://github.com/liferay/clay/pull/3419#discussion_r448737852), @pat270 brings an important point: .keyCode for identifying a pressed key is no longer recommended.

 See: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

```
This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.
```

The idea is replace these usages, @pat270 recommended using `event.key` due to compatibility with IE11, but according  https://caniuse.com/#feat=keyboardevent-key: `Has non-standard key identifiers and incorrect behaviour with AltGraph.`.

When grepping our [Clay codebase](https://gist.github.com/diegonvs/89654431236967272f4ff3bfc43b8311), I noticed We're using on the Modal, MultiSelect and on FocusScope component.

## Why would adopting this proposal be beneficial?

Avoid future problems with Keyboard Events.

## What are the alternatives to this proposal?

Use `.code`(but [no support for IE11](https://caniuse.com/#feat=keyboardevent-code))

wincent:
I'm not sure what keys we need to support, but the list of non-standard event names that we'd need to cover might be pretty short; ie. [this list from MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) might get us pretty far:

```
  switch (event.key) {
    case "Down": // IE/Edge specific value
    case "ArrowDown":
      // Do something for "down arrow" key press.
      break;
    case "Up": // IE/Edge specific value
    case "ArrowUp":
      // Do something for "up arrow" key press.
      break;
    case "Left": // IE/Edge specific value
    case "ArrowLeft":
      // Do something for "left arrow" key press.
      break;
    case "Right": // IE/Edge specific value
    case "ArrowRight":
      // Do something for "right arrow" key press.
      break;
    case "Enter":
      // Do something for "enter" or "return" key press.
      break;
    case "Esc": // IE/Edge specific value
    case "Escape":
      // Do something for "esc" key press.
      break;
```

bryceosterhaus:
[...]

PR #3513, opened 2020-07-14
chore: Replace `keyCode` usages with `key` for KeyboardEvents

diegonvs:
Some keys can be different depending the browser which is being running the application but React normalize it when creating the synthetic event on: https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/events/getEventKey.js#L12-L29

### How I tested It:
Regarding
> but React normalize it when creating the synthetic event

From https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values: 
>[1] Older browsers may return "Spacebar" instead of " " for the Space Bar key. Firefox did so until version 37, as did Internet Explorer 9, 10, and 11.

Spacebar: 

https://github.com/liferay/clay/pull/3513/files#diff-18ecf9fd4dc1fbbb2a2bff5e734cc990L52

1. Open our Storybook on IE11
2. Open ClayCardWithNavigation story
3. Try navigating via tabs and select the first card with Spacebar key
4. You'll see an alert opening

Closes https://github.com/liferay/clay/issues/3430

### DIFF
--- a/packages/clay-card/package.json
+++ b/packages/clay-card/package.json
@@ -32,7 +32,7 @@
 		"@clayui/label": "^3.3.1",
 		"@clayui/layout": "^3.1.1",
 		"@clayui/link": "^3.2.0",
-		"@clayui/shared": "^3.1.2",
+		"@clayui/shared": "^3.1.3",
 		"@clayui/sticker": "^3.1.1",
 		"classnames": "^2.2.6"
 	},
--- a/packages/clay-card/package.json
+++ b/packages/clay-card/package.json
@@ -32,6 +32,7 @@
 		"@clayui/label": "^3.3.1",
 		"@clayui/layout": "^3.1.1",
 		"@clayui/link": "^3.2.0",
+		"@clayui/shared": "^3.1.2",
 		"@clayui/sticker": "^3.1.1",
 		"classnames": "^2.2.6"
 	},
--- a/packages/clay-card/src/CardWithNavigation.tsx
+++ b/packages/clay-card/src/CardWithNavigation.tsx
@@ -54,11 +54,6 @@ interface IProps {
 	title?: React.ReactText;
 }
 
-const KEYCODES = {
-	ENTER: 13,
-	SPACE: 32,
-};
-
 const noop = () => {};
 
 export const ClayCardWithNavigation: React.FunctionComponent<IProps> = ({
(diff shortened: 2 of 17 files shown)


wincent:
> One small change I think would be helpful is to use the keys as constants at the top of the file and then reference those instead of the string directly.

Another possibility would be to move them into a file in `clay-shared` so that you're not repeating them in a number of places. That file would also provide you with a good place to put a comment explaining why/how the key names are normalized.

Consider:

```js
const SPACE = ' ';

if (event.key === SPACE) {
   // ...
}
```

Some well-intentioned developer comes along and says, "Hm, that might not work on IE; I'm going to change that":

```js
if (event.key === ' ' || event.key === 'Spacebar') {
  // ...
}
```

And it could sneak through code review, which means that it is relatively brittle.

On the other hand, imagine this:

```js
import {Keys} from '@clayui/shared';

if (event.key === Keys.SPACE ) {
   // ...
}
```

Developer might still think, "Hm, I wonder if that will work in IE", but when they look in `clay-shared` they'll see:

```js
/**
 * Key names for comparison with `Event.prototype.key`.
 *
 * Note that because React normalizes key names in its synthetic
 * event system, we don't have to worry about browser-specific
 * edge cases (such as "Spacebar" instead of " " in IE):
 *
 * See:
 *
 * - https://github.com/facebook/react/blob/b87aabdfe1/packages/react-dom/src/events/getEventKey.js#L12-L29
 * - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values
 */
export const Keys = {
  SPACE: ' ',
  // ...
};
```

In other words, less brittle, and people are more likely to copy the pattern because they'll see `import {Keys} from '@clayui/shared';` in various places around the repo.

Just an idea for you to consider anyway.

bryceosterhaus:
> @bryceosterhaus 
> I would say lets just start with copy/paste for each map and then we can evaluate down the road if we want to abstract to @clayui/shared



> @wincent 
> Another possibility would be to move them into a file in `clay-shared`


Sorry for the confusion @diegonvs 😂😂😂

_btw:_ I'm fine with both copypasta and shared

diegonvs:
Force pushing for fixing a dep version mismatch error :p

diegonvs:
Updated cd54a8191b46b0a0760ccc6ac31fbcefe6fd51bf

bryceosterhaus:
Hmm, not sure why the clayui.com build is failing... Fails for me locally on master as well, but the latest commit built fine
Screen Shot 2020-07-17 at 7 35 26 PM

bryceosterhaus:
Built fine at https://github.com/liferay/clay/pull/3526. @diegonvs I just triggered a re-deploy, maybe that'll fix it 🤷‍♂️

diegonvs:
Hey @bryceosterhaus:

Greener as amazon rainforest in 80ties:
Screen Shot 2020-07-20 at 17 41 45

PR #3515, opened 2020-07-15
fix(@clayui/modal): fixes error of closing the modal when the target element was removed from the DOM before the click event

matuzalemsteles:
See https://github.com/liferay-frontend/liferay-portal/pull/83#issuecomment-658840945

This should correct the case of an element being removed before reaching the click event.


packages/clay-modal/src/Hook.ts
```diff
@@ -81,12 +83,28 @@ const useUserInteractions = (
 
 	const handleDocumentClick = (event: Event) => {
 		if (event.defaultPrevented) {
+			mouseEventTargetRef.current = null;
+
 			return;
 		}
 
-		if (event.target === modalElementRef.current) {
+		if (
```
brunobasto, line 91: Not really sure how `event.target` could have been unintentionally changed so that it would fall into this `if` statement when it shouldn't. From talking to Matu it seems Chrome will change the target of the event during the bubbling sequence if the target element was removed in the middle of the bubbling.

[The spec](https://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-flow-bubbling) says 

> The chain of EventTargets from the event target to the top of the tree is determined before the initial dispatch of the event. If modifications occur to the tree during event processing, event flow will proceed based on the initial state of the tree.

So, in theory, if the element was removed it shouldn't change the flow, but maybe browsers don't implement this consistently.

jbalsas: When I saw this, my expectation was that if `mousedown` and `mouseup` didn't happen on the same target, then `onclick` wouldn't be triggered... in any case, we could definitely check both `mousedown` and `mouseup` rather than `click`...

matuzalemsteles: > When I saw this, my expectation was that if mousedown and mouseup didn't happen on the same target, then onclick wouldn't be triggered... in any case, we could definitely check both mousedown and mouseup rather than click, since those should be at least consistent?

Hmm, in the case we would stop listening to the `click` event and use the `mousedown`? looks good.

jbalsas: I'd say both, to make it closer to click.

Reacting on mousedown or mouseup can create unexpected interactions. A mousedown inside modal, and mouseup outside and viceversa will make the modal close weirdely.

I'd play around with that and I'd say we need to do mouseup but only if mousedown target is the same (thay would be what click is)

matuzalemsteles: Hmm I think in the `mouseup` in the case of Workflow closing the auto-complete in the so-called `mousedown` it is likely that the event.target will change from `mousedown` to `mouseup`. I think it ends up going back to the case. Really if we were to use both for tracker in the end we would do the same here I think. Why we probably need some information to check the integrity of the event during `mouseup` and `mousedown`, in which case it might be `event.target`.

Analyzing the cases better, I really think I was missing something, because there may be an event before `mouseup` and `mousedown` that can remove the element from the DOM or change the value and lose the `event.target`, perhaps thinking better this is not a good lesson here 😕. I realized that an onBlur event is called before the Autocomplete closes in Workflow, so we come back to the problem 😅 thinking about other solutions

matuzalemsteles: I tried to track how I could solve this problem but it is very complicated to think about circumventing the behavior that changes `event.target` to the element of `.fade.modal` that close Modal. So these are the cases, it seems that this behavior always existed before the recent changes in Modal. Some cases:

1. Clay components that use `React.Portal` is added as a sibling to the `.modal-dialog` element so that they can be visible and have no z-Index problems.
2. When a click event occurs in a DropDown.Item, for example, it can close DropDown, adding the `show` class, this makes DropDown invisible.
3. As he closed DropDown in the click event, the `event.target` from the document is still from the element that the user clicked on and so we did not close Modal

Second case:

1. When a click event occurs in DropDown.Item and closes DropDown before the `click` event
2. The Modal click event to decide to close the Modal, the `event.target` changes to the `.fade.modal` element and the modal is closed (I still unable to understand why it always goes to that element)

Third case:

1. We add the extra div that wraps the element `.modal-dialog` and the portals are added as a brother of `.modal-dialog` (the same we had before)
2. When a click event occurs in DropDown.Item and closes DropDown before the `click` event
3. In Modal's click event, `event.target` changes to the "wrapping" element, so, we don't close Modal because we use `.fade.modal` to check if it was closed or not.

If you take this third case and place the Autocomplete by closing the Autocomplete in the Item's MouseDown and having the Autocomplete keep an item out of Modal and try to click, the `event.target` it will be `.fade.modal` 🤷‍♂️.

This behavior of the browser deciding which is the element of `event.target` when the element was invisible before the click event, it is still very strange, I not being able to think about which rule it is following, I'm assuming it has to be an element visible and focusable perhaps.

jbalsas: So, what I meant is:

1. On `mousedown`, keep the target ref
2. On `mouseup`, check if the target is the same as the target ref from the `mousedown` event
    a. If it's not, do nothing
    b. If it is and it's outside the modal... dismiss

matuzalemsteles: @jbalsas I was thinking about doing this but maybe we would still have problems if an event does this before `mousedown`, it seems that the Autocomplete in Workflow is closed in the `onBlur` of the `input`, this event seems to be called before the `mousedown`, I will check this out in more detail today, maybe I missed something. It was late when I was checking this out, so my mind might not be thinking very well 😅

matuzalemsteles:
hey @jbalsas I continued with the idea of ​​using `mouseup` and `mousedown`, I did a test on today's master and it works well, in fact, `onBlur` was being called later.

I think we can go ahead with that, I'm just a little concerned if the Portal has dependencies on the click event, as we are closing Modal before the click.

jbalsas:
> I think we can go ahead with that, I'm just a little concerned if the Portal has dependencies on the click event, as we are closing Modal before the click.

Only one way to find out! 😉 

@bryceosterhaus said he'd be on PTO a couple of days... let's see if he can still look at this, or else, I'd trust @brunobasto judgement here. We can merge this and cut a modal release and see how tests behave in portal.

matuzalemsteles:
> @bryceosterhaus said he'd be on PTO a couple of days... let's see if he can still look at this, or else, I'd trust @brunobasto judgement here. We can merge this and cut a modal release and see how tests behave in portal.
> We could also wait for the next release, but I'd rather clear this up to avoid surprises :)

Sounds good to me!

bryceosterhaus:
taking a look now. Gonna test it out and if all looks good, ill cut a release of it today and send to portal

bryceosterhaus:
resent at https://github.com/liferay/clay/pull/3538

updated for merged conflict

PR #3519, opened 2020-07-16
fix(@clayui/color-picker): Add option to make Color Picker small

### LINKED ISSUE #3514: ClayColorPicker doesn't have a small variation

https://codesandbox.io/s/xenodochial-thunder-tt314?file=/src/index.js

Placing a ClayColorPicker inside a ClayForm.Group with the small property set to true or pass the property sizing="sm" to the ClayColorPicker

The ClayColorPicker is show in the small input variation (32px instead of 40px)

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.2.4  |
| React |         |

pat270:
fix(@clayui/color-picker): Removes fixed sizes for Splotch

fixes #3514

Not related to this issue, but I was wondering why we use a hidden input to store the color value when we already have an input in the `input-group`? Can't we just use that one?

### DIFF
--- a/packages/clay-color-picker/README.mdx
+++ b/packages/clay-color-picker/README.mdx
@@ -18,11 +18,12 @@ import {ColorPicker} from '$packages/clay-color-picker/docs/index';
 
 ## Types of Color Picker
 
-Color Picker is delivered in 3 different ways: default colors, custom colors and native.
+Color Picker is delivered in 4 different ways: default colors, custom colors, native and small.
 
 -   **Default colors**: a predefined list of colors to the dropdown that is not allowed to change or manipulate.
 -   **Custom colors**: using the [`colors`](#api-colors) API in conjunction with [`onColorsChange`](#api-onColorsChange) enables the user to modify colors and add new ones.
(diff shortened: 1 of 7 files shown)


pat270:
I updated with a blurb and example on storybook.

The disable input doesn't make sense. If the input is disabled I would assume the whole color-picker input-group should be disabled (color picker dropdown included).

Your comment made me realize we have the simple color-picker where it's just the dropdown button. I'm thinking it was to accommodate that. It's possible to use the regular color-picker, hide the `input-group-item` and remove `input-group-prepend` to achieve that.

marcoscv-work:
LGTM and yeah, it looks like the extra hidden input is not needed

jbalsas:
So, I haven't double checked this, but the goal of the hidden input (afaict) would be to make this component work seamlessly inside a traditional html form, so that the colour selected value is send if you submit a form without having to use JS to retrieve the value.

Seeing as we have 2 inputs, I assume maybe we can tweak it somehow, but please, keep this requirement in mind.

bryceosterhaus:
Looks good!

Issue #3521, opened 2020-07-17
Add two new spacers


marcoscv-work:
## Add two new spacers

In order to accomplish the new need for the Epic https://issues.liferay.com/browse/LPS-114841 and after conversations with Juan Antón and Sandro we will need to add two new spacers.
The task for portal is: https://issues.liferay.com/browse/LPS-117363
But it will only add the new two custom properties.

The spacers are:
```
9 = $spacer * 9
10 = $spacer * 10
```

PR #3522, opened 2020-07-17
feat(@clayui/css): Add new spacers

marcoscv-work:
Fixes: https://github.com/liferay/clay/issues/3521

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_globals.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_globals.scss
@@ -214,6 +214,12 @@ $spacers: map-deep-merge(
 		8: (
 			$spacer * 7.5,
 		),
+		9: (
+			$spacer * 9,
+		),
+		10: (
+			$spacer * 10,
+		),
 	),
 	$spacers
 );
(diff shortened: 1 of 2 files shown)


pat270:
@marcoscv-work @jbalsas this adds 18kb to `atlas.css`, 2kb zipped. Are we good with that size increase? Other than that it's good with me.

marcoscv-work:
I cry each time I see those increases but based on need from Echo team, new tokens, and their editor.. we have to :-(

jbalsas:
Why do we have to do the spacers deep-merge twice? Is it possible we're generating more than needed?

I take that the current spacers already account for 16kb zipped... definitely not ideal, but as @marcoscv-work says... it is what it is...

pat270:
We need `map-deep-merge` in `atlas.scss` so Sass will overwrite/replace/add to the `$spacers` map that is passed in from `_clay_variables.scss`. Then this new map will do the same to the `$spacers` map in `base.scss`. This is preferred so the user doesn't need to redeclare all the key value pairs when modifying a map.

We might be generating more than needed, but it's not from `map-deep-merge`. We are generating `margin` and `padding` utilities for every breakpoint with and without `!important`.

```
  .c-m-sm-10 {
    margin: 10rem;
  }
  .c-mt-sm-10,
  .c-my-sm-10 {
    margin-top: 10rem;
  }
  .c-mr-sm-10,
  .c-mx-sm-10 {
    margin-right: 10rem;
  }
  .c-mb-sm-10,
  .c-my-sm-10 {
    margin-bottom: 10rem;
  }
  .c-ml-sm-10,
  .c-mx-sm-10 {
    margin-left: 10rem;
  }
```

PR #3523, opened 2020-07-17
feat(clayui.com): Make code collapsed by default and add localStorage to keep user's preference

### LINKED ISSUE #3502: Should we collapse code examples by default?
@bryceosterhaus @matuzalemsteles @pat270 @jbalsas 

What do you guys think, should we collapse the code examples by default, to increase visibility and discoverability across the site? Or do we keep it as is right now?

This was brought up a couple times in the past, and most recently [here](https://github.com/liferay/clay/pull/3491#issuecomment-656872363)

kresimir-coko:
Fixes #3502 

It's been a while since I've worked with `localStorage` so I looked around the internet and I found [this neat article](https://joshwcomeau.com/react/persisting-react-state-in-localstorage/) that helped me understand exactly what's happening here.

I went pretty much line by line from the article's example seeing as I couldn't find a better way of approaching this implementation.

The idea being that when a user reveals at least one of the code examples and refreshes the page with at least one of the examples being revealed, it will be saved as a preference for the user until they collapse all examples on the page they're currently viewing.

I've tried simplifying this:
```
return stickyValue !== null
	? JSON.parse(stickyValue)
	: defaultValue;
```
but whenever I remove `JSON.parse` the text `true` or `false` gets passed instead of it's real value, even tried double bangs `!!stickyValue` but that had the same result, I think because even a string `false` is still truthy since it's not empty.

Also, removing the sanity check: `stickyValue !== null` provided to be bad, I assume because of rehydration, but not exactly sure, maybe someone can let me know why.

I kept the `useStickyState` as a function to provide better clarity in the future and for some future usage, I didn't export it into a separate file as I wasn't sure if that was our workflow when making custom hooks inside of the `clayui.com/components` workspace, I noticed we do have `Hooks/useResource.js` but I first wanted to get some feedback if this is the approach we want to use.

### DIFF
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -59,8 +59,22 @@ const Editor = ({
 	}
 
 	function useStickyState(defaultValue, key) {
+		const inBrowser = typeof window !== 'undefined';
+
+		const localStorage = {
+			getItem(key) {
+				return inBrowser ? window.localStorage.getItem(key) : null;
+			},
+
+			setItem(key, value) {
+				if (inBrowser) {
+					window.localStorage.setItem(key, value);
+				}
+			},
+		};
+
 		const [value, setValue] = React.useState(() => {
-			const stickyValue = window.localStorage.getItem(key);
+			const stickyValue = localStorage.getItem(key);
 
 			return stickyValue !== null
 				? JSON.parse(stickyValue)
@@ -68,7 +82,7 @@ const Editor = ({
 		});
 
 		React.useEffect(() => {
-			window.localStorage.setItem(key, JSON.stringify(value));
+			localStorage.setItem(key, JSON.stringify(value));
 		}, [key, value]);
 
 		return [value, setValue];
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -59,24 +59,9 @@ const Editor = ({
 		];
 	}
 
-	const inBrowser = typeof window !== 'undefined';
-
-	const localStorage = {
-		getItem(key) {
-			return inBrowser ? window.localStorage.getItem(key) : null;
-		},
-
-		setItem(key, value) {
-			if (inBrowser) {
-				window.localStorage.setItem(key, value);
-			}
-		},
-	};
-
 	const [collapseCode, setCollapseCode] = useStickyState(
 		true,
-		'collapse-code',
-		localStorage
+		'collapse-code'
 	);
 	const [activeIndex, setActiveIndex] = React.useState(0);
 	const [snippets, setSnippets] = React.useState(code);
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -12,6 +12,7 @@ import React, {useContext, useState} from 'react';
 import {LiveEditor, LiveError, LivePreview, LiveProvider} from 'react-live';
 
 import theme from '../../utils/react-live-theme';
+import useStickyState from '../Hooks/useStickyState';
 
 const spritemap = '/images/icons/icons.svg';
 
@@ -58,39 +59,24 @@ const Editor = ({
 		];
 	}
 
-	function useStickyState(defaultValue, key) {
-		const inBrowser = typeof window !== 'undefined';
+	const inBrowser = typeof window !== 'undefined';
 
-		const localStorage = {
-			getItem(key) {
-				return inBrowser ? window.localStorage.getItem(key) : null;
-			},
+	const localStorage = {
+		getItem(key) {
+			return inBrowser ? window.localStorage.getItem(key) : null;
+		},
 
-			setItem(key, value) {
-				if (inBrowser) {
-					window.localStorage.setItem(key, value);
-				}
-			},
-		};
-
-		const [value, setValue] = React.useState(() => {
-			const stickyValue = localStorage.getItem(key);
-
-			return stickyValue !== null
-				? JSON.parse(stickyValue)
-				: defaultValue;
-		});
-
-		React.useEffect(() => {
-			localStorage.setItem(key, JSON.stringify(value));
-		}, [key, value]);
-
-		return [value, setValue];
-	}
+		setItem(key, value) {
+			if (inBrowser) {
+				window.localStorage.setItem(key, value);
+			}
+		},
+	};
 
 	const [collapseCode, setCollapseCode] = useStickyState(
 		true,
-		'collapse-code'
+		'collapse-code',
+		localStorage
 	);
 	const [activeIndex, setActiveIndex] = React.useState(0);
 	const [snippets, setSnippets] = React.useState(code);
@@ -121,19 +107,31 @@ const Editor = ({
 				/>
 
 				<div style={{padding: '10px'}}>
-					<ClayButtonWithIcon
-						className="btn-copy"
-						displayType="unstyled"
-						small
-						spritemap={spritemap}
-						symbol={'paste'}
-						title={'Copy'}
-					/>
+					{snippets.length === 1 && collapseCode && (
+						<span className="align-items-center clay-p d-flex mt-0">
+							{'Code Sample (expand to see it)'}
+						</span>
+					)}
+
+					{!collapseCode && (
+						<ClayButtonWithIcon
+							className="btn-copy"
+							displayType="unstyled"
+							small
+							spritemap={spritemap}
+							symbol={'paste'}
+							title={'Copy'}
+						/>
+					)}
 
 					<ClayButtonWithIcon
 						className="btn-collapse"
 						displayType="unstyled"
-						onClick={() => setCollapseCode(!collapseCode)}
+						onClick={() => {
+							setActiveIndex(0);
+
+							setCollapseCode(!collapseCode);
+						}}
 						small
 						spritemap={spritemap}
 						symbol={collapseCode ? 'angle-right' : 'angle-down'}
@@ -144,12 +142,22 @@ const Editor = ({
 						<ClayTabs modern>
 							{snippets.map((snippet, index) => (
 								<ClayTabs.Item
-									active={activeIndex === index}
+									active={
+										!collapseCode && activeIndex === index
+									}
 									innerProps={{
 										'aria-controls': `tabpanel-${snippet.name}`,
 									}}
 									key={snippet.name}
-									onClick={() => setActiveIndex(index)}
+									onClick={() => {
+										if (activeIndex !== index) {
+											setCollapseCode(false);
+										} else {
+											setCollapseCode(!collapseCode);
+										}
+
+										setActiveIndex(index);
+									}}
 								>
 									{snippet.name}
 								</ClayTabs.Item>
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -58,7 +58,26 @@ const Editor = ({
 		];
 	}
 
-	const [collapseCode, setCollapseCode] = useState(false);
+	function useStickyState(defaultValue, key) {
+		const [value, setValue] = React.useState(() => {
+			const stickyValue = window.localStorage.getItem(key);
+
+			return stickyValue !== null
+				? JSON.parse(stickyValue)
+				: defaultValue;
+		});
+
+		React.useEffect(() => {
+			window.localStorage.setItem(key, JSON.stringify(value));
+		}, [key, value]);
+
+		return [value, setValue];
+	}
+
+	const [collapseCode, setCollapseCode] = useStickyState(
+		true,
+		'collapse-code'
+	);
 	const [activeIndex, setActiveIndex] = React.useState(0);
 	const [snippets, setSnippets] = React.useState(code);
 
--- a/clayui.com/src/components/Hooks/useStickyState.js
+++ b/clayui.com/src/components/Hooks/useStickyState.js
@@ -9,33 +9,41 @@ const inBrowser = typeof window !== 'undefined';
 
 const localStorage = {
 	getItem(key) {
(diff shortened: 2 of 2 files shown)


jbalsas:
Mmm.... deploy preview failed... shame, wanted to see this but no time for building myself :(

kresimir-coko:
> Mmm.... deploy preview failed... shame, wanted to see this but no time for building myself :(

Okay it is as I feared, `WebpackError: ReferenceError: window is not defined`, it's throwing this error because `window` is not defined server side, it's just the way Gatsby works, in development it's fine, but in production it throws that error, so I have to find a workaround. 🚧

wincent:
> so I have to find a workaround

```js
const inBrowser = typeof process === 'undefined' || String(process) !== '[object process]';

// or:

const inBrowser === typeof window !== 'undefined';

const localStorage = {
    getItem(key) {
        return inBrowser ? window.localStorage.getItem(key): undefined;
    },

    setItem(key, value) {
        if (inBrowser) {
            window.localStorage.setItem(key, value);
        }
    }
};
```

or something along those lines?

kresimir-coko:
I've pushed a sanity check inspired by @wincent's idea, there was a problem with JSON not being able to parse `undefined` but I circumvented that by making sure it never receives `undefined`. I've been testing with `yarn build` to make sure it works on Production, as running that command was giving me the same error as the Deploy Preview prior to this fix

clayui.com/src/components/Editor/index.js
```diff
@@ -55,16 +55,30 @@ const Editor = ({
...
+				return inBrowser ? window.localStorage.getItem(key) : null;
+			},
+
+			setItem(key, value) {
+				if (inBrowser) {
+					window.localStorage.setItem(key, value);
+				}
+			},
+		};
+
```
wincent, line 71: You could move all of this outside of `useStickyState` so as not to redefine it each time the hook is used.

jbalsas:
Hey @kresimir-coko, I think this is a good start! 2 things I saw (on a quick check)

Screen Shot 2020-07-20 at 11 46 08

Clicking on the tabs won't expand the code. I think it might be better if we did:
- Initially, no active tab? (because the user hasn't interacted with anything)
- Click on selects that language and expands the code sample
- Expanding from the expand button should select the first tab

Screen Shot 2020-07-20 at 11 46 14

In the case there aren't multiple languages, the tab area feels a bit off and might be indeed hard to discover. 

Maybe rather than what I wrote above, what we could do is add an additional header "Code Sample (expand to see it)" that's always present, so the expanding action is clear... and include the tabs within the collapsible part, so we don't have to deal with their state.

I'd bet @drakonux and @dsanz have insights into this as much as anyone else :)

kresimir-coko:
Interesting ideas there @jbalsas I haven't really considered them, but I like them! Let's see what others have to say and I'll jump on it when we reach a consensus

clayui.com/src/components/Editor/index.js
```diff
@@ -54,7 +54,40 @@ const Editor = ({
...
+			const stickyValue = localStorage.getItem(key);
+
+			return stickyValue !== null
+				? JSON.parse(stickyValue)
+				: defaultValue;
+		});
+
+		React.useEffect(() => {
+			localStorage.setItem(key, JSON.stringify(value));
+		}, [key, value]);
```
bryceosterhaus, line 82: Whats the purpose of the `JSON.parse` and `JSON.stringify` here? Since our value is just a boolean value, I think we would be fine to not use parse and stringify.

wincent: Given that local storage are [always strings](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), using `JSON` ensures you get the same type back out that you put in.

If you pass in boolean values `true`/`false`, they get stored as (effectively) `'true'`/`'false'`, which are both truthy when you read them back out, so the other places in the code where we're doing things like `!collapsed` won't work as expected.

clayui.com/src/components/Editor/index.js
wincent: And actually, now that I look at the broader context I see the entire definition of the `useStickyState` hook itself could/should be moved outside of the component.

kresimir-coko: I agree @wincent it can be moved outside, I'll play around with it soon. Thanks for taking the time to look at it in a broader context 🥇

dsanz:
I like the idea of a "Show me the code" header that would be collapsed by default and remembered per-user. In this case, I'd also remember the language preference in the localStorage, so that once you choose one, page reload/navigation sticks to that preference. In addition, the copy button should only be visible when code sample is shown.

In case some sample is not provided in the preferred language, we might show it in other language, plus some indicator that the remembered language is not available. I think this should not imply to "change" the localStorage preference, it's just a fallback for that specific sample not to be expanded and empty. If that's hard to implement, then we may show it empty with a "not available" text or something like that.

clayui.com/src/components/Editor/index.js
bryceosterhaus: @wincent sounds good!

kresimir-coko:
@jbalsas @dsanz is this what you had in mind? It's just a screenshot where I added the additional header, no functionality yet.
image

Now I'm unsure as to the design of this, it's kinda redundant in situations where we have the tabs, but in situations where we don't have the tabs, I can see it being used. Reading both your comments a couple times I got the feeling that you suggested to have that header visible all the time, even when the code is expanded, while also having tabs. Where does the user click to expand the code then, the additional header, the expand button, both?

What I think would be better, is to show that `Code sample (expand to see it)` text only when the example doesn't have the tabs, and show it to the left of the Copy and Expand icons, something like this:

image

Another thing we can also do, is replace the Copy and Expand icons with text, it's one word per action, so it won't take up too much space, and it shouldn't look terrible, something like this:

image

In all of these approaches, I'd straight up hide the `Code sample (expand to see it)` text/header once the code has been expanded, as it's, IMO, providing no value at that point.

P.S. Disregard the Copy button being shown when code is collapsed, I'm about to fix that.

jbalsas:
Yup, the second thing is what I think both of us were suggesting. I don't have a strong opinion on the UX for this :)

dsanz:
What if "Code Sample" is shown when not expanded, then it's substituted by the tabs when expanded? Right buttons would remain in place all the time (maybe the "copy" button shall be made visible only wneh expanding), so it would look like this (sorry for using "code" markup to show a toolbar, I know....):

When collapsed:
```
Code sample (expand to see it)                                                               Expand
```

When expanded (many tabs):
```
React | JSP | ...                                                                         Copy      Collapse
```

I'm proposing this based on @kresimir-coko suggestions: in the end, the "Code Sample" bar is redundant when we expand this. Even when we don't have tabs. I think it'd be better if we just show the language for which the (single) sample is being provided, like this

When expanded (just one tab):
```
React                                                                                      Copy      Collapse
```

Would that be aligned with your suggestions @kresimir-coko?

kresimir-coko:
> sorry for using "code" markup to show a toolbar

@dsanz don't worry about it, you got your point across, and in a simple way, I like it 😄 

That's a valid point you make about showing the tab even when it's just one tab present, because it has the space for it anyway and it can provide value to people by saying `HTML` even if it's obvious for some of us.

kresimir-coko:
So here's what I've done with this new commit:

- Moved `useStickyState` to it's own file inside the `Hooks` folder
- Move variables defined within `useStickyState` outside, to the component using it
- Hide Copy button when code is collapsed
- Add text signalling the user to expand the code when the code is collapsed and when it doesn't have any tabs
- Make no tab be selected by default when the code is collapsed
- Make clicking on currently expanded tab collapse that tab
- Make clicking on Expand open the 1st tab even if the last tab to be opened wasn't the first tab

I feel like I'm writing patch notes 👾 

I'm going on a vacation for the next couple weeks, so if this isn't final someone can pick it up and finalize it so it doesn't get dragged for long, but I think this should be final.

clayui.com/src/components/Editor/index.js
```diff
@@ -54,39 +55,24 @@ const Editor = ({
...
+				window.localStorage.setItem(key, value);
+			}
+		},
+	};
 
 	const [collapseCode, setCollapseCode] = useStickyState(
 		true,
-		'collapse-code'
+		'collapse-code',
+		localStorage
```
wincent, line 75: Having to pass in `localStorage` makes the hook less convenient to use now. If another file wants to use it, they will import it, and run into exactly the problem that you ran into that led you to write the `inBrowser` stuff and the `localStorage` wrapper. I think you should move all of _that_ too.

Counter-argument: users can now pass in an arbitrary object that conforms to the local storage interface. Such flexibility!

Counter-counter-argument:

squinting

One useful rule of thumb about whether this kind of thing should be in another file in the first place: I don't extract something that is used in exactly one place into a separate file unless the place where I am extracting it from is getting large and unwieldy (not really the case here?). Keeping things private until they need to be public gives you more freedom to change them, which is a good thing; because hard-to-change abstractions are The Enemy™ and easy-to-change abstractions are The Best™.

PS. Enjoy your vacations @kresimir-coko.

pat270:
If I disable `localStorage` in my browser, the site doesn't render. We should probably check if `localStorage` is available, if not render everything collapsed?

localStorage-error

wincent:
Good, er, `catch`, @pat270. A `try`/`catch` around attempts to access `localStorage` seems appropriate.

pat270:
Another thing from me, it seems kind of odd that we have two plugins for collapsing code (React / JSP pages and CSS pages). Maybe we could workout unifying the two so the code can stay in one place.

clayui.com/src/components/Editor/index.js
kresimir-coko: Alrighty @wincent that comment sounds like a fork in the road that just meets back into a single road at the other side.

I'll go ahead and add the `inBrowser` & `localStorage` to the exported hook. 

The reason why I didn't do it initially was that I anticipated that `const inBrowser = typeof window !== 'undefined';` wouldn't execute fast enough if it was imported through a separate file, and that we didn't want to declare `localStorage` whenever the hook was being called.

P.S. Vacation was great, the weather wasn't 100% but we managed 🏖️

kresimir-coko:
Hey @wincent & @bryceosterhaus I've pushed a couple new commits which are basically adding what was suggested when I left for vacation.

As per @wincent's comment I extracted the `inBrowser` and `localStorage` to the custom hook, and as per @pat270's watchful eye I've attempted to catch errors when `localStorage` is disabled.

I'm not great at `try catch` so I might have butchered it, but the `try` part works on my end, because I couldn't find out how to disable local storage on my Chrome. A lot of articles online point to an older setting that doesn't exist anymore, so @pat270 can you let me know how I can disable it so I can test this and to know for future.

clayui.com/src/components/Hooks/useStickyState.js
```diff
@@ -0,0 +1,46 @@
...
+
+		return stickyValue !== null ? JSON.parse(stickyValue) : defaultValue;
+	});
+
+	React.useEffect(() => {
+		try {
+			localStorage.setItem(key, JSON.stringify(value));
+		} catch (error) {
+			return;
+		}
```
wincent, line 40: Personally, I'd move this one closer to (or rather, inside of) your `localStorage` implementation, so that it is closer to where the actual exception might be thrown.

(Note: didn't suggest the same for your other `try`/`catch` because `JSON.parse` can throw, so it is appropriate to have a `try`/`catch` there.)

clayui.com/src/components/Hooks/useStickyState.js
```diff
@@ -0,0 +1,46 @@
...
+	const [value, setValue] = React.useState(() => {
+		let stickyValue;
+
+		try {
+			stickyValue = localStorage.getItem(key);
+		} catch (error) {
+			stickyValue = null;
+		}
+
+		return stickyValue !== null ? JSON.parse(stickyValue) : defaultValue;
```
wincent, line 32: One thing to bear in mind is that `JSON.parse` can throw too (eg. someone sticks garbage in their local storage → BOOM), so you might move that into the `try` too:

```js
const [value, setValue] = React.useState(() => {
    try {
        const value = localStorage.getItem(key);

        return value === null ? defaultValue : JSON.parse(value);
    } catch (error) {
        return defaultValue;
    }
});
```

You can omit the `(error)` entirely if we have Babel set up to allow optional `catch` binding in this repo (but I don't know if we do).

pat270:
@kresimir-coko you can check Block all cookies, that should disable localStorage.
localstorage

clayui.com/src/components/Editor/index.js
```diff
@@ -84,19 +88,31 @@ const Editor = ({
...
 				<div style={{padding: '10px'}}>
-					<ClayButtonWithIcon
-						className="btn-copy"
-						displayType="unstyled"
-						small
-						spritemap={spritemap}
-						symbol={'paste'}
-						title={'Copy'}
-					/>
+					{snippets.length === 1 && collapseCode && (
```
bryceosterhaus, line 91: Should this be `!!snippets.length` meaning that only show if snippets exist? Or is this explicitly checking against a single snippet

kresimir-coko: It's supposed to show the `Code Sample (expand to see it)` text only when there is a single snippet of code - when no tabs are displayed, primarily in the HTML/CSS examples, because they don't have language variants.

kresimir-coko:
@bryceosterhaus & @wincent I've added some extra `try...catch` blocks and located them closer to the thing we wanna test.

I've also tested that it works with Local Storage disabled. Note: it won't work on the `development` site, not even on `master`, but when serving a `production` site, it will work with or without Local Storage, as it does on the currently released version.

wincent:
> I've added some extra `try...catch` blocks

Good to see that you could do it without the optional catch bindings.

bryceosterhaus:
Sent update to https://github.com/liferay/clay/pull/3606

Issue #3336, opened 2020-06-15
Move upper-toolbar to toolbar package


bryceosterhaus:

kresimir-coko:
Not sure what the situation with this issue is @bryceosterhaus but I'll work on documentation for @clayui/toolbar since that's something we're missing and it might give you some foundation for this issue.

bryceosterhaus:
@kresimir-coko no situations here, just havent gotten the time to get to this. Feel free to pick it up!

PR #3528, opened 2020-07-20
feat(@clayui/toolbar): Add Upper Toolbar high-level component

kresimir-coko:
Hey @bryceosterhaus I started looking at #3336 and investigating, I'm looking for some inspiration as to how I should approach it. 

This PR has only one very simple commit right now, an example of how I imagined this should be done. Basically adding a UpperToolbar high-level component inside the Toolbar package.

Outside of that, I'm unsure how to deprecate the UpperToolbar on our side (Clay), is there some configuration that needs to be changed, or added to signal that it's deprecated? 

Here's what I assume needs to be done, before we replace usages in DXP:

- [ ] Mark `@clayui/upper-toolbar` as deprecated
- [ ] Add Upper Toolbar variant to `@clayui/toolbar`, including implementation, story, and documentation
- [ ] Remove Upper Toolbar page from clayui.com, and remove `draft` from Toolbar page

Let me know if I'm missing something and if you have some quick advice regarding some more obscure things that need to be done (like some special configuration that maybe needs to be added)

### DIFF
--- a/packages/clay-toolbar/src/Input.tsx
+++ b/packages/clay-toolbar/src/Input.tsx
@@ -7,9 +7,7 @@ import {ClayInput} from '@clayui/form';
 import classNames from 'classnames';
 import React from 'react';
 
-export interface IProps
-	extends React.InputHTMLAttributes<HTMLInputElement>,
-		React.ComponentProps<typeof ClayInput> {}
+export interface IProps extends React.ComponentProps<typeof ClayInput> {}
 
 const Input: React.FunctionComponent<IProps> = ({className, ...otherProps}) => (
 	<ClayInput.Group>
--- a/packages/clay-toolbar/src/Input.tsx
+++ b/packages/clay-toolbar/src/Input.tsx
@@ -1,0 +1,28 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ClayInput} from '@clayui/form';
+import classNames from 'classnames';
+import React from 'react';
+
+export interface IProps
+	extends React.InputHTMLAttributes<HTMLInputElement>,
+		React.ComponentProps<typeof ClayInput> {}
+
+const Input: React.FunctionComponent<IProps> = ({className, ...otherProps}) => (
+	<ClayInput.Group>
+		<ClayInput.GroupItem>
+			<ClayInput
+				className={classNames(className, 'form-control')}
+				type="text"
+				{...otherProps}
+			/>
+		</ClayInput.GroupItem>
+	</ClayInput.Group>
+);
+
+Input.displayName = 'ClayToolbarInput';
+
+export default Input;
--- a/packages/clay-toolbar/src/index.tsx
+++ b/packages/clay-toolbar/src/index.tsx
@@ -60,7 +60,7 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
 		{
 			'component-tbar': !subnav,
 			'subnav-tbar': !!subnav,
-			'tbar-light': !!light,
+			'tbar-light': light,
 		},
 		subnav && {
 			'subnav-tbar-disabled': subnav.disabled,
--- a/packages/clay-toolbar/src/index.tsx
+++ b/packages/clay-toolbar/src/index.tsx
@@ -7,6 +7,7 @@ import classNames from 'classnames';
 import React from 'react';
 
 import Action from './Action';
+import Input from './Input';
 import Item from './Item';
 import Label from './Label';
 import Link from './Link';
@@ -18,6 +19,11 @@ interface IProps extends React.HTMLAttributes<HTMLElement> {
 	 * Adds a helper class that turns the Toolbar inline at a specified breakpoint.
 	 */
 	inlineBreakpoint?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
+
+	/**
+	 * Determines if the tbar-light class should be added to the Toolbar, making it's background white.
+	 */
+	light?: boolean;
 
 	/**
 	 * Defines if the toolbar should have the `subnav-tbar` class.
@@ -33,6 +39,7 @@ interface IProps extends React.HTMLAttributes<HTMLElement> {
 const ClayToolbar: React.FunctionComponent<IProps> & {
 	Action: typeof Action;
 	Item: typeof Item;
+	Input: typeof Input;
 	Label: typeof Label;
 	Link: typeof Link;
 	Nav: typeof Nav;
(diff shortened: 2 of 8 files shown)


bryceosterhaus:
Hey @kresimir-coko here are the steps I think we need to take

- [ ] Move `clay-upper-toolbar` components to `toolbar`
- [ ] Abstract `Item` to its own low-level component
- [ ] Create stories showcasing upper-toolbar from https://liferay.design/lexicon/core-components/toolbars/toolbar-upper/#usage
- [ ] Point upper-toolbar docs to the new toolbar component


And then for deprecating upper-toolbar and releasing toolbar package, I can take care of those.

bryceosterhaus:
@kresimir-coko also see https://github.com/liferay/clay/issues/3300#issuecomment-637874889 for background on why we are creating a toolbar package.

We should be less focused on "upper-toolbar" and more focused on creating low-level components for toolbar variants(upper is one of them). See lexicon for details on toolbars as well.

kresimir-coko:
I've force pushed some commits that bring this closer to the finish line with a few questions from me

### Styles
@pat270 I've changed the default horizontal `padding` for `tbar-item` seeing as it was set to 4px compared to the expected 8px from [Lexicon](https://liferay.design/lexicon/core-components/toolbars/toolbar-upper/). LMK if I should've done it differently. Also let me know if the color classes I used are good in HTML, namely `text-light`.

### Upper Toolbar Component
@bryceosterhaus Investigating our current [implementation](https://github.com/liferay/clay/blob/master/packages/clay-upper-toolbar/src/index.tsx#L59) of Upper Toolbar I can see it's very barebones, basically the same as our Toolbar [implementation](https://github.com/liferay/clay/blob/master/packages/clay-toolbar/src/index.tsx#L33). Which brings me back to the [comment](https://github.com/liferay/clay/issues/3300#issuecomment-637874889) you linked, specifically this line:
> Upper-toolbar may not even need to be its own component anymore either since there isn't really anything specific to it.

Which is the reason why I removed the `UpperToolbar` component completely.

### Upper Toolbar Story
Which one of these is supposed to be Upper Toolbar, the one above or under the red line? https://prnt.sc/tzmr9p
Me not knowing this, I included 2 stories, each representing one of those toolbars, so LMK about this please.

packages/clay-toolbar/src/Input.tsx
```diff
@@ -0,0 +1,26 @@
...
+import classNames from 'classnames';
+import React from 'react';
+
+const Input: React.FunctionComponent<React.HTMLAttributes<
+	HTMLInputElement
+>> = ({className, ...otherProps}) => (
+	<ClayInput.Group>
+		<ClayInput.GroupItem>
+			<ClayInput
+				className={classNames(className, 'form-control-inline')}
```
pat270, line 16: The class `form-control-inline` doesn't exist in Bootstrap or Clay CSS. The class `form-control` should suffice.

packages/clay-toolbar/src/index.tsx
```diff
@@ -52,6 +58,9 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
 		{
 			'component-tbar': !subnav,
 			'subnav-tbar': !!subnav,
+			[`tbar-${
+				displayType === 'light' ? displayType : `${displayType}-l2`
```
pat270, line 62: What happens if you don't pass in `displayType`? I envision users wanting to pass in their own custom modifier class, maybe we can give 2 options `light` and `dark`, but don't output anything if displayType has no value.

Also there are two placeholders for dark, `.dark-d1` and `.dark-l2`. Lexicon has no definitions for these so, maybe we don't need a dark option at all.

packages/clay-toolbar/stories/index.tsx
```diff
@@ -211,4 +212,97 @@ storiesOf('Components|ClayToolbar', module)
...
+							small
+							spritemap={spritemap}
+							symbol="ellipsis-v"
+						/>
+					</ClayToolbar.Item>
+				</ClayToolbar.Nav>
+			</ClayLayout.ContainerFluid>
+		</ClayToolbar>
+	))
+	.add('Upper Toolbar #2', () => (
```
pat270, line 274: We don't need this example. This is covered by Navbar (Application Bar). I think having this example would mislead users to using the wrong component.

pat270:
@kresimir-coko changing the padding on `$tbar-item-padding-x` shouldn't be a problem. I don't think `tbar` is used in Portal other than in `sidebar`, but I could be wrong. The only issue I see is if a team is using the component and sized things in a way to make it fit in small screens with 4px padding; increasing the padding would make it break to new line.

packages/clay-toolbar/stories/index.tsx
kresimir-coko: Alright, great, that was my main point of confusion. Thanks for clarifying

packages/clay-toolbar/src/index.tsx
kresimir-coko: If `displayType` isn't provided, the toolbar's background is transparent.

> Lexicon has no definitions for these so, maybe we don't need a dark option at all.

Sure, I don't have any reason to argue, I'll omit the `displayType` prop and add a boolean `light` prop that adds the `tbar-light` class.

kresimir-coko:
I've done the following with the most recent pushed commits: 
- Removed the `displayType` prop and replaced it with `light` prop that adds the `tbar-light` class (would've done the same for dark if it existed, but it doesn't).
- Initially I didn't actually import Toolbar.Input so I fixed that now /yikes
- Remove 2nd Upper Toolbar story as per @pat270's comment
- Add Toolbar.Input to the Upper Toolbar story

Should be ready for review now!

pat270:
LGTM

packages/clay-toolbar/src/index.tsx
```diff
@@ -52,6 +60,7 @@ const ClayToolbar: React.FunctionComponent<IProps> & {
 		{
 			'component-tbar': !subnav,
 			'subnav-tbar': !!subnav,
+			'tbar-light': !!light,
```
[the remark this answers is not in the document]
kresimir-coko: I think that I got confused a bit by seeing `!!subav` but that has double bangs because that's not a simple prop. You're right, it's unnecessary to have them for `light`

kresimir-coko:
Hey @matuzalemsteles I've squashed the old, already reviewed commits and pushed the 3 new ones that tackle the things you said needed to be done:
- Deprecating Upper Toolbar
- Simplifying the props and extensions
- Expanding the packageStatus so it can output custom values too

matuzalemsteles:
Resent #3646

packages/clay-toolbar/src/index.tsx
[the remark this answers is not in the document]
wincent: You would think that this is the kind of thing that the TS compiler would be able to warn about (casting an already-boolean to a boolean); maybe in a future "strict" mode update...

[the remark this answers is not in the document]
kresimir-coko: Yea, wouldn't that be nice

PR #3532, opened 2020-07-20
fix(@clayui/list): When running on IE11, use `target.ownerDocument.activeElement` instead of `event.relatedTarget` on onBlur callbacks

### LINKED ISSUE #3531: event.relatedTarget may not be filled on onBlur callbacks in IE11
-   [✅ ] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

See https://github.com/facebook/react/issues/3751

See https://github.com/liferay/clay/blob/f33dc148b53b48317d95751a3d59d66b9daac055/packages/clay-list/src/Item.tsx#L66

This may causing a potential bug on ClayList component running in IE11

diegonvs:
See: https://github.com/facebook/react/issues/3751

Closes #3531

### DIFF
--- a/packages/clay-list/src/Item.tsx
+++ b/packages/clay-list/src/Item.tsx
@@ -6,8 +6,6 @@
 import classNames from 'classnames';
 import React from 'react';
 
-const isIE11 = typeof window !== 'undefined' && 'msCrypto' in window;
-
 interface IProps extends React.HTMLAttributes<HTMLLIElement> {
 	/**
 	 * Flag to indicate if the `list-group-item-action` class should be applied.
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
@diegonvs is there a specific bug that is appearing in ie11 because of this? I wasn't sure if there was something broken, or if this is fixing something that could be broken in the future.

diegonvs:
> @diegonvs is there a specific bug that is appearing in ie11 because of this?

Not related to an existing bug but I found this tip somewhere and when checking on our codebase I found this smell and I decided to send a fix proactively.

diegonvs:
> @diegonvs I'm not sure if using target.ownerDocument.activeElement would solve our case, I tested it locally using Chrome the behavior seems to break, so it would be good to test this in IE 11 and check if the behavior looks the same as tested in https://storybook.clayui.com/?path=/story/components-claylist--complex

When trying to test storybook on IE11 I can't see any icon(apparently spritemap is not being loaded).

Testing visually, nothing changed without using the fix or not see the following gif:

Kapture 2020-07-21 at 17 29 27

However, when console.log the received value of relatedTarget and seeing the fix in action:
Screen Shot 2020-07-21 at 17 44 19

diegonvs:
That's weird, I may not testing adequately: Focusing the List Items and seeing if the DOM changed properly with quick actions elements.

`focus` class is changing properly on the item element when focusing the elements with the fix and not.

bryceosterhaus:
@diegonvs Yeah I thinking we discard for now since its not an issue at the moment and can't seem to reproduce when it would be an issue. But this is good to keep in mind though, if we do encounter any sort of issue with focus/activeElement we can keep this in mind.

PR #3535, opened 2020-07-20
fix(@clayui/color-picker): Splotch should use classes provided by Clay CSS to style borders on light colors and active colors

### LINKED ISSUE #3525: @clayui/color-picker Dropdown Menu active swatch style doesn't match Lexicon
From Lexicon https://liferay.design/lexicon/core-components/forms/picker-color/#predefined-palette-panel:
lexicon-color-picker

Clay Color Picker:
clay-color-picker

We are also inlining the border styles and active outline via JS. Clay CSS has classes:
|state|class|
|---|---|
|light color swatches|`clay-color-btn-bordered`|
|active outline|`focus` (should be `.active`, need to map `.focus` to it in Clay CSS)|

pat270:
~fix(@clayui/color-picker: Splotch adds `tabindex="-1"` when a splotch is active~

~One thing I noticed when tabbing through the dropdown menu is an element with `tabindex="-1"` doesn't take it out of the tab order. I was wondering if we had something that is managing the focus outside the browser.~

Edit: Not fixing these in this pr.

fixes #3525

### DIFF
--- a/packages/clay-color-picker/src/Splotch.tsx
+++ b/packages/clay-color-picker/src/Splotch.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import classNames from 'classnames';
 import React from 'react';
 import tinycolor from 'tinycolor2';
 
@@ -33,13 +34,12 @@ const ClayColorPickerSplotch = React.forwardRef<HTMLButtonElement, IProps>(
 		return (
 			<button
 				{...otherProps}
-				className={`btn clay-color-btn ${className ? className : ''}`}
+				className={classNames('btn clay-color-btn', className, {
+					active,
+					'clay-color-btn-bordered': requireBorder,
+				})}
 				ref={ref}
 				style={{
-					...(active ? {outline: 'auto 3px #55ADFF'} : {}),
-					...(requireBorder
-						? {border: '1px solid #E7E7ED'}
-						: {borderWidth: 0}),
 					background: `#${value}`,
 					height: size,
 					width: size,
--- a/clayui.com/content/docs/components/css-color-picker.md
+++ b/clayui.com/content/docs/components/css-color-picker.md
@@ -129,7 +129,7 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/forms/picker-
 					</div>
 					<div class="clay-color-swatch">
 						<div class="clay-color-swatch-item">
-							<button class="btn clay-color-btn focus" title="#B2EDFF" style="background-color:#B2EDFF;"></button>
+							<button class="active btn clay-color-btn" title="#B2EDFF" style="background-color:#B2EDFF;" tabindex="-1"></button>
 						</div>
 						<div class="clay-color-swatch-item">
 							<button class="btn clay-color-btn" title="#45EDC5" style="background-color:#45EDC5;"></button>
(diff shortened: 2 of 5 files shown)


packages/clay-color-picker/src/Splotch.tsx
```diff
@@ -33,17 +33,16 @@ const ClayColorPickerSplotch = React.forwardRef<HTMLButtonElement, IProps>(
 		return (
 			<button
 				{...otherProps}
-				className={`btn clay-color-btn ${className ? className : ''}`}
+				className={`${active ? 'active ' : ''}btn clay-color-btn ${
+					requireBorder ? 'clay-color-btn-bordered ' : ''
+				}${className ? className : ''}`}
```
matuzalemsteles, line 38: To avoid this and organize better, we can add the `classnames` package, as we use it elsewhere, so it would look more or less:

```js
className={classnames('btn clay-color-btn', className, {
	active,
	'clay-color-btn-bordered': requireBorder
})}
```

packages/clay-color-picker/src/Splotch.tsx
```diff
@@ -33,17 +33,16 @@ const ClayColorPickerSplotch = React.forwardRef<HTMLButtonElement, IProps>(
...
 				style={{
-					...(active ? {outline: 'auto 3px #55ADFF'} : {}),
-					...(requireBorder
-						? {border: '1px solid #E7E7ED'}
-						: {borderWidth: 0}),
 					background: `#${value}`,
 					height: size,
 					width: size,
 				}}
+				tabIndex={active ? -1 : undefined}
```
matuzalemsteles, line 45: Thinking better, maybe we don't even need it anymore.

pat270: Where can I find documentation on FocusScope? I can only find an issue on reactjs/rfcs. I'm just trying to take the active button out of the tab order. We could always use `span`/`div` but would be nice if there was a way to take a button out of tab order.

Maybe we could create another issue for managing focus, there are some other things I would like to make better.

matuzalemsteles: @pat270 `<FocusScope />` is our own implementation to deal with the Focus of the elements when they are using `React.Portal`, unfortunately, we don't have documentation of it because we created it for internal use. https://github.com/liferay/clay/blob/master/packages/clay-shared/src/FocusScope.tsx

> I'm just trying to take the active button out of the tab order. We could always use span/div but would be nice if there was a way to take a button out of tab order.

Hmm, I think I understand now, do you want to ignore the active element in the order of the tabs, in the tab should you skip the active element to the next one?

In this case, FocusManager will ignore non-focusable and `disabled` elements.

> Maybe we could create another issue for managing focus, there are some other things I would like to make better.

It might be better to create an issue, to describe how we can handle the focus in ColorPicker.

pat270: Sounds good I'll make an issue.

pat270:
ok hopefully this passes. Was I supposed to commit updates to the `yarn.lock` file?

matuzalemsteles:
> ok hopefully this passes. Was I supposed to commit updates to the yarn.lock file?

I would say yes, but it depends on which packages have been updated just to make sure that nothing breaks 😁

pat270:
It updated/removed several dependencies, but nothing related to classnames package.

matuzalemsteles:
> It updated/removed several dependencies, but nothing related to classnames package.

I think we can leave that to another PR and analyze the changes in more detail.

bryceosterhaus:
LGTM!
fixed the merge conflict and sent at https://github.com/liferay/clay/pull/3555

PR #3543, opened 2020-07-22
docs(@clayui/sticker): fix bad example

### LINKED ISSUE #3541: User Icon (Sticker) example
We have the following example for User Icons:

```
<ClaySticker className="sticker-user-icon" size="xl">
   <div className="sticker-overlay">
      <img className="sticker-img" src="/images/long_user_image.png" />
   </div>
</ClaySticker>
```

but `ClaySticker` is wrapping the children with `sticker-overlay`, so the correct example should be

```
<ClaySticker className="sticker-user-icon" size="xl">
   <img className="sticker-img" src="/images/long_user_image.png" />
</ClaySticker>
```

bryceosterhaus:
fixes #3541

I also added a new API to allow for imgs directly instead of custom classes and child elements.

### DIFF
--- a/packages/clay-sticker/src/index.tsx
+++ b/packages/clay-sticker/src/index.tsx
@@ -17,7 +17,7 @@ export type DisplayType =
 	| 'unstyled'
 	| 'warning';
 
-type Shape = 'circle' | 'rounded';
+type Shape = 'circle' | 'user-icon';
 
 type Position = 'bottom-left' | 'bottom-right' | 'top-left' | 'top-right';
 
@@ -56,6 +56,30 @@ interface IProps extends React.HTMLAttributes<HTMLSpanElement> {
 	size?: Size;
 }
 
+const Overlay: React.FunctionComponent<
+	React.HTMLAttributes<HTMLSpanElement> & {
+		/**
+		 * Flag to indicate if `inline-item` class should be applied
+		 */
+		inline?: boolean;
+	}
+> = ({children, className, inline, ...otherProps}) => (
+	<span
+		className={classNames(className, 'sticker-overlay', {
+			'inline-item': inline,
+		})}
+		{...otherProps}
+	>
+		{children}
+	</span>
+);
+
+const Image: React.FunctionComponent<React.ImgHTMLAttributes<
+	HTMLImageElement
+>> = ({className, ...otherProps}) => (
+	<img className={classNames(className, 'sticker-img')} {...otherProps} />
+);
+
 const ClaySticker: React.FunctionComponent<IProps> = ({
 	children,
 	className,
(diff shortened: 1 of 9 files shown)


bryceosterhaus:
@matuzalemsteles I really like that idea, however I'm not sure its possible to check the return element type for each child. We could explicitly check for `` but that would break for any custom component that returns an img. I'll look more into it and see what we can do though

bryceosterhaus:
@matuzalemsteles instead of making it smart, I added low-level components for composition. This seemed to be consistent with the rest of our components and a simple change.

packages/clay-sticker/src/index.tsx
```diff
@@ -54,8 +54,37 @@ interface IProps extends React.HTMLAttributes<HTMLSpanElement> {
 	 * Sticker size.
 	 */
 	size?: Size;
+
+	/**
+	 * Flag to indicate if `sticker-user-icon` class should be applied.
+	 */
+	userIcon?: boolean;
```
matuzalemsteles, line 61: This name seems a bit strange when you use it with any Sticker and Image, the icon of a user always comes to my mind, it seems very specific. Looking at the class it looks like it just rounds the edges and some details, maybe the name of the API is better with `circle`, what do you think?

bryceosterhaus:
@matuzalemsteles I just re-looked through the css classes we have and turns out "sticker-circle" is already a shape and that "sticker-rounded" doesn't even exist. So I instead removed "rounded" and added "user-icon" as a shape. I think this fits the existing api well and keeps us more in-line with the css.

matuzalemsteles:
@bryceosterhaus I had forgotten about the shape API. yeah, it makes sense this is like shape and being a user-icon the difference for the circle is very subtle but it is different.

Strange that there is no 'rounded', perhaps it has been removed from the CSS. @pat270 this may have been removed recently? I think we would realize that rounding did not exist when implementing this component.

pat270:
Bootstrap has utilities `.rounded` (4px), `.rounded-circle` (100%), and `.rounded-0` (0). We pass it down to `.sticker-overlay` so it has the same border radius as the `sticker`. The syntax for rounded is `sticker rounded`.

|syntax|radius|notes|
|---|---|---|
|sticker|4px|-|
|sticker sticker-circle|5000px|Generic classname for non user icon stickers|
|sticker sticker-user-icon|5000px|Semantic classname easier to style user icons across the board|
|sticker rounded|4px|Support Bootstrap syntax|
|sticker rounded-0|0|Support Bootstrap syntax|
|sticker rounded-circle|100%|Support Bootstrap syntax|

matuzalemsteles:
Thanks @pat270. Hey @bryceosterhaus, so maybe it's still worth keeping the shape `rounded` but updating with the new syntax.

bryceosterhaus:
@matuzalemsteles yeah since rounded had no effect, I think this is essentially a bug fix

PR #3548, opened 2020-07-23
fix(clayui.com): Move `code-collapse.js` code to `docs-site.js` and u…

### LINKED ISSUE #3547: clayui.com code-collapse.js doesn't work on IE11
~Can't open and close side navigation menu items due to JS error in `code-collapse.js`~

Can't toggle code snippets for html examples (files with `.md` extension) in IE11.
code-snippet

Edit: The side navigation menu IE11 issue needs more debugging. We have other JS on the page using ES6 syntax. This breaks all JS functionality in IE11. Probably related to #3546.

pat270:
…pdate so it works in IE11

I wasn't able to test if it works because our dev site doesn't work in IE11, but I think it will.

fixes #3547


matuzalemsteles:
hey @pat270 looking at this now, maybe we can move that implementation to React.

pat270:
@matuzalemsteles I don't think we can use React for this. This code should only apply to html code examples in `.md` files.

pat270:
I realized I need to polyfill `.matches()` for IE11 also

matuzalemsteles:
@pat270 sorry for delay. When I say move to React, we would move to the `docs.js` template, which takes care of the `.md` and `.mdx` pages, so that we do not leave this code loose.

pat270:
I'll take a look thanks @matuzalemsteles

PR #3549, opened 2020-07-23
feat: Icon for Collections in DXP

drakonux:
Due to [LEXI-962](https://issues.liferay.com/browse/LEXI-962) we've decided to add a new icon to represent a Collection in DXP. 

Closed-book

This is a complex concept because 1 or more collections can be part of an asset library. Thus we have decided to design the icon thinking the relationship with the asset library icon (_books_). In addition, we also added a little bookmark in anticipation of the icon for adding/saving elements to a Collection. 

@pat270 I named it _closed-book_ what do you think? I would have liked to name it book-closed but checking the other book icon (_info-book_) it's wiser to follow that naming order, isn't?


pat270:
@drakonux yeah I agree `book-closed` is easier to find. `closed-book` shouldn't be that much of an issue we have a good search function on clayui.com. If we decide to rename it, we can deprecate `info-book` and `closed-book` and rename them to `book-*`.

PR #3556, opened 2020-07-24
fix(@clayui/css): Clay Range hides `.clay-range-thumb` and styles thumb pseudo elements so input range will work without JS

### LINKED ISSUE #3551: @clayui/css: Clay Range should use the default input type range thumb
Spawned from #3503. Clay Range should provide basic range input functionality even without JS. The visible thumb doesn't move when dragged, but the underlying input thumb is actually moving. We should show that instead of our custom thumb.

pat270:
fixes #3551

### DIFF
--- a/clayui.com/content/docs/components/css-slider.md
+++ b/clayui.com/content/docs/components/css-slider.md
@@ -35,6 +35,45 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/slider/'
 
 ## Custom Slider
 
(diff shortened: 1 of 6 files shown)


pat270:
@matuzalemsteles yes they should all "work", just progress indicator and tooltip won't update

matuzalemsteles:
Thanks, lgtm then

pat270:
Thanks @matuzalemsteles I'll go ahead and merge it

PR #3560, opened 2020-07-24
fix(@clayui/css): Atlas Alerts change `.alert .close` font-size to 12px

### LINKED ISSUE #3557: Alerts: Replace times with times-small
Hi Clayers!

I'm updating the Figma library with the new feature [autolayout ](https://issues.liferay.com/browse/LEXI-838)and I also reviewing distances, sizes and little details. I started with [Alerts ](https://issues.liferay.com/browse/LEXI-843)and I would like to change the following:

## Proposal

- Change times icon with times-small for Alerts

## Current

image

14x14. I think the font size: 0.875 changes the normal size of the icon to that value.
## Expected

12px

Using _times-small_ 12x12 (by design without changing the font-size)

pat270:
fixes #3557 

@drakonux We decided to change the `font-size` in Clay CSS because we would have to change the icon name in too many places.

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_alerts.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_alerts.scss
@@ -15,6 +15,7 @@ $alert-close-opacity: 1 !default;
 $alert-close: () !default;
 $alert-close: map-deep-merge(
 	(
+		font-size: 0.75rem,
 		opacity: $alert-close-opacity,
 		disabled: (
 			color: inherit,


drakonux:
Hey guys, sorry to bother you with this @pat270 @matuzalemsteles 

If finally, you accept the new changes for `times` and `times-small`: https://github.com/liferay/clay/pull/3567#

I think it would be wiser and safer to keep the size of the updated `times` icon for alerts (because it's big but not so big as it was). 

Alerts (1)

I wrote to you before requesting a smaller `times` icon, so this comment is just in the opposite direction I said 🙇🏽‍♂️. 

There was and still is an internal discussion about how to use correctly the times icons. Lexicon should provide clear guidance for product designers about how to place them correctly (I know, we have to do it) but for now please I propose to use continue using `times` for alerts and "big" UI components such us modals and panels.

matuzalemsteles:
Hmm Thanks @drakonux. Don't you think it is safer to fix this via CSS than to leave two options available and impose good practice to use `times-small` or do we want to provide that option even now after the icon updates?

pat270:
The alert icon size we can change back, not a problem.

PR #3567, opened 2020-07-30
fix(@clayui/css): SVG Icons normalize `check`, `times`, `times-small` and feat(@clayui/css): SVG Icons `check-small`

drakonux:
Due to a conversation with @pat270 yesterday we have several doubts about the consequences of this [commit](https://github.com/liferay/clay/pull/3537/commits/fba13b65f13f2ffb63cef108a91ace2a2b434ae8) can cause.

The commit updates:
- The thickness of the `check` icon (from 1px to 2px)
- The size of the icon (from ≈12x≈10 to 10x8)

First commit sent

So that commit potentially can affect clients that are currently using `check`. 
In addition, if the check icon is used isolated as an indicator in a dropdown menu with the new check icon causes:

Indicators

Thus we propose the following: 
- Update check
- Update times
- Update times-small
- Create check-small

Some visual test we did, before sending the pull request:

**New check and times icons in a dropdown menu**
new icons

**Alerts**
Alerts

**Modals**
Modals

Updating `check` icon will enlarge the size of the icon even more so the check within the checkbox will be very encased. 

Thus, instead of using `check` for the checkbox component, we propose to use the new `check-small` icon:

Checkbox Tests

**Extra ball**
We received a lot of complains about the size of the times-icon in the past. So it would be good this change to reduce it a bit and allow product designers to create visual hierarchies using times for modals and panels and times-small for smaller things. 

cc\ @victorvalle @jbalsas


pat270:
This is good. I updated the checkbox to use `check-small`. Thanks @drakonux!

Issue #3527, opened 2020-07-20
Show only the custom color selector in ClayColorPicker React component


victorg1991:
## What is your proposal?
Right now, It is not possible to show a ClayColorPicker that only shows the custom color selector without the color palette above. 
Something like this:
![Screenshot 2020-07-20 at 14 50 21](https://user-images.githubusercontent.com/11654230/87939776-d8d68a80-ca98-11ea-82e7-ff1c2a6c4563.png)

## Why would adopting this proposal be beneficial?
More flexibility in the case like me color palette is not needed

## What are the alternatives to this proposal?
Using the useNative prop in the ClayColorPicker will show the default color picker which doesn't have color palette.. but is is not ideal.

bryceosterhaus:
@drakonux, does this fit within the lexicon definition? If so, I can add an ability to toggle that palette on/off.

jbalsas:
Hey @bryceosterhaus, I think this seems like a reasonable request that should come at a small price.

I think ColorPicker is complex enough so that low-level components won't do to implement it, so having props to pick which features to use from it makes sense to me.

That being said... @victorg1991, are you sure you don't want the palette? This makes sense out of context, but in the context of StyleEditor (if that's where you're applying it...) maybe it would be good to have the palette populated with the colours for the primary tokens... just a thought 😉

drakonux:
> @drakonux, does this fit within the lexicon definition? If so, I can add an ability to toggle that palette on/off.

The Lexicon team published the Color Picker time ago and it has a "mode": [custom color panel](https://liferay.design/lexicon/core-components/forms/picker-color/#custom-color-panel) and this is the [technical document](https://docs.google.com/document/d/1IUGl5VOWh6lqRa1baRoshVcytc8XbqVMkk30vRW4SM8/edit#heading=h.jkrp0t7cj15n) we currently have.  

Does it fit? It's quite similar, without the colors cell set and the feature to allow to change between panels.  But I would have to know more about the usage they are going to do, their requirements and review the definition we made.

victorg1991:
@jbalsas I talked with Juan about it and we decided it was better to have only a custom color picker without the palette. The idea behind is that we are going to be able to reference other variables so in case the user wants the primary color it will link the current variable to the --primary variable instead of using the concrete color value of primary (choosing the color from the palette) Does it makes sense? :)

Maybe in the future, the color picker could receive a list of CSS variables and allow the user to select the variable instead of the specific color

[...]

PR #3570, opened 2020-07-30
feat(@clayui/color-picker): add prop for disabling palette for custom colors

bryceosterhaus:
Hey @victorg1991, is this what you would expect for https://github.com/liferay/clay/issues/3527? Check out the storybook demo https://5f2322ebba586a00071c8a01--next-storybook-clayui.netlify.app/?path=/story/components-claycolorpicker--custom-w-o-palette

### DIFF
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -96,6 +96,11 @@ interface IProps {
 	onColorsChange: (val: Array<string>) => void;
 
 	/**
+	 * Flag for showing and disabling the palette of colors
+	 */
+	showPalette?: boolean;
+
+	/**
 	 * Path to the location of the spritemap resource.
 	 */
 	spritemap?: string;
@@ -109,11 +114,12 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 	label,
 	onChange,
 	onColorsChange,
+	showPalette,
 	spritemap,
 }) => {
 	const inputRef = React.useRef(null);
 	const [activeSplotchIndex, setActiveSplotchIndex] = React.useState(0);
-	const [editorActive, setEditorActive] = React.useState(false);
+	const [editorActive, setEditorActive] = React.useState(!showPalette);
 
 	const color = tinycolor(colors[activeSplotchIndex]);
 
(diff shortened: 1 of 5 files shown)


victorg1991:
@bryceosterhaus Yep that's exactly what we need 😍👏🏻 Thank you!!

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -106,14 +108,15 @@ interface IProps {
...
+	disablePalette,
 	label,
 	onChange,
 	onColorsChange,
 	spritemap,
 }) => {
 	const inputRef = React.useRef(null);
 	const [activeSplotchIndex, setActiveSplotchIndex] = React.useState(0);
-	const [editorActive, setEditorActive] = React.useState(false);
+	const [editorActive, setEditorActive] = React.useState(disablePalette);
```
matuzalemsteles, line 119: Maybe, you don't even need to use the `disablePalette` as an initial value for the state.

packages/clay-color-picker/src/index.tsx
```diff
@@ -73,6 +73,8 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	 */
 	disabled?: boolean;
 
+	disablePalette?: boolean;
+
```
matuzalemsteles, line 77: Just a description here of the API.

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -80,6 +80,8 @@ interface IProps {
 	 */
 	colors: Array<string>;
 
+	disablePalette?: boolean;
```
jbalsas, line 83: Could we try to use the positive version of this API instead of a negative one? I find it easier to reason with than having to use a double-negative.

There's some rationale for this in [API Names and Begging the Negatives](https://ariya.io/2016/04/api-names-and-begging-the-negatives) which is also covered as part of [Fluent 2014: Design Strategies for JavaScript API](https://www.youtube.com/watch?v=y7RrP_fOYUQ)

@wincent, have we covered this in the past? Tried to check for an "API" good practices in guidelines but couldn't find one...

wincent: > @wincent, have we covered this in the past?

It occasionally comes up in code review but I haven't seen it often enough to prompt me adding it to the guidelines, but maybe we should.

From ["Remove Double Negative"](https://refactoring.com/catalog/removeDoubleNegative.html):

> Double negatives are often frowned on by mavens of natural language ... double negatives are just plain confusing. So kill them on sight.

In my experience, _most_ of the time things read better if you avoid negation. The negation is fine until you need to negate it.... (eg. `!disablePalette`, which reads as "don't disable palette", which is a mouthful). In JS because we can easily have things like default params it is generally a good to favor the positive form, because you can make the dominant case (enabled) implicit and low friction:

```
function foo({enablePalette = true} = {}) {
  console.log(enablePalette);
}
```

bryceosterhaus: @jbalsas yeah that makes sense. I think the rationale for the `disable*` props probably stems from the `disabled` attr on html elements. I'm not tied to it though, so I'll go ahead and change it to a positive form and default the value to true.

On the naming, I don't think we typically use `enable` as a prefix anywhere. I think we would use `palette` or maybe `showPalette`(positive form 🤷‍♂️)? Any strong preferences for this?

wincent: No strong prefs but `showPalette` sounds good to me.

Alternatives considered by discarded:

- `letYourPaletteShow`
- `paletteLivesMatter`
- `paletteOrNotToPaletteThatIsTheQuestionAndIThinkYouKnowTheAnswer`

bryceosterhaus: > letYourPaletteShow

😂  😂  😂  this one got me good

jbalsas: 🤣🤣🤣

showMeThePALETTE!!

wincent: > showMeThePALETTE!!

sick-reference

packages/clay-color-picker/src/Custom.tsx
bryceosterhaus: We need to set the initial value since the custom interface is only showed after toggling the icon. If we disable the palette, we want to show the custom interface be default

Issue #3540, opened 2020-07-21
@clayui/color-picker iOS maybe all touch screens HSB Map doesn't update color value on click


pat270:
https://storybook.clayui.com/?path=/story/components-claycolorpicker--custom-colors

iOS 13 and maybe all touch screens

Try to create a custom color by tapping on the HSB Map or Hue. The color doesn't update.

Sidenote: Also unable to drag the pointers on touch device

PR #3572, opened 2020-07-31
fix(@clayui/color-picker): Tapping on HSB Map or Hue Range on a mobile device should update colors and be draggable

pat270:
I changed the mouse event to pointer event. This is supported by IE11, Safari, Chrome, Firefox, iOS Safari 13+ and Chrome for Android (https://caniuse.com/#feat=pointer). I can't figure out how to mimic pointer events using testing-library. I think this feature is missing from js-dom from the research I did.

issue #3540

### DIFF
--- a/packages/clay-color-picker/src/GradientSelector.tsx
+++ b/packages/clay-color-picker/src/GradientSelector.tsx
@@ -6,7 +6,7 @@
 import React from 'react';
 import tinycolor from 'tinycolor2';
 
-import {useMousePosition} from './hooks';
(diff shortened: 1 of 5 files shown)


bryceosterhaus:
I just looked into the tests a bit and it looks like jsdom doesn't support PointerEvent (https://github.com/jsdom/jsdom/issues/2527).

I'm gonna see if we can find a workaround for this though since these tests are particularly helpful

bryceosterhaus:
Added tests at https://github.com/liferay/clay/pull/3572

PR #3573, opened 2020-08-01
fix(clayui.com): Move `code-collapse.js` code to `CodeToggle` component? and update so it works in IE11

### LINKED ISSUE #3547: clayui.com code-collapse.js doesn't work on IE11

pat270:
fix(clayui.com): Adds `focusable="false"` to code collapse svgs so IE11 doesn't focus them on tab
fix(clayui.com): CSS Pages should have code collapsed by default
fix(clayui.com): Clean up code CSS

Not sure if I did this right, tried to copy the pattern as best I could.

fixes #3547


pat270:
@matuzalemsteles I'll take a look. Code copy alert was always wonky in some places. The alert doesn't show up in the same place between React and CSS pages. https://clayui.com/docs/components/alert.html CSS page 2nd and 3rd example doesn't show alert message on copy.

matuzalemsteles:
Hmm, I'm thinking, if we could remove the alert, it's a little weird maybe adding in an Alert Toast would be better. But we can do that in another PR.

pat270:
@matuzalemsteles the Code Copy should be fixed. I wanted to make the notification a toast, but probably better to wait for #3523.

PR #3582, opened 2020-08-04
fix(@clayui/modal): Changes the body's tabIndex with scroll and focuses on the element only when pressing the Up/Down keyboard

### LINKED ISSUE #3581: DropDown, Modal and Popover add tabIndex="0" to the body with scroll
Adding `tabIndex="0"` to body elements that contain scroll as in `<ClayModal />` adds the border when the element is focused and also adds the element in the order of focus. Apparently, this seems to be a problem for a `div` that is focusable and without context just to help the scrollable container to be scrollable using the Up/Down keyboard. This also happens for components with `<ClayDropDown />` and `<ClayPopover />`.

We can avoid adding `tabIndex="0"` in the component `<ClayDropDown />` since it will always have focusable elements and this will help to scroll the container. For cases like Modal and Popover we must deal with it in another way.

Ticket reference https://issues.liferay.com/browse/LPS-116326

matuzalemsteles:
Fixes #3581 

I took a different approach to Modal and maybe we can replicate for Popover, unfortunately, I still need `tabIndex` and I'm also using `c-inner` to remove the outline, as suggested by @pat270.

The case of Modal is that its content is not controllable, so there may be focusable elements or not, so it is still necessary to support that the container scroll is done via keyboard using Up/Down, so I using the strategy of when the keyboard of Up/Down is pressed to change the focus to the body, so it will be possible to scroll and remove it from the focus order and we have a behavior similar to the page scroll.

For Modal's case, this makes sense since when Modal is open it removes the scroll from the page but for Popover's case, we still need to deal with it better.

### DIFF
--- a/packages/clay-modal/src/Body.tsx
+++ b/packages/clay-modal/src/Body.tsx
@@ -4,7 +4,7 @@
  */
 
 import classNames from 'classnames';
-import React from 'react';
+import React, {useEffect, useRef} from 'react';
 
 export interface IBodyProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
@@ -28,16 +28,43 @@ const ClayModalBody: React.FunctionComponent<IBodyProps> = ({
 	iFrameProps = {},
 	scrollable,
 	url,
-}: IBodyProps) => (
-	<div
-		className={classNames('modal-body', {
-			'inline-scroller': scrollable,
-			'modal-body-iframe': url,
-		})}
-		tabIndex={scrollable ? 0 : undefined}
-	>
-		{url ? <iframe {...iFrameProps} src={url} title={url} /> : children}
-	</div>
-);
+}: IBodyProps) => {
+	const elementRef = useRef<HTMLDivElement | null>(null);
+
+	useEffect(() => {
+		const onKeyDown = (event: KeyboardEvent) => {
+			if (
+				elementRef.current &&
+				(event.key === 'ArrowUp' || event.key === 'ArrowDown') &&
+				!elementRef.current.contains(document.activeElement)
+			) {
+				if (event.defaultPrevented) {
+					return;
+				}
+
+				elementRef.current.focus();
+			}
+		};
+
+		document.addEventListener('keydown', onKeyDown);
+
+		return () => {
+			document.removeEventListener('keydown', onKeyDown);
+		};
+	}, [elementRef]);
+
+	return (
+		<div
+			className={classNames('modal-body', {
+				'inline-scroller c-inner': scrollable,
+				'modal-body-iframe': url,
+			})}
+			ref={elementRef}
+			tabIndex={scrollable ? -1 : undefined}
+		>
+			{url ? <iframe {...iFrameProps} src={url} title={url} /> : children}
+		</div>
+	);
+};
 
 export default ClayModalBody;
--- a/packages/clay-modal/src/Body.tsx
+++ b/packages/clay-modal/src/Body.tsx
@@ -56,7 +56,7 @@ const ClayModalBody: React.FunctionComponent<IBodyProps> = ({
 	return (
 		<div
 			className={classNames('modal-body', {
-				'inline-scroller c-inner': scrollable,
+				'inline-scroller': scrollable,
 				'modal-body-iframe': url,
 			})}
 			ref={elementRef}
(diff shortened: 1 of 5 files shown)


packages/clay-modal/src/Body.tsx
```diff
@@ -28,16 +28,43 @@ const ClayModalBody: React.FunctionComponent<IBodyProps> = ({
...
+
+		return () => {
+			document.removeEventListener('keydown', onKeyDown);
+		};
+	}, [elementRef]);
+
+	return (
+		<div
+			className={classNames('modal-body', {
+				'inline-scroller c-inner': scrollable,
```
pat270, line 59: The markup should go:
```
<div class="modal-body inline-scroller" tabindex="0">
  <div class="c-inner" tabindex="-1">
  </div>
</div>
```
`c-inner` is an overlay that highjacks the click so focus outline won't show on `modal-body`. This applies to `popover` and `dropdown` also. I'll need to add styles to Clay CSS since we are going this route.

jbalsas: > We still want to show the focus outline when someone tabs into modal-body. With tabindex="-1" a keyboard user can't tab into modal-body if they tab out of it.

I'd say that we do **NOT** want to show the focus outline. We want to allow tabbing inside or interacting somehow, although as I already expressed in https://github.com/liferay/clay/issues/3583, tabbing into a box sounds wrong (could be the only way, though).

jbalsas: One quick hack that comes to mind would be to add an invisible tabbable element inside inside the scroller. Something like:

```
<div tabindex="0"></div>
```

This would avoid the focus ring while making the content scrollable. 

That being said, I have no idea how users handle focus. I don't see an outer focus ring in pages, so it seems natural not to have it, but at the same time, how do users know what are they scrolling? 

I'm honestly just at a loss here 🤷

matuzalemsteles: @pat270 and @jbalsas yeah, in the case of Modal I tried to simulate the same behavior that happens on a page that has a scroll, so we don't need to have a focus ring on the `inline-scroll`, the user can scroll the content by pressing the Up/Down keyboard, this is a little bit different from the page 😬, but it can help.

See this example in the storybook https://deploy-preview-3582--next-storybook-clayui.netlify.app/?path=/story/components-claymodal--default with the scrollable knob enabled.

pat270: @matuzalemsteles I checked out the demo. I see how it works now. We don't need `c-inner`,  just use `tabindex="-1"`. I added a rule where outlines aren't shown when `tabindex` is negative.

matuzalemsteles: @pat270 I had to add `c-inner` exactly why is it appearing outline, maybe there was a regression?

pat270: Strange, it works fine on the storybook you linked. I see the rule there:
```
[tabindex^='-'] {
  outline: 0;
}
```

matuzalemsteles: Lol, now I see it too 🤷‍♂️. Probably doing something wrong. Okay, I'll go ahead and remove, I'll do the same for Popover.

matuzalemsteles: I think we have some problems in the development environment, for some reason I can't see the rule in the development environment. I'll check this out later.

pat270:
Modal is working well. I can't seem to get to the popover to scroll with keyboard at the https://deploy-preview-3582--next-storybook-clayui.netlify.app/?path=/story/components-claypopover--popover-w-trigger demo.

matuzalemsteles:
@pat270 I still don't know how to handle Popover's case, so it only works if you click, so I'm thinking of keeping `tabIndex="0"`, I don't know if the same fix could be replicated for Popover, already that when Popover is open the user can continue interacting with the rest of the screen.

pat270:
Can we do something similar to modal? Add `element.ref` to popover and `elementRef.current.focus();` on open?

We can also do `tabindex="0"` with nested `c-inner`, but I think I need to add some CSS to make the spacing correct.

matuzalemsteles:
We can do this for Popover, focus on content when opening instead of using `tabindex="0"`, I can continue with that.

matuzalemsteles:
@pat270 I updated the case for the popover, you can test it here https://5f2c0efae3077c00085afc74--next-storybook-clayui.netlify.app/?path=/story/components-claypopover--popover-w-manual-show

pat270:
Working well for me 👍

matuzalemsteles:
Right, I'm going to go ahead with this. @jbalsas if you have anything else about it, we can go back and revert something.

PR #3585, opened 2020-08-05
fix(@clayui/css): Alerts text in `alert-autofit-row` break to new lin…

### LINKED ISSUE #3584: @clayui/css: Alerts `.alert-autofit-row` doesn't offset gutters
This causes text to break prematurely that leaves unused space.
premature
should be:
alert-should-be

https://github.com/liferay-frontend/liferay-portal/pull/208#issuecomment-669104206

pat270:
…e prematurely

fixes #3584


marcoscv-work:
LGTM

PR #3587, opened 2020-08-05
fix(@clayui/data-provider): checks if the request body is read using request.bodyUsed

### LINKED ISSUE #3586: DataProvider does not read the request body in IE 11 without a custom fetch
Apparently, after we implemented the custom fetch feature, especially this line of code https://github.com/liferay/clay/blob/master/packages/clay-data-provider/src/useResource.ts#L196, DataProvider stops reading the request body even without a custom fetch in IE 11. This is because `res.body` and therefore `res.body.locked` do not exist in the request in IE 11, we use this to find out if the body has been read before reading it again.

matuzalemsteles:
Fixes #3586

### DIFF
--- a/packages/clay-data-provider/src/useResource.ts
+++ b/packages/clay-data-provider/src/useResource.ts
@@ -193,7 +193,7 @@ const useResource = ({
 
 				promise = fn(getUrlFormat(link, variables), fetchOptions).then(
 					(res: Response) => {
-						if (res.ok && res.body && !res.body.locked) {
+						if (res.ok && !res.bodyUsed) {
 							return res.json();
 						}
 


packages/clay-data-provider/src/useResource.ts
```diff
@@ -193,7 +193,7 @@ const useResource = ({
 
 				promise = fn(getUrlFormat(link, variables), fetchOptions).then(
 					(res: Response) => {
-						if (res.ok && res.body && !res.body.locked) {
+						if (res.ok && !res.bodyUsed) {
```
wincent, line 196: Strangely, MDN [claims IE doesn't support `bodyUsed`](https://developer.mozilla.org/en-US/docs/Web/API/Body/bodyUsed), but I guess you tested it? (They also label it as "Experimental. Expect behavior to change in the future.", but [they do that for `body` too](https://developer.mozilla.org/en-US/docs/Web/API/Body/body)). 🤷

matuzalemsteles: Hey @wincent yeah, I had also noticed this, am I guessing that there might be some polyfill on the portal? I only see the `bodyUsed` when I was testing it in IE 11.

bryceosterhaus:
Merging for release today

PR #3590, opened 2020-08-06
fix(@clayui/css): C Kbd each key in a keyboard shortcut should be wrapped in its own `kbd` element

### LINKED ISSUE #3283: C-Kbd each key in the sequence should be individually wrapped in `kbd`
The keyboard combination ⌘+Shift+Y is marked up like:
```
<kbd class="c-kbd">⌘+Shift+Y</kbd>
```

It should be marked up like:
```
<kbd class="c-kbd">⌘</kbd>+<kbd class="c-kbd">Shift</kbd>+<kbd class="c-kbd">Y</kbd>
```
or
```
<kbd class="c-kbd">
  <kbd class="c-kbd">⌘</kbd>+<kbd class="c-kbd">Shift</kbd>+<kbd class="c-kbd">Y</kbd>
</kbd>
```
to denote each key in the sequence. The `+` isn't part of the sequence.

pat270:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd#Representing_keystrokes_within_an_input is the reason for this update.

fixes #3283

### DIFF
--- a/clayui.com/content/docs/components/css-dropdown.md
+++ b/clayui.com/content/docs/components/css-dropdown.md
@@ -21,6 +21,7 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/dropdowns/'
         -   [Start](#start)
         -   [End](#end)
         -   [Start and End](#start-and-end)
+    -   [Keyboard Shortcuts](#keyboard-shortcuts)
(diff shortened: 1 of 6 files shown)


marcoscv-work:
LGTM

PR #3596, opened 2020-08-10
fix(@clayui/color-picker): fix native usage across browsers

### LINKED ISSUE #3524: @clayui/color-picker Why do we use a hidden input to store the color value when we already have an input in the input-group?
Continuing from unrelated pr #3519 and trying to get to the bottom of this. I think we are getting closer.

My next theory is that we need it for the native `input[type="color"]`. I see that the hidden input changes to input to `type="color"` at https://storybook.clayui.com/?path=/story/components-claycolorpicker--native.

If this is the reason, I think we could approach this in another way. We could replace the Splotch with `input[type="color"]` and and overwrite the default styles of the input to match Lexicon.

bryceosterhaus:
fixes #3524

This addresses several issues with native mode
1. Native shoudl now work across browsers
3. If browser is IE11, we use don't use type="color" since it is not supported

One change was for type="color", we were using hex values without the hash, however this wrong since the html specification says the value must include a hash.

### DIFF
--- a/packages/clay-color-picker/src/index.tsx
+++ b/packages/clay-color-picker/src/index.tsx
@@ -149,6 +149,14 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 	value = 'FFFFFF',
 	...otherProps
 }: IProps) => {
+	if (value.indexOf('#') === 0) {
+		value = value.slice(1);
+	}
+
+	if (useNative && window.navigator.userAgent.indexOf('MSIE') !== -1) {
+		useNative = false;
+	}
+
 	const triggerElementRef = React.useRef<HTMLDivElement>(null);
 	const dropdownContainerRef = React.useRef<HTMLDivElement>(null);
 	const inputRef = React.useRef<HTMLInputElement>(null);
@@ -174,9 +182,15 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 							useNative ? onValueChange(e.target.value) : null
 						}
 						ref={valueInputRef}
-						style={{display: 'none'}}
+						style={{
+							height: 0,
+							margin: 0,
+							padding: 0,
+							visibility: 'hidden',
+							width: 0,
+						}}
 						type={useNative ? 'color' : 'text'}
-						value={value}
+						value={`#${hexInputValue}`}
 					/>
 				)}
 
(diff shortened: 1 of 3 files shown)


packages/clay-color-picker/src/index.tsx
```diff
@@ -149,6 +149,14 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 	value = 'FFFFFF',
 	...otherProps
 }: IProps) => {
+	if (value.indexOf('#') === 0) {
+		value = value.slice(1);
+	}
+
+	if (useNative && window.navigator.userAgent.indexOf('MSIE') !== -1) {
```
pat270, line 156: Instead of sniffing for IE, wouldn't using feature detection be better? Something like:
```
var input = document.createElement('input');
input.setAttribute('type', 'color');
var inputColorSupport = input.value === '' || true;

if (inputColorSupport) {
    useNative = false;
}
```

bryceosterhaus: Good idea, just updated

packages/clay-color-picker/src/index.tsx
```diff
@@ -174,9 +193,15 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
...
+						style={{
+							height: 0,
+							margin: 0,
+							padding: 0,
+							visibility: 'hidden',
+							width: 0,
+						}}
 						type={useNative ? 'color' : 'text'}
-						value={value}
+						value={`#${hexInputValue}`}
```
matuzalemsteles, line 204: @bryceosterhaus wondering if with the addition of the hash we could break some behaviors that expected the hex value to come without the hash. I see this in DDM Field with ColorPicker but I'm not sure if it would be a problem for the backend to handle it.

matuzalemsteles: Apparently there seems to be no problem, I don't see any special validation for that.

bryceosterhaus: Yeah this is could technically break some uses, but we were using the wrong html spec anyways so its really a fix for us. I was going to go through liferay-portal and update all instances when I release

matuzalemsteles: Makes sense, no problem. LGTM.

PR #3602, opened 2020-08-11
fix(clayui.com): Random fixes

pat270:
- chore(clayui.com): Supress console warning `A page wasn't found for index.html`
- fix(clayui.com): `/docs/examples.html` links are prefixed with `https://clayui.com/`
- fix(clayui.com): A page wasn't found for "/docs" message in browser console (hover over Docs on top nav)
- fix(clayui.com): Blogs `Warning: ClayIcon requires a `spritemap` via prop or ClayIconSpriteContext` in console (`yarn develop`)
- fix(clayui.com): Removes `13:7  warning  'spritemap' is assigned a value but never used  no-unused-vars` from blog.js (`yarn develop`)

### DIFF
--- a/clayui.com/src/components/LayoutNav/LayoutNav.js
+++ b/clayui.com/src/components/LayoutNav/LayoutNav.js
@@ -15,7 +15,7 @@ const spritemap = '/images/icons/icons.svg';
 
 const DEV = process.env.GATSBY_CLAY_NIGHTLY === 'true';
 
-export default () => {
+export default (location) => {
 	const [showAtlas, setShowAtlas] = useStateWithLocalStorage(
 		true,
 		'clay.showAtlas'
@@ -25,6 +25,8 @@ export default () => {
 		true,
 		'clay.showSiteCss'
 	);
+
+	const pathname = location.pathname || '';
 
 	return (
 		<nav className="navbar navbar-clay-site navbar-expand-lg navbar-light">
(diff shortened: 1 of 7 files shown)


marcoscv-work:
LGTM !

clayui.com/src/components/LayoutNav/LayoutNav.js
```diff
@@ -49,27 +51,33 @@ export default () => {
...
+									color: pathname.match(/blog\//g)
+										? '#b3472d'
+										: '',
+								}}
 							>
 								<span className="c-inner" tabIndex="-1">
 									{'Blog'}
 								</span>
-							</Link>
+							</a>
```
matuzalemsteles, line 80: Hey @pat270 I think that replacing this with a href will have a redirection made by the browser instead of redirecting via JS, I think it is safe to keep it even with `console.log` because we make this rule here https://github.com/liferay/clay/blob/master/clayui.com/gatsby/createPages.js#L210-L215.

pat270: @matuzalemsteles I can change it back. This doesn't fix the issue. There are still warnings on the development site. It might be a bug in the gatsby version we are using. I tried upgrading to a newer version but kept running into build issues.

matuzalemsteles: @pat270 no problem, I also think it could be a bug, because we are creating a redirect for this route, in theory I should not give this warning. I think we can open an issue to update Gatsby to a newer version and resolve these compilation errors in another PR.

matuzalemsteles:
Okay, I'm merging this.

PR #3606, opened 2020-08-12
fix(@clayui.com): code collapse

bryceosterhaus:
Small updates for https://github.com/liferay/clay/pull/3523

Going to merge and if we need any other updates we can add as necessary, this won't be live un clayui.com till next release

### DIFF
--- a/clayui.com/gatsby-browser.js
+++ b/clayui.com/gatsby-browser.js
@@ -9,10 +9,16 @@ export const onClientEntry = () => {
 	if (process.env.GATSBY_CLAY_NIGHTLY === 'true') {
 		const isNullOrTrue = (val) => val === 'true' || val === null;
 
-		const showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
-		const showSiteCss = isNullOrTrue(
-			localStorage.getItem('clay.showSiteCss')
-		);
+		let showAtlas = true;
+		let showSiteCss = true;
+
+		try {
+			showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
+
+			showSiteCss = isNullOrTrue(
+				localStorage.getItem('clay.showSiteCss')
+			);
+		} catch {} // eslint-disable-line no-empty
 
 		if (showAtlas) {
 			require('@clayui/css/src/scss/atlas.scss');
--- a/clayui.com/src/components/Editor/index.js
+++ b/clayui.com/src/components/Editor/index.js
@@ -59,8 +59,22 @@ const Editor = ({
 	}
 
 	function useStickyState(defaultValue, key) {
+		const inBrowser = typeof window !== 'undefined';
+
+		const localStorage = {
+			getItem(key) {
+				return inBrowser ? window.localStorage.getItem(key) : null;
+			},
+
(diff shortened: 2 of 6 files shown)


clayui.com/src/components/Editor/index.js
```diff
@@ -12,7 +12,7 @@ import React, {useContext, useState} from 'react';
 import {LiveEditor, LiveError, LivePreview, LiveProvider} from 'react-live';
 
 import theme from '../../utils/react-live-theme';
-import useStickyState from '../Hooks/useStickyState';
+import useStateWithLocalStorage from '../Hooks/useStateWithLocalStorage';
```
bryceosterhaus, line 15: @kresimir-coko I renamed this just so we can be explicit as possible

clayui.com/src/components/Editor/index.js
```diff
@@ -104,8 +104,8 @@ const Editor = ({
 							displayType="unstyled"
 							small
 							spritemap={spritemap}
-							symbol={'paste'}
-							title={'Copy'}
+							symbol="paste"
```
bryceosterhaus, line 107: SF

clayui.com/src/components/Editor/index.js
```diff
@@ -131,15 +131,17 @@ const Editor = ({
...
+										}`,
 									}}
 									key={snippet.name}
 									onClick={() => {
-										if (activeIndex !== index) {
-											setCollapseCode(false);
-										} else {
-											setCollapseCode(!collapseCode);
-										}
+										setCollapseCode(
```
bryceosterhaus, line 140: cleaned up the logic a bit to use a ternary

clayui.com/gatsby-browser.js
```diff
@@ -7,12 +7,18 @@
...
+		let showAtlas = true;
+		let showSiteCss = true;
+
+		try {
+			showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
+
+			showSiteCss = isNullOrTrue(
+				localStorage.getItem('clay.showSiteCss')
+			);
+		} catch {}
```
bryceosterhaus, line 21: @kresimir-coko I think this should fix the localstorage issue when it is disabled

kresimir-coko:
Looks good @bryceosterhaus thanks for wrapping it up 👍

clayui.com/src/components/Editor/index.js
wincent: `<devils-advocate>`At the same time, you've leaked the implementation details into the name, arguably making it a leakier abstraction now, and the new name is less descriptive of the purpose and behavior of the hook (ie. before, the word "sticky" conveyed that the state somehow sticks around beyond the normal lifecycle of the component; after, the term "with local storage" conveys that local storage is used but it doesn't really say how/why).`</devils-advocate>`

I have no problem with the change, but I just wanted to point out that "explicit" is not always necessarily better than "abstract".

wincent: An example I should have mentioned was, what if, for some reason, we decided to switch to `sessionStorage` from `localStorage` in the future? Would you rename the hook to `useStateWithSessionStorage`?

PR #3608, opened 2020-08-12
fix(@clayui/form): Makes ClayInput's `type` property has a value `text` by default

### LINKED ISSUE #3604: ClayInput should add the attribute `type="text"` if no type is passed in
An `input` element without a specified type attribute defaults to `text`. We should add this because devs sometimes use the CSS selector `input[type="text"]` to style text inputs across the board.

diegonvs:
... and a boy-scouting allowing _"textarea"_ to be a possible value for [ClayInput's dynamic `component` property](https://github.com/liferay/clay/blob/master/packages/clay-form/src/Input.tsx#L162).

Closes #3604


pat270:
Stuffing `textarea` inside `ClayInput` doesn't make sense to me since the pattern right now is mapping to each kind of form element. `<ClayInput component="textarea">` is an outlier, we don't write any other form component this way. It should be `<ClayTextarea>` imo.

bryceosterhaus:
Yeah I agree with @pat270, I think its weird to have the textarea prop and would be better served if we just exported a textarea component, even if that component is an implementation of ClayInput

diegonvs:
Got it, I'll create the textarea related discussion on another PR and remove the related commits in this one.

diegonvs:
Forcing pushing removing the textarea commits in 3,2,1...

packages/clay-form/src/Input.tsx
```diff
@@ -198,6 +199,7 @@ const ClayInput = React.forwardRef<HTMLInputElement, IProps>(
 				['input-group-inset-before']: insetBefore,
 			})}
 			ref={ref}
+			type={Component === 'input' ? type : undefined}
```
matuzalemsteles, line 202: A little strange, why not just pass the `type`? it seems safe to just pass the type attribute.

diegonvs: It was a leftover from textarea commits, since a textarea doesn't have `type` attr. I'll remove.

PR #3615, opened 2020-08-17
feat(clayui.com): Add a utility function for Local Storage

### LINKED ISSUE #3611: We should have a centralized localStorage module
With increasing implementations using `localStorage`, we should have a module that eliminates the site breaking when working with Local Storage.
The module should be imported in the several places where we use `localStorage` and abstracts over the check/fallback required to make it work flawlessly.
Inspired by [this Slack thread](https://liferay.slack.com/archives/G3JBR21HA/p1597231455165800)

kresimir-coko:
Fixes #3611 

The goal of this factory function is to standardize the way we use `localStorage`, and even `sessionStorage` in a safe and consistent way.

### Steps Taken
1. I've thoroughly read and investigated the factory function from this [article](https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/) that @wincent linked in Slack a few days ago. I figured why reinvent the wheel and ported the function into our codebase.

2. I've adapted some small things to our ESLint rules but outside of that the code is the same. I'm unsure how we feel about "stealing" code from articles like this.

3. I tested this in-depth, making sure the site works on master and on my custom branch to make sure it works as intended considering I haven't written the implementation myself.

### Notes
- `localStorage` is used in `clay/packages/clay-css` in a couple files and I'm unsure if we can include the factory function in those files since the factory function is in `clay/clayui.com/

### DIFF
--- a/clayui.com/gatsby-browser.js
+++ b/clayui.com/gatsby-browser.js
@@ -3,22 +3,18 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import localStorage from './src/utils/localStorage';
+
 // A stub function is needed because gatsby won't load this file otherwise
 // (https://github.com/gatsbyjs/gatsby/issues/6759)
 export const onClientEntry = () => {
 	if (process.env.GATSBY_CLAY_NIGHTLY === 'true') {
 		const isNullOrTrue = (val) => val === 'true' || val === null;
 
-		let showAtlas = true;
-		let showSiteCss = true;
-
-		try {
-			showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
-
-			showSiteCss = isNullOrTrue(
-				localStorage.getItem('clay.showSiteCss')
-			);
-		} catch {} // eslint-disable-line no-empty
+		const showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
+		const showSiteCss = isNullOrTrue(
+			localStorage.getItem('clay.showSiteCss')
+		);
 
 		if (showAtlas) {
 			require('@clayui/css/src/scss/atlas.scss');
--- a/clayui.com/src/components/Hooks/useStateWithLocalStorage.js
+++ b/clayui.com/src/components/Hooks/useStateWithLocalStorage.js
@@ -5,45 +5,17 @@
 
 import React from 'react';
 
-const inBrowser = typeof window !== 'undefined';
-
-const localStorage = {
-	getItem(key) {
-		try {
-			return inBrowser ? window.localStorage.getItem(key) : null;
-		} catch {
-			return null;
-		}
-	},
-
-	setItem(key, value) {
-		if (inBrowser) {
-			try {
-				window.localStorage.setItem(key, value);
-			} catch {
-				return;
-			}
-		}
-	},
-};
+import localStorage from '../../utils/localStorage';
 
 function useStateWithLocalStorage(defaultValue, key) {
 	const [value, setValue] = React.useState(() => {
-		try {
-			const stickyValue = localStorage.getItem(key);
+		const stickyValue = localStorage.getItem(key);
 
-			return stickyValue === null
-				? defaultValue
-				: JSON.parse(stickyValue);
-		} catch {
-			return defaultValue;
-		}
+		return stickyValue === null ? defaultValue : JSON.parse(stickyValue);
 	});
 
 	React.useEffect(() => {
-		try {
-			localStorage.setItem(key, JSON.stringify(value));
-		} catch {} // eslint-disable-line no-empty
+		localStorage.setItem(key, JSON.stringify(value));
 	}, [key, value]);
 
 	return [value, setValue];
--- a/clayui.com/src/components/Hooks/useStateWithLocalStorage.js
+++ b/clayui.com/src/components/Hooks/useStateWithLocalStorage.js
@@ -9,9 +9,15 @@ import localStorage from '../../utils/localStorage';
 
 function useStateWithLocalStorage(defaultValue, key) {
 	const [value, setValue] = React.useState(() => {
-		const stickyValue = localStorage.getItem(key);
+		try {
+			const stickyValue = localStorage.getItem(key);
 
-		return stickyValue === null ? defaultValue : JSON.parse(stickyValue);
+			return stickyValue === null
+				? defaultValue
+				: JSON.parse(stickyValue);
+		} catch {
+			return defaultValue;
+		}
 	});
 
 	React.useEffect(() => {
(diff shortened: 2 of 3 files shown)


wincent:
> 2\. I've adapted some small things to our ESLint rules but outside of that the code is the same. I'm unsure how we feel about "stealing" code from articles like this.

The post is quite explicit:

```
/* ISC License (ISC). Copyright 2017 Michal Zalecki */
```

And somebody even asked him about it in the comments:

> Would you mind adding a licensing statement to the gist so that there won't be any legal issues with using your code?

To which he replied:

> I've updated the gist adding ISC license.

Here's the text of [the ISC licesse](https://en.wikipedia.org/wiki/ISC_license):

```
Copyright <YEAR> <OWNER>

Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is hereby granted, provided that the above copyright notice and this permission notice appear in all copies.

THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```

So, you can't use this code without attribution. You can freely take the idea though and implement something similar yourself (IMO much preferable to cluttering the repo with more licensing boilerplate just to re-use a snippet of code).

kresimir-coko:
> So, you can't use this code without attribution. You can freely take the idea though and implement something similar yourself (IMO much preferable to cluttering the repo with more licensing boilerplate just to re-use a snippet of code).

Wow, first time I encountered that, idk how to feel about that, I assumed the world was a better place, how naive I was.
I'll tinked with it tomorrow then and rewrite it, keeping the concept.

wincent:
> I'll tinked with it tomorrow then and rewrite it, keeping the concept.

Suggestions:

-   Don't call it a "factory"; this isn't Java.
-   Just call it `localStorage`, so you can:

    ```js
    import localStorage from '../utils/localStorage';
    

    // and use it like you would `window.localStorage`
    console.log(localStorage.getItem('yo'));
    ```
-   Don't implement all the methods on `localStorage`, just `getItem` and `setItem` is probably going to be enough, now and forever. (And if we ever do need another method like `clear()` etc, we can add it then... but we won't...)
-   Forget about the whole `getStorage` abstraction; it's flexibility we don't need now, and probably won't ever.
-   If the whole thing is much more than 20 lines (excluding copyright header), it probably has more than it needs.

diegonvs:
I'm not sure if this package would help you but We have an older localStorage handler on metal-plugins: https://github.com/metal/metal-plugins/blob/master/packages/metal-storage/src/Storage.js

May we could revamp this code? I don't know but I just left this idea here :joy:

kresimir-coko:
I've changed the utility to fit our needs better, tested it with Local Storage enabled and disabled on both development and production sites. Both are behaving as they're supposed to. One thing I'm not certain about is the structure of the function, I made it similar to how we make our components, defining low-level `getItem` and `setItem` methods and then combining them into a high-level `localStorage` function.

clayui.com/gatsby-browser.js
```diff
@@ -3,22 +3,18 @@
...
-			showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
-
-			showSiteCss = isNullOrTrue(
-				localStorage.getItem('clay.showSiteCss')
-			);
-		} catch {} // eslint-disable-line no-empty
+		const showAtlas = isNullOrTrue(localStorage.getItem('clay.showAtlas'));
+		const showSiteCss = isNullOrTrue(
+			localStorage.getItem('clay.showSiteCss')
+		);
```
wincent, line 17: Well this looks much nicer now.

clayui.com/src/components/Hooks/useStateWithLocalStorage.js
```diff
@@ -5,45 +5,17 @@
...
-			const stickyValue = localStorage.getItem(key);
+		const stickyValue = localStorage.getItem(key);
 
-			return stickyValue === null
-				? defaultValue
-				: JSON.parse(stickyValue);
-		} catch {
-			return defaultValue;
-		}
+		return stickyValue === null ? defaultValue : JSON.parse(stickyValue);
```
wincent, line 14: Minor nit: `JSON.parse` could `throw` if garbage ever got into localStorage (and you have to treat localStorage as "untrusted input" because a user can stick whatever they want in there), so you probably still want a `try`/`catch` here.

`</paranoia>`

clayui.com/src/utils/localStorage.tsx
```diff
@@ -0,0 +1,35 @@
...
+		window.localStorage.setItem(name, value);
+	}
+};
+
+const localStorage = {
+	getItem,
+	setItem,
+};
+
+export default localStorage;
```
wincent, line 35: Nice and simple.

kylo-approves

> One thing I'm not certain about is the structure of the function, I made it similar to how we make our components, defining low-level `getItem` and `setItem` methods and then combining them into a high-level `localStorage` function.

This seems fine to me:

- Use of `default` export makes it easy to consume.
- Rarely use `getItem` without a matching `setItem` partner, so no need to keep them separate (for tree-shaking).
- Exporting single object matches the API that you're wrapping (ie. `window.localStorage` is a single object too).

clayui.com/src/utils/localStorage.tsx
```diff
@@ -0,0 +1,35 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const isSupported = () => {
+	try {
+		window.localStorage.setItem('_TEST_KEY_', '_TEST_KEY_');
```
wincent, line 8: No `removeItem`?

kresimir-coko: Probably slipped through the cracks while simplifying. Gonna take a look in a moment!

kresimir-coko:
@wincent I've made some changes:
- Added the `try...catch` block to avoid the `JSON.parse` acting up
- Implemented your structure of the main `localStorage` function
- Brought back `removeItem`

kresimir-coko:
Hey @matuzalemsteles can you give this a look since Greg is vacationing

PR #3619, opened 2020-08-17
feat(@clayui/form): Adds Textarea component

diegonvs:
From https://github.com/liferay/clay/pull/3608 conversations:

I tried to use Storybook to see if all styles(combining via props) of ClayInput can be applied on the `ClayTextarea` component.

See the following storybook examples with our current input styles/props:

Textarea with Group Separated:
textarea_group_separated

Textarea with Group Connected:
textarea_group_connected

Textarea with Group Mixed:
textarea_group_mixed

I’m not sure if we support (textarea input groups)[https://liferay.design/lexicon/core-components/forms/text-input-group/]

#### Inset properties:

With insetAfter:

textarea_insetAfter

With insetBefore:

textarea_insetBefore

Apparently insetAfter and insetBefore classes works on the textarea component  

#### When auditing DXP:

I found we probably have an issue when autoFocus a textArea, that calling textAreaRef.focus() is required. (See https://github.com/liferay/liferay-portal/blob/a0397bf6adc1a9d6e60fb43579475241720f6526/modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/page_editor/common/components/Textarea.js)

**In this PR?** ❌ 

#### When auditing other component libraries I found some interesting. props that can be added:

`autoSize`: Height autosize feature, can be set to true\\|false or an object { minRows: 2, maxRows: 6 }

From rc-textarea(antd): https://react-component.github.io/textarea/?path=/story/rc-textarea--readme

Similar to https://github.com/liferay/liferay-portal/blob/ac2a395b4c1eee877498c87db48a1fe619e79cdc/portal-web/docroot/html/taglib/aui/input/page.jsp#L353 (It needs a AUI plugin to resize it, maybe We need to apply this logic of autoSize for all our input fields.

**In this PR?** ❌ 

----- 

`onResize`: Whether We need width/height values updated on the textarea when resizing

From rc-textarea(antd): https://react-component.github.io/textarea/?path=/story/rc-textarea--readme

**In this PR?** ❌ 

----- 

`rowsMax`, `rowsMin`: The maximum and minimum amount of rows to be displayed

From mui(materia-ui): https://material-ui.com/pt/api/textarea-autosize/#textareaautosize-api

**In this PR?** ❌ 

### Question:

What are you thoughts on these properties? Should we add it carefully when people need or should We be add it proactively?

### DIFF
--- a/packages/clay-form/src/Textarea.tsx
+++ b/packages/clay-form/src/Textarea.tsx
@@ -1,0 +1,57 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+interface ITextareaProps
+	extends React.TextareaHTMLAttributes<HTMLTextAreaElement> {
+	/**
+	 * Textarea component to render. Can either be a string like 'textarea' or a component.
+	 */
+	component?: 'textarea' | React.ForwardRefExoticComponent<any>;
+
+	/**
+	 * Flag to indicate if `input-group-inset-after` class should be applied
+	 */
+	insetAfter?: boolean;
+
+	/**
+	 * Flag to indicate if `input-group-inset-before` class should be applied
+	 */
+	insetBefore?: boolean;
+
+	/**
+	 * Selects the height of the input.
+	 */
+	sizing?: 'lg' | 'sm';
+}
+
+const ClayTextarea = React.forwardRef<HTMLTextAreaElement, ITextareaProps>(
+	(
+		{
+			className,
+			component: Component = 'textarea',
+			insetAfter,
+			insetBefore,
+			sizing,
+			...otherProps
+		}: ITextareaProps,
+		ref
+	) => (
+		<Component
+			{...otherProps}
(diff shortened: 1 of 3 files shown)


jbalsas:
> Regarding the comments on new APIs, I think we can study case by case and validate if it is worth adding this to the component, if there are more use cases of this elsewhere in DXP, we can add these new APIs in another PR, if it is necessary. First, we start small the same as we are doing with new components.

Yeah, we shouldn't rush to merge this or even work on it unless we have a clear need for it. Has any team asked for it? Has this been validated by Lexicon?

matuzalemsteles:
@jbalsas I think we can keep this on hold until we have a real case, as this is not a priority now. I don't think it came from any team.

diegonvs:
> Yeah, we shouldn't rush to merge this or even work on it unless we have a clear need for it. Has any team asked for it? Has this been validated by Lexicon?


No, It didn't came from any team and some new properties(that aren't included on this PR) needs to be validated with Lexicon.

This PR is just a expansion from a conversation in another PR and for preventing users to use directly the `<textarea` element and use the same API of ClayInput.

bryceosterhaus:
@diegonvs I'm gonna close this PR for now and if we decide we need textarea in the future or a team requests it, we can re-open this.

Thanks!

PR #3626, opened 2020-08-18
fix(@clayui/popover): focus on the body only when the scroll is enabled

### LINKED ISSUE #3625: ClayPopover focuses on the body when the element has no scroll
We recently added the change of a `<ClayPopover />` focus on the body when the element has scroll enabled so that it can scroll via the keyboard, but we focus on the body of the popover regardless of whether it has scroll or not.

For a use case for the App Builder team, they are using Popover with the disabled scroll that opens when an element hovers in DropDown, but this will close DropDown once the focus goes to the body.

- Codesandbox: https://codesandbox.io/s/priceless-pare-1ufhk?file=/src/App.js

Ideally, this should not be the solution to this problem, because it will still happen when Popover has a scroll and even if we remove this behavior it will happen when the user clicks on Popover. Perhaps this is a misuse of Popover because Popover is not part of DropDown, thinking that instead of using Popover it should be Tooltip.

cc @interaminense @kevenleone

matuzalemsteles:
Fixes #3625

### DIFF
--- a/packages/clay-popover/src/index.tsx
+++ b/packages/clay-popover/src/index.tsx
@@ -124,10 +124,10 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 		}, [show]);
 
 		useEffect(() => {
-			if (popoverScrollerRef.current && show) {
+			if (!disableScroll && popoverScrollerRef.current && show) {
 				popoverScrollerRef.current.focus();
 			}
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
Okay, I was hoping the codesandbox was working, but apparently it's been breaking for a few months. I will investigate this later.

PR #3629, opened 2020-08-19
fix(@clayui/nav): Fix the button that shows the vertical navigation when on mobile

### LINKED ISSUE #2821: Vertical Navigation: remove Select behavior
## Context

Reviewing [LEXI-688](https://issues.liferay.com/browse/LEXI-688) we found that in mobile version the Vertical Navigation component changes to a [dropdown button](https://liferay.design/lexicon/core-components/buttons/) but it behaves as it were a [Selector](https://liferay.design/lexicon/core-components/forms/selector/). 

image
_The dropdown shows the "five" item selected_

## What is your proposal?

The vertical navigation in mobile version should not show the current level of the navigation. It [should not behave like a Select](https://issues.liferay.com/browse/LEXI-773) component, it's just a button that allow users to toggle the menu. 

Thus, we should not add that Select feature to this component and we should allow to customize the text shown.

Talking with @marcoscv-work makes more sense to make this change for the sake of accessibility.   

## Concerns

Even more, the dropdown button should be customizable by product teams to use it with an icon, icon + caret, icon + text, text + caret or just with text (the most known examples are a menu button using the hamburger icon or just the word "menu")

kresimir-coko:
Fixes #2821 

### Steps taken

- Changed the element that opens the vertical nav from `<button>` to `<ClayButton>`
- Changed the label prop from `activeLabel` to `buttonTriggerLabel`, and set it's default value to `Menu`, also made it optional
- Updated docs, tests & stories, and added the Clay Button dependency


packages/clay-nav/src/Vertical.tsx
```diff
@@ -46,9 +47,9 @@ interface IItemWithItems extends IItem {
 
 interface IProps {
 	/**
-	 * Label of item that is currently active.
+	 * Label of the button that appears on smaller resolutions to open the vertical navigation.
 	 */
-	activeLabel: string;
+	buttonToggleLabel?: string;
```
matuzalemsteles, line 52: @kresimir-coko making this change will cause a breaking change since we would be deprecated the `activeLabel` API, so I think we can just keep `activeLabel`.

kresimir-coko: You're right, my bad!

kresimir-coko:
Hey @matuzalemsteles I've brought back the `activeLabel` prop and also included a `triggerContent` prop (and appropriate story) which should add flexibility to the users that was requested in the issue:
> Even more, the dropdown button should be customizable by product teams to use it with an icon, icon + caret, icon + text, text + caret or just with text (the most known examples are a menu button using the hamburger icon or just the word "menu")

kresimir-coko:
@matuzalemsteles 
> why switching to `buttonToggleLabel`, `activeLabel` doesn't make sense?

My idea was that the name doesn't make sense once we deprecate it. Users will expect it to represent the currently active navigation item, but the request from the issue was to have text that isn't tied to the currently active navigation item:
> The vertical navigation in mobile version should not show the current level of the navigation.

So we can deprecate `activeLabel` (I guess by adding a comment in the line above it), and add `buttonToggleLabel` which represent exactly what we want it to represent - the text inside the button that opens the Vertical navigation.

I'm working on the Trigger component right now, so I'll send something soonTM

PR #3642, opened 2020-08-22
fix(clayui.com): should be able to toggle between Atlas and Base theme on the production

### LINKED ISSUE #3621: We should be able to toggle between Atlas and Base theme on the production site
Users should be able to view the Base theme on the production site. IIRC it's not easy to extend the Classic Theme in Portal.

pat270:
I ended up adding a task to compile `atlas.scss` and `base.scss` into `static/css` so we can toggle both themes on production. I feel like there is a gatsby/webpack way, but I couldn't figure it out.

I realized we can't toggle between nightly and current release. `clay/node_modules/@clayui/css` gets updated when I make a change in `clay-css`. I was banking on that staying the same.

I also wanted to replace `GATSBY_CLAY_NIGHTLY` with `process.env.NODE_ENV === 'development'` but wasn't sure what https://github.com/liferay/clay/blob/master/clayui.com/gatsby/createPages.js#L30 did.

fixes #3621

### DIFF
--- a/clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
+++ b/clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
@@ -120,17 +120,20 @@ const generateFiles = (pluginOptions) => {
 // Compiles and writes `atlas.css` and `base.css` files to `/static/css/`
 
 const generateCSSFiles = (pluginOptions) => {
+	const cssDir = path.join(pluginOptions.staticDir, 'css');
+	const scssDir = path.join(pluginOptions.clay, 'scss');
+
 	fs.writeFileSync(
-		path.join(pluginOptions.staticDir, 'css', 'atlas.css'),
+		path.join(cssDir, 'atlas.css'),
 		compileSass({
-			file: path.join(pluginOptions.clay, 'scss', 'atlas.scss'),
+			file: path.join(scssDir, 'atlas.scss'),
 		}).css.toString()
 	);
 
 	fs.writeFileSync(
-		path.join(pluginOptions.staticDir, 'css', 'base.css'),
+		path.join(cssDir, 'base.css'),
 		compileSass({
-			file: path.join(pluginOptions.clay, 'scss', 'base.scss'),
+			file: path.join(scssDir, 'base.scss'),
 		}).css.toString()
 	);
 };
--- a/clayui.com/gatsby-browser.js
+++ b/clayui.com/gatsby-browser.js
@@ -8,25 +8,17 @@ import localStorage from './src/utils/localStorage';
 // A stub function is needed because gatsby won't load this file otherwise
 // (https://github.com/gatsbyjs/gatsby/issues/6759)
 export const onClientEntry = () => {
-	if (process.env.GATSBY_CLAY_NIGHTLY === 'true') {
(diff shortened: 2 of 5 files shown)


clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
```diff
@@ -105,33 +117,53 @@ const generateFiles = (pluginOptions) => {
 	);
 };
 
+// Compiles and writes `atlas.css` and `base.css` files to `/static/css/`
+
+const generateCSSFiles = (pluginOptions) => {
```
matuzalemsteles, line 122: @pat270 to have no problem in the production build, we need to make sure to create the file `/static/css/`, because we are considering that it exists below. Instead of doing some node checking, we can create the css folder and add an empty file called `.keep`, which is a common pattern in repositories. For this to work we would have to improve the ignore of `static/css`.

pat270:
> Do you say to compile sass files with webpack on Gatsby? I think we can use gatsby/onCreateWebpackConfig.js but I'm not sure if it is always like a watch environment or just used in production mode.

@matuzalemsteles From what I gathered, gatsby takes all the scss files on the page and combines them into a single file. I need two global css files where I can toggle between the two.

clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
pat270: I just pushed a change doing some node checking. I'll change that to this.

matuzalemsteles: @pat270 it's okay to check this with the node as well.

pat270:
> This line in particular makes documents with nightly appear only when GATSBY_CLAY_NIGHTLY is true, changing this to process.env.NODE_ENV === 'development' I think it would not be a problem, I think it would have the same behavior.

This is a nitpick. `GATSBY_CLAY_NIGHTLY` leads me to believe we can view the current site on different versions of Clay CSS. It's not possible right now since everything is based off Clay CSS source. We can come back to this if there is a need for it.

matuzalemsteles:
> @matuzalemsteles From what I gathered, gatsby takes all the scss files on the page and combines them into a single file. I need two global css files where I can toggle between the two.

Hmm, I think the best thing would be to keep it in the plugin. This can be more problematic in making it work with the webpack.

> This is a nitpick. GATSBY_CLAY_NIGHTLY leads me to believe we can view the current site on different versions of Clay CSS. It's not possible right now since everything is based off Clay CSS source. We can come back to this if there is a need for it.

Ideally, everything around the GATSBY_CLAY_NIGHTLY flag is just so that certain behaviors, pages appear only in that environment, this does not apply to CSS because we were not making any additional rules, as you were removing the conditional of showing the toggle to show the theme between Atlas or Base, this only appeared for the nightly environment, this environment was just for you to work on but it wouldn't show it to us for example, since our work on clayui.com is usually to update documents. We can remove this if you wish and just have an environment that would be a development environment.

pat270:
@matuzalemsteles we can merge this. I'm still writing out the overview of Clay CSS. I'm going to place it under CSS Framework. I'll send a separate pr for that.

matuzalemsteles:
@pat270 for a moment I had forgotten that. Right, I'll merge this.

PR #3644, opened 2020-08-24
feat(@clayui/modal): Add zIndex prop 

### LINKED ISSUE #3643: ClayModal does not support zIndex
Currently, there's no way to open a `ClayModal` in a different zIndex. Right now, we're hardcoding `.modal { z-index: $zindex-modal}` in [`modals.scss`](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_modals.scss#L10-L20) and `.modal-backdrop { z-index: $zindex-modal-backdrop }` in [`modals.scss`](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_modals.scss#L145-L161).

These values are by default:
- Modal backdrop: `1040`
- Modal window: `1050`

The problem is that these values are hardcoded at build time and can't change, so if there's any element with a higher `zindex` than that, modals will show underneath. 

Check https://codesandbox.io/s/fervent-gates-v61vs for an example of the issue where a different element will block both the modal backdrop and the modal window.

## What is your proposal?
Add a `zIndex` property to `ClayModal` that can overtake the default values with custom `z-index` styles

- It would allow developers to show modals at their z-index of choice
- This is a blocker for [Allow "Browse Server" for links and images in CKEditor](https://issues.liferay.com/browse/LPS-119207) since CKEditor will be showing a Modal on `1100` that hides `ClayModal`.

- I don't see any other possibility, really...

kresimir-coko:
Fixes #3643 

Modal window's zIndex is set to the value provided by the user, and the backdrop value is decreased by 10 to maintain the connection between the 2 elements. I've tested this in storybook to make sure both styles are applying properly and overriding the default ones.

### DIFF
--- a/packages/clay-modal/src/Modal.tsx
+++ b/packages/clay-modal/src/Modal.tsx
@@ -41,6 +41,11 @@ interface IProps
 	 * hook, adds observer from `useModal` hook here.
 	 */
 	observer: Observer;
+
+	/**
+	 * Allows setting a custom z-index value, overriding the default one which is 1040, modal body z-index will be +10 of this value
+	 */
+	zIndex?: number;
 }
 
 const warningMessage = `You need to pass the 'observer' prop to ClayModal for everything to work fine, use the 'useModal' hook that exposes the observer.
@@ -62,6 +67,7 @@ const ClayModal: React.FunctionComponent<IProps> = ({
 	size,
 	spritemap,
 	status,
+	zIndex,
 	...otherProps
 }: IProps) => {
 	const modalElementRef = useRef<HTMLDivElement | null>(null);
@@ -84,6 +90,7 @@ const ClayModal: React.FunctionComponent<IProps> = ({
 				className={classNames('modal-backdrop fade', {
 					show,
 				})}
+				style={{zIndex}}
 			/>
 			<div
 				{...otherProps}
@@ -91,6 +98,7 @@ const ClayModal: React.FunctionComponent<IProps> = ({
 					show,
 				})}
 				ref={modalElementRef}
+				style={{zIndex: zIndex && zIndex + 10}}
 			>
 				<div
 					className={classNames('modal-dialog', {


packages/clay-modal/src/Modal.tsx
```diff
@@ -41,6 +41,11 @@ interface IProps
 	 * hook, adds observer from `useModal` hook here.
 	 */
 	observer: Observer;
+
+	/**
+	 * Allows setting a custom z-index value, overriding the default one which is 1050, backdrop z-index will be -10 of this value
+	 */
+	zIndex?: number;
```
matuzalemsteles, line 48: The only thing I would change here would be to invert, instead of decreasing -10 of the backdrop, add the chosen value on the backdrop and in `.fade.modal` add 10 more, this seems better. For example, if the dev wants a z-index greater than 1200 and he sets it to 1205, it is just a hypothesis but it can happen with a certain frequency and it can be a subtle problem to perceive. I think it is safer to increase than decrease for this case.

kresimir-coko: Sure, but I did add in the JSDoc that the backdrop is decreased by 10, so if something isn't working for the user they will go to the API table, right? I don't care much, just it seems that the Modal Body is the primary element in the component, so the backdrop should follow the body, rather than body following the backdrop.

jbalsas: We discussed this as a possible issue with an easy workaround... but it's true that on the cost of "correctness" we get a more robust solution as @matuzalemsteles points out. I'm not opposed it :)

kresimir-coko: Alright, gonna adapt it and send! 👍

jbalsas: You guys can keep talking about it, don't take my 2cts as definitive answer! :)

matuzalemsteles: > Sure, but I did add in the JSDoc that the backdrop is decreased by 10, so if something isn't working for the user they will go to the API table, right?

It's usually what we expect 🙂

> just it seems that the Modal Body is the primary element in the component, so the backdrop should follow the body, rather than body following the backdrop.

Depending on the perspective yes, maybe for those who are using it on the outside without knowing the component inside, it would not matter to him but it would be strange if the backdrop ends up falling below. I say this because we are applying a rule of having to increase above 10 the target amount to be safe.

kresimir-coko: It makes sense now that it the information settled in my head a bit, 
The core being:
- it doesn't hurt having it increase, 
- but it might hurt having it decrease

So we wanna avoid the `might hurt` part, and follow the `doesn't hurt` part

kresimir-coko: @matuzalemsteles I've force pushed the same commit but with new code, didn't wanna make a new commit since it's so simple and to avoid confusion later when looking at the changes some time from now

matuzalemsteles: Makes sense, thanks @kresimir-coko! I'll just wait for the CI for merge. @jbalsas is this urgent for this DXP release? I can release this package individually along with https://github.com/liferay/clay/pull/3639 and add it to the PR that is under review to update clay on the portal.

jbalsas: Hey @matuzalemsteles, yeah, as I wrote in https://github.com/liferay/clay/issues/3643:

> This is a blocker for [Allow "Browse Server" for links and images in CKEditor](https://issues.liferay.com/browse/LPS-119207) since CKEditor will be showing a Modal on `1100` that hides `ClayModal`.

Let's not add conflate the other PR already which is already seemingly stuck. Release the package and send a PR with the `LPS-119207` number so I can add the rest of the fix on top.

matuzalemsteles: 👍

PR #3646, opened 2020-08-24
feat(@clayui/toolbar): Deprecated ClayUpperToolbar and adds examples with ClayToolbar

matuzalemsteles:
From #3528

@kresimir-coko I just added a warning to the ClayUpperToolbar document. I realized that the ticket for this is closed, we must create another issue or reopen, I'm not sure if it was closed by mistake.

### DIFF
--- a/packages/clay-css/src/scss/variables/_tbar.scss
+++ b/packages/clay-css/src/scss/variables/_tbar.scss
@@ -1,4 +1,4 @@
-$tbar-item-padding-x: 0.25rem !default;
+$tbar-item-padding-x: 0.5rem !default;
 $tbar-item-padding-y: null !default;
 
 // Tbar Stacked
--- a/clayui.com/gatsby/onCreateNode.js
+++ b/clayui.com/gatsby/onCreateNode.js
@@ -60,7 +60,7 @@ module.exports = exports.onCreateNode = ({actions, getNode, node}) => {
 				pkgVersion.includes(subs)
 			);
 
-			pkgStatus = isBeta ? 'Beta' : 'Stable';
+			pkgStatus = pkgStatus ? pkgStatus : isBeta ? 'Beta' : 'Stable';
 		}
 
 		if (!slug) {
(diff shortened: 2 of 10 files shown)


matuzalemsteles:
@bryceosterhaus @kresimir-coko as part of your older conversations I think it's closer now to discontinuing the UpperToolbar, but I think we still need to do a few things before we go:

- Remove the draft from the Toolbar document
- Releases a new version of the Toolbar
- Deprecated UpperToolbar package at npm
- Check if there are cases in the portal with UpperToolbar and replace with Toolbar

@bryceosterhaus I will wait on your thoughts before merging this, it is safe to go ahead and continue iterate on the master.

bryceosterhaus:
@matuzalemsteles I think you are right on with the list of things that need to be done. We can remove draft and release tomorrow and start working on the other tasks in that list

packages/clay-css/src/scss/variables/_tbar.scss
```diff
@@ -1,4 +1,4 @@
-$tbar-item-padding-x: 0.25rem !default;
+$tbar-item-padding-x: 0.5rem !default;
```
bryceosterhaus, line 1: Where is this change from?

kresimir-coko: This is because the current padding is 4px, but on Lexicon it is 8px, @pat270 commented on this saying it should be fine: [comment](https://github.com/liferay/clay/pull/3528#issuecomment-675236246)

kresimir-coko:
Hey @matuzalemsteles I've pushed a commit primarily consisting of removing draft from the Toolbar, I do not know how to do the other steps but I'll be taking a close eye on what you and/or @bryceosterhaus do with it.

matuzalemsteles:
Thanks @kresimir-coko! I'm thinking, if we could migrate these components to use this new implementation, but we would have to see how much work there is on the portal, so that it doesn't get expensive.

> I think you are right on with the list of things that need to be done. We can remove draft and release tomorrow and start working on the other tasks in that list

Great, maybe the release, we can hold on until after the 7.3 release, it's very close! I may have forgotten something, so comment on it if something comes to mind. So I will continue with this PR and merge!

PR #3654, opened 2020-08-26
fix(clayui.com): Fix React and Markup tabs not being linkable

### LINKED ISSUE #2654: [clayui.com] Make the second tab (CSS/Markup) of the components linkable
Make the second tab (CSS/Markup) of the components linkable

image

At the moment I can not link a CSS/Markup page

kresimir-coko:
Fixes #2654

### DIFF
--- a/clayui.com/src/templates/docs.js
+++ b/clayui.com/src/templates/docs.js
@@ -176,13 +176,21 @@ export default (props) => {
 		location.hash === '#css' || !containsReact ? CSS_INDEX : REACT_INDEX
 	);
 
+	const [hasMounted, setHasMounted] = React.useState(false);
+
 	useEffect(() => {
+		setHasMounted(true);
+
 		document
 			.querySelectorAll('.clay-site-custom-checkbox-indeterminate')
 			.forEach((item) => {
 				item.indeterminate = true;
 			});
 	}, []);
+
+	if (!hasMounted) {
+		return null;
+	}
 
 	const reactTab = body || tab.body;
 


bryceosterhaus:
@kresimir-coko I'm not seeing this bug as reproducible locally in a dev environment. It might have something to do with the statically generated html that gatsby creates that doesn't happen on the dev server, so you might look into testing it with `yarn build && yarn serve`

kresimir-coko:
As I'm trying to figure this out, I might have encountered a potential roadblock: [location search or hash update cannot trigger page re-render](https://github.com/gatsbyjs/gatsby/issues/3339).

What happens is that the `activeIndex` does set to the proper value, but only after it get's reset to 0, like this: https://i.imgur.com/Y5Sldsf.mp4, do note the hash, the current active index and the selected tab while I'm refreshing the page in the video.

kresimir-coko:
I've figured out a fix, it entails having a `hasMounted` state that dictates whether the app will render the proper content, as it was rendering old the initial (`activeIndex` of value 0), regardless of the hash, as can be seen in the video linked in the comment above. What this does is it returns `null` if that case mentioned above is triggered and then it sets the `hasMounted` to true which renders the proper content.

I pushed a new commit, which overrides the old one, as the old one didn't make sense anymore.

clayui.com/src/templates/docs.js
```diff
@@ -176,14 +176,22 @@ export default (props) => {
 		location.hash === '#css' || !containsReact ? CSS_INDEX : REACT_INDEX
 	);
 
+	const [hasMounted, setHasMounted] = React.useState(false);
+
 	useEffect(() => {
+		setHasMounted(true);
+
```
matuzalemsteles, line 183: Ideally, a `useEffect` should deal with only one responsibility, it would be good to add it to another `useEffect` since it deals with something else.

clayui.com/src/templates/docs.js
```diff
@@ -176,14 +176,22 @@ export default (props) => {
 		location.hash === '#css' || !containsReact ? CSS_INDEX : REACT_INDEX
 	);
 
+	const [hasMounted, setHasMounted] = React.useState(false);
+
 	useEffect(() => {
+		setHasMounted(true);
```
wincent, line 182: As I am sure you know, it's generally an anti-pattern to call a `useState` setter from inside a `useEffect`, because you are provoking a second render. Are you _sure_ you need this? It looks to me like every single load is going to be doubled because of this.

I pull down your PR and add a `console.log(location.hash)` in the body here. I see:

1.  Load: http://localhost:8000/docs/components/alert.html
    - See it log `''` twice (ie. two renders).
    - See main tab load.
2.  Hit Command-R to reload.
    - See it log `''` twice (ie. two renders).
    - See main tab load.
    - Address bar obviously still the same.
3.  Click on CSS.
    - See it log `''` then `'#css'` (ie. two renders).
    - See CSS tab load, and URL bar shows: http://localhost:8000/docs/components/alert.html#css
4.  Hit Command-R to reload.
    - See it log `'#css'` twice (ie. two renders).
    - See CSS tab load.
    - Address bar obviously still the same.

When I revert your PR I see reloading still working.

I am I seeing something different because I'm looking at this in the hot-reloading environment?

matuzalemsteles: @wincent it looks like this works only in the development environment without the changes to that PR, but it doesn't work correctly when the build is done, try testing with `yarn build && yarn serve` or using `clayui.com`.

wincent: Is it possible that this has nothing to do with Gatsby and something to do with `clay-tab`?

Given a production build, without this PR, if I add this logging:

```js
console.log(location.hash, activeIndex);
console.log(data);
```

I see `location.hash` of `#css`, `activeIndex` of `1`, and I see the CSS HTML in `data.mdTab.html` (which gets assigned to `cssTab`). In other words, everything that is necessary to render the CSS tab correctly on the first render is present, but for some reason, `ClayTab` is still rendering the first tab as active.

matuzalemsteles: I'll check that out.

matuzalemsteles: It really is strange behavior, I think it is not related to the `ClayTab` component, as Gatsby does a build-time rendering to generate the static page, it seems that when you open the page it doesn’t cause a rendering because I can see the components with the fixes indexes and classes but are not reflected in the DOM. This is very strange, I will continue investigating tomorrow.

kresimir-coko: Alright @matuzalemsteles I'll let you investigate then, I'm excited to see what you find out. This issue has been like a sore thumb ever since I joined the team, I wanted to fix it day 1 but didn't know the intricacies of Gatsby.

kresimir-coko: Hey @matuzalemsteles did the investigation yield any results?

matuzalemsteles: @kresimir-coko I went back to it yesterday in the late afternoon, still not getting anywhere, just with assumptions, I will try not to use ClayTabs to check if it is not a problem with himself because Gatsby uses `ReactDOM.hydrate` to hydrate the container that was rendered before, I think that in the first rendering it should differentiate the tree from elements and cause the changes, but I'm not sure if this is true or is it really a bug here.

matuzalemsteles: Ok, I think I understood better the case of ReactDOM.hydrate, this is related to it, this will clarify better https://github.com/facebook/react/issues/13260#issuecomment-407686205.

In summary https://reactjs.org/docs/react-dom.html#hydrate:
> React expects that the rendered content is identical between the server and the client. It can patch up differences in text content, but you should treat mismatches as bugs and fix them. In development mode, React warns about mismatches during hydration. There are no guarantees that attribute differences will be patched up in case of mismatches. This is important for performance reasons because in most apps, mismatches are rare, and so validating all markup would be prohibitively expensive.

This is exactly our case, we changed the tab using attributes, specifically `className`, as `ReactDOM.hydrate` is used to hydrate the container, it will not change attributes, this is its standard behavior, this explains why the case of @kresimir-coko works above, in theory, we just have to cause a new rendering so that reconciliation can have the expected effects. We can continue with this implementation or this:

```js
const initialActiveIndex = !containsReact ? CSS_INDEX : REACT_INDEX;

const [activeIndex, setActiveIndex] = React.useState(initialActiveIndex);

useEffect(() => {
	setActiveIndex(location.hash === '#css' || initialActiveIndex);
}, []);
```

The downside of this is that we will see the transition from the home screen to the other, in the case implemented in this PR, as we will render the home page without content, it will not have this transition effect.

Although I don't like the transition from one page to another I prefer to go with it, because the user will not always start to see the CSS page and this is good because we will have the page already rendered at build time, returns a `null` will just send several pages without content and will be rendered at run time, and we can lose the benefit of SSR here.

wincent: This all seems a bit smelly to me and makes me think we might be doing this wrong. You shouldn't have to render twice to render correctly once. Have we thought about [the issue](https://github.com/gatsbyjs/gatsby/issues/3339) that @kresimir-coko originally linked?

> This is deliberate. Gatsby is a static site generator which means we need to be able serve static page versions of URLs. There's no way to (universally) serve different files for various locations. So either use pathnames

ie. We could render two pages:

- http://localhost:8000/docs/components/alert.html (main tab active)
- http://localhost:8000/docs/components/alert/css.html (other tab active)

Gatsby doesn't care whether we tell it to render 200 or 2,000 pages; the idea of prerendering them to static markup is that they be fast.

matuzalemsteles: > This all seems a bit smelly to me and makes me think we might be doing this wrong. You shouldn't have to render twice to render correctly once. Have we thought about the issue that @kresimir-coko originally linked?

Yeah, I think they are related exactly because at build time it is rendered using `ReactDOM.renderToString` and then `ReactDOM.hydrate` is used, according to the documentation he expects everything to be the same when it will hydrate and if there is any difference it would be considered a "bug".

Yes, I do not see any problem in rendering more pages, Gatsby should be able to handle twice as many pages here, it can be a good solution to go with that, we would probably have to change all our current rule, maybe it will simplify things too. What do you think about going with that? @bryceosterhaus @kresimir-coko

bryceosterhaus: > ie. We could render two pages:
> http://localhost:8000/docs/components/alert.html (main tab active)
> http://localhost:8000/docs/components/alert/css.html (other tab active)

Yeah I think this is the best route to go and would've been my preference initially but just took more time to implement. My initial work with the hash was only done because it was the simplest way of approaching it, although I wasn't aware that it didn't work in the prod build. So I'm all for adding a route per tab

matuzalemsteles: Perfect, I will try to send a draft today so that we can evaluate the implementations. @kresimir-coko this will probably affect your work with #3477, but I will try to make things more friendly and you probably won't need those implementations anymore to create more than one tab.

matuzalemsteles: Sent #3679

kresimir-coko: Alright, big brain talk here, I love it, I actually understood all of it 👍 

@matuzalemsteles As for affecting #3477 that's fine, it's still being worked on, and I'm not racing against time so if it takes longer that's alright with me.

matuzalemsteles:
@kresimir-coko I will close this PR since we merged #3679.

PR #3655, opened 2020-08-26
fix(@clayui/color-picker): maintains the value entered by the user when the color is not valid

### LINKED ISSUE #3651: @clayui/color-picker Can't clear the input once a color is selected
https://storybook.clayui.com/?path=/story/components-claycolorpicker--custom-colors

There are a few issues:

1) Try to clear the color, tabbing out always causes the color to return.
2) If there was no initial value, color picker always has `#` in the value. This is an issue in Portal under Portlet Look and Feel Configuration, changing border top color causes the other border colors to output `#`.
3) Can't use `transparent` or other css color names like `red`, `blue`, `green`.

matuzalemsteles:
Fixes #3651

This fixes an issue in issue #3651 but we still need to understand the other problem that I cannot reproduce. Marking this as a draft initially to correct the other problems, still ignoring the case of the ColorPicker accepting `transparent`, `red`, `black`... as colors.

### DIFF
--- a/packages/clay-color-picker/src/index.tsx
+++ b/packages/clay-color-picker/src/index.tsx
@@ -201,7 +201,7 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 							width: 0,
 						}}
 						type={useNative ? 'color' : 'text'}
-						value={`#${hexInputValue}`}
+						value={hexInputValue ? `#${hexInputValue}` : ''}
 					/>
 				)}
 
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
> If there was no initial value, color picker always has # in the value. This is an issue in Portal under Portlet Look and Feel Configuration, changing border top color causes the other border colors to output #.

I think we just need to update the logic for the hidden input to be
```jsx
value={hexInputValue ? `#${hexInputValue}`: ''}
```
And then the 3rd issue pat mentioned would likely be a new feature and a separate ticket

matuzalemsteles:
> I think we just need to update the logic for the hidden input to be

@bryceosterhaus Yeah, I'm wondering if that was really the problem. I didn't understand it very well, but we can do just that. @pat270 is that right? Sorry, I'm still confused.

pat270:
@matuzalemsteles yes that is the fix. I'm not sure if portal will still output `border-left-color:;` but we can't fix that in Clay.

bryceosterhaus:
👍

PR #3657, opened 2020-08-26
fix(@clayui/shared): fixes side effect of focusing on the first element when the active element does not exist in the scope

### LINKED ISSUE #3656: The focus manager is stuck when the focused active element is not in scope
We recently made some improvements in FocusManager to not deal with focus when the next/previous element does not exist in scope #3334. This is very important for when we render the loose component in one part of the application and the other elements are not in the React tree.

In the ticket https://issues.liferay.com/browse/LPS-119771 a problem was found in IE 11 that the focus is still stuck by the focus manager, analyzing the problem better because of the svg element inside the button that receives the focus (this only happens in IE 11 https://stackoverflow.com/questions/29393144/how-to-prevent-svg-elements-from-gaining-focus-with-tabs-in-ie11) right after focusing on the element button that the focus manager controls but does not control the svg focus on, but `focusManager.focusNext()` is called again because it is inside the FocusScope tree.

We can fix this by adding `focusable="false"` to svg for the Portal case and we can ignore these cases on the focus manager side, when he does not find the active element within the scope he goes back to the first element, this is a side effect that shouldn't happen.

matuzalemsteles:
Fixes #3656

This is a very specific case that can happen in IE 11 when there are svg elements. We will probably have to make a release and send about the ticket https://issues.liferay.com/browse/LPS-119771.

### DIFF
--- a/packages/clay-shared/src/useFocusManagement.ts
+++ b/packages/clay-shared/src/useFocusManagement.ts
@@ -208,6 +208,11 @@ export function useFocusManagement(scope: React.RefObject<null | HTMLElement>) {
 		const lastPosition = elements.length - 1;
 		let nextElement = null;
 
+		// Ignore when the active element is not in the scope.
+		if (position < 0) {
+			return null;
+		}
+
 		if (backwards) {
 			if (position === 0) {
 				// is out of scope, so we go back through the fiber to pick up the


bryceosterhaus:
LGTM!

PR #3673, opened 2020-08-31
feat(@clayui/multi-select): improves the filter for case-sensitive or non-sequential characters

### LINKED ISSUE #3672: The default MultiSelect filter does not handle case-sensitive or non-sequential characters very well
Because the filter uses `'two'.match('t')`, it does not deal with case-sensitive and non-sequential characters, this greatly reduces the filter experience, despite providing an API that the dev can customize your filter, we can provide something on Clay’s side that can handle these common cases well.

- Ticket: https://issues.liferay.com/browse/LPS-120103

matuzalemsteles:
Fixes #3672

### DIFF
--- a/packages/clay-multi-select/package.json
+++ b/packages/clay-multi-select/package.json
@@ -32,7 +32,8 @@
 		"@clayui/icon": "^3.0.5",
 		"@clayui/label": "^3.3.1",
 		"@clayui/shared": "^3.2.3",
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
Going ahead with this to be able to release this package and send it today to the Portal about the ticket https://issues.liferay.com/browse/LPS-120103

PR #3676, opened 2020-08-31
fix(clayui.com): Linking to CSS pages doesn't take you to the correct place

pat270:
This fixes an issue where linking to https://clayui.com/docs/components/input.html#textarea, which is in the CSS/Markup section doesn't take you there.

### DIFF
--- a/clayui.com/package.json
+++ b/clayui.com/package.json
@@ -45,6 +45,7 @@
 		"@mdx-js/mdx": "^1.6.16",
 		"@mdx-js/react": "^1.6.16",
 		"@mdx-js/tag": "^0.20.3",
+		"@reach/router": "1.3.4",
 		"classnames": "^2.2.6",
 		"clipboard": "^2.0.4",
 		"dotenv": "^6.0.0",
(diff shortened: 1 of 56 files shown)


clayui.com/package.json
```diff
@@ -55,6 +56,7 @@
 		"gatsby-plugin-react-helmet": "^3.3.10",
 		"gatsby-plugin-sass": "^2.3.12",
 		"gatsby-plugin-typescript": "^2.4.18",
+		"gatsby-react-router-scroll": "3.0.0",
```
matuzalemsteles, line 59: @pat270 from what I understand it needs is like `resolutions` too so you can force the use of this version in Gatsby or is that just enough?

pat270: This worked for me, but I can add the resolutions line.

matuzalemsteles: No problem, I say this because I see two versions of the package in yarn.lock:

https://github.com/liferay/clay/pull/3676/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.scss#diff-8ee2343978836a779dc9f8d6b794c3b2R11141

pat270:
@matuzalemsteles @bryceosterhaus I checked next.clayui.com and this pr doesn't seem to work in production. Going to send another one if I can fix it.

PR #3677, opened 2020-09-01
fix(@clayui/shared): useTransitionHeight should still expand and coll…

### LINKED ISSUE #3637: ClayPanel: Panel does not Collapse/Close after second hit on Microsoft Edge W10

here: https://codesandbox.io/s/agitated-cherry-r6rdt?file=/src/App.js

1. The panel should Collapse/Close after hit on Panel
2. A transition slide should be shown

## Environment

| Tech  | Version |
| ----- | ------- |
| @clayui/panel  | v3.2.3  |
| React | v16.12       |
| Microsoft Edge Windows 10 |  84.0.522.63 (Latest) |

pat270:
…apse if transitions aren't supported

fixes #3637

### DIFF
--- a/packages/clay-shared/src/useTransitionHeight.ts
+++ b/packages/clay-shared/src/useTransitionHeight.ts
@@ -40,16 +40,45 @@ export function useTransitionHeight(
 ) {
 	const [transitioning, setTransitioning] = React.useState<boolean>(false);
 
+	const transitionBuffer = 0.01;
+	const transitionTimerRef = React.useRef<NodeJS.Timeout | null>(null);
+
+	const fakeTransitionEnd = () => {
+		const {transitionDelay, transitionDuration} = window.getComputedStyle(
+			contentRef.current
+		);
+		const totalDuration =
+			(parseFloat(transitionDelay) +
+				parseFloat(transitionDuration) +
+				transitionBuffer) *
+			1000;
+
+		transitionTimerRef.current = setTimeout(() => {
+			setTransitioning(false);
+			setVisible(!visible);
+
+			if (!visible) {
+				removeCollapseHeight(contentRef);
+			}
+		}, totalDuration);
+	};
+
 	React.useEffect(() => {
 		if (transitioning) {
 			setCollapseHeight(contentRef);
 			if (visible) {
 				removeCollapseHeight(contentRef);
 			}
+
+			fakeTransitionEnd();
 		}
 	}, [transitioning]);
 
 	const handleTransitionEnd = (event: React.TransitionEvent) => {
+		if (transitionTimerRef.current) {
+			clearTimeout(transitionTimerRef.current);
+		}
+
 		if (event.target === contentRef.current && transitioning && !visible) {
 			setVisible(true);
 			setTransitioning(false);


packages/clay-shared/src/useTransitionHeight.ts
```diff
@@ -40,16 +40,42 @@ export function useTransitionHeight(
 ) {
 	const [transitioning, setTransitioning] = React.useState<boolean>(false);
 
+	let transitionTimer: any;
```
matuzalemsteles, line 43: This is very unsafe, as we can easily lose its value in a rendering, to maintain the ideal would be to use `useRef` here.

packages/clay-shared/src/useTransitionHeight.ts
```diff
@@ -40,16 +40,42 @@ export function useTransitionHeight(
...
+
+	const fakeTransitionEnd = () => {
+		const {transitionDelay, transitionDuration} = window.getComputedStyle(
+			contentRef.current
+		);
+		const totalDuration =
+			(parseFloat(transitionDelay) +
+				parseFloat(transitionDuration) +
+				0.01) *
```
matuzalemsteles, line 52: @pat270 this magic value here, is it just for the timeout not to run before calling a `transitionEnd`?

packages/clay-shared/src/useTransitionHeight.ts
pat270: Not sure how to use `useRef` for this case, this variable is to hold the `setTimeout` function so we can clear it in `handleTransitionEnd`. Can you give me an example?

Also I think it's better to manually fire a React `transitionEnd` event instead of reimplementing `handleTransitionEnd` in the `setTimeout`, but I couldn't find any documentation on it.

packages/clay-shared/src/useTransitionHeight.ts
pat270: Yes, this is a buffer. If `transitionEnd` fires the `setTimeout` is cleared in `handleTransitionEnd`.

pat270:
> MS Edge based on Chromium v84, the transition is a little broken

I don't have that browser installed to test. My guess is there is a delay before the panel instantly collapses and opens? This is expected for browsers that support CSS transitions, but for whatever reason that transition isn't happening.

packages/clay-shared/src/useTransitionHeight.ts
matuzalemsteles: > Not sure how to use useRef for this case, this variable is to hold the setTimeout function so we can clear it in handleTransitionEnd. Can you give me an example?

Sure, see this example:

```js
const transitionTimerRef = useRef<NodeJS.Timeout | null>(null);

// Write
transitionTimerRef.current = setTimeout

// Read
clearTimeout(transitionTimerRef.current);
```

> Also I think it's better to manually fire a React transitionEnd event instead of reimplementing handleTransitionEnd in the setTimeout, but I couldn't find any documentation on it.

To do this we would have to manually trigger an event on the node, but we would still have to have a timeout anyway. Starting it immediately could break the animation.

packages/clay-shared/src/useTransitionHeight.ts
matuzalemsteles: Maybe we may want to add this value to a constant to add more context. What do you think?

matuzalemsteles:
> My guess is there is a delay before the panel instantly collapses and opens? This is expected for browsers that support CSS transitions, but for whatever reason that transition isn't happening.

When I was analyzing, the `transitionEnd` was not triggered even after opening the panel for the first time, so I think it would dismiss this case, it only happens in this browser and only in Win 10, it also has for Mac now but it didn't happen, version v85 this was solved for Win 10, I was checking this out today too.

packages/clay-shared/src/useTransitionHeight.ts
bryceosterhaus: I like the idea of adding at least a comment above this value explaining why its needed. I dont know if its needed to be a constant at the top of the file since we don't use it in multiple places, but the comment with context would be great.

packages/clay-shared/src/useTransitionHeight.ts
bryceosterhaus: One note is sometimes TS complains about the type of `setTimeout`, which is why I think `NodeJS.Timeout` is used as the type. But just be aware it may complain

packages/clay-shared/src/useTransitionHeight.ts
matuzalemsteles: Yeah, this is another good option.

pat270:
I think I got all the changes, let me know if I missed any.

matuzalemsteles:
Hey @pat270 I'm going to analyze this with a little more care, I'm still not 100% satisfied with implementing a "fallback" to cover this case, just trying to predict that we won't swell this code by creating some roots for the cases, I'll try to think of other possibilities or just a way of trying to reuse more of the functions we have, such as trying to trigger the `transitionEnd`. What do you think? cc @bryceosterhaus

matuzalemsteles:
Resent #3710

PR #3679, opened 2020-09-01
refactor(clayui.com): Changes the implementation of Tabs from a single page to multiple pages

matuzalemsteles:
For context: https://github.com/liferay/clay/pull/3654#discussion_r481295365

Just a first implementation to change the current behavior of tabs from a single page to multiple pages. Further changes are still needed, I will work on that tomorrow. But it is very functional, I managed to remove some complications that we had in the `docs.js` template.

### DIFF
--- a/clayui.com/content/docs/components/css-alerts.md
+++ b/clayui.com/content/docs/components/css-alerts.md
@@ -2,6 +2,7 @@
 title: 'Alerts'
 description: 'Alerts are used to capture the attention of the user in an intrusive way. They can be used just to say that something went right, or perhaps to say that something needs to be reviewed.'
 lexiconDefinition: 'https://liferay.design/lexicon/core-components/alerts/'
+mainTabLink: 'docs/components/alert.html'
 ---
 
 <div class="nav-toc-absolute">
--- a/clayui.com/content/docs/components/css-alerts.md
+++ b/clayui.com/content/docs/components/css-alerts.md
@@ -2,7 +2,7 @@
 title: 'Alerts'
 description: 'Alerts are used to capture the attention of the user in an intrusive way. They can be used just to say that something went right, or perhaps to say that something needs to be reviewed.'
 lexiconDefinition: 'https://liferay.design/lexicon/core-components/alerts/'
-mainTabLink: 'docs/components/alert.html'
+mainTabURL: 'docs/components/alert.html'
 ---
 
 <div class="nav-toc-absolute">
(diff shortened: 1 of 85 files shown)


clayui.com/content/docs/components/css-alerts.md
```diff
@@ -2,6 +2,7 @@
 title: 'Alerts'
 description: 'Alerts are used to capture the attention of the user in an intrusive way. They can be used just to say that something went right, or perhaps to say that something needs to be reviewed.'
 lexiconDefinition: 'https://liferay.design/lexicon/core-components/alerts/'
+mainTabLink: 'docs/components/alert.html'
```
kresimir-coko, line 5: Shouldn't this be `mainTabURL`? My thinking being this isnt a Link but a URL that will be consumed by a Link

matuzalemsteles: Yeah, actually I haven't been using it with the intention of being a URL but a hyperlink, just a space to create links between documents, not necessarily being a URL, although I am practically using it for this purpose. But if you take it to the literal of the URL it would probably be a `mainTabPath`. I have no problem with continuing with this, I changed that name at least 2 times 😅

pat270:
@matuzalemsteles I like this, works like it should.

jbalsas:
> I like this, works like it should

robot

I like this assessment, completely devoid of emotion like it should

clayui.com/content/docs/components/css-alerts.md
bryceosterhaus: Personally I think `url` might be better, primarily because `path` can be easily confused with file path and my first thought was you were pointing to a specific file.

or maybe `mainTabURLPath` 😂 😂

matuzalemsteles: haha yeah, finding an appropriate name is a kind of hard 😆!! An option not to confuse would be `mainTabPathname` but 🤷‍♂️.

matuzalemsteles:
> I did notice that theres a bit of a delay when clicking on the tabs, any idea what takes the tab so long to load?

Yeah, now this a page, the source code of the page is loaded when the user clicks, unfortunately there will be a slow on the first click 😕. I think I can try using Gatsby's `Link` component it tries to download the page code in advance to try to smooth it out but I'm not sure.

bryceosterhaus:
@matuzalemsteles its not a huge issue and we can fix later, I had just noticed it for the first time.

matuzalemsteles:
> @matuzalemsteles its not a huge issue and we can fix later, I had just noticed it for the first time.

No problem, I pushed some changes now that can improve or decrease this slowness I hope 😁.

bryceosterhaus (approved):
LGTM! 

As for the `mainTab*` name, I'm really fine with whatever we choose, I don't think its a big issue either way.

kresimir-coko:
LGTM too!

matuzalemsteles:
So, I went with `mainTabURL` since I had a consensus for that side and I'm also fine with that 🙂.

PR #3680, opened 2020-09-02
fix(clayui.com): Fix Contributing page not properly redirecting

kresimir-coko:
Hey, so this one is pretty simple but that might be not a good thing in this case (I might have made it too simple by not making it generic enough).

I wanted to make it generic so that the check isn't testing against this specific file (`contributing.html`), but I couldn't figure it out within the confines of the `writeRedirectsFile` function.

I attempted changing the `FILE_PATH` conditionally but I couldn't get to a simpler solution than what I ended up with in this PR.

### DIFF
--- a/clayui.com/content/docs/contributing.md
+++ b/clayui.com/content/docs/contributing.md
@@ -1,5 +1,6 @@
 ---
 title: 'Contributing'
 order: 4
-redirect: 'https://github.com/liferay/clay/blob/master/CONTRIBUTING.md'
 ---
+
+See [CONTRIBUTING doc](https://github.com/liferay/clay/blob/master/CONTRIBUTING.md) on our Github repository.


clayui.com/gatsby/createPages.js
```diff
@@ -11,6 +11,9 @@ require('dotenv').config({
 	path: `.env.${process.env.NODE_ENV}`,
 });
 
+const CONTRIBUTING_URL =
+	'https://github.com/liferay/clay/blob/master/CONTRIBUTING.md';
+
```
wincent, line 16: I wouldn't define this because:

- It is used in exactly one place; _and_
- All the other `fromPath`, `toPath` pairs are just written inline as string literals, making this inconsistent.

clayui.com/content/docs/contributing.md
```diff
@@ -1,5 +1,4 @@
 ---
 title: 'Contributing'
 order: 4
-redirect: 'https://github.com/liferay/clay/blob/master/CONTRIBUTING.md'
```
wincent, line 4: Only one other file has a `redirect` property in the frontmatter (`content/docs/index.md`), so perhaps this feature is just broken.

One thing to note about it though is it has `layout: 'redirect'`; maybe that is what's needed to make this work.

clayui.com/plugins/gatsby-plugin-meta-redirect/gatsby-node.js
```diff
@@ -18,7 +18,10 @@ const writeRedirectsFile = async (redirects, folder, pathPrefix) => {
 	for (const redirect of redirects) {
 		const {fromPath, toPath} = redirect;
 
-		if (fromPath.endsWith('index.html')) {
+		if (
+			fromPath.endsWith('index.html') ||
+			fromPath.endsWith('contributing.html')
```
wincent, line 23: This is the bit that makes you wince... maybe this could just be:

```js
if (fromPath.endsWith('.html')) {
```

clayui.com/gatsby/createPages.js
```diff
@@ -128,6 +131,12 @@ module.exports = async ({actions, graphql}) => {
 		redirectInBrowser: true,
 		toPath: '/docs/css/content/typography.html',
 	});
+	actions.createRedirect({
+		fromPath: '/docs/contributing.html',
+		isPermanent: true,
+		redirectInBrowser: true,
+		toPath: CONTRIBUTING_URL,
+	});
```
bryceosterhaus, line 139: I think `createRedirect` is only used for internal redirects within the gatsby site. In the [docs](https://www.gatsbyjs.com/docs/actions/#createRedirect) it says the `toPath` should be a "URL of a created page"

matuzalemsteles:
I agree with @bryceosterhaus I think that just changing the file is enough for now. The redirects were designed only for the routes of the site if I am not mistaken, so we can think of other strategies for that. So that it is not necessary to make these conditions scattered just for this use case.

clayui.com/plugins/gatsby-plugin-meta-redirect/gatsby-node.js
kresimir-coko: Yea, this is the one that made me insecure about my solution.

kresimir-coko:
I'm glad I didn't miss something incredibly obvious, so I did follow @bryceosterhaus's advice, and just added the content that points to the Contributing doc.

It's a shame that Gatsby doesn't have a clear-cut way of supporting this kind of redirecting.

matuzalemsteles:
@kresimir-coko yeah, I think these internal Gatsby APIs don’t have that focus but we can still use `redirect` with `layout: redirect` and maybe do that on the Sidebar side, creating a link instead of waiting for Gatsby to create this rule in route, he would probably have to do something else since he can try to create this rule and we just want to ignore it.

kresimir-coko:
@matuzalemsteles I tried adding `layout: 'redirect'` alongside the `redirect: url` but that resulted in a white page. I tested that on the dev site.

matuzalemsteles:
> @matuzalemsteles I tried adding layout: 'redirect' alongside the redirect: url but that resulted in a white page. I tested that on the dev site.

Yeah, the idea was just to have the redirect information and not to create a page, just to collect it from the template side and decide when not to use the Gatsby Link component by an `a` tag with `href`. For example here https://github.com/liferay/clay/blob/master/clayui.com/src/components/Sidebar/Navigation.js#L125

We can go with that PR and test it another time.

kresimir-coko:
@matuzalemsteles I'm not a 100% sure I understand, do we want to just have the following (I submitted a commit with this above):
```markdown
---
title: 'Contributing'
order: 4
---

See [CONTRIBUTING doc](https://github.com/liferay/clay/blob/master/CONTRIBUTING.md) on our Github repository.
```

Or do you want me to make something more robust? Something to programatically decide if it's a `<Link />` or  `<a href="https://github.com/liferay/clay/blob/master/CONTRIBUTING.md" />` depending on some inputs?

matuzalemsteles:
@kresimir-coko don't worry about it, sorry! We can try this in a more programmatic way to decide this, just to make redirects outside the site, but we do it in another PR, for now, we are going with this.

matuzalemsteles (approved):
LGTM!

PR #3684, opened 2020-09-04
fix(clayui.com): Code examples on certain parts of the documentation …

pat270:
…should be expanded by default

docs(clayui.com): SCSS add more details about variables, functions, mixins

Couldn't figure out an elegant way to keep certain CSS code examples examples expanded on first load with prism js or gatsby-transformer-remark. I just hacked it with the keyword `code_expand`.

This also updates the SCSS page with more detail about variables, mixins, and functions

### DIFF
--- a/clayui.com/plugins/gatsby-remark-use-clipboard/index.js
+++ b/clayui.com/plugins/gatsby-remark-use-clipboard/index.js
@@ -5,14 +5,39 @@
 
 const visit = require('unist-util-visit');
 
+const parseOptions = (meta) => {
+	const availableOptions = {};
+
+	if (meta) {
+		meta.map((option) => {
+			const key = option.match(/.*(?=:)|.*$/);
+			const value = option.match(/(?<=:).+/);
+
+			if (key) {
+				availableOptions[key[0].trim()] =
+					value === null ? true : value[0].trim();
+			}
+		});
+	}
+
+	return availableOptions;
+};
+
 module.exports = ({markdownAST}) => {
 	visit(markdownAST, 'html', (node) => {
 		if (typeof node.lang === 'undefined') {
 			return;
 		}
 
+		const language = node.meta ? node.lang + node.meta : node.lang;
+
+		const {expanded} = parseOptions(language.match(/(?<=\{).+?(?=\})/g));
+
+		const expandedClass = expanded ? ' expanded' : '';
+		const hideClass = expanded ? ' hide' : '';
+
 		node.value = `
-			<div class="code-container">
+			<div class="code-container${expandedClass}">
 				<div class="copied-alert alert alert-fluid alert-info d-none" role="alert">
 					<div class="container">
 						<div class="alert-autofit-row autofit-row">
@@ -35,7 +60,7 @@ module.exports = ({markdownAST}) => {
 					</div>
 				</div>
 
-				<span class="clay-p code-sample-info">
+				<span class="clay-p code-sample-info${hideClass}">
 					Code Sample (expand to see it)
 				</span>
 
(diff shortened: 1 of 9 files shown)


pat270:
Having issues with `yarn format` adding semi colons at the end of snippets in:
\`\`\`jsx
code_expanded;

import ClayList from '@clayui/list';
\`\``
but not in:

\`\`\`jsx
code_expanded

import ClayDropDown from '@clayui/drop-down';

<ClayDropDown>
<ClayDropDown.Action>
<ClayDropDown.Caption>
<ClayDropDown.Divider>
<ClayDropDown.Group>
<ClayDropDown.Help>
<ClayDropDown.Menu>
<ClayDropDown.Item>
<ClayDropDown.ItemList>
<ClayDropDown.Search>
```

**Review by matuzalemsteles — COMMENTED** [review]
@pat270 this is great, really some snippets need to be `expanded`, I think we can try to follow the syntax that the prism ends up using to configure settings in the code block.

So instead of adding `code_expanded` within the code block, we can add it along with the code block lang. For example:
```markdown
```jsx{expanded: true}
...
```

In the `remark-use-clipboard` plugin, we can parse the options.
```js
const parseOptions = (language) => {
	const availableOptions = {
		expanded: null,
	};

	if (language && language.split(`{`).length > 1) {
		const options = language.split(`{`);

		options.forEach((option) => {
			option = option.slice(0, -1);
			const splitOption = option.replace(/ /g, ``).split(`:`);

			if (splitOption.length === 2 && splitOption[0] === 'expanded') {
				availableOptions.expanded = Boolean(splitOption[1].trim());
			}
		});
	}

	return availableOptions;
};

module.exports = ({markdownAST}) => {
	visit(markdownAST, 'html', (node) => {
		if (typeof node.lang === 'undefined') {
			return;
		}

		const language = node.meta ? node.lang + node.meta : node.lang;
		const {expanded} = parseOptions(language);
		...
```

This will make it a little easier and not have problems with the formatter and also with the configuration standards of this plugin. What do you think?

pat270:
Thanks @matuzalemsteles ! This is what I was looking for. I think I tried some variant of that and couldn't get anything to work with. I'll try this out right now.

pat270:
@matuzalemsteles Thanks for the help! I didn't like how they were doing it in `gatsby-remark-prismjs` so I put my own spin on it. It took me forever to figure out what was going on.

We will be able to pass in options two ways and still have access to other options I don't think we'll ever use.
```
```shell{expanded}{expanded: true}{promptUser: alice}{promptHost: dev.localhost}
```

clayui.com/plugins/gatsby-remark-use-clipboard/index.js
```diff
@@ -5,14 +5,39 @@
 
 const visit = require('unist-util-visit');
 
+const parseOptions = (meta) => {
+	const availableOptions = {};
+
+	if (meta) {
+		meta.map((opt) => {
```
kresimir-coko, line 12: I believe that we don't want to abbreviate variables like these, `opt` should be `option`. @wincent correct me if I'm wrong

kresimir-coko (commented):
Apart from the variable abbreviation it LGTM

clayui.com/plugins/gatsby-remark-use-clipboard/index.js
wincent: Yeah, in general nope. Rather than answering this in lots of places, I created this PR to liferay-frontend-guidelines: https://github.com/liferay/liferay-frontend-guidelines/pull/195

matuzalemsteles (approved):
That looks cleaner now, great work @pat270! LGTM!

clayui.com/plugins/gatsby-remark-use-clipboard/index.js
pat270: Changed `opt` to `option` let me know if there's anything else.

matuzalemsteles:
So, I will merge this.

PR #3688, opened 2020-09-05
fix(@clayui.com): Atlas Buttons `.btn-outline-secondary` should have …

### LINKED ISSUE #3687: @clayui/css Atlas .btn-outline-secondary color is incorrect
Caused by #3149 

btn-outline-secondary

`.btn-outline-secondary` should be color #6b6c7e instead of #272833

pat270:
…color #6b6c7e, caused by #3149

fixes #3687


marcoscv-work:
LGTM !

So do we need this change along portal?

PR #3689, opened 2020-09-07
feat(@clayui/dual-list-box): Make reorder & transfer button state more consistent

### LINKED ISSUE #3682: Reorder buttons ClayDualListBox

* https://clayui.com/docs/components/dual-list_box.html
* https://codesandbox.io/s/clayduallistbox-content-dashboard-configuration-0ze8o

Both are examples of `ClayDualListBox` with the out-of-the-box implementation.

You can view these results with the out-of-the-box component.

1. When there is only one element in the "In Use" box, disable both reorder buttons.
2. When there are two elements, and the user selects the first one, disable the reorder up button.
2. When there are two elements, and the user selects the second one, disable the reorder down button.

### Actual behavior screenshots
1.
Screenshot 2020-09-03 at 17 18 59

2.
Screenshot 2020-09-03 at 17 19 10

3.
Screenshot 2020-09-03 at 17 19 14

I don't know if it's a bug or an improvement. But I think it would be very nice to have it.

Thanks! 😄

kresimir-coko:
Fixes #3682 

I believe I've handled all cases of buttons being disabled except for one, and that is something has been transferred between select boxes but nothing is visually selected. After investigating, I noticed that the transferred item (or items if more have been selected), is kept in the `leftSelected` variable (in the case when transfering from left to right). I couldn't figure out a clean way of resetting the `leftSelected` value back to an empty array.

### DIFF
--- a/packages/clay-form/src/SelectBox.tsx
+++ b/packages/clay-form/src/SelectBox.tsx
@@ -154,18 +154,13 @@ const ClaySelectBox: React.FunctionComponent<IProps> = ({
 		Array.isArray(value) ? value : [value]
 	);
 
-	const selectedIndex = selectedIndexes.length === 1 && selectedIndexes[0];
-
 	const noItems = !items.length;
 
 	const noItemsSelected = !selectedIndexes.length;
 
-	const firstItemSelected =
-		selectedIndex === 0 || selectedIndexes.includes(0);
+	const firstItemSelected = selectedIndexes.includes(0);
 
-	const lastItemSelected =
-		selectedIndex === items.length - 1 ||
-		selectedIndexes.includes(items.length - 1);
+	const lastItemSelected = selectedIndexes.includes(items.length - 1);
 
 	return (
 		<div className={classNames(className, 'form-group')}>
--- a/packages/clay-form/src/SelectBox.tsx
+++ b/packages/clay-form/src/SelectBox.tsx
@@ -156,6 +156,17 @@ const ClaySelectBox: React.FunctionComponent<IProps> = ({
 
 	const selectedIndex = selectedIndexes.length === 1 && selectedIndexes[0];
 
+	const noItems = !items.length;
+
+	const noItemsSelected = !selectedIndexes.length;
+
+	const firstItemSelected =
+		selectedIndex === 0 || selectedIndexes.includes(0);
+
+	const lastItemSelected =
+		selectedIndex === items.length - 1 ||
+		selectedIndexes.includes(items.length - 1);
+
 	return (
 		<div className={classNames(className, 'form-group')}>
 			{label && (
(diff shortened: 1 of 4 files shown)


packages/clay-form/stories/index.tsx
```diff
@@ -87,11 +87,15 @@ storiesOf('Components|ClayDualListBox', module).add('default', () => {
 		<ClayDualListBox
 			disableLTR={
 				boolean('disableLTR', false) ||
-				secondSelectBoxItems.length >= rightMaxItems
+				secondSelectBoxItems.length >= rightMaxItems ||
+				!firstSelectBoxItems.length ||
+				!leftSelected.length
```
bryceosterhaus, line 92: I don't think we shouldn't put this responsibility on the user to manage these checks, I think we should be able to internalize.

packages/clay-form/src/SelectBox.tsx
```diff
@@ -212,7 +214,13 @@ const ClaySelectBox: React.FunctionComponent<IProps> = ({
...
 								aria-label={ariaLabels.reorderUp}
 								className="reorder-button reorder-button-up"
-								disabled={!value.length}
+								disabled={
+									!value.length ||
+									selectedIndex === 0 ||
+									selectedIndexes.includes(0) ||
+									!items.length ||
+									selectedIndexes.length === 0
+								}
```
bryceosterhaus, line 223: This expression is hard to follow, I think it'd be best to put into some named variables in order to clean it up. I also noticed that `selectedIndexes` is directly related to `value`, so you shouldn't need to have checks for both.

```jsx
const noItems = !items.length;
const noItemsSelected = !selectedIndexes.length;
const firstItemSelected = selectedIndex === 0 || selectedIndexes.includes(0);
```

bryceosterhaus (changes requested):

packages/clay-form/stories/index.tsx
kresimir-coko: Yea, that makes sense, I'll get on top of that!

packages/clay-form/src/SelectBox.tsx
kresimir-coko: You're right, I got caught up in making it work, so I didn't think much about readability.

kresimir-coko:
@bryceosterhaus I've sent a commit that should cover your requests

packages/clay-form/src/SelectBox.tsx
```diff
@@ -154,6 +154,19 @@ const ClaySelectBox: React.FunctionComponent<IProps> = ({
...
+	const noItems = !items.length;
+
+	const noItemsSelected = !selectedIndexes.length;
+
+	const firstItemSelected =
+		selectedIndex === 0 || selectedIndexes.includes(0);
+
+	const lastItemSelected =
+		selectedIndex === items.length - 1 ||
+		selectedIndexes.includes(items.length - 1);
```
bryceosterhaus, line 168: I don't think you need the `selectedIndex` variable. Can't this just be simplified to this?

```suggestion
	const noItems = !items.length;
	const noItemsSelected = !selectedIndexes.length;
	const firstItemSelected = selectedIndexes.includes(0);
	const lastItemSelected = selectedIndexes.includes(items.length - 1);
```

kresimir-coko: It can! @bryceosterhaus good job spotting that, because `selectedIndexes` is always an array, even with just one selected item, `selectedIndexes.includes()` applies always.

kresimir-coko:
I've pushed even more suggested changes, simplifying the checks and removing the `selectedIndex` variable.

bryceosterhaus (approved):
LGTM!

PR #3698, opened 2020-09-10
fix(clayui.com): Toggling Show Atlas Theme should update color values…

pat270:
… on Colors Page

fix(clayui.com): Separate site styles that rely on atlas and base themes, mainly css colors page

fix(clayui.com): We should output sourcemaps for Clay CSS. Unfortunately Sass sourcemaps doesn't work well with the way Clay CSS generating our CSS, but better than nothing.

This fixes https://clayui.com/docs/css/color.html color values to update when switching from Atlas to Base themes.


clayui.com/src/components/LayoutNav/LayoutNav.js
```diff
@@ -14,6 +14,35 @@ import Search from './Search';
...
+	};
+
+	const checkCSS = setInterval(() => {
+		const clayCSSTheme = window.getComputedStyle(themeEl);
+		const themeName = checked ? 'atlas' : 'base';
+
+		if (clayCSSTheme.content.replace(/"/g, '') === themeName) {
+			loaded();
+		}
+	}, 100);
```
bryceosterhaus, line 41: This seems a little odd to just always have running in the background without ever canceling the intervak. Can we listen in the `onChange` callback without the interval?

pat270: We can't because I'm getting the values based on the css file. We need to wait for it to load before we get the right values. This function only runs when `onChange` is fired. I suppose it could run forever if we end up changing the css file.

bryceosterhaus: So `onChange` is fired and then this interval starts to run and never stops? Can we cancel the interval after it is loaded?

pat270: Interval starts and is canceled at https://github.com/liferay/clay/pull/3698/commits/a6e0639d4fcfba3b5c434f46dcfd21ee06539d23#diff-f2ffdac3dda83b4a8838ef0e8dc88a56R23. 

I was saying if we end up removing https://github.com/liferay/clay/pull/3698/commits/a6e0639d4fcfba3b5c434f46dcfd21ee06539d23#diff-cca54bd05824f23cf8707bc5dff1c409R155-R157, it could run forever.

bryceosterhaus: ohh I see it now, my bad, I somehow skipped right over seeing that. In that case I think its fine and I'm good with merging this

PR #3702, opened 2020-09-16
chore(clayui.com): Co-locate API and Markup tab markdown files to their respective folders

### LINKED ISSUE #3701: Co-locate API and Markup files to their respective docs
This issue is a continuation of #3451 because we doubled the amount of files documenting components in a location that wasn't tied to their components, this issue should fix that.

kresimir-coko:
Fixes #3701 

Thanks @bryceosterhaus for the tip on `onCreateNode.js`, I adapted it a bit so that components where we have more sub-components (namely Form), don't get messed up if we just name them `api-tab.mdx`.


clayui.com/gatsby/onCreateNode.js
```diff
@@ -78,7 +78,10 @@ module.exports = exports.onCreateNode = ({actions, getNode, node}) => {
 						'@clayui/',
 						''
 					)}.html`;
-				} else {
+				} else if (
+					!relativePath.includes('api-') ||
+					!relativePath.includes('markup-')
```
matuzalemsteles, line 83: Instead of adding the prefix, you could use `mainTabURL` in the conditional. Since everyone who is as a tab will have `mainTabURL`.

matuzalemsteles (commented):
Oh great, a little change!

clayui.com/gatsby/onCreateNode.js
bryceosterhaus: @matuzalemsteles good idea!

kresimir-coko:
Seems that `yarn build` is complaining about `mainTab` not existing on this line: https://github.com/liferay/clay/blob/master/clayui.com/src/templates/docs.js#L157. Gonna have to take a look

clayui.com/gatsby/onCreateNode.js
```diff
@@ -78,7 +78,7 @@ module.exports = exports.onCreateNode = ({actions, getNode, node}) => {
 						'@clayui/',
 						''
 					)}.html`;
-				} else {
+				} else if (mainTabURL) {
```
matuzalemsteles, line 81: @kresimir-coko I see that all the main tabbed pages are using the wrong information at the top of the document, they are being pointed to `Button Group`. Maybe here we should ignore it when the page has `mainTabURL`?

```suggestion
				} else if (!mainTabURL) {
```

kresimir-coko: Interesting @matuzalemsteles, I can't replicate that locally, on dev or prod server. I'll try the suggestion

kresimir-coko:
It's interesting that `!mainTabURL` works, I assumed this block of code wanted to match all of the files that build out the tabs and rename them properly. Should be mergeable now @matuzalemsteles

clayui.com/gatsby/onCreateNode.js
matuzalemsteles: Hmm, this is really weird.

matuzalemsteles:
> It's interesting that !mainTabURL works, I assumed this block of code wanted to match all of the files that build out the tabs and rename them properly. Should be mergeable now @matuzalemsteles

I think it's working because https://github.com/liferay/clay/pull/3702/files#diff-1ed3aafc4aadd6656fe082aa8048f293R96 will take care of creating the new slugs for the prefixed files or better with `mainTabURL`.

We may have to organize our logics a bit in this file to try to take a little out of the redundancies and complexities that are appearing and also try to write some descriptions when it falls in these cases.

matuzalemsteles (approved):

Issue #3696, opened 2020-09-10
Esc to dismiss a tooltip


victorvalle:
After reviewing [LEXI-600](https://issues.liferay.com/browse/LEXI-600), Lexicon finds that implementing "esc" command key to dismiss a tooltip would be a positive improvement in terms of accessibility for the Tooltip component.

jbalsas:
Sounds good to me... some of this might need to be implemented in DXP side directly, so you might file an LPS as well :) 

My only concern is that `ESC` might trigger other actions, so I'd investigate a little to see what are the possible ramifications of such a behaviour.

One ridiculous example is that `ESC` with focus on Chrome DevTools opens a panel. Native tooltips can be dismissed using `ESC` (so this proposal is aligned with default behaviour). Since tooltips show without focus, you can have focus somewhere and trigger something while dismissing the tooltip. I imagine this interacting with autocompletes, closing dropdowns, modals... things like that...

matuzalemsteles:
I have the same feeling that Chema, we will have problems with DropDowns and Modals, mainly in Modals, maybe we can prevent him from closing everything, I can test this next week.

victorvalle:
Hi @matuzalemsteles , you closed this ticket without the outcome of your exploration. Could you write some lines about what happened? Otherwise, it just looks to me it was closed without any work done on it.

Thanks

victorvalle:
Sorry for my last comment, I see in #3704 that you did the tests and implementation.
Thanks!

PR #3704, opened 2020-09-17
chore(@clayui/tooltip): close the tooltip when pressing Esc

matuzalemsteles:
This is just an initial implementation of #3696, the problem with this is you can still close DropDown when it is open and there is a Tooltip, so I thought about adding the event in the capture phase so we could get the event first before others but it also depends when the event was added, the order influences, in this implementation here I always add the event when I open the tooltip and then remove it to avoid problems. I think we managed to get around some behaviors within Clay but not something in a DXP world where people can add events and we have no control. Suggestions?

I added the Tooltip to the Button inside the Modal with an Autocomplete in the ClayModal story for testing.

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -3,9 +3,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {ClayPortal} from '@clayui/shared';
+import {ClayPortal, Keys} from '@clayui/shared';
 import domAlign from 'dom-align';
-import React from 'react';
+import React, {useCallback} from 'react';
 import warning from 'warning';
 
 import ClayTooltip from './Tooltip';
@@ -139,7 +139,7 @@ const TooltipProvider: React.FunctionComponent<{
 	const titleNodeRef = React.useRef<HTMLElement | null>(null);
 	const tooltipRef = React.useRef<HTMLElement | null>(null);
 
-	const handleHide = ({target}: any) => {
+	const handleHide = useCallback(({target}: any) => {
 		if (!titleNodeRef.current) {
 			return;
 		}
@@ -147,6 +147,7 @@ const TooltipProvider: React.FunctionComponent<{
 		const dataTitle = titleNodeRef.current.getAttribute('data-title');
 
 		if (dataTitle) {
+			document.removeEventListener('keyup', handleEsc, true);
 			target.removeEventListener('click', handleHide);
 			titleNodeRef.current.setAttribute('title', dataTitle);
 			titleNodeRef.current.removeAttribute('data-title');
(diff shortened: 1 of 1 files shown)


packages/clay-tooltip/src/TooltipProvider.tsx
```diff
@@ -139,14 +139,15 @@ const TooltipProvider: React.FunctionComponent<{
...
-	const handleHide = ({target}: any) => {
+	const handleHide = useCallback(({target}: any) => {
 		if (!titleNodeRef.current) {
 			return;
 		}
 
 		const dataTitle = titleNodeRef.current.getAttribute('data-title');
 
 		if (dataTitle) {
+			document.removeEventListener('keyup', handleEsc, true);
```
kresimir-coko, line 150: I'm interested why you're providing `true` as the 3rd argument here, as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) it's an `options` argument. Does it mean that the 3rd argument is implicitly `capture`:
> `capture`: A Boolean which indicates that events of this type will be dispatched to the registered listener before being dispatched to any EventTarget beneath it in the DOM tree.

matuzalemsteles: @kresimir-coko yeah, essentially I want to be able to capture the event as soon as possible to stop it, so I add `true`, otherwise, I would be picking up on Bubbling which would be the last phase.

The idea is to close the tooltip with Esc but there may be problems when it is in a Modal or DropDown for example, that Esc closes both, so I need to get the event first and cancel and close the Tooltip.

bryceosterhaus:
@matuzalemsteles  I think this looks good to me, I don't know if we need to be too concerned about a tooltip in a dropdown, I think that use case is much smaller and less annoying than the modal use case. There isn't a ton of state lost, if they accidentally close the drop down with esc, whereas the modal could contain much more state information that would be annoying when accidentally closing.

Semi-related: I did notice when using escape for autocomplete will close the whole modal, could we add the same logic there?

matuzalemsteles:
> @matuzalemsteles I think this looks good to me, I don't know if we need to be too concerned about a tooltip in a dropdown, I think that use case is much smaller and less annoying than the modal use case. There isn't a ton of state lost, if they accidentally close the drop down with esc, whereas the modal could contain much more state information that would be annoying when accidentally closing.

Makes sense, I'll fix the commits.

> Semi-related: I did notice when using escape for autocomplete will close the whole modal, could we add the same logic there?

@bryceosterhaus I think it's related to `DropDown.Menu`, which also closes via Esc. I will test this.

matuzalemsteles:
@bryceosterhaus updated, I will add the implementation in `DropDown.Menu` in another PR.

packages/clay-tooltip/src/TooltipProvider.tsx
kresimir-coko: 👍  thanks for the explanation

bryceosterhaus (approved):
LGTM!

PR #3706, opened 2020-09-18
docs(clayui.com): Document how we develop components in Clay for viewing on GitHub

### LINKED ISSUE #3705: Document how we build new components
Going into depth with how we develop implementations of our components, like the fact that props need to be named `IProps`, `IItemProps`, casting component props to `IProps` so they show up in the API table.

Documenting High vs Low level components concepts, common pitfalls, etc.

Child of #3699

kresimir-coko:
Fixes #3705 

Let me know if I missed the mark with this, I think I covered everything as concise as I could.

### DIFF
--- a/docs/developing_components.md
+++ b/docs/developing_components.md
@@ -2,6 +2,7 @@
 
 This article explains best practices, concepts, and common pitfalls when making components for Clay.
 We go over concepts of high and low level components, composition, and advice on tying it all together.
+When we make new components we must determine how high-level the component should be, and know what the component needs to offer, to better understand this, read [component offerings](https://github.com/liferay/clay/wiki/Component-Offerings)
 
 ## High-level vs low-level components and composition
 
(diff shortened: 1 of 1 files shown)


matuzalemsteles (commented):
@kresimir-coko It looks good to me, there are some other places that we also talk about how we offer the components, it might be interesting to link them to have a better understanding of the reasons and why.

- [Component Offerings](https://github.com/liferay/clay/wiki/Component-Offerings)
- [Design System: Compositional philosophy of components](https://dev.to/matuzalemsteles/design-system-compositional-philosophy-of-components-1cc4) (Optional, much of the writing there is learning from here)

As this would probably be another technical document, I would go into this a little more, showing the examples what to do and what not to do in the sessions, it would be good.

docs/developing_components.md
```diff
@@ -0,0 +1,51 @@
...
+```tsx
+const ClayComponent: React.FunctionComponent<IProps> & {
+    Item: typeof Item;
+    Label: typeof Label;
+} = ({...});
+
+ClayToolbar.Item = Item;
+ClayToolbar.Label = Label;
+
+export default ClayComponent;
```
bryceosterhaus, line 50: Another pattern we use and often works better with ts:

```jsx
export default Object.assign(ClayForm, {
	FeedbackGroup,
	FeedbackIndicator,
	FeedbackItem,
	Group,
	Text,
});
```

bryceosterhaus (commented):
Looks good, I just had one comment on how we export.

Additionally, I think it would be good to make a note about specifying the `displayName` for components. This is important so that it can be easier to read in the React dev tools.

docs/developing_components.md
kresimir-coko: I'll just recommend this pattern then, since it often works better with TS, and we could be consistent and only use one pattern for these kinda things, so I'll just recommend this one.

kresimir-coko:
I've made the suggested changes @matuzalemsteles & @bryceosterhaus let me know if I did it in a way you guys envisioned 😄

matuzalemsteles (commented):
LGTM! I think it's a good start. We will be adding more information and cases over time. What do you think?

bryceosterhaus:
LGTM!

PR #3710, opened 2020-09-22
fix(@clayui/shared): useTransitionHeight should still expand and collapse if transitions aren't supported

### LINKED ISSUE #3637: ClayPanel: Panel does not Collapse/Close after second hit on Microsoft Edge W10

matuzalemsteles:
Fixes #3637, From #3677

I made some changes on top of @pat270's PR, instead of implementing a fallback in support of the `transitionEnd` event, I decided to remove it and rewrite the implementation on top of `setTimeout`.

### DIFF
--- a/packages/clay-navigation-bar/src/__tests__/index.tsx
+++ b/packages/clay-navigation-bar/src/__tests__/index.tsx
@@ -113,6 +113,8 @@ describe('ClayNavigationBar', () => {
 	});
 
 	it('collapses the previously expanded dropdown when trigger element is clicked', async () => {
+		jest.useFakeTimers();
+
 		const {container, getByTestId} = render(
 			<ClayNavigationBar
 				inverted
@@ -143,10 +145,16 @@ describe('ClayNavigationBar', () => {
 		);
 
 		fireEvent.click(getByTestId('navbarToggler'), 'Trigger Label');
-		fireEvent.transitionEnd(getByTestId('NavigationBarDropdown'));
+
+		act(() => {
+			jest.runAllTimers();
+		});
 
 		fireEvent.click(getByTestId('navbarToggler'), 'Trigger Label');
-		fireEvent.transitionEnd(getByTestId('NavigationBarDropdown'));
+
+		act(() => {
+			jest.runAllTimers();
+		});
 
 		let navigationBarDropdown;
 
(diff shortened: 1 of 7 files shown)


pat270:
LGTM

packages/clay-navigation-bar/src/__tests__/index.tsx
```diff
@@ -143,10 +145,16 @@ describe('ClayNavigationBar', () => {
 		);
 
 		fireEvent.click(getByTestId('navbarToggler'), 'Trigger Label');
-		fireEvent.transitionEnd(getByTestId('NavigationBarDropdown'));
+
+		act(() => {
+			jest.runAllTimers();
+		});
```
bryceosterhaus, line 151: How come you have to wrap this in an act?

bryceosterhaus (commented):
Looks good to me, just had one question on the tests

packages/clay-navigation-bar/src/__tests__/index.tsx
matuzalemsteles: This was throwing some errors on the console, as we are using `setTimeout` and has scheduled a function that will change the states, this probably causes new renderings later and we have to wrap this within an `act`.

bryceosterhaus: Hmm, that seems odd. I would think you could just run it without the `act`.

Oh well, im fine with merging!

matuzalemsteles: Yeah, I remember that I had a lot of problems with this and I only solved it by adding `act`, otherwise we will see a lot of warnings on the console like this one.

Screen Shot 2020-09-24 at 18 32 39

I think it makes sense since we will be running all the timeouts that will cause some rendering.

matuzalemsteles:
@bryceosterhaus I will go ahead with this PR according to your last message, in case you see any problems we can create another PR for this.

PR #3711, opened 2020-09-22
feat: add displayName for all components

### LINKED ISSUE #3709: Add `displayName` to every react component
To help developers and read components in the devtools easier, we need to add `displayName` for all of our components.

I am also going to add an eslint rule to enforce this.

bryceosterhaus:
fixes #3709

This ensures we have proper names in the dev tools

### DIFF
--- a/clayui.com/.eslintrc.js
+++ b/clayui.com/.eslintrc.js
@@ -1,0 +1,10 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+module.exports = {
+	rules: {
+		'react/display-name': 'off',
+	},
+};
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,6 +34,7 @@ module.exports = {
 			},
 		],
 		'prefer-template': 'error',
+		'react/display-name': 'error',
 		'react/jsx-boolean-value': 'error',
 		'react/jsx-no-bind': [
 			'error',
--- a/packages/clay-autocomplete/src/Input.tsx
+++ b/packages/clay-autocomplete/src/Input.tsx
@@ -20,4 +20,6 @@ const ClayAutocompleteInput = React.forwardRef<HTMLInputElement, IProps>(
 	}
 );
 
+ClayAutocompleteInput.displayName = 'ClayAutocompleteInput';
+
 export default ClayAutocompleteInput;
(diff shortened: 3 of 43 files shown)


matuzalemsteles (approved):
LGTM!

clayui.com/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
```
kresimir-coko, line 2: Maybe this should be 2020?

wincent: We should probably [let the linter handle this](https://github.com/liferay/clay/blob/master/copyright.js), and if the globs in the top-level package.json aren't letting ESLint "see" this file, we should change the globs.

matuzalemsteles: 🤔 really strange the linter does not catch. Apparently the globs are correct.

wincent: Are you sure? We have a dotfile glob only at the top level, not for nested directories.

matuzalemsteles: Hmm, I'm guessing that the script in [package.json](https://github.com/liferay/clay/blob/master/package.json#L30) will go through any directory using glob `\".*.js\" \"**/*.{js,ts,tsx}\"`. I may be mistaken but it should search nested, right?

bryceosterhaus: We seem to have a handful of lingering copyrights that reference 2019. I'll update that in a separate PR and merge this separately

pat270: @kresimir-coko @bryceosterhaus we should leave the copyright year alone.

> How to write a good copyright statement and license notice
the year of the first publication – a good date would be the year in which you created the file and then do not touch it anymore;

Reference: https://liferay.dev/blogs/-/blogs/how-and-why-to-properly-write-copyright-statements-in-your-code#how-to-write-a-good-copyright-statement-and-license-notice

wincent: @matuzalemsteles:

> Hmm, I'm guessing that the script in [package.json](https://github.com/liferay/clay/blob/master/package.json#L30) will go through any directory using glob `\".*.js\" \"**/*.{js,ts,tsx}\"`. I may be mistaken but it should search nested, right?

I think you're mistaken, _but I may be mistaken_ 😂

My understanding is that `**/*.{js,ts,tsx}` matches `.js`, `.ts`, and `.tsx` in any directory, but it doesn't match dotfiles. Conversely, `.*.js` matches dotfiles, but only at the top level.

Easy way to find out if this is true: try changing the globs.

> we should leave the copyright year alone

This is why I said leave it to the linter — it does exactly what you described.

matuzalemsteles: @wincent oh, you're right, I wasn't thinking about dotfiles 😂, sorry.

PR #3712, opened 2020-09-22
fix(@clayui/drop-down): fixes DropDown error closing Modal when pressing Esc

matuzalemsteles:
From https://github.com/liferay/clay/pull/3704#issuecomment-696863577

I made some changes to DropDown that removes the `useDropdownCloseInteractions` hook that was just being used for `DropDown.Menu` and moved the implementation to the component.


packages/clay-drop-down/src/Menu.tsx
```diff
@@ -173,15 +171,58 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>(
...
+						focusRefOnEsc.current.focus();
+					}
+
+					onSetActive(false);
+				}
+			};
+
+			if (active) {
+				document.addEventListener('keyup', handleEsc, true);
+			}
```
bryceosterhaus, line 218: This could potentially be added multiple times right? I think we need to also remove the listener for whenever `active` is false.

matuzalemsteles: Yeah, we are doing this on `return` after that. With dependency with `active`, we will ensure that it is removed when it is false, but it is not added.

bryceosterhaus: Oh I see, I thought the return only happened when the component was being unmounted and destroyed

matuzalemsteles: > I thought the return only happened when the component was being unmounted and destroyed

Yeah, it will only happen when we have an empty dependency on `useEffect` but as in this case I'm set `active`, it will call the previous cleaning up before calling the effect again.

bryceosterhaus (approved):
LGTM!

PR #3717, opened 2020-09-23
fix(clayui.com): fixes the error when trying to iterate through a NodeList

### LINKED ISSUE #3716: Clay's documentation site is broken in IE 11
Accessing clayui.com/docs is broken in IE 11, generates an error in the console:

```
Object doesn't support property or method 'forEach'
```

matuzalemsteles:
Fixes #3716

This is a little weird that it didn't break before or we didn't really notice.

### DIFF
--- a/clayui.com/src/templates/docs.js
+++ b/clayui.com/src/templates/docs.js
@@ -172,11 +172,13 @@ export default (props) => {
 	}, []);
 
 	useEffect(() => {
-		document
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
I will merge this so that we can have the site live again in IE 11, at least next.clayui.com. Feel free to comment and we can send another PR with the changes.

PR #3719, opened 2020-09-24
feat(@clayui/charts): Split API into it's file, add notice we're switching to Recharts, and cast props to IProps

### LINKED ISSUE #3714: Expand documentation on Charts
Looking at ClayCharts I see that we have very little documentation on it. I think it would be beneficial to provide at least some documentation on it. At the very least some basic information even if it's just a link to `billboard.js`, or something like that. If I wanted to use the component now, I wouldn't know how to really start and use it.

I quickly looked over their documentation, and I got to a conclusion that we could provide information that leverages `billboard.js`'s docs and write something user-friendly, even if we do duplicate some of the information. 
Information like where to look for docs, some most common props, and also a list with a short description of the extra charts we do provide.

I'd do this myself but I've never worked with Charts, so I'd need some guidance if I wanted to tackle it.

kresimir-coko:
Fixes #3714 

I went ahead and made this super simple in the end after pondering on the offered advice. Only meaningful thing I added is the notice that we're switching to Recharts, which from what I understand we will make a wrapper around, similar to how we did with Billboard?

Other than that, @matuzalemsteles :
> maybe write a documentation explaining better the examples we have in the Storybook and how you can reach the Lexicon specification with recharts.

Not a 100% sure what you meant with this, I do understand the idea behind it, but not sure how to explain the examples as they're pretty basic Rechart charts. As for reaching Lexicon specs, shouldn't Lexicon be enough for that?

### DIFF
--- a/packages/clay-charts/README.mdx
+++ b/packages/clay-charts/README.mdx
@@ -13,9 +13,17 @@ import {Chart, GeoMap, Predictive} from '$packages/clay-charts/docs/index';
 -   [Basic](#basic)
 -   [GeoMap](#geomap)
 -   [Predictive](#predictive)
--   [API](#api)
 
 </div>
+</div>
+
+<div class="clay-site-alert alert alert-warning">
+	Charts in its current state will be deprecated soon in favor of a wrapper
+	around <a href="https://recharts.org/">Recharts</a>. See examples of charts
+	built with recharts on&nbsp;
+	<a href="https://storybook.clayui.com/?path=/story/demos-recharts--line">
+		Storybook
+	</a>.
 </div>
 
 ## Basic
@@ -29,19 +37,3 @@ import {Chart, GeoMap, Predictive} from '$packages/clay-charts/docs/index';
 ## Predictive
 
 <Predictive />
-
-## API
-
-### Basic
-
-Clay Charts is primarily a wrapper around the [Billboard.js charting library](https://naver.github.io/billboard.js/), see Billboard.js API for more in-depth API.
-
-<div>[APITable "clay-charts/src/index.tsx"]</div>
-
-### GeoMap
-
-<div>[APITable "clay-charts/src/GeoMap.tsx"]</div>
-
-### Predictive
-
-<div>[APITable "clay-charts/src/Predictive.tsx"]</div>
--- a/packages/clay-charts/docs/api-chart.mdx
+++ b/packages/clay-charts/docs/api-chart.mdx
@@ -1,0 +1,30 @@
+---
+title: 'Chart'
+description: 'Charts is a React wrapper around the Billboard.js library with a few additional charts provided.'
+lexiconDefinition: 'https://liferay.design/lexicon/core-components/charts/'
+mainTabURL: 'docs/components/charts.html'
+---
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
+-   [Basic](#basic)
+-   [GeoMap](#geomap)
+-   [Predictive](#predictive)
+
+</div>
+</div>
+
+Clay Charts is primarily a wrapper around the [Billboard.js charting library](https://naver.github.io/billboard.js/), see Billboard.js API for more in-depth API.
+
+## Basic
+
+<div>[APITable "clay-charts/src/index.tsx"]</div>
+
+## GeoMap
+
+<div>[APITable "clay-charts/src/GeoMap.tsx"]</div>
+
+## Predictive
+
+<div>[APITable "clay-charts/src/Predictive.tsx"]</div>
--- a/packages/clay-charts/src/BillboardWrapper.tsx
+++ b/packages/clay-charts/src/BillboardWrapper.tsx
@@ -15,7 +15,7 @@ const BillboardWrapper: React.FunctionComponent<IProps> = ({
 	forwardRef,
 	elementProps = {},
 	...otherProps
-}) => {
+}: IProps) => {
 	const elementRef = React.useRef<HTMLDivElement>(null);
 
 	const updateChart = React.useCallback((args: any) => {
--- a/packages/clay-charts/src/GeoMap.tsx
+++ b/packages/clay-charts/src/GeoMap.tsx
@@ -235,7 +235,7 @@ const Geomap: React.FunctionComponent<IProps> = ({
 	elementProps = {},
 	forwardRef,
 	...otherProps
-}) => {
+}: IProps) => {
 	const elementRef = React.useRef<HTMLDivElement>(null);
 
 	React.useEffect(() => {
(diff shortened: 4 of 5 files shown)


matuzalemsteles:
>  I understand we will make a wrapper around, similar to how we did with Billboard?

We were thinking of doing that but for now, we will not be working on it. There's more context here https://github.com/liferay/clay/pull/2779.

> Not a 100% sure what you meant with this, I do understand the idea behind it, but not sure how to explain the examples as they're pretty basic Rechart charts. As for reaching Lexicon specs, shouldn't Lexicon be enough for that?

The idea is to create a small mini-tutorial how people can get to the Chart UI that Lexicon defines, for now, our examples in the Storybook are still very simple and do not arrive in the expected look

- [Clay Bar](https://storybook.clayui.com/?path=/story/demos-recharts--bar)
- [Lexicon Bar](https://liferay.design/lexicon/core-components/charts/chart-bar/)

kresimir-coko:
Oh okay, so I can do this by myself attempting to make Recharts fit Lexicon specs, alright @matuzalemsteles good to know, I'll do that then.

kresimir-coko:
## Notes

- The lines can have dashed lines, but the chart legend cannot have a dashed representation like it's shown in Lexicon
- To make dashed lines, we need to add a `strokeDasharray` prop on the `Line` that we want to have a dashed line. The value of the prop determines the length of the line and the spacing between lines
- The "dot" on the line cannot be customized easily, we would need to make a custom `Dot` component that could have a few variants to fit the Lexicon spec
- @bryceosterhaus made a custom Tooltip component, which needs some work on the styling side, the left margin/padding is bigger than on Lexicon
- The `Legend` component requires inline styles to be passed through it's `wrapperStyle` prop
- X and Y Axes and the labels related to them will need to be styled by us to achieve Lexicon looks

P.S. I did some work on this, to get it closer to Lexicon, but some things from above remain to be done. I'd love to hear your thoughts @matuzalemsteles, @pat270 & @bryceosterhaus before going further, for example if I should tweak the CSS mentioned above, etc.

matuzalemsteles:
Hey @kresimir-coko, these are some of the things where I think we should have some kind of wrapper over recharts because we need to make a lot of fine adjustments to the CSS to look like the Lexicon. I think that even if we try to add a tutorial/guides to help them reach the Lexicon spec, we will ask them to do a lot more things on the CSS side. @bryceosterhaus maybe we should start reevaluating this what do you think?

bryceosterhaus:
@matuzalemsteles I've talked to a few teams using recharts and lexicon and one thing that is pretty consistent is that each team has slightly different use cases which needs custom tweaking for their chart. Initially when I investigated the charting issues, it became apparent that it was incredibly difficult to create a "one size fits all" solution. Initially I thought creating a wrapper around Recharts, but that seemed to run into the same issues.

Really I think the best way forward is to help teams create the implementation they need, and then typically they get the hang of recharts and no longer need our help. I also try to create any demos in our storybooks for any odd edge cases that teams need. So really, I don't think trying to create and maintain a wrapper is a great way forward, I think its best to provide examples and help teams use Recharts.

packages/demos/stories/Recharts.tsx
```diff
@@ -85,6 +85,33 @@ const DATA = [
 ];
 
 storiesOf('Demos|Recharts', module)
+	.add('lexicon', () => (
```
bryceosterhaus, line 88: Technically I think all of these demos would be "Lexicon" since they try to follow that specification. If we need to update our line chart, lets just do that instead of making a new chart.

packages/clay-charts/README.mdx
```diff
@@ -13,11 +13,19 @@ import {Chart, GeoMap, Predictive} from '$packages/clay-charts/docs/index';
...
 -   [GeoMap](#geomap)
 -   [Predictive](#predictive)
--   [API](#api)
 
 </div>
 </div>
 
+<div class="clay-site-alert alert alert-warning">
+	Charts in its current state will be deprecated soon in favor of a wrapper
+	around <a href="https://recharts.org/">Recharts</a>. See examples of charts
```
bryceosterhaus, line 22:
```suggestion
	Charts in its current state will be deprecated soon in favor using <a href="https://recharts.org/">Recharts</a>. See examples of charts
```

matuzalemsteles:
> @matuzalemsteles I've talked to a few teams using recharts and lexicon and one thing that is pretty consistent is that each team has slightly different use cases which needs custom tweaking for their chart. Initially when I investigated the charting issues, it became apparent that it was incredibly difficult to create a "one size fits all" solution. Initially I thought creating a wrapper around Recharts, but that seemed to run into the same issues.

Yeah, I believe that each team has some differences but I think that there may be points in common, when I say that we would create a wrapper it could be just a low-level component that already has some customizations, for example:

- Borders
- Subtitles
- Axis Y and X visuals
- Fonts and spacings

I think that just creates a foundation for teams to create on top of Lexicon specifications without harming it.

> Really I think the best way forward is to help teams create the implementation they need, and then typically they get the hang of recharts and no longer need our help. I also try to create any demos in our storybooks for any odd edge cases that teams need. So really, I don't think trying to create and maintain a wrapper is a great way forward, I think its best to provide examples and help teams use Recharts.

We can continue with this and guide the teams

kresimir-coko:
1. First I tried making just a wrapper that we use in stories like below, but that obviously didn't work as I couldn't send any props down to the `LineChart` that way.
```jsx
<ClayRechartsWrapper props>
   	{children}
</ClayRechartsWrapper>
```
2. Then I focused on trying to fix problems via CSS but since that didn't work (due to the whole chart being an `svg`) I went to the next step
3. The latest idea I had was to add the `LineChart` directly into the `Wrapper`, like this:
```jsx
const ClayRechartsLineChart: React.FunctionComponent<IProps> = ({
	displayAxisLines,
	labelDistance
}) => (
	<div className="c-recharts-wrapper">
		<LineChart data={data}>
			...
			<XAxis
				axisLine={displayAxisLines}
				dataKey="name"
			/>
			...
			<Line
				dataKey="uv"
				dot={{stroke: colors[0], strokeWidth: 2}}
				stroke={colors[0]}
				strokeDasharray="5 5"
			/>
			...
		</LineChart>
	</div>
);
```

## Notes
- It seems that styling these might be a problem, as they're all svg elements like `text`, `tspan`, `line`, `g`
- I went ahead and sent the last step, it's most definitely not done by any definition of the word, but just so you guys can see my progress
- I think that **step 3** is the way to go, with the caveat of somehow (IDK how yet) to generalize it so that we don't have to have all Charts in the wrapper component

bryceosterhaus:
@kresimir-coko I think we can worry about the wrapper/non-wrapper stuff in a different issue since it will take more investigation and working with teams using recharts. For now, I think its best to keep it simple and just note about using recharts over clay-charts

kresimir-coko:
@bryceosterhaus 
> just note about using recharts over clay-charts

I've reverted back to the first commit here and pushed that. I think that should be enough for now, like you said, and open a new issue that will focus on adding a wrapper so we don't bloat this issue.

bryceosterhaus:
Looks good! Thanks @kresimir-coko

PR #3721, opened 2020-09-24
DO NOT MERGE: testing the Drilldown component's API Table generation

matuzalemsteles:
This is just a test to see if this resolves the error of the [Drilldown API](https://next.clayui.com/docs/components/drop-down/api.html#dropdownwithdrilldown), this is just an error in production but locally I cannot see this error in the development environment or in the build that is generated for production.

### DIFF
--- a/clayui.com/react-docgen/package.json
+++ b/clayui.com/react-docgen/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "@babel/core": "^7.4.4",
     "@babel/runtime": "^7.0.0",
-    "ast-types": "^0.12.4",
+    "ast-types": "^0.14.2",
     "async": "^2.1.4",
     "commander": "^2.19.0",
     "doctrine": "^3.0.0",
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
Just a follow-up comment:

The error that is happening in the generation of the `DropDownWithDrilldown` API Table, seems to be related to `ast-types` that the version that react-docgen was using did not yet support `ChainExpression`.

```
did not recognize object of type "ChainExpression"
```

I updated the version but still had some other problems, as we had brought the react-docgen codebase to our repository to correct some errors that were still pending in react-docgen PRs, apparently, they were merged and released, so I'm removing react -docgen from our codebase and testing with its latest version but I still have some other errors going on.

```
Cannot read property 'toString' of undefined
```

It seems something related to the new ChainExpression (handleTSIndexedAccessType), although just updating I think that some changes still need to be made in react-docgen to deal with some edge cases.

matuzalemsteles:
I realized that this `toString` error is not related to the previous error, Bryce had already worked on this for some time but his PR was left unanswered recently. https://github.com/reactjs/react-docgen/pull/400

matuzalemsteles:
Closing this to send a fix to another PR.

PR #3723, opened 2020-09-24
fix(@clayui/css): Atlas Custom Radio don't use % for icon background-size. Chrome has trouble rendering it when whitespace is removed.

pat270:


matuzalemsteles (approved):
LGTM!

PR #3729, opened 2020-09-28
fix(clayui.com): Fixes error when trying to generate the ClayDropDownWithDrilldown component API Table

matuzalemsteles:
I had done some tests before and left some comments here #3721. What I did here was just update the react-docgen codebase to the latest one, we should try to see if we can keep up with @bryceosterhaus PR https://github.com/reactjs/react-docgen/pull/400 to just use the npm package.

### DIFF
--- a/clayui.com/plugins/gatsby-remark-api-table/index.js
+++ b/clayui.com/plugins/gatsby-remark-api-table/index.js
@@ -106,7 +106,7 @@ module.exports = ({markdownAST}) => {
 						// eslint-disable-next-line require-atomic-updates
 						node.value = `<ul class="list-api">
 							<li class="list-api-item">
-							<code class="list-api-item-type">{'${error.message}'}</code>
+							<code class="list-api-item-type">{"${error.message}"}</code>
 							</li>
 						</ul>`;
 						// eslint-disable-next-line no-console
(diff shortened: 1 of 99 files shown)


bryceosterhaus:
@matuzalemsteles I recently tried to see if we could avoid using our own fork, but still didn't work from their releases. I'm starting to think the project is just semi-dead at this point...

matuzalemsteles:
> @matuzalemsteles I recently tried to see if we could avoid using our own fork, but still didn't work from their releases. I'm starting to think the project is just semi-dead at this point...

Yeah, I agree. Some days ago I was researching how we could build some better API Tables, with support for importing interfaces, links with more details to get a better experience. But I think we're going to have to create this manually, something like this https://react-spectrum.adobe.com/react-stately/useAsyncList.html#interface. I think it is worth the effort as it is a very important part of our documentation.

bryceosterhaus (approved):
LGTM! Looks like it generates the tables on the deploy preview

wincent:
> I'm starting to think the project is just semi-dead at this point...

You could try mentioning `@danez` directly on it. The person who originally wrote the project left FB and is probably not super engaged with it right now.

Issue #3728, opened 2020-09-28
Icon: Update icon-display-content


hold-shift:
Due to [LEXI-904](https://issues.liferay.com/browse/LEXI-904) we've updated the display-content icon

<img width="209" alt="Screenshot 2020-05-06 at 10 50 12" src="https://user-images.githubusercontent.com/61543739/94448999-dcc70d00-01ab-11eb-97fa-1b7c3c22a7c1.png">

Please replace it in the icon library.

Find the SVGs on the ticket.
Thanks!

matuzalemsteles:
Fixed by #3731

Issue #3727, opened 2020-09-28
Icon: Update Product menu icons


hold-shift:
Due to [LEXI-971 ](https://issues.liferay.com/browse/LEXI-971)we've updated the product-menu-open & product-menu-closed icons.

![Captura de pantalla 2020-09-28 a las 16 35 47](https://user-images.githubusercontent.com/61543739/94446265-b2c01b80-01a8-11eb-84c1-3ddc9002fb78.png)

Please replace it in the icon library.

Find the SVGs on the ticket.
Thanks!

matuzalemsteles:
Fixed by #3731

PR #3731, opened 2020-09-28
fix(@clayui/css): update icons

### LINKED ISSUE #3726: Icon: Update Hidden Icon
Due to [LEXI-1017](https://issues.liferay.com/browse/LEXI-1017) we've updated the hidden icon.

Captura de pantalla 2020-09-28 a las 16 24 38

Please replace it on

 the icon library.

Find the SVG on the ticket.
Thanks!

bryceosterhaus:
fixes #3726 #3727 #3728

@pat270, do I need to update icon info anywhere else? I thought I remembered needing to update some other file.


pat270:
@bryceosterhaus We don't need to update anything else. We only need to update aliases on clayui.com if needed. 

I removed some white space and added class names where applicable.

bryceosterhaus:
Thanks for the update! I think the alias should be good since it was just updating an existing icon

PR #3732, opened 2020-09-28
feat(@clayui/color-picker): add ability to use named colors

### LINKED ISSUE #3681: @clayui/color-picker Should accept valid css values for the color property
Spawned from:
https://github.com/liferay/clay/issues/3651#issuecomment-680968921
https://github.com/liferay-frontend/liferay-portal/pull/348#issuecomment-685760858

CSS Color names:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Color

Other valid values:
```
inherit
initial
unset
transparent
var(--my-custom-propery-var) // css custom properties
currentColor // on svg element
```

These should be implemented separately from `rgba` and custom color names since no conversion is needed. We can output the value straight to the browser.

bryceosterhaus:
fixes #3681

Are there any edge cases I missed here? You should now be able to used named colors and hex values

### DIFF
--- a/packages/clay-color-picker/src/index.tsx
+++ b/packages/clay-color-picker/src/index.tsx
@@ -3,8 +3,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import DropDown from '@clayui/drop-down';
+import ClayDropDown from '@clayui/drop-down';
 import {ClayInput} from '@clayui/form';
+import ClayIcon from '@clayui/icon';
 import {FocusScope, sub} from '@clayui/shared';
 import React from 'react';
 import tinycolor from 'tinycolor2';
@@ -12,7 +13,6 @@ import tinycolor from 'tinycolor2';
 import Basic from './Basic';
 import Custom from './Custom';
 import Splotch from './Splotch';
-import {useHexInput} from './hooks';
 
 const DEFAULT_COLORS = [
 	'000000',
@@ -72,7 +72,6 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	 * Flag for adding ColorPicker in disabled state
 	 */
 	disabled?: boolean;
-
 	/**
 	 * The label describing the collection of colors in the menu
 	 */
@@ -137,7 +136,6 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 	disabled,
 	label,
 	name,
-	onBlur = () => {},
 	onColorsChange,
 	onValueChange = () => {},
 	showHex = true,
@@ -149,7 +147,9 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 	value = 'FFFFFF',
 	...otherProps
 }: IProps) => {
-	if (value.indexOf('#') === 0) {
+	const isHex = tinycolor(value).getFormat() === 'hex';
+
+	if (isHex && value.indexOf('#') === 0) {
 		value = value.slice(1);
 	}
 
@@ -175,13 +175,6 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 	const splotchRef = React.useRef<HTMLButtonElement>(null);
 
 	const [active, setActive] = React.useState(false);
-	const [hexInputValue, setHexInputValue] = useHexInput(value);
-
-	React.useEffect(() => {
-		if (document.activeElement !== inputRef.current) {
-			setHexInputValue(value);
-		}
-	}, [value]);
 
 	return (
 		<FocusScope arrowKeysUpDown={false}>
@@ -201,7 +194,7 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 							width: 0,
 						}}
 						type={useNative ? 'color' : 'text'}
-						value={hexInputValue ? `#${hexInputValue}` : ''}
+						value={value ? `${isHex ? '#' : ''}${value}` : ''}
 					/>
 				)}
 
@@ -229,7 +222,7 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 						</ClayInput.GroupText>
 					</ClayInput.GroupItem>
 
-					<DropDown.Menu
+					<ClayDropDown.Menu
 						active={active}
 						alignElementRef={triggerElementRef}
 						className="clay-color-dropdown-menu"
@@ -241,8 +234,8 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 							<Basic
 								colors={colors || DEFAULT_COLORS}
 								label={label}
-								onChange={(value) => {
-									onValueChange(value);
+								onChange={(newVal) => {
+									onValueChange(newVal);
 									setActive((val: boolean) => !val);
 
 									if (splotchRef.current) {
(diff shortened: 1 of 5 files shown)


pat270:
This is a lot more useable than before. I'm unable to use `inherit` and `var(--my-custom-property)`. If we are able to include these values, it should cover most use cases in Portal.

bryceosterhaus:
@pat270 I'm not sure we will be able to support either of those. If this package was explicitly a css color picker, it might make sense, but I think it would be odd to support both of those when they rely on context outside of the component.

For example, `inherit` would have to have knowledge about the css on the page in order to know how to display the color. And then for `var(--my-custom-property)`, you also comes from an outside context of the component.

Does that make sense? I'm trying to wrap my head around how and why we support those use cases for non-css implementations.

pat270:
This is confusing since this relies on the browser? We need HTML + CSS to render the color picker. We should have at least some kind of option or attribute we can set to allow valid CSS values to pass through the input. Not allowing these limits what you can do with the stylebook feature in Portal.

bryceosterhaus:
@pat270 yeah thats true. Let me take a look at the stylebook feature in portal, I'm not familiar with it, so maybe it'll help me think of something.

I'll also take a look and see if and how other color-pickers allow this.

bryceosterhaus:
@pat270 I was able to come up with something. I didn't realize it was using used as a css selector for style book.

This may lead to some odd non-css scenarios, but I think its fairly safe.

packages/clay-color-picker/src/index.tsx
```diff
@@ -280,44 +290,43 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
...
 					)}
 				</ClayInput.Group>
+
+				{invalidColor && (
+					<ClayForm.FeedbackGroup>
+						<ClayForm.FeedbackItem>
+							{feedbackText}
+						</ClayForm.FeedbackItem>
+					</ClayForm.FeedbackGroup>
+				)}
```
matuzalemsteles, line 329: @bryceosterhaus I didn't get to check it out but whoever uses the ColorPicker will probably have to know when it is invalid to correctly format the validation markup as shown here https://next.clayui.com/docs/components/form.html#validation.

I think it would be better to expose an `onValidation` to inform which error happened and let the developer take care of that mainly so we don't start adding more use cases within that component. What do you think?

packages/clay-color-picker/src/index.tsx
```diff
@@ -280,44 +290,43 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
...
 							/>
-
-							<ClayInput.GroupInsetItem before tag="label">
-								{'#'}
-							</ClayInput.GroupInsetItem>
+							{isHex && (
+								<ClayInput.GroupInsetItem before tag="label">
+									{'#'}
+								</ClayInput.GroupInsetItem>
+							)}
```
matuzalemsteles, line 318: It would be interesting to have some other symbol for the other cases so that there are no blinks to remove the element and then put it back on when you keep switching.

packages/clay-color-picker/src/index.tsx
```diff
@@ -149,7 +160,13 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 	value = 'FFFFFF',
 	...otherProps
 }: IProps) => {
-	if (value.indexOf('#') === 0) {
+	const [invalidColor, setInvalidColor] = React.useState(false);
+
+	const colorVal = tinycolor(value);
+
+	const isHex = colorVal.getFormat() === 'hex';
```
matuzalemsteles, line 167: Maybe just simplify for one line?
```suggestion
	const isHex = tinycolor(value).getFormat() === 'hex';
```

matuzalemsteles (commented):
I had a different impression about these changes but it really seems simpler now.

packages/clay-color-picker/src/index.tsx
bryceosterhaus: @matuzalemsteles oh yeah thats a great idea, i'll switch that up to be a callback and allow them to insert their own content.

packages/clay-color-picker/src/index.tsx
bryceosterhaus: Good idea, ill check out other icon examples

packages/clay-color-picker/src/index.tsx
bryceosterhaus: Actually, since the user has the `onChange` callback, there is no need to provide internal validation. The user can decide what is a valid color, rather than us deciding that for them

bryceosterhaus:
@matuzalemsteles 

Added
- `feedbackContent` prop in case the user wants to validate the input.
- Instead of removing the `#` I am replacing it with an icon with the `color-picker` symbol.

packages/clay-color-picker/src/index.tsx
```diff
@@ -281,43 +286,30 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
...
+										style={{width: 10}}
+										symbol="color-picker"
+									/>
+								)}
 							</ClayInput.GroupInsetItem>
 						</ClayInput.GroupItem>
 					)}
 				</ClayInput.Group>
+
+				{feedbackContent}
```
matuzalemsteles, line 312: @bryceosterhaus to prevent us from adding this new API maybe the dev can compose with `ClayForm.Group` to create the validation on the outside, this can work better and give the more control.

```jsx
<ClayForm.Group className="has-error">
	<ClayColorPickerWithState
		{...props}
		colors={customColors}
		onColorsChange={setCustoms}
		onValueChange={setColor}
		value={color}
	/>
	<ClayForm.FeedbackGroup>
		<ClayForm.FeedbackItem>
			{"This is a form-feedback-item."}
		</ClayForm.FeedbackItem>
		<ClayForm.FeedbackItem>
			<ClayForm.FeedbackIndicator
				spritemap={spritemap}
				symbol="exclamation-full"
			/>
			{"This is a form-feedback-item with a error feedback indicator."}
		</ClayForm.FeedbackItem>
	</ClayForm.FeedbackGroup>
</ClayForm.Group>
```

matuzalemsteles: When I looked at it I was even thinking that the `title` API could also be avoided 🤔

packages/clay-color-picker/src/index.tsx
```diff
@@ -281,43 +286,30 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
...
 							<ClayInput.GroupInsetItem before tag="label">
-								{'#'}
+								{isHex ? (
+									'#'
+								) : (
+									<ClayIcon
+										spritemap={spritemap}
+										style={{width: 10}}
+										symbol="color-picker"
+									/>
```
matuzalemsteles, line 305: Cool!

packages/clay-color-picker/src/index.tsx
```diff
@@ -88,6 +93,11 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	 */
 	onColorsChange?: (val: Array<string>) => void;
 
+	/**
+	 * Callback for when the validity of the color value changes
+	 */
+	onValidation?: (val: boolean) => void;
+
```
matuzalemsteles, line 100: You probably forgot to remove 😅

matuzalemsteles (commented):
In general, it looks good, I added a comment, let me know what you think.

packages/clay-color-picker/src/index.tsx
bryceosterhaus: yep! I rely on eslint too much sometimes lol

packages/clay-color-picker/src/index.tsx
bryceosterhaus: Since we already provide the title API, I'm just gonna leave that as is, but I like the idea with the feedback. I'm gonna create a storybook example for that as well.

bryceosterhaus:
@matuzalemsteles updated!

packages/clay-color-picker/src/index.tsx
matuzalemsteles: Yeah, maybe we can plan to remove the `title` API in an upcoming major version.

matuzalemsteles (approved):
Cool, great work! LGTM.

PR #3738, opened 2020-10-01
fix(@clayui/drop-down): Increase clickable area of text and icon

kresimir-coko:
Alrighty, it took me a bit of digging through my notes to find that in order to see this example I need to fire up a gulp server inside clay-css. We should definitely add that into the GitHub readme thingy.

Once I did get it fired up I realized the main culprit, `dropdown-item` is always the parent of the item, and always has the same `padding`, which makes sense. What doesn't make sense is that it's child in case of simple text, like `Folder` in the above example doesn't have an HTML element wrapped around it, so it's just `<div class="dropdown-item">Folder</div>`.

My proposal is to add another layer of HTML, like a `span` to that `Folder` text, so that we can target `dropdown-item` with CSS and remove the `padding` from it. 

I sent a WIP example of the styles just to get feedback from you @pat270 if I should be doing it in a different file, or a certain way with these variables and imports, haven't worked in a proper CSS project yet.

I also sent a commit that fixes a couple small things in Drilldown, lemme know what you feel about them @bryceosterhaus

### DIFF
--- a/packages/clay-css/src/content/dropdowns_drill_down.html
+++ b/packages/clay-css/src/content/dropdowns_drill_down.html
@@ -166,7 +166,7 @@ section: Components
 							<li>
 								<div class="dropdown-item">
 									<a class="dropdown-item-indicator-text-end" href="#1">Document</a>
-									<button class="btn btn-unstyled dropdown-item-indicator-end" data-drilldown="next" data-target="#_s950dwymf" role="button">
+									<button class="btn btn-unstyled dropdown-item-indicator-end" data-drilldown="next" data-target="#_s950dwymf" type="button">
 										<svg class="lexicon-icon lexicon-icon-angle-right" focusable="false" role="presentation">
 											<use xlink:href="{{rootPath}}/images/icons/icons.svg#angle-right" />
 										</svg>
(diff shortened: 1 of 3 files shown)


pat270:
@kresimir-coko I was thinking about using a hack on `.drilldown .dropdown-item-indicator-text-end`:

```
  margin-top: -0.59375rem;
  margin-bottom: -0.59375rem;
  margin-left: -1rem;
  padding-left: inherit;
  padding-top: inherit;
  padding-bottom: inherit;
  width: auto;
```

We need to make sure we use variables and not fixed values, since a variable change will break it.

kresimir-coko:
@pat270 how did u arrive at this hack, specifically, how did u come to the very specific `-0.59375rem` ? I tried it out and it works. I've force pushed a commit with your hack. I've fallen out of touch with my CSS 😢 

I also noticed that the back button here has a small clickable area, so I tried messing around with it but failed to arrive to an acceptable solution.

image

jbalsas:
> @pat270 how did u arrive at this hack, specifically, how did u come to the very specific -0.59375rem ? I tried it out and it works. I've force pushed a commit with your hack. I've fallen out of touch with my CSS 😢

magicspell

pat270:
@kresimir-coko `0.59375rem` is the top/bottom padding on the `dropdown-item`. The negative margin offsets the padding and inheriting it applies it on `dropdown-item-indicator-text-end`. I learned this from Bootstrap's `row` and `col` components.

pat270 (approved):

PR #3740, opened 2020-10-05
feat(clayui.com): Put form components under Form folder in the navigation

### LINKED ISSUE #3736: Should we put all @clayui/form components under a Form section in the navigation
I feel like, for clarity, we should add all components that are a part of the `@clayui/form` package under a Form category in the navigation. I, myself, have been confused when importing packages like Select, Checkbox, Input to discover that they're under the `@clayui/form`package.

This is shown semi-prominently in those components, but when clicking around in the UI, and going straight down to the examples or API, I often find myself skipping this heading.

image

I think that putting these child components into their parent tab, which corresponds the package they belong to, will be enough for improving that clarity and hierarchy.

kresimir-coko:
Fixes #3736 

I'm pretty happy with this result, and I learned how the navigation get's prepared and filled up.

It's pretty simple, even though i took me a couple days to get here, but I did the following:
- added a `formPackageMember` to the frontmatter of components that belong to the `@clayui/form` package
- made Form a folder by assigning it in `getSection`, not sure if we can add that too to its frontmatter
- made the Form always open by adding `alwaysActive` to its frontmatter
- move components that belong to `@clayui/form` to that folder and remove them from the regular nav

### DIFF
--- a/clayui.com/src/utils/getSection.js
+++ b/clayui.com/src/utils/getSection.js
@@ -20,6 +20,10 @@ const sortByOrderAndTitle = (a, b) => {
 	}
 };
 
+const navigationRelationships = {
+	'/docs/components/form.html': [],
+};
+
 const toSectionElements = (
 	slug,
 	pathname,
@@ -38,7 +42,6 @@ const toSectionElements = (
 	const link = `/${slug}`;
 	const parentLink = `/${slug.substring(0, slug.lastIndexOf('/') + 1)}`;
 	const isFolder = lastSlug === 'index';
-	const isFormFolder = lastSlug === 'form';
 	const isRoot =
 		(slugs.length === 3 && isFolder) || (slugs.length === 2 && !isFolder);
 
@@ -48,7 +51,6 @@ const toSectionElements = (
 		id,
 		indexVisible,
 		isFolder,
-		isFormFolder,
 		isRoot,
 		link,
 		navigationParent,
@@ -60,8 +62,14 @@ const toSectionElements = (
 };
 
 const toSectionItem = (item, paths) => {
-	if (item.isFormFolder) {
-		item.isFolder = true;
+	const isParentPackage = Object.keys(navigationRelationships).includes(
+		`/${item.pathname}`
+	);
+
+	if (item.navigationParent) {
+		navigationRelationships[item.navigationParent].push(
+			`/${item.pathname}`
+		);
 	}
 
 	if (item.isFolder) {
@@ -75,14 +83,14 @@ const toSectionItem = (item, paths) => {
 			.filter((path) => !path.navigationParent)
 			.map((path) => toSectionItem(path, paths))
 			.sort(sortByOrderAndTitle);
+	}
 
-		if (item.isFormFolder) {
-			item.items = paths
-				.filter((path) => path.navigationParent)
-				.filter((path) => path.link !== item.link)
-				.map((path) => toSectionItem(path, paths))
-				.sort(sortByOrderAndTitle);
-		}
+	if (isParentPackage) {
+		item.items = paths
+			.filter((path) => path.navigationParent)
+			.filter((path) => path.link !== item.link)
+			.map((path) => toSectionItem(path, paths))
+			.sort(sortByOrderAndTitle);
 	}
 
 	return item;
--- a/clayui.com/src/utils/getSection.js
+++ b/clayui.com/src/utils/getSection.js
@@ -27,7 +27,7 @@ const toSectionElements = (
 	order,
 	alwaysActive,
 	draft,
-	formPackageMember,
+	navigationParent,
 	indexVisible
 ) => {
 	const slugs = slug.split('/');
@@ -45,13 +45,13 @@ const toSectionElements = (
 	return {
 		alwaysActive,
 		draft,
-		formPackageMember,
 		id,
 		indexVisible,
 		isFolder,
 		isFormFolder,
 		isRoot,
 		link,
+		navigationParent,
 		order,
 		parentLink,
 		pathname,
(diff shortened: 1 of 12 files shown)


bryceosterhaus (commented):
This looks pretty good! I think it actually looks nice having the form components nested and its nice that the URLs are all the same. @matuzalemsteles what do you think?

One thought on the code, instead of hard coding frontmatter for `formPackageMember` could we do something like `navigationParent` and then specify the parent page? That way in the future, if we had another collection of pages, we could easily just add the frontmatter to nest it under another page. For this use case then, it would look like `navigationParent: /docs/components/form.html`

matuzalemsteles:
> This looks pretty good! I think it actually looks nice having the form components nested and its nice that the URLs are all the same. @matuzalemsteles what do you think?

I don't see any problems with that I like it too 🙂.

> One thought on the code, instead of hard coding frontmatter for formPackageMember could we do something like navigationParent and then specify the parent page? That way in the future, if we had another collection of pages, we could easily just add the frontmatter to nest it under another page. For this use case then, it would look like navigationParent: /docs/components/form.html

Yeah, I would say that we should always try to avoid adding specific cases to our page creation or to queries in GraphQL.

kresimir-coko:
Hey @bryceosterhaus & @matuzalemsteles I made the suggested changes. I tried to add a `parentPackage` property to forms `README.mdx` but didn't have time to figure it out fully. The goal of it would be so that that we don't have to check for `item.isFormFolder`. I sent a commit without that, so if you guys think I should pursue that I can do it tomorrow. Otherwise, it should be good to go now 😄

clayui.com/src/utils/getSection.js
```diff
@@ -56,6 +60,10 @@ const toSectionElements = (
 };
 
 const toSectionItem = (item, paths) => {
+	if (item.isFormFolder) {
+		item.isFolder = true;
+	}
+
```
matuzalemsteles, line 66: I think we can remove `isFormFolder` from here and work only with the `navigationParent` information to create the grouping, you can try to do the following.

- Sweep the items' Array and create an object by grouping the corresponding paths to use later
  ```js
  {
     'path/some': ['path/children', '']
  }
  ```
- Sweep the elements to group, look for the path corresponding to the object created previously, and add all the corresponding elements of that path
  ```js
  elements
	.map((item, index, arr) => {
		const path = Object.keys(groups).find(path => path === item.slug);

		if (path) {
			return {
				...item,
				items: groups[path].map(path => arr.find(item => item.slug === path))
			}
		}

		return item;
	})
  ```
- Sweep elements to filter and remove any elements with `navigationParent`

I have not tested these examples above and it is also likely that you will have to do something else.

matuzalemsteles (commented):
@kresimir-coko I think we can just make some adjustments before we go with that.

kresimir-coko:
Hey @matuzalemsteles I made it a bit more generalized I think with the most recent commit. I think it can be improved but I've exhausted all my mental capabilities for today, this one really took a toll on me, but it was worth it 😄 .

The idea was that in the future if we wanted to add new ones, along with adding the parent path in the frontmatter, we would also add an entry in this object below, which I use in the implementation further along:
```jsx
const navigationRelationships = {
	'/docs/components/form.html': [],
};
```

bryceosterhaus:
@kresimir-coko I just pushed an update to your branch, I think with these changes we can make it more dynamic, now every time that frontmatter is added, it will dynamically create the structure for us.

clayui.com/src/utils/getSection.js
```diff
@@ -61,42 +57,32 @@ const toSectionElements = (
...
 			.sort(sortByOrderAndTitle);
 	}
 
-	if (isParentPackage) {
+	if (parentPaths.includes(item.pathname)) {
 		item.items = paths
-			.filter((path) => path.navigationParent)
-			.filter((path) => path.link !== item.link)
-			.map((path) => toSectionItem(path, paths))
+			.filter(path => path.navigationParent === item.pathname)
```
bryceosterhaus, line 76: This check ensures that only the items with that navigationParent get added to the right path, otherwise they get added to every parent.

clayui.com/src/utils/getSection.js
```diff
@@ -122,10 +108,16 @@ const getSection = (data) => {
 		);
 	});
 
-	const rootElements = elements.filter((path) => path.isRoot);
+	const rootElements = elements.filter(path => path.isRoot);
+
+	const parentPaths = elements.reduce((acc, {navigationParent}) => {
```
bryceosterhaus, line 113: I pulled this outside of `toSectionItem` to reduce the amount of times it would be called. Otherwise we would have to check on every call to that function

clayui.com/src/utils/getSection.js
```diff
@@ -20,10 +20,6 @@ const sortByOrderAndTitle = (a, b) => {
 	}
 };
 
-const navigationRelationships = {
-	'/docs/components/form.html': [],
-};
```
bryceosterhaus: this is now no longer needed since we can just check if any of the child elements have the path as a parent

packages/clay-form/CHECKBOX.mdx
```diff
@@ -3,7 +3,7 @@ title: 'Checkbox'
 description: 'A checkbox is a component that lets the user select the value that is written in its corresponding text label. A user can select multiple checkboxes from a group at the same time.'
 lexiconDefinition: 'https://liferay.design/lexicon/core-components/forms/radio-check-toggle/'
 packageNpm: '@clayui/form'
-navigationParent: '/docs/components/form.html'
+navigationParent: 'docs/components/form.html'
```
bryceosterhaus, line 6: I removed the leading `/` so that we don't have to manually add that when comparing paths

matuzalemsteles (commented):
No problem @kresimir-coko! Thanks @bryceosterhaus for simplifying, good job. LGTM! There are only a few formatting errors.

packages/clay-form/CHECKBOX.mdx
kresimir-coko: Yea this is smarter than manually adding it when comparing paths. 👍

clayui.com/src/utils/getSection.js
kresimir-coko: Smart 👍 I was going with what I knew could work.

kresimir-coko:
I've investigated your solution @bryceosterhaus and it's very clean, I like it. LGTM 👍

PR #3746, opened 2020-10-09
fix(@clayui/form): Fix custom radio button image

eduardoallegrini:
In the [ticket LPS-121351](https://issues.liferay.com/browse/LPS-121351), we had a minor visual bug in Windows OS due to the [Bootstrap background image](https://getbootstrap.com/docs/4.5/components/forms/#radios) we were using for the custom radio button and the size of our radio button defined by lexicon. I've updated the image and the size following the [Lexicon specification for radio buttons](https://www.figma.com/file/A7ZNuyNHxj7kzJ6SYwQbQ3/Component-Library-Template-(Intra-Test)?node-id=1096%3A11397) and the [Clay pattern for the custom checkbox](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/atlas/variables/_custom-forms.scss#L105)

__Before:__
Capture2

__After:__
Capture

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_custom-forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_custom-forms.scss
@@ -114,7 +114,8 @@ $custom-checkbox-indicator-border-radius: 0.125rem !default; // 2px
 
 // Custom Radio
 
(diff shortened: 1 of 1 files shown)


pat270:
This LGTM! I can't test Windows Chrome, but I trust it works.

PR #3747, opened 2020-10-09
fix(@clayui/icon): Add svg4everybody to fix long loading of Icons page

### LINKED ISSUE #3744: The Icons page is frozen for a while
When accessing the [Icon component page](https://clayui.com/docs/components/icon.html), my browser freezes for some (5) seconds before becoming responsive.

Haven't fully explored the reasons but I assume we're doing some heavy work with the data or the icons...

icons

kresimir-coko:
Fixes #3744 

Let me know if this is a suboptimal location to invoke `svg4everybody`, I tried it in `IconSearch` as that's the component that gets called but there it didn't do anything.

The red flag to me is the need for updating that many snapshots, but it might be okay.

### DIFF
--- a/clayui.com/package.json
+++ b/clayui.com/package.json
@@ -67,7 +67,8 @@
 		"react-docgen": "file:./react-docgen",
 		"react-dom": "^16.12.0",
 		"react-helmet": "^5.2.0",
-		"react-live": "^2.0.1"
+		"react-live": "^2.0.1",
+		"svg4everybody": "^2.1.9"
 	},
 	"devDependencies": {
 		"cpx": "^1.5.0",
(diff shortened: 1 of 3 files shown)


packages/clay-icon/src/index.tsx
```diff
@@ -26,6 +27,12 @@ const ClayIcon = React.forwardRef<SVGSVGElement, IProps>(
 	({className, spritemap, symbol, ...otherProps}: IProps, ref) => {
 		let spriteMapVal = React.useContext(ClayIconSpriteContext);
 
+		React.useEffect(() => {
+			svg4everybody({
+				polyfill: true,
+			});
+		}, []);
+
```
matuzalemsteles, line 35: @kresimir-coko I was thinking that we would add this only on clayui.com, because we already have it in DXP, so maybe add it to the doc template.

kresimir-coko: Good idea, it didn't occur to me, but it works! 👍

kresimir-coko:
Hey @matuzalemsteles I force pushed a new commit because I had to undo the whole updating snapshots, `yarn.lock`, etc. So I assumed it'd be cleaner to just force push a single commit.

bryceosterhaus (approved):
LGTM!

@matuzalemsteles I'll let you merge it, in case you have any other concerns

matuzalemsteles (approved):
LGTM!

PR #3748, opened 2020-10-19
feat(@clayui/css): add sticky functionality for table headers

### LINKED ISSUE #1244: Table Header fixed
Hey guys,

In the **Analytics Cloud** project, we have a component that uses a table with the clay's markup, however we had to customize this table because we had a need to fix the header as shown the gif below:

table-header-fixed

**Is it possible to create a markup in clay to fix the header?**

cc: @fcodias, @oliveiraaraujo

bryceosterhaus:
fixes #1244

I saw that the definition now included the ability for sticky behavior.

@pat270, you probably have a better idea on this, but thought I would take a stab at it. Let me know if I need to change anything.

### DIFF
--- a/packages/clay-css/src/scss/components/_tables.scss
+++ b/packages/clay-css/src/scss/components/_tables.scss
@@ -28,6 +28,16 @@ th {
 	margin-bottom: $table-margin-bottom;
 	width: 100%;
 
+	&.sticky-head {
+		position: relative;
+
+		thead th {
+			z-index: 1;
+			position: sticky;
+			top: 0;
+		}
+	}
+
 	thead {
 		td,
 		th {
--- a/packages/clay-table/docs/markup-table.md
+++ b/packages/clay-table/docs/markup-table.md
@@ -349,6 +349,162 @@ A table is styled like a list. The active state can be invoked by adding class `
 
 ## Variants(#css-variants)
 
+### Sticky Header
+
+Use `.sticky-head` on your table that has a fixed height and `overflow: auto;` to have the table header stick to the top of the table on scroll.
+
+<div class="sheet-example">
+    <div style="height: 200px; overflow: auto;">
+        <table class="table sticky-head">
+            <thead>
+                <tr>
+                    <th>Country</th>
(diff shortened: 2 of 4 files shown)


pat270:
Hey @bryceosterhaus, Maybe it would be better to have it on `table-responsive` since it needs `overflow: auto;` and `height` to work? Also I checked on Safari 13 and it's not working correctly (might be Safari implementation) I didn't search too hard for the answer.

pat270:
Ok I tested it more, seems like it works on Chrome and Safari with `.sticky-head thead th`. The other selector `.sticky-head thead` breaks it for Safari.

bryceosterhaus:
@pat270 just updated and I think it works on safari now. 

I tried out putting it on table-responsive, but the table needs a wrapper with overflow and a specific height and that didn't seem to work with table-responsive, I could easily be doing something wrong though...

packages/clay-css/src/scss/components/_tables.scss
```diff
@@ -28,6 +28,16 @@ th {
 	margin-bottom: $table-margin-bottom;
 	width: 100%;
 
+	&.sticky-head {
```
matuzalemsteles, line 31: @pat270 @bryceosterhaus do you think it shouldn't have the same nomenclature as the other classes? something like `.table-sticky-head`?

bryceosterhaus: Ill defer to @pat270 for whatever he thinks is best

pat270: Sorry I completely forgot about this. I agree with @matuzalemsteles `table-sticky-head` is consistent with the current naming structure. I'm hesitant to add this because a majority of our tables have `table-responsive` which adds `overflow-x: auto`. `position: sticky` only works when the parent container is `overflow: visible` if the browsers scrollbar is on document. 

My next question is are we ok with sticky header tables overflowing horizontally in smaller screens? In the long run I do think we need to use JS to make the experience worthwhile.

bryceosterhaus:
Going to close this for now until it becomes more of a necessity

PR #3753, opened 2020-10-21
feat(@clayui/card): add API to pass props to the checkbox

### LINKED ISSUE #3752: `ClayCardWithInfo` does not pass `checkboxProps`  to `ClayCheckbox`
- I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

Yes, please see https://codesandbox.io/s/blissful-breeze-395f9

Render a `<ClayCardWithInfo />` with the `checkboxProps` prop., i.e.

```jsx
<ClayCardWithInfo
      checkboxProps={{ name: "checkbox-name", value: "checkbox-value" }}
      spritemap={spritemap}
      name="ClayCardWithInfo"
      onSelectChange={() => {}}
/>
```

The `input` element (which is a `checkbox`) should have the properties you passed in the `checkboxProps` property, but it doesn't because they [aren't](https://github.com/liferay/clay/blob/fbe41067227135f5498e162bc6d6b21ab0ae8b30/packages/clay-card/src/CardWithInfo.tsx#L163-L169) passed to the `ClayCheckBox` component. However, `CardWithUser` [does](https://github.com/liferay/clay/blob/fbe41067227135f5498e162bc6d6b21ab0ae8b30/packages/clay-card/src/CardWithUser.tsx#L126) pass that prop to it's `ClayCheckBox` child component.

matuzalemsteles:
Fixes #3752

I have the feeling that we should be avoiding props that serve to pass props to sub-components nested within the high-level... I don't have a strong opinion on that yet but I think we should put it in mind.

### DIFF
--- a/packages/clay-card/src/CardWithHorizontal.tsx
+++ b/packages/clay-card/src/CardWithHorizontal.tsx
@@ -16,6 +16,11 @@ interface IProps {
 	actions?: React.ComponentProps<typeof ClayDropDownWithItems>['items'];
 
 	/**
+	 * Props to add to the checkbox element
+	 */
+	checkboxProps?: React.HTMLAttributes<HTMLInputElement>;
+
+	/**
 	 * Flag to indicate that all interactions on the card will be disabled.
 	 */
 	disabled?: boolean;
@@ -58,6 +63,7 @@ interface IProps {
 
 export const ClayCardWithHorizontal: React.FunctionComponent<IProps> = ({
 	actions,
+	checkboxProps = {},
 	disabled,
 	dropDownTriggerProps = {},
 	href,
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
LGTM!

> I have the feeling that we should be avoiding props that serve to pass props to sub-components nested within the high-level...

This has also felt kind of odd to me, but I'm also not sure the best way to provide an API for form and accessibility type attributes that are nested deeply in the component.

One nice part about this sort of API is that we only have to add one prop instead of `checkboxName`, `checkboxValue`, `checkboxTitle`, etc.

I'll continue to think about this though, maybe we can discover someone better for v4.

matuzalemsteles:
Yeah, I also don't know yet how we could avoid this in high-level components but just to keep that in mind to have something in v4.

julien:
@matuzalemsteles, @bryceosterhaus thanks for the quick merge.

> Yeah, I also don't know yet how we could avoid this in high-level components but just to keep that in mind to have something in v4.

Sounds like a good topic for an interesting conversation.
If we looked at all the use cases we might be able to find a pattern, ideas, etc.

matuzalemsteles:
@julien Of course, we can talk more about it and explore ideas.

PR #3758, opened 2020-10-22
feat(@clayui/card): add support for base HTMLDivElement and HTMLAnchorElement props in high-level components

### LINKED ISSUE #3757: Add support for id attribute
We need to support pass an `id` to the components to be rendered as the `id` of the element. I identified that you can't do it in `CardWithUser` and `CardWithInfo` but this is probably happening almost everywhere.

Thanks!

matuzalemsteles:
Fixes #3757

I ended up causing the components to extend the base HTML interfaces to extend the number of props that can be passed to the component so we don't have to add it in the future since `<ClayCard />` already does the same.

### DIFF
--- a/packages/clay-card/src/CardWithHorizontal.tsx
+++ b/packages/clay-card/src/CardWithHorizontal.tsx
@@ -12,7 +12,7 @@ import React from 'react';
 
 import ClayCard from './Card';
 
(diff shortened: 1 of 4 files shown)


bryceosterhaus:
lgtm!

Issue #3754, opened 2020-10-21
ClayPanel: can expanding one panel collapse all other panels in a group?


rdai10:
Given the current API, is it possible to achieve the behavior where expanding one panel in a group of panels will collapse all other panels? Like the example illustrated [here](http://pat270.github.io/lexicon/v2.15.1/content/panels/#accordions-with-panels-and-collapse-plugin).

I noticed that each panel seems to be its distinct entity. While there is the `defaultExpanded` prop, it only sets the initial value for the `expanded` state. And the state management is internal to the panel component. Just want to make sure I didn't miss anything.

Thanks!

bryceosterhaus:
@rdai10 I think you are right in seeing that we don't provide a current mechanism for that. I think we made the mistake of not thinking about how someone would control this. I will send a PR to add an API for this. Thanks for catching this!

rdai10:
Thanks @bryceosterhaus !

bryceosterhaus:
@rdai10 sorry this took so long to fix. This should be available in the latest clay release and also in dxp master via https://issues.liferay.com/browse/LPS-125356

PR #3760, opened 2020-10-22
fix: allow for controlling components that expand vertically

bryceosterhaus:
So I am trying to address this issue, #3754, but its proving to be more difficult than I thought. 

### The Problem
When added the hook for transitioning height, we didn't account well for controlled components. The current implementation requires that the `animate` function is called for the height to transition smoothly. In cases where we control the height via a prop, its not easy to transition the height.

I'm curious to get another set of eyes on this and if anyone has any ideas on the best approach for this. 

Part of the difficulty is that our transitions are done via css classes, so there is a timing issue when applying `collapsing`, `show`, and then setting the height after those classes are added.

Any input is appreciated!

cc @matuzalemsteles @diegonvs


packages/clay-panel/src/index.tsx
```diff
@@ -120,7 +136,7 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 								collapsed: showIconCollapsed,
 							}
 						)}
-						onClick={animate}
+						onClick={() => setExpanded(!expanded)}
```
matuzalemsteles, line 139: Why not call `animate` here in conjunction with `setExpanded`?

matuzalemsteles (commented):
Hey @bryceosterhaus I didn't quite understand the idea, are you trying to eliminate the need for a trigger? I think we can test two things.

- Keep the implementation as before but remove `animate` and make it an `useEffect` that will be triggered by `expanded`. (I didn't get to test this but I imagine it might work)
- Keep the implementation as before but together with the call to `setExpanded`, call the `animate` method, this maybe does the same thing as before because it received an `onChange` method, so maybe you have to bring this out of hook.

A question, the implementation of before did not work well? just passing `expanded` or `setExpanded` to the `useTransitionHeight` hook, whether the value is controlled or internal.

The important thing is to always know when expanded changed to add the element's height and then remove and do the same for when it changes to `false` and `true`. The closing behavior on the lap is sometimes a little more complicated, you have to deal with a different sequence. I'll take a closer look at it tomorrow and see if I can bring anything else to help.

packages/clay-panel/src/index.tsx
bryceosterhaus: We want the ability to animate the component whenever the `expand` prop is changed. We can add animate here, but that doesn't allow us to actually animate it by changing the prop from the parent component

bryceosterhaus:
I'm trying to rewrite this component to allow for a "controlled" api of `expand`. I tried to add a `useEffect` with an `[expand]` dependency, but the problem is that `useTransitionHeight` will fire `setExpand`, so this leads to an infinite loop of `animate -> setExpand -> (expand changes) -> useEffect -> animate`.

I'm going to see if react-css-transitions can help solve this...

bryceosterhaus:
@matuzalemsteles I finally got around to working on this again, go ahead and check out my changes. This preserves the old API but now allows for controlling some components if they are expanded and animating when that value changes. Let me know what you think or if you see any bugs

matuzalemsteles (commented):
@bryceosterhaus I did some tests and observed some strange behaviors, added some comments... let me know if you can also reproduce.

bryceosterhaus:
@matuzalemsteles updated to just use `react-transition-group` and it seems to be much more consistent with this approach. I also went ahead and created a hook for the "internal state" approach we take for backwards compatibility.

I haven't updated nav or nav-bar yet, wanted to see what you thought on this first.

matuzalemsteles:
Hey @bryceosterhaus it really looks like it got a lot better on the Panel and maybe simplifying it further, I think it's good to go with that for the other components, I think we don't even need the `useTransitionHeight` hook anymore.

packages/clay-shared/src/setElementFullHeight.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function setElementFullHeight(element: HTMLElement) {
+	if (element) {
+		// Cloned into a new array since `.reduce` is not a method on an HTMLCollection
+		const height = Array.prototype.slice
+			.call(element.children)
+			.reduce(
+				(acc: number, child: HTMLElement) => acc + child.clientHeight,
+				0
+			);
```
wincent, line 17: I don't think I'd create a new object only to have it immediately GC'ed like this, especially if it involves me having to use a relatively esoteric `slice` + `call`. A simple `for` loop would do and arguably be more readable (no need for type annotations, no need for `reduce`, which some people find hard to read etc):

```js
let height = 0;

for (let i = 0; i < element.children.length; i++) {
    height += element.children[i].clientHeight;
}
```

packages/clay-shared/src/useInternalState.ts
```diff
@@ -0,0 +1,34 @@
...
+	value,
+}: IArgs<TValue>) {
+	const [internalValue, setInteralValue] = React.useState(initialValue);
+
+	if (typeof value === 'undefined') {
+		value = internalValue;
+	}
+
+	if (typeof onChange === 'undefined') {
+		onChange = setInteralValue;
```
wincent, line 30: Typo here: `setInteralValue` → `setInternalValue`

GnaneshVadlamudi:
Hi @bryceosterhaus 
I hope above mentioned issue got fixed. But not able to find the use case scenario in Clay site. Kindly let me know whether the changes got released or not  ? or the use case scenario.

wincent:
@GnaneshVadlamudi: The last release was on [2021-2-11](https://github.com/liferay/clay/blob/master/CHANGELOG.md#2021-2-11), and includes all PRs merged prior to that date (including this one).

GnaneshVadlamudi:
@wincent  Thank you so much for immediate response. But unable to find the use case scenario or API Usage for issue : 
#3754 : can expanding one panel collapse all other panels in a group

wincent:
@GnaneshVadlamudi: Does [this storybook example](https://storybook.clayui.com/?path=/story/components-claypanel--collapsable) which [was updated in this PR](https://github.com/liferay/clay/pull/3760/files#diff-64ad94b041078381b22c219622de7f06e7d514474222dd321020cf2017e16ce7) cover you are looking for? The idea is that as a controlled component, you can pass in the state for which panel should be open/closed at any given time from above via props; that's what the "Toggle Expand" button does in the example — so you'd just want to maintain a list of panels in the parent component and ensure that when you toggle one of them, only that one gets `expanded: true` and all the others get `false`.

matuzalemsteles:
Hey @GnaneshVadlamudin we don't have this specific example here, the goal here was just to provide an API that allows you to achieve that goal, as Greg commented above. Well, I did an example in the [codesandbox](https://codesandbox.io/s/quick-start-clay-finale-forked-sgngt?file=/src/index.js).

PR #3768, opened 2020-10-27
chore(@clayui/card): uses the ClaySticker.Image component internally

### LINKED ISSUE #3764: Markup Inconsistency between CardWithUser React component and html markup
React component CardWithUser with an image is rendering the following image markup:

```
<img alt="thumbnail" class="aspect-ratio-item-fluid" src="/images/long_user_image.png">
```

While in the Markup view what's being rendered is

```
<img alt="thumbnail" class="sticker-img" src="/images/long_user_image.png">
```

So in the react component we're rendering `aspect-ration-item-fluid` class but in the markup example `sticker-img` appears.

### LINKED ISSUE #3765: Why are we using `<img>` instead of `ClaySticker.Image` in `CardWithUser`?
https://github.com/liferay/clay/blob/master/packages/clay-card/src/CardWithUser.tsx#L120

Here we're rendering directly an image instead of using `ClaySticker.Image`. Why is this?

IMHO I think we should use it

```
<ClaySticker.Image
	alt={userImageAlt}
	className="aspect-ratio-item-fluid"
	src={userImageSrc}
/>
```

matuzalemsteles:
Fixes #3765 and Fixes #3764

I removed the use of `aspect-ratio-item-fluid` that was being added because it ends up doing the same as `sticker-img`.

### DIFF
--- a/packages/clay-card/src/CardWithUser.tsx
+++ b/packages/clay-card/src/CardWithUser.tsx
@@ -118,11 +118,7 @@ export const ClayCardWithUser: React.FunctionComponent<IProps> = ({
 				shape="circle"
 			>
 				{userImageSrc && (
-					<img
-						alt={userImageAlt}
-						className="aspect-ratio-item-fluid"
-						src={userImageSrc}
-					/>
+					<ClaySticker.Image alt={userImageAlt} src={userImageSrc} />
 				)}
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
LGTM!

Issue #3773, opened 2020-10-30
 CardWithInfo can't render a ClaySticker with an image


julien:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## Is there an example you can provide via codesandbox.com?

Yes, please see https://codesandbox.io/s/sharp-rgb-k30gg to understand the "problem"

## What are the steps to reproduce?

Render a [`<ClayCardWithInfo />`](https://github.com/liferay/clay/blob/8b18dc5e48f730aa60ef1fdc51a6f75edabc3494/packages/clay-card/src/CardWithInfo.tsx) and use `stickerProps`

## What is the expected result?

I'd like to be able to pass an `src` (the name of this prop doesn't really matter as long as it makes sense) attribute to [`stickerProps`](https://github.com/liferay/clay/blob/8b18dc5e48f730aa60ef1fdc51a6f75edabc3494/packages/clay-card/src/CardWithInfo.tsx#L155-L156) to be able to render a card with a sticker that displays an image instead of text.

While this is probably doable in pure JSX (although I haven't found a way to do it), in my case, I'm rendering cards via a jsp tag, that accepts an [`stickerImageSrc`](https://github.com/liferay/liferay-portal/blob/02396f764ad152bde79f062c7cb1ca941a670234/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/resources/META-INF/liferay-clay.tld#L1832) attribute and I need some help to figure this out.

Thanks!

matuzalemsteles:
@julien maybe you can do something like that.

```jsx
<ClayCardWithInfo
	stickerProps={{
		content: <ClaySticker.Image alt="" src="" />
	}}
/>
```

julien:
@matuzalemsteles yes that looks totally usable :+1: 
Before closing, I'll just ask @carloslancha if he wants to add anything else? (i.e. @carloslancha do you think this should be done automatically by adding a new property on `stickerProps`)

matuzalemsteles:
@julien I think this was a little inconsistent with the other components that do the same, in this case, the `<ClayCardWithUser /> `, so maybe create a `stickerDisplayType` and `stickerContent` and deprecated `stickerProps`, since in that case, we can have a text and an image.

cc @bryceosterhaus

bryceosterhaus:
[...]

PR #3775, opened 2020-11-03
chore: add breaking change note for 4.x

bryceosterhaus:
Stems from discussion at https://github.com/liferay/clay/issues/3773

This gives us a way to remember requests like these when we decide to break at 4.x

### DIFF
--- a/packages/clay-card/BREAKING.md
+++ b/packages/clay-card/BREAKING.md
@@ -1,1 +1,3 @@
 # List of Breaking Changes for 4.x
+
+[Update `stickerProps`](https://github.com/liferay/clay/issues/3773#issuecomment-720658343)


wincent:
Is it a foregone conclusion, then, that we _are_ going to having a breaking version update? Last time we spoke about this was [here](https://liferay.slack.com/archives/G3JBR21HA/p1600754672048400?thread_ts=1600712150.041900&cid=G3JBR21HA). Some excerpts:

@wincent:

> i'd be looking for a very strong justification for why we need to make breaking changes (and require a major semver update), but  I'm not seeing it in that document yet. my feeling is that clay 3.x is still "young" and could be evolved significantly before we get to the point where we can't continue improving it without causing pain for customers

@jbalsas:

> I just get the impression that devs are happy that they’re mostly getting things done, which means that there’s no incentive in putting them through any kind of upgrade process unless the benefit outweighs the cost

@matuzalemsteles 

> I see no problem in waiting longer for bigger changes if we want to wait and give more time.

Basically, there was a long discussion in there about surveying developers before considering moving to v4.0.

bryceosterhaus:
> Is it a foregone conclusion, then, that we are going to having a breaking version update?

**Nope!** There hasn't been any conclusion about a 4.x version, and this doc isn't here to dictate if/when we will.

The purpose of keeping docs like this is so that we can jot any ideas we have that would require a breaking change and this way we can view one single place to look whenever we decide to make that change. So for this case, rather than just trying to remember we wanted to make some API changes for Card, this way we can make note of any specific changes for this component we had thought about.

PR #3780, opened 2020-11-05
fix(@clayui/card): Avoid setting `href` on `span` element

### LINKED ISSUE #3779: Clay Card Description (3.x) passes `href` attribute to `span` element
During a [code review](https://github.com/liferay-frontend/liferay-portal/pull/448#discussion_r512231080), @carloslancha pointed out that in some cases, [Card.Description sets an `href` attribute on a `span` element](https://github.com/liferay/clay/blob/6b00a1e60deaa3c415dbed5227c3ca6017ac14ea/packages/clay-card/src/Description.tsx#L62) - and we're wondering if that's the correct behaviour.

julien:
If the card is disabled, a `span` element will be rendered and no
`href` attribute is required

Fixes #3779

### DIFF
--- a/packages/clay-card/src/Description.tsx
+++ b/packages/clay-card/src/Description.tsx
@@ -59,7 +59,10 @@ const ClayCardDescription: React.FunctionComponent<ICardDescriptionProps> = ({
 		>
 			{truncate ? (
 				<span className="text-truncate-inline">
-					<InnerTag className="text-truncate" href={href}>
+					<InnerTag
+						className="text-truncate"
+						href={disabled ? undefined : href}
+					>
 						{children}
 					</InnerTag>
 				</span>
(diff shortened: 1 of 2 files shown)


julien:
@bryceosterhaus, @matuzalemsteles I ran `yarn ci` locally before sending the PR and I see this is failing.

Am I supposed to run or send anything else?

matuzalemsteles:
@julien you probably need to run `yarn build` before running `yarn ci` or `yarn develop` on clayui.com.

matuzalemsteles (approved):
LGTM! I put it to run the tests and the stats again, the stats were breaking, giving a very high diff, it seems something unrelated to this PR.

julien:
@matuzalemsteles thanks for the quick reply.
In case I need to pushed updated build files (which I didn't do), let me know

matuzalemsteles:
@julien this is not necessary, don't worry. @bryceosterhaus I see that the stats are having a very sharp reduction, this could be a bug? maybe some faulty merge caused a big spike and is now recalculating... do you have any idea why this is happening?

bryceosterhaus:
@matuzalemsteles that looks really odd. I'm gonna re-run the stats and see if it was just a one-off mistake

bryceosterhaus:
@matuzalemsteles yeah my only guess is that something at HEAD might be off for the CI to calculate. I think this is fine to merge, but lets keep an eye on if things change or if it stays broken.

PR #3801, opened 2020-11-18
fix(@clayui/time-picker): make sure input always has a leading 0 for single digits

### LINKED ISSUE #3782: Time Picker is missing the leading 0 for single digit minutes
Screen Shot 2020-11-05 at 10 33 44 AM

| Tech  | Version |
| ----- | ------- |
| @clayui/time-picker  | v3.2.0 |

### LINKED ISSUE #3784: Time Picker uses 0:00 rather than 12:00 in AM/PM mode
Screen Shot 2020-11-05 at 10 33 59 AM

| Tech  | Version |
| ----- | ------- |
| @clayui/time-picker  | v3.2.0 |

bryceosterhaus:
fixes #3784
fixes #3782

The interactions should now be more consistent with the html native time picker

### DIFF
--- a/packages/clay-time-picker/src/index.tsx
+++ b/packages/clay-time-picker/src/index.tsx
@@ -193,18 +193,18 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
 	) => {
 		const config = useConfig[configName];
 		const intrinsicValue = Number(value);
-		const setValue = (newValue: string | number) =>
-			onInputChange({
+		const setValue = (newValue: string | number) => {
+			const newVal =
+				configName === TimeType.ampm
+					? newValue
+					: handleMaxAndMin(String(newValue), config as ConfigMaxMin);
+
+			return onInputChange({
 				...values,
 				// eslint-disable-next-line sort-keys
-				[configName]:
-					configName === TimeType.ampm
-						? newValue
-						: handleMaxAndMin(
-								String(newValue),
-								config as ConfigMaxMin
-						  ),
+				[configName]: String(newVal).padStart(2, '0'),
 			});
+		};
 
 		switch (event.key) {
 			case 'Backspace':
@@ -238,11 +238,29 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
 				break;
 			default:
 				if (regex.test(event.key) && configName !== TimeType.ampm) {
+					const maxFirstDigit =
+						configName === TimeType.hours
+							? use12Hours
+								? 1
+								: 2
+							: 5;
+
+					const newVal =
+						Number(value) > maxFirstDigit
+							? `0${event.key}`
+							: (value && value !== DEFAULT_VALUE ? value : '') +
+							  event.key;
+
+					setValue(newVal);
+				} else if (
+					configName === TimeType.ampm &&
+					(event.key === 'a' || event.key === 'p')
+				) {
 					setValue(
-						(value && value !== DEFAULT_VALUE ? value : '') +
-							event.key
+						event.key === 'a'
+							? (config as ConfigAmpm).am
+							: (config as ConfigAmpm).pm
 					);
-					(configName === TimeType.ampm && event.key === 'a') ||
 				}
 				break;
 		}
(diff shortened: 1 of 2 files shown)


matuzalemsteles (commented):
@bryceosterhaus I can still reproduce the case of #3782, I think we should not allow that with the `use12Hours` mode activated the hour and minutes cannot be 00:00 and should start with 01:00.

bryceosterhaus:
@matuzalemsteles how are you reproducing? 

Here is what I am seeing on the deploy preview 
time-picker


If its in 24 hour mode, you cannot input a number greater than 23 for the hours or 59 for the minutes.

If its in 12 hour mode, you cannot input a number greater than 12 for the hours or 59 for the minutes.

matuzalemsteles:
@bryceosterhaus yeah this I can reproduce well, what I'm talking about is that with the 12-hour mode, the time cannot start with `00:00`, we should start with` 01:00`. I found some information about it here https://www.timeanddate.com/time/am-and-pm.html

bryceosterhaus:
@matuzalemsteles ah gotcha, that shouldn't be too hard to update. Ill send a change today

bryceosterhaus:
@matuzalemsteles updated

packages/clay-time-picker/src/index.tsx
```diff
@@ -238,9 +238,33 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
...
+								: 2
+							: 5;
+
+					const keyVal =
+						Number(event.key) < (config as ConfigMaxMin).min
+							? (config as ConfigMaxMin).min
+							: event.key;
+
+					const newVal =
+						Number(value) > maxFirstDigit
```
matuzalemsteles, line 254: You can use a variable that is already doing this, just so you don't have to convert again.

```suggestion
						intrinsicValue > maxFirstDigit
```

packages/clay-time-picker/src/index.tsx
```diff
@@ -238,9 +238,33 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
 				break;
 			default:
 				if (regex.test(event.key) && configName !== TimeType.ampm) {
+					const maxFirstDigit =
+						configName === TimeType.hours
+							? use12Hours
+								? 1
+								: 2
+							: 5;
```
matuzalemsteles, line 246: @bryceosterhaus just a doubt, I can't quite understand why this value is 5 here 😅.

packages/clay-time-picker/src/index.tsx
```diff
@@ -238,9 +238,33 @@ const ClayTimePicker: React.FunctionComponent<IProps> = ({
 				break;
 			default:
 				if (regex.test(event.key) && configName !== TimeType.ampm) {
+					const maxFirstDigit =
+						configName === TimeType.hours
+							? use12Hours
+								? 1
+								: 2
```
matuzalemsteles, line 245: I would say the same for that value 😅

matuzalemsteles (commented):
@bryceosterhaus I just added a few comments

packages/clay-time-picker/src/index.tsx
bryceosterhaus: This is because we never want to allow the first digit in the minutes input to be greater than 5. It's the max value for the "10"s spot in minutes. That way we allow 0-59, but not 60+.

I guess we could do this programatically by taking the "10"s digit from the config.max

packages/clay-time-picker/src/index.tsx
bryceosterhaus: Same as above, except for hours. We should probably do programmatically from the config.

bryceosterhaus:
@matuzalemsteles updated to derive numbers from the min/max instead

matuzalemsteles (commented):
Hey @bryceosterhaus I know this is not related to these changes but it seems that there are some conflicts of using the keyboard to change the numbers in the TimePicker that is inside the DatePicker, apparently, conflict with the FocusScope, you may have to add an `event.stopPropagation()` here... Another thing is that we may have to update the initial value or it must be formatting the Team to `HH:mm` in DatePicker.

bryceosterhaus:
@matuzalemsteles yeah we should fix that. Could you open an additional issue to track that change?

matuzalemsteles:
@bryceosterhaus I just created issue #3810 by linking to the comment so we can track it. I think we can continue with this PR.

Issue #3786, opened 2020-11-06
Clay Card (3.x): Default Icon for ClaySticker is "documents-and-media"


julien:
Not a big issue, but I was wondering why the [value for the sticker's icon in `ClayCardWithInfo`](https://github.com/liferay/clay/blob/9270625e8bafa20c1e77ab2cbe67ebf5560d5584/packages/clay-card/src/CardWithInfo.tsx#L121) is `'documents-and-media'`.

I'm asking because looking at Lexicon's card definitions, it seems that [some cards have no sticker at all](https://liferay.design/lexicon/core-components/cards/cards-variation/#blog-card)

carloslancha:
Actually there's a different default icon for Image cards (`document-image`) and File cards (`document-default`), so this could be considered a bug.

Also, those default icons changed last week after a conversation with @victorvalle . Victor, did you open any ticket or issue to update this? (I guess any change in docs should be followed by a jira ticket or Github issue)

https://docs.google.com/document/d/1st2moaRVphrsFrDJUz-ic26aH-V46YkQam75X6BHPY0/edit#heading=h.l82je57kortn

Thx guys

victorvalle:
I didn't create any tickets. I thought you were adding this requirement on the fly. If I have to just let me know it.

carloslancha:
We did it for the Tags but they need to know about it here. Don't worry about the issue, we can use this to do it. Thx!

bryceosterhaus:
Likely was un-intentional that we had the wrong default icon. Whatever lexicon defaults to, we should as well. I see no problem updating the default icon and labeling it as a bug fix

carloslancha:
I'll send a pr tomorrow updating the component to use correct defaults depending on displayType.

Thx @bryceosterhaus !

PR #3803, opened 2020-11-19
fix(@clayui/card): Change default symbols and add displayType attribute to CardWithInfo so the the correct default symbol and sticker symbol could be used depending on the displayType

carloslancha:
Fixes https://github.com/liferay/clay/issues/3786

### DIFF
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -15,6 +15,8 @@ import React from 'react';
 
 import ClayCard from './Card';
 
+type CardWithInfoDisplayType = 'file' | 'image';
+
 interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	/**
 	 * List of actions in the dropdown menu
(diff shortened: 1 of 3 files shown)


packages/clay-card/src/CardWithInfo.tsx
```diff
@@ -36,6 +38,11 @@ interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	 */
 	disabled?: boolean;
 
+	/**
+	 * Determines the style of the card
+	 */
+	displayType?: CardWithInfoDisplayType;
```
matuzalemsteles, line 44: Just out of curiosity, why did you create a new type instead of adding it directly here? I see no problem with staying that way just to know.

matuzalemsteles (approved):
LGTM! Overall it looks good, I liked the idea of ​​`symbol` overlaying `displayType`.

packages/clay-card/src/CardWithInfo.tsx
carloslancha: I'm just following your pattern in https://github.com/liferay/clay/blob/1a9e591e623274890362b881474e545ffbf07415/packages/clay-card/src/Card.tsx#L18

matuzalemsteles: Lol, I probably missed this during the review 🤪, but thanks anyway. Feel free to change something if you think it doesn't make sense.

carloslancha:
Merging this one since it was approved. Thx.

Issue #3807, opened 2020-11-23
Dropdown triggers on Card components have the wrong display type 


markocikos:
Steps to reproduce:
1. Place `frontend-taglib-clay-sample-web` on a page and open it.

What is the expected result?
- On cards, trigger icons to open dropdown should be `unstyled` buttons.

What actually happens
- On cards, trigger icons to open dropdown are `primary` buttons. (see screenshot)
![screen](https://user-images.githubusercontent.com/5435511/99957838-9d223900-2d88-11eb-818b-a331963d9d3a.png)

| Tech  | Version |
| ----- | ------- |
| @clayui/card  | v3.5.1  |

PR #3808, opened 2020-11-23
fix(@clayui/card): Set display type of dropdown triggers to `unstyled`

markocikos:
This PR fixes the bug in https://github.com/liferay/clay/issues/3807

There is no need to update tags on the portal side, we are [already setting](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/java/com/liferay/frontend/taglib/clay/servlet/taglib/DropdownActionsTag.java#L31) the `unstlyed` display style in server-side rendering.

This issue was unnoticed because button styles were overriden by `.component-action` styles, that is why trigger looks correct in the storybook. But, in recent portal changes, higher priority was given to `btn-primary` class, that exposed the issue.

To test without portal, just on storybook:
1. Remove `color` and `background-color` from `.component-action`
2. Compare before and after PR

@julien @carloslancha 

Before PR:
before

After PR:
after

### DIFF
--- a/packages/clay-card/src/CardWithHorizontal.tsx
+++ b/packages/clay-card/src/CardWithHorizontal.tsx
@@ -104,6 +104,7 @@ export const ClayCardWithHorizontal: React.FunctionComponent<IProps> = ({
 									{...dropDownTriggerProps}
 									className="component-action"
 									disabled={disabled}
+									displayType="unstyled"
 									spritemap={spritemap}
 									symbol="ellipsis-v"
 								/>
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -223,6 +223,7 @@ export const ClayCardWithInfo: React.FunctionComponent<IProps> = ({
 										{...dropDownTriggerProps}
 										className="component-action"
 										disabled={disabled}
+										displayType="unstyled"
 										spritemap={spritemap}
 										symbol="ellipsis-v"
 									/>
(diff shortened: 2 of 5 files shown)


carloslancha:
LGTM. I'm surprised / worried this doesn't break any test.

matuzalemsteles (approved):
LGTM! Yeah, strange that it only breaks on the Portal, was this due to some recent change in the styles of the Portal? Just thinking about the root cause, something related to the css loading order...

markocikos:
> LGTM! Yeah, strange that it only breaks on the Portal, was this due to some recent change in the styles of the Portal? Just thinking about the root cause, something related to the css loading order...

On portal it breaks because of [these stlyes](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-classic/src/css/custom_properties/_custom_properties_set.scss#L58-L64). Additional `:not(.dropdown-item)` selector bumps it in priority.

screen

julien (approved):

carloslancha:
Can we add a test case to check the resulting snapshot when actions are passed to the cards? That's why no tests are failing here when we're changing the markup.

matuzalemsteles:
Thanks @markocikos for the details!

> Can we add a test case to check the resulting snapshot when actions are passed to the cards? That's why no tests are failing here when we're changing the markup.

Are you saying this on the Clay side or Portal? Yeah, it would be nice to add a case like this, I think we are only testing something specific to the actions and not just generating the snapshot.

carloslancha:
>Are you saying this on the Clay side or Portal? Yeah, it would be nice to add a case like this, I think we are only testing something specific to the actions and not just generating the snapshot.

Here in Clay. Yes, we're only testing action interactions so we're missing markup changes there.

markocikos:
@carloslancha @matuzalemsteles I added commit with extended snapshot tests that cover dropdown trigger markups.

carloslancha:
Thx @markocikos !

carloslancha:
Hey guys, I'm merging this so we have it in the next release

Issue #3814, opened 2020-11-25
Flush and fluid should not be used together in CardWithInfo


carloslancha:
See https://liferay.slack.com/archives/G3JBR21HA/p1606316882037300 for more info

PR #3815, opened 2020-11-25
fix: flush and fluid should not be used together in CardWithInfo

carloslancha:
Fixes https://github.com/liferay/clay/issues/3814

Do not require any tag update in portal, but https://issues.liferay.com/browse/LPS-123984 depends on this


matuzalemsteles (approved):
LGTM! @carloslancha I think we can create a "action: taglib" label to add to the PR to say that changes also need to be made to the taglib.

carloslancha:
>I think we can create a "action: taglib" label to add to the PR to say that changes also need to be made to the taglib.

Sounds good Matu, thx!

PR #3817, opened 2020-11-27
fix(@clayui/card): Only render sticker if stickerProps are passed

### LINKED ISSUE #3816: Clay Card (3.x) sticker should not always be rendered
Related to what I commented in #3786, stickers [should not always be rendered in cards](https://github.com/liferay/clay/blob/bfb2529dc57c9415a5873d7b9e8b22b3d7be428d/packages/clay-card/src/CardWithInfo.tsx#L173-L198).

As an example, this should not happen:

julien:
Fixes #3816

### DIFF
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -170,32 +170,34 @@ export const ClayCardWithInfo: React.FunctionComponent<IProps> = ({
 				/>
 			)}
 
-			<ClaySticker
-				displayType={
-					stickerProps && stickerProps.displayType
-						? stickerProps.displayType
-						: 'primary'
-				}
-				position="bottom-left"
-				{...stickerProps}
-			>
-				{stickerProps ? (
-					stickerProps.children ? (
-						stickerProps.children
+			{stickerProps && (
+				<ClaySticker
+					displayType={
+						stickerProps && stickerProps.displayType
+							? stickerProps.displayType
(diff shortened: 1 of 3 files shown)


packages/clay-card/src/CardWithInfo.tsx
```diff
@@ -170,32 +170,34 @@ export const ClayCardWithInfo: React.FunctionComponent<IProps> = ({
...
-				)}
-			</ClaySticker>
+						<ClayIcon
+							spritemap={spritemap}
+							symbol={
+								isCardType.image
+									? 'document-image'
+									: 'document-default'
+							}
+						/>
```
matuzalemsteles, line 197: @julien I think this guy will no longer be rendered after this change, we may have to improve this conditional if we still want to support this case.

julien:
So after speaking with @carloslancha this change is not what we want since doing this we're changing the default behavior that we want to preserve (specified by Lexicon). What we need here is a way to be able to not render the sticker while keeping the default behavior.
Closing this one and we'll send a followup.

packages/clay-card/src/CardWithInfo.tsx
julien: @matuzalemsteles exactly what we were talking about with @carloslancha :+1:

matuzalemsteles: Thanks @julien for the info.

PR #3819, opened 2020-11-27
fix: Make stickerProps nullable and minor refactor

carloslancha:
Fixes https://github.com/liferay/clay/issues/3816

With this pr we're adding support to pass `stickerProps={null}`, which will cause that the sticker will not be rendered.

In DXP we'll need to add a `showSticker` attribute to `vertical-card`, `file-card` and `image-card` tags, defaults to `true` and in `VerticalCard.js` pass `stickerProps={null}` if `showSticker` is `false`.

We'll need to add `isSticker` method to `VerticalCard.java` model, defaults to `true`.

### DIFF
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -93,11 +93,9 @@ interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	/**
 	 * Values used in displaying bottom-left icon
 	 */
-	stickerProps?:
-		| (IClayStickerProps & {
-				content?: React.ReactNode;
-		  })
-		| null;
+	stickerProps?: IClayStickerProps & {
+		content?: React.ReactNode;
+	};
 
 	/**
 	 * Name of icon
@@ -125,7 +123,7 @@ export const ClayCardWithInfo: React.FunctionComponent<IProps> = ({
 	onSelectChange,
 	selected = false,
 	spritemap,
-	stickerProps,
+	stickerProps = {},
 	symbol,
 	title,
 	...otherProps
@@ -171,22 +169,20 @@ export const ClayCardWithInfo: React.FunctionComponent<IProps> = ({
 				/>
 			)}
 
-			{stickerProps !== null && (
+			{stickerProps && (
 				<ClaySticker
 					displayType={
-						stickerProps && stickerProps.displayType
+						stickerProps.displayType
 							? stickerProps.displayType
 							: 'primary'
 					}
 					position="bottom-left"
 					{...stickerProps}
 				>
-					{stickerProps ? (
-						stickerProps.children ? (
-							stickerProps.children
-						) : (
-							stickerProps.content
-						)
+					{stickerProps.children ? (
+						stickerProps.children
+					) : stickerProps.content ? (
+						stickerProps.content
 					) : (
 						<ClayIcon
 							spritemap={spritemap}
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -11,7 +11,7 @@ import ClayLabel from '@clayui/label';
 import ClayLayout from '@clayui/layout';
 import ClaySticker, {IClayStickerProps} from '@clayui/sticker';
 import classNames from 'classnames';
-import React, {useMemo} from 'react';
+import React from 'react';
 
 import ClayCard from './Card';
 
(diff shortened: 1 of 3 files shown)


packages/clay-card/src/CardWithInfo.tsx
```diff
@@ -135,20 +135,31 @@ export const ClayCardWithInfo: React.FunctionComponent<IProps> = ({
...
+		if (!symbol) {
+			if (isCardType.image) {
+				return 'camera';
+			}
+
+			return 'documents-and-media';
+		}
+
+		return symbol;
+	}, [isCardType]);
```
wincent, line 148: I don't think you want `useMemo` for this. The cost of memoization is probably close to the cost of the operation you're trying to memoize, and possibly greater than it:

- Without memoization = one or two simple `if` statements, zero or one property accesses.
- With memoization = an array instantiation, a function call, iteration over the array with an `Object.is()` call for each item in it (1).

Similar analysis for `stickSymbol` below.

wincent (commented):
Just commenting on the `useMemo`, but will leave actual review up to _le Clay Folk_.

packages/clay-card/src/CardWithInfo.tsx
carloslancha: Better now?

wincent: > Better now?

never-enough

Instead of making functions and calling them, just do:

```js
const contentSymbol = symbol
    ? symbol
    : isCardType.image
    ? 'camera'
    : 'documents-and-media';

const stickeSymbol = isCardType.image ? 'document-image' : 'document-default';
```

You avoid the function call overhead (creating a stack frame, calling, returning etc). These simple ternary expressions are so fast that you can consider them to be "free".

carloslancha: Done

julien (approved):

markocikos (approved):
LGTM!

jbalsas:
Don't just look at it, get it merged! ;)

packages/clay-card/src/CardWithInfo.tsx
```diff
@@ -95,9 +93,11 @@ interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
...
 	 * Values used in displaying bottom-left icon
 	 */
-	stickerProps?: IClayStickerProps & {
-		content?: React.ReactNode;
-	};
+	stickerProps?:
+		| (IClayStickerProps & {
+				content?: React.ReactNode;
+		  })
+		| null;
```
matuzalemsteles, line 100: I think this is not an impediment but it is a little strange to have an API behavior that is different when it is set to `null`, I am probably not used to it... but maybe we can use a boolean here to solve this? it is more common in my view.

carloslancha: I don't have any preference here, I did it in this way 'cause it felt weird to pass a boolean to a prop called `stickerProps`. It's like
- what time is it?
- yes

But you're more familiar with the rest of the code in Clay, so you decide @matuzalemsteles

matuzalemsteles: @carloslancha makes sense, that's a good argument 🙂! I don't have very strong opinions about it, the rest of our codebase doesn't have similar behaviors, so this would be the first behavior. I think we can make some changes about it, for example instead of just checking `null` we can set `stickerProps` to the default value with `{}` and just check in the conditional if there is any value for `stickerProps`.

carloslancha: @matuzalemsteles I like that idea. I pushed some changes with that (that also fixes a wrong behavior with custom stickers not rendering default icon if no `children` or `content` is passed we didn't catch before. Please take a look and merge it if it's fine, I'm planing to cut a release later today. Thx!

matuzalemsteles (approved):
LGTM!

packages/clay-card/src/CardWithInfo.tsx
wincent: Obviously too late now, but I liked the first approach (using `null`) better.

- Boolean is no good for this because, like @carloslancha said, you don't want the answer to the question "what time is it?" to be "yes" or "no". A good heuristic for using a boolean is: are there clear semantics for what _both_ `true` and `false` should mean? In this example, `false` could communicate "no sticker", but `true` doesn't communicate a useful inverse.
- `null` is the value that JS uses to _explicitly_ signal "nothing here", as opposed to `undefined`, which just _implicitly_ means "I didn't put anything here" — so, for example, `undefined` can be useful to allow a default to take effect (pass nothing and you get a reasonable default), vs `null` to indicate "I want nothing here". In general, you also can't tell the difference between someone passing `undefined` and not passing anything — so `null` really is the right tool for _signalling_ "nothing".

carloslancha: We didn't go with the `boolean` approach @wincent but what you say makes sense, we're passing `undefined` when we want nothing no be rendered but `null` is more explicit, thx!

@matuzalemsteles would you mind applying this feedback (use of `null` instead `undefined`) to https://github.com/liferay/clay/pull/3827 ? Thx!

matuzalemsteles: Makes sense! @carloslancha no problem, you want to get back with this behavior for the `ClayCardWithInfo`?

carloslancha: @matuzalemsteles just add `null` as a valid value of `stickerProps`, that should be enough :)

carloslancha: @matuzalemsteles did we added this change in the refactor pr?

matuzalemsteles: @carloslancha Sorry, I ended up forgetting to add this to the PR, I will send another one now, making this change. Thanks for remember.

carloslancha: Thx @matuzalemsteles !

matuzalemsteles: Done!

Issue #3789, opened 2020-11-06
ClayCardWithHorizontal can generate duplicate CSS class that breaks styles


markocikos:
Steps to reproduce:
1. Go to [horizontal card storybook](https://storybook.clayui.com/?path=/story/components-claycard--cardwithhorizontal).
1. On the left card, find container `<div class='card'>` and add style `border: 1px solid black`

or, to see usage in portal that I ran into:

1. On this line in `depot-web`, change `<clay:horizontal-card-v2>` to `<clay:horizontal-card>` and redeploy. This changes implementation from Clay v2 to Clay v3.
1. In Global Menu, create an Asset Library.
1. On the Asset Library item, click on title to open its dashboard.

What is the expected result?
- Horizontal card looks the same as it did in Clay v2 (except for bottom border in storybook)

What actually happens:
- Horizontal card looks differently:
![screen](https://user-images.githubusercontent.com/5435511/98388270-d8b3c800-2052-11eb-95f0-895598612afc.png)

This happens because `card` CSS class is duplicated, along with styles on it.

HTML of v2:
```
<div class='card card-horizontal ...'>
    <div class='card-body'>...</div>
</div>
```

HTML of v3:
```
<div class='card  ...'>
    <div class='card card-horizontal ...'>
        <div class='card-body'>...</div>
    </div>
</div>
```

Notes:
- Without the `border` hack, Chrome collapses two bottom margins into one, so the issue is poorly visible. Behavior might be different of other browsers.
- This does not happen in all usages of horizontal card, only if `selectable` is `false`. That is the left card in storybook.
- The styles that add margin twice are in [Clay CSS](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_cards.scss#L13). The style that adds border is in portal.
- Root cause of the issue is that the HTML from v2 to v3 is different, which is a hurdle in itself, but the additional fact that `card` class is duplicated makes me think this could simply be a bug.
- [First place](https://github.com/liferay/clay/blob/master/packages/clay-card/src/Card.tsx#L71) where `card` is inserted, [second place](https://github.com/liferay/clay/blob/master/packages/clay-card/src/Body.tsx#L25).
- Possible solutions:
  - Remove one of the `card` classes. That might break other things but would be the cleanest solution.
  - Modify Clay CSS to cover this case.
[...]

PR #3820, opened 2020-11-30
fix(@clayui/card): prevent double 'card' class on ClayCardWithHorizontal

markocikos:
Fixes: https://github.com/liferay/clay/issues/3789

In this PR we are removing the case of double `card` class:
- Outer class is [added](https://github.com/liferay/clay/blob/master/packages/clay-card/src/Card.tsx#L71) when `!selectable`
- Inner class is [added](https://github.com/liferay/clay/blob/master/packages/clay-card/src/Body.tsx#L24-L26) when the card is `horizontal` and `!interactive`. 

In analysis I found that removing inner card would we problematic, but removing outer would be doable after we remove the only case that relies on outer CSS class: horizontal interactive card. We achieved this with navigation card, so we're good to go.

As this is released, we need to update server side rendering in horizontal tag to match markup, specifically we need to delete [these three lines](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/java/com/liferay/frontend/taglib/clay/servlet/taglib/HorizontalCardTag.java#L114-L116):
```
else {
	cssClasses.add("card");
}
```

Notes:
- This is no longer a fix for a blocker, as the case in the original ticket was fixed by a different solution, a navigation card. This is now more of an improvement, to prevent future headaches.
- There is a bit of SF in this PR: `(horizontal && interactive) || interactive` is the same as `interactive`
- `|| undefined` on the end of `classNames` is to prevent `class=""` in markup. See more [here](https://github.com/JedWatson/classnames/issues/115).

@carloslancha @julien

### DIFF
--- a/packages/clay-card/src/Card.tsx
+++ b/packages/clay-card/src/Card.tsx
@@ -69,8 +69,7 @@ const ClayCard: React.FunctionComponent<IProps> & {
 				{...otherProps}
 				className={classNames(className, {
 					card: !selectable && (!horizontal || interactive),
-					'card-interactive card-interactive-primary card-type-template':
-						interactive,
+					'card-interactive card-interactive-primary card-type-template': interactive,
 					'card-type-asset':
 						isCardType.file || isCardType.image || isCardType.user,
 					'card-type-directory form-check form-check-card form-check-middle-left':
@@ -85,7 +84,7 @@ const ClayCard: React.FunctionComponent<IProps> & {
 					'template-card': interactive && !horizontal,
 					'template-card-horizontal': horizontal && interactive,
 					'user-card': isCardType.user,
-				}) || undefined}
+				})}
 				href={interactive ? href : undefined}
 				onClick={onClick}
 				role={onClick ? 'button' : undefined}
--- a/packages/clay-card/src/Card.tsx
+++ b/packages/clay-card/src/Card.tsx
@@ -68,9 +68,9 @@ const ClayCard: React.FunctionComponent<IProps> & {
 			<TagHeaderName
 				{...otherProps}
 				className={classNames(className, {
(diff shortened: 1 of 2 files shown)


markocikos:
Hmm, `yarn test` on the whole repo is passing locally. :thinking: 

Honestly, if this PR starts any complication, I am fine with just closing it and the related ticket.

carloslancha:
This is a genuine "Looks not SO horrible to me".

We need to take a deeper look to all this logic, it's too difficult to understand, I spotted a couple of redundant condition and I'm pretty sure that we'll be facing more issues like this one.

The sentence @markocikos told me when I asked some stuff about this "that was my first approach but I quickly went into a rabbit hole" is enough to me to think that this should be refactor/rewritten/simplify/whatever.

So let's move on with this one but we need to take some time to simplify this.

matuzalemsteles:
Yeah I agree, these are some of the visions I had for ClayCard too, I sent an alternative to that in PR #3827. I made some initial changes and added some comments.

markocikos:
@carloslancha I reverted the markup cleanup and updated test. I squashed the last commit it with previous `auto SF` commit for easier review.

carloslancha:
Thanks @markocikos !

@matuzalemsteles we're going with this solution for now since your refactor changes a lot of markup (that looks pretty good) and we're a bit worried about breaking something including it on this release without testing it properly in DXP and we need this release ASAP just to close the card tags migration epic and a support issue.

We can go with yours in the next release, so we can have the time to test it. Sounds right?

Thx everyone!

matuzalemsteles:
@carloslancha no problem, we can look at it with better attention anyway.

PR #3825, opened 2020-12-01
fix(@clayui/css): Tables don't require the class `table-autofit` for column utilities

### LINKED ISSUE #3818: [ClayCSS] Move .table-cell-expand class outside .table-autofit
We are upgrading the portal classes from the compat-layer
https://github.com/liferay-frontend/liferay-portal/pull/554/commits/4bf50d07972ce4fb9bbe0d50d79432b1ea8600d5
- In ClayCSS `.table-cell-expand` class works when is children of `.table-autofit`
- In Portal the equivalent class worked independently
- We suspect in Portal some `.table-cell-expand` are not children of `.table-autofit`

https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_tables.scss#L624

---

We could also remove the `%table-cell-contract` since is not used anywhere
https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_tables.scss#L140

pat270:
Hey @eduardoallegrini,
There are some gotchas with using these utilities outside of `.table-autofit`. Every table cell needs a class on it (`table-cell-contract`, `table-cell-expand`, `table-cell-expand-small`, `table-cell-expand-smaller`, or `table-cell-expand-smallest`). There were some inconsistencies with table utility naming. I added aliases to make them more consistent.

For portal, we need to replace `table-cell-field` => `table-cell-contract`. The class `table-cell-content` => `table-cell-expand`.

@carloslancha I'll merge this once everything passes.

fixes #3818

### DIFF
--- a/packages/clay-css/src/content/tables.html
+++ b/packages/clay-css/src/content/tables.html
@@ -1924,6 +1924,89 @@ section: Components
 	</div>
 </div>
 
+<div class="clay-site-row-spacer row">
+	<div class="col-md-12">
+		<h3>Basic HTML Table with Expand</h3>
+
+		<div class="table-responsive">
+			<table>
+				<caption>Caption: Table with filler content</caption>
+				<thead>
+					<tr>
+						<th class="table-cell-expand">Country</th>
+						<th class="table-cell-expand-small">Purchasing Power Parity</th>
+						<th class="table-cell-expand-small">Official Exchange Rate</th>
+						<th class="table-cell-contract">Exports</th>
+						<th class="table-cell-contract table-cell-text-center"></th>
+					</tr>
+				</thead>
+				<tbody>
+					<tr>
+						<td class="table-cell-expand">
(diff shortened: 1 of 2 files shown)


carloslancha:
@pat270 do we need to update something in portal during the integration (need = something will break when upgrading version 😅 ) or it will look the same as in previous versions?

pat270:
@carloslancha nothing will break all the class names are still the same. The only thing that might cause a break is CSS specificity related, but I don't think it will be an issue.

pat270:
@carloslancha @eduardoallegrini I miss read the comment. Yes things will need to be updated in portal.
```
table-cell-field => table-cell-contract
table-cell-content => table-cell-expand
```
are the minimum updates needed.

eduardoallegrini:
we are updating them here https://issues.liferay.com/browse/LPS-102101

PR #3827, opened 2020-12-01
refactor(@clayui/card): Simplifies the use of the ClayCard component and separates it into components by responsibilities

### LINKED ISSUE #3789: ClayCardWithHorizontal can generate duplicate CSS class that breaks styles

matuzalemsteles:
Fixes #3789

This is an initial implementation to start simplifying the low-level component of ClayCard, it is currently very complex and was in [our sights for improvements and revision for v4](https://paper.dropbox.com/doc/Clay-4.x-API-review-and-improvements--BAejQ~GmATHhMLeDKFcBBZqqAQ-uUNnd4tTNFTU7lNP6yQfP). The changes here do not break the API, consequently, it corrects #3789 and there were many changes in the snapshot due to the fact that we removed the double declaration from `card` to horizontal and changes in class ordering.

The idea here was to break the ClayCard component into two more new components `<ClayCardHorizontal />` and `<ClayCardNavigation />`. These two new components are being used only internally but are not yet being public, we will work on that in the next interaction.

In the next major version, the idea is to deprecate the `interactive` API and the respective behaviors of the created components and remove the `ClayCardHybrid` component, which at the moment we will publicly mask as `ClayCard`.

- CardNavigation is a component that deals with a totally different template than ClayCard, visually speaking and markup, so we can start to separate it from the traditional Card we have to avoid confusion and complexity within the code.
- CardHorizontal follows the same idea as the component above, the markup structure is different in the `selectable` mode compared to the other ClayCard options.

Probably a change in the taglib, I am adding the `.card-type-directory` class together with the `.card card-horizontal` class.

### DIFF
--- a/packages/clay-card/src/Card.tsx
+++ b/packages/clay-card/src/Card.tsx
@@ -3,25 +3,24 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import ClayLink from '@clayui/link';
 import classNames from 'classnames';
 import React from 'react';
 
 import AspectRatio from './AspectRatio';
 import Body from './Body';
 import Caption from './Caption';
+import {ClayCardHorizontal} from './CardHorizontal';
+import {ClayCardNavigation} from './CardNavigation';
 import Context, {IContext} from './Context';
 import Description from './Description';
 import Group from './Group';
 import Row from './Row';
 
-type CardDisplayType = 'file' | 'image' | 'user';
-
 export interface ICardProps extends IContext {
 	/**
 	 * Determines the style of the card
 	 */
-	displayType?: CardDisplayType;
+	displayType?: 'file' | 'image' | 'user';
 
 	/**
 	 * Flag that indicates if the card can be selectable.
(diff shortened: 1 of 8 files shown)


packages/clay-card/src/CardHorizontal.tsx
```diff
@@ -0,0 +1,49 @@
...
+		{...otherProps}
+	>
+		{children}
+	</div>
+);
+
+export const ClayCardHorizontal: React.FunctionComponent<IProps> & {
+	Body: typeof ClayCardHorizontalBody;
+} = ({children, className, selectable, ...otherProps}) => (
+	<Context.Provider value={{horizontal: true, interactive: false}}>
```
matuzalemsteles, line 32: All components are using `Context` just to maintain compatibility and also because we use it directly on high-level components, we want to try to get rid of it in the future.

packages/clay-card/src/Card.tsx
```diff
@@ -35,78 +34,69 @@ interface IProps
...
 	);
 };
 
-ClayCard.AspectRatio = AspectRatio;
-ClayCard.Body = Body;
-ClayCard.Caption = Caption;
-ClayCard.Description = Description;
-ClayCard.Group = Group;
-ClayCard.Row = Row;
+const ClayCardHybrid: React.FunctionComponent<IProps> & {
```
matuzalemsteles, line 72: This component is created only to group all the behaviors that ClayCard currently has but that were separated into other components, in order to maintain compatibility. In the next major version, the idea is to remove and deprecated these behaviors and dry the ClayCard component low-level.

julien:
> Probably a change in the taglib, I am adding the .card-type-directory class together with the .card card-horizontal class.

Hey @matuzalemsteles just a question does that mean you plan updating the tag?

packages/clay-card/src/Card.tsx
julien: I know naming is hard, and maybe `ClayCardHybrid` makes sense, but if we can come up with another name I think it might be worth it

matuzalemsteles: @julien yeah, I really can't find another name for it but I would have no problem the idea of it is just to be internal and we removed it soon, also why I changed the `displayName` to `ClayCard` so that it is not visible outside.

matuzalemsteles:
> Hey @matuzalemsteles just a question does that mean you plan updating the tag?

@julien probably yes, depending on who does the release. I also see no problem in case I have to update.

julien:
@matuzalemsteles great thanks. 

I think that starting from now, if we make changes to the markup and if that component is used by a tag (which will probably be the case for most) we should take care of updating the tag. (cc @bryceosterhaus)

matuzalemsteles:
@julien yeah, we had discussed this in a thread on slack, now I can't find it anymore 🤷‍♂️. We had agreed to add this documented as part of the release process 😉.

julien:
@matuzalemsteles perfect :+1:

packages/clay-card/src/Card.tsx
```diff
@@ -35,78 +34,69 @@ interface IProps
...
 	className,
 	displayType,
-	horizontal,
-	href,
-	interactive,
-	onClick,
 	selectable = false,
 	...otherProps
-}: IProps) => {
+}) => {
```
bryceosterhaus, line 43: I think the `IProps` is casted here for our docs site. Could you confirm whether that API table still picks up this component without the type?

packages/clay-card/src/Card.tsx
bryceosterhaus: `Hybrid` should be fine and remains consistent with our naming in Label and Modal

bryceosterhaus (commented):
Looks good! Thanks for working on this @matuzalemsteles. I had a concern with the generated API table and wasn't sure if we were able to get the whole list of API from the Hybrid component to be available in our docs. Could you just verify that?

matuzalemsteles:
Thanks @bryceosterhaus! I will try to work on it in the afternoon and fix conflicts.

packages/clay-card/src/Card.tsx
matuzalemsteles: Hmm I'm going to take a look, I think it was to take the default export that is being `ClayCardHybrid` now. I'll check what's going on.

matuzalemsteles:
@bryceosterhaus about the generation of our API for Table it seems to have the same behavior as before, in this case it is not showing the `IContext` APIs that `ICardProps` extend but this is a fault of react-docgen that does not extract the interfaces from other files, is something we need to try to change soon. I got to work on a PoC a while ago but without much success and the alternative I saw was to build our own mechanism...

bryceosterhaus:
> it seems to have the same behavior as before

My main concern was loss of API docs, but since its the same, this is good for me! Feel free to merge

matuzalemsteles:
@bryceosterhaus of course, we are missing two APIs (`horizontal`, `interactive`) that are not visible in the table because of the problem I mentioned above, I will create an issue later so that we can track this table generation problem and think about how to improve it.

PR #3836, opened 2020-12-08
feat(@clayui/css): Mixins `clay-button-variant` allow styling `.activ…

### LINKED ISSUE #3797: @clayui/css Mixin clay-button-variant should be able to style only .active
There are times when we only want to style the `.active` class instead of `:active`. This mixin should allow that.

See https://github.com/liferay/clay/blob/e0a84fb91adebbf8005fce3bcb15abdb24ebd12e/packages/clay-css/src/scss/mixins/_buttons.scss#L375-L389

pat270:
…e` separately from `:active` with `active-class`

fixes #3797


marcoscv-work:
LGTM !

PR #3837, opened 2020-12-08
fix(@clayui/css): Mixins clay-link active-class should also style `&[…

### LINKED ISSUE #3767: @clayui/css Mixin `clay-link` add option for styling collapse items
Collapsible `nav-link` that is open should be able to be styled differently. See https://clayui.com/docs/components/nav/markup.html#css-nav-nested.

pat270:
…aria-expanded='true']` and `&.show` so we can style collapse plugin open state

fixes #3767

### DIFF
--- a/packages/clay-css/src/scss/mixins/_links.scss
+++ b/packages/clay-css/src/scss/mixins/_links.scss
@@ -351,6 +351,8 @@
 			@include clay-css($active);
 		}
 
+		&[aria-expanded='true'],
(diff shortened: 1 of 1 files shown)


marcoscv-work:
I love the use of `aria-expanded` or other aria or native attributes to provide styles in order to help (or even force) to use more semantic markup !

marcoscv-work:
LGTM !

Issue #3806, opened 2020-11-20
Table caption is not correctly color contrasted and the visual order is not correct


marcoscv-work:
## Add a `<caption>` to a Clay table

![image](https://user-images.githubusercontent.com/7963804/99804118-ccdc0180-2b3a-11eb-92cf-649109d3382a.png)

1) The contrast is not accessible:

![image](https://user-images.githubusercontent.com/7963804/99804467-58ee2900-2b3b-11eb-8a62-71aab4bea4ea.png)

2) The order is different in the code and in the render, it should be fixed:

![image](https://user-images.githubusercontent.com/7963804/99804568-7c18d880-2b3b-11eb-9bbc-f6b44c18a6ee.png)
but:
![image](https://user-images.githubusercontent.com/7963804/99804596-86d36d80-2b3b-11eb-9d0e-d57a01e53a21.png)

By CSS property:
![image](https://user-images.githubusercontent.com/7963804/99804655-9eaaf180-2b3b-11eb-9bf4-44b27c1f530b.png)

Captions can be added from ckeditor:
https://issues.liferay.com/browse/LPS-123682

PR #3842, opened 2020-12-09
fix(clay-css): Tables caption should be at the top and Atlas theme te…

pat270:
…xt color should be darker

issue #3806


pat270:
Going to merge since we don't have tests for 2.x

pat270:
@bryceosterhaus just merged it.

Issue #3838, opened 2020-12-09
ClayToggle should have `label` root element


markocikos:
[ClayToggle](https://github.com/liferay/clay/blob/master/packages/clay-form/src/Toggle.tsx) should have markup that matches [documentation](https://clayui.com/docs/components/toggle-switch/markup.html), including `label` as the root element. See more on benefits of `label` associated with `input` [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label).

Currently, the root element is `span`.

| Tech  | Version |
| ----- | ------- |
| @clayui/form  | v3.13.0  |

PR #3843, opened 2020-12-09
fix(@clayui/form): ClayToggle should use a label element instead of span

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/3838

### DIFF
--- a/packages/clay-form/src/Toggle.tsx
+++ b/packages/clay-form/src/Toggle.tsx
@@ -45,7 +45,7 @@ const ClayToggle = React.forwardRef<HTMLLabelElement, IToggleProps>(
 		}
 
 		return (
-			<span
+			<label
 				className={classNames(
 					'toggle-switch',
 					'simple-toggle-switch',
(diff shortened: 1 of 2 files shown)


markocikos (approved):
LGTM!

matuzalemsteles (commented):
Hey @bryceosterhaus looking at the markup it seems that we no longer have the `.toggle-switch-check-bar` wrapper, @pat270 can you confirm that this is it?

packages/clay-form/src/Toggle.tsx
```diff
@@ -45,7 +45,7 @@ const ClayToggle = React.forwardRef<HTMLLabelElement, IToggleProps>(
 		}
 
 		return (
-			<span
+			<label
 				className={classNames(
 					'toggle-switch',
 					'simple-toggle-switch',
```
pat270, line 51: Some users might not want to use `simple-toggle-switch`. I've seen commerce use toggle switches in narrow columns that make `simple-toggle-switch` look funny. Do we have a way to only use `toggle-switch`?

pat270:
@matuzalemsteles `.toggle-switch-check-bar` is new from v2.11.1+. It aligns the hidden checkbox to the toggle-switch overlay. Our documentation on clayui.com is incorrect.

PR #3844, opened 2020-12-09
chore(release): publish v2.22.0

bryceosterhaus:


bryceosterhaus:
travisci is being weird, I think this is still safe since https://travis-ci.org/github/liferay/clay/builds/748646037 is passing

PR #3853, opened 2020-12-15
feat(@clayui/multi-select): Use ClayButtonWithIcon instead of <button>

### LINKED ISSUE #3850: ClayMultiSelect is using a `button` instead of `ButtonWithIcon`
I was wonderning why the [`MultiSelect` component renders a `button`](https://github.com/liferay/clay/blob/03e1f07a341634dbfececd97bce9bd5f3b2f5fcb/packages/clay-multi-select/src/index.tsx#L360) instead of a [`ClayButtonWithIcon`](https://github.com/liferay/clay/blob/master/packages/clay-button/src/ButtonWithIcon.tsx)?

No that it's wrong but I was just wondering why.

We're creating a java tag for this component and we'd like both the tag and our React component to be consistent when it comes to what we're rendering.

kresimir-coko:
Fixes #3850 

I've tested this in Storybook and on DXP, it seemed to work fine from my light testing.

### DIFF
--- a/packages/clay-multi-select/src/index.tsx
+++ b/packages/clay-multi-select/src/index.tsx
@@ -359,6 +359,7 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 							<ClayInput.GroupItem shrink>
 								<ClayButtonWithIcon
 									className="component-action"
+									displayType="unstyled"
 									onClick={() => {
 										onClearAllButtonClick();
 
--- a/packages/clay-multi-select/src/index.tsx
+++ b/packages/clay-multi-select/src/index.tsx
@@ -4,9 +4,9 @@
  */
 
 import ClayAutocomplete from '@clayui/autocomplete';
+import {ClayButtonWithIcon} from '@clayui/button';
 import ClayDropDown from '@clayui/drop-down';
 import {ClayInput} from '@clayui/form';
-import ClayIcon from '@clayui/icon';
 import ClayLabel from '@clayui/label';
 import {FocusScope, Keys, noop, sub} from '@clayui/shared';
 import classNames from 'classnames';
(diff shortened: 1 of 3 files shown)


matuzalemsteles (commented):
@kresimir-coko LGTM! Visually it seems to work well I just see that now we are adding the border when it is focused I don't know if that would be a problem and we are also adding some unnecessary classes that don't seem to have much effect, like `btn-primary` and `btn-monospaced`.

packages/clay-multi-select/src/index.tsx
```diff
@@ -357,7 +357,7 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 						!disabledClearAll &&
 						(inputValue || items.length > 0) && (
 							<ClayInput.GroupItem shrink>
-								<button
+								<ClayButtonWithIcon
```
julien, line 360: @kresimir-coko don't forget to pass a `displayType`, by using the default "primary" it's not going to look like we want it to.

kresimir-coko: Good catch @julien, I'll take a look at that soon

julien: :+1:

kresimir-coko:
@matuzalemsteles @julien  I've pushed a change that adds a `displayType="unstyled"` which should be close to what we want in DXP. I believe that we need to change the CSS if we wanted to completely remove the background on hover.
As for the border on focus, I find it adds "free" A11y, which is great.

julien:
@kresimir-coko sounds good, concerning the CSS override in portal let's ping @marcoscv-work and @edalgrin

matuzalemsteles:
@kresimir-coko no problem, I think keeping the fund and the A11y looks like it was expected, I was looking at the documentation now when it was requested https://docs.google.com/document/d/1LJ8yiO8V5mSEGOjaYzHd5Nex_De_8fMRQAEKQoHMiiM/edit.

matuzalemsteles:
Well, I don't see any big problems going with this... @bryceosterhaus do you see anything?

bryceosterhaus (approved):
LGTM!

PR #3854, opened 2020-12-15
chore(@clayui/card): fixes the test of not rendering a sticker on CardWithInfo and improves type

matuzalemsteles:
From https://github.com/liferay/clay/pull/3819#discussion_r534104466

We already had this behavior being done where setting `null` to `stickerProps` he did not render the Sticker in CardWithInfo and passing the `undefined` he would render the default Sticker. I just did correct the test, as we have a default value for `stickerProps`, when the value of `stickerProps` is `undefined` it starts with its default value so rendering the default Sticker ie just `null` will not render the sticker.


bryceosterhaus:
LGTM!

Issue #3879, opened 2021-01-07
Labels: Review usage of the times icon for component library


hold-shift:
Hi there!

Due to LEXI-728, We are reviewing the usage of the times icon in Labels in order to reflect in our library what is in the published working component. The usage expected should include the "times-small" icon at 16x16px.

**Current**
![Captura de pantalla 2021-01-07 a las 8 13 05](https://user-images.githubusercontent.com/61543739/103862981-52524b80-50c0-11eb-8fea-a2ffbbd2718e.png)

**Expected**
Using "times-small" icon at 16x16px within the surrounding frame. With few exceptions, we are using this size as a base. We recommend keeping them that way to better manage a future update of the icon library without altering their proportions/path size.

pat270:
We had discussion of this exact situation for Alerts at https://github.com/liferay/clay/issues/3557. I have no problem with this.

/cc @matuzalemsteles @bryceosterhaus

bryceosterhaus:
I can take a look at this today

pat270:
I ended up reverting the change made in #3557 but I can't seem to find the reason. I need to make some tweaks to the CSS to size the clickable area to be 16x16. We can swap out the icon to `times-small` or change the font size of `times` later.

bryceosterhaus:
@pat270 yeah it looks like it needs to change from `0.875rem` on `.alert .close` to `1rem`. 

And then additionally we need to update the react components to use `times-small` instead of `times`

bryceosterhaus:
Some testing... 
`1rem` on `times-small` results in 14x14
`1rem` on `times` results in 16x16

@hold-shift do you happen to have a list of places where `times-small` is supposed to be used over `times`?

hold-shift:
Sure. I can only reach the use of `times-small` in the label component. Due to we are only applying it here, the decision was to unify it with the `times` icon, but it does not seem convenient since scaling the icon also scales the stroke. This icon is meant to appear inserted in 16x16px boxes too, so we need this alternate version to place it there without overlapping edges.

bryceosterhaus:
@hold-shift so the only place we should use `times-small` is in label?

hold-shift:
Yes, as far as I can reach. [In this Figma File](https://www.figma.com/file/tqJ2SAR9VXbvU9UelnKDk3/LEXI-728-Review-usage-of-the-times-icon-for-component-library?node-id=36%3A0), you can see the components that we manage for both icons.

[...]

PR #3882, opened 2021-01-11
fix(@clayui/label): close icon should be times-small

bryceosterhaus:
Part of the fix for https://github.com/liferay/clay/issues/3879

@pat270 I wasn't totally sure what css changes to make for this, but I updated the components so that Label now uses times-small. Let me know what changes need to be made to address this issue in full.


pat270:
@bryceosterhaus I pushed the CSS changes here, should be good to go.

carloslancha:
Hey @bryceosterhaus this will need `LabelTag` to be updated in portal. How are we tracking this finally? Adding some labels in the issue/pr? Just checking it after the release? 

(Sorry if this was explained before 😅 )

PR #3894, opened 2021-01-18
feat(@clayui/drop-down): add props for making menu different widths

### LINKED ISSUE #3790: Making ClayDropdown.Menu width customisable
In commerce would like to to use the dropdown menu to show some global search results. 
The current default width is not enough and it's not customisable.  

In ClayUI.com / dropdown / markup section I found [this](https://clayui.com/docs/components/drop-down/markup.html#css-wide) but the react component doesn't handle it:
https://github.com/liferay/clay/blob/9270625e8bafa20c1e77ab2cbe67ebf5560d5584/packages/clay-drop-down/src/Menu.tsx#L256

Could we add a prop to define the preferable width for the dropdown? 
`default`, `wide` and `full` are enough for our user cases

bryceosterhaus:
fixes #3790

This adds the css utils to the react components.

@pat270, would you mind just verifying this is how its supposed to work? You can check out the storybook preview here and then toggle the "knobs" to see the different widths.

### DIFF
--- a/packages/clay-drop-down/README.mdx
+++ b/packages/clay-drop-down/README.mdx
@@ -18,6 +18,7 @@ import {
     -   [Using dropdown without a trigger](#using-dropdown-without-a-trigger)
 -   [DropDownWithItems](#dropdownwithitems)
 -   [DropDownWithDrilldown](#dropdownwithdrilldown)
+-   [Caveats](#caveats)
 
 </div>
 </div>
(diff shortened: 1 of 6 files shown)


pat270:
@bryceosterhaus yep this is exactly how it should work.

matuzalemsteles (commented):
It is worth adding a consideration about the use of `width` with `auto`, some people may think that `auto` will respect the size of the node parent but since the Menu is rendered in the body due to the `Portal` , that will not happen. I think we can document that.

bryceosterhaus:
@matuzalemsteles good point, I will update docs and add that note

bryceosterhaus:
@matuzalemsteles updated docs

matuzalemsteles (commented):
LGTM! Thanks @bryceosterhaus

PR #3897, opened 2021-01-20
fix(@clayui/panel): make sure <button> has a type

### LINKED ISSUE #3896: ClayPanel: The page is refreshed after click on the Header, inside a Form Tag

Here: https://codesandbox.io/s/jovial-wind-bgk8k?file=/src/panel.js

The page should not be reloaded inside a Form Tag

## Environment

| Tech  | Version |
| ----- | ------- |
| @clayui/panel  | v3.3.4  |
| React |  16.12.0       |

bryceosterhaus:
fixes #3896

### DIFF
--- a/packages/clay-panel/src/index.tsx
+++ b/packages/clay-panel/src/index.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import ClayButton from '@clayui/button';
 import ClayIcon from '@clayui/icon';
 import {setElementFullHeight, useInternalState} from '@clayui/shared';
 import classNames from 'classnames';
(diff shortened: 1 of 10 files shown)


packages/clay-panel/src/index.tsx
```diff
@@ -127,6 +127,7 @@ const ClayPanel: React.FunctionComponent<IProps> & {
 						)}
 						onClick={() => setInternalExpanded(!internalExpanded)}
 						role="tab"
+						type="button"
```
matuzalemsteles, line 130: @bryceosterhaus in that case, would we be able to use `ClayButton`?

matuzalemsteles (commented):
Just left a comment! 🙂

packages/clay-panel/src/index.tsx
bryceosterhaus: Good catch. I just updated to make it a little more robust and updated a few more components.

matuzalemsteles (approved):
LGTM

PR #3903, opened 2021-01-25
fix(@clayui/css): Atlas `color-yiq` function not useable in Atlas var…

### LINKED ISSUE #3708: Bootstrap 4's `color-yiq` function doesn't work inside Atlas variables
https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/bootstrap/_functions.scss#L63-L75

~Bootstrap 4 assumes the variable `$yiq-contrasted-threshold` is declared before the function is ever called. This makes it so we can't use this in variables declared in `_clay_variables.scss`. We should duplicate the function and add a parameter `$threshold: 150` instead of using `$yiq-contrasted-threshold` directly.~

We just need to define `$yiq-contrasted-threshold` in Atlas `_globals.scss`

pat270:
…iables due to `$yiq-contrasted-threshold` undefined

fixes #3708

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_globals.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_globals.scss
@@ -166,6 +166,7 @@ $theme-colors: map-deep-merge(
 	$theme-colors
 );
 
+$yiq-contrasted-threshold: 150 !default;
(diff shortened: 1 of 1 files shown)


marcoscv-work:
LGTM!

PR #3904, opened 2021-01-26
chore(docs): add script for generating global changelog for each release

bryceosterhaus:
Since we are planning on releasing tomorrow I thought I could throw together a super crude idea of auto-generating the global changelog with each release. Let me know what you think

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
 		"format:check": "prettier --list-different -- \"**/*.{js,ts,tsx,md,mdx,json,scss}\"",
 		"genBuildSize:compare": "NODE_ENV=production node scripts/check-size/index.js --compare",
 		"genBuildSize": "NODE_ENV=production node scripts/check-size/index.js",
+		"genGlobalChangeLog": "node scripts/generate-global-changelog.js && prettier --write -- \"CHANGELOG.md\"",
 		"lerna": "lerna bootstrap -- --no-optional --no-package-lock",
 		"lint:changed": "git ls-files -mz \"**/*.js\" \"**/*.ts\" \"**/*.tsx\" | xargs -0 eslint",
 		"lint:fix": "eslint --fix \".*.js\" \"**/*.{js,ts,tsx}\"",
@@ -33,7 +34,7 @@
 		"storybook:build": "build-storybook -c .storybook -o .storybook-static",
 		"storybook": "start-storybook -p 8888",
 		"test": "jest",
-		"version": "yarn run format:all"
+		"version": "yarn genGlobalChangeLog && prettier --write -- \"packages/**/CHANGELOG.md\" && git add -- \"CHANGELOG.md\" \"packages/clay-css/src/scss/_license-text.scss\""
 	},
 	"devDependencies": {
 		"@babel/cli": "^7.10.1",
--- a/scripts/generate-global-changelog.js
+++ b/scripts/generate-global-changelog.js
@@ -1,0 +1,135 @@
+/**
+ * SPDX-FileCopyrightText: © 2021 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {execSync} = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+const dateObj = new Date();
+
+const month = dateObj.getMonth() + 1;
+const day = dateObj.getDate();
+const year = dateObj.getFullYear();
+
+const releaseDateString = `${year}-${month}-${day}`;
+
+const globalChangeLog = path.join(__dirname, '../CHANGELOG.md');
+
+const VERSION_HEADING = /^#+ \[\d/;
+
+const VERSION_LINK = /^#+ (\[\d+\.\d+\.\d+\]\(\S+\))/;
+
+function getChanges(file, packageName) {
+	let latestVersionHeader;
+	const changes = [];
+
+	const lines = fs.readFileSync(file, 'utf8').split('\n');
+
+	for (const line of lines) {
+		if (VERSION_HEADING.test(line)) {
+			if (latestVersionHeader) {
+				// Making it into this block indicates that we are now looking at the
+				// second to last release, one that should already be in the CHANGELOG.
+				// This means we already have the content we need.
+				break;
+			}
+
+			latestVersionHeader = line;
+		}
+
+		// The second condition makes sure we don't add the header to our changes.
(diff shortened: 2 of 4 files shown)


brunofernandezg:
I'd appreciate it a lot :pray:

scripts/generate-global-changelog.js
```diff
@@ -0,0 +1,108 @@
...
+const day = dateObj.getDate();
+const year = dateObj.getFullYear();
+
+const dateString = `${year}-${month}-${day}`;
+
+const globalChangeLog = path.join(__dirname, '../CHANGELOG.md');
+
+function main() {
+	const diffOutput = execSync(
+		'git diff HEAD^ --name-only -- "*CHANGELOG.md"'
```
wincent, line 27: You can get rid of the `IGNORE_LIST` and make this less imprecise if you switch to something like:

```
git diff HEAD^ --name-only -- 'packages/clay-*/CHANGELOG.md'
```

scripts/generate-global-changelog.js
```diff
@@ -0,0 +1,108 @@
...
+	).toString();
+
+	const changelogContents = fs.readFileSync(globalChangeLog).toString();
```
wincent, line 30: Pro-Tip™: don't need the `toString()` call if you specify the expected encoding (which is probably a good idea anyway):

```js
const changelogContents = fs.readFileSync(globalChangeLog, 'utf8');
```

scripts/generate-global-changelog.js
```diff
@@ -0,0 +1,108 @@
...
+	if (changelogContents.includes(`## ${dateString}`)) {
+		console.error(
+			`CHANGELOG has already been generated for the ${dateString} release`
+		);
+
+		return;
+	}
+
+	const files = diffOutput
+		.split(/\n/)
```
wincent, line 41: Obviously not going to cause problems here, but as a matter of best practice, if you're going to be scripting with the output of Git commands you should use `NUL`-delimited output formats whenever they're available (and they usually are); that way, even if a filename contained a newline, you'd still parse the output correctly.

ie. add `-z` to your `git diff` invocation above, and instead `.split('\0')`

scripts/generate-global-changelog.js
```diff
@@ -0,0 +1,108 @@
...
+
+	for (let i = 0; i < files.length; i++) {
+		const file = files[i];
+
+		try {
+			const fileDiff = execSync(`git diff HEAD^ -U0 -- "${file}"`)
+				.toString()
+				.split(/\n@@.*/)[1]
+				.trim()
+				.replace(/(\n|^)\+/g, '\n');
```
wincent, line 55: One of the problems with `git diff` is that there are a _lot_ of settings that people can put in their `~/.gitconfig` that can change the output (see the references to `diff.` in `man git-config`), so it would probably be more robust to just read the file contents.

```js
const VERSION_HEADING = /^#+ \[\d/;

function getChanges(file) {
    const changes = [];
    
    for (const line of fs.readFileSync(file, 'utf8').split('\n')) {
        if (VERSION_HEADING.test(line)) {
            if (changes.length) {
                break;
            }
            else {
                changes.push(line);
            }
        }
    }

    return changes.join('\n');
}
```

(Untested, written in browser, so may have bugs.)

scripts/generate-global-changelog.js
```diff
@@ -0,0 +1,108 @@
...
+				new RegExp(versionLinkRegex)
+			);
+
+			const versionLink = versionLinkMatch ? versionLinkMatch[0] : null;
+
+			// Be sure to properly nest headers in proper order.
+			// This order is CHANGELOG => date => package => fix type => commit
+			let content = fileDiff.replace(/(\n|^)#\s/gm, '\n#### ');
+			content = content.replace(/(\n|^)##\s/gm, '\n##### ');
+			content = content.replace(/(\n|^)###\s/gm, '\n###### ');
```
wincent, line 80: If your goal is just to add `###` to each heading, I think you could do these in a single step:

```js
let content = fileDiff.replace(/^(#+)/gm, '$1###');
```

Note that because of the `m` flag, `^` will match at the beginning of any line, so you don't need to explicitly deal with `\n` anywhere.

If you want to keep the separate `replace()` calls you could chain them like you did on lines 58-60 above, for consistency.

scripts/generate-global-changelog.js
```diff
@@ -0,0 +1,108 @@
...
+const year = dateObj.getFullYear();
+
+const dateString = `${year}-${month}-${day}`;
+
+const globalChangeLog = path.join(__dirname, '../CHANGELOG.md');
+
+function main() {
+	const diffOutput = execSync(
+		'git diff HEAD^ --name-only -- "*CHANGELOG.md"'
+	).toString();
```
wincent, line 28: Are there any sanity-checks that we can/should add here? Like, print a warning if the worktree is dirty, or if there are no changes? etc (I don't have any specific ideas here, but just trying to think of possibly "failure modes" where the script is called at the wrong time or in the wrong way.)

bryceosterhaus:
@wincent updated with your suggestions. Reading the file directly instead of the diff was a great suggestion!

I'll likely merge this today so that we can get it working for the release later today.

package.json
```diff
@@ -33,7 +34,7 @@
 		"storybook:build": "build-storybook -c .storybook -o .storybook-static",
 		"storybook": "start-storybook -p 8888",
 		"test": "jest",
-		"version": "yarn run format:all"
+		"version": "yarn genGlobalChangeLog && prettier --write -- \"packages/**/CHANGELOG.md\" && git add -- \"CHANGELOG.md\" \"packages/clay-css/src/scss/_license-text.scss\""
```
bryceosterhaus, line 37: I had to add these `git add` commands so that lerna would include the changes to the CHANGELOG to the publish commit. I was hoping to use [their config option](https://github.com/lerna/lerna/blob/main/commands/version/README.md#--no-granular-pathspec) for this, but I wasn't able to get it working properly.

The license file was an issue we needed to address and it was the same solution as the CHANGELOG so I included that here.

scripts/generate-global-changelog.js
bryceosterhaus: I added a check to make sure we actually have changelogs that have changed.

We expect the tree to be dirty since this script is called AFTER changelogs are updated but BEFORE they are commited

package.json
wincent: Oh well, at least it is a very specifically targeted `git add`, so it's unlikely to do anything nefarious.

PR #3906, opened 2021-01-26
feat(@clayui/button): add api for stacking button group vertically

### LINKED ISSUE #3788: A property called `vertical` for ClayButton.Group
## What is your proposal?

There is a way to stack [buttons vertically mentioned on markup documentation](https://clayui.com/docs/components/button-group/markup.html#vertical-variation) but We don't have a property for adding this behaviour on ClayButton.Group

More intuitive Button Group component

Implement `<div class="btn-group-vertical" role="group">` markup instead

bryceosterhaus:
fixes #3788


matuzalemsteles (approved):
LGTM!

PR #3907, opened 2021-01-26
feat(@clayui/drop-down): add api for controlling the active state of the menu

### LINKED ISSUE #3776: onActiveChange on ClayDropDownWithItems
## What is your question?

Recently I was migrating a component from 2.x to 3.x, this component receives a property called `functions` where the shape is an array containing objects shaped as 2.x API, in this case `{label: string, tooltip: string, value: string}`.  and I would not iterate whole array adding onClick functions before passing it to `items` property on ClayDropDownWithItems or maybe use a low-level composition, because the UI is simpler and I can reach the expected markup with just ClayDropDownWithItems.

The question is:

Does make sense adding `onActiveChange` to ClayDropDownWithItems?

bryceosterhaus:
fixes #3776

### DIFF
--- a/packages/clay-drop-down/src/Menu.tsx
+++ b/packages/clay-drop-down/src/Menu.tsx
@@ -121,6 +121,11 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	alignmentPosition?: number | TPointOptions;
 
 	/**
+	 * Flag to indicate if clicking outside of the menu should automatically close it.
+	 */
+	closeOnClickOutside?: boolean;
+
+	/**
 	 * Flag to indicate if menu is displaying a clay-icon on the left.
 	 */
 	hasLeftSymbols?: boolean;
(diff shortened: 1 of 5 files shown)


packages/clay-drop-down/src/Menu.tsx
```diff
@@ -187,6 +193,10 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>(
 
 		useEffect(() => {
 			const handleClick = (event: MouseEvent) => {
+				if (!closeOnClickOutside) {
+					return;
+				}
+
```
matuzalemsteles, line 199: @bryceosterhaus maybe we can move this validation when attaching the event, so we can avoid adding the event that will not be used. What do you think?

matuzalemsteles (commented):
LGTM! I just added a comment but it doesn't stop you from moving on.

packages/clay-drop-down/src/Menu.tsx
bryceosterhaus: Good idea!

PR #3917, opened 2021-02-02
chore: add gh action for commenting DXP release on closed issues in a milestone

### LINKED ISSUE #3898: Changelog Updates Proposal
### The Problem

It's difficult for DXP devs to know when a clay feature or issue has been resolved and available in DXP. This is likely due to an issue of traceability and discoverability.

#### Current Process

If a DXP dev has created an issue in this repo and sees that a fix is merged and thus closing their ticket. In order to determine if that fix is in DXP they would need to check the following.
- Is it released yet? Check if there has been a release since that PR has been merged.
- If released, check the version released for that module.
- Go to DXP and check if that version exists in `frontend-taglib-clay/package.json` for the given module.

However, that is more of a "clay centric" approach and requires the dev to know their way around the clay repo and assumes they had an issue they are specifically curious about.

More often than not, I think people just want to know "whats in the latest release?" when they see an LPS ticket in DXP with "Update @clayui/* dependencies". In order to know that a dev would need to look at the list of packages released and then go to each individual changelog. If 10+ packages have been released, this requires 10+ different docs and isn't ideal.

#### Proposal 

In order to fix this, I think we can implement a few processes to address this.

1. Use GH actions to automatically label issues with the "fix version". Each release we would spawn a GH action that would apply labels(something like `@clayui/label@v3.5.0`) to each issue that was fixed by the current release.
2. Generate more helpful changelogs for each package. I think we need to not only link to the commit, but also the issue that the commit stemmed from.
3. Generate a "global" changelog for each release. This would live at the root of the clay repo, and it would include the list of issues fixed after each publish. Example below
4. Copy/paste the global change log changes to the LPS ticket that is used to merge the release into DXP.

 --------
# Global Change Log

#### Bug Fixes

-   Close icon should be times-small (#3879)

#### Features

- Supports type `null` for displayType (#3896)

-------

#### Conclusion

I think if we adopt these changes (most of which would be automated), it would significantly increase the traceability and discoverability of what is released and how to know if it's available in DXP.

bryceosterhaus:
fixes #3898

This action works as follows

Event: Whenever we close a milestone
- Gets every closed issue in a milestone
- Grabs the Jira issue from the milestone title
- Comments on each issue and adds the link to track the LPS

### DIFF
--- a/.github/workflows/dxp-release-labeler.yml
+++ b/.github/workflows/dxp-release-labeler.yml
@@ -1,0 +1,36 @@
+name: 'DXP Release Commenter'
+on:
+    milestone:
+        types: [closed]
+
+jobs:
+    labeler:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/github-script@v3
+              with:
+                  github-token: ${{secrets.GITHUB_TOKEN}}
+                  script: |
+                      const jiraIssueMatch = context.payload.milestone.title.match(/^LPS-\d+/);
+
+                      if (jiraIssueMatch) {
+                        const LPS = jiraIssueMatch;
+
+                        const issues = await github.issues.listForRepo({
+                          milestone: context.payload.milestone.number,
+                          owner: context.repo.owner,
+                          repo: context.repo.repo,
+                          state: 'closed'
+                        });
+
+                        if (issues.data) {
+                          for (const issue of issues.data) {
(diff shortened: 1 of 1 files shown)


wincent (approved):
Looks sensible. Hope it works! 😁

brunofernandezg:
looks great 👍

PR #3927, opened 2021-02-10
feat(@clayui/drop-down): add util for aligning dropdown on scroll

### LINKED ISSUE #3804: ClayDropdown.Menu panel misalignment within a ClayModal when the modal body is scrollable

https://codesandbox.io/s/peaceful-bhabha-lfv6d?file=/src/index.js

If you use ClayModal.Body as scrollable and use a ClayDropdown.Menu component within, the dropdown panel will not be realigned to the trigger element.

Kapture 2020-11-19 at 14 42 24

Dropdown panel still aligned with the trigger element.

Using `observeRect` from '@reach/observe-rect' may solve it. As the example below:

https://github.com/liferay-frontend/liferay-portal/pull/430/files#diff-298ea5905081aa95c932d1681a7bcf60538269e3e9a96379a504259364432271R99-R101

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.8.0  |
| React |  XXXX   |

bryceosterhaus:
fixes #3804

I added our own observer util with inspiration from observe-rect and seems to work as intended.

There is still an issue with the z-index of the drop-down, and I'm not quite sure what we want to do about that. Perhaps the best option is to check if the dropdown is still in the view(modal or document) and if it is not, we close the drop-down. Although I am fearful that might lead to poor UX.

Kapture 2021-02-10 at 15 33 09

### DIFF
--- a/packages/clay-breadcrumb/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-breadcrumb/src/__tests__/__snapshots__/index.tsx.snap
@@ -38,6 +38,7 @@ exports[`ClayBreadcrumb calls callback when an item is clicked 1`] = `
       <button
         class="dropdown-toggle breadcrumb-link btn btn-unstyled"
         data-testid="breadcrumbDropdownTrigger"
+        style=""
         type="button"
       >
         <svg
@@ -116,6 +117,7 @@ exports[`ClayBreadcrumb renders 1`] = `
       <button
         class="dropdown-toggle breadcrumb-link btn btn-unstyled"
         data-testid="breadcrumbDropdownTrigger"
+        style=""
         type="button"
       >
         <svg
@@ -195,6 +197,7 @@ exports[`ClayBreadcrumb renders with properties passed by \`ellipsisProps\` 1`] 
       <button
         class="dropdown-toggle breadcrumb-link btn btn-unstyled"
         data-testid="breadcrumbDropdownTrigger"
+        style=""
         type="button"
       >
         <svg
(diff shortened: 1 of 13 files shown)


pat270:
This might be a good case for allowing relative dropdowns. See https://github.com/liferay/clay/issues/3790#issuecomment-726961721. Is it possible to create a portal inside modal-body?

bryceosterhaus:
@pat270 yeah our current implementation of Portal will automatically put the child portal within the parent portal. So for this case, Modal is rendered in a portal and then we have dropdown inside a portal which is inside of the modal portal.

Looks like we would need to move this up a few levels to `modal-body` though

Screen Shot 2021-02-10 at 3 51 53 PM

matuzalemsteles:
This works great, great work! Regarding the `z-index` it would really be a difficult thing to do via CSS, but I think if disabling the suspension of elements can we have other problems in the case for Modal?

Here are some of the examples https://storybook.clayui.com/?path=/story/components-claymodal--w-autocomplete-and-dropdown, in which case they are not on `modal-body` it seems quite correct, otherwise we would have to expand the body so that the user could see the rest of the content but because it is an absolute position it would be even more difficult, it would have to be manipulated via JavaScript as well.

bryceosterhaus:
I experimented with putting the portal deeper within the modal.

Rendering inside of `modal-dialog`(current behavior)
Screen Shot 2021-02-16 at 2 55 29 PM

Rendering inside of `modal-content`(moving sub portal down one level)
Screen Shot 2021-02-16 at 3 13 43 PM

Moving the portal inside of `modal-body` is a bit more complicated since we require the modal to be composed with the ModalBody component.

If we move the portal inside of `modal-content` so that it doesn't overflow the modal, we might run into issues with autocomplete and dropdowns that render below the modal since we are now inside of a `overflow:hidden;` element and no pop-ups will render outside of it....

My thought is that we just keep the current behavior(this PR) and iterate once we run into any major UX issues.

matuzalemsteles:
@bryceosterhaus yeah, this is a very complicated behavior, although when I move it to position itself over all the elements inside a scroll in the case of the header, I think this would be the least problematic behavior so far. It makes sense to me, to go with this now.

PR #3916, opened 2021-01-28
feat(@clayui/color-picker): update color-picker API to allow for both color palettes

bryceosterhaus:
fixes #3910 

@pat270 can you verify some of these css changes I made? I noticed the styling was wonky when we added both palettes to the drop down. I took the specs from https://docs.google.com/document/d/1IUGl5VOWh6lqRa1baRoshVcytc8XbqVMkk30vRW4SM8/edit#heading=h.7dwfrx5xjmtw

PR #3931, opened 2021-02-11
fix(@clayui/color-picker): remove last-row classes and API that was added on accident

bryceosterhaus:
I accidentally merged a new API in this PR https://github.com/liferay/clay/pull/3916. I initially needed the API for adding the class name for last row, however that was removed in a later commit and I didn't remember to remove the API and markup from the component.


bryceosterhaus:
fairly trivial

PR #3933, opened 2021-02-12
fix(@clayui/css): _global-functions.scss move autogenerated $lx-icons variables to separate file

### LINKED ISSUE #3469: @clayui.com/css _global-functions.scss move autogenerated $lx-icons variables to separate file
Spawned from https://github.com/liferay/clay/pull/3458#issuecomment-654632178.

> It's easier to track autogenerated things independently so one knows what can be edited manually and what shouldn't. It also helps in cases like this where simply re-running the generation scripts is enough to get back to a consistent state.

We should consider moving the autogenerated `$lx-icons` variable to a separate file. I'm thinking of placing it in the same directory `clay-css/src/scss/functions/_lx-icons-generated.scss` and `@import '_lx-icons-generated.scss';` at the top of `_global-functions.scss`?

pat270:
fixes #3469


edalgrin:
Just started reviewing :)

edalgrin (approved):
LGTM!

Issue #3934, opened 2021-02-15
Clay Dropdown includes non-localized information for placeholder and aria-label attribute


antonio-ortega:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## Is there an example you can provide via codesandbox.com?
No, I can't.

## What are the steps to reproduce?
Using Clay 2.x, make use of ClayDropdown component passing searchable:true.

If using Liferay, in 7.2.x branch:

1. Create a new Form.
2. Add a "Select from list" field.
3. Fill, at least, 6 options for this new field.
4. Deploy Forms widget into a page.
5. Click on Select field and assert the text "Search for..." is used as placeholder and aria-label attribute.
6. Change your language to something different than English.

## What is the expected result?
"Search for..." text is translated into the new language.

However, that's not happening, since that text is hardcoded in ClayDropdownBase.soy ([here](https://github.com/liferay/clay/blob/2.x/packages/clay-dropdown/src/ClayDropdownBase.soy#L159))

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v2.x  |

PR #3935, opened 2021-02-15
fix(clay-dropdown) Passing  aria-label and placeholder as parameter so they can be localized

antonio-ortega:
Hi guys,

Can you have a look at this pull request, please?

It's fixing [issue#3934](https://github.com/liferay/clay/issues/3934).

I ended up adding non-localized default values to avoid breaking changes. This way, nothing is going to change for those users who are already making use of clay-dropdown and once we publish the new version, we can modify the uses of it in Liferay Portal.

Thanks.

Regards.

### DIFF
--- a/packages/clay-dropdown/src/ClayDropdownBase.soy
+++ b/packages/clay-dropdown/src/ClayDropdownBase.soy
@@ -156,7 +156,14 @@
 		<div class="dropdown-section">
 			<div class="input-group input-group-sm">
 				<div class="input-group-item">
-					<input aria-label="Search for..." class="form-control input-group-inset input-group-inset-after" data-oninput="{$_handleSearch}" placeholder="Search for..." type="text" ref="searchInput" />
+					<input 
+						aria-label="{msg desc=""}search-for{/msg}"
+						class="form-control input-group-inset input-group-inset-after" 
+						data-oninput="{$_handleSearch}" 
+						placeholder="{msg desc=""}search-for{/msg}" 
+						type="text" 
+						ref="searchInput" 
+					/>
 
 					<span class="input-group-inset-item input-group-inset-item-after">
 						{call ClayButton.render}


packages/clay-dropdown/src/ClayDropdownBase.soy
```diff
@@ -156,7 +162,7 @@
 		<div class="dropdown-section">
 			<div class="input-group input-group-sm">
 				<div class="input-group-item">
-					<input aria-label="Search for..." class="form-control input-group-inset input-group-inset-after" data-oninput="{$_handleSearch}" placeholder="Search for..." type="text" ref="searchInput" />
+					<input aria-label="{$ariaLabel}" class="form-control input-group-inset input-group-inset-after" data-oninput="{$_handleSearch}" placeholder="{$placeholder}" type="text" ref="searchInput" />
```
matuzalemsteles, line 165: Hey @antonio-ortega instead of exposing an API to configure we can use the `{msg desc=""}search-for{/msg}`? We usually did it in other places.

matuzalemsteles (commented):
Hey @antonio-ortega sorry for the delay, I just leaving a comment to reassess whether we need an API or just using `{msg}` would solve the problem.

You could also correct the commit confirmation message, it's missing `:` after the scope, apparently CI also complained about some things. If you don't have time to do this I can work on it later. Thanks!

packages/clay-dropdown/src/ClayDropdownBase.soy
antonio-ortega: That make complete sense. My lack of knowledge made me send that silly way to localize these values. I've just forced-pushed my changes in this branch with your suggestions! Thanks @matuzalemsteles !

matuzalemsteles: @antonio-ortega do not worry!

bryceosterhaus:
@antonio-ortega going to continue work in this PR https://github.com/liferay/clay/pull/3938

Sorry, I had duplicated your work and didn't see your PR initially 😬

PR #3936, opened 2021-02-16
feat(@clayui/panel): add 'show' class to collapsable panel's button

### LINKED ISSUE #3766: Collapsible ClayPanel add the class `show` to the button when open
We have instances where we need to style the open and close button in the open state, but there is no class to hook to.

clay-panel

The only attribute that we have is `aria-expanded="true"`. It would be nice to have a class that is easier to spot.

bryceosterhaus:
fixes #3766


matuzalemsteles (approved):
LGTM! Well despite adding this here, I didn't see any CSS effects being applied, is this expected, or will @pat270 still send a PR with the changes?

bryceosterhaus:
@matuzalemsteles no expected css changes on the clay side. This change is primarily for DXP developers to easily spot and target the button for panels to show that they are active.

PR #3938, opened 2021-02-16
fix(clay-dropdown): add localizable placeholder

### LINKED ISSUE #3934: Clay Dropdown includes non-localized information for placeholder and aria-label attribute

bryceosterhaus:
fixes #3934

I think this is how you add localized terms in .soy but I'm not super positive... Anyone see issues with this? I verified that search-for is a key in dxp.


matuzalemsteles:
Hey @bryceosterhaus I think you ended up doing the same as PR #3935, maybe we can continue with this PR? As I had commented on the other PR, I prefer that we do so than exposing an API to configure in this case.

cc @antonio-ortega

bryceosterhaus:
@matuzalemsteles my bad, I just saw that other PR right now. Not sure how I missed that one. I'll check it out and see what to do next.

PR #3942, opened 2021-02-18
feat(@clayui/shared): select next focusable element in doc when focus leaves the focus scope

### LINKED ISSUE #3340: Unable to move focus from dropdown menu item to next button

bryceosterhaus:
fixes #3340

@matuzalemsteles maybe you could take a look at this just to verify I didn't miss anything with FocusScope.

Essentially this now allows the focus manager to jump out of the react tree and on to the regular dom ordering of focusable items. Check out the first storybook for DropDown for an example.


packages/clay-drop-down/stories/index.tsx
```diff
@@ -96,22 +96,42 @@ const DropDownWithState: React.FunctionComponent<any> = ({
...
+			document
+				.getElementById('testtest')
+				.append(
+					document
+						.createRange()
+						.createContextualFragment(
+							'<button onclick="alert();">test</button>'
+						)
+				);
+		}, []);
```
bryceosterhaus, line 110: This is just a hack for storybook to render an element outside of the react tree for testing purposes

packages/clay-drop-down/stories/index.tsx
```diff
@@ -96,22 +96,44 @@ const DropDownWithState: React.FunctionComponent<any> = ({
...
+			if (node && document) {
+				node.append(
+					document
+						.createRange()
+						.createContextualFragment(
+							'<button onclick="alert();">test</button>'
+						)
+				);
+			}
+		}, []);
```
bryceosterhaus, line 112: This is just a hack for storybook to render an element outside of the react tree for testing purposes

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -232,6 +271,19 @@ export function useFocusManagement(scope: React.RefObject<null | HTMLElement>) {
 				// when have React Portal.
 				if (nextElement) {
 					setInteractionObserver(nextElement);
+				} else {
+					// When the next element is not in scope, we check for the next
+					// focusable element in the document that is not in the React tree.
+					// This is particularly used for Portals.
+					const docElements = collectDocumentFocusableElements();
+
+					const index = docElements.indexOf(elements[0]);
```
matuzalemsteles, line 280: @bryceosterhaus for cases where we can have a lot of elements in the scope, wouldn't that break? it wouldn't have to be `docElements.indexOf(elements[position]);`.

Maybe I'm understanding this case, probably the other elements are the DropDown buttons that are inside a Portal, if you take the last one you will not find the closest one to it, because it is in the body and it would have to be the DropDown trigger button. I think we would have to test this more in other use cases, considering more elements on the screen.

Anyway, I fell into the same challenge when it was developed, oriented towards DOM, the Portal was always the problem, I had to make some inferences to try to get the element that was not in the Portal and then get the next element from it.

One element to test this is the Pagination Bar, which maybe you can find some cases.

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -232,6 +271,19 @@ export function useFocusManagement(scope: React.RefObject<null | HTMLElement>) {
 				// when have React Portal.
 				if (nextElement) {
 					setInteractionObserver(nextElement);
+				} else {
+					// When the next element is not in scope, we check for the next
+					// focusable element in the document that is not in the React tree.
+					// This is particularly used for Portals.
+					const docElements = collectDocumentFocusableElements();
```
matuzalemsteles, line 278: Perhaps we should make the same case for when it is backwards? use of shift + tab.

matuzalemsteles (commented):
Hey @bryceosterhaus I put some considerations on some cases that I think may have a problem. Perhaps we can prepare a story with more elements and use cases.

bryceosterhaus:
@matuzalemsteles I re-visited this from scratch as my previous implementation really only worked for one specific use case. I my new approach is more generic and _should_ cover a multitude of use cases that I manually tested against.

In the process of this, I also removed a bunch of code that I don't think was necessary anymore, however I'm not super confidant about it. Could you test this out and give it a look? I'm sure you might remember some edge cases we had that I might not have accounted for. Feel free to add more stories to test as well. Once this starts to look and work how we want, I am going to go through and add some unit tests to make sure we don't accidentally break anything in the future.

matuzalemsteles:
@bryceosterhaus great, I took a quick look but a great job I will still look at the details and use cases more carefully in the late afternoon.

matuzalemsteles:
Just a status on this, after I added the message a few long days ago, I’ll set aside time in the morning tomorrow to look at this and add the use cases and concerns.

jbalsas:
Labelling as `on hold` and converting to Draft until @matuzalemsteles can get to it

matuzalemsteles:
Starting to see it now!

matuzalemsteles:
Hey @bryceosterhaus I know you are coming back and updating yourself so don't be in a hurry with that, I'm just analyzing the changes here and doing some manual tests, noting the example that was added in the story it seems that we have some strange behaviors when leaving the tree React to the DOM using `tab` and `shift` + `tab`, apparently, we had no navigation problems in use cases when it is inside the React the only tree.

matuzalemsteles (commented):
Hey @bryceosterhaus I think in general this is working very well, I think your story is covering the main use cases, so it looks very complete, maybe we should bring this to the tests I don't know if we can do this to avoid exposing the FocusScope to the public. A great job, you cleaned up some things.

I think the only point here is that we will not have the `shift` +` tab` to return to an element that is inside a Portal and just in case when it was an element of the DOM and not in the React tree, you can reproduce this in the story, try:

- `Tab` until you get to the `html2` element
- Press `shift` + `tab`

He will end up respecting the order of the document, turning the focus to React instead of going back to `link 3` for example, which would be the last element if he didn’t have the Portal.

I added the listener to the element exactly to try to cover this case but it ended up getting a little complex but we can probably handle it in a simpler way. It is also worth seeing if it is worth covering this case again since most of our components with Portal disappear, this is normally used for `DropDown.Menu` but we are not implementing this behavior in `DatePicker` or `ColorPicker`.

Let me know what you think, anything, I can take that and try to add the tests if you want.

bryceosterhaus:
Good catch on that bug. I'm not quite sure the best way to solve that, as I don't really want to have to listen for each element in the document and when it loses focus.

The logic to handle this would likely have to look something like...
- listen for blur on non-react element.
- When element blurs, check the previous element in the document.
- IF that element is in a react tree, focus the last element in that tree.

I'm just not totally sure how we can accomplish that cleanly and all within "FocusScope".

I'll keep investigating today and see if I can come up with anything

packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -5,261 +5,241 @@
 
 import React from 'react';
 
-import {Keys} from './Keys';
+// import {Keys} from './Keys';
```
matuzalemsteles, line 8: Just to remember to remove 🙂

matuzalemsteles:
Yeah I also think that it would be very expensive.

Well, I'm not sure if it would be very easy to deal with this by adding the focus listener to the scope or document, maybe there are some match problems. Previously I added the listener to the next element that is out of scope and as soon as I press `shift` + `tab` it goes back to the last element of the React tree but I had a lot of problems with that because I had to remove the listener in cases where it no longer existed or lost because the element was rebuilt for some reason.

bryceosterhaus:
@matuzalemsteles updated and added a test.

I investigated ways to handle backwards tabbing but couldn't find a good solution. I think for our case, we can get away with it for now since our primary use case is an overlay and its fine to shift+tab back to the root react element rather than the last focusable node in that react tree.

Let me know what you think!

packages/clay-shared/stories/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+/* eslint-disable no-sparse-arrays */
+
+import '@clayui/css/lib/css/atlas.css';
+import {storiesOf} from '@storybook/react';
+import React from 'react';
+
+import {ClayPortal, FocusScope} from '../src';
+
+storiesOf('Components|FocusScope', module).add('default', () => {
```
matuzalemsteles, line 14: Hey @bryceosterhaus I wanted not to make this visible in our storybook, can we make sure that this story does not appear in production only in the development environment? This would really be good for exploratory testing.

packages/clay-shared/src/__tests__/FocusScope.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {ClayPortal, FocusScope} from '..';
+
+describe('FocusScope', () => {
+	afterEach(() => {
+		jest.clearAllTimers();
+
+		cleanup();
+	});
+
+	it('manages focus order when using portals', () => {
```
matuzalemsteles, line 20: This looks good, I think it's a good start for us to be able to add more cases here.

packages/clay-time-picker/src/__tests__/index.tsx
```diff
@@ -71,7 +71,7 @@ describe('IncrementalInteractions', () => {
 
 	// Once we find a solution so that useFocusManagement can get rid of the offsetParent
 	// to check if the element is visible, in tests this will always be null.
-	it.skip('pressing arrow right should move focus to minutes input', async () => {
+	it('pressing arrow right should move focus to minutes input', async () => {
```
matuzalemsteles, line 74: Great to see it being activated! We can probably now remove the comment above.

matuzalemsteles (commented):
@bryceosterhaus yeah, I think we can get along with this for now, I just see the Date Picker and Color Picker components that need it but I think this is not critical because this case is only when the Date Picker is the last element in the React where the next focused element is outside the tree, so we can send the improvements we have here and we can review this in the future.

bryceosterhaus:
@matuzalemsteles updated!

The only note is that I still am publishing that story in prod mode. I think this is useful for us to be able to compare the current use on storybook.clayui.com and the preview builds. I did prefix it under a separate heading though so that its label as "internal only"

matuzalemsteles (approved):
Yeah no problem I think it helps, I also agree that we should have this in the storybook. Great Work! LGTM!

wincent:
Posted a couple of tiny follow-ups here:

- https://github.com/liferay/clay/pull/4031
- https://github.com/liferay/clay/pull/4032

PR #3944, opened 2021-02-18
chore: update lerna mode to use fixed instead of independent

### LINKED ISSUE #3939: Can we make the global changelog more useful?
So we have [a global changelog](https://github.com/liferay/clay/blob/master/CHANGELOG.md) now (since [#3898](https://github.com/liferay/clay/issues/3898)) and the bot that comments with an LPS link when PRs make it into releases that are bound for liferay-portal (since [#3917](https://github.com/liferay/clay/pull/3917)).

One small obvious change we could make to improve the usefulness of the global change log would be to aggregate the updates which are just "noise" (the "version bump only for package" ones) so that the ones which are "signal" (marked in pink in this screenshot) would stand out:

Screenshot 2021-02-15 -113757-A1aOmOS8@2x

Something like this:

```
# YYYY-MM-DD release

## @clayui/package-with-real-changes (x.y.z)

### Features

- ...

### Bug Fixes

- ...

## Version bumps only

- @clayui/package-without-significant-changes-a (x.y.z)
- @clayui/package-without-significant-changes-b (x.y.z)
```

A bigger change than just formatting the changelog in a more useful way would be to switch from independent releases to releasing everything at once. We've discussed this before (at least in [#3214](https://github.com/liferay/clay/issues/3214) and other places too, no doubt), but it's worth remembering that it is a possibility. IIRC, the main argument against is "why bump the version number for all the packages that didn't actually change in a release? isn't that misleading? wasteful?" but on the flip side it represents a massive simplification from the perspective of the user...

Instead of having about 39 different Clay packages with (literally) 23 different version numbers like we currently do in `liferay-portal`:

```
"@clayui/alert": "3.5.1",
"@clayui/autocomplete": "3.2.6",
"@clayui/badge": "3.2.0",
"@clayui/breadcrumb": "3.3.6",
"@clayui/button": "3.6.0",
"@clayui/card": "3.6.6",
"@clayui/charts": "3.2.0",
"@clayui/color-picker": "3.5.0",
"@clayui/css": "3.24.1",
"@clayui/data-provider": "3.3.10",
"@clayui/date-picker": "3.8.0",
"@clayui/drop-down": "3.9.1",
"@clayui/empty-state": "3.2.0",
"@clayui/form": "3.14.4",
"@clayui/icon": "3.1.0",
"@clayui/label": "3.4.1",
"@clayui/layout": "3.3.0",
"@clayui/link": "3.2.0",
"@clayui/list": "3.4.7",
"@clayui/loading-indicator": "3.2.0",
"@clayui/management-toolbar": "3.3.0",
"@clayui/modal": "3.8.5",
"@clayui/multi-select": "3.9.4",
"@clayui/multi-step-nav": "3.3.6",
"@clayui/nav": "3.4.5",
"@clayui/navigation-bar": "3.3.5",
"@clayui/pagination": "3.3.6",
"@clayui/pagination-bar": "3.2.6",
"@clayui/panel": "3.3.6",
"@clayui/popover": "3.5.5",
"@clayui/progress-bar": "3.2.0",
"@clayui/shared": "3.5.1",
"@clayui/slider": "3.2.0",
"@clayui/sticker": "3.3.0",
"@clayui/table": "3.1.0",
"@clayui/tabs": "3.3.5",
"@clayui/time-picker": "3.2.6",
"@clayui/toolbar": "3.0.6",
"@clayui/tooltip": "3.4.5",
"@clayui/upper-toolbar": "3.1.14",
```

we could have 39 packages with a single version number that could be easily understood by developers as "version x.y.z of all these Clay packages" that are all known to work correctly and cohesively together. This is what we did for Clay v2, as you probably recall.

Another argument against going away from independent versioning was that we'd lose the ability to make breaking changes (ie. say we wanted to bump _one_ package to v4 to make a breaking change, we'd have to bump _all_ of them). But the pressure to avoid breaking changes at all costs is probably a feature and not a bug. 🤔

Anyway, if we did that, the global changelog would lose the "version bump" updates entirely and could just be:

```
# x.y.z

## @clayui/package-with-actual-changes

### Features

- ...
```

(ie. packages without actual changes would just be omitted from the changelog entirely).

Anyway, I think both of these options are worth considering if you want to keep Clay as a separate repo, because it helps alleviate some of the pain points and makes it easier to fend off the case for moving it into DXP (ie. [#3849](https://github.com/liferay/clay/issues/3849)).

bryceosterhaus:
fixes #3939

Putting in draft just to get some eyes on this first before going with it

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
 	"lerna": "3.4.0",
-	"version": "independent",
+	"version": "3.24.1",
 	"npmClient": "yarn",
 	"useWorkspaces": true,
 	"command": {
@@ -10,9 +10,12 @@
 				"**/__tests__/**",
 				"**/*.md",
 				"**/*.mdx",
-				"**/docs/*"
+				"**/docs/*",
+				"**/packages/demos/**",
+				"**/packages/generator-clay-component/**",
+				"**/packages/browserlist-config-clay/**"
 			],
-			"message": "chore(release): publish updated packages"
+			"message": "chore(release): publish %s packages"
 		}
 	},
 	"packages": ["packages/*"]


jbalsas:
Makes sense!

wincent:
My feelings right now:

partay

matuzalemsteles:
lgtm!!!

Issue #3949, opened 2021-02-25
ColorPicker's RGB Input shouldn't accept any value other than numbers from 0 to 255


carloslancha:
The Custom ColorPicker is currently allowing to enter any character to the RGB inputs, so when you enter any non-numeric value it automatically changes to NaN.

PR #3950, opened 2021-02-25
fix(@clayui/color-picker): RGB inputs should only accept numeric values from 0 to 255

carloslancha:
Fixes https://github.com/liferay/clay/issues/3949

### DIFF
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -42,6 +42,26 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
 	const inputRef = React.useRef(null);
 	const [inputValue, setInputValue] = React.useState(value);
 
+	const handleOnChange = (event: any) => {
+		const value = event.target.value;
+
+		if (value === '') {
+			return;
+		}
+
+		let newVal = Number(value);
+
+		if (newVal < 0) {
+			newVal = 0;
+		} else if (newVal > 255) {
+			newVal = 255;
+		}
+
+		setInputValue(newVal);
+
+		onChange({[name]: newVal});
+	};
+
 	React.useEffect(() => {
 		if (document.activeElement !== inputRef.current) {
 			setInputValue(value);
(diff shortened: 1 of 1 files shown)


packages/clay-color-picker/src/Custom.tsx
```diff
@@ -42,6 +42,26 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
...
 
+	const handleOnChange = (event: any) => {
+		const value = event.target.value;
+
+		if (value === '') {
+			return;
+		}
+
+		let newVal = Number(value);
+
```
julien, line 53: This looks OK to me, but I was wondering why we weren't using `parseInt` 

Why not use?

```js
let newVal = parseInt(value, 10);`
```

Is there anything better about using the `Number` constructor (without `new`) to do the cast?
(Non blocking comment)

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -55,15 +75,11 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
...
-							setInputValue(newVal);
-
-							onChange({[name]: newVal});
-						}}
+						max="255"
+						min="0"
+						onChange={handleOnChange}
 						ref={inputRef}
-						type="text"
+						type="number"
```
julien, line 82: Makes sense since this is going to accept (numeric) values between 0 and 255 :+1:

julien (commented):
LGTM, just left a non blocking comment that we can discuss any time

packages/clay-color-picker/src/Custom.tsx
carloslancha: I think it's the same, I kept previous behavior, don't know if there's any advantage on any of them.

julien: Got it.

I know that there is a difference,  for example

```js
parseInt("");  
```
Will return NaN

And 

```js
Number(""); 
```
Will return 0

carloslancha: Then we should keep `Number`, 'cause if `""` is received as value (that happens when you enter a non-numeric character in an number input) we `return` to avoid any update.

julien: Well, you could still do the same with `parseInt`, since you're returning before casting the string to a number, right?

julien: Anyway, I don't think it makes much of a difference, I just personally don't like using constructors without `new` but the result is going to be the same.

Thanks for the interesting conversation Carlos :smile: 

I'll let the clay experts do the review :+1:

bryceosterhaus: > I'll let the clay experts do the review 👍

image

bryceosterhaus:
Updated logic re-sent at https://github.com/liferay/clay/pull/3953

bryceosterhaus:
Other PR I sent was 🗑️

PR #3952, opened 2021-02-25
fix(@clayui/drop-down): make sure element exists before passing to domAlign

### LINKED ISSUE #3951: dom-align package errors in DropDown.Menu
In our latest release `v3.25.0`, the `DropDown.Menu` occasionally emits errors if the component is mounted/unmounted quickly. This causes dom-align to to be called with an null value for the element.

bryceosterhaus:
fixes #3951

Bug was reproducible by going to storybook and quickly clicking into different DropDown stories

### DIFF
--- a/packages/clay-drop-down/src/Menu.tsx
+++ b/packages/clay-drop-down/src/Menu.tsx
@@ -244,7 +244,7 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>(
 			};
 		}, [active]);
 
-		const align = React.useCallback(() => {
+		const align = () => {
 			if (alignElementRef && alignElementRef.current) {
 				let points = alignmentPosition;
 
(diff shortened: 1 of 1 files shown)


packages/clay-drop-down/src/Menu.tsx
```diff
@@ -254,20 +254,22 @@ const ClayDropDownMenu = React.forwardRef<HTMLDivElement, IProps>(
...
-					{
-						offset: offsetFn(points),
-						overflow: {
-							adjustX: autoBestAlign,
-							adjustY: autoBestAlign,
-						},
-						points,
-					}
-				);
+				if ((ref as React.RefObject<HTMLElement>).current) {
```
wincent, line 257: So this is what you mention in the commit message. Is the removal of `useCallback` also a required part of the fix? Or unrelated?

bryceosterhaus: The check before calling domAlign was the necessary part to fix the bug.

Removing `useCallback` was motivated more out of simplicity since `align` really isn't expensive or necessary to be memoized. I initially thought `useCallback`'s dependencies might be the bug, which led me to realize it was probably an over-complication of the code.

wincent: Cool. Next time, that should go in the commit message (or better still, be a separate `refactor:` commit).

But overall lg2m.

kim-jong-approval

wincent (approved):

PR #3953, opened 2021-02-25
fix(@clayui/color-picker): RGB inputs should only accept numeric values from 0 to 255

### LINKED ISSUE #3949: ColorPicker's RGB Input shouldn't accept any value other than numbers from 0 to 255

bryceosterhaus:
fixes #3949

follow up from https://github.com/liferay/clay/pull/3950

@carloslancha I took a look into this and I think all we actually needed was just the type on the input and also the min/max. I removed the other changes you made and squashed it into your commit.

Let me know if I missed something

### DIFF
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -55,6 +55,8 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
 					<ClayInput
 						data-testid={`${name}Input`}
 						insetBefore
+						max="255"
+						min="0"
 						onChange={(event) => {
 							const newVal = Number(event.target.value);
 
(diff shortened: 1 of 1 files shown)


wincent (approved):
Continuing today's theme of `#DictatorApprovals`...

putin-approval

 lg2m.

carloslancha:
@bryceosterhaus if you remove the validation/control in the `onChange` handler you can type manually any number higher 255 and lower than 0 (negative). I added that control so if you type any number higher than that it automatically changes to 255, and the same with lower than 0.

wincent:
Interesting, seems you have to explicit tell the browser to do [constraint validation](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation).

bryceosterhaus:
@carloslancha ah I see, good catch. Going to close this PR and go with your original one.

PR #3955, opened 2021-02-25
chore: update release docs for 2.x

### LINKED ISSUE #3948: Evaluate how we release 2.x
Conversation started at https://github.com/liferay-frontend/liferay-portal/pull/814

> @wincent 
> I'd take the opposite stance. I'd rather know that "vX.Y.Z" of Clay means "the set of Clay packages that we know work well together", and liferay-portal always will have the latest (ie. the last-known/best-working version of every Clay package). We have a global changelog, so if I care to know the details, I'd look at that. Furthermore, when we update in liferay-portal, we take care of updating the version numbers in every module, so there's no burden for the consumers; they don't even need to know an update happened (especially because we don't make breaking changes). The counterscenario is that you wind up with what we currently have in v3, which is a bunch of different version numbers (as I reported here, there are 39 different Clay v3 packages and 23 distinct v3 version numbers at once in liferay-portal) — that seems much more confusing.

I like the idea of being able to think "this set of Clay packages work well together", and bumping all versions at the same time is one way to accomplish that. When thinking about that though, it makes me think that devs want a singular "all-in-one" package, rather than multiple segmented packages. Which luckily for our case, 2.x has that in the `clay` package. So maybe instead of providing every individual package to DXP we just provide `clay` and that way its all based off a single version.

bryceosterhaus:
fixes #3948

Updated docs so that we release all 2.x packages under the same version at the same time

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -194,18 +194,32 @@ You may want to use [Github search](https://help.github.com/articles/searching-i
 
 To publish a new version
 
-1.  `git checkout 2.x`
-2.  Make sure you have all of the tags from upstream. `git fetch $REMOTE --tags`
-3.  Run `lerna version --conventional-commits --no-push` and the new version change.
 
-    -   You can also use `lerna changed` to see an output of what packages will be updated before running `lerna version`.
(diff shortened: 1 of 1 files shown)


wincent (approved):
I don't know how to verify this other than by running the steps, so...

fingers-crossed

PR #3957, opened 2021-03-01
fix: issue 3958 (LPS-128384 removing firefox styles for fieldsets)

jgarcialr:
Fixes: #3958

Basically in older versions of firefox we add this piece of code to workaround fieldset behavior in firefox.
From firefox v86 this workaround doesn't behave as in older versions and prevents fieldset and their children nodes to get 100% of width of their parent node

### DIFF
--- a/src/scss/lexicon-base/_forms.scss
+++ b/src/scss/lexicon-base/_forms.scss
@@ -5,10 +5,6 @@ input[type="radio"] {
 
 fieldset {
 	word-wrap: break-word;
-
-	@-moz-document url-prefix() { // FF Fieldset workaround
-		display: table-cell;
-	}
 }
 
 label,


wincent:
[Could we please get a short description](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/pull_requests.md)? The [referenced LPS](https://issues.liferay.com/browse/LPS-128384) also doesn't say what the rationale for the change is, just _what_ the change is. I gather it's because some code was removed from DXP in the first half of 2020 in [LPS-112983](https://issues.liferay.com/browse/LPS-112983), but it would be good to have that explicitly confirmed.

jgarcialr:
hi @wincent 
I created a new issue related to this PR
https://github.com/liferay/clay/issues/3958

Basically in older versions of firefox we add this piece of code to workaround fieldset behavior in firefox.
From firefox v86 this workaround doesn't behave as in older versions and prevents fieldset and their children nodes to get 100% of width of their parent node

jbalsas:
I don't even remember how to publish `1.x`... hopefully @pat270 does! 🙏

wincent:
Thanks @jgarcialr. Semantic Pull Request bot is complaining because it wants commit messages (and PR titles) to start with `fix:` (force-pushing to update after amending is fine)

Also:

> how to publish 1.x

double-take

pat270:
This LGTM. I'm going to close this for https://github.com/liferay/clay/pull/3964 which has a commit message that makes Semantic PR happy.

PR #3960, opened 2021-03-01
fix: use isomorphic helpers for useLayoutEffect

### LINKED ISSUE #3956: Support of SSR in Clay Components
## What is your proposal?

Support of SSR in Clay Components

I know that Clay were designed to attend the Liferay Portal needs, but as Clay is Open-Source and free to use, would be great if Clay supports Server Side Rendering 

https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

Basically [this comment](https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85#gistcomment-2911761) could help to solve the SSR problem in this first moment

One possible solution

```
const useIsomorphicLayoutEffect =
  typeof window !== 'undefined' ? useLayoutEffect : useEffect
```

---

My actual problem, i have a NextJS application and when i use ClayDropDown, this error message is shown
```

Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
    at /@clayui/drop-down/lib/Menu.js:109:21
```

Thanks in advance !

bryceosterhaus:
fixes #3956

Thought it was overkill to add a helper in "shared" for this, so I just added it to each package individually.

### DIFF
--- a/packages/clay-charts/src/BillboardWrapper.tsx
+++ b/packages/clay-charts/src/BillboardWrapper.tsx
@@ -11,6 +11,9 @@ interface IProps extends ChartOptions {
 	elementProps?: React.HTMLAttributes<HTMLDivElement>;
 }
 
+const useIsomorphicLayoutEffect =
+	typeof window === 'undefined' ? React.useEffect : React.useLayoutEffect;
+
 const BillboardWrapper: React.FunctionComponent<IProps> = ({
 	forwardRef,
 	elementProps = {},
@@ -42,7 +45,7 @@ const BillboardWrapper: React.FunctionComponent<IProps> = ({
 		}
 	}, []);
 
-	React.useLayoutEffect(() => {
+	useIsomorphicLayoutEffect(() => {
 		requestAnimationFrame(() => updateChart(otherProps));
 
 		return () => {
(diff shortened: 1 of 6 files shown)


packages/clay-charts/src/BillboardWrapper.tsx
```diff
@@ -11,6 +11,9 @@ interface IProps extends ChartOptions {
 	elementProps?: React.HTMLAttributes<HTMLDivElement>;
 }
 
+const useIsomorphicLayoutEffect =
+	typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;
```
wincent, line 15: (Non-blocking) nit: a ternary like this is probably more readable without the negation.

```js
const a = b === c ? d : e;
```

vs.

```js
const a = b !== c ? e : d;
```

In this case because `'undefined'` is itself a kind of negation (ie. "not defined"), I think the case is even stronger because it reads like a _double_ negation (ie. "if window is not not defined").

```js
const useIsomorphicLayoutEffect =
	typeof window === 'undefined' ? React.useEffect : React.useLayoutEffect;
```

wincent (approved):
> Thought it was overkill to add a helper in "shared" for this, so I just added it to each package individually.

Fair enough. Everybody has a threshold for what they consider overkill. I used to work with a guy who would say, "two or more, use a `for`". 😂 Given that that is basically the maximal expression of "DRY", I was never sure if he was joking or just very zealous.

Issue #3958, opened 2021-03-01
Clay 1.x, remove fieldset workaround for firefox


jgarcialr:
Inside file _forms.scss
we have this piece of code to workaround fieldset on firefox
```
	@-moz-document url-prefix() { // FF Fieldset workaround
		display: table-cell;
	}
```
From firefox v86 this piece of code makes fieldsets and their childs not getting 100% width from it's parents.

If you remove this piece of code, it works as expected.
(LPS-128384)
Clay version   v.1.x

pat270:
I cut a new version of 1.x should be under lexicon-ux@1.0.31.

@jbalsas @wincent There are some issues with the release script that I'm going to fix and send a pr for. One of the major issues I found was that the script creates a 1.x-develop branch which we phased out some time ago. Can one of you delete it from the repo? I don't think I have permissions to do that.

wincent:
> One of the major issues I found was that the script creates a 1.x-develop branch which we phased out some time ago. Can one of you delete it from the repo? I don't think I have permissions to do that.

Done, but you do have permission to do it; if you have rights to push, that means you have the rights to modify branches (ie. not just adding commits, but force-updating branches and deleting them too).

How:

Fetch all the changes with `git remote update --prune`.

Inspect the branches to see what is pointing where. For this, I used `git graph upstream-rw/1.x{,-develop,-stable}*` (where `graph` is [an alias](https://github.com/wincent/wincent/blob/f71806fbb6217ad44517c661d1b35cee023342b6/aspects/dotfiles/templates/.gitconfig.erb#L30) for the rather horrid, `git --paginate log --graph --pretty=format:'%C(auto)%h%Creset%C(auto)%d%Creset %s %C(magenta bold)(%cr)%Creset %C(cyan)<%aN>%Creset'` — see the screenshot further down).

Given that the branches look to be in reasonable places (ie. `1.x-develop` pointing at the `f25b9b145` which is the `HEAD` of `1.x`, and `1.x-stable` very stale and pointing off at `30a616772` which is the last release from over 2 years ago), do a dry-run of the deletion:

```
git push upstream-rw :1.x-develop :1.x-stable --dry-run
To github.com:liferay/clay.git
 - [deleted]             1.x-develop
 - [deleted]             1.x-stable
```

Looks safe, so proceed to do the real thing:

```
git push upstream-rw :1.x-develop :1.x-stable                                                                                                                                                               
To github.com:liferay/clay.git
 - [deleted]             1.x-develop
[...]

PR #3965, opened 2021-03-01
chore(lexicon-ux): Update release tasks

pat270:
- Remove message `and push to gh-pages` from release task we removed it a while ago (this is being done manually at the moment)
- Release task is based only on 1.x now
- ~Release task should push changes to 1.x-stable instead of 1.x-develop, also changed this a while ago~

### DIFF
--- a/tasks/release.js
+++ b/tasks/release.js
@@ -15,8 +15,6 @@ module.exports = function(gulp, plugins, _, config) {
 
 	var TEMP_PATH = path.join(process.cwd(), 'temp');
 
-	var BRANCH_DEV = '1.x-develop';
-
 	var BRANCH_RELEASE = '1.x';
 
 	var getGitRemote = function() {
@@ -120,7 +118,6 @@ module.exports = function(gulp, plugins, _, config) {
 		function(done) {
 			var branchName = 'bower-staging-' + Math.random().toString(16).replace(/[^0-9a-fA-F]/g, '');
 			var browserCommitMessage = 'Browser files for v' + bumpedVersion;
-			var mergeCommitMessage = 'Merging ' + BRANCH_RELEASE + '@v' + bumpedVersion + ' into ' + BRANCH_DEV;
 			var rebuildCommitMessage = 'Rebuild v' + bumpedVersion;
 			var releaseCommitMessage = 'Release v' + bumpedVersion;
 			var tagMessage = 'Version ' + bumpedVersion;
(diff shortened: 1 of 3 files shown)


tasks/release.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(gulp, plugins, _, config) {
 
 	var TEMP_PATH = path.join(process.cwd(), 'temp');
 
-	var BRANCH_DEV = '1.x-develop';
+	var BRANCH_STABLE = '1.x-stable';
```
wincent, line 18: So I probably made a mistake here. I thought we weren't updating `stable` any more as it is kind of pointless (I don't think we've ever cut a release that was from `stable` and not `master` in the last couple of years).

In my [redundant version of this PR](https://github.com/liferay/clay/pull/3967) I removed this, and over on that [other issue](https://github.com/liferay/clay/issues/3958), I actually [_deleted_ the `1.x-stable` branch](https://github.com/liferay/clay/issues/3958#issuecomment-788733603) because I thought it was old/stale (it was pointing at the last release from over two years before). If you want to resurrect it, you can do so with something like:

```
git checkout 1.x-stable # seeing as you probably still have it locally
git merge 1.x --ff-only
git push upstream-rw 1.x-stable --dry-run
git push upstream-rw 1.x-stable
```

If you don't still have `1.x-stable` in your local clone, you can recreate it easily enough:

```
git checkout 1.x
git checkout -b 1.x-stable
git push upstream-rw 1.x-stable --dry-run
git push upstream-rw 1.x-stable
```

tasks/release.js
```diff
@@ -120,7 +120,7 @@ module.exports = function(gulp, plugins, _, config) {
 		function(done) {
 			var branchName = 'bower-staging-' + Math.random().toString(16).replace(/[^0-9a-fA-F]/g, '');
 			var browserCommitMessage = 'Browser files for v' + bumpedVersion;
-			var mergeCommitMessage = 'Merging ' + BRANCH_RELEASE + '@v' + bumpedVersion + ' into ' + BRANCH_DEV;
+			var mergeCommitMessage = 'Merging ' + BRANCH_RELEASE + '@v' + bumpedVersion + ' into ' + BRANCH_STABLE;
```
wincent, line 123: Funning thing is, I don't think this merge commit has _ever_ gotten created. If I look at the history of _every_ commit on every branch (21k lines) with `git log --all --grep 'into 1.x-develop'` I don't find any evidence that we've ever actually done this (ie. zero hits, including when you look at where 1.x-develop used to point before I deleted it, which is 3039054de).

My guess is that the merge fails for some reason and we just silently swallow the error.

tasks/release.js
pat270: Removing it sounds like the better option. I'll take what you did in the redundant pr and add it here.

wincent (approved):
lg2m 🤞

bryceosterhaus:
LGTM! I had no idea 1.x was even in use. Glad to know we have docs for this.

PR #3966, opened 2021-03-01
brainstorm of @clayui/core package

bryceosterhaus:
Just an idea we had talked about in a few threads, potentially combine all of our  packages into a singular release.

Note that I left out the following packages
- @clayui/css (since its css specific)
- @clayui/charts (excluded since we aren't encouraging its use and it significantly increases the size)
- @clayui/shared (not intended to be used externally)

Happy to brainstorm on this more. I created a npm release of `@clayui/corev3.0.0-alpha`. Feel free to test in CSB.

### DIFF
--- a/packages/clay-core/src/index.tsx
+++ b/packages/clay-core/src/index.tsx
@@ -1,0 +1,120 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * Packages Not Included:
+ * - @clayui/charts (Not commonly used)
+ * - @clayui/shared (internal use only)
+ * - @clayui/upper-toolbar (deprecated)
+ */
+
+export {default as ClayAlert} from '@clayui/alert';
+export * from '@clayui/alert';
+
+export {default as ClayAutocomplete} from '@clayui/autocomplete';
+export * from '@clayui/autocomplete';
+
+export {default as ClayBadge} from '@clayui/badge';
+export * from '@clayui/badge';
+
+export {default as ClayBreadcrumb} from '@clayui/breadcrumb';
+export * from '@clayui/breadcrumb';
+
+export {default as ClayButton} from '@clayui/button';
+export * from '@clayui/button';
+
+export {default as ClayCard} from '@clayui/card';
+export * from '@clayui/card';
+
+export {default as ClayColorPicker} from '@clayui/color-picker';
+export * from '@clayui/color-picker';
+
+export {default as ClayDatePicker} from '@clayui/date-picker';
+export * from '@clayui/date-picker';
+
+export {default as ClayDropDown} from '@clayui/drop-down';
+export * from '@clayui/drop-down';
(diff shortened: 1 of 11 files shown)


packages/clay-core/src/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Packages Not Included:
+ * - @clayui/charts (Not commonly used)
+ * - @clayui/shared (internal use only)
+ * - @clayui/upper-toolbar (deprecated)
+ */
+
+export {default as ClayAlert} from '@clayui/alert';
+export * from '@clayui/alert';
```
wincent, line 14: My only doubt here is about all these `export *`...

- Do we want to do this for packages that only have a `default` export? (`@clayui/alert` is one, AFAIK; this line does nothing for those).
- What is going to happen if we have a colliding name? (I gather Babel will silently overwrite with last-one-wins semantics, but that's just a guess).

Obviously writing it like you have is less work, but being explicit would avoid those two doubts, and give us a nice "table of contents" of everything in the API. Could probably write it like:

```js
export {default as ClayAlert} from '@clayui/alert';

export {default as ClayButton, ClayButtonWithIcon} from '@clayui/button';

// etc
```

The obvious cost there is you have to maintain the list.

I notice we have a few different ways of doing exports, and it would be nice if they were all consistent, but I don't think we can change that without breaking things:

- Assigning properties directly to `default` export (eg. [`ClayAlert`](https://github.com/liferay/clay/blob/6a4753c60df96f39265d77d3cee2eb0fc81c9035/packages/clay-alert/src/index.tsx#L185-L188)).
- Using `Object.assign` to add properties to `default` export (eg. [`ClayAutocomplete`](https://github.com/liferay/clay/blob/6a4753c60df96f39265d77d3cee2eb0fc81c9035/packages/clay-autocomplete/src/index.tsx#L79-L84)).
- Using `default` + named exports (eg. [`ClayButton`](https://github.com/liferay/clay/blob/6a4753c60df96f39265d77d3cee2eb0fc81c9035/packages/clay-button/src/index.tsx#L9-L10), although the `default` export ends up using the `Object.assign` trick [in another file](https://github.com/liferay/clay/blob/6a4753c60df96f39265d77d3cee2eb0fc81c9035/packages/clay-button/src/Button.tsx#L89)).
- Using `default` export alone (eg. [`ClayBreadcrumb`](https://github.com/liferay/clay/blob/6a4753c60df96f39265d77d3cee2eb0fc81c9035/packages/clay-breadcrumb/src/index.tsx#L93), which has/composes [other modules](https://github.com/liferay/clay/tree/6a4753c60df96f39265d77d3cee2eb0fc81c9035/packages/clay-breadcrumb/src) but doesn't expose them directly).
- ... not sure if others, but that's just what I see in a quick scan.

jbalsas:
Labelling as `on hold`, I like the approach and it could simplify the process of updating Clay in Liferay... maybe @ambrinchaudhary has some thoughts about this now that she's cut a couple of releases and integrated them into DXP

bryceosterhaus:
Going to close until we decide to go forward with this

PR #3967, opened 2021-03-02
fix(1.x): don't create/update 1.x-develop branch

wincent:
As noted [here](https://github.com/liferay/clay/issues/3958#issuecomment-788321662), we don't want to update 1.x-develop as we phased out that "develop" pattern some time ago. So, this PR removes all references.

Test plan: I guess we'll find out if this works in the next release, if there ever is one... 😱


wincent:
Ha, I hadn't gotten through my whole inbox yet, so didn't see you'd already done this: https://github.com/liferay/clay/pull/3965/files

PR #3991, opened 2021-03-18
feat(@clayui/css): Date Picker adds Date Range styles

### LINKED ISSUE #3986: [Date Picker] Date Range
Hi Clay team,

Due to [LEXI-1127](https://issues.liferay.com/browse/LEXI-1127) team-tango has requested a new feature for the date picker component. 

Date Range

I said feature because I can't say if it's a feature or a variant. We documented it as a variant because it has two parts to introduce the date but except that it's just a change in the behavior of selecting dates. However, you must guide us and decide what this will be. 

Please review the definition added to the original [date picker document](https://docs.google.com/document/d/1iPFtX7Zku7N-sElSfpGSBlfpYvMu4ATCkVVCU_oB3a4/edit?usp=sharing).

ps. We would like to introduce some improvements about input masking BUT we don't want to propose those changes now because team-tango didn't request those changes and we don't want to slow down the development of this request. If you say it would be good to propose it soon please let us know and we will work as soon as possible to introduce new changes in the main component. 

ps2. Besides we would like to renew all the definition doc so we plan to update it on [LEXI-975](https://issues.liferay.com/browse/LEXI-975)

pat270:
TODO: Still need to write documentation for this.

Had to wrap days and dates in `<div class="date-picker-col">` to achieve Lexicon's Date Range pattern. I tried to keep the markup as is, but it was too messy. Good news is the old markup will still work, just no styles for Date Range. We still need to implement the JS for this.

date-range

fixes #3986

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,12 @@ yarn workspace @clayui/css run build && yarn storybook
 yarn test
 ```
 
+8. Update snapshots:
+
+```
+yarn run jest --u
+```
+
 #### clayui.com
 
 To contribute to the documentation and the site in general, you can try to run locally to test your changes:
--- a/clayui.com/src/styles/site/_content-site.scss
+++ b/clayui.com/src/styles/site/_content-site.scss
@@ -118,8 +118,14 @@
 		margin-right: 15px;
 		vertical-align: top;
 
+		&.sheet-example {
+			display: block;
+			margin-right: 0;
+		}
+
 		.dropdown-menu {
 			display: block;
+			float: none;
 			position: static;
 		}
 
--- a/packages/clay-css/src/content/form_elements_date_picker.html
+++ b/packages/clay-css/src/content/form_elements_date_picker.html
@@ -5,6 +5,384 @@ section: Components
 
 <div class="clay-site-row-spacer row">
 	<div class="col-md-12">
+		<h3>Date Picker Date Range</h3>
+
+		<blockquote class="blockquote">
+			<p>The class `c-selected` must be added to `date-picker-col` for all dates in the range.</p>
+			<p>The modifier class `c-selected-start` must be added to `date-picker-col` for the first date in the range.</p>
+			<p>The class `c-selected-end` must be added to `date-picker-col` for the last date in the range.</p>
+		</blockquote>
+
+		<div class="date-picker-dropdown-menu dropdown-menu show" style="position:static;">
+			<div class="date-picker-calendar">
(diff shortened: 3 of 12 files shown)


matuzalemsteles (commented):
Hey @pat270, just to make sure that’s right. Well, I see that we have now with these changes:

- `date-picker-col` let's wrap everything up with it now right? regardless of whether it's a range or not?
- `c-selected` just to be used in the case of the range, right?
- `c-selected-start` start the range
- `c-selected-end` ends the range

The behavior of `c-selected` is different from `active` right or maybe we should just use `c-selected` from now on?

pat270:
@matuzalemsteles yes, wrap everything. `c-selected` behavior is different from `active`.  If we want to use `c-selected` (no active) from now on, I need to add a modifier class to the dropdown menu. My gut says it's better to keep them separate, but no evidence to prove it.

marcoscv-work:
it looks awesome to me! :-D

matuzalemsteles:
@pat270 no problem, well, I think we can keep them separate but it would be nice if we could join the two classes, we will have to deal with two classes in different places to mark the active element depending on whether it is range or not. Maybe it is not a big problem too...

pat270:
@matuzalemsteles Can we try maintaining both `c-selected` and `active` first? It's easy for me to add a rule set that that uses `c-selected` only, but want to keep it backwards compatible for those that might have implemented their own date picker using the CSS (it's in 2.x). Bootstrap also adds their state classes on whatever element is receiving the style changes (e.g., `a`, `button`, `li`). It would be nice to keep it consistent.

If it's too much work to maintain both, I can add this.

matuzalemsteles (approved):
LGTM

marcoscv-work:
Then can we merge this PR? :-D

matuzalemsteles:
@marcoscv-work yeah, I don't know if @pat270 has anything else to add, otherwise, we can go ahead with that.

pat270:
I was saving this one until it is safe to send stuff to Portal master. If it's ok we can merge it.

marcoscv-work:
I think we can fix conflicts and merge!

matuzalemsteles:
@pat270 we can merge that tomorrow after the release to add that to the release on 21st, @diegonvs is already working on the component-side implementation.

pat270:
@matuzalemsteles just merged!

matuzalemsteles:
Thanks @pat270!

PR #3995, opened 2021-03-20
docs(clayui.com): ClayButton adds example of Base Button

### LINKED ISSUE #3993: @clayui/button should be able to output only the `btn` class
The base `btn` class provides the minimum amount of styles to create a normalized button across browsers. We should provide a `displayType='none'` to make it easier to create custom variants. Forcing a displayType based on the variants provided by Clay creates additional overhead when creating custom variants.

pat270:
base-btn

fixes #3993


marcoscv-work:
LGTM

PR #3996, opened 2021-03-22
 fix(@clayui/date-picker): Date Navigation Controls are wrong color

### LINKED ISSUE #3992: @clayui/date-picker: Navigation controls are the wrong color, should be $secondary
Date Picker Navigation Controls (angle left, dot, and angle right) are the wrong color. This is due to having to use `btn-unstyled` if we want to mimic a base button. We really should be able to output only the `btn` class for buttons.

date-picker-nav-controls

pat270:
This adds clay css classes but also leaves `btn-monospaced` and `btn-sm` just incase.

fixes #3992

### DIFF
--- a/packages/clay-date-picker/src/DateNavigation.tsx
+++ b/packages/clay-date-picker/src/DateNavigation.tsx
@@ -126,8 +126,9 @@ const ClayDatePickerDateNavigation: React.FunctionComponent<IProps> = ({
 				<div className="date-picker-nav-controls date-picker-nav-item date-picker-nav-item-expand">
 					<Button
 						aria-label={ariaLabels.buttonPreviousMonth}
+						className="nav-btn nav-btn-monospaced"
 						disabled={disabled}
-						displayType="unstyled"
+						displayType={null}
 						monospaced
 						onClick={handlePreviousMonthClicked}
 						small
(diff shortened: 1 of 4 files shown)


packages/clay-date-picker/src/DateNavigation.tsx
```diff
@@ -126,8 +126,9 @@ const ClayDatePickerDateNavigation: React.FunctionComponent<IProps> = ({
 				<div className="date-picker-nav-controls date-picker-nav-item date-picker-nav-item-expand">
 					<Button
 						aria-label={ariaLabels.buttonPreviousMonth}
+						className="nav-btn nav-btn-monospaced"
```
matuzalemsteles, line 129: Hey @pat270 I noticed some differences, we are adding `nav-btn-monospaced` but we have enabled `monospaced`, at least removing `btn-monospaced` I can not see any differences, I think we can disable this property of the buttons. I also noticed that `btn-sm` causes a small change, it apparently makes the icons a little smaller, would that be intentional?

pat270: It's actually not supposed to have `btn-sm btn-monospaced`. I just left them in because there's a chance some projects might be depending on those classes. If you want them gone, I can remove them.

matuzalemsteles: Hmm makes sense. Anyway, I think it would be worth removing it and checking the Portal if it is used and replacing it otherwise it can stay a long time here waiting for the best day to remove it. Does it make sense to you?

pat270: That's good with me. I removed and pushed changes up.

matuzalemsteles (approved):
LGTM! Just check the lint error before the merge.

marcoscv-work:
LGTM!

Issue #3987, opened 2021-03-16
@clayui/css update mixins to use the `clay-css` pattern


pat270:
The `clay-css` pattern allows more customization through Sass maps without having to declare a variable name for every property. It also must be backward compatible.

- [x] clay-form-control-variant #4002
- [x] clay-select-variant #4046
- [x] clay-label-variant #4070
- [x] clay-loading-animation-variant #4171
- [x] clay-nav-variant #4072
- [x] clay-menubar-vertical-expand #3988
- [x] clay-menubar-vertical-variant #3998
- [x] sheet-footer-btn-block #4171

matuzalemsteles:
Hey @pat270 I was thinking a little about it, once we finished it, we could publish a post on clayui.com to add more information about this change, the benefits, what changes, the advantages, how to migrate, examples... what do you think?

pat270:
@matuzalemsteles This would be good.

matuzalemsteles:
Hey @pat270 since we have that ready and the last one has been merged, it would be good to write a blog post on clayui.com to explain this change and plans to migrate to mixins users, it would also be good to have an idea of how to remove backward compatibility with variables old ones. In that case it would be to get an idea of when we plan to do this, whether it's the next major version or two new major versions from here.

pat270:
@matuzalemsteles Sure I can start working on a blog post. Removing backward compatibility is easy, we just need to delete some setters in the mixins. I suppose we can remove in the next major version and supply the backward compatible mixins to whoever needs it. They can be easily overridden. Hopefully the markup won't change a whole lot, then we will have a problem.

matuzalemsteles:
Perfect @pat270! Depending on the complexity this can involve in a codebase migration from Clay users that use themes, for example, we can keep this for one more major version if necessary, I expect that we won't have too many markup changes in the next major version 😅

PR #4002, opened 2021-03-24
fix(@clayui/css): Mixins `clay-form-control-variant` old keys should win over new keys

pat270:
This updates the way we merge Sass map keys in `clay-form-control-variant` so deprecated keys will win over the new ones. This is for backward compatibility with older versions of Clay CSS.

issue #3987


marcoscv-work:
LGTM! do we need something more to merge it? will we wait for the rest of https://github.com/liferay/clay/issues/3987 tasks or this is not needed it?

PR #4003, opened 2021-03-26
CVE-2018-20677

holatuwol:
## Steps to reproduce

1. Build clay 1.x and run `gulp serve` (or `node node_modules/.bin/gulp serve` if you don't have gulp installed globally)
2. Navigate to the local address. In my case, it was http://localhost:8081/
3. Paste the following into the Javascript console

```
$('body').affix({
  target: '<img src=x onerror=alert(0)>'
})
```

* Expected behavior is that CVE-2018-20677 is addressed and there is no alert.
* Actual behavior is that CVE-2018-20677 has not been addressed and there is an alert.

## Other notes

I don't know how to regenerate bootstrap.js and bootstrap.js.min, so I just manually applied the fixes.


pat270:
@holatuwol How do you feel about upgrading Bootstrap to 3.4.1 and its affect on Portal? It looks like this fix is contained there among other things.

holatuwol:
I'm comfortable with a solution that upgrades it to 3.4.1, since the release notes indicate only additions and not removals (so it should be backwards compatible). I have no idea how to add a completely new release of Bootstrap to Clay, though.

pat270:
I've compiled the release notes below. The ones in bold are the changes I think could affect themes. It could break some, but might be worth it since there are other XSS fixes bundled. The CSS changes seem pretty minor though.

I have no problem updating Bootstrap myself FYI.

3.4.1
- **Security: Fixed an XSS vulnerability (CVE-2019-8331) in our tooltip and popover plugins by implementing a new HTML sanitizer**
- Handle bad selectors (#) in data-target for Dropdowns
- Clarified tooltip selector documentation
- Added support for NuGet contentFiles

3.4.0
- **New: Added a .row-no-gutters class.**
- **Fixed: Resolved an XSS issue in Alert, Carousel, Collapse, Dropdown, Modal, and Tab components. See https://snyk.io/vuln/npm:bootstrap:20160627 for details.**
- **Fixed: Added padding to .navbar-fixed-\* on modal open**
- **Fixed: Removed the double border on <abbr> elements.**
- New: Added docs searching via Algolia.
- Removed Gist creation in web-based Customizer since anonymous gists were disabled long ago by GitHub.
- Removed drag and drop support from Customizer since it didn't work anymore.
- Added a dropdown to the docs nav for newer and previous versions.
- Update the docs to use a new baseurl, /docs/3.4/, to version the v3.x documentation like we do with v4.
- Reorganized the v3 docs CSS to use Less.
- Switched to BrowserStack for tests.
- Updated links to always use https and fix broken URLs.
- Replaced ZeroClipboard with clipboard.js

v3.3.7
JavaScript
- **#19192 Fix keyboard navigation for toggle buttons (checkbox, radio, single toggle) following a mouse click**
- **#19659: Clear tooltip's $element to prevent leaking memory**
- **#20019: Avoid calling jQuery('#') since it's a syntax error in jQuery 3**
- #20259: Backport jQuery-related JS unit test fixes from v4
- **#20278: button.js: Set disabled property in addition to disabled attribute, for jQuery 3 compatibility**
- **#20313: Avoid using $.offset() on SVGs since it gives incorrect results in jQuery 3**
- #20338: Update jQuery version check and jQuery dependency version range to allow jQuery 3

CSS
- No significant changes.

Accessibility
- #19704: Add aria-label and callout about labeling pagination components

v3.3.6
CSS
- **#16492: Use @input-border-radius variables on input groups.**
- **#16529: Un-style the caret on `<select>`s in IE10+.**
- **#16562, #16563: Create and implement a new @dl-horizontal-breakpoint variable to replace the @grid-float-breakpoint on .dl-horizontal.**
- **#16646, #16651: Fix hover border covering active links in pagination.**
- **#16684: Fix border-radius not being set on two corners of vertical btn group.**
- **#16718: Full 4-digit Unicode escape sequences in CSS for asterisk and plus Glyphicons.**
- **#16769: Remove the min-height on modal headers and replace with a clearfix to avoid collapsing when no title exists.**
- **#16779: Adjust the padding on jumbotrons within containers.**
- **#16785: Round the padding on containers to match the new rounding of negative margins on rows.**
- **#16796: Ensure carousel controls are clickable in IE9.**
- **#16804: Indicate which buttons in a button group are selected, even when disabled.**
- **#16824: Correct a math error in the padding-top of .control-labels in large and small horizontal forms.**
- **#16975: Use @carousel-control-font-size for in specific pixel values.**
- **#17001: Fix borders on focused inputs in input groups.**

JS
- **#16865: Fix triggering of shown.bs.dropdown & hidden.bs.dropdown events so that relatedTarget property gets set properly.**
- #16896: Note jQuery v3 incompatibility.

holatuwol:
I checked with Supportability, and Joel indicated that anything that might break themes isn't handled in a special way, so I'm okay with upgrading Bootstrap.

pat270:
Sounds good to me, closing this in favor of #4005.

PR #4007, opened 2021-04-05
fix(@clayui/date-picker): disable focus control using the arrow keys

### LINKED ISSUE #3981: [Date Picker] Unexpected behavior using keyboard arrows
Reviewing the current behavior of the [date picker](https://clayui.com/docs/components/date-picker.html) component I noticed that the calendar behaves in a different way I expected:

focus-jump-using-keyboard

**Steps to Reproduce**
1. Open the calendar
2. Click on a date
3. Click on the input (to change manually the date)
4. Press the right arrow ➡️  
5. Click on the input (to change manually the date)
6. Press the right arrow ➡️ 
7. ~Delete the date~

**Expected behavior**

1. Pressing ⬅️ ➡️ moves left/right one char position the cursor
 
2. Besides, when the cursor is placed:

AT THE START of the string:

- Pressing ⬆️ or ⬅️: nothing
- Pressing ⬇️ : move the cursor to the end of the string

AT THE END of the string:

- Pressing ⬆️ : move the cursor to the start of the string
- Pressing ⬇️ or ➡️ : nothing

cc\ @marcoscv-work

matuzalemsteles:
Fixes #3981

We are disabling focus control using the arrow keys because it causes inappropriate behavior in the Date Picker, when there is input instead of going through the strings it will jump to the next element. We have no way of identifying whether the user skips to the next element or continues to go through the input string.

The best case for this would be to deal with a better focus control only within the Date Picker, disregarding the input.

This change still allows focus control but only using `tab` and `shift + tab`.

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -372,7 +372,7 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 		}, [handleFocus]);
 
 		return (
-			<FocusScope arrowKeysLeftRight>
+			<FocusScope>
 				<div className="date-picker">
 					<ClayInput.Group id={id} ref={triggerElementRef}>
 						<ClayInput.GroupItem>


matuzalemsteles:
Merging this so that we can add to tomorrow's release!

wincent:
> Merging this so that we can add to tomorrow's release!

i-ll-allow-it

PR #4009, opened 2021-04-06
feat(@clayui/css): Clay Admin

pat270:
This is my initial attempt to create a Clay Admin Theme. The CSS gets compiled to `cadmin.css` that we can load inside Portal when a user has admin permissions. In my Liferay Portal branch, I'm loading it inside the Product Menu theme contributor with `@import 'cadmin';`. I'm also loading variables similar to `atlas-variables` with `@import 'cadmin-variables';`.

### How It Works

The styles are scoped by abusing the `:not()` selector. It increases the specificity of the selector based on the type of selector used. For example, if we use `:not(#id)`, it will increase our selector by an `id`. If we use a class, it will increase it by a class. We can also chain negation selectors to increase it by any number of `id`'s or `class`es. The default specificity is 2 id's, 1 class, and an element.

```
$cadmin-selector: ':not(#__):not(#___)';

html#{$cadmin-selector} .cadmin {}
```

I'm using the `cadmin` class as the marker to designate where scoping should begin. Generally we want to wrap whatever component we want to scope in `<div class="cadmin"></div>`. There are some exceptions like with `dropdown` or `modal` where we can put `cadmin` directly on the component element. This was done because it required work to wrap it.

#### How Component Teams Should Use It

Clay Admin plugs in exactly like Clay CSS. Component teams that want to scope their CSS have to import and wrap their custom code with:

```
@import 'cadmin-variables';

html#{$cadmin-selector} .cadmin {
  .custom-btn {}
}
```

If they want themeable code (not scoped):

```
@import 'atlas-variables';
@import 'cadmin-variables';

html#{$cadmin-selector} .cadmin {
  // use Clay Admin stuff here
}

.themeable-component {
  // use Clay CSS stuff here
}
```

#### Why Sass Placeholders

All the CSS are inside placeholders so that we can still scope hard to change components in Portal.

```
html#{$cadmin-selector} .cadmin {
  .hard-to-change-class-that-is-deep-inside-a-taglib {
    @extend %cadmin-btn !optional;
  }
}
```

### Conclusion

I think this will be good for our product. It opens up the Bootstrap namespace so devs can directly style Bootstrap classes without worrying about admin controls. It is the same for Clay CSS variables. Devs can choose to change values and not worry about it affecting admin.


jbalsas:
@wincent, @marcoscv-work, @edalgrin, @p2kmgcl, @matuzalemsteles, could you guys weigh in on this one and [Create a Shadow component to prevent fragment's CSS to propagate to admin UI](https://github.com/p2kmgcl/liferay-portal/pull/729) to see if we can agree on what direction we'd like to take?

pat270:
cadmin-flow-chart
Here is a flow chart to help illustrate how Clay Admin is integrated in Portal. The Clay Admin CSS file is included at the top of Product Menu Theme Contributor. Ideally, I would like this to be included near `atlas.css`, but I'm trying to use what's already available. There will be some specificity issues with scoped module CSS and Clay Admin if we load it this way. If we load Clay Admin at the top, it shouldn't be a problem.

Clay Admin CSS is only loaded once via Product Menu. The file `cadmin-variables` includes functions, variables, mixins, and placeholders similar to `atlas-variables`. No CSS is output from `cadmin-variables` unless you explicitly use it.

wincent:
> @wincent, @marcoscv-work, @edalgrin, @p2kmgcl, @matuzalemsteles, could you guys weigh in on this one and [Create a Shadow component to prevent fragment's CSS to propagate to admin UI](https://github.com/p2kmgcl/liferay-portal/pull/729) to see if we can agree on what direction we'd like to take?

This is a very clever approach, but it is difficult for a mere mortal like me to understand, so I'd really appreciate it if we could get some input from the actual CSS experts mentioned above. 😀 

Speaking from my position as a CSS dabbler, I worry about the complexity and maintainability of this approach (ie. will we have multiple people capable of understanding and modifying this?). I also feel a bit nervous that the way are imposing the separation styles is by entering into a specificity arms race, a race which by definition is losable because somebody can always come along and trump the other side with a higher specificity; that is to say, we don't have "real" isolation here, maybe just the appearance of it — would be interested to hear your thoughts on that @pat270. Another thing that may be of concern is the performance of the more complex selectors. I know browsers are faster now, but in the old days we used to be wary of the longer selectors and "special" ones like `:not`. Might be worth trying to measure the impact (although admittedly, at least the impact should be restricted to logged-in admin users).

In contrast, the Shadow DOM approach feels like a more natural fit for the task at hand — it really does deliver "isolation", as a matter of design as opposed to clever [CSS tricks](https://css-tricks.com/) 😂 — although as noted on @p2kmgcl's PR, it is not all roses and rainbows. For example, the "isolation", while effective, is not actually intended to apply in 100% of cases (as [explained here](https://open-wc.org/guides/knowledge/styling/styles-piercing-shadow-dom/)). My gut instinct however, is that the Shadow DOM is still going to be the more maintainable solution in the long run. But as I said earlier, I am not a CSS expert, so it's hard to know if this is my hipster JS coder bias manifesting itself unconciously.

edalgrin:
🤯 This is very complex but I think I grasp the concept and here's some personal opinions/doubts 

> :not(#__):not(#___)

this is really _really_ ugly but I guess it's necessary 😞

> I worry about the complexity and maintainability of this approach

I share this concern with Greg, one of the point we are discussing in a different-yet-related epic is to simplify the architecture of the themes, with this method we are adding it back in another context

---

Using the proposed approach we are duplicating a lot of code and update a component in clay-css might required a change in at least 2 places which is a not-recommended-practice in modular stuff – a solution for this issue could be making clay-admin-variables a dependency of (clay-)atlas-variables, e.g.:
```scss
$h1: $cadmin-h1;
```

---

Having said that, at the moment I still prefer this solution to the shadow-dom one, because it's easier to apply and we are able to anticipate most of the issues with the _bleeding styles_

pat270:
@wincent It's complex and annoying scrolling through the styles panel in dev tools, but you get used to the computed styles tab quick :laughing:.

> complexity and maintainability of this approach (ie. will we have multiple people capable of understanding and modifying this?)

I'm not sure if you're talking about the Clay repo or the implementation in Portal. For the Portal implementation they just need to know:

```
@import 'cadmin-variables';

html#{$cadmin-selector} {
  .cadmin {
    .full-admin-component-no-outside-styles {}
  }

  .mixed-content-admin-styles {}
}
```

There might be some edge cases where they need to prevent inheritance with `@extend %cadmin-reset !optional;`.

> we don't have "real" isolation here

Yes this is just a hack. There will always be a way to get around it. This is essentially the same trick that [styled-components](https://styled-components.com/docs/basics) uses. They generate unique class names that have low probability of being duplicated. It can still be overwritten by a selector with a specificity of 2 classes. They also have a setting for increasing specificity.

Everyone hates the global nature of CSS, but I think it's one of its greatest strengths. It allows me to take any page and completely transform it without having to touch the underlying code.

> imposing the separation styles is by entering into a specificity arms race

I'm not sure if this will turn into a specificity arms race since I get the feeling no one wants to touch our admin code. As long as we write our selectors properly, I don't foresee complaints like we had in 6.1 / 6.2 with `aui` on the `html` element :crossed_fingers:.

> performance of the more complex selectors

I have no data on this, but my bet is on crazy CSS selectors being more performant over shadow DOM. I believe the general problem is with selectors like `header > div > div`. Browsers read selectors from right to left. It would traverse the document and grab all the `div` elements then check to see if it was a direct descendant of `div` then check to see if it's a direct descendant of `header`. Like you said browsers are more efficient and computers have more power, but I wouldn't mind doing some performance tests (not sure how though).

> this is really really ugly

@edalgrin Don't worry it gets uglier. It will only be visible in mixed content though (e.g., fragment editor stuff). Client markup shouldn't show these selectors in dev tools.

> making clay-admin-variables a dependency of (clay-)atlas-variables

We can't make Clay Admin a dependency of Clay Atlas. This is literally the reason why we have this problem. We need to treat them as two separate frameworks. Clay Atlas is for the client and Clay Admin is for us. We should duplicate code for now until we figure out where the separation should be. Once we figure that out we can start removing bloat from Clay Atlas.

I'm firm believer that we will use both. Shadow DOM has its limitations with mixed content (page-editor) and will probably require complete rewrites. Specificity hack can plug into existing code with less invasive changes.

pat270:
cadmin-scoping
Here's another diagram that might clarify what's going on. The boxes in orange are the isolated areas. The `cadmin` container blocks inherited styles like `font-family`, `font-size`, `color` from passing through.

The containers `page-editor__topper` is unique because it's an admin element with user facing content inside. We just scope the admin styles to `html:not(#__):not(#___)` and don't use the `cadmin` container because we still want inherited properties to pass through to the user facing content.

Components that can use the `cadmin` container are good candidates for Shadow DOM while mixed components are a bit tricker. This is where the specificity hack can make our lives easier.

wincent:
> I'm not sure if you're talking about the Clay repo or the implementation in Portal.

I was specifically thinking of the implementation in the Clay repo. As you say, the liferay-portal side doesn't look bad at all.

> but I wouldn't mind doing some performance tests (not sure how though).

Yeah, it's tricky because we don't have a full implementation of either proposal at this time to be able to do an A/B comparison (or even a comparison of "with isolation version A" vs "without isolation"); and even then, unless you run a lot of tests in an extremely controlled environment, you might not be able to pick up a reliable signal of a difference. But it is still probably worth at least having a dig around in the dev tools profiler to see if there is any obvious insight to be gleaned (or see if Lighthouse reports potential problems).

marcoscv-work:

marcoscv-work (approved):
I prefer this solution instead of shadow, why? because it is more simple to understand to me.
So I wonder if there is a need to make it simple to understand or not, I mean, any customer will need to take a look into this? or do we prefer to make it invisible or imperceptible?

About the PR, I've been testing and the result looks pretty good. just some questions.
- Have we tested it in portal? can we do it in an easy way?
- Do we cover all admin controls portal needs?
- I've tested the CSS site, but just to not forgetting something..  do we have any indications to test correctly the PR? :-D

`cadmin.css` weight is 119KB, it looks much less than the 609KB from `atlas.css`

My total approval to keep going in this direction and help if I can!

matuzalemsteles:
Well, coming up with that now. Just adding a few cents to the conversation but I tend to go with that implementation compared to using the Shadow DOM.

- The downside of this proposal is that we will have to duplicate code if that's what I understood, then we will have to send more code to the client, would there be any way to try to reuse it or does this imply due to separating things by namespace?
- Having duplicate code can increase the degree of maintenance since we have to be extra careful when making changes so that things also reflect for the admin.

Just thinking a little bit about it, but at least when I look at CSS isolation I see a lot more strategies being used with CSS-in-JS (Be it solutions centered on Runtime or build) since you can control it at build time to generate random classes and achieve maximum reuse. Just thinking if we can explore something similar to this on our side, not implementing a CSS-in-JS library but some tool to help us with that, maybe duplicating the code but generating new random classes or creating the namespace, well that it can of course imply a lot of other problems that was not visible but just launching it in case anyone has any more ideas.

jbalsas:
> The downside of this proposal is that we will have to duplicate code if that's what I understood, then we will have to send more code to the client, would there be any way to try to reuse it or does this imply due to separating things by namespace?

Yeah... we should find a way to trim some fat before adding this in or else we're just doing exactly the opposite of what we've been saying we're doing.

edalgrin:
> /cc @edalgrin, @marcoscv-work, is there anywhere in our plans to improve CSS/Theme performance room for this?

Aren't we already working on this https://github.com/liferay-frontend/liferay-portal/pull/961 ?

wincent:
> I prefer this solution instead of shadow, why? because it is more simple to understand to me.
> So I wonder if there is a need to make it simple to understand or not, I mean, any customer will need to take a look into this?

I wasn't thinking of customers when I said this:

> I worry about the complexity and maintainability of this approach (ie. will we have multiple people capable of understanding and modifying this?).

My main concern is that _we_ can understand and modify it. Not just understand it "intellectually", but be capable of working on it in the practical sense — making changes, fixing bugs, updating... @pat270 obviously can, and you say that you can too, so that's good. I just don't want to find ourselves in a situation where we have a bus factor of 1 (ie. only one person can effectively work and iterate on this). Ideally bus factor would be higher still, but we have to work within the constraints that exist...

Now, at some point, a customer _is_ going to ask how to customize admin styles in this new world. It may be difficult, it must be possible, but it doesn't have to be easy. I really think we have to optimize of _us_ first this time. The solution has to be robust, it has to be something we can evolve and modify over time, and it it needs to do all that with adequate performance.

edalgrin:
> My main concern is that _we_ can understand and modify it. Not just understand it "intellectually", but be capable of working on it in the practical sense — making changes, fixing bugs, updating... @pat270 obviously can, and you say that you can too, so that's good. I just don't want to find ourselves in a situation where we have a bus factor of 1 (ie. only one person can effectively work and iterate on this). Ideally bus factor would be higher still, but we have to work within the constraints that exist...
> 
> Now, at some point, a customer _is_ going to ask how to customize admin styles in this new world. It may be difficult, it must be possible, but it doesn't have to be easy. I really think we have to optimize of _us_ first this time. The solution has to be robust, it has to be something we can evolve and modify over time, and it it needs to do all that with adequate performance.

I think it will be the same as the current clay-css customization method, since this new `cadmin` is a reduced duplication of it. Now, to improve/fix/extend the system, it will be the same as improve/fix/extend clay-css, the real difficulty could be creating the initial namespace-thing (which I think Patrick already did) and maintaining clay-css and clay-admin-css in sync (not a requirement)

pat270:
I ended up scrapping my original idea and duplicated Clay CSS in the `cadmin` directory. All variables are prefixed with `$cadmin-`, styles are scoped, and values in `px`. This should cover everything Clay CSS covers. I still need to update my Portal draft to accept this new version.

pat270:
Just pushed up another iteration of this. I cleaned up the CSS isolation a bit.  Instead of resetting styles based on Clay Components, I targeted the HTML elements instead. This cleans up the selectors a lot in dev tools. This covers everything we have, meaning we can isolate anything we want. Most of the work now will be updating the markup / css inside Portal.

jbalsas:
> I ended up scrapping my original idea and duplicated Clay CSS in the cadmin directory

Hey @pat270, not sure if I'm understanding this properly, but with this we somehow have everything clay twice? So, if we ever have a bug in one component, we'll need to fix it in at least 2 different places?

On top of that... what would be the impact on a page that already has `atlas` if it also needs `cadmin`? Does this mean we'd be doubling the amount of CSS we deliver?

pat270:
@jbalsas, yes we are loading two. The second one (cadmin) only shows up when `product-menu-theme-contributor` is on the page. We will have to fix it in two places, unfortunately, but we should consider them two separate themes. The only shared pieces between the two are Sass functions and mixins.

Any module that is imports `atlas-variables` and `cadmin-variables` should have no additional weight since those files don't output any CSS. The only issue right now is license text is output twice.

This is the first step. I didn't want to start removing stuff from Clay CSS until we had a working demo and this gets approved. If this is something we want, we can start disabling admin only components from Clay CSS (e.g., code, drilldown, range, reorder, clay-color, time, date-picker, dual-listbox, multi-step-nav, menubar, navigation-bar).

In Cadmin we can remove all Bootstrap components we aren't using.

We can further trim down Clay CSS by moving component CSS into the module that uses them, like forms. Multi Step Nav and Timelines come to mind. We can move management-bar, color-picker, date-picker, dual-listbox into the taglib so they are only loaded when needed.

pat270:
This one is outside the scope of this pull, but I'd like to get rid of `frontend-css-common` and `frontend-css-web`. There is a lot of deprecated CSS and the CSS that we do use can probably be moved into the modules.

marcoscv-work:
All this work is pretty important to understand that finally, we are duplicating Clay to keep admin controls design, but the full version of Clay is loaded in the site.. when Clay actually was created to fill the gaps between bootstrap and liferay-administration-components.. so what if we try to invert it?.
Loading full clay for what was created for: admin, and try to create a reduction of clay for sites, and step by step cleaning the no-needed components are being loaded in the site, in this direction, we finally would have a reduction of clay for sites, or why not, only bootstrap, or even only the components we essentially use in the page editor for the site: helpers and the grid

edalgrin:
> This one is outside the scope of this pull, but I'd like to get rid of `frontend-css-common` and `frontend-css-web`. There is a lot of deprecated CSS and the CSS that we do use can probably be moved into the modules.

I'm taking a look at that https://docs.google.com/document/d/1RJTd3KRHvWFG07qkRJvG0rMnMSjhWXZeNAJNDaEJhYg/edit#

marcoscv-work:
> > This one is outside the scope of this pull, but I'd like to get rid of `frontend-css-common` and `frontend-css-web`. There is a lot of deprecated CSS and the CSS that we do use can probably be moved into the modules.
> 
> I'm taking a look at that https://docs.google.com/document/d/1RJTd3KRHvWFG07qkRJvG0rMnMSjhWXZeNAJNDaEJhYg/edit#

Yeah we already working on that and any help is welcomed :-P

pat270:
This PR is ready for review. Some stats:
```
atlas.css: 611kb
atlas.css.gz: 72kb
base.css: 564kb
base.css.gz: 67kb
cadmin.css: 829kb
cadmin.css.gz: 72kb
```
The scoping selector adds a lot of weight to the uncompressed version. The good news is that it compresses well since it's repeated for every selector. The compressed size is the same as `atlas.css`. We could possibly lower the size by pruning our `utilities` and `navbar` breakpoints. We only really use `md` no need for `xs`, `sm`, `lg`, `xl`. This would apply to `navbar`, `navigation-bar`, `management-bar`. We can also remove more Bootstrap components from `form`, `form-validation`, `list-group`, `tables`.

This can plug into Portal today as is, since nothing is relying on it. We can convert admin components by module at our own pace.

edalgrin:
Just started reviewing :)

edalgrin:
I going to be honest, I'm scared of this solution. Creating a whole new Clay just for admin will probably generate other issues, this is not a "just a part of clay" anymore, it's the whole library 😬 and maintaining both libraries could become complicated especially if we decide to follow different development paths, we might have performance issues as well (which we are trying to remove in another epic) and if we are planning to use custom properties for this separation in the future, the proposed solution is not helping to achieve that

I think @marcoscv-work were working on the same objective from a different perspective: "create clay-front-office" where we use the current clay-css for admin and a new smaller library for the "front-office" (at least that was the premise? 🤔 )

So my proposal is to keep this PR on hold until we are sure this is the best solution we can offer now.

marcoscv-work:
> I going to be honest, I'm scared of this solution. Creating a whole new Clay just for admin will probably generate other issues, this is not a "just a part of clay" anymore, it's the whole library 😬 and maintaining both libraries could become complicated especially if we decide to follow different development paths, we might have performance issues as well (which we are trying to remove in another epic) and if we are planning to use custom properties for this separation in the future, the proposed solution is not helping to achieve that
> 
> I think @marcoscv-work were working on the same objective from a different perspective: "create clay-front-office" where we use the current clay-css for admin and a new smaller library for the "front-office" (at least that was the premise? 🤔 )
> 
> So my proposal is to keep this PR on hold until we are sure this is the best solution we can offer now.

The proposal is not to create another workflow, just using the current one and not creating a super complex version of clay in parallel.

For example, moving modules like grid or helpers to a general load and then move CSS components to modules that use them, but all of them coming directly from Clay.

If we want to update something, we will update clay as always.

If you will use something super complex like documents and media on your site you need to assume you will load that extra CSS but is not a good idea to load super complex components CSS just in case a non-usual widget appears on the screen

Please remember this is not the usual case, probably is the most complex one

marcoscv-work:
I have a similar feeling, I'm a little bit scared of this solution since we create a version of clay for admin in a really complex way and at least for me, difficult to understand.

I think this is done thinking of optimizing the size of the new cadmin file, but do we really think to optimize something for admin? overall when the cost is complexity?

On the other hand, I think Clay is for admin, and we don't need to create a new version of clay for admin, probably we could work on isolate styles from clay witch we do really need for sites (since sites are what we need to optimize).

pat270:
> where we use the current clay-css for admin

This pr is exactly that. It's converting Clay CSS to admin only. Now, Clay CSS can become anything we want it to be. Clay CSS => clay-front-office, sure. Just remove everything except `cadmin` and write what you want.

> maintaining both libraries could become complicated especially if we decide to follow different development paths

clay-front-office is the same. We will be maintaining two libraries.

> I think this is done thinking of optimizing the size of the new cadmin file, but do we really think to optimize something for admin? overall when the cost is complexity?

We optimize both. We duplicate then remove what we don't need from Clay CSS and Cadmin. Duplicating lets us optimize Clay CSS to 0kb.

matuzalemsteles:
Hey @pat270 I think I have the same feeling that this can be a little complex maybe due to the duplication of code that we have and because we will have to "load two versions of Clay" if that's what I understand.

Just thinking a little about it, as both will always be the same, to avoid manual code duplication, because we didn’t explore the possibility of doing this in an automated way, I see that we have a lot of duplication of the components but the only different thing I was able to see in my superficial review are the `$cadmin` variables.

We can do this at build time, perhaps, to generate the assets for `cadmin.css` and `atlas.css` from the same source but for `cadmin` just by changing the variables and creating if the values are the same in both places. I don't know if there is any more restriction. For some configurations that are specific, we can keep in the respective file and we can think of something like "merge", to allow overrides.

pat270:
@matuzalemsteles It's definitely possible to automate this. I don't think it will be worth it. The major issue is in a year, Clay CSS and Cadmin shouldn't look anything like each other. Cadmin gives us the freedom to remove a lot of the CSS from Clay. It really doesn't make sense to keep both in sync since they are serving two purposes. Clay CSS should be for the user and Cadmin for us.

I'm envisioning Clay CSS to provide a CSS reset, buttons, forms, grid, utilities or some subset of components provided by Clay. Everything else would be gone, left up to the user to decide if they need it.

pat270:
I went ahead and removed some components from Clay CSS as a test (not committed in the pr):
```
@import 'components/_carousel';
@import 'components/_jumbotron';
@import 'components/_spinners';
@import 'components/_toasts';
@import 'components/_media';
@import 'components/_drilldown';
@import 'components/_range';
@import 'components/_reorder';
@import 'components/_clay-color';
@import 'components/_time';
@import 'components/_date-picker';
@import 'components/_dual-listbox';
@import 'components/_form-validation';
@import 'components/_multi-step-nav';
@import 'components/_menubar';
@import 'components/_application-bar';
@import 'components/_navigation-bar';
@import 'components/_panels';
@import 'components/_popovers';
@import 'components/_progress-bars';
@import 'components/_quick-action';
@import 'components/_slideout';
@import 'components/_tbar';
@import 'components/_timelines';
@import 'components/_toggle-switch';
@import 'components/_tooltip';
```
This brought `atlas.css` down 171kb to 447kb (55kb gzip). The screenshot below shows what a Widget Page looks like with this deployed.

widget-page-reduce-atlas

There are more places we can optimize in Clay CSS. We could probably bring it down to the 300k range. For a page that doesn't use OOTB widgets, we could bring this really low.

wincent:
As this thread has gone on, it shows exactly why [I expressed concern above](https://github.com/liferay/clay/pull/4009#issuecomment-814763466) about the "bus factor" involved:

> Speaking from my position as a CSS dabbler, I worry about the complexity and maintainability of this approach (ie. will we have multiple people capable of understanding and modifying this?)

In the recent comments, we have:

- @edalgrin [saying](https://github.com/liferay/clay/pull/4009#issuecomment-828324112) "I going to be honest, I'm scared of this solution."
- @marcoscv-work [says](https://github.com/liferay/clay/pull/4009#issuecomment-828409172) "I have a similar feeling, I'm a little bit scared of this solution since we create a version of clay for admin in a really complex way and at least for me, difficult to understand."
- @matuzalemsteles [adds](https://github.com/liferay/clay/pull/4009#issuecomment-828666403) "I think I have the same feeling that this can be a little complex"

And illustrating the difficult of grokking this, @pat270 [informs us](https://github.com/liferay/clay/pull/4009#issuecomment-828595519) that there is actually nothing to worry about: "This pr is exactly that." (ie. the intent and functioning of the PR is _not_ clear, _even to CSS experts_) etc.

So, I just want to raise this as a red flag... If we're going to do this, _how_ are we going to make sure that people other than Patrick can understand and maintain it. He's going to have to take a holiday every now and then, you know...

Is the answer documentation? Should somebody else be sharing the work in this PR with Patrick to prove that we have somebody who can meaningfully collaborate on it? Other ideas?

marcoscv-work:
Anyway, I think all these works are a good example of what we need and what we have and make it much more clear the way to optimize.

I really like the Patrick example cleaning components with `atlas.css` down 171kb to 447kb (55kb gzip). Maybe we can make those "import" optional, and little by little continuing removing things until we only have what we really and essentially need.

I think will not have problems if we make it clear and well documented since we broke with bootstrap everything we decide now for admin controls will be fine and should be aligned with our needs (I think is important the simplicity).

On the other hand, all that we decide to do with our admin framework shouldn't affect customers, for example, if we don't have high demands for our grid, it's fine if we keep using flex for years, but we should offer a way to use new things (like css grid). in our customer sites.
In this way, we could be much more disruptive or less demanding with Clay changes, that's the reason I'm more and more sure that we should split clay admin controls and site styles, it would resolve weight problems in sites and admin controls in editors at the same time.

Of course, probably we cannot do it in a perfect way right now, but we could follow some simple steps to get it:

- List all widgets with admin controls
- Create Clay admin (encapsulating styles and removing no needed components).
- Adding the 'cadmin' to all admin controls in control-panel and widgets
- Load Clay admin when:
  1) Some widget with admin controls is on the screen (we could create ADT's without those admin controls, avoiding load Clay admin)
  2) We are administrators
- Provide examples of:
  1) Site with Clay for sites (full version, probably Classic theme)
   2) Site without Clay (light version and full configurable)

jbalsas:
> Load Clay admin when:
> - Some widget with admin controls is on the screen (we could create ADT's without those admin controls, avoiding load Clay admin)
> - We are administrators

I think this could be a rather simple change... we could touch **ALL ootb widgets** so that they include `atlas.css` (or however we call it) and remove it by default from the theme leaving only whatever's necessary in terms of general site layout and appearance. That would mean that we don't push our styles by default, but any ootb widget will still function properly (at the expense of extra and maybe duplicated css).

pat270:
@wincent
dead-to-me

Contributing here and in Portal is the best way. I tried documenting the best I could in the PR.

marcoscv-work:
> > Load Clay admin when:
> > 
> > * Some widget with admin controls is on the screen (we could create ADT's without those admin controls, avoiding load Clay admin)
> > * We are administrators
> 
> I think this could be a rather simple change... we could touch **ALL ootb widgets** so that they include `atlas.css` (or however we call it) and remove it by default from the theme leaving only whatever's necessary in terms of general site layout and appearance. That would mean that we don't push our styles by default, but any ootb widget will still function properly (at the expense of extra and maybe duplicated css).

Sounds perfect for me, we could even adapt each widget with components they are using really quickly. Example with documents and media here:

https://github.com/marcoscv-work/clay/commits/poc-admin-css

matuzalemsteles:
So, I think I'm getting a better understanding of this, but I see that we already want to create a distance of Admin vs Clay for customers, it is still unclear to me what we would remove from Admin or Clay to reduce the size, but if we continue with this idea makes sense to keep a `cadmin`.

To make it clearer, what would be admin in DXP? Do product portlets also enter as part of `cadmin`?

pat270:
Only specific parts of a Product Portlet would be part of `cadmin`. I'll take Forms Portlet as an example.

forms-widget

The Alert and Select Form link would be Admin. When you click Select Form, the Modal that pops up would be Admin.

It's up to the component teams to decide what parts of their app should be admin only. The can scope them by wrapping what they want isolated in `<div class="cadmin">`.

marcoscv-work:
Everything would be wrapped by `.cadmin` in order to save some KB.. does make sense to reduce selectors names? for example `cadmin` to `cad` or something like that

---

Continuing with the Patrick example for forms and trying to be clear and summarize the conversation, I see two options:

- In some way.. Portlets with uses of 'cadmin' communicate to the portal we need to serve `cadmin.css`.
- Portlets have their own version of `cadmin` in this example, probably only with components Froms needs: lists, froms, etc..
   - For this case, we would need a common configuration in order to be read for all portlets `cadmins`.

matuzalemsteles:
@pat270 I think it got a little clearer, but one thing that is not clear to me or not visible yet is when you say that we can remove more things from Clay for customers, which we would necessarily be removing what goes on in their minds 😁.

What I imagine is, maybe components that were made for the admin part of DXP but removing this would not be very unsafe since it has always been available to customers, and removing it could break some of them that they may be using.

Once we do that too, we have to worry about the documentation, in theory, we add all the components or the vast majority of it to clayui.com, we want to make it documentation for customers or just for internal, somehow we would have to document it `cadmin` things.

pat270:
@matuzalemsteles You make a good point. We don't need to remove/disable it from the Clay Repo. It can be done at the Portal level. We could include a warning (on clayui.com) about it being disabled in Portal and instructions on how to enable it.

My gut says no one outside is using our specialized CSS components and wouldn't cause too much disruption if we provide enough warning and time for those that do to use the Cadmin version. They wouldn't need to include the whole `cadmin.css` file. They would import whatever pieces they need into their own CSS files.

pat270:
> does make sense to reduce selectors names?

@marcoscv-work  It doesn't matter to me, we can pick whatever name we like. Since I assume things will be minified and gzipped in production a long selector name doesn't impact as much. Gzip is good and compressing repeated selectors (cadmin.gz is the same size as atlas.gz).

If it's not gzipped, then a shorter selector name would reduce KBs. Changing `.cadmin` to `.cad` reduces to 776kb and changing `:not(#__):not(#___)` to `:not(#_):not(#_)` reduces it more to 751kb. A total reduction of 51kb (802 - 751).

marcoscv-work:

marcoscv-work:
Reduce names and selector looks good to me:

Something that surprises me is the weight of `cadmin.css`. even seeing all the unnecessary components you already removed or optimized.

pat270:
@marcoscv-work the scoping selector is adding the extra size. I fear trying to reduce the size by piecing components out will do more harm than good.

marcoscv-work:
> @marcoscv-work the scoping selector is adding the extra size. I fear trying to reduce the size by piecing components out will do more harm than good.

So then we keep `cadmin`.

I'm testing the last changes, everything looks much better so I think we can consider this PR the most advanced and close to what we need.

@jbalsas @wincent @pat270 @edalgrin Do we need something more to consider merging this solution?
Wait to see if it is fully possible on the portal?
Maybe a most advanced portal testing?

pat270:
> piecing components out

@marcoscv-work that wasn't related to reducing the selector. It was about trying to reduce total Cadmin size by moving components into Portal modules. I was waiting for more feedback on the selector name before I started changing everything. If no one has objections I can do that today.

New Selector: `html:not(#_):not(#_) .cad`
New variable prefix: `$cad-primary-l1`

Maybe we could change the directory name from `cadmin` to something more descriptive like `clay-admin`?

wincent:
> @jbalsas @wincent @pat270 @edalgrin Do we need something more to consider merging this solution?

I'd say first up, wait for @pat270 to say it's ready. Then we do one "final" pass. Consensus seems to be in favor of this solution, but I am not going to be around to foot the bill of any choice here, so you shouldn't care about my opinion. 😂

>  I was waiting for more feedback on the selector name before I started changing everything. If no one has objections I can do that today.
> 
> New Selector: `html:not(#_):not(#_) .cad`

`cad` is just obfuscation, and as you have said in more than one place (I think) @pat270, gzip is the right tool for managing size. Sadly, DXP does not ship with the `GzipFilter` active by default, so we are pretty much counting on anybody running DXP in production to have enough clue to know that they have to put `nginx` or something similar in front to handle compression. Over on the JS side, we advise people to [avoid abbreviations](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/naming.md#avoid-abbreviations), because we should optimize our source for consumption by _humans_ and use _tools_ (ie. compression, minification etc) to optimize it for consumption by machines.

matuzalemsteles:
> Consensus seems to be in favor of this solution, but I am not going to be around to foot the bill of any choice here, so you shouldn't care about my opinion. 😂

Hey @wincent despite that I would like to hear your opinions regardless 😁

> You make a good point. We don't need to remove/disable it from the Clay Repo. It can be done at the Portal level. We could include a warning (on clayui.com) about it being disabled in Portal and instructions on how to enable it.

@pat270 since we only do this within DXP, then we will still have all the components for both right? Admin and Clay. In Clay repo.

pat270:
@matuzalemsteles for now yes. I wrote that because of your documentation concern. If we include `cadmin` in clayui.com all we need to do with our Admin components is apply the `cadmin` class to the parent.

marcoscv-work:
> > piecing components out
> 
> @marcoscv-work that wasn't related to reducing the selector. It was about trying to reduce total Cadmin size by moving components into Portal modules. I was waiting for more feedback on the selector name before I started changing everything. If no one has objections I can do that today.
> 
> New Selector: `html:not(#_):not(#_) .cad`
> New variable prefix: `$cad-primary-l1`
> 
> Maybe we could change the directory name from `cadmin` to something more descriptive like `clay-admin`?

No objections here :-P
About the directory more descriptive like `clay-admin`. Sounds good to me.

wincent:
> Hey @wincent despite that I would like to hear your opinions regardless 😁

I don't have anything new to add here beyond what I previously said. (1) As a JS hipster, I am _personally_ attracted to the Shadow DOM solution, but a very significant contributor to that preference is that it's more in comfort zone, so I'm trying to be objective and not let that cloud my judgment; and (2) My main concern with the CSS alternative is maintainability (in terms of it being something that multiple team members can reasonably do), so that depends on people affirming that they understand it and could meaningfully contribute to it.

marcoscv-work:
I think we have improved the maintainability part, and probably it can be understood easily by team members, for customers or teams it should be easy to apply, just make your admin controls surrounded by `cadmin` class.

Just some ideas I think we can summary and adjust if we not agree:
- Clay admin should prioritize maintainability over optimization (since loading time is not so important for admin users).
- If finally, we move some admin styles to widgets we should try to keep a global `Clay admin` configuration by some type of propagation
- We will try to keep descriptive names and we will assume CSS will be always gziped (keeping GZipFilter active).

Personally, I think this approach will help with our Bootstrap break decision and Clay CSS growth
If we are able to apply this solution some parts of our products will have new spaces for improvement, live editors, themes, or even Liferay online.

matuzalemsteles:
Thanks @wincent, well I also have just the same sustainability concern as this, still trying to understand how it can be in the future, it is not 100% clear yet... but I'm fine since @pat270 and @marcoscv-work see value in this strategy since they are more experts.

I would love for us to continue still exploring better ways to do this even though we already have this implementation ready to merge, just like @wincent I'm a little biased with JS, so I would be curious to see how we could try to implement CSS-in-JS, I saw in some products like [Framer](https://www.framer.com) which is the most important piece for them to isolate the CSS of their platform so that the components added by users do not interfere, of course, they have other scenarios and a zero environment to try to implement this and in DXP things are totally different, but I would be curious to explore that side.

marcoscv-work (approved):
I agree CSS is much more performant than JS for this.

I think we have advantages in following the SASS format to build the admin solution, much easy to explain or document (from a CSS current structure perspective). so I agree with merging this PR if we have a consensus.

I've approved merging this PR!

marcoscv-work:
Trying to resolve it and advance in the implementation of a solution I'm trying to summarize the agreements in this conversation and other conversations in other channels/meetings:

After this long conversation (35 days from the first Patrick PR?) we have decided the performance for logged-in users with admin-controls is not so important and a priority to optimize, so:

- We agreed about the use of clay-admin and we have been working on adjustments last month
- We have decided to encapsulate admin general components with the class `.cadmin`
- We have decided to load clay admin components in widgets they may need.

These agreements are opening a door to
- The site optimization (our biggest CSS issue and something we have been facing for years):
  - Since we are separating clay admin from clay-site we could update the site by using only some clay modules, or even rewrite them without breaking admin controls, so the possibilities to optimize for us and our customer will change radically.
- Fix all issues with page editors and the difference between admin and themes
- Make Liferay online much more independence
- Opening a new world of possibilities for customization by custom properties since they will not need to modify admin controls and we are breaking with that idea
  - I remember when we work on custom properties, our main problem was that in most cases we were breaking admin controls so we have the CSS with a lot of exceptions and we expend like a month fixing issues related to custom properties
---
I think assuming that "best performance improvement" is CSS-in-JS can be an error.
Splitting of the clay CSS in admin modules to be loaded in parts by JS when parts of the administration are needed is a big project itself that even we are not sure about its viability, so we would need much more time to validate it when actually we have agreed on the important part is the **site optimization**, no the admin controls views optimizations.
On the other hand, I've been working on a portal PR to cover admin controls we would need to show Documents and media correctly and the result was that we need to load almost all admin modules, and the difference with Patrick clay admin are really poor, overall if we compare numbers when gziped is activated.

What I mean is that I cannot see the real justification to invest time to work on CSS-in-JS since our objectives (prioritize the site, no the admin view) and our CSS structure is not much compatible with CSS-in-JS idea. and of course, the problem of not been compatible with splitting in little modules is not a clay gap because clay never was conceptualized to do that.

From my point of view, this solution is the simplest because:
- We use CSS inheriting directly from clay so better to understand for us and for customers
- We don't need to implement new technology or philosophy (CSS-in-JS).
- We don't need to change our concept of Clay, just applying it for what was created for: admin controls
- The time to implement is the shortest.

matuzalemsteles:
Thanks @edalgrin and @marcoscv-work for summarizing the agreements here, well I think everyone seems more inclined to go with this, I don't want to stop it from entering, so we can go ahead with it and merge. @pat270 I will work on component issues so that I don't delay changes on the DXP side.

Regarding CSS-in-JS, I agree that this is a total paradigm shift and that it could require many changes and we would also have to verify the feasibility of this. It is worth remembering that there is a great advantage here, we can follow several directions in the CSS-in-JS world, build at build time and prepare for runtime, we can also reuse structure as we wish, so the possibilities are endless here I don't think that it would be a limitation to cover the cases we need but to summarize it, we don't need to focus on that now but as I said in the future we may want to reevaluate the options. We can have a clearer view of this in the future...

brunofernandezg:
It seems we reached an agreement :)

Are we ready to split the epic in tasks and transition it to "in progress" to start working on it?

We have another spike open:  https://issues.liferay.com/browse/LPS-131432

Do we need to finish it before working on the rest of the epic?

cc @dsanz

pat270:
We should create a separate issue for CSS in JS. We can do this incrementally if we hash out the fundamentals. We currently have an alpha? version of design tokens. If we can create a detailed design token structure, we can use that as the key (single source of truth) for all our CSS projects and maybe even other projects (e.g., Clay CSS, Portal CSS, CSS in JS, JS modules, and more). The hardest part is structuring it in a way that describes everything correctly.

brunofernandezg:
Thanks for the explanation!

> we need this inside Portal before we can start actual work on isolating components
If I understood correctly we are already doing actual work. Although there are still a couple of investigation tasks that will help us define how implement the epic, we already made a decision and we are going to implement it in any case.
Couldn't
- Identify admin only UI to scope in Portal and apply the cadmin class
- Find the best way to include cadmin.css in Portal
be seen as part of the work of developing the Epic?

of course the outcome will be the same in any case, but it would change the perception of an Epic still in analysis to one in which actual work is being done (and won't be undone) and we are progressing.

What do you think if we:
- Transition the epic from "in analysis" to "in progress"
and maybe turn both spikes into tasks (if you feel more confortable colling them spikes, no problem)? 
- Identify admin only UI to scope in Portal and apply the cadmin class
- Find the best way to include cadmin.css in Portal

Thanks :)

edalgrin (approved):

marcoscv-work:
So looks like everybody agrees! we can merge it! Patrick, you deserve it!

pat270:
Thanks @marcoscv-work ! Now there are two.
tada

pat270:
@brunofernandezg I converted the issues to subtasks. I can start sending them in after the next clay release.

brunofernandezg:
Great! Thanks for taking care of it :)

PR #4008, opened 2021-04-05
Implement date range on ClayDatePicker

diegonvs:
This PR is horrible ™️ . It's just a status of what I'm working on

This PR is on top of https://github.com/liferay/clay/pull/3991.

## In resume:

Replace `selectedDay` with `dateRange` that contains information about `fromDate` and `toDate` (these names can perhaps be changed to `startDate` and `endDate`). When not under `enableDateRange` We are going to have the `fromDate` value as default value.

## Known Issues:

By clicking on the date picker to select dates everything is working however we have two problems:

<s>1. The date-fns `isWithinInterval` function throws an error instead of returning a Boolean value. When editing the Date Input manually it fails because the date being entered is invalid and the error erupts preventing the user from continuing typing. I'm thinking of creating an internal value state just to control this input and not undergo any verification. </s> 

Fixed ✅ , I implemented my version of `isWithinInterval`

2. There is a visual bug in which when you have selected a date outside the range of the month as a start or end date, for example, when the date panel is in April and the date selected is May 1st, the date is still disabled however continues with the `c-selected-first` style and this causes a broken visual aspect.

See: 
image

What would the style look like in this @drakonux @hold-shift use case?

## Additional fixes

Added an aria-label option for the button with the calendar symbol: buttonChooseDate

PR #4012, opened 2021-04-10
feat(@clayui/date-picker): Adds ClayDatePicker Range variation

diegonvs:
It came from the draft sent here: https://github.com/liferay/clay/pull/4008

# In resume:

Replace selectedDay state with selectedDays that contains a tuple with info about start and end dates. When not under `range` We are going to have a unique element on the tuple and it can be used as default value.

# Possible breaking changes(?)

onValueChange callback will not have two possible types as before(Date | string). Only strings will be accepted.

# Boy Scouting

Some a11y improvements like:

* Added an aria-label option for the button with the calendar symbol: buttonChooseDate
* Added an aria-label for each day on the calendar improving the message to screen reader from `2012 12 21` to `Thusday Dec 12 2012`
* Added aria-modal="true" and role="dialog" as https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html shows for the DatePicker dialog component :)

### DIFF
--- a/packages/clay-date-picker/src/DayNumber.tsx
+++ b/packages/clay-date-picker/src/DayNumber.tsx
@@ -7,43 +7,84 @@ import {Keys} from '@clayui/shared';
 import classnames from 'classnames';
 import React from 'react';
 
-import {IDay, formatDate, setDate} from './Helpers';
+import {IDay, setDate} from './Helpers';
+import {TDaysSelected} from './types';
 
 interface IProps {
 	day: IDay;
-	daySelected: Date;
+	daysSelected: TDaysSelected;
 	disabled?: boolean;
-	onClick: (date: Date) => void;
+	range?: boolean;
+	onClick: (date: Date, range?: boolean) => void;
+}
+
+interface IInterval {
+	start: Date;
+	end: Date;
+}
+
+function isWithinInterval(dirtyDate: Date, dirtyInterval: IInterval) {
+	const interval = dirtyInterval || {};
+	const time = dirtyDate.getTime();
+	const startTime = interval.start.getTime();
+	const endTime = interval.end.getTime();
+
+	if (!(startTime <= endTime)) {
+		return false;
+	}
+
+	return time >= startTime && time <= endTime;
 }
 
 const ClayDatePickerDayNumber: React.FunctionComponent<IProps> = ({
 	day,
-	daySelected,
+	daysSelected,
 	disabled,
 	onClick,
+	range,
 }) => {
 	const {date, outside} = day;
+	const [startDate, endDate] = daysSelected;
+
+	const isSelectedToDate = date.toDateString() === endDate.toDateString();
+	const isSelectedFromDate = date.toDateString() === startDate.toDateString();
 
 	const classNames = classnames(
 		'date-picker-date date-picker-calendar-item',
 		{
-			active: date.toDateString() === daySelected.toDateString(),
+			active: isSelectedFromDate || (range && isSelectedToDate),
 			disabled: outside || disabled,
 		}
 	);
 
 	return (
-		<div className="date-picker-col">
+		<div
+			className={classnames(
+				'date-picker-col',
+				range && {
+					'c-selected':
+						!isSelectedFromDate &&
+						!isSelectedToDate &&
+						isWithinInterval(date, {
+							end: endDate,
+							start: startDate,
+						}),
+					'c-selected c-selected-end':
+						isSelectedToDate &&
+						!(isSelectedToDate && isSelectedFromDate),
+					'c-selected c-selected-start':
+						isSelectedFromDate &&
+						!(isSelectedToDate && isSelectedFromDate),
+				}
+			)}
+		>
 			<button
-				aria-label={formatDate(
-					setDate(date, {
-						hours: 12,
-						milliseconds: 0,
-						minutes: 0,
-						seconds: 0,
-					}),
-					'yyyy MM dd'
-				)}
+				aria-label={setDate(date, {
+					hours: 12,
+					milliseconds: 0,
+					minutes: 0,
+					seconds: 0,
+				}).toLocaleDateString()}
 				className={classNames}
 				disabled={outside}
 				onClick={() => onClick(date)}
--- a/packages/clay-date-picker/src/Helpers.ts
+++ b/packages/clay-date-picker/src/Helpers.ts
@@ -6,6 +6,8 @@
 import {default as formatDate} from 'date-fns/format';
 import {default as parseDate} from 'date-fns/parse';
 
+import {IYears, TDaysSelected} from './types';
+
 export {formatDate, parseDate};
 
 export interface IDay {
@@ -16,6 +18,36 @@ export interface IDay {
 export type WeekDays = Array<IDay>;
 
 export type Month = Array<WeekDays>;
+
+export const RANGE_SEPARATOR = ' - ';
+
+export function isDateRangeWithinYears(year: number, years: IYears) {
+	return year >= years.start && years.end >= year;
+}
+
+export function fromStringToRange(
+	value: string,
+	dateFormat: string,
+	referenceDate: Date
+): TDaysSelected {
+	const [fromDateString, toDateString] = value.split(RANGE_SEPARATOR);
+
+	return [
+		parseDate(fromDateString, dateFormat, referenceDate),
+		toDateString
+			? parseDate(toDateString, dateFormat, referenceDate)
+			: referenceDate,
+	];
+}
+
+export function fromRangeToString(range: TDaysSelected, dateFormat: string) {
+	const [startDate, endDate] = range;
+
+	return `${formatDate(startDate, dateFormat)}${RANGE_SEPARATOR}${formatDate(
+		endDate,
+		dateFormat
+	)}`;
+}
 
 /**
  * Clone a date object.
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -17,13 +17,21 @@ import React from 'react';
 import DateNavigation from './DateNavigation';
 import DayNumber from './DayNumber';
 import DaysTable from './DaysTable';
-import {isValid, parseDate, setDate} from './Helpers';
+import {
+	formatDate,
+	fromRangeToString,
+	fromStringToRange,
+	isDateRangeWithinYears,
+	isValid,
+	parseDate,
+	setDate,
+} from './Helpers';
 import {useCurrentTime, useWeeks} from './Hooks';
 import InputDate from './InputDate';
 import TimePicker from './TimePicker';
 import Weekday from './Weekday';
 import WeekdayHeader from './WeekdayHeader';
-import {FirstDayOfWeek, IAriaLabels, IYears} from './types';
+import {FirstDayOfWeek, IAriaLabels, IYears, TDaysSelected} from './types';
 
 export {FirstDayOfWeek};
 
@@ -93,15 +101,17 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	 *
 	 * Second argument gives the type that caused the value change
 	 */
-	onValueChange: (
-		value: Date | string,
-		type?: 'click' | 'input' | 'time'
-	) => void;
+	onValueChange: (value: string, type?: 'click' | 'input' | 'time') => void;
 
 	/**
 	 * Describe a brief tip to help users interact.
 	 */
 	placeholder?: string;
+
+	/***
+	 * Flag to indicate the user will use the date-range date-picker
+	 */
+	range?: boolean;
 
 	/**
 	 * Path to the location of the spritemap resource.
@@ -150,7 +160,8 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	onExpandedChange?: TInternalStateOnChange<boolean>;
 }
 
-const dateNow = new Date();
+const DATE_NOW = new Date();
+
 const TIME_FORMAT = 'HH:mm';
 
 /**
@@ -163,6 +174,7 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 	(
 		{
 			ariaLabels = {
+				buttonChooseDate: 'Choose date',
 				buttonDot: 'Select current date',
 				buttonNextMonth: 'Select the next month',
 				buttonPreviousMonth: 'Select the previous month',
@@ -174,7 +186,7 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			footerElement,
 			id,
 			initialExpanded = false,
-			initialMonth = new Date(),
+			initialMonth = DATE_NOW,
 			inputName,
 			months = [
 				'January',
@@ -194,6 +206,7 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			onNavigation = () => {},
 			onValueChange = () => {},
 			placeholder,
+			range,
 			spritemap,
 			time = false,
 			timezone,
@@ -201,8 +214,8 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			value,
 			weekdaysShort = ['S', 'M', 'T', 'W', 'T', 'F', 'S'],
 			years = {
-				end: dateNow.getFullYear(),
-				start: dateNow.getFullYear(),
+				end: DATE_NOW.getFullYear(),
+				start: DATE_NOW.getFullYear(),
 			},
 			...otherProps
 		}: IProps,
@@ -227,17 +240,15 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			normalizeDate(initialMonth)
 		);
 
+		// TODO - Rename to daysSelected
+		const [daysSelected, setDaysSelected] = React.useState<TDaysSelected>(
+			() => [normalizeDate(initialMonth), normalizeDate(initialMonth)]
+		);
+
 		/**
 		 * Indicates the time selected by the user.
 		 */
 		const [currentTime, setCurrentTime] = useCurrentTime(TIME_FORMAT);
-
-		/**
-		 * The day selected by the user.
-		 */
-		const [daySelected, setDaySelected] = React.useState<Date>(
-			initialMonth
-		);
 
 		/**
 		 * An array of the weeks and days list for the current month
@@ -282,12 +293,64 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 		};
 
 		/**
+		 * Holds the logic for how to behave when clicking on a date,
+		 * on the DatePicker using date ranges
+		 */
+		const handleDateRangeClicked = (
+			date: Date,
+			daysSelected: TDaysSelected
+		) => {
+			const [startDate, endDate] = daysSelected;
+
+			if (date < startDate) {
+				setDaysSelected([date, endDate]);
+
+				onValueChange(
+					fromRangeToString(daysSelected, dateFormat),
+					'click'
+				);
+
+				return;
+			}
+
+			setDaysSelected([startDate, date]);
+
+			onValueChange(
+				fromRangeToString([startDate, date], dateFormat),
+				'click'
+			);
+		};
+
+		/**
 		 * Handles the click on element of the day
 		 */
 		const handleDayClicked = (date: Date) => {
-			setDaySelected(date);
+			if (range) {
+				handleDateRangeClicked(date, daysSelected);
 
-			onValueChange(date, 'click');
+				return;
+			}
+
+			setDaysSelected([date, date]);
+
+			const dateFormatted = formatDate(date, dateFormat);
+
+			if (time) {
+				const dateString = formatDate(
+					parseDate(
+						`${dateFormatted} ${currentTime}`,
+						`${dateFormat} ${TIME_FORMAT}`,
+						date
+					),
+					`${dateFormat} ${TIME_FORMAT}`
+				);
+
+				onValueChange(dateString, 'click');
+
+				return;
+			}
+
+			onValueChange(dateFormatted, 'click');
 		};
 
 		/**
@@ -297,17 +360,30 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 		 */
 		const inputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
 			const {value} = event.target;
+
 			const format = time ? `${dateFormat} ${TIME_FORMAT}` : dateFormat;
 
-			const date = parseDate(value, format, new Date());
-			const year = date.getFullYear();
+			const [fromDate, toDate] = fromStringToRange(
+				value,
+				format,
+				DATE_NOW
+			);
 
-			if (isValid(date) && year >= years.start && years.end >= year) {
-				changeMonth(date);
-				setDaySelected(date);
+			const yearFrom = fromDate.getFullYear();
+			const yearTo = toDate.getFullYear();
+
+			if (
+				isValid(fromDate) &&
+				isValid(toDate) &&
+				isDateRangeWithinYears(yearFrom, years) &&
+				isDateRangeWithinYears(yearTo, years)
+			) {
+				changeMonth(fromDate);
+
+				setDaysSelected([fromDate, toDate]);
 
 				if (time) {
-					setCurrentTime(date.getHours(), date.getMinutes());
+					setCurrentTime(fromDate.getHours(), fromDate.getMinutes());
 				}
 			}
 
@@ -319,25 +395,37 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 		 * is no date selected.
 		 */
 		const handleDotClicked = () => {
-			changeMonth(dateNow);
-			setDaySelected(dateNow);
-			onValueChange(dateNow);
+			changeMonth(DATE_NOW);
+
+			setDaysSelected([DATE_NOW, daysSelected[1]]);
+
+			if (range) {
+				return onValueChange(
+					fromRangeToString([DATE_NOW, daysSelected[1]], dateFormat)
+				);
+			}
+
+			const time = parseDate(currentTime, TIME_FORMAT, DATE_NOW);
+
+			onValueChange(formatDate(time, `${dateFormat} ${TIME_FORMAT}`));
 		};
 
 		const handleTimeChange = (
 			hours: number | string,
 			minutes: number | string
 		) => {
-			const format = time ? `${dateFormat} ${TIME_FORMAT}` : dateFormat;
+			const format = `${dateFormat} ${TIME_FORMAT}`;
+			const [day] = daysSelected;
 
-			// Hack to force InputDate to add `currentTime` to the value of
-			// the input when the value was edited by the user.
-			if (
-				typeof value === 'string' &&
-				isValid(parseDate(value, format, new Date()))
-			) {
-				onValueChange(parseDate(value, format, new Date()), 'time');
-			}
+			const dateParsed = parseDate(value, format, DATE_NOW);
+			const newDate = isValid(dateParsed)
+				? dateParsed
+				: new Date(day.getTime());
+
+			newDate.setHours(Number(hours));
+			newDate.setMinutes(Number(minutes));
+
+			onValueChange(formatDate(newDate, format), 'time');
 
 			setCurrentTime(hours, minutes);
 		};
@@ -379,20 +467,18 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 							<InputDate
 								{...otherProps}
 								ariaLabel={ariaLabels.input}
-								currentTime={currentTime}
-								dateFormat={dateFormat}
 								disabled={disabled}
 								inputName={inputName}
 								onChange={inputChange}
 								placeholder={placeholder}
 								ref={ref}
-								time={time}
 								useNative={useNative}
 								value={value}
 							/>
 							{!useNative && (
 								<ClayInput.GroupInsetItem after>
 									<Button
+										aria-label={ariaLabels.buttonChooseDate}
 										className="date-picker-dropdown-toggle"
 										data-testid="date-button"
 										disabled={disabled}
@@ -418,7 +504,11 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 							onSetActive={setExpandedValue}
 							ref={dropdownContainerRef}
 						>
-							<div className="date-picker-calendar">
+							<div
+								aria-modal="true"
+								className="date-picker-calendar"
+								role="dialog"
+							>
 								<DateNavigation
 									ariaLabels={ariaLabels}
 									currentMonth={currentMonth}
@@ -445,10 +535,11 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 										{({day, key}) => (
 											<DayNumber
 												day={day}
-												daySelected={daySelected}
+												daysSelected={daysSelected}
 												disabled={disabled}
 												key={key}
 												onClick={handleDayClicked}
+												range={range}
 											/>
 										)}
 									</DaysTable>
--- a/packages/clay-date-picker/src/InputDate.tsx
+++ b/packages/clay-date-picker/src/InputDate.tsx
@@ -6,16 +6,11 @@
 import {ClayInput} from '@clayui/form';
 import React from 'react';
 
-import {formatDate, isValid} from './Helpers';
-
 interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	ariaLabel?: string;
-	currentTime: string;
-	dateFormat: string;
 	disabled?: boolean;
 	inputName?: string;
 	placeholder?: string;
-	time: boolean;
 	useNative: boolean;
 	value: string;
 }
@@ -24,41 +19,23 @@ const ClayDatePickerInputDate = React.forwardRef<HTMLInputElement, IProps>(
 	(
 		{
 			ariaLabel,
-			currentTime,
-			dateFormat,
 			inputName = 'datePicker',
-			time = false,
 			useNative = false,
 			value = '',
 			...otherProps
 		},
 		ref
 	) => {
-		const isValidValue = (value: string | Date): string => {
-			if (value instanceof Date && isValid(value)) {
-				const date = formatDate(value, dateFormat);
-
-				return time ? `${date} ${currentTime}` : date;
-			}
-
-			return value as string;
-		};
-
-		const memoizedValue = React.useMemo(() => isValidValue(value), [
-			value,
-			currentTime,
-		]);
-
 		return (
 			<>
-				<input name={inputName} type="hidden" value={memoizedValue} />
+				<input name={inputName} type="hidden" value={value} />
 				<ClayInput
 					{...otherProps}
 					aria-label={ariaLabel}
 					insetAfter={!useNative}
 					ref={ref}
 					type={useNative ? 'date' : 'text'}
-					value={memoizedValue}
+					value={value}
 				/>
 			</>
 		);
--- a/packages/clay-date-picker/src/types.ts
+++ b/packages/clay-date-picker/src/types.ts
@@ -14,6 +14,7 @@ export enum FirstDayOfWeek {
 }
 
 export interface IAriaLabels {
+	buttonChooseDate: string;
 	buttonDot: string;
 	buttonNextMonth: string;
 	buttonPreviousMonth: string;
@@ -24,3 +25,5 @@ export interface IYears {
 	end: number;
 	start: number;
 }
+
+export type TDaysSelected = [Date, Date];
--- a/packages/clay-date-picker/stories/index.tsx
+++ b/packages/clay-date-picker/stories/index.tsx
@@ -15,12 +15,24 @@ const ClayDatePickerWithState = (props: {[key: string]: any}) => {
 	const [value, setValue] = React.useState<string | Date>('');
 
 	return (
-		<ClayDatePicker
-			{...props}
-			onValueChange={setValue}
-			spritemap={spritemap}
-			value={value as string}
-		/>
+		<>
+			<label>{'Date Picker'}</label>
+			<ClayDatePicker
+				{...props}
+				ariaLabels={{
+					buttonChooseDate: `Choose Date, selected date is ${
+						value.toLocaleString() ?? value
+					}`,
+					buttonDot: 'Go to today',
+					buttonNextMonth: 'Next month',
+					buttonPreviousMonth: 'Previous month',
+					input: value.toLocaleString(),
+				}}
+				onValueChange={setValue}
+				spritemap={spritemap}
+				value={value as string}
+			/>
+		</>
 	);
 };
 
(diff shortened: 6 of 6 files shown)


packages/clay-date-picker/src/Helpers.ts
```diff
@@ -17,6 +19,36 @@ export type WeekDays = Array<IDay>;
 
 export type Month = Array<WeekDays>;
 
+export const RANGE_SEPARATOR = ' - ';
+
+export function isDateRangeWithinYears(year: number, years: IYears) {
+	return year >= years.start && years.end >= year;
```
wincent, line 25: While this works, it doesn't read quite as naturally as it could. This is subtle, but let me see if I can explain what I mean.

Say you have a range of years that starts at `A` and ends at `Z`, and you want to find out whether year `J` falls somewhere in the middle. ie. on a timeline, is J in the middle:

```
- - - - A - - - - - - - - - - J - - - - - - Z - - - - →
      (start year)    (candidate year)   (end year)
```

It's possible to write you function so that the `true` result happens (ie. the "happy" path), the variables corresponding to `A`, `J` and `Z` appear in your expression that order:

```
//         A           J       J         Z
return years.start <= year && year <= years.end;
```

(In your version, the order is `J` `A`, `Z`, `J`)

Anyway, very subjective, but food for thought.

packages/clay-date-picker/src/Helpers.ts
```diff
@@ -17,6 +19,36 @@ export type WeekDays = Array<IDay>;
 
 export type Month = Array<WeekDays>;
 
+export const RANGE_SEPARATOR = ' - ';
+
+export function isDateRangeWithinYears(year: number, years: IYears) {
```
wincent, line 24: This function name doesn't seem to describe what it does. Name says that it tells you whether a date range (presumably with a start and end) falls within `years` (which implies a range). But it looks to me like what it actually does is tell you whether a year falls within a range (of years).

So maybe it should be called something like `isYearWithinYearRange()` (or `isYearWithinYears()`).

packages/clay-date-picker/src/index.tsx
```diff
@@ -17,13 +17,21 @@ import React from 'react';
...
-import {isValid, parseDate, setDate} from './Helpers';
+import {
+	formatDate,
+	fromRangeToString,
+	fromStringToRange,
+	isDateRangeWithinYears,
+	isValid,
+	parseDate,
+	setDate,
+} from './Helpers';
```
wincent, line 28: Seems like several of these functions are used _only_ in this file (eg. `fromRangeToString`, `fromStringToRange`, `isDateRangeWithinYears`) so by putting them there, you are just obliging the reader to jump into another file to see what they do. My recommendation would be to put functions close to where they are needed (improving "locality" is worth doing for many other kinds of abstractions too, not just functions).

packages/clay-date-picker/src/index.tsx
```diff
@@ -150,7 +160,8 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	onExpandedChange?: TInternalStateOnChange<boolean>;
 }
 
-const dateNow = new Date();
+const DATE_NOW = new Date();
```
wincent, line 163: I know this was like this before, but this seems bad. It means that "now" means the time the module was loaded and not when the component was rendered. That may sound a bit edge casey, but I would expect this to misbehave around midnight at the end of the month, and the end of the year. The cost of avoiding the bug looks basically zero to me (just move it inside the component), so may as well do it. What do you think?

packages/clay-date-picker/src/index.tsx
```diff
@@ -227,18 +240,16 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			normalizeDate(initialMonth)
 		);
 
+		// TODO - Rename to daysSelected
```
wincent, line 243: Kill this comment.

packages/clay-date-picker/src/index.tsx
```diff
@@ -227,18 +240,16 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+		const [daysSelected, setDaysSelected] = React.useState<TDaysSelected>(
```
wincent, line 244: Weird that TS can't infer the type here...

wincent (commented):
Not a real review. Just some weekend code-reading. Feel free to ignore me.

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,84 @@ import {Keys} from '@clayui/shared';
...
+	end: Date;
+}
+
+function isWithinInterval(dirtyDate: Date, dirtyInterval: IInterval) {
+	const interval = dirtyInterval || {};
+	const time = dirtyDate.getTime();
+	const startTime = interval.start.getTime();
+	const endTime = interval.end.getTime();
+
+	if (!(startTime <= endTime)) {
```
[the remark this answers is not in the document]
diegonvs: `if (startTime > endTime) {` it seems more adequated. :)

> under what conditions do you expect this function might be called with an interval that fails this test?

This function was used to see if a given date is within an interval or not for marking the DayNumber as selected or not.

The happy path:

If the date `02/13/2021(mm/dd/yyyy)` was given as a "dirtyDate" and the interval was starting as `02/11/2021` and ending in `02/14/2021`

this function will fail to pass a value outside the range. Like passing `02/13/2021` and a interval(start|end): `02/14/2021 | 02/20/2021`

I need to write tests for it but when testing it intensively while developing the feature it did not bring any strange behavior.

packages/clay-date-picker/src/Helpers.ts
diegonvs: isYearWithinYears seems appropriate :)

packages/clay-date-picker/src/index.tsx
diegonvs: Makes sense. Initially I didn't move these functions to Helpers file but after the first review I moved it for reducing the LOC in the same file. :)

diegonvs: So, just makes sense moving the Helpers files if a function is used in more than more file

packages/clay-date-picker/src/index.tsx
diegonvs: Yes, it also makes sense. So, Can I just use `new Date();` in every place that I use?

packages/clay-date-picker/src/index.tsx
diegonvs: 🙈 🙈 🙈 🙈 🙈

packages/clay-date-picker/src/index.tsx
diegonvs: I created `TDaysSelected` just to have a better name than <[Date, Date]>. So, isn't a good practice?

diegonvs:
> Not a real review. Just some weekend code-reading. Feel free to ignore me.

Thank you for your amazing comments 💖

packages/clay-date-picker/src/Helpers.ts
matuzalemsteles: Makes sense for me!!

packages/clay-date-picker/src/index.tsx
```diff
@@ -281,13 +292,65 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+					),
+					`${dateFormat} ${TIME_FORMAT}`
+				);
+
+				onValueChange(dateString, 'click');
+
+				return;
+			}
+
+			onValueChange(dateFormatted, 'click');
```
matuzalemsteles, line 353: Why not keep it all together instead of separating it into two functions, just thinking about simplifying it a little bit here, maybe something like this:

```js
/**
 * Handles the click on element of the day
 */
const handleDayClicked = (date: Date) => {
	const [startDate, endDate] = daysSelected;

	let newDaysSelected ;
	let daysSelectedToString;

	if (range) {
		newDaysSelected = date < startDate ? [date, endDate] : [startDate, date];

		daysSelectedToString = fromRangeToString(newDaysSelected, dateFormat);
	} else {
		// Ideally, just set a date: [date]
		newDaysSelected = [date, date];

		daysSelectedToString = formatDate(date, dateFormat);

		if (time) {
			daysSelectedToString = formatDate(
				parseDate(
					`${daysSelectedToString} ${currentTime}`,
					`${dateFormat} ${TIME_FORMAT}`,
					date
				),
				`${dateFormat} ${TIME_FORMAT}`
			);
		}
	}

	setDaysSelected(newDaysSelected);
	onValueChange(daysSelectedToString, 'click');
}
```

You can avoid repeatedly calling `onValueChange` and `setDaysSelected` and you can slightly improve the readability of the use cases.

The bad thing about maintaining it as in PR is that it is difficult to read the use cases that exist, seeing many calls of `onValueChange` or `setDaysSelected` is like creating several blocks of edge cases and it is difficult to make one sequential reading of what is happening. 

Separating into another function can help as you did but it will just repeat more code and more lines and maybe complicate it a little more because it will not be reused elsewhere.

packages/clay-date-picker/src/index.tsx
```diff
@@ -319,25 +395,37 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+
+			if (range) {
+				return onValueChange(
+					fromRangeToString([DATE_NOW, daysSelected[1]], dateFormat)
+				);
+			}
+
+			const time = parseDate(currentTime, TIME_FORMAT, DATE_NOW);
+
+			onValueChange(formatDate(time, `${dateFormat} ${TIME_FORMAT}`));
```
matuzalemsteles, line 410: You can achieve the same thing like this:

```js
onValueChange(
	range
		? fromRangeToString([DATE_NOW, daysSelected[1]], dateFormat)
		: formatDate(
			parseDate(currentTime, TIME_FORMAT, DATE_NOW),
			`${dateFormat} ${TIME_FORMAT}`
		)
);
```

matuzalemsteles (commented):
@diegonvs I added some more comments to try to simplify some more things, let me know if you have any questions.

packages/clay-date-picker/src/index.tsx
wincent: I wasn't thinking about best practices or anything; just surprised that TS didn't infer it automatically (because it is usually pretty good at figuring out the types involved in a `useState` call). I mean, you might think it could be confused by the `Object.keys().reduce()` stuff in `setDate()` (which is called from `normalizeDate()`), and `setDate()` doesn't have a return type annotation... but my editor says TS _is_ inferring it `setDate` as:

```
function setDate(...blah}): Date
```

and the return type (`Date`) is correct.

Just testing this locally, I see that if you remove the explicit annotation the type of the `useState` gets inferred as `Date[]` as opposed to `[Date, Date]` (ie. array instead of tuple), so what you have here is stricter, and stricter is good. Personally though, I would just give TS a hint (`as const`) that you want a tuple here, and drop the annotation:

```ts
const [daysSelected, setDaysSelected] = React.useState(
	() => [normalizeDate(initialMonth), normalizeDate(initialMonth)] as const
);
```

Same with the `useState<Date>` a few lines above; if you drop the `<Date>` it can still figure out the type:

```ts
const [currentMonth, setCurrentMonth] = React.useState(() =>
	normalizeDate(initialMonth)
);
```

My rationale here is that our plans to adopt TS in liferay-portal are going to face less resistance from people unfamiliar with TS if we avoid the syntax and clutter of type annotations in places that they can be inferred (the type info is always only a tool-tip away, anyway).

Separate topic is whether:

- `TDaysSelected`
- `[Date, Date]`

is better or more readable. For simple types, I prefer to "inline" them because I can tell immediately _exactly_ what `[Date, Date]` is, whereas with `TDaysSelected` I have to do some guess-work: is it an object with day keys and `true`/`false` values? An array with one slot for each day of the week? One slot for each day of the month? Is it a `Set` containing a bunch of `Date` objects? etc.

packages/clay-date-picker/src/index.tsx
wincent: > I would just rename the variable to something like `const NEW_DATE = new Date()`.

Yep, that's exactly what I was thinking when I said:

> just move it inside the component

packages/clay-date-picker/src/index.tsx
diegonvs: I didn't knew about `as const` for tuples. Thanks for helping :)))

About the `TDaysSelected` makes sense don't use a separated type definition on simpler types. I'm removing and replacing with `readonly [Date, Date]` ;)

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,84 @@ import {Keys} from '@clayui/shared';
...
+interface IInterval {
+	start: Date;
+	end: Date;
+}
+
+function isWithinInterval(dirtyDate: Date, dirtyInterval: IInterval) {
+	const interval = dirtyInterval || {};
+	const time = dirtyDate.getTime();
+	const startTime = interval.start.getTime();
+	const endTime = interval.end.getTime();
```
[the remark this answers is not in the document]
diegonvs: I can't see any differences when removing `|| {}`. In this case, should I warn the user about the missing `end|start` properties? I can't see defining a default value for it like new Date(). It may broke or introduce some weird usages(like ppl considering when not passing a start or an end the new Date() will be considered.

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,84 @@ import {Keys} from '@clayui/shared';
...
+	daysSelected,
 	disabled,
 	onClick,
+	range,
 }) => {
 	const {date, outside} = day;
+	const [startDate, endDate] = daysSelected;
+
+	const isSelectedToDate = date.toDateString() === endDate.toDateString();
+	const isSelectedFromDate = date.toDateString() === startDate.toDateString();
```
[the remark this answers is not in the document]
diegonvs: makes sense. I forget update it ;)

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,84 @@ import {Keys} from '@clayui/shared';
...
+import {TDaysSelected} from './types';
 
 interface IProps {
 	day: IDay;
-	daySelected: Date;
+	daysSelected: TDaysSelected;
 	disabled?: boolean;
-	onClick: (date: Date) => void;
+	range?: boolean;
+	onClick: (date: Date, range?: boolean) => void;
```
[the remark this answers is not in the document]
diegonvs: Yes! I forgot!

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,84 @@ import {Keys} from '@clayui/shared';
...
+	const [startDate, endDate] = daysSelected;
+
+	const isSelectedToDate = date.toDateString() === endDate.toDateString();
+	const isSelectedFromDate = date.toDateString() === startDate.toDateString();
 
 	const classNames = classnames(
 		'date-picker-date date-picker-calendar-item',
 		{
-			active: date.toDateString() === daySelected.toDateString(),
+			active: isSelectedFromDate || (range && isSelectedToDate),
```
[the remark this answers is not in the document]
diegonvs: Yes, it can be just `isSelectedFromDate ` :)
Screen Shot 2021-04-13 at 17 36 02

Another improvement is from: `hasStartDateSelected && !(hasEndDateSelected && hasStartDateSelected),` we can just `a && !hasEndDateSelected,` :)
Screen Shot 2021-04-13 at 17 39 47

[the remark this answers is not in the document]
diegonvs: I'm not good at boole's algebra for now but I use: https://www.dcode.fr/boolean-expressions-calculator for simplifying :)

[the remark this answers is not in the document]
diegonvs: Result:

Screen Shot 2021-04-13 at 17 43 04

packages/clay-date-picker/src/index.tsx
```diff
@@ -227,18 +240,16 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			normalizeDate(initialMonth)
 		);
 
+		// TODO - Rename to daysSelected
+		const [daysSelected, setDaysSelected] = React.useState<TDaysSelected>(
+			() => [normalizeDate(initialMonth), normalizeDate(initialMonth)]
```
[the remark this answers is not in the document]
diegonvs: Moved normalizeDate outside component function!

packages/clay-date-picker/src/DayNumber.tsx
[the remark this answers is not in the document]
matuzalemsteles: I was not referring to whether the conditional was right or not but I had the impression that `active` was only to be added when it is not a range but we need it regardless. Glad you managed to improve.

@pat270 I probably got it wrong here when you were saying it was different but I had a feeling that the `c-selected` would do the same as `active`, sorry for the confusion!! now it makes more sense 😅

diegonvs:
`date entered in the input element should reflect on the date picker` and `value entered in the input value must be reflected in the time picker` tests will fail.

Looks like on Jest environment, day button `.active` styles are not being applied when writing the correct date on the input. However, testing locally it works well. I tried to use fake timers and use async testing-library methods but It didn't worked :/

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,80 @@ import {Keys} from '@clayui/shared';
...
-						seconds: 0,
-					}),
-					'yyyy MM dd'
-				)}
+				aria-label={setDate(date, {
+					hours: 12,
+					milliseconds: 0,
+					minutes: 0,
+					seconds: 0,
+				}).toDateString()}
```
matuzalemsteles, line 83: The only problem with using it like this is that we will have problems with internationalization.

packages/clay-date-picker/src/index.tsx
```diff
@@ -27,6 +32,39 @@ import {FirstDayOfWeek, IAriaLabels, IYears} from './types';
...
+	range: readonly [Date, Date],
+	dateFormat: string
+) {
+	const [startDate, endDate] = range;
+
+	return `${formatDate(startDate, dateFormat)}${RANGE_SEPARATOR}${formatDate(
+		endDate,
+		dateFormat
+	)}`;
+}
```
matuzalemsteles, line 66: Although I have a preference to keep it in the helpers, I have the feeling that the component is more bloated with information, the separation of files helps me to read better as opposed to harming at least for me. I have a preference to keep this below the interface and to consider the interface declaration first in the component file, it is easier to know what the component expects.

cc @wincent

packages/clay-date-picker/src/index.tsx
```diff
@@ -281,13 +330,62 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+					`${dateFormat} ${TIME_FORMAT}`
+				);
 
-			onValueChange(date, 'click');
+				onValueChange(dateString, 'click');
+
+				return;
+			}
+
+			onValueChange(dateFormatted, 'click');
```
matuzalemsteles, line 388: I still think we can consider the comment I made about it https://github.com/liferay/clay/pull/4012#discussion_r612015851. We can talk more tomorrow because you didn't follow through.

packages/clay-date-picker/src/index.tsx
```diff
@@ -319,24 +430,46 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+				const time = parseDate(currentTime, TIME_FORMAT, NEW_DATE);
+
+				return onValueChange(
+					formatDate(time, `${dateFormat} ${TIME_FORMAT}`)
+				);
+			}
+
+			const dateFormatted = formatDate(NEW_DATE, dateFormat);
+
+			return onValueChange(dateFormatted);
```
matuzalemsteles, line 453: The same here.

```js
onValueChange(
	range
		? fromRangeToString([DATE_NOW, daysSelected[1]], dateFormat)
		: time
			? formatDate(
				parseDate(currentTime, TIME_FORMAT, DATE_NOW),
				`${dateFormat} ${TIME_FORMAT}`
			)
			: formatDate(NEW_DATE, dateFormat)
);
```

If the ternary gets too weird, we still have other options to consider:

```js
if (range) {
	onValueChange(
		fromRangeToString([NEW_DATE, daysSelected[1]], dateFormat)
	);
} else if (time) {
	onValueChange(
		formatDate(
			parseDate(currentTime, TIME_FORMAT, NEW_DATE),
			`${dateFormat} ${TIME_FORMAT}`
		)
	);
} else {
	onValueChange(formatDate(NEW_DATE, dateFormat));
}
```

or (Particularly in that way I look a little weird yet)

```js
let dateFormatted;

if (range) {
	dateFormatted = fromRangeToString([NEW_DATE, daysSelected[1]], dateFormat);
} else if (time) {
	dateFormatted = formatDate(
		parseDate(currentTime, TIME_FORMAT, NEW_DATE),
		`${dateFormat} ${TIME_FORMAT}`
	)
} else {
	dateFormatted = formatDate(NEW_DATE, dateFormat);
}

onValueChange(dateFormatted);
```

@wincent any thoughts on that?

packages/clay-date-picker/src/index.tsx
```diff
@@ -445,10 +580,11 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 										{({day, key}) => (
 											<DayNumber
```
matuzalemsteles, line 581: Regarding this behavior change in `DayNumber` it is still valid considering my previous comment about it https://github.com/liferay/clay/pull/4012#discussion_r612004289. We can talk more tomorrow because you didn't follow through.

matuzalemsteles (commented):
Adding a few more follow-up comments

wincent:
> `date entered in the input element should reflect on the date picker` and `value entered in the input value must be reflected in the time picker` tests will fail.
> 
> Looks like on Jest environment, day button `.active` styles are not being applied when writing the correct date on the input. However, testing locally it works well. I tried to use fake timers and use async testing-library methods but It didn't worked :/

Pulling this down I see lots of failed snapshots, plus 3 other failures ([full 9k-line gist here](https://gist.github.com/wincent/fc16404da1eda7325b871396d324be9a)). Do we have something locale-aware in the test suite? If so, we should force the locale to some known value so that the tests behave the same regardless of where they are run.

```
Test Suites: 3 failed, 50 passed, 53 total
Tests:       28 failed, 2 skipped, 359 passed, 389 total
Snapshots:   17 failed, 253 passed, 270 total
```

Ignoring the snapshot failures I get a more reasonable 900 lines of output ([gist](https://gist.github.com/wincent/ac7dadf102dbca7b51ee61cff8ecf509)):

```
Test Suites: 1 failed, 52 passed, 53 total
Tests:       11 failed, 2 skipped, 376 passed, 389 total
Snapshots:   17 updated, 253 passed, 270 total
```

I'm not going to dig any further because I think the possible local issues are just confounding diagnosis here and we should resolve those first. Could also be something local to my system 🤷; I just pulled the PR, ran `yarn run build` and `yarn run test`.

packages/clay-date-picker/src/index.tsx
wincent: There are arguments for and against, but I come down on the side of preferring locality (or colocation, or whatever you want to call it). Here are some reasons:

- If you put it in a separate file, the user has to jump to another place to look at it.
- If all you want to do is get it "out of the way", you can move it to the bottom of the same file (because functions are all "hoisted" in JS; ie. usable in code that appears above their definition); this way, a person landing in the file can still see the important stuff (ie. the component) at the top.
- Having it in a separate file means having to `import` it, which clutters the header area of the file with more boilerplate.
- Keeping it in the same file means that you can rename, move this file and know this internal function that it needs will always go along for the ride. Conversely, if you extract functions out into another file, every time you want to move that other file, you have to update all the imports. In other words, splitting stuff into more files adds friction and makes iteration/evolution slower.
- Putting it in a separate file means that anybody looking at the function cannot change it without first searching for an analysis all usage sites (eg. they can't rename it, they can't change the behavior etc); conversely, if the function is in the same file as the only user, it is easy to rename/refactor, and contemplate other behavior changes. Another way of saying this is that the cognitive and adminisrative load of reading and understanding an _internal_ function is always going to be lower than that of an _exported_ function.
- File size is a weak argument for wanting to extract functions; modern editors and computers can easily handle large files, and provide good tools for visualizing and organizing them (eg. code-folding, split panes etc).

And even though all of those arguments apply to functions, they resonate with a broader trend in the industry to recognize the value of colocation in general: GraphQL libraries colocate query fragments with the components that need the corresponding data; CSS-in-JS libraries colocate style definitions inside the components that need those styles; even JSX itself is an example (remember the "rethink best practices" meme? That was all about saying that what we had been calling "separation of concerns" was actually "separation of technologies", so splitting stuff into JS + HTML + CSS was wrong, and a better unit of abstraction and encapsulation was the "component", which is why we should "mix" together presentational, structural, and behavioral concerns in the same module, redefining what we consider to be a "concern"... that's actual an argument for colocation in disguise).

Obviously, it's up to you how you divide up your code and none of this is blocking feedback. I just honestly consider this to be good advice and want to share it. Similar arguments have been made by people much smarter than me (eg. [Carmack, here](https://twitter.com/id_aa_carmack/status/1333605572617363458)):

> A long time ago, I wrote about how a big function can be better than many small ones: http://number-none.com/blow/john_carmack_on_inlined_code.html I think it also applies to processes in a system.  I just said this: "A million line codebase is almost cozy today; I think we could combine many things and benefit."

So, at the surface level, he's talking about distrubuted systems. But what he's really talking about is colocation.

packages/clay-date-picker/src/DayNumber.tsx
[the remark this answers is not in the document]
wincent: > should I warn the user about the missing `end|start` properties?

In general when it comes to warnings, it's a trade-off. 🤷 

- If you warn, it might be spammy, but also possibly helpful.
- And irrespective of whether you warn, you should degrade "gracefully" in some way (requires you to come up with a definition of what "degrade gracefully" means, but avoids the spam, at the cost of potential eliding helpful information).

In this specific instance, they'd only be missing because of _you_ didn't provide them in your fallback object (`{}`), so I don't think I'd want to warn about that. In any case, I only see one caller of `isWithinInterval`, and it always passes an object there, so it doesn't look like the fallback would ever be used. So I'd say, just delete the `|| {}` (in fact, I am almost surprised that TS doesn't warn that its a useless expression). That still leaves open the broader question about prop validation, but that would/should occur "upstream" of here, in the component. Getting rid of the fallback removes the obvious local inconsistency internal to this function.

packages/clay-date-picker/src/index.tsx
wincent: Nested ternaries are obviously a bit harder to read than spelling it out with `if`/`else if`/`else`. I tend to reserve them for very specific cases (like in JSX, if the conditions are very simple, and brief enough not to wrap excessively).

So I like option 2 or 3. Option 2 is more verbose (😢 — more stuff to read), but option 3 has a `let` (😢 — more cognitive load because you have to look for mutation/assigment). I'd probably lean towards 3 for now, knowing that [once JS gets "do expressions"](https://github.com/tc39/proposal-do-expressions) you will be able to have the best of both worlds, perhaps (a `const` assignment plus DRY-ing up the calls to `onValueChange()`):

```js
const dateFormatted = do {
	if (range) {
		fromRangeToString([NEW_DATE, daysSelected[1]], dateFormat)
	} else if (time) {
		formatDate(
			parseDate(currentTime, TIME_FORMAT, NEW_DATE),
			`${dateFormat} ${TIME_FORMAT}`
		)
	} else {
		formatDate(NEW_DATE, dateFormat)
	}
};

onValueChange(dateFormatted);
```

Only stage 1 at this time though, so don't hold your breath.

packages/clay-date-picker/src/DayNumber.tsx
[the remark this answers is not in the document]
diegonvs: > In general when it comes to warnings, it's a trade-off. 🤷

> If you warn, it might be spammy, but also possibly helpful.
> And irrespective of whether you warn, you should degrade "gracefully" in some way (requires you to come up with a definition of what "degrade gracefully" means, but avoids the spam, at the cost of potential eliding helpful information).

Thanks for the explanation! 

Considering the current usage I understood the fallback may be unnecessary. So, I'll remove `|| {}` :)

packages/clay-date-picker/src/index.tsx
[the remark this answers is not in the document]
diegonvs: Added ;)

```tsx
		/**
		 * DaysSelected is a tuple that represents [startDate, endDate]
		 * in the cases where We have a date range and when `range` property
		 * is disabled we will just use the first element of the tuple(startDate)
		 */
		const [daysSelected, setDaysSelected] = React.useState(
			() =>
				[
					normalizeDate(initialMonth),
					normalizeDate(initialMonth),
				] as const
		);
```

packages/clay-date-picker/src/DayNumber.tsx
diegonvs: I forgot to test when changing browser's language. I'll take a look.

diegonvs: Yes, according [a website](https://www.globalyzer.com/gzserver/help/reference/localeSensitiveMethods/js_toDateString.html) when using `toDateString` We may enter in some i18n issues.[ I found an utility for it on date-fns](https://date-fns.org/v2.20.3/docs/I18n), we can pass a locale to the format function. [However, We need to provide the locale in `ISO 639-1`](https://date-fns.org/v2.20.3/docs/Locale).

packages/clay-date-picker/src/index.tsx
matuzalemsteles: @wincent Yeah, these are great arguments, it's a different perspective from that point of view looking at how components "should" be structured so as not to require separation of technology or just to bring things together in the same file, I agree that this reduced the cognitive load for make changes.

Well, about large files it is really quite debatable or swell with a lot of information to put the intrinsic complexity under the carpet also a point but it is also quite relative and varies from case to case, for example, during migration I saw file with minimum 2 thousand lines, some 3 thousand and a large application in just one component that made "millions" of rules, so I would consider this a large file and had an explicit high complexity, all the things related to the component/app were there, that of course imposes other problems that are not related to the locality but as the component was created without responsibility what I saw myself wasting hours trying to understand to discover the edge cases, in the end give up trying to understand why it was Metal.js and went with a solution of zero, this issue doesn’t really matter, but the result was breaking the component into small pieces with responsibilities where each small component could be composed and reused in other places, so in the end it became about 10 files which makes it easier to understand part of the application and makes it easy to reason what’s happening and helps to discover flaws more quickly... although locality is not imposed here I mean that this is more a matter of balance and depends a lot on the case and how it is done.

For some cases allocating everything can mean adding complexity or just keeping everything close by with something very structured, like, let's keep HTML + CSS + JavaScript close, it took us a while to add it all close because it required us to rethink the shape of everyone together, PHP started a part of this but quickly started to generate problems at scale, then we started to evolve to the frameworks, without considering MVC here..., on the JavaScript side, some find Vue interesting for doing this, uniting everything in the same file, some find that can get complex, so we have JSX + CSS in JS + JS in the same file, they are different technologies but they all converge to JS that approximate the same corresponding technology, this means that to understand and benefit from the locality we would need to structure how each technology it would be structured within the same file or how to achieve this without getting complex to understand or scale. I hope you understand the point of view that I am raising here.

Bringing it to our reality in Clay, we can sometimes think that 500 lines or 200 is something very big, depending on the point of view and the component maybe it is something very big or maybe not, I think that should not be taken into account so much for to say if we should take all the functions and put it on the same file where it is being used but we should in a way evaluate if it makes sense in the proposal: DatePicker `index.js` needs to know this, it just concentrates the composition of all components, assembling the blocks... well I understood the objectives of the locality but I see that there must always be a scale and it varies a lot from case to case I don’t have the cake recipe of when we should follow this against the locality so I won’t be able to argue much here.

Thank you for bringing this point of view!

packages/clay-date-picker/src/index.tsx
matuzalemsteles: Yeah, the ternary here ended up degrading the reading here, sad... I'm also not a big fan of option 3 because it requires mutation in the variable but I wanted to avoid the repetitive function call.

the expression `do` could really help, I remember seeing it being used in JSX some time ago but it seems to have been stagnant 🤷‍♂️. Maybe in some parallel world, we can reduce this if we structure other parts of the component to converge in a simpler logical change 😄.

packages/clay-date-picker/src/index.tsx
wincent: Yeah, the answer is always "it depends". One has to be careful about transferring one's experiences from 2,000 or 3,000 line "nightmare" files in a project that (from what I've heard) had really tied itself into knots, into different contexts.

You can easily find counter examples where well-written multi-thousand-line files have worked out brilliantly... (look in the Git project, or the Linux kernel, or TypeScript etc). Size alone should _not_ be the criteria for judging whether something is "maintainable" or not.

matuzalemsteles: Yes, I totally agree. I am not bringing this up as being big will be bad but it really depends, just so that other people when they see it try to take into account the points before going with locality or not.

As I had commented to get to the point of having everything together without achieving what was done in PHP at the beginning, it demanded that we do something well structured, so for sure this comes to the point here of taking as examples to reach a healthy locality or not. Considering the same ones you mentioned, Linux Kernel, TypeScript, At first, even the codebase in Go try to do this as good practice, I'm not sure... (Or just by limitation)

diegonvs:
Regarding the questions mentioned on the PR description, @drakonux answered [here](https://github.com/liferay/clay/pull/4008#issuecomment-819244473) and I solved many cases subsequently. See https://github.com/liferay/clay/pull/4008#issuecomment-819789480. The only one that I'm waiting a definition is the first one. 
Maybe @pat270 can take the CSS part and @matuzalemsteles forward with my work here. I'll be on PTO til 4th May.

matuzalemsteles:
Well, maybe I will look at it later but there are some problems here that we still need to fix, I can't set the start date after the end date or between it, just coming back, this is very strange because I can not define a range in the coming months it will always mark the end date.

matuzalemsteles:
Hey @drakonux could you take a look at the usage experience to select range? mainly to change the start range when there is already a defined range end, at first I thought this was a little strange, I am not sure if this had been covered in the document but I realized that to select a start date forward I have to click on the date end to reallocate the start date and then select the end date.

https://deploy-preview-4012--next-storybook-clayui.netlify.app/?path=/story/components-claydatepicker--w-date-range

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,81 @@ import {Keys} from '@clayui/shared';
...
+						!hasEndDateSelected &&
+						isWithinInterval(date, {
+							end: endDate,
+							start: startDate,
+						}),
+					'c-selected c-selected-end':
+						hasEndDateSelected && !hasStartDateSelected,
+					'c-selected c-selected-start':
+						hasStartDateSelected && !hasEndDateSelected,
+				}
```
bryceosterhaus, line 75: Can we make it so only one class is declared per key? I think this would work the same...

```js
{
	'c-selected':
		hasStartDateSelected ||
		hasEndDateSelected ||
		isWithinInterval(date, {
			end: endDate,
			start: startDate,
		}),
	'c-selected-end': hasEndDateSelected && !hasStartDateSelected,
	'c-selected-start': hasStartDateSelected && !hasEndDateSelected,
}
```

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
```diff
@@ -11,6 +11,7 @@ import React from 'react';
 const spritemap = 'icons.svg';
```
bryceosterhaus, line 11: It worries me a bit that we had to update these tests. Was this done for locale or is there a particular reason we had to do so?

bryceosterhaus (commented):
I tried to get the full context of this component but I am sure I missed something. I only had a few small things I noted.

The ux of the range functionality seemed really clean to me, great job!

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,81 @@ import {Keys} from '@clayui/shared';
...
 
-import {IDay, formatDate, setDate} from './Helpers';
+import {IDay, setDate} from './Helpers';
 
 interface IProps {
 	day: IDay;
-	daySelected: Date;
+	daysSelected: readonly [Date, Date];
 	disabled?: boolean;
+	range?: boolean;
```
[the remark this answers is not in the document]
matuzalemsteles: I think not we have compatibility issues here because `daysSelected` is an internal state of the component, so we have control over that. Well, what I thought about this is that `daysSelected` behaves like a tuple but when it is not a range, we will have only `[Date]` and when it is range `[Date, Date]`.

Currently, this code is not aligned in this sense, it only considers that it will always be a tuple with two dates, so when it is not a range, the dates will be the same.

packages/clay-date-picker/src/DayNumber.tsx
matuzalemsteles: I think yes, I don't remember why it didn't go on with that.

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
matuzalemsteles: As the spritemap will have no effect in tests and the components need this information, so we pass on any value.

bryceosterhaus: @matuzalemsteles sorry, I meant for this whole file. I just left the comment at the first line I saw in the file

packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -7,43 +7,81 @@ import {Keys} from '@clayui/shared';
...
 	onClick: (date: Date) => void;
 }
 
+interface IInterval {
+	start: Date;
+	end: Date;
+}
+
+function isWithinInterval(dirtyDate: Date, interval: IInterval) {
```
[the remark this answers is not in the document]
matuzalemsteles: Do you mean to call the first `date` of the `interval` instead of having `dirtyDate`? I think this will break the functionality because we want to know if `date` is within the range, maybe the name of the function is not helping?

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
matuzalemsteles: Oh okay 😅. Are you trying to tell if these changes in testing make sense? I think the last time Diego talked about it was because he made the changes to the `aria-label` of `<DayNumber />` to be more accessible.

packages/clay-date-picker/src/DayNumber.tsx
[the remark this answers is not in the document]
bryceosterhaus: I am just saying that the name of the argument to the function should be "date" instead of "dirtyDate". The functionality of the util is fine, just thought the naming of the argument could be simplified.

Using "dirtyDate" doesn't give much value since there is no "cleanDate" to compare it against. It also really doesn't matter what kind of date is passed to the function, it is simply checking if its within a given interval.

packages/clay-date-picker/src/__tests__/IncrementalInteractions.tsx
bryceosterhaus: ah I see. Yeah I was just making sure that we aren't introducing any breaking changes for people. Typically I check out tests because if we require any changes to how we test, it sometimes indicates that users will also have to make a change

matuzalemsteles: Yeah I totally agree

matuzalemsteles:
Resent with some fixes #4019

PR #4019, opened 2021-04-19
feat(@clayui/date-picker): Adds ClayDatePicker Range variation

matuzalemsteles:
I'm just adding a few more commits to correct some things, for context see the original PR: #4012

We still have some problems that we need to fix:

- We have a CSS bug where the `active` state of the date is disabled it does not appear, maybe we should just show color of the `active` with less opacity instead of removing it? cc @pat270 @drakonux 
- ~~The experience for changing the start range when an end date is already selected seems a little confusing https://github.com/liferay/clay/pull/4012#issuecomment-822558650~~

Despite having some small details to fix I think that both do not prevent this PR from entering the master.

### DIFF
--- a/packages/clay-date-picker/README.mdx
+++ b/packages/clay-date-picker/README.mdx
@@ -6,27 +6,31 @@ packageNpm: '@clayui/date-picker'
 ---
 
 import {
-	DatePickerWithState,
-	DatePickerWithTime,
 	DatePickerCustomFooter,
 	DatePickerLocale,
+	DatePickerWithRange,
+	DatePickerWithState,
+	DatePickerWithTime,
 } from '$packages/clay-date-picker/docs/index';
 
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
 -   [Time](#time)
+-   [Range](#range)
 -   [Internationalization](#internationalization)
 -   [Custom Footer](#custom-footer)
+-   [Programatically Expand Dropdown](#programatically-expand-dropdown)
+-   [Note about Moment.js](#note-about-moment.js)
 
 </div>
 </div>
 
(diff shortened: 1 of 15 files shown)


matuzalemsteles:
> The experience for changing the start range when an end date is already selected seems a little confusing #4012

I was revisiting the Date Picker documents to check the requirements and noticed that we missed some use cases and that it will mainly correct some confusion when selecting the start date.

- A range is selected:  Any date selection removes the current range and sets the start date with the new date chosen
- If the end date belongs to a different month than the start date when it is selected, the calendar view comes back to the start date month.

Some bugs also found in this implementation:

- Clicking on the dot does not add a normalized date which breaks some logic to add the `c-selected` class

I'm going to work on that now so that we can have it available in tomorrow's release.

pat270:
I spoke with @diegonvs about this over slack and thought it was odd that we disabled selectable dates. In any case, I can make c-selected work with disabled if that is something we don't want to look into right now.

matuzalemsteles:
Hey @pat270, I think the `c-selected` class is working well in the disabled state, only the `active` class is not, so I think we can just try to leave `active` with a low opacity instead of removing the entire background.

Screen Shot 2021-04-20 at 13 52 55

The 29th is the end date.

matuzalemsteles:
I think this is now ready to go ahead or go into a next review, I had to make some changes to correct the problems listed in my previous comment.

I also added some test cases and fixed the snapshots that were merged breaking a behavior (see commit message ad859f7688cea1752796e3a459633baad06c5ff5 and a36e9e13df2eacc291bf15dc93e109d17b6996d7), added tests to cover the main use cases for the Range and documentation on clayui.com.

Tomorrow it will be a holiday in Brazil and I will be in PTO and I will be back on Monday, @bryceosterhaus you can look at this if you need any changes, it would be good if we can send it tomorrow for the release but if needs big changes we can move it for the next release.

bryceosterhaus:
I don't see any major changes that are needed. Going to merge for release and we can follow up with finer improvements if needed

Issue #4017, opened 2021-04-16
isLoading prop on multi select component


FabioDiegoMastrorilli:
In a new implementation we would like to use a multi select as an autocomplete.

During the fetch we would like to notify the user that something is going on by displaying a loading indicator. 
The visual feedback should be the same as a normal autocomplete (a loading indicator on the right)

My proposals are 2:

1. adding a prop which might be called `isLoading` to a multi select
2. allowing the autocomplete component to show multiple values by passing them via prop (transforming it in a multi select)

matuzalemsteles:
Hey @FabioDiegoMastrorilli, if I understand correctly you want to do something similar like the example of the autocomplete https://storybook.clayui.com/?path=/story/components-clayautocomplete--w-async-data?

> allowing the autocomplete component to show multiple values by passing them via prop (transforming it in a multi select)

This was not very clear to me, could you explain it better?

FabioDiegoMastrorilli:
Yes, you got it right! :) 
I'll try to explain it better by sharing some designs:

What you see here can be created by using an Autocomplete component so when the user starts typing something, a spinner will be visible so he will understand that there is some data fetching going on.
![image](https://user-images.githubusercontent.com/44569796/115212590-f8badd80-a100-11eb-9975-aa564a1218fe.png)

The difference is that, when you click on a result, the selected value will be visible in the input (the way it looks in a multi select components):
![image](https://user-images.githubusercontent.com/44569796/115212847-3cade280-a101-11eb-963c-4b12d7bdc918.png)

Due to these requirements, in order to achieve these UIs, we can work on adding something on the multiselect component (to let it show the spinner - the quickest solution) **OR** on the Autocomplete (by passing a list of values so it can show them as labels within the input).

Maybe the `Autocomplete` component can wrap a `Multiselect` so we call it by using `Autocomplete.Multiselect`.

Hope I explained myself better :)

bryceosterhaus:
Makes sense to me that we should provide this. I think the best route would be to provide an async mechanism for `Multiselect`'s `sourceItems` prop. This way this component both handles hard-coded data and async data. If its async data, we would automatically handle the loading indicator for the user.

I can investigate it this week

matuzalemsteles:
Hmm yeah, this can be interesting @bryceosterhaus instead of providing an `isLoading` API if `sourceItems` is async we can do this internally.

[...]

PR #4029, opened 2021-04-26
feat(@clayui/multi-select): add async functionality for source items

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/4017

@FabioDiegoMastrorilli can you verify the "async" example in the storybook and see if thats what you would expect?

### DIFF
--- a/packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
@@ -140,8 +140,8 @@ exports[`ClayMultiSelect renders a custom menu 1`] = `
   <div>
     <div>
       <div
-        class="dropdown-menu autocomplete-dropdown-menu show"
-        style="max-width: none; left: -999px; top: -995px; width: 0px;"
+        class="dropdown-menu autocomplete-dropdown-menu"
+        style="max-width: none; left: -999px; top: -995px;"
       >
         <ul
           class="list-unstyled"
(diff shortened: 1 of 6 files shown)


packages/clay-multi-select/stories/index.tsx
```diff
@@ -235,6 +235,30 @@ storiesOf('Components|ClayMultiSelect', module)
...
+			<ClayMultiSelect
+				inputName="myInput"
+				inputValue={value}
+				items={items}
+				onChange={setValue}
+				onItemsChange={setItems}
+				sourceItems={asyncData}
+				spritemap={spritemap}
+			/>
+		);
```
matuzalemsteles, line 260: I had this in mind for `sourceItems` to accept only one `Promise` and we hope it resolves to get the data I think it would be easier than accepting a function that returns a `Promise`, something like this:

```jsx
<ClayMultiSelect
	inputName="myInput"
	inputValue={value}
	items={items}
	onChange={setValue}
	onItemsChange={setItems}
	sourceItems={new Promise(...)}
	spritemap={spritemap}
/>
```

bryceosterhaus: @matuzalemsteles passing a Promise as a prop or argument seems odd to me for some reason. I feel like I am used to seeing functions that return a promise in similar APIs, I recall . This way we can also pass the query directly to the function. 

If we had a util function for calling the api it might look something like this, which seems strange to invoke on render rather than on the value change. At this level, you might as well hoist the async logic out and just us an `isLoading` prop.
```jsx
function myapp() {
  const itemFetcher = () => {
    // some business logic for wiring up call.
    return fetch(`/some/api?q=${value}`);
  };

  return (
    <ClayMultiSelect
      inputValue={value}
      items={items}
      onChange={setValue}
      onItemsChange={setItems}
      sourceItems={itemFetcher()}
    />
  );
}
```

bryceosterhaus: I could be looking at this wrong though. Can you think of other examples where we pass a promise directly?

wincent: One argument in favor of a function is that detecting it is a tiny bit easier (eg. `typeof blah === 'function'`) compared with identifying a "thenable".

But with promises, they are generally pretty annoying to work with because they can't be cancelled, and they are opaque (ie. you can't ask them if they are resolved or not — this is why there are so many `PublicPromise` abstractions out there, which are wrappers around promises that add methods for introspecting their internal state).

Having said that, a function that returns a promise isn't any better in this regard; you still get a promise in the end, with all of its downsides.

One thing I notice about your implementation, @bryceosterhaus, is that you have a `then()` but no `catch()`; let's that add that to the list of things to hate about promises — that it is far too easy to overlook error-handling in them.

Just thinking out loud here, but a "best of both worlds" approach might be to start using `async`/`await` in Clay, although it does have obvious costs (eg. because Clay is used on older branches, we can't assume native browser support and would have to transpile). That way you could accept an async function, or a function that returns a promise, from the calling code, an inside Clay, we would at least have the somewhat more pleasing:

```ts
try {
	const asyncItems = await sourceItems(inputValue);

	setInternalSourceItems(asyncItems);

	filterAndSet(asyncItems);
}
catch {
	// Do something graceful; I don't actually know what...
}
finally {
	setIsLoading(false);
}
```

Ah, and one final thought... all those `set` calls run the risk of firing after the component unmounts, triggering an error. Any async function in a React components needs to be protected by some kind of "is mounted" check, like we do in DXP with [the `useIsMounted()` hook](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-react-web/src/main/resources/META-INF/resources/js/hooks/useIsMounted.ts); not sure if we already have something like that somewhere in Clay.

Anyway, food for thought.

matuzalemsteles: Well, the idea of receiving only `Promise` instead of invoking a function that returns a `Promise` is that we don’t have the responsibility to call the function and we leave the responsibility to the developer who can call the function at any time he wants so he can add his own rules although he can do the same in this implementation we limit this so that `sourceItems` will be called whenever there is a change in `inputValue`, for example, the use of `useResource` he deals with the responsibility to revalidate the request when a `variable` changes, that is, we allow this to be composed by the developer and he can arrive at different business rules, where he can revalidate the request to any change in `inputValue` or just invoke just once.

The [Autocomplete example in the Storybook](https://storybook.clayui.com/?path=/story/components-clayautocomplete--w-async-data) shows some of these benefits, we add loading only when a request happens, when the user tries to search for a request that is in the cache we will return from the cache, this brings some beneficious, which we would not be able to do here and if to support we would have to do more things. I'm just thinking of removing the responsibility of this from the component.

> Just thinking out loud here, but a "best of both worlds" approach might be to start using async/await in Clay

Yeah, I think it would be worth trying to start using `async` / `await` but it really brings the responsibility for whoever uses it, dealing with transpilation.

> Any async function in a React components needs to be protected by some kind of "is mounted" check, like we do in DXP with the useIsMounted() hook...

Yeah, I think it's super valid to add an implementation of this to `shared`, we usually don't deal with asynchronous operations on other components except the implementation of `useResource` that has this responsibility, it would be good to just keep there...

FabioDiegoMastrorilli: Thank you very much guys!
I just visually link the loading spinner to the data fetch action so my assumption is that the code which manages both the things should be in the same place.

My only (really small point) is that something like this: 
```
getData(query).then(updateSourceItems()) 
```
**looks** a bit more predictable than:

```
updateSourceItems(getData(query))
```
since dealing with unsolved promises It's pretty rare to me: where I don't use a `.then` they are wrapped in some `Promise` method like all, allSettled, etc...

That being said, this PR solves the problem I asked to be solved so to me is totally ok :) 
The `isLoading` would just make me slightly happier but for that I can get a dog :P

matuzalemsteles: Well I'm almost accepting the idea of just having `isLoading` or another name to show `LoadingIndicator`  🙈

bryceosterhaus:
So I was looking at this more and I think I am more convinced that our current API is pretty flawed with how we handle so much internally. Like @matuzalemsteles said, we should be deferring the business logic of filtering items and the async nature to the parent component. Right now we are using `useLayoutEffect` to filter and run the async data source, which should have been an initial indicator that keeping this logic internally isn't the best idea.

I took a look at DXP and only one use case actually uses the `filter` prop, so I am thinking we just remove that prop and re-write a bit so that the consumer has more control of the source items. This would mean we go with the original idea of a "isLoading" prop, and I like that the more I think about it since it gives the user more control of when they want to label it as loading. There could be a use case where the loading indicator needs to be used before the data can even be fetched.

See my latest commit for this direction.

Technically this is a breaking change, but we can just manually update DXP ourselves and can provide details on how to easily port between using filter and just using onChange

matuzalemsteles:
Yeah, I agree, it's very difficult to balance this... I think it could be the most correct way here but the bad thing about removing this filter behavior now from `sourceItems` is that by default we have a filter defined in the `filter` API, then it may be that the components that use `ClayMultiSelect` use the default filter.

> I'm not sure why we even added the filter prop in the first place, all we needed the parent component to do is something like this.

Probably because we wanted to use `fuzzy.match` for filtering but I think that native `.match` can already help us here because we use `fuzzy.match` internally in `ClayAutocomplete.Item` to mark the characters that satisfy the search.

bryceosterhaus:
@matuzalemsteles for capability with `fuzzy` matching, we could export a util function that they can use for their sourceItems.

I can send an update to portal to show what it would look like to update DXP

matuzalemsteles:
@bryceosterhaus move on, I think it might be a good idea 🤔.

packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -568,14 +568,21 @@ exports[`Interactions add non-sequential text should filter the items when it is
...
+          </li>
         </ul>
       </div>
     </div>
   </div>
 </body>
 `;
 
-exports[`Interactions adding text doesn not filter the items with a custom function 1`] = `
+exports[`Interactions adding text doesn't not filter the items with a custom function 1`] = `
```
matuzalemsteles, line 585: @bryceosterhaus perhaps tests like this may no longer make sense since `sourceItems` no longer filters. What do you think?

packages/clay-multi-select/src/index.tsx
```diff
@@ -347,6 +334,10 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 							onFocus={(e) => {
 								onFocus(e);
 								setIsFocused(true);
+
+								if (sourceItems.length !== 0) {
+									setActive(true);
+								}
```
matuzalemsteles, line 340: Just out of curiosity and to understand this here. Since we are already doing the same in `useEffect` when it is focused, shouldn't this be more necessary here?

packages/clay-multi-select/src/index.tsx
```diff
@@ -379,14 +376,13 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 							</ClayInput.GroupItem>
 						)}
 
-					{sourceItems.length > 0 && (
+					{!isLoading && sourceItems.length > 0 && (
```
matuzalemsteles, line 379: I think we shouldn't stop showing `DropDown` when doing some asynchronous operation because we may want to show the data that was in cache but we are revalidating, `LoadingIndicator` should already be the essential information to show the user that a request is being made and if the developer wants to follow this behavior, he is free to define the `sourceItems` for an empty `Array` to arrive at the same behavior here. What do you think?

In theory, we should follow this specification https://docs.google.com/document/d/1LJ8yiO8V5mSEGOjaYzHd5Nex_De_8fMRQAEKQoHMiiM/edit, to show `Loading...` text on DropDown, but I am not sure if this was included in the official documentation or it was just a proposal but I think it was also a proposal we shouldn't fix that here.

packages/clay-multi-select/src/__tests__/index.tsx
```diff
@@ -221,17 +224,6 @@ describe('Interactions', () => {
...
-
-		fireEvent.change(
-			document.querySelectorAll(
-				'input[type=text]'
-			)[0] as HTMLInputElement,
-			{target: {value: 'bar'}}
-		);
-
-		expect(document.querySelectorAll('.dropdown-item').length).toEqual(1);
-
```
matuzalemsteles, line 234: What was the reason for removing this? Just to understand 😄.

matuzalemsteles (commented):
@bryceosterhaus overall it looks good, I added some comments. Only one thing I worry about is that this should be a breaking change but we can get away from the rule a little bit here and make changes to the DXP, but it would be good to avoid and perhaps leave changes like this for a future major version.

bryceosterhaus:
@matuzalemsteles I updated the tests structure a bit and also applied some of the changes you noted. I'm also not stoked that its technically a breaking change, but I also think its the right direction to go. One positive to note is at least the break is fixable by just adding our utility filter. I'm pretty sure DXP is our only consumer of this component as well, so I'm not too concerned about other apps breaking.

packages/clay-multi-select/src/__tests__/index.tsx
```diff
@@ -252,19 +244,8 @@ describe('Interactions', () => {
...
 
-		fireEvent.change(
-			document.querySelectorAll(
-				'input[type=text]'
-			)[0] as HTMLInputElement,
-			{target: {value: 'bar'}}
-		);
-
-		expect(document.querySelectorAll('.dropdown-item').length).toEqual(1);
-
```
bryceosterhaus, line 267: These were removed because they had nothing to do with the purpose of the test, making sure the onItemsChange was called.

matuzalemsteles: Makes sense, thanks

matuzalemsteles (approved):
LGTM! Yeah, I agree! I think we can handle this change well.

Before sending forward just a look at the Storybook mistakes, there appear to be some type errors in the new Storybook story.

bryceosterhaus:
@matuzalemsteles I updated some types to help address the issue. I also didn't realize you could `export type ...` until recently so I went ahead and added that to suppress some console warnings for our storybook build

matuzalemsteles:
Great, feel free to merge!

PR #4030, opened 2021-04-26
feat(@clayui/tooltip): change the Provider to singleton and adds the listeners to the document.body

### LINKED ISSUE #4013: [Tooltip Provider] Improves the event track for applications using React.Portal
After we had problems (https://github.com/liferay-frontend/liferay-portal/pull/946) with using the Tooltip Provider within DXP, we had two main problems that prevent the use of Tooltip Provider in DXP.

1. Components within the application tree using `React.Portal` that renders on the Body
2. Elements outside the React tree

We were unable to solve these two current problems because internally we added two mouse event props in `children`, this also creates a problem if the props are not allocated in the correct element which can further aggravate the problem.

The use of `React.Portal` complicates the problem because we want to take advantage of React's synthetic event instead of dealing with it manually. Once the element is rendered in the body, the `onMouseOut` or `onMouseOver` event will not be triggered because it is not in the same tree. To detect elements outside the React tree we need to add events manually and in the body to be able to detect all elements independently of the tree.

The problem of going in the direction of manually adding the event to the body is that in DXP we can have several Tooltip Providers being rendered on the same page, this also makes it a Singleton where we would need to start or add the event only once regardless of the number of providers that are rendered on the page so that there is no conflict.

matuzalemsteles:
Fixes #4013

Well, the change here is quite simple, we stopped using the synthetic events of the React that were added in `children` to add events to `document.body` to cover the use cases with `React.Portal` and as a consequence to avoid conflicts with other providers on the same page which is quite common in DXP, making the Provider a singleton.

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -206,7 +206,11 @@ const TooltipProvider: React.FunctionComponent<{
 				(window as any)[SYMBOL_TOOLTIP] = false;
 
 				document.body.removeEventListener('mouseout', handleHide, true);
-				document.body.removeEventListener('mouseover', handleShow, true);
+				document.body.removeEventListener(
+					'mouseover',
+					handleShow,
+					true
+				);
 			};
 		}
 	}, [handleHide]);
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -5,8 +5,7 @@
 
 import {ClayPortal, Keys} from '@clayui/shared';
 import domAlign from 'dom-align';
-import React, {useCallback} from 'react';
-import warning from 'warning';
+import React, {useCallback, useEffect} from 'react';
 
 import ClayTooltip from './Tooltip';
 
@@ -43,6 +42,8 @@ const ALIGNMENTS_INVERSE_MAP = {
 	trbr: 'bottom-right',
 } as const;
 
+const SYMBOL_TOOLTIP = Symbol('clay.tooltip.attached');
+
 interface IState {
 	align?: typeof ALIGNMENTS[number];
 	message?: string;
@@ -76,42 +77,6 @@ const reducer = (state: IState, {type, ...payload}: IAction): IState => {
 	}
 };
 
-function matches(
-	element: HTMLElement & {
-		msMatchesSelector?: HTMLElement['matches'];
-	},
-	selectorString: string
-) {
-	if (element.matches) {
-		return element.matches(selectorString);
-	} else if (element.msMatchesSelector) {
-		return element.msMatchesSelector(selectorString);
-	} else if (element.webkitMatchesSelector) {
-		return element.webkitMatchesSelector(selectorString);
-	} else {
-		return false;
-	}
-}
-
-function closestAncestor(node: HTMLElement, s: string) {
-	const el = node;
-	let ancestor: HTMLElement | null = node;
-
-	if (!document.documentElement.contains(el)) {
-		return null;
-	}
-
-	do {
-		if (matches(ancestor, s)) {
-			return ancestor;
-		}
-
-		ancestor = ancestor.parentElement;
-	} while (ancestor !== null);
-
-	return null;
-}
-
 type TContentRenderer = (props: {
 	targetNode?: HTMLElement | null;
 	title: string;
@@ -119,7 +84,6 @@ type TContentRenderer = (props: {
 
 const TooltipProvider: React.FunctionComponent<{
 	autoAlign?: boolean;
-	children: React.ReactElement;
 	contentRenderer?: TContentRenderer;
 	delay?: number;
 }> = ({
@@ -201,7 +165,7 @@ const TooltipProvider: React.FunctionComponent<{
 		}
 	}, []);
 
-	React.useEffect(() => {
+	useEffect(() => {
 		if (
 			titleNodeRef.current &&
 			(tooltipRef as React.RefObject<HTMLDivElement>).current
@@ -231,10 +195,21 @@ const TooltipProvider: React.FunctionComponent<{
 		}
 	}, [align, show]);
 
-	warning(
-		children.type !== React.Fragment,
-		'<TooltipProvider />: React Fragment is not allowed as a child to TooltipProvider. Child must be a single HTML element.'
-	);
+	useEffect(() => {
+		if (!(window as any)[SYMBOL_TOOLTIP]) {
+			(window as any)[SYMBOL_TOOLTIP] = true;
+
+			document.body.addEventListener('mouseout', handleHide, true);
+			document.body.addEventListener('mouseover', handleShow, true);
+
+			return () => {
+				(window as any)[SYMBOL_TOOLTIP] = false;
+
+				document.body.removeEventListener('mouseout', handleHide, true);
+				document.body.removeEventListener('mouseover', handleShow, true);
+			};
+		}
+	}, [handleHide]);
 
 	return (
 		<>
@@ -249,13 +224,45 @@ const TooltipProvider: React.FunctionComponent<{
 				</ClayPortal>
 			)}
 
-			{React.cloneElement(children, {
-				...children.props,
-				onMouseOut: handleHide,
-				onMouseOver: handleShow,
-			})}
+			{children}
 		</>
 	);
 };
+
+function matches(
+	element: HTMLElement & {
+		msMatchesSelector?: HTMLElement['matches'];
+	},
+	selectorString: string
+) {
+	if (element.matches) {
+		return element.matches(selectorString);
+	} else if (element.msMatchesSelector) {
+		return element.msMatchesSelector(selectorString);
+	} else if (element.webkitMatchesSelector) {
+		return element.webkitMatchesSelector(selectorString);
+	} else {
+		return false;
+	}
+}
+
+function closestAncestor(node: HTMLElement, s: string) {
+	const el = node;
+	let ancestor: HTMLElement | null = node;
+
+	if (!document.documentElement.contains(el)) {
+		return null;
+	}
+
+	do {
+		if (matches(ancestor, s)) {
+			return ancestor;
+		}
+
+		ancestor = ancestor.parentElement;
+	} while (ancestor !== null);
+
+	return null;
+}
 
 export default TooltipProvider;
--- a/packages/clay-tooltip/stories/index.tsx
+++ b/packages/clay-tooltip/stories/index.tsx
@@ -44,45 +44,43 @@ storiesOf('Components|ClayTooltip', module)
 			<button title="Default">{'No Tooltip'}</button>
 
 			<ClayTooltipProvider>
-				<div>
-					<button title="Default">{'Default'}</button>
+				<button title="Default">{'Default'}</button>
 
-					{/* Note that the newline has to be a string within braces. */}
-					<button title={'Line1\nLine2'}>
-						{'With a line Break'}
-					</button>
+				{/* Note that the newline has to be a string within braces. */}
+				<button title={'Line1\nLine2'}>{'With a line Break'}</button>
 
-					<button data-tooltip-align="left" title="Left">
-						{'Left'}
-					</button>
+				<button data-tooltip-align="left" title="Left">
+					{'Left'}
+				</button>
 
-					<button data-tooltip-align="right" title="Right">
-						{'Right'}
-					</button>
+				<button data-tooltip-align="right" title="Right">
+					{'Right'}
+				</button>
 
-					<button data-tooltip-align="top" title="Top">
-						{'Top'}
-					</button>
+				<button data-tooltip-align="top" title="Top">
+					{'Top'}
+				</button>
+			</ClayTooltipProvider>
 
-					<button
-						data-tooltip-align="bottom"
-						data-tooltip-delay="0"
-						title="Bottom"
-					>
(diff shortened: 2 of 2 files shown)


packages/clay-tooltip/src/TooltipProvider.tsx
```diff
@@ -231,10 +195,25 @@ const TooltipProvider: React.FunctionComponent<{
 		}
 	}, [align, show]);
 
-	warning(
-		children.type !== React.Fragment,
-		'<TooltipProvider />: React Fragment is not allowed as a child to TooltipProvider. Child must be a single HTML element.'
-	);
+	useEffect(() => {
+		if (!(window as any)[SYMBOL_TOOLTIP]) {
+			(window as any)[SYMBOL_TOOLTIP] = true;
```
bryceosterhaus, line 200: These `any`s scare me a bit. I'm also not sure the best way to handle global window values....

packages/clay-tooltip/stories/index.tsx
```diff
@@ -44,45 +44,43 @@ storiesOf('Components|ClayTooltip', module)
 			<button title="Default">{'No Tooltip'}</button>
```
bryceosterhaus, line 44: This now shows a tooltip in the Story since it is global.

Screen Shot 2021-04-26 at 3 37 04 PM

bryceosterhaus (commented):
By making this a singleton, we no longer allow overriding the tooltip right? This would also remove the capability of scoping a tooltip to a specific set of children. [For example, this story doesn't work as expected.](https://deploy-preview-4030--next-storybook-clayui.netlify.app/?path=/story/components-claytooltip--claytooltipprovider-w-custom-jsx)

Also, this would be a significant breaking change right? Might need to think about our versioning here or providing both APIs to avoid breaking.

Instead of making ClayTooltip a global singleton in the Clay component, would it actually make more sense to just remove the Provider portion on the Clay side and implement it on the DXP side specifically?

wincent:
> we stopped using the synthetic events of the React

Tiny nit here: the fact that the events are [synthetic](https://reactjs.org/docs/events.html) isn't actually interesting at all — the critical thing is that React < v17 uses a delegated event listener on the document, while [React >= v17 uses a listener on the container where the root component is mounted](https://reactjs.org/blog/2020/08/10/react-v17-rc.html).

> Might need to think about our versioning here or providing both APIs to avoid breaking.

Even though it sucks to maintain two versions (or to maintain a more complex backwards-compatible version), I would go to extreme lengths to avoid a major version bump in any Clay package. Happy to write a Russian novel 🇷🇺  about the reasons why if anyone wants to know 😂.

packages/clay-tooltip/src/TooltipProvider.tsx
wincent: You can augment the `Window` interface as described in [some of the answers here](https://stackoverflow.com/questions/12709074/how-do-you-explicitly-set-a-new-property-on-window-in-typescript) but having done this in several projects and several versions of TS, I've always found it to be a tremendous pain, incredibly delicate, and sensitive to versions, config settings, project structure etc. Basically, it's really hard, and I suspect (wild speculation) that one of the reasons they don't particularly prioritize making it easy is because they don't want to encourage people spewing stuff into the global scope. Just a theory though.

Here's my most recent example of fighting with this, where I [add `Liferay.State`](https://github.com/liferay/liferay-portal/blob/4b4ff53717645e7bf4c4b373b5cb5d218fb1207a/modules/apps/frontend-js/frontend-js-state-web/src/main/resources/META-INF/resources/State.ts#L510-L524) to the global `window` object in DXP. @bryceosterhaus, you are probably going to have to take this kind of thing into consideration when you do [LPS-131257](https://issues.liferay.com/browse/LPS-131257).

matuzalemsteles: Yeah, I didn't even get to extend the `Window` interface, it's what I normally do in a NodeJS environment for `process.env`. But we can try to do the same thing @wincent did for State. Probably just adding to the same file or creating a `window.d.ts`?

packages/clay-tooltip/stories/index.tsx
matuzalemsteles: Yeah, this is the limiter or regression that we will have using this implementation, we can try to reverse this with some strategies.

matuzalemsteles:
@bryceosterhaus yes I agree, making a singleton will have some unwanted behaviors, he will no longer be looking at a specific tree this is what I like most but we have problems with `React.Portal` and also with the props in `children` that may not arrive a necessary element.

Keeping a version control maybe we can even have to evaluate the cost of this but also if we don’t go with that, that would be a breaking change and a major version bump.

Well I think we can try to leave some things on the DXP side and some here to maintain compatibility or just to make `TooltipProvider` make sense as well. For example:

- By default, `TooltipProvider` adds events to `children`
- We added a new API to pass the element or reference of the element where we must add the event

I think we can alleviate the side effects of making a singleton, because if you follow this by default, actually using the `contentRenderer` API and others may not make sense, because when there is a global `TooltipProvider` it will overwrite any other.

> Even though it sucks to maintain two versions (or to maintain a more complex backwards-compatible version), I would go to extreme lengths to avoid a major version bump in any Clay package. Happy to write a Russian novel 🇷🇺 about the reasons why if anyone wants to know 😂.

haha please, I would like to read 😁

wincent:
> > Even though it sucks to maintain two versions (or to maintain a more complex backwards-compatible version), I would go to extreme lengths to avoid a major version bump in any Clay package. Happy to write a Russian novel 🇷🇺 about the reasons why if anyone wants to know 😂.
>
> haha please, I would like to read 😁

Oh no, now I've forgotten what I was gonna say... 😂

Something along the lines of:

- **One general reason:** Overall, the kind of customers who pay for DXP don't like having to rebuild there stuff just because we decided to break it. In order to justify breakage, the upsides have to be overwhelming. That's why — if you look across the DXP codebase as it exists now and also in the commit history — you'll see we often go to great lengths to keep things working in a backwards compatible way, even if it makes it more expensive for us. That is, we'd rather absorb the costs than pass them on to customers. The exceptions are much rarer in Liferay than in typical "non-enterprise" products; we only break things when there is a clear case to be made that the upside value to the broader community considerably outweighs the downside costs to the customer. To be used seriously at scale, Clay needs to be boring in the same way that React is boring: v15 came out in early April 2016; v16 came out in late September 2017, 1.5 years later; v17 showed the slowing pace, coming out in October 2020, more than 3 years later, and it [didn't even have any actually significant breaking changes](https://reactjs.org/blog/2020/08/10/react-v17-rc.html)! Facebook has many tens of thousands of React components in their monorepo (not an exaggeration) — they simply can't afford to waste their time scurry around breaking "userland" code. There are more important things to do. And breaking changes are not just extra work for customers; they're extra work for us too — divergent behavior makes it harder to field support questions (what version is being used? in what combination with other packages?) and increases the burden of maintaining the documentation (not only do we have to write the docs for each version, but we have to implement complex support for allowing users to switch between versions, make it easy for them to find the right version, search within the right version etc; or alternatively, we do nothing and let our documentation suck). Basically, the juice isn't worth the squeeze.
- **One specific reason:** Nothing is happening literally right now, but Brian has been pushing us (Frontend Infra and Dev Tools) to make it easier for customers to start and continue using our libraries. One of the options that is on the table is a Clay "umbrella" package like we had in the v2 days that would mean you could add _one_ dependency to your `package.json` and do all of your imports from that, instead of remembering and juggling a big set of different packages and version numbers. Obviously, to make that work we have to make sure our buildtime and runtime tooling still delivers the smallest possible payload to the client, (it's not actually desirable to _literally_ have "all of Clay" come down to the browser just because somebody wants to stick a button in a window). But the point is, if we have an umbrella package, we don't want individual packages introducing breaking changes that would force the entire umbrella to get a major version bump.

Having said all that, we do have _some_ options available to us if we really want to make breaking changes. For example, we can prepare easy-to-use code-mods with breaking releases that users can run to update their codebases. This is one of the tricks that FB used rather a lot to deal with the whole "tens of thousands of usages" thing, and also to make their open source projects good choices for the outside world that wanted to build large things on top of them. But even with code-mods or brilliant docs, I still think we should go to great lengths to avoid breaking changes unless we're reeeeeally sure.

packages/clay-tooltip/src/TooltipProvider.tsx
wincent: So in `@liferay/frontend-js-state-web` I initially tried to declare all this as an ambient type in a separate file, with a view towards centralizing it in the future, but then the whole project references thing caused things to blow up, requiring the nonsense that I ended up stick down the bottom of that file. Because of the sensitivity to versions, config, and structure, there is no guarantee that a technique you see working in one place will work when transplanted to another. 😢

matuzalemsteles:
I totally agree @wincent, this is really something that we should weigh here. I think that taking this into account, we don't want to cause a major version now mainly because of a change in `TooltipProvider`. I think one thing to add about this is that also taking time on a version is that we let the library mature and that people can get used to the "way it is done" (I was not able to express myself very well here but it is more or less that 🙂).

So, I think we can keep the default behavior of `TooltipProvider` and provide an API to change the source of the listener, and maybe we can keep that as part of that API or create a new one to make it a singleton. Does it make sense?

Something like this:

1. Adds the `on` API makes it a singleton and adds the listener to `document.body`
```jsx
<ClayTooltipProvider on={document.body}>
  {...}
</ClayTooltipProvider>
```

2. Add the `on` API and it is necessary to make it explicit that it will be a singleton
```jsx
<ClayTooltipProvider on={document.body} singleton>
  {...}
</ClayTooltipProvider>
```

I have a preference for option 1, because in theory when you want to specify the element is when you want to control outside the React tree when you don't want that, you probably don't need to specify `on`.

bryceosterhaus:
> I would go to extreme lengths to avoid a major version bump in any Clay package.

I fully agree, I brought up the major version bump as more of an indicator that our changes seem a bit drastic to me. 

In taking a step back, the problem we are trying to address is very DXP specific where we want a single instance of tooltip to work across multiple react trees, as well as JSPs and anything server rendered. Currently, `ClayTooltipProvider` works pretty well in the case of a single React tree and your classic monolithic React app. I think instead of modifying our Provider to work across multiple react trees and non-js generated html, we move this sort of logic to DXP. I see it similar to our global state manager in DXP, it didn't make sense to create a new "state management" module here in Clay because it was a very DXP specific use case. We even have a tooltip specific module in DXP where we could add this to (`frontend-js-tooltip-support-web`). I think all we would need to do is update the delegate's selector logic to match `[title]` and `[data-title]`, I believe this is how we use to manage all DXP tooltips in the past as well?

In order to provide support for overriding how tooltip renders, we could add some logic to `ClayTooltipProvider` that would add a class or data attribute to the element that triggers the tooltip. Maybe something like `data-tooltip-override` and we add that to our `frontend-js-tooltip-support-web` module, `[title]:not([data-tooltip-override])`.

This solution would provide both
- global tooltip manager for the majority of cases
- ClayTooltipProvider would have ability to "override" the global manager for a specific react tree

Am I missing something in this or an edge case that is causing us issues?

matuzalemsteles:
Hmm well, looking more closely at the `frontend-js-tooltip-support-web` it looks something similar to what we do with` ClayTooltipProvider` I'm not sure why we don't use it instead but I think we should just choose one, instead of keeping both running on the same page so they don't have conflicts.

@bryceosterhaus `<ClayTooltipProvider />` has some limitations:

- Does not work with Portals (Many problems appeared in DXP because many components within the application used `React.Portal` in the body, as well as our components)
- There is a problem that we add events to `children` if children are a React component, it can happen that they are not added correctly to the root of the application. I think @markocikos get some cases of that

The `ClayTooltipProvider` needs to attach some element, we even suggested the introduction of a `div` as a wrapper to avoid things like this but this brings up other problems that were discussed in the PR https://github.com/liferay-frontend/liferay-portal/pull/946

bryceosterhaus:
@matuzalemsteles I can try to reconcile DXP with ClayTooltipProvider a bit and see what might be done. One option to make the provider more robust for a react tree with portals would be to parse through the fiber (similar to focus scope).

If you aren't currently looking into it, let me dive into it tomorrow and see what I can do. I just was working on the tooltip-support-web module, so I have some familiarity with both.

matuzalemsteles:
@bryceosterhaus I'm just waiting for a common consensus to proceed with the solution but feel free to dive into that option, the more we have, the better to decide which direction to take. Thanks!

About fiber, I don't have an idea on how it could help us but I would be curious to see how it can help us 🙂.

bryceosterhaus:
@matuzalemsteles I tested ClayTooltipProvider with a ClayPortal and it seems to work fine in our storybook. Is there a specific DXP case you can link me to that is an issue?

matuzalemsteles:
@bryceosterhaus apparently, there are some problems in the Page Editor Sidebar and elsewhere, I think they were reported in this https://issues.liferay.com/browse/LPS-129309 ticket also in the comment.

bryceosterhaus:
@matuzalemsteles I'll check out those specific examples, my guess is that something in the react tree is behaving strange. Also, I saw [in the react docs](https://reactjs.org/docs/portals.html#event-bubbling-through-portals) that the mouseover even does bubble up through the react portal, so that shouldn't be an issue.

matuzalemsteles:
@bryceosterhaus that's great, weird that it's not working in these DXP cases. It can probably be related that the `props` are not being added in the correct location.

matuzalemsteles:
I will close this since this solution is no longer viable to solve this problem. For more context https://github.com/liferay-frontend/liferay-portal/pull/1016

PR #4031, opened 2021-04-27
refactor(@clayui/shared): don't coerce `contentEditable`

wincent:
Not sure whether this would actually cause a bug, so labeling this as a refactor commit.

`contentEditable` can be a string (eg. `'true'`)or a bool (or `undefined`), so we probably shouldn't coerce it (eg. `!!'false'` is `true`). Instead, explicitly check against `true` and `'true'`.

Originally started looking at this because the `!!` in [this PR](https://github.com/liferay/clay/pull/3942) looked a bit off.

### DIFF
--- a/packages/clay-shared/src/useFocusManagement.ts
+++ b/packages/clay-shared/src/useFocusManagement.ts
@@ -43,7 +43,8 @@ export function isFocusable({
 
 	if (
 		(tabIndex != null && tabIndex >= minTabIndex) ||
-		!!contentEditable === true
+		contentEditable === true ||
+		contentEditable === 'true'
 	) {
 		return true;
 	}


packages/clay-shared/src/useFocusManagement.ts
```diff
@@ -43,7 +43,7 @@ export function isFocusable({
 
 	if (
 		(tabIndex != null && tabIndex >= minTabIndex) ||
-		!!contentEditable === true
+		contentEditable === true
```
bryceosterhaus, line 46: I think contentEditable can also be a string (not saying the coercion makes sense with a string), so we might want to check for both the boolean and string value of true?
```suggestion
		contentEditable === 'true' || contentEditable === true
```

matuzalemsteles: Oh yeah, this is probably because we also get it now directly from the DOM.

wincent: > I think contentEditable can also be a string

Ha, good catch. I need to revise my operator precendence table. 😂

wincent:
Update and force-pushed a less-incorrect version. 😀

PR #4038, opened 2021-04-30
LEXI-1167: Add rotate icon

drakonux:
Due to LEXI-1167 the lima team has requested a new icon for the new image editor.


drakonux:
I tried @matuzalemsteles and @pat270 but as you can see this is not my area of expertise 😅. What I did wrong for second time? the branch merge?

drakonux:
Tremble developers here I am hahahaha 😅

matuzalemsteles:
hahah Thanks a lot @drakonux, don't worry about that, we can take care of it from now on.

pat270:
@drakonux I made a minor change to the svg file. Instead of the class `lexicon-icon-body` we should use `lexicon-icon-outline`. We stopped using `lexicon-icon-body` after we changed the way we designed icons in 2.x.

I also `cd clay/clayui.com;yarn develop;` and committed the generated files. It runs a task to copy svgs from clay-css to clayui.com.

drakonux:
> @drakonux I made a minor change to the svg file. Instead of the class `lexicon-icon-body` we should use `lexicon-icon-outline`. We stopped using `lexicon-icon-body` after we changed the way we designed icons in 2.x.

That's right. I forgot to change it because I was trying to fix my node problem 🤦 . 

However, I would like to ask you: so do we plan to not use lexicon-icon-body ever? for example the "body" of an icon as the icon guide shows? 

> I also `cd clay/clayui.com;yarn develop;` and committed the generated files. It runs a task to copy svgs from clay-css to clayui.com.

Ah, so I would have also to upload the generated files, gtk.

Thanks a lot @pat270 🤗

pat270:
@drakonux `lexicon-icon-body` is used for the background of an icon. All the icons coming in from the Lexicon Team recently have had no path for the background. Some icons that have them are `add-cell` and `adjust`. You can see at https://clayui.com/docs/components/icon.html. Not sure how useful it is, but gives you more options if you want to color them.

PR #4040, opened 2021-04-30
fix(@clayui/css): Progress Bar sets a `min-width` on `progress-group-…

### LINKED ISSUE #4024: ClayProgressBar takes dynamic space for the progress value

* ClayProgressBar demo: https://codesandbox.io/s/clayprogressbar-demo-z6x24

Screenshot 2021-04-26 at 16 15 56

Use the default API to create a progress bar with the Clay component ClayProgressBar (out-of-the-box behavior).

The bar takes more or less space depending on the number/s or image in the right side (see attached screenshot). This means that if we have a progress that continuously change (because it's updating the progress value), the user experience is that the component has a glitch.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.x  |
| React | 16.12.0 |

pat270:
…addon` to prevent resizing progress bar when numbers are changed to icons

fix(@clayui/css): Progress Bar adds Sass map `$progress-group-addon`

fixes #4024

### DIFF
--- a/packages/clay-css/src/scss/components/_progress-bars.scss
+++ b/packages/clay-css/src/scss/components/_progress-bars.scss
@@ -85,17 +85,18 @@
 }
 
 .progress-group-addon {
-	font-size: $progress-group-addon-font-size;
-	font-weight: $progress-group-addon-font-weight;
-	margin-right: $progress-group-addon-spacer-x;
-	text-align: center;
+	@include clay-css($progress-group-addon);
 
 	&:first-child {
-		padding-left: 0;
+		$first-child: setter(map-get($progress-group-addon, first-child), ());
+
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM!

jbalsas:
Just sanity-checking here, since the comment says "_prevent resizing progress bar when numbers are changed to icons_" but the progress bar also changes in size when the number goes from 1 (4%) to 2 digits (10%). Did this fix also addressed that case?

pat270:
@jbalsas Yes that is addressed. There is an issue when the number changes to 3 digits. It's really hard to tell, but would be fixed by setting the min-width to 40px instead of 32.

PR #4041, opened 2021-04-30
feat(@clayui/panel): adds the small API to Panel.Group

### LINKED ISSUE #4025: ClayPanel small version
## What is your proposal?

Create a way to use a small version of `ClayPanel` to use in limited spaces such as sidebars.

Avoid having CSS styles in several places to add a small version of a ClayPanel. In this moment, for example for sidebars, we have custom CSS using `.sidebar-sm.*`. Also, we have a CSS class to remove the padding for the component with `panel-group-flush`.

* ClayPanel demo: https://codesandbox.io/s/claypanel-demo-p8sne?file=/src/App.js

Screenshot 2021-04-26 at 16 58 37

I saw in several places in the portal that we have custom components doing the same as the ClayPanel and maybe we can migrate/update those to use the proper ones from Clay. Several are placed in side sidebars or info sidebars.

As we have in the Pagination Bar component, I'd like to use the ClayPanel with a CSS class such as `panel-sm` or adding `small` in the attributes to make it small (font-size, spacing, icon size, etc).

The alternatives now is to use the ClayPanel as it is, add the `panel-group-flush` CSS class and use it but I don't know if the icon has the right size for sidebars or it's calculated in the theme.

Thanks!

matuzalemsteles:
Fixes #4025

Just by adding the new `small` API to` <ClayPanel.Group />` which renders small panels within the group, I also removed the use of `react-test-renderer` in favor of using only `testing-library`, I went through a few tests and we still use it a lot, I will create another issue later to track it and remove the use of `react-test-renderer` from our test suite.

### DIFF
--- a/packages/clay-panel/src/Group.tsx
+++ b/packages/clay-panel/src/Group.tsx
@@ -30,6 +30,11 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	 * This class generally should be used inside card, sheet, or a type of padded container.
 	 */
 	flush?: boolean;
+
+	/**
+	 * Flag to enable the small variation for panels within a group of panels.
+	 */
+	small?: boolean;
 }
 
 const ClayPanelGroup: React.FunctionComponent<IProps> = ({
@@ -39,6 +44,7 @@ const ClayPanelGroup: React.FunctionComponent<IProps> = ({
 	fluidFirst,
 	fluidLast,
 	flush,
+	small,
 	...otherProps
 }: IProps) => {
 	return (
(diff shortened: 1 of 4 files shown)


bryceosterhaus (commented):
Looks good to me! I am guessing that the updates to the snapshots were because of the switch to react-testing-library?

matuzalemsteles:
Hey @bryceosterhaus yeah, the use of `testing-library` creates a `div` to render the components, probably because it is the `container` to isolate.

bryceosterhaus:
@matuzalemsteles makes sense! Feel free to merge

PR #4047, opened 2021-05-05
fix(@clayui/css): Mixins `clay-select-variant` update to use `clay-cs…

### LINKED ISSUE #4046: @clayui/css Mixins update `clay-select-variant` to use `clay-css` pattern, also deprecated keys should win over new keys
Piecing out parts of https://github.com/liferay/clay/issues/3987 into separate issues to make it easier to track. Since Clay CSS provides so many variables, it would be better if our variables could follow a pattern that mimics the scss. Here's an example:

Old way:
```
$input-select: (
  color: white,
  hover-color: black,
  disabled-option: (
    color: gray,
  ),
);
```

New way:
```
$input-select: (
  color: white,
  hover: (
    color: black,
  ),
  disabled: (
    option: (
      color: gray,
    ),
  ),
);
```

The old way will still be compatible, until we decide to remove it in the next major version.

pat270:
…s` pattern, also deprecated keys should win over new keys

fix(@clayui/css): Update Sass maps that use `clay-select-variant` to use new keys: `$date-picker-nav-select` and `$input-select`

fixes #4046

### DIFF
--- a/packages/clay-css/src/scss/variables/_date-picker.scss
+++ b/packages/clay-css/src/scss/variables/_date-picker.scss
@@ -51,7 +51,6 @@ $date-picker-nav-btn-monospaced: map-deep-merge(
 );
 
 $date-picker-nav-select: () !default;
-$date-picker-nav-select: map-deep-merge((), $date-picker-nav-select);
 
 // Date Picker Row
 
--- a/packages/clay-css/src/scss/atlas/variables/_date-picker.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_date-picker.scss
@@ -26,7 +26,7 @@ $date-picker-nav-btn: map-deep-merge(
 $date-picker-nav-select: () !default;
 $date-picker-nav-select: map-deep-merge(
 	(
-		bg: transparent,
+		background-color: transparent,
 		border-color: transparent,
 		color: $secondary,
 		font-size: $font-size-sm,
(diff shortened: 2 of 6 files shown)


packages/clay-css/src/scss/variables/_date-picker.scss
```diff
@@ -51,7 +51,6 @@ $date-picker-nav-btn-monospaced: map-deep-merge(
 );
 
 $date-picker-nav-select: () !default;
-$date-picker-nav-select: map-deep-merge((), $date-picker-nav-select);
```
matuzalemsteles, line 54: Just a question about the unrelated this change:

- Double declaration in sequence: the last declaration rewrite the last declaration, right? why double declaration?
- Once you removed the last declaration, what's the expected side effect? if it exist, of course.

matuzalemsteles (commented):
Hey @pat270 I'm taking a look at this to understand more about how everything is working around Clay CSS, so I'm going to ask more questions from now on, probably always doubts.

packages/clay-css/src/scss/variables/_date-picker.scss
pat270: @matuzalemsteles Not a problem, `map-merge` takes whatever was declared in the first map and combines the two. If there are two keys with the same name in both maps, whatever is declared in the second map wins between the two. This is a way (hack?) to provide default values with Sass maps.

An example that may help explain better:
```
// This one replaces `$map: () !default;` below it
$map: (
  background-color: #eee,
  color: red,
) !default;

$map: () !default;
$map: map-merge((
  color: black,
  font-size: 16px,
), $map);

// $map will be:
$map: (
  background-color: #eee,
  color: red,
  font-size: 16px,
);
```

If we don't use `map-merge` while providing default values:

```
$map: (
  background-color: #eee,
  color: red,
) !default;

$map: (
  color: black,
  font-size: 16px,
) !default;

// $map will be:
$map: (
  background-color: #eee,
  color: red,
);
```

The reason we don't need it here is because we aren't providing any default values for `$date-picker-nav-select`.

matuzalemsteles: Hmm makes sense, thanks for the clarification.

matuzalemsteles (approved):
LGTM

marcoscv-work:
LGTM and merge!

marcoscv-work:
I'm not sure why it was for 13 days here...

Issue #4027, opened 2021-04-26
Clay Shared Portal allow passing class names and id's


pat270:
Cadmin relies on a CSS class to mark specific content admin only. The class makes it difficult to overwrite styles. Currently, it's not possible to apply this class to `modal-backdrop` in ClayModal. I think it would be best if we could allow passing attributes to the React Portal div. It would give us the most freedom to scope how we want. This feature would relieve my headaches with Dropdown, Modal, Tooltip, and Popover.

PR #4067, opened 2021-05-13
feat(@clayui/shared): Clay Shared Portal allow passing class names and id's

ambrinchaudhary:
Fixes https://github.com/liferay/clay/issues/4027

> Cadmin relies on a CSS class to mark specific content admin only. The class makes it difficult to overwrite styles. Currently, it's not possible to apply this class to modal-backdrop in ClayModal. I think it would be best if we could allow passing attributes to the React Portal div. It would give us the most freedom to scope how we want. This feature would relieve my headaches with Dropdown, Modal, Tooltip, and Popover.

I'm opening this pull as Draft so we can discuss the proper approach for this issue. 

The original issue says that `className` and `id` should be allowed to pass as otherProps to our `ClayPortal` component. So the idea is to add these attributes to the [root element](https://github.com/liferay/clay/blob/master/packages/clay-shared/src/Portal.tsx#L30). The [first commit](https://github.com/liferay/clay/pull/4067/commits/d7fbe225710620794c16d6ca73466f0e314b3c13) fixes this. 

But if wer want to extend these `otherProps` to any other react attributes, then we need to render the [root element](https://github.com/liferay/clay/blob/master/packages/clay-shared/src/Portal.tsx#L30) (div) with react, and here is where I am finding problems and what I want to be discussed. Work is in progress in the [second commit](https://github.com/liferay/clay/pull/4067/commits/fa5ac9afc553304a4acd9b4047c03c0f94f4293b), and it makes other tests fails... so it is not working properly. I've also tried another aproach which was adding this div to the [content](https://github.com/liferay/clay/blob/master/packages/clay-shared/src/Portal.tsx#L60), but also resulted in test failures for other components. Just wanted to know if this is what we want to achieve and if you have any other suggestions for doing this. 

Thanks!


packages/clay-shared/src/Portal.tsx
```diff
@@ -26,9 +26,8 @@ export const ClayPortal: React.FunctionComponent<
 	}
 > = ({children, containerRef, subPortalRef, ...otherProps}) => {
 	const parentPortalRef = React.useContext(ClayPortalContext);
-	const portalRef = React.useRef(
-		typeof document !== 'undefined' ? document.createElement('div') : null
```
ambrinchaudhary, line 30: If this div must accept the `otherProps`, it has to be rendered in React

packages/clay-shared/src/Portal.tsx
```diff
@@ -41,15 +40,17 @@ export const ClayPortal: React.FunctionComponent<
...
-			if (otherProps?.className) {
-				portalRef.current.classList.add(otherProps.className);
-			}
-			if (otherProps?.id) {
-				portalRef.current.id = otherProps.id;
-			}
+		if (elToMountTo) {
+			const myDiv = document.createElement('div');
+
+			ReactDOM.render(<div {...otherProps}></div> , myDiv, () => {
```
ambrinchaudhary, line 46: `ReactDOM.render` does not allow you to render directly in the `document.body` element. Thus, I had to create a temporal div `myDiv`, and assign its firstChild to the `portalRef.current`. This should result in the same that we had before while doing `elToMountTo.appendChild(portalRef.current);`.

packages/clay-shared/src/Portal.tsx
```diff
@@ -41,15 +40,17 @@ export const ClayPortal: React.FunctionComponent<
...
+				portalRef.current = myDiv.firstChild as Element;
+
+				elToMountTo.appendChild(portalRef.current);
 
-			elToMountTo.appendChild(portalRef.current);
+				setTest('re-render'); //this way we force a re-render when portalRef.current has a value,
```
ambrinchaudhary, line 51: Notice that this is done in the ReactDOM.render callback, so to force a re-render of the component (otherwise the `portalRef.current` is null and the createPortal was not being called here) I had to use the states of React. Probably there is a best way to do this... I'm all hears.

packages/clay-shared/src/Portal.tsx
```diff
@@ -42,6 +42,13 @@ export const ClayPortal: React.FunctionComponent<
 				: closestParent;
 
 		if (elToMountTo && portalRef.current) {
+			if (otherProps?.className) {
+				portalRef.current.classList.add(otherProps.className);
+			}
+			if (otherProps?.id) {
+				portalRef.current.id = otherProps.id;
+			}
+
```
matuzalemsteles: Maybe we can create a function around the `document.createElement` so that we can create the element already with the attributes and we can call it directly in `useRef`. I have a feeling this would be cleaner, what do you think?

Something like that:

```ts
const createElement = (nodeName: string, attributes: Record<string, string>) => {
  const element = document.createElement(nodeName);

  Object.keys(attributes).forEach((key) => {
     const value = attributes[key];

     element.setAttribute(key, value);
  });

  return element;
}

```

matuzalemsteles (commented):
Thanks @ambrinchaudhary for investigating this and for the details, well I think to really support the generic props that are React, we are really going to have to work to convert the properties to the element, I think that maybe this is not worth the effort now since you have encountered problems but if we have to endure this at some point in the future we will already have a vision of we can try other options.

So I think we can ignore the case of passing generic props and for now, it only supports `id` and `className`.

packages/clay-shared/src/Portal.tsx
ambrinchaudhary: Thank @matuzalemsteles, I've done it here https://github.com/liferay/clay/pull/4067/commits/231e1640295c39b5bf5f49dd021114bad67a7b9e. Although, I was not able to type `attributes` as you mentioned above, it returned me this error: 
```
Argument of type '{ defaultChecked?: boolean | undefined; defaultValue?: string | string[] | undefined; suppressContentEditableWarning?: boolean | undefined; suppressHydrationWarning?: boolean | undefined; ... 247 more ...; onTransitionEndCapture?: TransitionEventHandler<...> | undefined; }' is not assignable to parameter of type 'Record<string, string>'.
```

And following that, I'm not sure if I can type it with something else than "any", what do you think?

julien: > And following that, I'm not sure if I can type it with something else than "any", what do you think?

I think that if we're using TypeScript it's to actually avoid using `any`, but that's just my opinion.

matuzalemsteles: Yeah it's always good to avoid using `any`, probably this TypeScript error is because we are doing the union for `React.HTMLAttributes<HTMLDivElement>`, we could remove this and add the types just for `id` and `className`, then it would not conflict with the `attributes` typing.

https://github.com/liferay/clay/blob/231e1640295c39b5bf5f49dd021114bad67a7b9e/packages/clay-shared/src/Portal.tsx#L16

packages/clay-shared/src/Portal.tsx
```diff
@@ -24,10 +24,26 @@ export const ClayPortal: React.FunctionComponent<
 		 */
 		subPortalRef?: React.RefObject<Element>;
 	}
-> = ({children, containerRef, subPortalRef}) => {
+> = ({children, containerRef, subPortalRef, ...otherProps}) => {
 	const parentPortalRef = React.useContext(ClayPortalContext);
+
+	const createElement = (nodeName: string, attributes: any) => {
```
matuzalemsteles, line 30: It would be a good idea to move this method out of our component just to avoid creating a new reference or function declaration every time a new rendering occurs in the component.

packages/clay-shared/src/Portal.tsx
ambrinchaudhary: Hi @matuzalemsteles @julien please check commit https://github.com/liferay/clay/pull/4067/commits/4faedf64195f59851748a9d0d151157738213e32 to see if I understood you correctly. Otherwise, apologises and do not hesitate to contact me via slack. Thanks!

packages/clay-shared/src/Portal.tsx
```diff
@@ -10,24 +10,57 @@ const ClayPortalContext = React.createContext<React.RefObject<Element | null> |
...
+		element.id = id;
 	}
-> = ({children, containerRef, subPortalRef}) => {
+
+	return element;
+};
+
+ClayPortalContext.displayName = 'ClayPortalContext';
+
+interface IProps extends React.HTMLAttributes<HTMLDivElement> {
```
matuzalemsteles, line 29: Hey @ambrinchaudhary, I think, it was not very clear, I'm sorry. The idea is that we can remove the `extend React.HTMLAttributes<HTMLDivElement>` from `IProps`, because we don’t need all the features of an `HTMLDivElement` and we don’t use it either, so in order not to provide the wrong types, we can just remove the `extend` and add the explicit type for `className` and `id`, just like you already did here. This way you can make `createElement` generic as before.

ambrinchaudhary: Ok, I think I got you ;-). But I thinkg that removing the extend implies that `children` has to be typed as well....I'm getting errors on that but will work on this tomorrow. Thanks for your explanation.

matuzalemsteles: @ambrinchaudhary I forget that 😅. Yeah, this is also necessary, add the explicit type, it may have to be `React.ReactElement` and `React.ReactNode`, probably.

ambrinchaudhary: Done in https://github.com/liferay/clay/pull/4067/commits/b9e8809ff1a06120aa0e417e1268b83c22315cf5. Although... I'm not sure if this is cleaner than passing directly the `className` and the `id` attributes, which are the ones that we  want to support... or do we really want to do this generic? Sure I'm missing something... so I let it up to you ;-) .

julien: @matuzalemsteles, just to add some context, @ambrinchaudhary and I were talking about the `createElement` function and we don't really understand why it needs to be generic. Could you tell us why you think this is needed?

matuzalemsteles: Of course. In my mind is that instead of manually defining `id` and `class` we can just isolate the `createElement` function from the rest, that is, it doesn’t know what attribute we are passing through or why we are just defining `id` and `class`, so when we see this at some point in the future it becomes simpler to look at it and be able to add new attributes if necessary.

So we were able to isolate `createElement` from the rest instead of attaching its signature to just `id` or `class`, for example, it is generic but we have full control of what we pass on to it, if we want to be more secure we can just `createElement('div', {id, class: className})`. Does it make sense?

Normally we may just want to define `id` and `class`, it also makes sense that this is not wrong. We also have the idea that generalizing too early can be bad but at some point in the future when we look at this, let's think about why we just defined that, why not simplify the method?, was that a use case?... I usually speak that when a person sees something in the code they will accept what they have and continue or they will refactor. In my view, this seems like a healthy change for us to keep around, even if it grows it seems that it's logic will not change.

matuzalemsteles (approved):
LGTM! Good job, I hope you haven't been frustrated by the changes 😬

packages/clay-shared/src/Portal.tsx
julien: @matuzalemsteles I understand how this `createElement` function works, I just thought it might be unnecessary given that we're only going to create `div` elements and only set their `id` and `class` attributes. I'm not saying it's bad to support more names and attributes (and on top of that we're calling the function ourselves so there's no risk of seeing `createElement('foo', {bar: 'baz'})`) but if we really wanted to implement a "robust" `createElement` function it could quickly become complex. For example, are we  going to let pass any value for our `name` parameter, do we want to validate the attributes, etc... Since we're not exporting it, there's no harm done and we can push this forward as far as I'm concerned but I just wanted to check it with you.

matuzalemsteles: @julien yeah I agree, it's makes sense too. I think we can always try to find a balance, currently this implementation is not very complex, it allows to evolve, minimally generic to create an element with several native attributes, like `aria`... etc, so your responsibility is not to validate or deal with other extreme cases that we would have to deal with, like updating the values after they change, that would probably be somewhere else ... what I mean is that we can generalize a little bit just to not keep some things hardcoded, like manually defining the `element.id = id`. For example, as this is only used in this file which makes sense, we would have to do something like:

```js
const createDivElement = (
	className: string,
	id: string,
) => {
	const element = document.createElement('div');

	element.setAttribute('class', className);
	element.setAttribute('id', id);

	return element;
};
```

Does that make sense to you?

ambrinchaudhary:
LGTM! Good job, I hope you haven't been frustrated by the changes 😬

packages/clay-shared/src/Portal.tsx
julien: Hey @matuzalemsteles I agree about finding a balance. Your suggestion is *exactly* what I had in mind, and my reasoning is the following: I'd rather not do anything more than what's needed to make it work. 

I'd go with the solution you suggested in the previous comment (just `className` and `id` parameters for now), and if we need to allow for more attributes in the future, we can always go with the more complete version (i.e [b7445bc](https://github.com/liferay/clay/commit/b7445bcd6f16cdda39ef7bd0af58bda2d3b1949e)) (after a second thought, attribute validation seems like a wrong idea, since even the native APIs let you create any kind of element with any attribute without throwing exceptions :smile: ) . 

I *have to admit* that the current change doesn't add a lot complexity, so you could say that I "complained for nothing"; but  it's always a good thing to take a moment and think about a change to be sure that we're all on the same track. Thanks!

ambrinchaudhary: Ok then, I added a new commit with simplifying our function. I doubt about putting back or not the extend of `React.HTMLAttributes<HTMLDivElement>,` but if we are not using it, better not to add it, right?

matuzalemsteles: @julien don't worry, this is good for evaluating our options 🙂.

>  Ok then, I added a new commit with simplifying our function. I doubt about putting back or not the extend of React.HTMLAttributes<HTMLDivElement>, but if we are not using it, better not to add it, right?

Yes, if it is not being used then I think it is better not to add it. Because when we use the component we will be seeing a typing that is not true, accepting properties that are not actually being accepted.

matuzalemsteles:
Ok, I think everyone agrees with that, so I’m going to merge it. Thanks for the work @ambrinchaudhary

PR #4068, opened 2021-05-13
fix(@clayui/css): Mixins `border-radius` shouldn't output the default…

### LINKED ISSUE #4066: @clayui/css Management Bar, Navigation Bar, and other navbars have rounded corners
mgmt-bar-rounded

This is due to changes from #4009.

pat270:
… radius if a null parameter is passed

fixes #4066

### DIFF
--- a/packages/clay-css/src/scss/mixins/_border-radius.scss
+++ b/packages/clay-css/src/scss/mixins/_border-radius.scss
@@ -1,4 +1,12 @@
-@mixin border-radius($radius: null, $fallback-border-radius: false) {
+@mixin border-radius(
+	$radius:
+		if(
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM! I only have one question, this only happens when we have the cadmin and clay on the same page, right?

pat270:
@matuzalemsteles This happens when a `null` value is passed to the mixin, it would default to one of the border-radius variables. This change makes it only use one of the border-radius variables if no parameter is passed (e.g., `@include border-radius();`). Clay and Cadmin share the function and mixin libraries, everything else is independent.

matuzalemsteles:
I got it, thanks @pat270

PR #4072, opened 2021-05-14
fix(@clayui/css): Mixins `clay-nav-variant` convert to use new (easie…

### LINKED ISSUE #4071: @clayui/css Mixin `clay-navbar-variant` should use new Sass map pattern
See #3987

pat270:
…r to remember) Sass map keys. The old keys will still work and win over new keys.

- Adds option `enabled` to prevent styles from being output
- Adds option `nav-item` to style `.nav-item` class in the mixin
- Adds option `nav-link` to style `.nav-link` class in the mixin, replaces `link-*-*` option

fixes #4071

### DIFF
--- a/packages/clay-css/src/scss/mixins/_nav.scss
+++ b/packages/clay-css/src/scss/mixins/_nav.scss
@@ -29,78 +29,164 @@
 /// A mixin to create a `.nav` color variant.
 /// @param {Map} $map - A map of `key: value` pairs. The keys and value types are listed below:
 /// @example
+/// // `bg` is deprecated use `background-color` instead
 /// bg: {Color | String | Null},
+/// // `link-bg` is deprecated use Sass map `nav-link` instead
 /// link-bg: {Color | String | Null},
+/// // `link-border-radius` is deprecated use Sass map `nav-link` instead
 /// link-border-radius: {Number | String | List | Null},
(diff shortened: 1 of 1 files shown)


packages/clay-css/src/scss/mixins/_nav.scss
```diff
@@ -29,78 +29,164 @@
 /// A mixin to create a `.nav` color variant.
 /// @param {Map} $map - A map of `key: value` pairs. The keys and value types are listed below:
 /// @example
+/// // `bg` is deprecated use `background-color` instead
```
matuzalemsteles, line 32: Since we are replacing these keys with another one, can this cause any breaking changes? could that be used by anyone?

pat270: @matuzalemsteles I've included an alias at https://github.com/liferay/clay/blob/3fa114a3a6f4f2109054e6a08ca026627b9a0e4a/packages/clay-css/src/scss/mixins/_nav.scss#L73-L74. Both `bg` and `background-color` work, if for some reason both are used, `bg` will win over `background-color`.

matuzalemsteles: Oh I got it. So, security here is using a `setter` to ensure compatibility, this is interesting. Thanks @pat270

PR #4080, opened 2021-05-19
feat(@clayui/color-picker): Add the new DropDownContainerProps API

### LINKED ISSUE #4028: Clay Color Picker allow passing classes to ClayDropdown.Menu
https://github.com/liferay/clay/blob/master/packages/clay-color-picker/src/index.tsx#L233

This issue is related to #4027. I'm unable to scope dropdown-menu's in ClayColorPicker because I can't pass class names to ClayDropdown.Menu.

matuzalemsteles:
Fixes #4028

Adding new APIs to add properties to get to `<ClayPortal />`, ideally, we wanted to get rid of having APIs with `*Props` as this may mean that we can go in mismatch the composition or our own foundations but this has become very common in our components.

I'm following the pattern here to maintain consistency but I have some ideas about this to avoid adding more APIs with `*Props` and creating a consistent pattern for all components without having to hurt our foundations. I will try to create a proposal for us to discuss.

### DIFF
--- a/packages/clay-color-picker/src/index.tsx
+++ b/packages/clay-color-picker/src/index.tsx
@@ -72,6 +72,14 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	 * Flag for adding ColorPicker in disabled state
 	 */
 	disabled?: boolean;
+
+	/**
+	 * Props to add to the DropDown container.
+	 */
+	dropDownContainerProps?: React.ComponentProps<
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
Considering that we want this for today's release to help Patrick with the implementation of CSS Isolate and that doesn’t bring too many problems since we have the same pattern within our components and doesn’t present major degradation problems for our APIs, I’ll merge that 🙂 but feel free to leave your concerns so we can continue to discuss these standards.

PR #4082, opened 2021-05-19
fix(@clayui/css): Global Functions update `setter` to accept more tha…

### LINKED ISSUE #4076: @clayui/css function setter should have clearer parameter names
The Clay CSS function `setter` has two parameter names `$var` and `$val`. The second parameter name isn't an accurate description of what it accepts. The function compares the value of two variables and returns the first one if it doesn't equal `null`. If it is `null` it returns the value of the second parameter. We should rename these to `$var1` and `$var2` or something that better describes what you pass in.

https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/functions/_global-functions.scss#L65

pat270:
…n 2 variables (e.g., `setter($var1, $var2, $var3, $var4, $var5)`)

fixes #4076

Instead of renaming the two parameters, it dawned on me that I could use an arglist. The `setter` function now accepts any number of variables and returns the first one that isn't `null`. The code below:

```
box-shadow: setter(
    map-get($map, active-focus-box-shadow),
    setter(
        map-get($active-focus, box-shadow),
        map-get($focus, box-shadow)
    )
),
```

can be changed to:

```
box-shadow: setter(
    map-get($map, active-focus-box-shadow),
    map-get($active-focus, box-shadow),
    map-get($focus, box-shadow)
),
```

### DIFF
--- a/packages/clay-css/src/scss/functions/_global-functions.scss
+++ b/packages/clay-css/src/scss/functions/_global-functions.scss
@@ -58,20 +58,21 @@
 	@return $newMap;
 }
 
-/// A helper function for setting default values in variables inside mixins if no value is declared. If the value of a variable is `clay-unset`, `setter` returns a value of `null` which prevents Sass from outputting the declaration. If the value of a variable is `null`, `setter` returns the default, `$val`.
-/// @param {Any} $var - The Sass variable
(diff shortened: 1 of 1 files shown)


marcoscv-work:
cool thing!

LGTM!

PR #4084, opened 2021-05-19
feat: Add the new ContainerProps API to Popover, TooltipProvider, Modal and DropDown

### LINKED ISSUE #4083: Allow adding `className` and `id` to components using ClayPortal
Since we added these behaviors through issue #4027 and support for ColorPicker was added in #4028, we also need to add support for components that use ClayPortal to allow passing `className` and `id`.

- DropDown
- Modal
- TooltipProvider
- Popover

matuzalemsteles:
Fixes #4083

It just does the same thing that was done for ClayColorPicker, we are adding for all components that depend on `<ClayPortal />`.


matuzalemsteles:
The same in #4080, I will merge it for have in the release today

PR #4100, opened 2021-05-27
feat(@clayui/drop-down): Make it possible to render dividers

### LINKED ISSUE #4089: [DropdownWithDrilldown] Cannot use dividers inside menu's

Here https://codesandbox.io/s/xenodochial-darwin-ljybh?file=/src/App.js

(It has no styles, because I didn't found the Clay template, but you can see the empty button)

1. Open the sandbox
2. Open the dropdown
3. Behold!

Having a divider (or at least an error instead of an empty button)

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.28.0  |
| React | v17.0.2 (same result with v16.x) |

julien:
In this change, we're adding the possibility to render dividers
in the `ClayDropDownWithDrilldown` component. The change is rather
straightforward: if a menu item with the `type: 'divider'` is passed
in the `menus` prop, it will be rendered as a `<Divider />`.

Fixes #4089

### DIFF
--- a/packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
+++ b/packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
@@ -1,47 +1,357 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`ClayDropDownWithDrilldown renders 1`] = `
-Document {
-  "location": Location {
-    "assign": [Function],
-    "hash": "",
-    "host": "localhost",
-    "hostname": "localhost",
-    "href": "http://localhost/",
-    "origin": "http://localhost",
-    "pathname": "/",
-    "port": "",
-    "protocol": "http:",
-    "reload": [Function],
-    "replace": [Function],
-    "search": "",
-    "toString": [Function],
-  },
-  Symbol(SameObject caches): Object {
-    "styleSheets": StyleSheetList {},
-  },
-}
+<body>
+  <div>
+    <div
+      class="dropdown"
+    >
+      <button
+        class="dropdown-toggle"
+        data-testid="trigger"
+        style=""
+      />
+    </div>
+  </div>
+  <div>
+    <div>
+      <div
+        class="dropdown-menu drilldown dropdown-menu-indicator-end show"
+        style="left: -999px; top: -995px;"
+      >
+        <div
+          class="drilldown-inner"
+        >
+          <div
+            class="drilldown-item drilldown-current"
+          >
+            <ul
+              class="inline-scroller"
+            >
+              <li>
+                <a
+                  class="dropdown-item"
+                  data-testid="menu-item-Hash Link"
+                  href="#"
+                >
+                  <span
+                    class="dropdown-item-indicator-text-end"
+                  >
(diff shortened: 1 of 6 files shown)


packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
```diff
@@ -22,3 +22,26 @@ Document {
...
+    "protocol": "http:",
+    "reload": [Function],
+    "replace": [Function],
+    "search": "",
+    "toString": [Function],
+  },
+  Symbol(SameObject caches): Object {
+    "styleSheets": StyleSheetList {},
+  },
+}
```
matuzalemsteles, line 46: This is a little intriguing, I think it wasn’t meant to be rendered that way and I don’t know how I didn’t realize it before 🤷‍♂️. We may have to change everyone to render `document.body`

packages/clay-drop-down/src/drilldown/Menu.tsx
```diff
@@ -88,55 +93,59 @@ const DrilldownMenu: React.FunctionComponent<IProps> = ({
...
-										onClick={(
-											event: React.SyntheticEvent
-										) => {
-											if (onClick) {
-												onClick(event);
-											}
+							) =>
+								type === 'divider' ? (
+									<Divider key={`${j}-divider`} />
+								) : (
```
matuzalemsteles, line 103: A good strategy to avoid conditionals for cases like this is to create a dictionary of valid types and access it to render the respective components, we can decrease some of the noise, we do the same for `DropDownWithItems` but we don't need to worry about that now since we support only one `type`.

matuzalemsteles (commented):
In general, it looks good to me, I just added a few comments but it doesn't block that PR.

packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
julien: I hadn't seen that either, let me have a look into it and get back to you.

packages/clay-drop-down/src/drilldown/Menu.tsx
julien: I thought about using a look up table, but for a simple case like this it seemed like overkill, that said let me have second look. I don't think there's any rush here, and we can probably agree quickly.

matuzalemsteles: Yes it can be a possibility too, well I'm fine if we go like this, we can change that in the future if we have more use cases.

packages/clay-drop-down/src/__tests__/__snapshots__/DropDownWithDrilldown.tsx.snap
julien: Yeah :see_no_evil: didn't even realize we were checking `document` and not `document.body`, updated in  [02be8af](https://github.com/liferay/clay/pull/4100/commits/02be8af2b2171891ad3ae0bfd922d9362213a9d5)

matuzalemsteles (approved):
LGTM

PR #4101, opened 2021-05-27
chore(@clayui/css): Remove unneeded RTL icons

### LINKED ISSUE #4077: Wondering if we need/use some RTL icons
## Just wondering if we are currently using RTL icons and why we created some of them

I think we could remove some of them since their meaning is not changing or providing new or better information in RTL version. They are:

- copy-rtl (it is a standard icon, just a quick review in other tools to see they don't usually change it in RTL)
- flag-empty-rtl (the wind changed? haha)
- flag-full-rtl
- paste-plaintext-rtl (no meaning changes)
- paste-word-rtl (no meaning changes)

image

julien:
Fixes #4077


matuzalemsteles (approved):
LGTM

PR #4102, opened 2021-05-27
feat(@clayui/drop-down): add the new contextual type to create a cascading menu

### LINKED ISSUE #4097: Cascading Menu
dropdown-multi-level
_credits @hold-shift_

## What is your proposal?
Due to [LEXI-1196](https://issues.liferay.com/browse/LEXI-1196) we propose to develop a Cascading Menu, an evolution/addon of the Dropdown Menu. Please review the proposal: [Figma file](https://www.figma.com/file/uTmPPMQ8UfWQyKIepyZ3p9/lexi-1196-cascading-menu?node-id=0%3A1), [definition doc](https://docs.google.com/document/d/1gnFu8Fwb_JEl1Lat_THfzXN9vkCJ7Q4f6zM82ioj1Wo/edit?usp=sharing). 

This would help product designers to use the proper Dropdown Menu depending on the scenario. For example, a Cascading Menu is designed for contextual menus with many actions (recommended for just 1 level) and a Drill Down Menu for deep hierarchies (2+ levels or navigation).

Currently, product designers are using a Drill Down Menu which is a multilevel solution too but it doesn't work well because a Cascading Menu is the most known pattern for a nested level. 

## Deadline
The echo-team requested to have this for **DXP 7.4 GA2** (see LEXI ticket)

matuzalemsteles:
Fixes #4097 

Well add this just as a draft just missing to understand how the visibility of the cascading menu will work, it is still not very clear if when hovering the mouse outside the cascading menu it should disappear, looking at the Figma prototype it looks like this but I'm not sure about that.

- [Spec](https://docs.google.com/document/d/1gnFu8Fwb_JEl1Lat_THfzXN9vkCJ7Q4f6zM82ioj1Wo/edit?usp=sharing)

I thought that just adding a level would be easier but in fact, our API allows you to add another level, so implementing it in a recursive way is more natural for our implementation. this is because we only want one level. To create a cascading menu is the same as creating a `group` but just change the type to `contextual` (I'm not sure about that name yet, suggestion is welcome), this allows you to switch between a group or create one menu.

```js
const items = [
	{label: 'Folder'},
	{type: 'divider'},
	{
		items: [
			{label: 'Basic Document'},
			{label: 'Contract'},
			{label: 'Marketing Banner'},
			{label: 'Spreadsheet'},
			{label: 'Presentation'},
		],
		label: 'Document',
		type: 'contextual',
	},
	{label: 'Shortcut'},
	{label: 'Repository'},
];
```

### Contextual Menu

Well, I followed a pattern that is common for contextual menus, where we build a safe area that allows the user to navigate to the menu without the menu closing, this video shows this aria well. Some examples like https://css-tricks.com/menus-with-dynamic-hit-areas/

- [Demo](https://deploy-preview-4102--next-storybook-clayui.netlify.app/?path=/story/components-claydropdown--w-cascading-menu)

https://user-images.githubusercontent.com/13750819/119883129-8f9b6680-bf05-11eb-9849-c36ec2874947.mp4

@drakonux could you confirm what the visibility behavior would be?

### DIFF
--- a/packages/clay-drop-down/README.mdx
+++ b/packages/clay-drop-down/README.mdx
@@ -17,6 +17,7 @@ import {
 -   [Composing](#composing)
     -   [Using dropdown without a trigger](#using-dropdown-without-a-trigger)
 -   [DropDownWithItems](#dropdownwithitems)
+    -   [Cascading Menu](#cascading-menu)
 -   [DropDownWithDrilldown](#dropdownwithdrilldown)
 -   [Caveats](#caveats)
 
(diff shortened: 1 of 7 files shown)


matuzalemsteles:
Well the interactions and behaviors are following the specifications of the documentation and we added a new `type` to support the cascading menu, the API does not differ from the others, it is possible to convert a type `group` to `contextual` just by changing the property, so nothing different to create contextual menu.

I'm going ahead with this PR to include in today's release, if we find any behavioral or interaction problems we can include the improvements in the next release.

PR #4107, opened 2021-05-31
 feat(@clayui/drop-down): Allow the menu to be toggled

### LINKED ISSUE #4103: [DropdownWithDrilldown] Possibility to close the Drop Down when an option is chosen
## What is your proposal?
Due to this [LPS-132227](https://issues.liferay.com/browse/LPS-132227), we propose that when an option is clicked, the Drop Down menu is closed automatically.

This could give the component more flexibility and give the option to automatically close the menu when an action is clicked.
 
There are no alternatives.

julien:
Two new props have been added to `ClayDropDownWithDrillDown`:
`active` and `onActiveChange`. Setting `active` to `true` will make the
menu initially `active` (open), and `onActiveChange` will allow the menu
to be toggled.

Fixes #4103

### DIFF
--- a/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
+++ b/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {TInternalStateOnChange, useInternalState} from '@clayui/shared';
 import classNames from 'classnames';
 import React from 'react';
 
@@ -11,6 +12,11 @@ import ClayDropDownMenu from './Menu';
 import Drilldown from './drilldown';
 
 interface IProps extends React.HTMLAttributes<HTMLDivElement> {
+	/**
+	 * Flag to indicate if the menu should be initially active (open).
+	 */
+	active?: boolean;
+
 	/**
 	 * Default position of menu element. Values come from `./Menu`.
 	 */
@@ -54,6 +60,11 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	offsetFn?: React.ComponentProps<typeof ClayDropDown>['offsetFn'];
 
 	/**
+	 * Callback the will be invoked when the active prop is changed.
+	 */
+	onActiveChange?: TInternalStateOnChange<boolean>;
+
+	/**
 	 * Path to spritemap
 	 */
 	spritemap?: string;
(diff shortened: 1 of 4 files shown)


packages/clay-drop-down/src/DropDownWithDrilldown.tsx
```diff
@@ -82,7 +88,7 @@ export const ClayDropDownWithDrilldown: React.FunctionComponent<IProps> = ({
 	spritemap,
 	trigger,
 }: IProps) => {
-	const [active, setActive] = React.useState(false);
+	const [active, setActive] = React.useState(initiallyOpen);
```
matuzalemsteles, line 91: Hey @julien I think that just adding a new property to define the initial state will probably not support the desired use case which is to close DrillDown when clicking on the element, in which case we need to make DrillDown controlled by the developer, support two new properties, `active `and `onActiveChange` to allow the state to be controlled. We have a hook that handles this, to control the state internally when these two properties do not exist.

```js
const [internalActive, setInternalActive] = useInternalState({
	initialValue: false,
	onChange: onActiveChange,
	value: active,
});
```

Example https://github.com/liferay/clay/blob/master/packages/clay-drop-down/src/DropDownWithItems.tsx#L402-L406

julien: Hey @matuzalemsteles sure,
After talking with @veroglez, I thought we just needed the drop down to be initially open. I'll update.

julien: @matuzalemsteles looking at this 

```js
const [internalActive, setInternalActive] = useInternalState({
	initialValue: false,
	onChange: onActiveChange,
	value: active,
});
```

One thing I don't understand is why we would set `initialValue` to `false`, since we want to add the possibility for the menu to be open immediately.

matuzalemsteles: @julien in this case it is just to define the initial value, if we want to make it configurable we must also have a new property called `initialValue`.

matuzalemsteles: > After talking with @veroglez, I thought we just needed the drop down to be initially open. I'll update.

Hmm, it didn't seem very clear on the issue but either way, we can make it consistent with `DropDownWithItems`.

julien: :+1:

julien: T.I.L.:  [`useInternalState`](https://github.com/liferay/clay/blob/d639da89b4ac9eddb66a0b5270da046c4da6565d/packages/clay-shared/src/useInternalState.ts#L19-L35)

matuzalemsteles (approved):
lgtm

PR #4116, opened 2021-06-07
fix(@clayui/css): Panel adjust spacing for `panel-unstyled` to preven…

### LINKED ISSUE #4109: ClayPanel makes a little visual jump when being closed

Easiest, we only have to go to https://clayui.com/docs/components/panel.html#clay-sidebar, and in the first example, change `displayType` from `secondary` to `unstyled`

1) Render a ClayPanel with `displayType` `unstyled`
2) Open and close the panel
3) We can see how the panel makes a "little visual jump" when being closed

 The panel is closed smoothly and it doesn't make any jump

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.28.0  |

pat270:
…t visual jump on close.

fix(@clayui/css): Panel adds Sass maps `$panel-group-panel-unstyled`, `$panel-group-sm-panel-unstyled`, `$panel-group-flush-panel-unstyled`

fix(@clayui/css): Cadmin Panel apply these updates

fixes #4109

### DIFF
--- a/packages/clay-css/src/scss/variables/_panels.scss
+++ b/packages/clay-css/src/scss/variables/_panels.scss
@@ -97,6 +97,14 @@ $panel-title-small-margin-left: null !default;
 
 $panel-group-panel-margin-bottom: -$panel-border-bottom-width !default;
 
+$panel-group-panel-unstyled: () !default;
+$panel-group-panel-unstyled: map-deep-merge(
+	(
+		margin-bottom: 1.5rem,
+	),
+	$panel-group-panel-unstyled
+);
+
 // Panel Group Sm
 
 $panel-group-sm: () !default;
@@ -111,6 +119,14 @@ $panel-group-sm: map-deep-merge(
 	$panel-group-sm
 );
 
+$panel-group-sm-panel-unstyled: () !default;
+$panel-group-sm-panel-unstyled: map-deep-merge(
+	(
+		margin-bottom: 1rem,
+	),
+	$panel-group-sm-panel-unstyled
+);
+
 // Panel Group Flush
 
 $panel-group-flush-body-margin-bottom: 1.5rem !default;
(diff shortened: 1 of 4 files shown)


packages/clay-css/src/scss/variables/_panels.scss
```diff
@@ -145,17 +172,30 @@ $panel-unstyled-header-link: () !default;
...
-		header-border-width: 0 0 1px 0,
-		header-margin-bottom: 1rem,
-		header-collapsed-border-color: $gray-500,
+		margin-bottom: 1.5rem,
+		header: (
+			border-color: $gray-500,
+			border-radius: 0,
+			border-style: solid,
+			border-width: 0 0 1px 0,
+			padding-left: 0,
```
matuzalemsteles, line 184: @pat270 in theory who should remove the padding here shouldn't be `flush` with the panel group? could adding this as a default break any other behavior that spacing is expected?

packages/clay-css/src/scss/variables/_panels.scss
```diff
@@ -145,17 +172,30 @@ $panel-unstyled-header-link: () !default;
...
+		),
+		header-collapsed: (
+			border-color: $gray-500,
+		),
```
matuzalemsteles, line 188: I've noticed some patterns that we are trying to adopt with `map-deep`, this is more of a question, but why not add this inside the `header`, is this a nesting limitation?

matuzalemsteles (commented):
Overall looks good, thanks @pat270! I just added a comment before we go with this

packages/clay-css/src/scss/variables/_panels.scss
pat270: Originally I was expecting `panel-group-flush` to remove the padding, but Lexicon always has the text aligned with the header border. If we want it to adhere to Lexicon spec without `panel-group-flush` we should remove it here.

packages/clay-css/src/scss/variables/_panels.scss
pat270: @matuzalemsteles no limitation. I'm not sure when I last touched the `clay-panel-variant` mixin. Patterns might have changed a bit since then. We could do:
```
header: (
    collapsed: (),
    c-inner: (),
    link: (),
)
```

pat270: I would have to update the mixin, let me know if you want it done in a separate pr, otherwise I'll update it here.

matuzalemsteles: @pat270 thanks for the information, well if you think this is valid, we can go ahead with it, feel free to send it in this PR or another one, whichever is better.

packages/clay-css/src/scss/variables/_panels.scss
matuzalemsteles: Makes sense, so we can treat this as a bug.

pat270:
@matuzalemsteles I updated the mixin and maps to the new pattern.

matuzalemsteles (approved):
LGTM! Thanks @pat270

PR #4118, opened 2021-06-08
docs(@clayui/drop-down): Add a note about DropDownWithDrillDown menus

### LINKED ISSUE #4110: [DropdownWithDrilldown] The order of the menus affects the behavior of the component

I have two menus: mainMenu and addItemsMenu (this is the child menu). If the child menu is placed first in the object received in the props menus (as shown in the code), the following behavior occurs:
```
<ClayDropDownWithDrilldown
   initialActiveMenu="mainMenu"
   menus={{
	addItemsMenu,
	mainMenu,
   }}
   trigger={
   ...
>
```

https://user-images.githubusercontent.com/9701095/120442820-51c79380-c386-11eb-8188-3c50c80f5345.mov

https://user-images.githubusercontent.com/9701095/120442853-5b50fb80-c386-11eb-9d85-8ee94def0178.mov

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.x |

julien:
Fixes #4110

### DIFF
--- a/packages/clay-drop-down/README.mdx
+++ b/packages/clay-drop-down/README.mdx
@@ -159,6 +159,13 @@ const menus = {
 };
 ```
 
+<div class="clay-site-alert alert alert-warning">
+	An important thing to have in mind, is that the `DropDownWithDrilldown`
+	component, will render its menus in the order that they are specified. This
+	means that if you specify the menus in the wrong order, the menu animation
+	will not behave correctly.
+</div>
+
 ## Caveats
 
 One caveat with the drop down menu is that it is rendered inside of a [React Portal](https://reactjs.org/docs/portals.html) and is rendered directly to the `body` element. This means if you are using the `menuWidth` prop set to `auto`, it will not respect the size of the node parent of the drop down, since the menu is rendered directly to the body.


packages/clay-drop-down/README.mdx
```diff
@@ -162,3 +162,4 @@ const menus = {
 ## Caveats
 
 One caveat with the drop down menu is that it is rendered inside of a [React Portal](https://reactjs.org/docs/portals.html) and is rendered directly to the `body` element. This means if you are using the `menuWidth` prop set to `auto`, it will not respect the size of the node parent of the drop down, since the menu is rendered directly to the body.
+Another important thing to have in mind, is that `<DropDownWithDrilldown />`, will render it's menus in the order that they are specified. This means that if you specify the menus in the wrong order, the menu animation will not behave correctly.
```
matuzalemsteles, line 165: I've been thinking about adding this to the DropDownWithDrilldown section as an alert to make it more visible to people, what do you think?

Something like that:

```html
<div class="clay-site-alert alert alert-warning">
    Another important thing to have in mind, is that <code class="language-text"><DropDownWithDrilldown /></code>, will render it's menus in the order that they are specified. This means that if you specify the menus in the wrong order, the menu animation will not behave correctly.
</div>
```

julien: Hey @matuzalemsteles, I had the idea of using an "alert" but without the code example because it renders something like this and (for me) it's not clear at all 



I read this as "Another important thing to have in mind is that "Click me" ....", ... what is "Click me" ? On top of that to see what we're talking about we're going to have to click on the button to expand the code snippet.

I think something like this would be easier to understand, but tell me what you think when you can

matuzalemsteles: Oh no, sorry this will render the example component, I forgot about the mdx in action here when adding `<DropDownWithDrilldown />` because this is imported above, I think the second image is enough for us to get on with it.

julien: No problem, I thought it was a mistake too. I updated that in [761121e](https://github.com/liferay/clay/commit/761121ee59510731c34e252b1c958269695742d1)

matuzalemsteles (approved):
LGTM!

PR #4119, opened 2021-06-08
fix(@clayui/css): Modal and Cadmin Modal make `.modal-body.inline-scr…

### LINKED ISSUE #4113: Portal CSS modal inline-scroller class to Clay
## This pattern looks like a general need in Liferay portal, can we add it to Clay in some way?

```scss
@media (min-height: 400px) {
	.modal-body.inline-scroller {
		max-height: 320px;
	}
}
```

Right now it is a CSS file included in the module `frontend-css-web`, wich we want to remove

Should be this pattern of max-height validated by Lexicon? @drakonux

pat270:
…oller` `max-height: 320px`

fix(@clayui/css): Modal and Cadmin Modal adds `$modal-body` and `$cadmin-modal-body` Sass maps, respectively

fixes #4113

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_modals.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_modals.scss
@@ -180,17 +180,15 @@
 }
 
 .modal-body {
-	background-color: $cadmin-modal-body-bg;
-	color: $cadmin-modal-body-color;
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
LGTM

Issue #4121, opened 2021-06-09
Add a new type to @clayui/nav in Vertical interface


kevenleone:
## What is your proposal?

Hi everyone, I'm using this @clayui/nav to build a sidebar menu, and I'm facing a Type Error that i want to propose the change.

In this interface

```
interface IItem extends React.ComponentProps<typeof Nav.Item> {
    active?: boolean;
    onClick?: () => void;
    href?: string;
    label?: string;
}
```

The label only accepts a string as value, but in the real scenario, you can pass React Component as well.
I had this particular use case to need to use a React Component in the label spot

```
{
      items: [
        {
          label: i18n.get("scheduled"),
          onClick: () => router.push("/session"),
        },
      ],
      label: (
        <>
          <ClayIcon symbol="home-full" className="mr-2" />
          {i18n.get("sessions")}
        </>
      ),
    },
```

So my suggestion is just to add React.ReactElement along with String type.

```
interface IItem extends React.ComponentProps<typeof Nav.Item> {
    active?: boolean;
    onClick?: () => void;
    href?: string;
    label?: string | React.ReactNode;
}
```

## Why would adopting this proposal be beneficial?

This change can avoid type error, since is already possible to pass a React Component to the label

## What are the alternatives to this proposal?

I don't see any

Issue #4120, opened 2021-06-09
@clayui/nav console warning


kevenleone:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## Is there an example you can provide via codesandbox.com?

https://codesandbox.io/s/youthful-rgb-csfes?file=/src/index.tsx:526-533

## What are the steps to reproduce?

Write some code with `ClayVerticalNav`

## What is the expected result?

Don't see any warn in the console

## Actual Result

Hi everyone ! I've been using this @clayui/nav component, in particular `ClayVerticalNav` and I'm seeing this error in the console

```
Warning: Received `false` for a non-boolean attribute `className`.

If you want to write it to the DOM, pass a string instead: className="false" or className={value.toString()}.

If you used to conditionally omit it with className={condition && value}, pass className={condition ? value : undefined} instead.
    in div (created by Item)
    in Transition (created by CSSTransition)
    in CSSTransition (created by Item)
    in li (created by NavItem)
    in NavItem (created by Item)
    in Item (created by ClayVerticalNav)
    in ul (created by Nav)
    in Nav (created by ClayVerticalNav)
    in div (created by ClayVerticalNav)
    in nav (created by ClayVerticalNav)
    in ClayVerticalNav (at NavComponent.tsx:13)
    in div (at NavComponent.tsx:12)
    in div (at NavComponent.tsx:11)
    in Sidebar (at src/index.tsx:49)
    in div (at src/index.tsx:47)
    in App (at src/index.tsx:57)
    in AppWithIcon (at src/index.tsx:63)
```

## Environment

| Tech  | Version |
| ----- | ------- |
| @clayui/nav  | v3.29.0  |
| React |  16, 17  |

Issue #4117, opened 2021-06-07
Add TreeView Component


matuzalemsteles:
This is a final implementation of #4111 where we need to build a TreeView component, as in the specification it can be very flexible to cover many use cases, so we should build this component in line with the composition like we do with low-level components, but maybe already we can apply what we can learn from #4112 and bring the new patterns to this new component.

The most important part of this component is also dealing with the rendering of large items, we will need the result of the exploration done in #4115 as it may impact the final API design but we can go on developing this component without waiting for the result and we can apply the performance features soon after.

matuzalemsteles:
This issue was divided into other issues to facilitate the work and give more sense of progress since the Tree View spec is quite large and involves more complex parts, the objective of this issue is to deliver an initial API of the components and the composition that allows use as TreeView and flexible to create new variants, the other resources will be created from the delivery of this issue.

- #4115
- #4207
- #4208
- #4209
- #4210

Issue #4112, opened 2021-06-02
Explore the possibility of standardizing collection components


matuzalemsteles:
Well, we currently have several components that are used for listing items of some kind, for example, DropDown, Autocomplete, Breadcrumb, DualListBox, Select, List, MultiSelect, Nav, NavigationBar... we have a lot of components and we need to make them consistent so that the same learning acquired in one component can be repeated when using in other components.

The purpose of this issue is to list APIs in common between the components that list items and also indifferences the idea is to analyze to get an overview of how the current state is and think about how to standardize.

### Goals

- Common APIs and Indifferences
- Component particularities

PR #4123, opened 2021-06-09
fix(@clayui/nav): Adjust condition, to avoid set Boolean values in className property

kevenleone:
Hi everyone, I opened 2 Issues

https://github.com/liferay/clay/issues/4120
https://github.com/liferay/clay/issues/4121

Fixing and Improving some warnings from @clayui/nav

Feel free to suggest changes, thanks!

### DIFF
--- a/packages/clay-nav/src/Vertical.tsx
+++ b/packages/clay-nav/src/Vertical.tsx
@@ -32,7 +32,7 @@ interface IItem extends React.ComponentProps<typeof Nav.Item> {
 	/**
 	 * Value of item.
 	 */
-	label?: string;
+	label?: string | React.ReactNode;
 }
 
 interface IItemWithItems extends IItem {
--- a/packages/clay-nav/src/Vertical.tsx
+++ b/packages/clay-nav/src/Vertical.tsx
@@ -132,7 +132,7 @@ function Item({
 
 			{subItems && (
 				<CSSTransition
-					className={!expanded && 'collapse'}
+					className={expanded ? undefined : 'collapse'}
 					classNames={{
 						enter: 'collapsing',
 						enterActive: `show`,


packages/clay-nav/src/Vertical.tsx
```diff
@@ -132,7 +132,7 @@ function Item({
 
 			{subItems && (
 				<CSSTransition
-					className={!expanded && 'collapse'}
+					className={expanded ? '' : 'collapse'}
```
matuzalemsteles, line 135: To avoid rendering with an empty `class`, we can just add an `undefined` value.

```suggestion
					className={expanded ? undefined : 'collapse'}
```

packages/clay-nav/src/Vertical.tsx
```diff
@@ -32,7 +32,7 @@ interface IItem extends React.ComponentProps<typeof Nav.Item> {
 	/**
 	 * Value of item.
 	 */
-	label?: string;
+	label?: string | React.ReactElement;
```
matuzalemsteles, line 35: I think instead of the `ReactElement` type we should add `ReactNode` instead because it covers more use cases, like text and not necessarily React elements, like `label={<div />}`.

matuzalemsteles: @julien maybe something to keep an eye out for, this is very related to #4112 and #4117. This is important for us to design a more flexible API for collections in general that should reflect in the TreeView since Vertical Navigation will be deprecated in favor of it.

matuzalemsteles (changes requested):
Hey @kevenleone thanks for your contribution, I think it makes sense to allow the label to add an element, this is probably pretty much in line with the vision to be replaced by a Tree View soon. I just left a few comments before going ahead

packages/clay-nav/src/Vertical.tsx
kevenleone: Sounds good !

packages/clay-nav/src/Vertical.tsx
kevenleone: I agree with you, @matuzalemsteles, thanks for the comments.

I'll push the recommended changes, let me know if you need anything else !

Ty

matuzalemsteles (approved):
LGTM

julien (approved):

PR #4126, opened 2021-06-11
docs: Adds RFC process and documentation template

matuzalemsteles:
Well initially we've already made an attempt to document our decisions with the [ADR](https://github.com/joelparkerhenderson/architecture-decision-record) model (#2017) and then moved to the Wiki (#2818) because we don't use it as much or because we don't see more value, but this has changed as Clay has evolved and more and more we need to make more decisions from Clay's future, API decisions... which are reflected in developers, a lot of these decisions are focused on specific issues of a component or a PR that adds a new API that we can ignore in certain moments the impact this has on the experience of using the components and harm consistency, over time this is very easy to be distorted even if someone is from the beginning.

The RFC idea may be similar to the idea of [ADRs](https://github.com/joelparkerhenderson/architecture-decision-record) but it comes closer to the idea of documenting proposals, decisions that have substantial to Clay, not necessarily just architecturally related but also closer to what other projects use, to help in decision making and help the principles of the project.

- React
- Vue
- Gastby
- Swift
- Rust
- NPM
- Ember
- ....

One of the views I have on the things I added in the "Clay strategy and solid foundation" document will be reflected in RFCs generated to discuss, such as #4112.

Leave your thoughts and worries behind.

### DIFF
--- a/docs/0000-proposal-template.md
+++ b/docs/0000-proposal-template.md
@@ -1,0 +1,43 @@
+# Feature name
+
+-   Start Date: (fill me in with today's date, YYYY-MM-DD)
+-   Authors: [Author 1](https://github.com/claydev), [Author 2](https://github.com/claydev)
+-   RFC PR: (leave this empty)
+-   Clay Issue: (leave this empty)
+-   Status: **Awaiting implementation**
+
+_During the review process, add the following fields as needed:_
+
+-   Implementation: [liferay/clay#NNNNN](https://github.com/liferay/clay/NNNNN) or available in version 3.x.x or available under feature flag.
+
+## Introduction
+
+A short description of what the feature is.
+
+## Motivation
+
+Describe the problems this proposal seeks to solve. What use cases does it support? Motivate why this new change would help developers.
+
+It focuses on explaining the motivation so that if this RFC is not accepted, the motivation can be used to develop alternative solutions.
+
+## Proposed solution
+
+Describe your solution to the problem. Take examples and describe how they work. Show how your solution adds value and is better than current alternative solutions: it's cleaner, more flexible...
+
+## Detailed design
+
+Describe the solution design in enough detail for someone familiar with Clay to understand and for someone familiar with the implementation to be able to implement it. If a new API is involved, show the full API and its documentation comments detailing what it does.
+
+## Alternatives considered
+
+Describe alternative approaches to addressing the same problem, and why you chose this approach instead.
+
+## Adoption strategy
+
+If we implement this proposal, how will existing Clay developers adopt it? Is this a significant change? Can we write a codemod? Should we coordinate with other projects?
+
+## How we teach this
+
+What names and terminology work best for these concepts and why? How is this idea best presented?
+
+When this proposal is accepted would it mean that Clay's documentation should be reorganized or changed?
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,52 @@
+# Clay internal documents
+
+It consists of short documents that explain how flows, tools, how processes work on the Clay project.
+
+Documents are not necessarily technical but any document that helps to understand certain motivations or that centralizes information that has a benefit for the entire team.
+
+There are two types of documents, repository infrastructure documents, release cycles or explanations of internal tools, and [RFCs documents](#rfcs) that are used to discuss and centralize "substantial" changes in Clay that reflect on components or documentation that are sourced to users.
+
+## Docs
+
+-   [Developing components](developing_components.md)
+-   [How are clayui.com pages created](creating_pages.md)
+-   [Clay CSS Version History](clay_css_version_history.md)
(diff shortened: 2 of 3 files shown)


matuzalemsteles:
Hey @pat270 maybe this can also help with more complex decisions like cadmin can turn into a proposal with more detail to allow everyone to better understand the reason and the payoffs behind it. What do you think?

docs/README.md
```diff
@@ -0,0 +1,52 @@
...
+The RFC process is a great opportunity to draw more attention to your proposal. Even if the proposals appear to be "obvious" they can be improved significantly, since the team can discuss together with other people and they can give their opinion.
+
+The RFC process can also be helpful to encourage discussions about a proposed feature as it is being designed, and incorporate important constraints into the design while it's easier to change, before the design has been fully implemented.
+
+### What is the process
+
+To add a new feature or make a major change to Clay components or change direction, an RFC must first be submitted and that it first merged as a markdown file. Once merged, the RFC is awaiting implementation.
+
+-   Fork this repository
+-   Copy `0000-proposal-template.md` for `proposals/0000-my-feature.md` (`my-feature` is descriptive)
```
bryceosterhaus, line 40: It might involve less overheard if instead of forking and copying files, they could open an RFC issue. Or maybe "Discussions" would be a worthy feature to try for this?

To enable, go to repo settings
Screen Shot 2021-06-15 at 4 26 19 PM

bryceosterhaus: And an RFC markdown file could be added after the issue is successfully work through.

docs/0000-proposal-template.md
```diff
@@ -0,0 +1,43 @@
+# Feature name
```
bryceosterhaus, line 1: Could this be a github issue template instead? Once it "passes" by the clay team, it can be created as a markdown file.

bryceosterhaus (commented):
Looks good to me! I just had a suggestion on gh templates and possibly use "discussions"

docs/README.md
bryceosterhaus: https://docs.github.com/en/discussions/quickstart

matuzalemsteles: Hey @bryceosterhaus I think it might make sense and we can try maybe the only thing I think is bad about some of them is just the proposal review flow, for example, in "issues" and "Discussions" I see one more initial step to understand what it is the problem of what we create the solid proposal and start to discuss there. With the flow of PRs is that we can select parts of the text and comment on, how to start a conversation on this subject, if I'm not mistaken I think we can even do with "Discussions" but we would have to copy and start a thread below the comment to keep things organized. I saw an Apple use case with Swift, that before creating an RFC you need to discuss this issue in the forum and the RFC header should link to this discussion, I imagine for example issue #4112, trying to collect more information to get have an overview and then open an RFC to discuss the proposal. Does this make sense to you?

docs/0000-proposal-template.md
matuzalemsteles: I think this fits into this conversation https://github.com/liferay/clay/pull/4126#discussion_r652737041

docs/README.md
bryceosterhaus: > I imagine for example issue #4112, trying to collect more information to get have an overview and then open an RFC to discuss the proposal.

This makes sense, it might be fragmented in the RFC if you are pointing to different issue comments. If the RFC was created through an issue, you could still use #4112 as a starting point and then close the issue and point to the official RFC issue.


> With the flow of PRs is that we can select parts of the text and comment on, how to start a conversation on this subject,

Yeah this is definitely a benefit to a PR. My only thought for not using a PR is that it requires the users to take a few additional steps, _clone_, _fork_, _make changes in editor_, and _send pr_. Whereas if it was all contained in the web via issues, they don't have to take those steps. I do think if we can offer the least path of resistance, we will get more feedback from other teams. I assume other developers are a lot like myself _(lazy)_ and that means the more steps I have to take, I am less likely to do it.

Regardless if it is an issue or PR though, I think the proposal process would be helpful and is a good idea and so go with whatever method you think would be best.

matuzalemsteles: Yeah that's true, it might be a bit of friction for some people as it will have to be fork, commit... I think if we want to go with some of them, probably the best scenario would be the "Discussions" for having a better flow to keep conversations organized I think. But anyway I would like to test this with the PR initially and we can add more recommendations to encourage more use.

I imagine who will be using this more would be us than other people are proposing, I would be very happy if someone can contribute to us but I hope it helps us to record the substantial changes and also to think better about them before doing.

matuzalemsteles:
So, I'll continue with that and merge, I expected some more comments but we can discuss this at any time, just to reinforce the idea of this is to focus on substantial value proposals that mean we already know what the problem is and we use the process to describe how we will solve this problem and what the changes and impacts will be.

PR #4127, opened 2021-06-11
fix(@clayui/css): Accessibility change request, overwrite 'outline: 0 !…

georgel-pop-lr:
…important;' with 'box-shadow: $component-focus-box-shadow;' in clay/packages/clay-css/src/scss/bootstrap/_reboot.scss on '[tabindex="-1"]:focus:not(:focus-visible)'

For more details please check: https://issues.liferay.com/browse/LPS-133314

fixes liferay#4124

### DIFF
--- a/packages/clay-css/src/scss/_components.scss
+++ b/packages/clay-css/src/scss/_components.scss
@@ -2,6 +2,8 @@
 	@import url($font-import-url);
 }
 
+@import "components/_reboot.scss";
+
 @import "components/_grid";
 
 @import "components/_alerts";
--- a/packages/clay-css/src/scss/components/_reboot.scss
+++ b/packages/clay-css/src/scss/components/_reboot.scss
@@ -1,0 +1,3 @@
+[tabindex="-1"]:focus:not(:focus-visible) {
+  box-shadow: $component-focus-box-shadow;
+}


pat270:
@georgel-pop-lr It would be better to create a file `_reboot.scss` in https://github.com/liferay/clay/tree/2.x/packages/clay-css/src/scss/components. Then import it on line 4 https://github.com/liferay/clay/blob/2.x/packages/clay-css/src/scss/_components.scss#L4. Then add in `components/_reboot.scss`:

```
[tabindex="-1"]:focus:not(:focus-visible) {
    box-shadow: $component-focus-box-shadow;
}
```

This will probably introduce some regressions where we don't want the focus outline to show (https://github.com/liferay/clay/issues/3436), but we can take it case by case.  That issue made me realize this is also in master at https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_reboot.scss#L162-L164.

pat270 (changes requested):
See comment.

georgel-pop-lr:
> tline to show

@pat270 I made the requested changes,  but I'm just a bit confused of this approach. 

From my understanding, we need here to remove or overwrite this CSS rule 'outline: 0 !important;'. 

How this would be overwritten if we only add ' box-shadow: $component-focus-box-shadow;' ?

Regards.

pat270:
@georgel-pop-lr It doesn't get overwritten, just adds the `box-shadow` to replace the focus outline that is removed.

georgel-pop-lr:
> ocus outline tha

Okay thanks for explaining, got it, I remove this part from bootstrap and added in a separate file, so we have it explicitly there. 

Change pushed, please check.

Regards.

pat270:
@georgel-pop-lr I probably didn't explain it correctly. What you had before this last change is correct. We are adding both styles. The original style from Bootstrap `outline: 0 !important;` removes the focus outline, but we recreate it using `box-shadow` in Clay.

georgel-pop-lr:
> we recreate it using

Hi @pat270,

I think I understood correctly the first time. But not sure how this would make the client's report any better, the tool identifies this rule `outline: 0 !important;` and doesn't like it because the accessibility tool thinks it makes the inputs less accessible, so I think we either need to remove or replace it with other value but not 0 (zero).

Not sure if after this change, the tool would think everything is okay and would pass this accessibility test. 

Regards.

pat270:
@georgel-pop-lr That makes sense. My hunch was the tool is more sophisticated since that was the only violation it caught. We use the `outline: 0` hack everywhere. I think this will pass, but you probably know more about the tool than I do. I'm good with nothing could be done or try to fix.

georgel-pop-lr:
> @georgel-pop-lr That makes sense. My hunch was the tool is more sophisticated since that was the only violation it caught. We use the `outline: 0` hack everywhere. I think this will pass, but you probably know more about the tool than I do. I'm good with nothing could be done or try to fix.

@pat270 Don't know more than you do about the tool, just trying to make my point on this, so we go in the right direction and find the best solution for this case.

I agree that we can try and see if it works with this solution, nothing to lose. 

I will try to explain to the client why we did this as you mention it before "The CSS property outline is very hard to style consistently across browsers which is why box-shadow is used as a hack.".

Regards.

pat270 (approved):

pat270:
This LGTM!

georgel-pop-lr:
> This LGTM!

Hi @pat270,

When the new version would be released? 

Regards.

pat270:
@georgel-pop-lr I think by EOD tomorrow.

PR #4137, opened 2021-06-16
feat(@clayui/css): Adds Cadmin Treeview

### LINKED ISSUE #4111: Add CSS to TreeView Component
The Lexicon team has developed a new component to handle TreeView that covers many use cases as well as deprecating Vertical Navigation and using TreeView instead. We must implement it on Clay's side, initially, the epic is under analysis but we can start the CSS development and later investigate how we will create the component and solve some problems that led to its creation.

- [Epic](https://issues.liferay.com/browse/LPS-133328)
- [Spec](https://docs.google.com/document/d/1fggZBRGAIkGPbnkRrGEUOT0hHpkF-YjFmREee0XCoI0/edit)

pat270:
I'm not 100% happy with this implementation. The markup is huge since I'm trying to cover all use cases. I tried to pare it down, but couldn't find a good solution for icon placement and working with all the different combinations of html elements (`a`, `button`). If the markup causes laggy rendering issues in the React component, I was thinking about adding a version with lighter markup for specific cases. 🤷 

Also we need the Cadmin pr to get merged in Portal before we can use this component. I can focus on getting that merged next. We can break things now.

treeview

fixes #4111

### DIFF
--- a/clayui.com/content/docs/components/markup-treeview.md
+++ b/clayui.com/content/docs/components/markup-treeview.md
@@ -1,0 +1,7751 @@
+---
+title: 'Treeview'
+description: 'A tree view is a component-based on nodes that are shown in a hierarchical structure.'
+lexiconDefinition: 'https://liferay.design/lexicon/core-components/tree-view/'
+---
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
+-   [Treeview](#css-treeview)
+    -   [Group](#css-treeview-group)
+    -   [Item](#css-treeview-item)
+    -   [Link](#css-treeview-link)
+    -   [Component Expander](#css-treeview-component-expander)
+    -   [Component Action](#css-treeview-component-action)
+-   [Variations](#css-variations)
+    -   [Expander on Hover](#css-treeview-expander-on-hover)
+    -   [Light](#css-treeview-light)
+    -   [Dark](#css-treeview-dark)
+
+</div>
+</div>
+
+<div class="clay-site-alert alert alert-warning">This component is a Cadmin only component.</div>
+
+The component must be wrapped in:
+
+```html{expanded}
+<div class="cadmin">
+	...
+</div>
+```
+
+## Treeview(#css-treeview)
+
+The treeview provides a way to display information in a hierarchical structure by using collapsible items (nodes). You can navigate between these items using either your mouse device or keyboard. The most common example of a treeview is a folder structure for file systems, but it can be used for showing any hierarchical relationships.
+
+<div class="clay-site-alert alert alert-info">
+	See <a href="https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html" rel="noreferrer noopener" target="_blank">https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html</a> for accessibility patterns.
+</div>
+
+<div class="clay-site-alert alert alert-warning">
+	Treeview Link indentation must be provided by javascript through inline styles, use the styles `padding-left: 24px;` on `.treeview-link` and `margin-left: -24px` on `.treeview-link > .c-inner`. Increase the indentation for each level by increasing or decreasing `padding-left` or `margin-left` by 24px.
+</div>
+
+<div class="cadmin">
+	<ul class="treeview treeview-light treeview-nested" role="tree">
+		<li class="treeview-item" role="none">
+			<div aria-controls="treeviewCollapse01" aria-expanded="true" class="treeview-link" data-target="#treeviewCollapse01" data-toggle="collapse" role="treeitem" tabindex="0">
+				<span class="c-inner" tabindex="-2">
+					<span class="autofit-row">
+						<span class="autofit-col">
+							<button aria-controls="treeviewCollapse01" aria-expanded="true" class="btn btn-monospaced component-expander" data-target="#treeviewCollapse01" data-toggle="collapse" tabindex="-1" type="button">
+								<span class="c-inner" tabindex="-2">
+									<svg class="lexicon-icon lexicon-icon-angle-down" focusable="false" role="presentation">
+										<use xlink:href="/images/icons/icons.svg#angle-down" />
+									</svg>
+									<svg class="lexicon-icon lexicon-icon-angle-right component-expanded-d-none" focusable="false" role="presentation">
+										<use xlink:href="/images/icons/icons.svg#angle-right" />
+									</svg>
+								</span>
+							</button>
+						</span>
+						<span class="autofit-col">
+							<span class="component-icon">
+								<svg class="lexicon-icon lexicon-icon-folder" focusable="false" role="presentation">
+									<use xlink:href="/images/icons/icons.svg#folder" />
+								</svg>
+							</span>
+						</span>
+						<span class="autofit-col autofit-col-expand">
+							<span class="component-text">
+								<span class="text-truncate-inline" title="Liferay Drive"><span class="text-truncate">Liferay Drive</span></span>
+							</span>
+						</span>
+						<span class="autofit-col">
+							<button class="btn btn-monospaced component-action" tabindex="-1" type="button">
+								<span class="c-inner" tabindex="-2">
+									<svg class="lexicon-icon lexicon-icon-times-circle-full" focusable="false" role="presentation">
+										<use xlink:href="/images/icons/icons.svg#times-circle-full" />
+									</svg>
+								</span>
+							</button>
+						</span>
+						<span class="autofit-col">
+							<button class="btn btn-monospaced component-action" tabindex="-1" type="button">
+								<span class="c-inner" tabindex="-2">
+									<svg class="lexicon-icon lexicon-icon-ellipsis-v" focusable="false" role="presentation">
+										<use xlink:href="/images/icons/icons.svg#ellipsis-v" />
+									</svg>
+								</span>
+							</button>
+						</span>
+					</span>
+				</span>
+			</div>
+			<div class="collapse show" id="treeviewCollapse01">
+				<ul class="treeview-group" role="group">
+					<li class="treeview-item" role="none">
+						<div aria-controls="treeviewCollapse02" aria-expanded="false" class="collapsed treeview-link" data-target="#treeviewCollapse02" data-toggle="collapse" role="treeitem" tabindex="-1" style="padding-left:24px;">
+							<span class="c-inner" tabindex="-2" style="margin-left:-24px;">
+								<span class="autofit-row">
+									<span class="autofit-col">
+										<button aria-controls="treeviewCollapse02" aria-expanded="false" class="btn btn-monospaced component-expander" data-target="#treeviewCollapse02" data-toggle="collapse" tabindex="-1" type="button">
+											<span class="c-inner" tabindex="-2">
+												<svg class="lexicon-icon lexicon-icon-angle-down" focusable="false" role="presentation">
+													<use xlink:href="/images/icons/icons.svg#angle-down" />
+												</svg>
+												<svg class="lexicon-icon lexicon-icon-angle-right component-expanded-d-none" focusable="false" role="presentation">
+													<use xlink:href="/images/icons/icons.svg#angle-right" />
+												</svg>
+											</span>
+										</button>
+									</span>
+									<span class="autofit-col">
+										<span class="component-icon">
+											<svg class="lexicon-icon lexicon-icon-folder" focusable="false" role="presentation">
+												<use xlink:href="/images/icons/icons.svg#folder" />
+											</svg>
+										</span>
+									</span>
+									<span class="autofit-col autofit-col-expand">
+										<span class="component-text">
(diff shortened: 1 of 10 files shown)


julien:
Looks nice

packages/clay-css/src/scss/cadmin/components/_treeview.scss
```diff
@@ -0,0 +1,324 @@
...
+	}
+}
+
+.treeview-action-btn {
+	$action-btn: setter(map-get($cadmin-treeview, action-btn), ());
+
+	@include clay-button-variant($action-btn);
+}
+
+.treeview-expander-btn {
```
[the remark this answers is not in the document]
matuzalemsteles: Apparently this also only needs to be shown when hovering.

matuzalemsteles (commented):
Hey @pat270 looks like a great job, I added some comments about some issues. But a few concerns to keep in mind:

This component just for `cadmin` worries me a little about how we document it, I get the feeling that our documentation is for anyone to use, and when I see that this is just for `cadmin` I think it would be something internal that people don't should use it but at the same time we should document it as part of our documentation because it's used by the product teams, I'm not sure it would be a good idea to just maintain specific components for a theme.

I keep in mind that `cadmin` would be more of a theme, I'm not sure it would be the most ideal term.

Another question is regarding the change from `dark` to `light`, I'm noticing that we're starting to adopt it more often and we're usually dealing with this in each component because that way we can isolate, I'm wondering if we can draw a strategy for this to avoid dealing with each component, what I have in mind is:

- Avoid adding colors directly to a specific class.
- Keep the class strategy to switch between the two color themes, like a `.dark`
- Create a pattern to switch between colors, for example, we create a class for colors and their values ​​in dark mode.
  - For example, we have `.tone-text-primary`, `.tone-bg-primary`

```html
<ul class="treeview dark">
  <li class="treeview-item">
    <div class="treeview-link">
      <span class="autofit-row">
        <span class="autofit-col">
          <span class="text-truncate-inline tone-text-secondary-l1">...</span>
        </span>
      </span>
    </div>
  </li>
</ul>
```

We can handle this in two ways, one via converting the color via css when the dark scope is added to the component and we change it to the equivalent color, another would be to change colors manually by controlling via JS, for example according to the property, we would change to another class that matches the color. This is just an idea to try and reuse more in our components, let me know what you think.

packages/clay-css/src/scss/cadmin/components/_treeview.scss
```diff
@@ -0,0 +1,324 @@
...
+	$icon: setter(map-get($cadmin-treeview, icon), ());
+
+	@include clay-css($icon);
+}
+
+.treeview-text {
+	$text: setter(map-get($cadmin-treeview, text), ());
+
+	@include clay-css($text);
+}
```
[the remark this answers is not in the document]
pat270: This is a good point, we do have some reusable classes `component-action`, `component-link`, `component-title`, and `component-subtitle`. I can add `component-text` and maybe some others?

packages/clay-css/src/scss/cadmin/components/_treeview.scss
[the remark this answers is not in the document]
pat270: Not a problem, we could change this to `btn-treeview-expand` or possible create a `component-???`. From my understanding from the specs, it just needs to change color on hover?

packages/clay-css/src/scss/cadmin/components/_treeview.scss
```diff
@@ -0,0 +1,324 @@
...
+
+.treeview-nested {
```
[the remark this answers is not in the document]
pat270: `treeview-nested` goes 4/5 levels depending where level 1 starts. I can't add N levels because the CSS gets too large. This is due to using padding to create the nesting. We can create N levels using margin without large CSS, but requires some pseudo element hacks. I wanted to leave those open for end user customizations. I can look into this more.

pat270:
> Keep the class strategy to switch between the two color themes, like a .dark

We can start exploring this. This is the pattern from Bootstrap so I thought it was best to keep it consistent. I'm in the camp of providing both options. The `treeview` component by itself is unstyled, adding a `dark` and `light` selector wouldn't increase the size at all.

packages/clay-css/src/scss/cadmin/components/_treeview.scss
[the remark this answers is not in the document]
matuzalemsteles: I think it might be a good idea, we can see how it looks.

packages/clay-css/src/scss/cadmin/components/_treeview.scss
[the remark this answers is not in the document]
matuzalemsteles: I think if that's too specific, then we go to `btn-treeview-expand` if not, ok, creating a generic `component-` might make sense.

Yes, this needs to change color on mouseover too, but when [hovering over the container, show the expand](https://docs.google.com/document/d/1fggZBRGAIkGPbnkRrGEUOT0hHpkF-YjFmREee0XCoI0/edit#heading=h.fc4dekozxsni) too.

packages/clay-css/src/scss/cadmin/components/_treeview.scss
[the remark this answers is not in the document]
matuzalemsteles: Hmm, ok, I think understand, If we need N levels, we add `.treeview-nested` in the depth? Can we maybe move the padding to `.treeview-group`, as it's being nested, adding `24px`, or would that be limiting because of hovering on the items?

matuzalemsteles:
> We can start exploring this. This is the pattern from Bootstrap so I thought it was best to keep it consistent. I'm in the camp of providing both options. The treeview component by itself is unstyled, adding a dark and light selector wouldn't increase the size at all.

I see, I think we can go with this as it's a pattern for other components, I'm thinking more about trying to reuse this for other components or even offering a list of color classes... We can think of some ideas and then come up with a solution safe for us to keep.

pat270:
@matuzalemsteles I can't create a treeview to accommodate N levels and adhere to Lexicon specs solely with CSS. The issue is hover, active, focus states for nested items that extend the full width of the treeview component combined with collapsible items with animation. 

treeview-link
It causes style shifting on collapse. The best we can do is provide 5, 10, 15, 20 or whatever number of levels that covers most cases.

N levels is easy if we don't care about the styles expanding the whole container. See screenshot below:

treeview-n-levels

matuzalemsteles:
@pat270 I understand, so this also affects the `expand` animation right? I'm not sure if it's very safe to offer a limited amount of levels, as it can be used in Page Editor, Categories... which can have many levels easily, I don't know if it would be very safe. Hey @liferay-lexicon can you guys help with that? we have some limitations to reconcile the levels with hover, focus...

pat270:
@matuzalemsteles yes, expand too.

drakonux:
Hey guys, well, I don't know how to help you, so ask us specific questions to discuss options with @hold-shift.

Sadly I can't give you a maximum number of levels, because I can't limit how our customers are going to use it. And, I would say that to set that number is your responsibility and it should be based on a technical reason. And it should be as high as you can, and Clay/Lexicon should agree on a solution together to see what answer (interaction/message/whatever) we give when a super deep tree-view is loaded.

About light/dark schemes on components, I think we proposed a dark scheme version for Vertical Bar in the past, but potentially many components could have a dark mode. So, it could be something to work if you think it makes sense. We believe it would be handy for dark backgrounds such as DXP currently has with the new image editor (buttons in a dark background) or the new proposals for Application Menu (that have many buttons, selectors and more basic components within).

pat270:
@matuzalemsteles I updated with generic class names and added the hover thing. I'll update the nesting stuff when we get confirmation from the Lexicon team.

matuzalemsteles:
Thanks @drakonux! Well I think we're going to have to pick a middle ground, it's very hard to know what levels the client can create, setting a specific amount we'll probably be adding more code and it gets more complex, right @pat270? Maybe instead of keeping the hover end-to-end, can we just keep like in @pat270's screenshot? so we could keep N levels, but maybe we still have to deal with the animation.

drakonux:
Ok, so we would have a dead interaction area on the left of each node. However, I'm not very happy with it. More levels, more indentation, and thus more problems interacting with the nodes. 

If you can show me a POC @pat270, maybe I change my mind, or at least I would figure it out better.

matuzalemsteles:
Yeah, this might be an interaction issue, @pat270 we might be able to set the left pixel values via JavaScript, that might also be an option.

pat270:
@drakonux Yes, there will be a dead area on the left. I just put up a demo at https://pat270.github.io/clay3-test-site/vclay-treeview/content/cadmin-treeview/. It's the first example.

@matuzalemsteles that might be the easiest way to get it to work. I'll let you know if I can think of anything else.

julien:
>  I just put up a demo at https://pat270.github.io/clay3-test-site/vclay-treeview/content/cadmin-treeview/. It's the first example.



:joy: this made my day

pat270:
@julien The credit goes to the late @natecavanaugh. He replaced my random hash file name with that text. RIP brother!

drakonux:
Hey, guys sorry for the delay. I see the right interaction in the POC @pat270 !! Did you be able to do it? 

no-dead-zone

matuzalemsteles:
Hey @drakonux, I think it's the first TreeView on the page 😅

brunofernandezg:
Does this mean that the proposed solution will have the "dead zone" in the left or not?

cc. @liferay-lexicon

drakonux:
Hey guys, you know that normally we are flexible with adapt the implementation if we find problems building a component or solving a problem, but in this case, we think it's important to have that "dead area" operative. So we don't see to keep it as a good option, could you invest more time to have it? or it is what it is.

nhpatt:
What do you think @pat270, does it makes sense to investigate it further? or we modify it via JS?

pat270:
@nhpatt This makes sense, adding padding inline via JS is the way to go.

packages/clay-css/src/scss/cadmin/components/_treeview.scss
```diff
@@ -0,0 +1,421 @@
...
+		.treeview-group
+		.treeview-group
+		> .treeview-item
+		> .treeview-link {
+		padding-left: 96px;
+
+		> .c-inner {
+			margin-left: -96px;
+		}
+	}
```
matuzalemsteles, line 230: Since we're going to deal with this from the JS side, would you be better off removing `treeview-nested` using `padding-left` and `margin-left`? so it is easier to deal with instead of creating a logic to add these values only at level 5 or more for example.

matuzalemsteles (commented):
I think we're getting there to merge this, just want to see if we can remove `treeview-nested` for easier JS control or remove some duplication.

Also one last thing that would be interesting is if we could keep the even spacing between the elements inside the TreeView, as each one has a different spacing made us have specific classes for `icon`, `text`, `action` and for `item`. @drakonux could these spacing for these specific elements be uniform? @pat270 if the spacings were uniform would we be able to remove these extra classes?

packages/clay-css/src/scss/cadmin/components/_treeview.scss
pat270: @matuzalemsteles Yes, removing all the `padding-left` and `margin-left` would be the right thing to do. I think we should still keep `treeview-nested` though.

matuzalemsteles: @pat270 no problem, we can go with it.

pat270: I found a hack I could use. We can do this without adding inline styles via JS. I'll update the pr today.

matuzalemsteles: Oh great 🤩

pat270: @matuzalemsteles I failed, we need to go back to original plan with JS.

matuzalemsteles: 😢

clayui.com/content/docs/components/markup-treeview.md
```diff
@@ -0,0 +1,7751 @@
...
+<div class="clay-site-alert alert alert-info">
+	See <a href="https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html" rel="noreferrer noopener" target="_blank">https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html</a> for accessibility patterns.
+</div>
+
+<div class="clay-site-alert alert alert-warning">
+	Treeview Link indentation must be provided by javascript through inline styles, use the styles `padding-left: 24px;` on `.treeview-link` and `margin-left: -24px` on `.treeview-link > .c-inner`. Increase the indentation for each level by increasing or decreasing `padding-left` or `margin-left` by 24px.
+</div>
+
+<div class="cadmin">
+	<ul class="treeview treeview-light treeview-nested" role="tree">
```
matuzalemsteles, line 47: I saw that you changed the name of the class `treeview-nested` to `treeview-nested-margins`, should we replace it here too?

matuzalemsteles (commented):
Hey @pat270 I think overall it's good, great work! I just added a comment to check before going with it.

clayui.com/content/docs/components/markup-treeview.md
pat270: @matuzalemsteles `treeview-nested-margins` is the type with the dead space on the left. I left it in as an alternative. `treeview-nested` is the correct name for the version that uses padding. It gives the option to set padding via CSS if for whatever reason a dev can't use the React plugin.

matuzalemsteles: Makes sense. Let's merge this. Do you have anything else to work on in the Treeview?

pat270: Nothing else, I'm going to merge this.

Issue #4139, opened 2021-06-17
<ClayCardWithHorizontal/> spritemap should be optional


freddi301:
-   [x] I have searched the [issues](https://github.com/liferay/clay/issues) of this repository and believe that this is not a duplicate.

## What are the steps to reproduce?

```jsx
<ClayIconSpriteContext.Provider value={spritemap}>
  <ClayCardWithHorizontal />
</ClayIconSpriteContext.Provider>
```

## What is the expected result?

No missing property error

## Environment

`package.json`
```json
{
  "dependencies": {
     "@clayui/card": "^3.30.0",
     "@clayui/css": "^3.29.0",
     "@clayui/icon": "^3.1.0"
  }
}

PR #4140, opened 2021-06-17
fix(@clayui/toolbar): spritemap should be optional

bryceosterhaus:
fixes https://github.com/liferay/clay/issues/4139

saw the issue and was an easy fix so thought I would send over

### DIFF
--- a/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
+++ b/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
@@ -99,7 +99,7 @@ interface IProps extends React.ComponentProps<typeof PaginationBar> {
 	/**
 	 * Path to spritemap from clay-css.
 	 */
-	spritemap: string;
+	spritemap?: string;
 
 	/**
 	 * The total number of items in the pagination list.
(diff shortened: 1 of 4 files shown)


packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -113,95 +113,96 @@ const DEFAULT_LABELS = {
...
-
-		if (!href) {
-			item.onClick = () => {
-				if (onDeltaChange) {
-					onDeltaChange(label as number);
-				}
-			};
+export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> =
+	({
+		activeDelta,
```
matuzalemsteles, line 118: This is just format, right? I don't see any other changes here

matuzalemsteles (approved):
LGTM, thanks @bryceosterhaus

bryceosterhaus:
@matuzalemsteles my bad. I think I had a different version of prettier in my editor and it formatted on save and I didn't take a look before sending since I only added `?`.

I sent an update

matuzalemsteles:
Thank you @bryceosterhaus, don't worry!

PR #4150, opened 2021-06-23
fix(@clayui/css): Removes nested `setter` functions, only need to cal…

### LINKED ISSUE #4146: @clayui/css Update places where we nest multiple `setter` functions to use a single function
The `setter` function was updated in https://github.com/liferay/clay/issues/4076 to accept an Argslist as a parameter. We can pass in multiple Sass maps now. We need to update the places where we nest `setter` to only call the function once.

pat270:
…l it once.

fixes #4146


matuzalemsteles (approved):
LGTM

PR #4156, opened 2021-06-28
chore(@clayui/color-picker): upgrade tinycolor2 to 1.4.2

### LINKED ISSUE #4155: ClayColorPicker: Upgrade tinycolor2 to 1.4.2
From https://github.com/liferay-frontend/liferay-portal/pull/1201.

The new version removes jquery dependencies, so it's a good reason to update on Clay's side. Also Bryce brought up the idea that we can remove this dependency on Clay, it's also an option we can evaluate, I'll create this issue just to update and keep away from jquery 🙂 and we can evaluate the effort to remove tinycolor2 and implement our own solution.

matuzalemsteles:
Fixes #4155

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3737,7 +3737,7 @@
     "@types/react-dom" "*"
     "@types/testing-library__dom" "*"
 
-"@types/tinycolor2@^1.4.1":
+"@types/tinycolor2@^1.4.2":
   version "1.4.2"
   resolved "https://registry.yarnpkg.com/@types/tinycolor2/-/tinycolor2-1.4.2.tgz#721ca5c5d1a2988b4a886e35c2ffc5735b6afbdf"
   integrity sha512-PeHg/AtdW6aaIO2a+98Xj7rWY4KC1E6yOy7AFknJQ7VXUGNrMlyxDFxJo7HqLtjQms/ZhhQX52mLVW/EX3JGOw==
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -23803,6 +23803,11 @@ tinycolor2@^1.4.1:
   resolved "https://registry.yarnpkg.com/tinycolor2/-/tinycolor2-1.4.1.tgz#f4fad333447bc0b07d4dc8e9209d8f39a8ac77e8"
   integrity sha1-9PrTM0R7wLB9TcjpIJ2POaisd+g=
 
+tinycolor2@^1.4.2:
```
bryceosterhaus, line 23806: Why was our old version not in the yarn.lock?

bryceosterhaus: Also... Where is the jquery dependency? I was just looking at https://bundlephobia.com/package/tinycolor2@1.4.1

bryceosterhaus: Looks like the jquery dependency was just for its demo. So I don't think it was actually packaged into our bundles

matuzalemsteles: We still have a dependency on 1.4.1 because the storybook uses a component called react-color which depends on the version of tinycolor2 in version 1.4.1 🙂, we can evaluate updating the storybook later to harvest new features and maybe not depend on 1.4.1 anymore. As Storybook dependencies don't impact in the end, so we don't have any problems.

matuzalemsteles: Apparently, it actually doesn't seem to have a jquery dependency but I can only see jquery removal in demos https://github.com/bgrins/TinyColor/commit/250a1e2421242b336770d50c2b5e1eae292bc727 🤷‍♂️

matuzalemsteles: Your last message only appeared to me later 🤦‍♂️

bryceosterhaus:
lgtm!

julien (approved):

PR #4157, opened 2021-06-28
feat(@clayui/list): Add `forwardRef` to ClayList components

### LINKED ISSUE #4149: ClayList.ItemTitle should accept a ref
We should add forwardRef to ClayList.ItemTitle. I found this by using Next.js' Link component which requires the child to accept a ref when using the `passHref` prop

```jsx
import Link from 'next/link';
import ClayList from '@clayui/list';

<Link href="/foo/bar" passHref>
  <ClayList.ItemTitle>test</ClayList.ItemTitle>
</Link>;
```

Try to pass a ref attribute to 

Ref does not get applied to element

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.x  |

matuzalemsteles:
Fixes #4149

Well, I went ahead and also added `forwardRef` to the other ClayList components, we only had a few, so it might be good practice to add to the others.

@bryceosterhaus I added it to the `ClayList.ItemTitle` component but I'm not sure if that solves your problem, from what I understand from the `next/link` it will pass the `href` to the `children`, in this case we have a `div` that wraps `children` of `ClayList.ItemTitle` or if it is a Link.

### DIFF
--- a/packages/clay-list/src/Header.tsx
+++ b/packages/clay-list/src/Header.tsx
@@ -6,18 +6,20 @@
 import classNames from 'classnames';
 import React from 'react';
 
-const ClayListHeader: React.FunctionComponent<React.HTMLAttributes<
-	HTMLLIElement
->> = ({children, className, ...otherProps}) => {
+const ClayListHeader = React.forwardRef<
+	HTMLLIElement,
+	React.HTMLAttributes<HTMLLIElement>
+>(({children, className, ...otherProps}, ref) => {
 	return (
 		<li
 			{...otherProps}
 			className={classNames('list-group-header', className)}
+			ref={ref}
 		>
 			<p className="list-group-header-title">{children}</p>
 		</li>
 	);
-};
+});
 
 ClayListHeader.displayName = 'ClayListHeader';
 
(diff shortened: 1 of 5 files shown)


bryceosterhaus:
@matuzalemsteles oh interesting! In that case, would it make sense to add an "as" or "containerElement" prop where you can pass in either an html element or a custom component?

The other option is for me to use use the css class which isn't terrible.

matuzalemsteles:
@bryceosterhaus hmm, I've been taking a look at the `next/link` implementation, it's a little confusing to understand but apparently if you pass `passHref`, it will add `href` as a property, so in that case I think it will work , since our component accepts `href` property and handles it.

https://github.com/vercel/next.js/blob/canary/packages/next/client/link.tsx#L320, The use of `ref` is not clear to me, but it seems to use it for other interaction cases.

bryceosterhaus:
@matuzalemsteles I realized I could've just tested this in a sandbox... It seems to work with the forward ref now and doesn't complain. You can remove the forward ref and see the console error

https://codesandbox.io/s/nice-frog-k6p05?file=/src/index.js

matuzalemsteles:
Oh perfect, seems to work fine so, ok, so let's get on with it.

PR #4160, opened 2021-06-30
fix(@clayui/list): fix error when building types

matuzalemsteles:
When building types, this error appears: "Default export of the module has or is using private name 'IItemProps'." Probably a regression of #4157 or because we are exporting the types in another way with `forwardRef`, strange thing is that CI didn't catch that, I'll check later if we're checking the types.


matuzalemsteles:
This is not trivial, so I'm going to merge to get to the release

PR #4168, opened 2021-07-15
fix(@clayui/css): Forms `form-group-sm div.form-control` should be 32px tall and grow to fix content inside, similar to a textarea element

### LINKED ISSUE #4164: @clayui/css form-group-sm with contenteditable form-control doesn't expand
form-group-sm

The Markup:
```html
<div class="form-group-sm">
  <div contenteditable="true" class="form-control"><div>
</div>
```

Some notes: we need to reset `min-height: 0` and set `height: auto` for `form-control` in `form-group-sm`.

pat270:
fixes #4164

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_forms.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_forms.scss
@@ -178,6 +178,12 @@ label {
 		width: 0;
 	}
 
+	// CKEditor markup work around
+
+	&[contenteditable] p {
+		margin-bottom: 0;
+	}
+
 	.label {
 		@include clay-label-size($cadmin-form-control-label-size);
 	}
@@ -726,6 +732,12 @@ textarea.form-control-sm,
 	.form-control,
 	.form-control-plaintext {
 		@extend %clay-form-control-sm;
+	}
+
+	div.form-control,
+	.form-control[contenteditable] {
+		height: auto;
+		min-height: $cadmin-input-height-sm;
 	}
 
 	select.form-control {
--- a/packages/clay-css/src/content/form_group.html
+++ b/packages/clay-css/src/content/form_group.html
@@ -202,3 +202,19 @@ section: Components
 		</div>
 	</div>
 </div>
+
+<div class="clay-site-row-spacer row">
+	<div class="col-md-12">
+		<h3>Form Group Sm with div.form-control</h3>
+
+		<div class="form-group-sm">
+			<label>This is a fake input, just for show</label>
+			<div class="form-control"></div>
+		</div>
+
+		<div class="form-group-sm">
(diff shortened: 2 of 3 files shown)


packages/clay-css/src/scss/cadmin/components/_forms.scss
```diff
@@ -178,6 +178,12 @@ label {
 		width: 0;
 	}
 
+	// CKEditor markup work around
+
+	&[contenteditable] p {
+		margin-bottom: 0;
+	}
```
matuzalemsteles, line 185: Hey @pat270, this is a little strange to keep on Clay's side, in that case wouldn't it be better to keep it in the CKEditor skins we have in DXP? I'm not sure if there were any limitations

matuzalemsteles (commented):
LGTM, just a little concerned about keeping a CKEditor workaround here

packages/clay-css/src/scss/cadmin/components/_forms.scss
pat270: @matuzalemsteles This is a good point. I'm not sure where the right place would be. This markup is generic enough that it makes sense to include it in Clay. Clay is adding the margin so makes sense to remove here? :shrug:

matuzalemsteles: I imagine that despite being generic we are only covering for the case of CKEditor, as in the comment above this caught my attention 😅, probably if it's something very specific for CKEditor and Clay to work together I think this should live in the skins of CKEditor, I'm not sure if the `mono-lexicon` skins are in DXP or ckeditor's repo. But if you think this shouldn't be an issue for non-CKEditor cases we can go with that.

pat270: I just remembered, CKE does take care of that. Cadmin increased the specificity of our styles and overrides the CKE styles. We should keep it here.

matuzalemsteles: Hmm, Just for me to understand, CKE adding some margin, is this intentional for some use cases? For example, aren't all CKE implementations using `form-control`, or just a few? if only a few use it, it might make sense to keep it here. Since without the `form-control` the CKE must need the margin.

pat270: Browsers add `margin-bottom: 1em` by default. I believe our version of CKE uses `form-control`? I doubt they would use Bootstrap classes by default unless it was a coincidence. But I agree lets just keep it here.

matuzalemsteles: Oh, I see, okay. Let's get on with it!

PR #4177, opened 2021-07-21
feat(@clayui/css): Buttons use nested property names in variables to stay as close as possible to the way our CSS is written

### LINKED ISSUE #4176: @clayui/css Atlas Button variables are using deprecated property names
https://github.com/liferay/clay/blob/b387faaabe638447d4a1d3e4a68869355b45b6d8/packages/clay-css/src/scss/atlas/variables/_buttons.scss#L50-L69

Property names like `hover-bg`, `focus-box-shadow` are deprecated. We should stay consistent with CSS property names in our Sass map variables.

See https://github.com/liferay/clay/issues/3987 for some work on deprecated property names.

pat270:
fixes #4176

I still need to document our mixins and functions. I'd like to use Sassdoc but haven't found a way to convert it to clayui.com. I can document directly in the source using comments which is nice for someone combing through the code. We should probably discuss this in a separate issue.

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_buttons.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_buttons.scss
@@ -50,20 +50,26 @@ $btn-primary: () !default;
 $btn-primary: map-deep-merge(
 	(
 		box-shadow: $c-unset,
-		hover-bg: $primary-d1,
-		hover-border-color: transparent,
 		hover: (
+			background-color: $primary-d1,
 			background-image: clay-enable-gradients($primary-d1),
+			border-color: transparent,
 		),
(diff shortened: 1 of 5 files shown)


matuzalemsteles (approved):
LGTM! For the documentation, it can be a good start in the code itself and we can create an issue to try to use sassdoc or another tool that can generate for us some kind of metadata that we can use to add to clayui.com, like what we do for the API Table of React components.

matuzalemsteles:
Hey @pat270 I'm going to merge this and create an issue so we can explore this possibility.

pat270:
Thanks @matuzalemsteles !

PR #4178, opened 2021-07-22
feat(@clay/pagination-bar): Make `deltas` optional

### LINKED ISSUE #4175: [ClayPaginationBar Taglib] Make paginationBarDeltas attribute optional
## What is your proposal?
Due to this [LPS-123825](https://issues.liferay.com/browse/LPS-123825), we propose to make the `paginationBarDeltas` attribute optional and consequently the dropdown is hidden.

The `activeDelta` parameter (items per page), can be obtained from other source and doesn't necessarily has to come from the delta array `paginationBarDeltas`. For this reason it should make sense that displaying the dropdown menu with the `deltas` is optional.

Currently, there are no alternatives.

julien:
With this change, we only render the DropDown menu when the `deltas`
prop is passed

Fixes #4175

### DIFF
--- a/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
+++ b/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
@@ -116,7 +116,7 @@ const DEFAULT_LABELS = {
 export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> = ({
 	activeDelta,
 	activePage = 1,
-	deltas = defaultDeltas,
+	deltas,
 	disabledPages,
 	ellipsisBuffer,
 	hrefConstructor,
@@ -127,6 +127,13 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> = 
 	totalItems,
 	...otherProps
 }: IProps) => {
+	let showDropDown = true;
+
+	if (!deltas) {
+		deltas = defaultDeltas;
+		showDropDown = false;
+	}
+
 	if (!activeDelta) {
 		activeDelta = deltas[0].label;
 	}
@@ -162,22 +169,24 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> = 
 
 	return (
 		<PaginationBar {...otherProps}>
-			<PaginationBar.DropDown
-				items={items}
-				trigger={
-					<ClayButton
-						data-testid="selectPaginationBar"
-						displayType="unstyled"
-					>
-						{sub(labels.perPageItems, [activeDelta])}
+			{showDropDown && (
+				<PaginationBar.DropDown
+					items={items}
+					trigger={
+						<ClayButton
+							data-testid="selectPaginationBar"
+							displayType="unstyled"
+						>
+							{sub(labels.perPageItems, [activeDelta])}
 
-						<ClayIcon
-							spritemap={spritemap}
-							symbol="caret-double-l"
-						/>
-					</ClayButton>
-				}
-			/>
+							<ClayIcon
+								spritemap={spritemap}
+								symbol="caret-double-l"
+							/>
+						</ClayButton>
+					}
+				/>
+			)}
 
 			<PaginationBar.Results>
 				{sub(labels.paginationResults, [
--- a/packages/clay-pagination-bar/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-pagination-bar/src/__tests__/__snapshots__/index.tsx.snap
@@ -6,26 +6,6 @@ exports[`ClayPaginationBar renders 1`] = `
     class="pagination-bar"
   >
     <div
-      class="dropdown pagination-items-per-page"
-    >
-      <button
-        class="dropdown-toggle btn btn-unstyled"
-        data-testid="selectPaginationBar"
-        style=""
-        type="button"
-      >
-        10 items
-        <svg
-          class="lexicon-icon lexicon-icon-caret-double-l"
-          role="presentation"
-        >
-          <use
-            xlink:href="path/to/spritemap#caret-double-l"
-          />
-        </svg>
-      </button>
-    </div>
-    <div
       class="pagination-results"
     >
       Showing 1 to 10 of 100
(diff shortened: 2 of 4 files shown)


julien:
@matuzalemsteles just so you know, the change for the tag has also been sent [here](https://github.com/liferay-frontend/liferay-portal/pull/1276)

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -127,6 +127,13 @@ export const ClayPaginationBarWithBasicItems: React.FunctionComponent<IProps> =
 	totalItems,
 	...otherProps
 }: IProps) => {
+	let showDropDown = true;
+
+	if (!deltas) {
+		deltas = defaultDeltas;
```
diegonvs, line 133: I think `defaultDeltas` are unnecessary in this case.

thinking about the usage:

* When passing `deltas`: 

`defaultDeltas` will be ignored and the given `deltas` will be used ✅ 

* When not passing `deltas`:

`defaultDeltas` value will be used in this case, to not introduce a breaking change ❌ 

* When using a "falsy" value to `deltas` :

`defaultDeltas` still be used. ❌ 

Would you mind creating another flag like `disableItemsPerPage` or `disableDeltas` and just use it instead of `showDropDown`? What do you think?

matuzalemsteles:
Hey @julien I'm still trying to figure out what the use cases for DropDown would be disabled another point is that making `deltas` optional would be recommended for taglib but in the React component and this reflects in taglib also that we have the default behavior when the value is not passed, changing this behavior would be a breaking change now, so to maintain compatibility we would have to create an API that allows you to disable delta regardless of the value you are passing and we can remove it in a next major version to make this behavior default, if the value of `deltas` is not passed DropDown is not shown.

If we go with this solution here, we would be breaking this behavior. What do you think? We can make the `ClayPaginationBar` taglib already have this behavior by default and use the React API for that. For example:

- `paginationBarDeltas` if is `null`, sets the `disableDeltas` API on the React side to `true`

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
julien: Hey @diegonvs 

> I think `defaultDeltas` is unnecessary in this case.

I don't understand what you mean, because it's already used (i.e. before my changes) if the `deltas` prop is [not passed](https://github.com/liferay/clay/blob/d8251c47262ccff9e715baf050c2713b075b11fa/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx#L119) and if the `activeDelta` prop isn't passed it gets it's value from there.

> When using a "falsy" value to deltas

I don't understand what you mean by a "falsy" value since `deltas`, see [here](https://codesandbox.io/s/determined-mendeleev-mowsd) 

> Would you mind creating another flag like disableItemsPerPage or disableDeltas and just use it instead of showDropDown? What do you think?

I don't mind adding another prop, but it would be nice to make it explicit: `disableDeltas` sounds like something that would be used to "ignore the deltas that are passed", but it's not the case. The only thing we want to do is "hide" the DropDown, so I'd make it as explicit and simple as possible. 

Actually, several people asked me what this "delta" prop means because they don't understand the meaning and it's already creating confusion.

julien:
Hey @matuzalemsteles thanks for the feedback

>  I'm still trying to figure out what the use cases for DropDown would be disabled

You don't have to try :smile: In this case the team using the component simply doesn't want the dropdown to be rendered because it's only going to show a single value that can't be changed: and they find this unnecessary (for me it makes sense)

> another point is that making deltas optional would be recommended for taglib but in the React component and this reflects in taglib also that we have the default behavior when the value is not passed, changing this behavior would be a breaking change now

I'm not sure I fully understand, but `deltas` is already [optional](https://github.com/liferay/clay/blob/d8251c47262ccff9e715baf050c2713b075b11fa/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx#L57) (correct me if I'm wrong) - the mistake that I made is that in the tag I marked it as [`required`](https://github.com/liferay/liferay-portal/blob/b795c02a239c901105c23612628a13f0097cab4e/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/resources/META-INF/liferay-clay.tld#L2194)

You can already use `ClayPaginationBarWithBasicItems` without passing a `deltas` prop, like this for example
```jsx
<ClayPaginationBarWithBasicItems
        totalItems="100"
        />
```

So, I don't understand why you mention a breaking change.

A *much more simple* change would be adding a new prop called `hideDropDown` or `showDropDown` (a boolean) or something else (can't come up with a better name right now but that's what this thread is for) that would just simply avoid rendering the dropdown if it's passed leaving the rest unchanged.

matuzalemsteles:
> I'm not sure I fully understand, but deltas is already optional (correct me if I'm wrong) - the mistake that I made is that in the tag I marked it as required

Sorry, maybe what I said was confused. Yes `deltas` here is optional, the only change meant I'm commenting on is about changing the behavior of the default value of `deltas`, for example:

- The default value of `deltas` will be `defaultDeltas` and its default behavior is to render DropDown of `deltas`
- In this case we still keep a default value for `deltas` but it won't render `DropDown`

It's exactly the change that behavior I'm talking about that causes a breaking change and the suggestion in this case to avoid it would be as you said, go with the creation of API to disable `DropDown` in order not to change the behavior of `deltas`.

For the case of Taglib, we don't need to add this API, because we just created it, so the risk is much lower, in this case we just use this API on the React side to make this behavior possible in Taglib.

For the API name, I suck too but having `disable` sounds a bit strange in APIs, particularly speaking, I think we can decide on two options `showDeltasDropDown` with the default value `true` or `disableDeltasDropDown`, I think I prefer `showDeltasDropDown` just because I see the same in some of our components as `ColorPicker`...

julien:
Hey @matuzalemsteles thanks for the suggestion, I'll go with `showDeltasDropDown`

julien:
Updated in https://github.com/liferay/clay/pull/4178/commits/6c65f43da7e64c3ed271539f776cb6ab7d0df633

matuzalemsteles (approved):
LGTM, Thanks @julien

PR #4179, opened 2021-07-22
fix(@clayui/color-picker): fix error of "hidden" input add extra margin

### LINKED ISSUE #4173: @clayui/color-picker label has 4px space
The `label` element text is slightly misaligned in Clay Color Picker due to a hidden input that is 4px by 4px. We should add `position: absolute;` to the hidden input so the input doesn't take that space.

color-picker-space

Credit to https://github.com/lfrsales/73-formed-theme/blob/master/src/css/formed/widgets/forms.scss#L104-L106. I was trying to figure out the reason for that style.

matuzalemsteles:
Fixes #4173

Hey @pat270 I realized that we can't add a `type=hidden` because we use this `input` as support for the native state, we need it for when I click on `Splotch` to open the native input, so I had to go with the first solution of adding the `position: absolute;`. This is related to this PR #3596

This PR also fixes an error on `FocusScope`, when you press `Tab + Shift` you get stuck on `input` which is not visible, so I added the behavior to ignore focus when `tabIndex` of `input` is smaller than the minimum and if not, keep the default behavior of ignoring only with type hidden and file.

### DIFF
--- a/packages/clay-color-picker/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-color-picker/src/__tests__/__snapshots__/index.tsx.snap
@@ -5,12 +5,6 @@ exports[`Interactions opens custom color picker drop down when clicked 1`] = `
   <div
     class="clay-color-picker"
   >
-    <input
-      name="colorPicker2"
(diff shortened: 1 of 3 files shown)


pat270:
LGTM!

Issue #4187, opened 2021-07-28
Tooltip is freezing on the screen


markocikos:
See description and steps to reproduce in https://issues.liferay.com/browse/LPS-135411

| Tech  | Version |
| ----- | ------- |
| @clayui/tooltip  | 3.29.0  |

PR #4188, opened 2021-07-28
fix(@clayui/tooltip): Hide tooltip on `dragstart`

markocikos:
Fixes: https://github.com/liferay/clay/issues/4187

The root cause of the issue is that none of the existing hide events trigger when node is being dragged. Funnily enough, this includes `mouseout`; it makes sense as the node is all the time below the mouse. Once the drop happens, node is removed from DOM, leaving a dangling tooltip in portal.

To fix this, in this PR we are adding `dragstart` to list of events that trigger tooltip hide.

Before PR:
before

After PR:
after

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -58,6 +58,7 @@ const initialState: IState = {
 };
 
 const TRIGGER_HIDE_EVENTS = [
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM! Oh perfect @markocikos, this makes perfect sense, thanks for checking this out and sent the fix.

PR #4189, opened 2021-07-28
fix(@clayui/autocomplete): fixes a potential XSS vulnerability in autocomplete by removing the use of `dangerouslySetInnerHTML`

### LINKED ISSUE #4169: Potential XSS issue with Autocomplete when rendering items
Well the behavior is hard to reproduce by itself with just the use of the MultiSelect or Autocomplete component, but there is a problem in DXP that a Categorization saves the value of `>` and it is passed for `Autocomplete` items causing it to display an alert. Check ticket for more information. https://issues.liferay.com/browse/LPS-135537

Should we escape or just try to get rid of the `dangerouslySetInnerHTML` usage in the `<Autocomplete.Item />` component but we preserve the fuzzy.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.30.0  |
| React | v16.12.0 |

matuzalemsteles:
Fixes #4169

Well to avoid XSS issues in Autocomplete because values can come from user input, I'm removing the use of `dangerouslySetInnerHTML` and "parsing" it to better handle the fuzzy output.

### DIFF
--- a/packages/clay-autocomplete/src/Item.tsx
+++ b/packages/clay-autocomplete/src/Item.tsx
@@ -22,7 +22,7 @@ export interface IProps extends React.ComponentProps<typeof ClayDropDown.Item> {
 	value: string;
 }
 
-const optionsFuzzy = {post: '</strong>', pre: '<strong>'};
+const optionsFuzzy = {post: '|+', pre: '+|'};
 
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
I'm going to merge this to include in today's release.

PR #4196, opened 2021-07-29
feat(@clayui/css): Reboot `body` element should use clay-css mixin for generating properties

### LINKED ISSUE #4194: @clayui/css body element should use clay-css mixin for generating properties
https://github.com/liferay/clay/blob/b3d1b171cc411bc8325f15a96f84e0458b9e94f0/packages/clay-css/src/scss/components/_reboot.scss#L33-L34

Currently, we are unable to change certain CSS properties in the `body` element through variables without it propagating through all components that use `$font-size-base`. We should be able to modify the `font-size` only for the `body` element.

pat270:
fixes #4194


matuzalemsteles (approved):
LGTM

PR #4201, opened 2021-08-03
fix(@clayui/css): Removes the use of `$theme-colors` Sass map to generate utility properties and replaces with new maps

### LINKED ISSUE #4180: @clayui/css $theme-colors Sass map should accept CSS Custom Properties
The map:
```scss
$theme-colors: (
  primary: var(--primary, #{$primary}),
  secondary: var(--secondary, #{$secondary}),
  success: var(--success, #{$success}),
  info: var(--info, #{$info}),
  warning: var(--warning, #{$warning}),
  danger: var(--danger, #{$danger}),
  light: var(--light, #{$light}),
  dark: var(--dark, #{$dark}),
);
```

causes Bootstrap's `theme-color-level` function to choke.

pat270:
fixes #4180

I couldn't get the `$theme-colors` map to play nicely with CSS variables, ended up creating separate maps for each utility that depends on it. This is much better since it allows better customization of colors and support for CSS variables.

```
$bg-theme-colors
$bg-gradient-theme-colors
$border-theme-colors
$text-theme-colors
$list-group-item-theme-colors
$table-row-theme-colors
```

This pr also adds some nice functions for converting a color string to type color from https://github.com/KittyGiraudel/SassyJSON/tree/master/stylesheets/decode. This will allow fallback values from CSS variables to work with Sass color functions if needed.

### DIFF
--- a/packages/clay-css/src/scss/components/_list-group.scss
+++ b/packages/clay-css/src/scss/components/_list-group.scss
@@ -409,21 +409,26 @@
 
 // List Group Item Variants
 
-@each $color, $value in $theme-colors {
+@each $color, $value in $list-group-item-theme-colors {
 	.list-group-item-#{$color} {
-		background-color: theme-color-level($color, -9);
-		color: theme-color-level($color, 6);
+		@include clay-css(setter($value, ()));
 
 		&.list-group-item-action {
-			@include hover-focus() {
-				background-color: darken(theme-color-level($color, -9), 5%);
-				color: theme-color-level($color, 6);
+			$active: setter(map-get($value, active), ());
+			$hover: setter(map-get($value, hover), ());
+
+			$focus: setter(map-get($value, focus), $hover);
+
+			&:hover {
+				@include clay-css($hover);
+			}
+
+			&:focus {
+				@include clay-css($focus);
 			}
 
 			&.active {
-				background-color: theme-color-level($color, 6);
-				border-color: theme-color-level($color, 6);
-				color: $white;
+				@include clay-css($active);
 			}
 		}
 	}
--- a/packages/clay-css/src/scss/components/_tables.scss
+++ b/packages/clay-css/src/scss/components/_tables.scss
@@ -226,12 +226,41 @@ caption {
 
 // Table Row Backgrounds
 
-@each $color, $value in $theme-colors {
-	@include table-row-variant(
-		$color,
-		theme-color-level($color, $table-bg-level),
-		theme-color-level($color, $table-border-level)
-	);
+@each $color, $value in $table-row-theme-colors {
+	.table-#{$color} {
+		&,
+		> th,
+		> td {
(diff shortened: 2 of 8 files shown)


packages/clay-css/src/scss/components/_list-group.scss
```diff
@@ -409,21 +409,26 @@
 
 // List Group Item Variants
 
-@each $color, $value in $theme-colors {
+@each $color, $value in $list-group-item-theme-colors {
```
matuzalemsteles, line 412: @pat270 Couldn't this be a breaking change or was `$theme-colors` only used internally?

matuzalemsteles (commented):
I'm a little confused by this but I understand why creating new variables, I have a concern about this, this would be more of a way to change colors using Sass variables right?

For example I can change colors by selectively changing the colors in `$primary`..., `$theme-colors` and now the specific colors of some parts with `$bg-theme-colors`, `$list-group- item-theme-colors`... by default they all follow `$primary`, `$secondary`... colors right? to me it looks ok, add a list of colors by categories but why add a group of colors for the specific component `$list-group-item-theme-colors`? would have a way around it? it's a little weird because it only applies to 2 components in this case and not to the others.

packages/clay-css/src/scss/components/_list-group.scss
pat270: @matuzalemsteles Yes this is a breaking change since changing `$theme-colors` no longer affects this component. The way `$theme-colors` generates variant colors was so bad I doubt anyone was reliably using this.

pat270:
Bootstrap used `$theme-colors` to generate `list-group-item-*` variants. I wouldn't have added `$list-group-item-theme-colors` if they weren't being generated by `$theme-colors`.

packages/clay-css/src/scss/components/_list-group.scss
matuzalemsteles: Hmm I see, it makes sense not to link, well I'll trust you 🙂, anyway in case anyone missed it, we can recommend replacing it because `theme-colors` has problems accepting CSS variables as arguments.

matuzalemsteles:
> Bootstrap used $theme-colors to generate list-group-item-* variants. I wouldn't have added $list-group-item-theme-colors if they weren't being generated by $theme-colors.

No problem with that anyway, just one thing for us to think about whether it makes sense to keep it or it would be better to refactor. But we can go ahead with this anyway.

Once we make this change do we also need to change the DXP?

pat270:
I'd like to get rid of it among other things. I'll compile a list and file an issue, we can discuss it there?

matuzalemsteles:
@pat270 perfect, it would be great to have a list or somewhere with these things so we can have a more macro view of the work.

matuzalemsteles:
@pat270 I'll merge this.

PR #4204, opened 2021-08-06
fix(@clayui/css): Cadmin Modal .modal.show should have display: block

### LINKED ISSUE #4203: @clayui/css .modal.show should have `display: block` and some Cadmin modal selectors not applying
`.modal.show` doesn't have `display: block` because Clay and Bootstrap modal plugins add it for us. There is a case in DXP with the old aui modal where it doesn't add `display: block` inline.

Cadmin `.modal-full-screen`, `.modal-full-screen-sm-down`, `.modal-height-*`, and `.modal-{color}` variants should work if the `cadmin` class is on the `.modal` element.

pat270:
fix(@clayui/css): Cadmin Modal .modal-full-screen, .modal-full-screen-sm-down, .modal-height-*, and .modal-{color} variants should work if the cadmin is class is applied on the same element

fixes #4203

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_modals.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_modals.scss
@@ -44,6 +44,8 @@
 
 .modal.show,
 &.modal.show {
+	display: block;
+
 	.modal-dialog {
 		transform: $cadmin-modal-show-transform;
 	}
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM

PR #4211, opened 2021-08-10
fix(@clayui/css): Forms .form-control-select with long text shouldn't break to new line and should have overflow ellipsis, similar behavior to select.form-control

### LINKED ISSUE #4206: @clayui/css .form-control-select.form-control-sm should have bigger padding-right value to offset the icon
form-control-select

The text flows into the double-caret-l icon due to `padding-right` being too small. We also should account for `form-control-select` being used with an `a` tag, hover/focus styles be removed/reset.

pat270:
.form-control-select.form-control-sm should have space on the right for the double-caret-l icon

form-control-sel

fixes #4206

### DIFF
--- a/packages/clay-css/src/content/form_elements.html
+++ b/packages/clay-css/src/content/form_elements.html
@@ -192,11 +192,33 @@ section: Components
 			<option>Sample 2</option>
 			<option>Sample 3</option>
 			<option>Sample 4</option>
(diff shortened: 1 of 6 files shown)


matuzalemsteles (approved):
great fix 🙂, thanks. LGTM!

Issue #4205, opened 2021-08-06
@clayui/css Remove unused components with fallback to add back in


pat270:
Related to https://github.com/liferay/clay/issues/1104.

We have a lot of components in Clay CSS that are specifically created for DXP's admin ui. These may be of use to site creators, but most likely not. We should try and remove them to reduce the CSS size. We also have a lot of Bootstrap components that aren't used. I've compiled a list of components that could easily be removed without regressions in DXP.

Clay CSS Components:
- application-bar
- navigation-bar breakpoints sm, lg, xl
- management-bar breakpoints sm, lg, xl
- timeline

Bootstrap Components:
- card-deck, card-group, card-columns, accordion
- carousel
- custom-switch, custom-select, custom-select-sm, custom-select-lg, custom-file, custom-range
- html5 form validator
- col-form-label, form-row, form-inline
- clean up input group styles
- jumbotron
- spinner
- toasts

We could get even more specific by providing different sets of components to Classic and Admin Theme. Classic Theme would have the minimum, whatever that is.

matuzalemsteles:
It makes sense to remove stuff from Bootstrap but I'm not sure about Clay CSS components, as we're adopting a `cadmin` specific component strategy I'm a little worried about how we're going to keep this documented, or should we separate the documentation at the session level, it might look like a little confusing for some, although these components are specific to `cadmin`.

nhpatt:
Bootstrap looks safe. Clay CSS I'm also worried about feature parity and documentation. We can analyze how much we can win (or if it's where we can get most bang for our buck) for removing those and decide.

pat270:
I was thinking we could add a disclaimer that it's not provided by default and document a way to add it if they need it. Application Bar == Control Menu in DXP, but doesn't use the markup and CSS. We could comment out the [import](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/_components.scss#L62) and write up how to bring it back (uncomment it).

For our docs, we would supply Clay CSS everything so our docs won't look broken. I know this is confusing, but I don't see how we can provide an upgrade path for those who might have developed something based on a component we provided.

matuzalemsteles:
@pat270 it might make sense for some components, for example `cadmin` components we should export somewhere and document them, management bar component is used a lot in both taglib and React components, so ignore it from clayui.com documentation might be a little weird, although we think some of these components are DXP specific I still think that because they are in the Lexicon they are part of the Design System experience and could be used anywhere.

[...]

PR #4217, opened 2021-08-12
fix(@clayui/css): Deprecate unused Bootstrap 4 components

pat270:
#4205

I still need to add documentation on deprecated components. I'm thinking we should include it in Getting Started under deprecated (idk) since none of these components are documented on clayui.com? 

This pr adds a setting `$enable-bs4-deprecated: false;` and deprecates the Bootstrap 4 components:
- Cards
    - `.card-header-tabs`, `.card-header-pills`, `.card-img-*`, `.card-deck`, `.card-group`, `.card-columns`, `.accordion`
- Carousel
- Custom Forms
    - `.custom-switch`, `.custom-select`, `.custom-select-lg`, `.custom-select-sm`, `.custom-file`, `.custom-range`
- HTML5 Form Validation
    - `.was-validated`, `.invalid-feedback`, `.valid-feedback`, `.#{$state}-feedback`, `.#{$state}-tooltip`
- Form
    - `.col-form-label`, `.col-form-label-lg`, `.col-form-label-sm`, `.form-row`, `.form-inline`
- Jumbotron
- Nav Pills
- Spinners
- Toasts
- Clay CSS Component Table Drag. This was implemented in DXP.

```
Before:
atlas.css: 610kb
atlas.css.gz: 72kb
base.css: 563kb
base.css.gz: 67kb

After:
atlas.css: 583kb (27kb reduction)
atlas.css.gz: 68kb (4kb reduction)
base.css: 536kb (27kb reduction)
base.css.gz: 63kb (4kb reduction)
```


matuzalemsteles (commented):
@pat270 that's interesting, I took a look at the archives to see if we had some related documents on clayui.com but we really don't have what could be good for this case that we can know that maybe people don't use, maybe clients should use due to bootstrap.

In terms of adding this information in some doc, the best case would be to create a blog post about it on clayui.com to talk about it, how to activate it, why we're doing it, and what the goal is. We can create a new Settings documentation to talk about enabling these features via a variable and saying what this will enable can be a way too.

I'm still wondering if the best way here would be to leave this still in v3 and wait for v4 to remove it, it's kind of counterintuitive to remove this now if the goal is just to shrink our bundle, we know it will shrink but that wouldn't be critical for now I'm not sure but if there is an option to keep the flag enabled by default and remove in v4 I think that would be the ideal scenario. It's very difficult to know who isn't using it or if there really is someone using it, especially saying that Clay CSS is also used outside the Liferay.

pat270:
@matuzalemsteles we can set `$enable-bs4-deprecated: true;`. It will output all those components. You're right, it's probably better to leave it in and let people opt out of it. Removing it in V4 sounds good, but I'm guessing we will have to provide some kind of fallback if others are using it. I think we can disable in V4 and remove in V5 😆 .

Blog post sounds good I will take a look at writing something up.

matuzalemsteles:
> Removing it in V4 sounds good, but I'm guessing we will have to provide some kind of fallback if others are using it. I think we can disable in V4 and remove in V5 😆 .

😅 Well, we can do that, hopefully, it doesn't take too long to remove 🙂. In this version we can say that we are deprecating these bootstrap components and that people should start using others instead. The blog post would be a great place to talk about this and the depreciation plan.

nhpatt:
Just to reiterate ideally we should do these changes in a 3 step process, first opt-in (I have a flag that enables me to remove those bootstrap components), then opt-out (a flag to add/enable them), the remove all together.

Seems hard for these kind of changes but we should not remove them directly at least without deprecating them first in one version (this one for example).

matuzalemsteles:
@nhpatt Yeah, I agree. I think it's safe to remove this in v5 completely and in v4 add the flag to enable. It would be good to prepare a blog post on clayui.com to inform you about this plan and if it involves other components, if you need help I can help you.

nhpatt:
I'm not so sure about the versions as they are usually very costly for external developers (it's something that we can discuss when planning the release). But I would deprecate with a flag to remove now if possible.

pat270:
Thanks for the feedback guys, I changed the flag to be opt-in for disabling. This outputs all the CSS unless you set `$enable-bs4-deprecated: false !default;`

matuzalemsteles (approved):
LGTM

PR #4220, opened 2021-08-13
fix(@clayui/css): Wrap *-theme-colors map keys in single quotes. Key …

### LINKED ISSUE #4219: @clayui/css Wrap keys in quotes for theme-colors Sass maps
Sass generates an annoying warning:

> WARNING: You probably don't mean to use the color value white in interpolation here.
It may end up represented as white, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "white").
If you really want to use the color value here, use '"" + $color'.

We should wrap the [color keys](https://github.com/liferay/clay/blob/1d1e05d6b8c59a898dd30c2b0e58d51902cb162c/packages/clay-css/src/scss/variables/_utilities.scss#L132-L188) in quotes

pat270:
…names that match CSS colors makes Sass throw this warning without quotes:

> WARNING: You probably don't mean to use the color value white in interpolation here.
It may end up represented as white, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "white").
If you really want to use the color value here, use '"" + $color'.

fixes #4219

### DIFF
--- a/packages/clay-css/src/scss/variables/_globals.scss
+++ b/packages/clay-css/src/scss/variables/_globals.scss
@@ -124,14 +124,14 @@ $dark-l2: lighten($dark, 32.94) !default;
 $theme-colors: () !default;
 $theme-colors: map-merge(
 	(
-		primary: $primary,
-		secondary: $secondary,
-		success: $success,
-		info: $info,
-		warning: $warning,
-		danger: $danger,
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
LGTM

PR #4225, opened 2021-08-17
fix(@clayui/css): Cadmin Modals allow adding cadmin directly to modal…

### LINKED ISSUE #4224: @clayui/css Cadmin Modals allow placing the cadmin class directly on the modal-header, modal-body, and modal-footer elements
https://issues.liferay.com/browse/LPS-137335

The issue above is a unique situation where we need to isolate everything except the content inside modal-body. We can isolate header, footer and all content inside it by placing the `cadmin` class directly on it. The parent element styles (modal-backdrop, modal, modal-content, etc) will be isolated in the same way isolating inline control styles is done.

pat270:
…-header, modal-body, and modal-footer elements

fixes #4224

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_modals.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_modals.scss
@@ -158,7 +158,8 @@
 	width: 100%;
 }
 
-.modal-header {
+.modal-header,
+&.modal-header {
 	align-items: center;
 	background-color: $cadmin-modal-header-bg;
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
@pat270 does that mean we can add `<div className="cadmin modal-footer" />` together?

pat270:
@matuzalemsteles yes we can add it directly like that.

matuzalemsteles (approved):
LGTM

Issue #4240, opened 2021-08-27
Facilitate the experience of testing Clay locally on DXP


matuzalemsteles:
Well currently we have a lot of problems testing local Clay in DXP, we already had two problems with that, one of them was giving React type conflicts from Clay monorepo to DXP monorepo types.

```
../../../../../../clay/node_modules/@types/react/index.d.ts(2817,14): error TS2300: Duplicate identifier 'LibraryManagedAttributes'.
../../../node_modules/@types/react/index.d.ts(2835,14): error TS2300: Duplicate identifier 'LibraryManagedAttributes'.
../../../node_modules/@types/react/index.d.ts(2848,13): error TS2717: Subsequent property declarations must have the same type.  Property 'a' must be of type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>', but here has type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>'.
```

The idea is to try to simplify creating some bash script or nodejs as appropriate to simplify the linking of Clay modules in DXP.

julien:
@matuzalemsteles just tested and here are the steps, I did to test "local" changes:

- Run the shell script posted in the last comment
- Navigate to `liferay-portal` repository
- Navigate to `modules/apps/frontend-taglib/frontend-taglib-clay`
- Link packages (in my case `yarn link @clayui/alert @clayui/button @clayui/label`)
- Stop tomcat (When deploying the `frontend-taglib-clay` module, if tomcat is running it usually crashes)
- Deploy the OSGi module (`gradlew deploy -a`)
- Start tomcat 
- Navigate to http://localhost:8080

(My change was totally pointless but it was just to test, so don't be scared of what you see in this screenshot)

Result:
![](https://user-images.githubusercontent.com/5572/131660939-9d3b6dc2-8f64-46f2-bb8f-a54c18d56d1d.png)

julien:
Adding another comment, just to say that personally if it's just a matter of linking the desired packages I would rather not maintain a script and have detailed instructions in our README or Wiki. 
It just boils down to something like this 

Run and build the package from the clay repo

```
yarn workspace @clayui/PACKAGE_NAME build
yarn workspace @clayui/PACKAGE_NAME link
```

To use it from an OSGi module in DXP

```
yarn link @clayui/PACKAGE_NAME
gradlew deploy -a
```

But I'm curious to know what the rest of the team thinks.

PR #4241, opened 2021-08-27
feat: Adds RTL OOTB support to methods using dom-align

### LINKED ISSUE #4238: Components that use dom-align do not follow RTL for adjust position
Components are used inside DXP and support languages with RTL directions, Clay components that use `dom-align` to position elements that are rendered on the body, do not take into account the RTL position in the positioning.

We can add OOTB support, once the page is in the RTL position we do the inverse direction (as mentioned by Bryce) in the desired placement in the `dom-align`.

matuzalemsteles:
Fixes #4238

From #4239, I added some changes to this PR of the @carloslancha, moved the method to our `@clayui/shared` package to be able to use it in other components that also use `dom-align`, I also went with @jbalsas' advice to get the element direction, since there may be the use case that we have different directions on the same page in some scenarios, seems safer than checking the document.

### DIFF
--- a/custom-types/dom-align/index.d.ts
+++ b/custom-types/dom-align/index.d.ts
@@ -4,27 +4,25 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-interface IConfig {
-	points: readonly [string, string];
-	offset: readonly [number, number];
-	targetOffset: readonly [string, string];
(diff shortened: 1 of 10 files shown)


matuzalemsteles:
Oh thanks @julien, I will take this into consideration for #4240

PR #4243, opened 2021-08-30
fix(@clayui/shared): fixes error when importing the type that is not available in the module

matuzalemsteles:
Well, the types for `dom-align` are defined inside Clay but we don't export them because they are `declare module` in our types root, it wouldn't make much sense either, in this case, I'm just adding the types that are needed as part of the `shared` package to allow error-free distribution.

### DIFF
--- a/packages/clay-shared/src/doAlign.ts
+++ b/packages/clay-shared/src/doAlign.ts
@@ -3,11 +3,15 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
The error will only happen when depending on the component or using the types. That's why it doesn't fail our CI. https://github.com/liferay-frontend/liferay-portal/pull/1403#issuecomment-908485781

Anyway, I'm going to go ahead with this PR since it's not trivial to just copy the `custom-types` types for this case and allow me to go ahead with the patch release.

PR #4218, opened 2021-08-13
feat(@clayui/core): Add initial sketch of components to TreeView

matuzalemsteles:
Fixes #4117

Well, this is an initial draft for implementation of components for TreeView, the idea is that it looks like a low-level and high-level component, the idea of TreeView is that it is very flexible to build other components like Vertical Navigation or just customize the Node to behave in different variations. An important part of this design is that the TreeView is data agnostic and allows to build a static or dynamic TreeView, we still don't have all the resources because we will be building with other issues.

This component is already being built in `@clayui/core` but it is not being exposed so we are not officially releasing it even if a new version is cut.

You can see the stories that have been added with some basic examples.

## Design Principles

This component follows some different behaviors when compared to other components but this is already a reflection of the upcoming evolution and refinement of our other components, this can also help with insights for #4112.

Well, this mostly follows our low-level delivery proposal and I'm using a different strategy to balance high-level with low-level and still allow customizations, differently we don't have a component with many APIs or an `items` that define the properties, unlike the data here is agnostic and the developer is responsible for assembling the `Nodes` or `Item` here according to its data structure and adding new behaviors if is necessary.

```jsx
<TreeView
  items={[
    {name: 'Liferay Drive'},
    {name: 'Repositories'},
    {name: 'Documents and Media'},
  ]}
>
  {(item) => (
    <TreeView.Item>
      <Icon symbol="folder" />
      {item.name}
    </TreeView.Item>
  )}
</TreeView>
```

TreeView uses the [Render Props](https://reactjs.org/docs/render-props.html) technique to handle different behaviors, static or dynamic values, so we've also been able to abstract some of the complexity that markup adds, as many wrappers to conform to the design spec, we've abstracted here for reduce the amount of components and make the component more "smart".

### Static content

Static content is what we normally use for low-level components with just small block composition.

```jsx
<TreeView>
  <TreeView.Item>Liferay</TreeView.Item>
</TreeView>
```

The static component has some different behaviors to allow you to abstract the markup and reduce the writing of more code, for the cases when it is necessary to build a sublevel for the Node it is necessary to add the component `<TreeView.ItemStack />` explicitly to create the Node If the Node has more information besides Text, another component can also be used.

```jsx
<TreeView>
  <TreeView.Item>
    <TreeView.ItemStack>
      <Icon symbol="folder" />
      Liferay
    </TreeView.ItemStack>
    <TreeView.Group>
      {...}
    </TreeView.Group>
  </TreeView.Item>
</TreeView>
```

[...]

PR #4253, opened 2021-09-06
feat(@clayui/core): add initial dnd implementation for TreeView

julien:
(W.I.P. Do not merge yet, thanks)

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -40,6 +40,7 @@ export function TreeView<T>({
 	nestedKey,
 	onExpandedChange,
 	onSelectionChange,
+	rootItem,
 	selectedKeys,
 	showExpanderOnHover = true,
 	...otherProps
@@ -57,6 +58,7 @@ export function TreeView<T>({
 				? (children as ChildrenFunction<Object>)
 				: undefined,
 		expanderIcons,
+		items,
 		nestedKey,
 		showExpanderOnHover,
 		...state,
@@ -72,7 +74,9 @@ export function TreeView<T>({
 			role="tree"
 		>
 			<TreeViewContext.Provider value={context}>
-				<Collection<T> items={items}>{children}</Collection>
+				<Collection<T> items={items} rootItem={rootItem}>
+					{children}
+				</Collection>
 			</TreeViewContext.Provider>
 		</ul>
 	);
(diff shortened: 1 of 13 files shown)


packages/clay-core/src/tree-view/TreeView.tsx
```diff
@@ -0,0 +1,122 @@
...
+import {DndProvider} from 'react-dnd';
+import HTML5Backend from 'react-dnd-html5-backend';
+
+import {ChildrenFunction, Collection, ICollectionProps} from './Collection';
+import {TreeViewGroup} from './TreeViewGroup';
+import {TreeViewItem, TreeViewItemStack} from './TreeViewItem';
+import {Icons, TreeViewContext} from './context';
+import {ITreeProps, useTree} from './useTree';
+
+function generateItemIndices<T>(
```
julien, line 17: @matuzalemsteles I guess that this function could be inside (or part of) a custom hook.

Initially what I had done is something like this (in pseudo code)

```tsx
export function TreeView<T>({
        ...
	items: initialItems,
	...
}: ITreeViewProps<T>) {
   const [items, setItems] = React.useState(generateItemIndices(initialItems));
   ...
}
```

turns out that broke everything (nothing was rendered) and I couldn't figure out why so I went with this "ugly solution" just to have something working. A hook would probably be the good place to also "move" the items which is currently done in `TreeViewItem`. Apart from that, there's not much but I also noticed that when dropping items in some cases there's no update. If you have any ideas let me know.

packages/clay-core/src/tree-view/TreeView.tsx
```diff
@@ -0,0 +1,122 @@
...
+	items: Array<T> | Record<string, T> | undefined
+) {
+	if (!items) {
```
julien, line 20: @matuzalemsteles  About this, I know why we defined items as `Array<T> | Record<string, T> | undefined` (we talked about in either #4218 or #4222), but there's got to be a better way to deal with these union types, I need to clean up.

packages/clay-core/src/tree-view/TreeView.tsx
matuzalemsteles: @julien yeah I'm taking a look at it now. Probably as you are manipulating the nested structure, this does not cause a rendering, the ideal would be to create a copy of the structure and then manipulate and save in state, I am thinking of some solutions now.

matuzalemsteles:
Hey @julien I did some exploration today, send a PR #4254 to review and verify some details. See my changes and comments in PR #4254, still very early.

julien:
Closing here since it seems we're going to use #4254

PR #4254, opened 2021-09-07
chore(@clayui/core): uses RFC 6902 (JSON Patch) implementation to move item in drop action

matuzalemsteles:
🚨 This is a very draft PR created on top of PR #4253 for exploration. Not considering a final resolution.

This is just a draft for an attempt to implement [RFC 6902 (JSON Patch)](https://datatracker.ietf.org/doc/html/rfc6902), it also won't cover all the cases that the RFC covers but uses the standards to implement the `move` operation which is what interests us most for perform the `drop` operation, also somewhat based on the [Immer](https://immerjs.github.io/immer/patches/) API which is used for immutability.

It still has to add the implementation of applying patches and dealing with immutability which is being very difficult to do.

In [section 4.4](https://datatracker.ietf.org/doc/html/rfc6902/#section-4.4) the `path` and `from` are a string but we are using an array of indices to simplify the case here.

I'm not considering adding the index to tree structure to `items` because that would be an operation we need to call again when implementing "load more" and it's a little expensive depending on the depth and size, just testing other alternatives.

A bigger difficulty here is dealing with the immutable tree because we can't manipulate the data as mutable, otherwise, we can cause side effects, like losing the data or not being able to update the state so that React does a new rendering, I've been thinking about testing Immer to deal with this, I haven't explored it yet. Also keeping in mind the idea of being as performative as possible to avoid very expensive operations, so following some patterns that avoid traversing the tree the same must be done to deal with immutable data.

> ⚠️ Still missing to implement the final result

### DIFF
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -33,7 +33,6 @@
 		"@clayui/modal": "^3.35.2",
 		"@clayui/provider": "^3.35.2",
 		"@clayui/shared": "^3.35.2",
-		"@storybook/react": "^6.3.8",
 		"classnames": "^2.2.6",
 		"react-dnd": "^10.0.2",
 		"react-dnd-html5-backend": "^10.0.2",
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -32,7 +32,9 @@
 		"@clayui/modal": "^3.35.2",
 		"@clayui/provider": "^3.35.2",
 		"@clayui/layout": "^3.32.0",
-		"classnames": "^2.2.6"
+		"@clayui/shared": "^3.32.0",
+		"classnames": "^2.2.6",
+		"react-transition-group": "^4.4.1"
 	},
 	"peerDependencies": {
 		"@clayui/css": "3.x",
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -30,7 +30,9 @@
 		"@clayui/drop-down": "^3.35.2",
 		"@clayui/icon": "^3.32.0",
 		"@clayui/modal": "^3.35.2",
-		"@clayui/provider": "^3.35.2"
+		"@clayui/provider": "^3.35.2",
+		"@clayui/layout": "^3.32.0",
+		"classnames": "^2.2.6"
 	},
 	"peerDependencies": {
 		"@clayui/css": "3.x",
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -29,11 +29,14 @@
 		"@clayui/button": "^3.32.0",
 		"@clayui/drop-down": "^3.35.2",
 		"@clayui/icon": "^3.32.0",
+		"@clayui/layout": "^3.32.0",
 		"@clayui/modal": "^3.35.2",
 		"@clayui/provider": "^3.35.2",
-		"@clayui/layout": "^3.32.0",
-		"@clayui/shared": "^3.32.0",
+		"@clayui/shared": "^3.35.2",
+		"@storybook/react": "^6.3.8",
 		"classnames": "^2.2.6",
+		"react-dnd": "^10.0.2",
+		"react-dnd-html5-backend": "^10.0.2",
 		"react-transition-group": "^4.4.1"
 	},
 	"peerDependencies": {
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -34,6 +34,7 @@ export function Collection<T extends Record<any, any>>({
 }: ICollectionProps<T>) {
 	const {key: parentKey} = useItem();
 	const {items: rootItems} = useTreeViewContext();
+	const dragAndDropEnabled = typeof children === 'function';
 
 	const items =
 		typeof compoundItems === 'object' &&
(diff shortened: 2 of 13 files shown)


julien:
Hey @matuzalemsteles  the document you linked was an interesting read,
thanks for sharing. That said I have some doubts about how this
fits in with drag and drop, because in the "abstract" they mention the 
HTTP PATCH method, which is what got me confused.

> Abstract
>   JSON Patch defines a JSON document structure for expressing a
   sequence of operations to apply to a JavaScript Object Notation
   (JSON) document; it is suitable for use with the HTTP PATCH method.
   The "application/json-patch+json" media type is used to identify such
   patch documents.

Seeing your changes, I know we want to use this concept to manipulate 
items, but if you could describe exactly how this will work it would nice
just to have everyone on the same track and avoid confusion.

I just need to understand some details, for example how the `path` is 
going to look like with lots of nested items, but this is much better 
than what it looked like in #4253

julien:
@matuzalemsteles one other thing I noticed is that all items don't 
get rendered. For example when changing the ["dynamic"](https://github.com/liferay/clay/blob/7bf180e3234f7190cc2662f85879f114ed43b5a6/packages/clay-core/stories/TreeView.stories.tsx#L130-L165) story, 
so that the "children" also contain "children", i.e. using this 

```
[
    {
        children: [
            {
                name: 'Blogs',
                children: [
                    { name: 'My Blog'  },
                    { name: 'Your Blog'},
                ]
            },
            {name: 'Documents and Media'},
        ],
        name: 'Liferay Drive',
    },
    {
        children: [
            {name: 'Blogs'},
            {name: 'Documents and Media'},
        ],
        name: 'Repositories',
    },
    {
        children: [{name: 'PDF'}, {name: 'Word'}],
        name: 'Documents and Media',
    },
]
```

The children of the `Blog` item don't get rendered, and seeing [this bit](https://github.com/liferay/clay/blob/7bf180e3234f7190cc2662f85879f114ed43b5a6/packages/clay-core/stories/TreeView.stories.tsx#L152-L164) I understand why.

I don't think that's very "reasonable", in the sense that the one should just
be able to pass any number of "nested items" and the TreeView should be able 
to render them.

julien:
@matuzalemsteles forget my last comment, played around a bit 
with that and using a custom component that renders itself does the trick

```jsx
function TreeViewItem({item}) {
	return (
		<TreeView.Item>
			<TreeView.ItemStack>
				<Icon symbol="folder" />
				{item.name}
			</TreeView.ItemStack>
			<TreeView.Group items={item.children}>
				{(item) => {
					if (item.children) {
						return <TreeViewItem item={item} />;
					} else {
						return <TreeView.Item>{item.name}</TreeView.Item>;
					}
				}}
			</TreeView.Group>
		</TreeView.Item>
	);
}
```

julien:
@matuzalemsteles I submitted #4255 based on this (in it's own branch)
we'll just need to improve the "move" functionality so that it returns
a copy of the tree

matuzalemsteles:
@julien many comments 😅 I will answer the ones that I think remained unanswered. But it looks like you've made a lot of strides here, so that's great.

> forget my last comment, played around a bit with that and using a custom component that renders itself does the trick.

Yes this is very configurable according to the expected structure of what is passed, so it is very much up to the developer to configure it according to its structure or as he knows it.


> Seeing your changes, I know we want to use this concept to manipulate items, but if you could describe exactly how this will work it would nice just to have everyone on the same track and avoid confusion.

Well, looking at PR #4255 now I think you understand the idea behind using this 😁, in the end, we don't want to implement everything the RFC says but adapt the idea of applying small changes to the structure, as JSON structures are similar and allow a tree, so we can adapt that to make changes to the structure in a standardized way. We just using the `move` operation which is the most valuable to us.

Although they mention the PATCH method is more a matter of communicating with the server, we don't need to worry about that. In the end this RFC is proposed to avoid problems such as sending the entire structure to the server with the changes because it can have a very large payload and the latency will also be high, so this pattern is used to send an operation with a very small payload asking for making changes to the structure, this is easy to apply in Frontend and backend to ensure a better user experience. In our case, since we don't use this, we just want the pattern of operations to apply the patch to the structure we have at runtime.

> we'll just need to improve the "move" functionality so that it returns a copy of the tree

Yeah, I'm going to try to explore this more, but doing it in a performance way is killing my neurons 😅

matuzalemsteles:
I will close this in favor of keeping PR #4255 as it will be better for us to collaborate.

julien:
@matuzalemsteles just some "final" thoughts

> Yes this is very configurable according to the expected structure of what is passed, so it is very much up to the developer to configure it according to its structure or as he knows it.

I agree, and as far as I'm concerned this is a good reason to use render props (so I'd be OK to merge #4218)
I'll add a comment there too.

> Well, looking at PR #4255 now I think you understand the idea behind using this grin, in the end, we don't want to implement everything the RFC says but adapt the idea of applying small changes to the structure, as JSON structures are similar and allow a tree, so we can adapt that to make changes to the structure in a standardized way. We just using the move operation which is the most valuable to us.

Yes I understood that, my point was that by mentioning "JSON Patch" some people might be confused, so even though this is still an "internal" conversation, once we release the component, we shouldn't mention this, or at least not link to the RFC in our documentation or in code comments.

matuzalemsteles:
> Yes I understood that, my point was that by mentioning "JSON Patch" some people might be confused, so even though this is still an "internal" conversation, once we release the component, we shouldn't mention this, or at least not link to the RFC in our documentation or in code comments.

Yeah, this is very internal we don't need to expose it. Only if we want to expose patches that are made as an API... but yeah, we don't need to mention that, just an implementation detail.

matuzalemsteles:
😂 haha but this is a valid point too.

Issue #4191, opened 2021-07-28
Expose Clay's packages in the `@clayui/core` package


matuzalemsteles:
We started working on the flow of merging all of Clay's packages into one for easy maintenance, simplifying their release and updating in DXP.

#4162 introduced an initial implementation for this, adding only a few packages but we still need to add all the other packages. The purpose of this ticket is to expose the rest of these packages and also make the component types available.

ethib137:
FYI: @brianchandotcom is very interested in the progress of this ticket in order to simplify the management of Clay packages in remote apps such as this one: https://github.com/liferay/liferay-portal/blob/master/modules/apps/site-initializer/site-initializer-raylife/extra/remote-app/package.json#L20

matuzalemsteles:
Thanks @ethib137 for the information! We might want to prioritize this after we finish migrating the TreeView. @dsanz What do you think?

julien:
@ethib137 thanks for the heads up. @matuzalemsteles let me know if we can talk about this via Slack - @sergiojimcos and I might be able to start working on this if it's our top priority.

dsanz:
We're considering moving clay to DXP in order to make it easier for clay to reach dxp. In this effort we must consider the need for simplifying the management of dependencies. We recenlty added the concept of target platform, which applies to external developers when dealing with JS projects for liferay.

I believe we shall consider applying these ideas to the remote apps too. Just noting this to start some brainstorming, ccing  @izaera as we already talked about this internally so he'd (hopefully) like to participate too ;)

izaera:
We are already [pulling all clay dependencies in target platforms](https://github.com/liferay/liferay-frontend-projects/blob/f59673a585bf1143cdf07130fe5628629d32344e/target-platforms/packages/portal-7.4/package.json#L8-L80), but that's for portlet projects (those deployed via JAR file).

In the case of remote apps, [we already have a project generator](https://github.com/liferay/liferay-frontend-projects/blob/f59673a585bf1143cdf07130fe5628629d32344e/projects/js-toolkit/docs/manuals/liferay-cli.md) that configures a build based on webpack that, when used with clay, would internalize clay deps into the generated javascript bundle.

[...]

PR #4256, opened 2021-09-07
chore(release): publish v3.35.3 packages

julien:
This is me just testing the steps described in 
the [release process](https://github.com/liferay/clay/blob/2697c1e53795b15e58240d863417d63811f5e5dd/CONTRIBUTING.md#release-process), so I'll make this a draft, but I still would 
like to see it pass. 

@matuzalemsteles, @bryceosterhaus if you have time 
and see anything fishy, let me know. Thanks


julien:
Well that failed quickly.
I've got a feeling that this error from the (CI/test) job:

```
  @clayui/provider:  BAD
    clay-core -> ^3.35.3
    clay-shared -> ^3.35.2
```

is something that needs to fixed manually.
Running `yarn checkDeps` in the `master` branch
also fails but with other errors.

```
Add missing dependencies with:

(cd packages/clay-core && yarn add @storybook/react classnames react-dnd react-dnd-html5-backend @clayui/shared react-transition-group @clayui/layout)
```

bryceosterhaus:
Okay I just got a chance to check this out.

```
 @clayui/provider:  BAD
    clay-core -> ^3.35.3
    clay-shared -> ^3.35.2
```

At first I was confused by this, but then I realized I was reading it wrong. This error is saying

"The `@clayui/provider` dependency is declared in clay-core as `^3.35.3` and in clay-shared as `^3.35.2`".

Meaning, these versions are out of sync.

If you bump the peerDependency in clay-shared so that it is dependent on `^3.35.3`, that problem should be solved.

As for running `yarn checkDeps` locally, you need to also make sure that you clean your built `lib` folders. I suspect that you have old built files since they are untracked by git. `checkDeps` is based on the built files and not the source files. It may be worthwhile to add a `clean` type of script to checkDeps so that we make sure the lib folders are wiped.

julien:
@bryceosterhaus thanks for the help and time.

Concerning the first error, you were right I did have to bump manually, 
which is why you'll see [this](https://github.com/liferay/clay/pull/4256/commits/446a10569e34c162ec169b227f3eac6ab07b416c) commit, (we can talk about it this is 
still a "test pr for myself" but this might be needed).

Everything else concerning the release process went fine.

Before running `yarn checkDeps` locally, I used 

```sh
find packages/ -mindepth 2 -maxdepth 2 -name lib -type d | xargs rm -rf {}
```

Which worked fine for me (and that we could use), but since I'm 
not on macOS I'll also ask @matuzalemsteles to try
that on his machine, because I'd just like to be sure that works 
for him too (and I've had issues with xargs on mac in the past).

Anyway, even if I'm going to delete this branch in a bit, 
I'd still like to keep it open for us to discuss.

matuzalemsteles:
@julien yeah, unfortunately as `@clayui/provider` is a peerDependency we have to manually update in `@clayui/shared`, I wanted to try to get rid of it but the circular dependency is very difficult to avoid, even if I put the files in `shared` we will have circular dependency due to the dependencies of the `provider` 😭. I hope I can remove this with the issue #4191 entry.

> Which worked fine for me (and that we could use), but since I'm not on macOS [...]

Yeah that worked fine on my machine I think it would be safe to add that. I'm not sure if we should add this to a script, maybe just as part of the release process.

julien:
> Which worked fine for me (and that we could use), but since I'm not on macOS [...]

> Yeah that worked fine on my machine I think it would be safe to add that. I'm not sure if we should add this to a script, maybe just as part of the release process.

Great, thanks for notifying. I think I can edit the release process document to include that - it's a very simple command so I agree that we don't need a script (or npm script) for that either

julien:
Closing this, since it was a "test run" pull request :) Thanks for the help guys.

PR #4258, opened 2021-09-07
fix(@clayui/css): Buttons create Sass placeholders for Clay CSS btn-*…

### LINKED ISSUE #4248: clayui/css: Buttons create Sass placeholders for `btn-primary` etc to allow theme devs to map their button classes to it
We should provide Sass placeholders for button classes. This would allow devs to map their custom button classes to Clay CSS' button classes. If a dev would like to have their main button be called `btn-type-1`, they could map it to Clay CSS' main button `btn-primary` with:
```scss
.btn-type-1 {
    @extend %btn-primary;
}
```
This would output:
```css
.btn-type-1,
.btn-primary {
    ...
}
```
Liferay Portal provides a _clay_custom.scss file where devs can apply this.

pat270:
… classes to allow theme devs to @extend their button classes from it

fixes #4248

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_buttons.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_buttons.scss
@@ -330,8 +330,12 @@ input[type='button'] {
 // Button Variants
 
 @each $cadmin-color, $cadmin-value in $cadmin-btn-palette {
-	.btn-#{$cadmin-color} {
+	%btn-#{$cadmin-color} {
 		@include clay-button-variant($cadmin-value);
+	}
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM! This is really cool, would this be interesting to be adopted as a pattern for other components?

pat270:
@matuzalemsteles Yes we probably could. I just haven't tested on more complex components like `input-group`.

matuzalemsteles:
Hm yeah, no problem @pat270! We can investigate this later.

PR #4268, opened 2021-09-14
chore(@clayui/form): add missing key prop to `OptGroup`

### LINKED ISSUE #4267: ClaySelectWithOption does not have proper keys

1) Enable development mode
2) Render a ClaySelectWithOption with at least a group and a separated option
3) Open browser console
4) A warning is shown because the OptGroup does not have a key
image
image

Users can set a key to ClaySelectWithOptions options. If it's not possible, at least the index could be added as key to each OptGroup, the same way is added to options.

## Environment
Google Chrome

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.35.3  |

julien:
Fixes #4267


matuzalemsteles (approved):
LGTM!

PR #4272, opened 2021-09-16
docs(clayui.css): Adds Scss API Documentation

### LINKED ISSUE #4185: Explore the possibility of using sassdoc to add on clayui.com
From #4177

The idea is to try to use some tool that can be used like Sassdoc to extract the Sass documentation from Clay, to expose the variables, mixins, functions that are available on clayui.com in each component's page, just like we do with the API for components in React.

pat270:
This uses sassdoc to generate Scss docs from comments in Clay CSS source. The html gets generated to `static/scss-api/` and I added a link in the main nav to point to it. It uses the default theme provided by sassdoc. We can create our own custom theme if we want.

fixes #4185


pat270:
`Error:   245:4   error  Expected object keys to be in natural ascending order. 'sideNavigation' should be before 'sidebar'  sort-keys` doesn't look correct to me.

matuzalemsteles:
Wow, I'll look deeper into this tomorrow.

pat270:
Might be looking at the value instead of key?
```
			sideNavigation: 'Side Navigation',
			sidebar: 'Sidebar',
```

julien:
@pat270 what I understand is that since N > b then `sideNavigation` gets sorted before `sidebar`.

pat270:
Thanks @julien! This completely makes sense when you put it that way.

clayui.com/gatsby-node.js
```diff
@@ -3,6 +3,12 @@
...
 
+const express = require('express');
+
 exports.createPages = require('./gatsby/createPages');
 exports.onCreateNode = require('./gatsby/onCreateNode');
 exports.onCreateWebpackConfig = require('./gatsby/onCreateWebpackConfig');
+
+exports.onCreateDevServer = ({app}) => {
+	app.use(express.static('static'));
+};
```
matuzalemsteles, line 14: Hmm this is weird that we have to do this to serve a static file, I thought this was OOTB.

clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
```diff
@@ -185,13 +186,88 @@ const generateCSSFiles = (pluginOptions) => {
...
+		package: {
+			description: pjson.description,
+			homepage: path.join('..', 'docs', 'get-started', 'index.html'),
+			license: pjson.license,
+			title: 'Clay CSS',
+			version: pjson.version,
+		},
+	};
+
+	sassdoc(path.join(pluginOptions.clayCssSrc, 'scss'), config);
```
matuzalemsteles, line 259: I had a different idea of this, actually I didn't want to use `sassdoc` to generate the static files but if it was possible to use the API to return the JSON data and so we could use it in different places in the documentation, for example, listed in the tab of `HTML/CSS` of each component. But we can explore that later.

clayui.com/src/components/LayoutNav/LayoutNav.js
```diff
@@ -74,6 +74,18 @@ export default (location) => {
...
 						</li>
+						<li className="nav-item">
+							<a
+								className="nav-link"
+								href="/scss-api/index.html"
+								rel="noopener noreferrer"
+								target="_blank"
+							>
+								<span className="c-inner" tabIndex="-1">
+									{'Scss API'}
```
matuzalemsteles, line 85: Shouldn't we call this the `Sass API`?

clayui.com/gatsby-node.js
pat270: @matuzalemsteles this is only in development mode for html files. I found an old issue on the gatsby repo where it errors when using yarn develop. This was the workaround others used. I don't think they fixed it.

clayui.com/src/components/LayoutNav/LayoutNav.js
pat270: Probably, I should fix it.

clayui.com/gatsby-node.js
matuzalemsteles: No problem, I'll check this later if we still need it, the our Gatsby version is already getting out of date and they will release the v4 version, and we're still on v2.24 😅

clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
pat270: I didn't know they had an API we could use. I have no idea how to use it, but it sounds good.

matuzalemsteles: Yeah, this seems to be possible just by calling `sassdoc.parse` which will generate the raw data in a JSON.

matuzalemsteles:
Hey @pat270 I don't see a problem in merging this now as we can deliver great value with it and we can create some issues to integrate better with clayui.com to improve the user experience or just stay like that too if the developers prefer.

pat270:
@matuzalemsteles sounds good, going to merge it.

PR #4273, opened 2021-09-16
fix(@clayui/tooltip): forces tooltip to be repositioned when the x-axis is modified

### LINKED ISSUE #4249: [Tooltip] The tooltip is off-center with respect to the application menu button at browser width <1250px
This issue was discovered fixing this [other issue](https://issues.liferay.com/browse/LPS-138306), therefore this part should be deployed.

- Sign in as admin
- Go to Application Menu > Applications > Remote Apps 
- Resize browser to width less than 1250px
- See the tooltip in the Application Menu button
Captura de pantalla 2021-09-01 a las 12 39 11

The tooltip should be centered with respect to the button.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.x  |

matuzalemsteles:
Fixes #4249

The idea here is to force `align` to recalculate the Tooltip position in a position contrary to the center alignment which can happen the x-axis is modified because the Tooltip is on a smaller screen, this avoids having the x-axis modified by choosing the best position.

This fixes when you are in the top or bottom position. See the List Page demo example and play around with increasing and decreasing the Storybook Sidebar to decrease content and reproduce reported use cases, Tooltip has been added at both ends of the Management Toolbar, in Checkbox and in the Add button.

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -48,6 +48,12 @@ const ALIGNMENTS_INVERSE_MAP = {
 	trbr: 'bottom-right',
 } as const;
 
+const ALIGNMENTS_FORCE_MAP = {
+	...ALIGNMENTS_INVERSE_MAP,
+	bctc: 'top-left',
+	tcbc: 'bottom-left',
+} as const;
+
 interface IState {
 	align?: typeof ALIGNMENTS[number];
 	message?: string;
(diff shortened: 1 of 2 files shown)


julien (commented):
I tested the "List page" demo and I think the result is what we want. 

On a side note (and that's just my opinion), having a JS component that does what the browser does out of the box seems like a bad idea :shrug:

julien (approved):

PR #4276, opened 2021-09-17
fix(@clayui/css): Atlas Global Variables define all global variables for easier copy and paste for those extending Clay CSS

### LINKED ISSUE #4195: @clayui/css Atlas Theme variables shouldn't rely on Base Theme variables
Atlas Theme relies on Base Theme variables to generate `atlas.css`. We should remove this dependency because:

1) When you want to extend Atlas you need to search through two variable directories to figure out what's going on (similar to classic and styled theme). This is pretty annoying.
2) We are unsetting a lot of properties in Atlas that were set in Base Theme.
3) Easier to copy and paste variables into a theme that extends Clay CSS without running into variable not defined errors.
4) There are more I just can't think of them.

pat270:
Decided not to define every variable in Atlas due to https://github.com/liferay/clay/issues/4195#issuecomment-897202074. This only defines all globals in Atlas to minimize the `Undefined variable` error due to missing global variables.

fixes #4195

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_globals.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_globals.scss
@@ -12,60 +12,32 @@ $c-unset: $clay-unset !default;
 
 $clay-unset-placeholder: clay-unset-placeholder !default;
 
+// Settings
+
 $atlas-theme: true !default;
 
+$enable-bs4-deprecated: true !default;
 $enable-lexicon-flat-colors: true !default;
 $enable-scaling-components: true !default;
+$enable-c-inner: true !default;
+$scaling-breakpoint-down: sm !default;
 
-$enable-rounded: true !default;
+$enable-caret: false !default;
+$enable-deprecation-messages: true !default;
 $enable-gradients: false !default;
+$enable-grid-classes: true !default;
+$enable-pointer-cursor-for-buttons: true !default;
+$enable-prefers-reduced-motion-media-query: true !default;
+$enable-print-styles: true !default;
+$enable-responsive-font-sizes: false !default;
+$enable-rounded: true !default;
 $enable-shadows: true !default;
 $enable-transitions: true !default;
+$enable-validation-icons: true !default;
 
-$grid-breakpoints: (
-	xs: 0,
-	sm: 576px,
-	md: 768px,
-	lg: 992px,
-	xl: 1280px,
-) !default;
+// Deprecated, no longer affects any compiled CSS
 
-$container-max-widths: (
-	sm: 540px,
-	md: 720px,
-	lg: 960px,
-	xl: 1248px,
-) !default;
-
-$grid-gutter-width: 24px !default;
-
-$moz-osx-font-smoothing: grayscale !default;
-$webkit-font-smoothing: antialiased !default;
-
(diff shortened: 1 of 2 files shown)


packages/clay-css/src/scss/atlas/variables/_globals.scss
```diff
@@ -12,60 +12,32 @@ $c-unset: $clay-unset !default;
 
 $clay-unset-placeholder: clay-unset-placeholder !default;
 
+// Settings
+
```
matuzalemsteles, line 16: Hey @pat270 sounds good to me, but a question, all these variables we added are coming from globals, right? I have the same concern you raised in the issue about having to keep two files and this can get confusing as we have to iterate and is subject to more failure.

I understand that the problem here is to lower errors like `Undefined variable` because the developer might be trying to extend from `atlas`, to avoid that then he would have to add the globals together right? Why not have an atlas "endpoint" that joins the files needed for an atlas theme to work or just join globals and the atlas variables? wouldn't that make more sense?

pat270: We do that with `atlas.scss` and `_atlas-variables.scss`. This pr isn't necessary. It's a quality of life adjustment for devs that want to extend Atlas. In Liferay Themes, there is only one "safe" file for overwriting variables, `_clay_variables.scss`. It's the first variable import in `atlas.scss`. 

```
@import '_license-text';

@import 'functions/_global-functions';

@import '_clay_variables.scss'; // <---

@import 'atlas/_variables';

@import '_variables';

@import '_mixins';

@import '_components';
```

It gets annoying when you copy paste a component map you want to modify, but errors due to a global variable missing at the top of `_clay_variables.scss` (e.g., I only copied over the global variables defined in Atlas). Currently Atlas only changes the global variables that need to be changed. By defining all variables available in both themes, it makes it a lot easier to copy paste and go.

matuzalemsteles: I see. I understand this is to improve the development experience and not anything against it but can we do something that we can solve without needing to duplicate unused variables? I imagine that if in the global variables of the atlas we already import the global variables, that could resolve it?

This is also a case where one just wants to import the `@import 'atlas/_variables';` right?

pat270: Can't think of another way other than documenting copy paste https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/atlas/variables/_globals.scss and https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/variables/_globals.scss at the top of your `_clay_variables.scss` file. It would result in duplicates for them.

Inserting `_clay_variables.scss` later in the import chain will cause the `!default` pattern not to work if they want to make their theme extendable or if they created a variable theme based on it due to Clay using it.

matuzalemsteles: I thought Sass could avoid import duplicates but in fact, it shouldn't anyway... This makes sense @pat270, maybe to avoid all this we would have to modify the import order and locations. Well, I'm fine with moving on with that.

matuzalemsteles (approved):

PR #4278, opened 2021-09-17
fix(@clayui/css): Atlas removes Sassdoc comments. All documentation t…

### LINKED ISSUE #4275: @clayui/css: Atlas remove Sassdoc comments from variables
All available variables are defined in the base theme. We shouldn't need to output Atlas variable documentation for our scss-api docs since it will all be repeated.

pat270:
…hat gets generated should go in Base

fixes #4275


matuzalemsteles (approved):
LGTM

PR #4280, opened 2021-09-20
fix(@clayui/css): Global Variables deprecate unused Bootstrap 4 variables

### LINKED ISSUE #4279: @clayui/css: Document Deprecated Global Variables
There are a lot of variables from Bootstrap that aren't being officially used. We should deprecate them. This one is for Global Variables. Some examples:

```
$escaped-characters: (('<', '%3c'), ('>', '%3e'), ('#', '%23')) !default;
$enable-hover-media-query: false !default;
$caret-width: 0.3em !default;
$caret-vertical-align: $caret-width * 0.85 !default;
$caret-spacing: $caret-width * 0.85 !default;
```

pat270:
fixes #4279

### DIFF
--- a/packages/clay-css/src/scss/variables/_custom-forms.scss
+++ b/packages/clay-css/src/scss/variables/_custom-forms.scss
@@ -180,7 +180,9 @@ $custom-control-primary-label-text: map-deep-merge(
 
 $custom-checkbox-indicator-border-radius: $border-radius !default;
 
-$custom-checkbox-indicator-icon-checked: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='8' height='8' viewBox='0 0 8 8'><path fill='#{$custom-control-indicator-checked-color}' d='M6.564.75l-3.59 3.612-1.538-1.55L0 4.26l2.974 2.99L8 2.193z'/></svg>") !default;
+$custom-checkbox-indicator-icon-checked: clay-svg-url(
+	"<svg xmlns='http://www.w3.org/2000/svg' width='8' height='8' viewBox='0 0 8 8'><path fill='#{$custom-control-indicator-checked-color}' d='M6.564.75l-3.59 3.612-1.538-1.55L0 4.26l2.974 2.99L8 2.193z'/></svg>"
+) !default;
 $custom-checkbox-indicator-icon-checked-bg-size: null !default;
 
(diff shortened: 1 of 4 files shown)


packages/clay-css/src/scss/variables/_custom-forms.scss
```diff
@@ -191,14 +193,18 @@ $custom-control-indicator-indeterminate-border-color: $custom-checkbox-indicator
...
 $custom-checkbox-indicator-icon-indeterminate-bg-size: null !default;
 
 // Custom Radio
 
 $custom-radio-indicator-border-radius: 50% !default;
 
-$custom-radio-indicator-icon-checked: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='12' height='12' viewBox='-4 -4 8 8'><circle r='3' fill='#{$custom-control-indicator-checked-color}'/></svg>") !default;
+$custom-radio-indicator-icon-checked: clay-svg-url(
+	"<svg xmlns='http://www.w3.org/2000/svg' width='12' height='12' viewBox='-4 -4 8 8'><circle r='3' fill='#{$custom-control-indicator-checked-color}'/></svg>"
+) !default;
```
matuzalemsteles, line 207: @pat270 just a thought in case someone is using it, maybe it will pass with `url()`, wouldn't we have to keep that yet? maybe just a fallback if passing `url` still use `escape-svg` here. What do you think?

matuzalemsteles (approved):
Overall LGTM, I just left a comment to verify the case.

packages/clay-css/src/scss/variables/_custom-forms.scss
pat270: You're right, we should account for that. I'll update this pr today.

pat270: @matuzalemsteles I left the `escape-svg()` functions inside the components. We can run `clay-svg-url` through `escape-svg` without any errors. I tried to remove all uses of it, but I think we need to wait until the next major version.

matuzalemsteles: @pat270 we can keep this on our radar to remove `escape-svg` and replace it in the next major version with no problems.

matuzalemsteles (approved):
LGTM

matuzalemsteles:
@pat270 feel free to merge if ready.

PR #4286, opened 2021-09-22
feat(@clayui/css): Move .hide from liferay-portal to clay-css

edalgrin:
In the attempt of reduce the CSS footprint in Liferay Portal, we are moving CSS classes where they make more sense.

https://issues.liferay.com/browse/LPS-134071

---

In this case, we are moving the class `.hide`, an alias of `.d-none`.

https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled/css/base/_util.scss#L1-L3

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_utilities-functional-important.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_utilities-functional-important.scss
@@ -186,6 +186,10 @@
 	}
 }
 
+.hide {
+	display: none !important;
+}
+
 // Display Print
 
 @media print {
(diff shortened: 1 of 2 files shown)


edalgrin:
> It makes sense to me, anyway I thought it was already in Clay 😅. @pat270 what do you think?

@matuzalemsteles this might help https://github.com/liferay-frontend/liferay-portal/pull/1463#discussion_r712421080

matuzalemsteles:
@edalgrin perfect, @pat270 approved, so I'll go ahead with that. Thanks!

matuzalemsteles:
@edalgrin as soon as you send more PRs could you create an equivalent issue too? this is just so that issue can track with Jira and we can make it visible that you've worked with this in the releases changelog.

edalgrin:
Sure!

PR #4287, opened 2021-09-22
fix(@clayui/css): Remove caret from .dropdown-toggle

edalgrin:
In the attempt of reduce the CSS footprint in Liferay Portal, we are moving CSS classes where they make more sense.

https://issues.liferay.com/browse/LPS-134071

---

In this case, we are removing the caret from `.dropdown-toggle` since it's overwritten in DXP

https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled/css/portal/_dropdowns.scss#L21-L29

### DIFF
--- a/packages/clay-css/src/scss/components/_dropdowns.scss
+++ b/packages/clay-css/src/scss/components/_dropdowns.scss
@@ -7,8 +7,6 @@
 
 .dropdown-toggle {
 	white-space: nowrap;
-
-	@include caret();
 }
 
 .dropdown-header {
@@ -246,10 +244,6 @@
 		margin-top: 0;
 		top: auto;
 	}
-
-	.dropdown-toggle {
-		@include caret(up);
-	}
 }
 
 .dropright {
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
Hey @edalgrin is this an old implementation that doesn't work or is no longer used? I don't remember seeing this on DropDown but @pat270 can check it out better than me. Anyways I think if this is still used why not remove it from DXP and just keep it here? since this has already been used on Clay, it should probably be used by other people, just thinking.

edalgrin:
I think we already removed in Cadmin and looking at the [lexicon specs](https://liferay.design/lexicon/core-components/dropdowns/) I didn't find anything about the caret, may it be something from bootstrap?

matuzalemsteles:
Yeah, I also thought it was weird I hadn't seen it before, maybe it's inherited from bootstrap or some old version we had, @pat270 you can confirm here 😁

pat270:
Yeah this is Bootstrap's. It adds a caret down icon on `.dropdown-toggle` and only shows up when `$enable-caret: true`. We don't use this at all, but I think we should remove it in the next major version.

pat270:
Ok going to close this one, we will save it for 4.x

PR #4290, opened 2021-09-22
fix(@clyaui/tooltip): fixes error when not forcing Tooltip repositioning when X axis is modified

### LINKED ISSUE #4289: TooltipProvider does not force Tooltip repositioning when X axis is changed
Well, we've added the possibility to reposition the Tooltip (#4273) when the X axis is modified to fix issue #4249, but the behavior is still being reproduced.

matuzalemsteles:
Fixes #4289

I think I ended up commenting with `adjustX` to false doing tests 🤦‍♂️ and I also didn't notice that in the Storybook of the PR build it wasn't working. I tested it on DXP and it works fine too.

Screen Shot 2021-09-22 at 20 21 04

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -343,7 +343,7 @@ const TooltipProvider: React.FunctionComponent<
 
 			const alignment = doAlign({
 				overflow: {
-					adjustX: false,
+					adjustX: autoAlign,
 					adjustY: autoAlign,
 				},
 				points,


julien:
@matuzalemsteles I actually saw in the previous PR that `adjustX` had been set to `false` but I really thought that is what you intended to do, anyway I tested this and it works as expected - thanks. I'll add a screenshot in the original issue.

julien:
@matuzalemsteles after merging I noticed the title of the commit had a small typo, I don't think it's a huge problem but just wanted to report.

PR #4297, opened 2021-09-23
fix(@clayui/css): Cards and Panels remove decision making if statements from border-radius and border-width

### LINKED ISSUE #4295: Can't use css variables in $border-radius.
Setting $border-radius to a css custom property causes clay to throw an error during compile.

$border-radius: var(--border-radius);

pat270:
fixes #4295

- Global Functions `math-sign` should return a negative calc value if a CSS custom property is passed in
- Inverse Sass variable values should use the `math-sign` function

### DIFF
--- a/packages/clay-css/src/scss/components/_cards.scss
+++ b/packages/clay-css/src/scss/components/_cards.scss
@@ -186,14 +186,14 @@
 @if ($enable-bs4-deprecated) {
 	.card-header-tabs {
 		border-bottom: 0;
-		margin-bottom: -$card-spacer-y;
-		margin-left: -$card-spacer-x / 2;
-		margin-right: -$card-spacer-x / 2;
+		margin-bottom: math-sign($card-spacer-y);
+		margin-left: calc(#{math-sign($card-spacer-x)} / 2);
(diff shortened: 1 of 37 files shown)


packages/clay-css/src/scss/components/_cards.scss
```diff
@@ -18,15 +18,13 @@
...
+	.autofit-col {
+		&:first-child {
+			@include border-left-radius($card-inner-border-radius);
+		}
 
-			&:last-child {
-				@include border-right-radius($card-inner-border-radius);
-			}
+		&:last-child {
+			@include border-right-radius($card-inner-border-radius);
```
matuzalemsteles, line 27: @pat270 the only side effect this adds is that it will create the classes even when the `border-radius` is zero right?

matuzalemsteles (approved):
LGTM!

packages/clay-css/src/scss/components/_cards.scss
pat270: @matuzalemsteles yes, I tried to save some CSS a long time ago. It doesn't really work well.

PR #4301, opened 2021-09-24
fix(@clayui/css): Forms `fieldset[disabled] .form-control` should use values from the `$input` Sass map and move rule-set closer to the `.form-control` rule-set

### LINKED ISSUE #4285: @clayui/css: Forms `fieldset[disabled] .form-control` should use `$input` map values
https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_forms.scss#L391-L417

`fieldset[disabled]` uses `$input-label-disabled-*` variables. Themes that customize `form-control` using the newer map variable api doesn't properly translate to this component.

pat270:
fixes #4285

This also fixes a similar issue with `fieldset[disabled] label`. Forms `label.disabled` should use properties declared in `$input-label` map and move the `label.disabled` rule-set closer to the `label` rule-set.

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_forms.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_forms.scss
@@ -69,6 +69,22 @@ label {
 	@include clay-css($cadmin-form-control-label-text);
 }
 
+// Label Disabled
+
+fieldset[disabled] label,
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
LGTM

PR #4307, opened 2021-09-27
feat(@clayui/core): implements DnD in TreeView

matuzalemsteles:
This is the final implementation for the TreeView implementation of DnD, this PR contains some commits that were squash to reduce the noise a bit, I like to keep it because normally the commit gets too big and confusing to understand but we can talk about that later. The entire context of implementing this functionality is in PR #4255.

I'm also sending another PR to keep the other PR's history available.


julien (approved):
LGTM

matuzalemsteles:
Well I'm going ahead with that and merging but feel free to look at the PR, comment and if you need anything else we can send you anytime.

@bryceosterhaus feel free to look at the implementation as well.

PR #10, opened 2015-09-02
#8 - Document sidebar component

pat270:

PR #4316, opened 2021-09-29
feat(@clayui/core): adds `onLoadMore` API to TreeView

matuzalemsteles:
Well, this is a very early draft for implementing "load more" or allowing the TreeView to load a large amount of data and still work fine without causing any slowdowns or crashes, yet this component doesn't implement any improvement related to loading a render a big payload.

One of the first attempts here is to allow you to load parts of the structure asynchronously, the final implementation is not yet ready to deal with this, there are some tradeoffs we will have to deal with and we can also try different strategies on how to approach this:

- The TreeView shouldn't handle the responsibility of loading asynchronous data or handling it, it's data agnostic. This is responsible for the `useResource` hook which will load the data with the possibility of aggregating the data and fetching using a "cursor". Currently `useResource` doesn't support this type of loading and also doesn't know how to handle multiple requests what we would have to do here is create a hook like `useTreeResource` to handle the data and aggregate.
- The TreeView uses the return from `onLoadMore` which must be a promise with the items and internally we do the aggregation and update the state.

There are some tradeoffs that we still have to deal with, the implementation of "load more" will harm the multiple selection, because we depend on creating the structure layout using the flat structure to reduce very high operating costs, and not using this will be slow but there are strategies that can mitigate this problem, such as treating the selected items only the visible ones and providing methods that retrieve all the ids of the selected items, because we would have to traverse the entire tree looking for these ids, so it would be less costly to do this every time select an item.

There is also a problem that "load more" does not solve, as the nodes are being expanded, it gets slower and slower. I did some tests on the storybook demo that renders a large number of nodes distributed in 10 items per level and the maximum depth of 5 levels, we reached 111,110 items, when having at least 120 nodes open up to half of this it is possible to notice a slowness too high, reaching around ~200ms for React to complete the commit in the DOM, this number can also be higher if you use the performance to evaluate instead of checking the React Profiler. We can also see a slow page with a smaller amount of expanded items if the same level has much more than 10, so maybe having 150 items may be that just opening 10 nodes can be very slow, the "load more" doesn't solve it because it just saves the using memory but not rendering, I still insist that the best way to handle rendering is to use virtualization.

There are many performance fronts that we are tackling so far, the multiple selection implementation, the expandable APIs, and the DnD implementation that avoids traversing the tree every time you do an operation but we haven't solved the rendering part yet.

### DIFF
--- a/packages/clay-core/src/tree-view/useTree.ts
+++ b/packages/clay-core/src/tree-view/useTree.ts
@@ -109,20 +109,27 @@ export function useTree<T>(props: ITreeProps<T>): ITreeState<T> {
 	};
 }
 
-// RFC 6902 4.4
+// Operation of `move` value to the same document structure, removing from
+// `from` and then adding to `path`.
+// RFC 6902 (JSON Patch) 4.4
 type PatchMove = {
 	op: 'move';
 	from: Array<number>;
 	path: Array<number>;
 };
 
-// RFC 6902 4.1
+// Operation of `add` value to the document structure.
+// RFC 6902 (JSON Patch) 4.1
 type PatchAdd = {
 	op: 'add';
 	path: Array<number>;
 	value: unknown;
 };
 
+// Patch refers to the implementation of RFC 6902 operations (JSON Patch)
+// https://datatracker.ietf.org/doc/html/rfc6902, we just borrow the document
+// structure to make partial updates to a JSON document.
+// Implementation Detail https://github.com/liferay/clay/pull/4254.
 type Patch = PatchMove | PatchAdd;
 
 export function createImmutableTree<T extends Array<Record<string, any>>>(
--- a/packages/clay-core/src/tree-view/useTree.ts
+++ b/packages/clay-core/src/tree-view/useTree.ts
@@ -35,6 +35,7 @@ export interface ITreeProps<T>
 
 export interface ITreeState<T> extends Pick<ICollectionProps<T>, 'items'> {
 	expandedKeys: Set<Key>;
+	insert: (path: Array<number>, value: unknown) => void;
 	open: (key: Key) => void;
 	reorder: (from: Array<number>, path: Array<number>) => void;
 	selection: IMultipleSelectionState;
@@ -62,7 +63,15 @@ export function useTree<T>(props: ITreeProps<T>): ITreeState<T> {
 	const reorder = (from: Array<number>, path: Array<number>) => {
 		const tree = createImmutableTree(items, props.nestedKey!);
 
-		tree.produce(from, path);
+		tree.produce({from, op: 'move', path});
+
+		setItems(tree.applyPatches());
+	};
+
+	const insert = (path: Array<number>, value: unknown) => {
+		const tree = createImmutableTree(items, props.nestedKey!);
+
+		tree.produce({op: 'add', path, value});
 
 		setItems(tree.applyPatches());
 	};
@@ -91,6 +100,7 @@ export function useTree<T>(props: ITreeProps<T>): ITreeState<T> {
 
 	return {
 		expandedKeys,
+		insert,
 		items,
 		open,
 		reorder,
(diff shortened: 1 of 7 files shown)


matuzalemsteles:
cc @nhpatt @dsanz

matuzalemsteles:
Well, I think this is ready for a review 🙂, sent some new commits and rebase it with the master to get the implementation of DnD which implements the `createImmutableTree` method to facilitate the implementation of the second strategy I'm following here, `onLoadMore ` returns a promise with the items that we will add to the tree structure, this allows you to create your own rule to load the data depending on the structure and according to your rule.

There's no way for us to know when to load more data, in the `headless-admin-taxonomy` APIs it seems to have a property like `numberOfTaxonomyCategories` that shows if there are "children", but I'm not sure if this is true for all other cases of use in DXP, this seems to be just a pattern for Categories that are linked to Vocabulary, so it's up to whoever listens to `onLoadMore` to decide whether to load the data or not.

`onLoadMore` follows a very simple principle for calling the callback, if there is no `group` rendered, assumes it might have data to be loaded and then the callback is called, see the async load example in Storybook.

Although this is pretty simple I still think it's worth adding the `useTreeResource` implementation, because in the `onLoadMore` implementation only the developer can only use `fetch` but `useResource` has more features, like retry, network status, cache and other strategies.

```js
const {items, loadMore} = useTreeResource(...);

<TreeView
  items={items}
  onLoadMore={loadMore}
>
  {...}
</TreeView>
```

So far I haven't implemented the loading state to show that the TreeView or Node is loading data either, we don't have that defined in the spec. cc @drakonux

drakonux:
Hey, @matuzalemsteles, it wasn't defined because I needed to know what options we considered to enhance the performance vs. the old one. So, I wrote a couple of lines about it but not in detail. 

Therefore, if you need it soon, please open a request in LEXI project with your deadline, that will allow us to prioritize it 💪

packages/clay-core/src/tree-view/useTree.ts
```diff
@@ -100,12 +110,21 @@ export function useTree<T>(props: ITreeProps<T>): ITreeState<T> {
...
 
 // RFC 6902 4.4
-type Patch = {
+type PatchMove = {
 	op: 'move';
 	from: Array<number>;
 	path: Array<number>;
 };
 
+// RFC 6902 4.1
```
julien, line 119: Nothing wrong here, but I'm just wondering if those "RFC" related comments are going to cause confusion. I know what it means, but that's because I'm in the squad and have context about this, but I'm thinking of anyone using Clay (or simply in another squad).

julien (approved):
LGTM, the performance is pretty good when looking at the new "stories", and not much code was needed to add this: good job.

packages/clay-core/src/tree-view/useTree.ts
matuzalemsteles: Yeah I add them to add references and create context, the idea is that people don't deviate from the implementation in the future or remember that this follows a pattern rather than just created by us, I can add more comments to also reduce confusion. What do you think?

julien: Well, looking at the code (once one is familiar with the component's API) it's clear what these patches do ... but I think adding a small comment would help yes, good idea, that works for me.

bryceosterhaus: Viewing this without context, I'm not sure what the RFC is referring to. Would it be possible to include a permalink to something here as well? Something like `// RFC 6902 https://github.com/liferay/clay/issues/10`

matuzalemsteles: Yeah, it's an option too.

julien: Thanks for the feedback @bryceosterhaus.

@matuzalemsteles the comments you added look good enough for me. Since our TreeView component is still in progress, I'll wait to get a "second round" of comments for Bryce - since he doesn't have context if this clarified his doubts I think we're all good.

julien:
@matuzalemsteles last time I reviewed it, I approved, I could have merged it, sorry for the delay.

PR #4321, opened 2021-10-01
feat(@clayui/css): Alerts adds components `.alert-inline`, `alert-autofit-stacked`, `alert-autofit-stacked-sm-down`, `alert-autofit-stacked-xs-down`, `alert-indicator-start`

### LINKED ISSUE #4294: [Alert] Inline Actions
## Proposal

One line
Please read:

- [LEXI-1232](https://issues.liferay.com/browse/LEXI-1232)
- [Inline actions for Alerts ](https://liferay.atlassian.net/l/c/5WgL2bmr)

Most of the alerts used in DXP have two or more lines, mainly because the max-width set constrains them. Providing this option will allow product teams to have wider alerts with actions.

Besides, the Dialect team requested this for their use cases. 

We recommend to not use inline actions when there are more than two lines, in that case to use block actions is a better choice. 

**Responsive Behavior**
Could we change an alert with inline actions to block actions when the screen space is narrowed?

pat270:
fixes #4294

This adds components so we can create Alerts according to the specs mentioned in the issue. These new alerts should still work if tossed into the `alert-notifications-fixed` container.

The markup:

alert-inline-actions

```html
<div class="alert alert-dismissible alert-indicator-start alert-inline alert-success">
  <div class="alert-autofit-row autofit-row">
      <div class="autofit-section">
        <span class="alert-indicator"></span>
        <strong class="lead">Success:</strong>
        <span>File uploaded. Would you like to add categories automatically?</span>
      </div>
    </div>
    <div class="autofit-col">
      <div class="autofit-section">
        <div class="btn-group">
          <div class="btn-group-item">
            <button class="btn btn-success btn-sm" type="button">Add</button>
          </div>
        </div>
      </div>
    </div>
  </div>
  <button aria-label="Close" class="close" data-dismiss="alert" type="button"></button>
</div>
```

We can use these new utilities to layout the current Toast alert if we wanted.

alert-current

```html
<div class="alert alert-autofit-stacked alert-indicator-start alert-info">
  <div class="alert-autofit-row autofit-row">
      <div class="autofit-section">
        <span class="alert-indicator"></span>
        <strong class="lead">With a button</strong>
        This is an alert with a button!
      </div>
    </div>
    <div class="autofit-col">
      <div class="autofit-section">
        <div class="btn-group">
          <div class="btn-group-item">
            <button class="alert-btn btn" type="button">View</button>
          </div>
        </div>
      </div>
    </div>
  </div>
</div>
```

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_alerts.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_alerts.scss
@@ -1,5 +1,5 @@
 $alert-padding-x: 1rem !default;
-$alert-padding-y: 1.03125rem !default;
+$alert-padding-y: 0.9375rem !default;
 
 // For top only border use: 2px 0 0 0
 $alert-border-width: 0.0625rem !default;
@@ -27,7 +27,7 @@ $alert-close: map-deep-merge(
 	$alert-close
 );
 
-$alert-font-size: 1rem !default; // 14px
+$alert-font-size: 0.875rem !default; // 14px
 
 $alert-lead-font-size: $alert-font-size !default;
 $alert-lead-font-weight: $font-weight-semi-bold !default;
@@ -36,7 +36,7 @@ $alert-link-decoration: underline !default;
 $alert-link-hover-decoration: none !default;
 $alert-link-font-weight: $font-weight-semi-bold !default;
 
-$alert-indicator-font-size: inherit !default;
+$alert-indicator-font-size: 1rem !default;
 
 $alert-title: () !default;
 $alert-title: map-deep-merge(
@@ -70,8 +70,8 @@ $alert-footer: map-deep-merge(
 $alert-fluid-container: () !default;
 $alert-fluid-container: map-merge(
 	(
-		padding-bottom: 1.0625rem,
-		padding-top: 1.0625rem,
+		padding-bottom: 0.96875rem,
+		padding-top: 0.96875rem,
 	),
 	$alert-fluid-container
 );
--- a/packages/clay-css/src/scss/atlas/variables/_alerts.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_alerts.scss
@@ -27,7 +27,7 @@ $alert-close: map-deep-merge(
 	$alert-close
 );
 
-$alert-font-size: 0.875rem !default; // 14px
+$alert-font-size: 1rem !default; // 14px
 
 $alert-lead-font-size: $alert-font-size !default;
 $alert-lead-font-weight: $font-weight-semi-bold !default;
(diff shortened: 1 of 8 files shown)


matuzalemsteles (commented):
@pat270 in general LGTM, just a use case that looks like we need to support looking at the document but I think we already support it https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1891729698/LEXI-1232+Alert+Inline+actions#Usage, but testing this in this implementation it looks like the alert size is fixed and the text margins don't seem to be respected.

I tried to investigate what is causing this but it seems to be related to the margins added to `.alert-autofit-row` when it is `inline`. But I'm not sure if that would really be it because it would break when there's only one line.

pat270:
@matuzalemsteles I can't seem to visualize what you are saying. Do you have a screenshot or an example I can copy paste?

matuzalemsteles:
of course it is here!

Screen Shot 2021-10-04 at 14 32 27

pat270:
@matuzalemsteles I see now. This is as intended since Lexicon added a rule about maximum two lines for inline alert. I just wanted to keep it simple, but we can add more space between text and top/bottom border. I just need to adjust negative margins on `alert-autofit-row`.

matuzalemsteles:
@pat270 no problem, I just tested it because in the confluence documentation there is this possibility.

julien:
LGTM, but I'd also consider increasing space between the text and the "border"

packages/clay-css/src/scss/atlas/variables/_alerts.scss
```diff
@@ -27,7 +27,7 @@ $alert-close: map-deep-merge(
 	$alert-close
 );
 
-$alert-font-size: 0.875rem !default; // 14px
+$alert-font-size: 1rem !default; // 14px
```
matuzalemsteles, line 30: @pat270 is this expected? I noticed that the alerts will undergo a visual change, it seems to have gotten bigger, in addition to the font size that has also changed, see this comparison:

|. Before | After |
|--------|-------|
| Screen Shot 2021-10-05 at 18 51 37 | Screen Shot 2021-10-05 at 18 52 24 |

matuzalemsteles: Oh ok, no problem.

pat270: This one should be good to go. I changed the wrong variable, should have been `$alert-indicator-font-size`. I also changed the padding so alert height should be the same.

matuzalemsteles (approved):
LGTM!

PR #4322, opened 2021-10-04
fix(@clayui/form): change the Dual List Box ordering side following the right side as the default behavior

### LINKED ISSUE #4312: Allow configuring the reordering of any side in DualListBox component
Currently the component implemented in React of DualListBox does not allow configuring the reordering, this is enabled by default only for the left side but some time ago the configuration was requested to modify the sides, for more context see #3985.

The purpose of this "issue" is to allow you to choose which side will have the reordering so it allows the team to choose depends on the context.

matuzalemsteles:
Fixes #4312

See issue https://github.com/liferay/clay/issues/4312 to understand the context of the change and why we are following this as a fix rather than not changing this default behavior now.

### DIFF
--- a/packages/clay-form/src/DualListBox.tsx
+++ b/packages/clay-form/src/DualListBox.tsx
@@ -150,9 +150,7 @@ const ClayDualListBox: React.FunctionComponent<IProps> = ({
 						onItemsChange([newLeftItems, rightItems])
 					}
 					onSelectChange={handleLeftSelectedChange}
-					showArrows
 					size={size}
-					spritemap={spritemap}
 					value={leftSelected}
 				/>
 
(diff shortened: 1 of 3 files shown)


julien (approved):
LGTM

julien:
I might wait for @carloslancha to have a look, because I know he was interested, but I'm OK with this change.

carloslancha:
LGTM. Just a small nit: [here](https://github.com/liferay/clay/pull/4322/files#diff-cf0c416b55a9a2267c0e760adfe730dde3c16a4722a3b36e9d9e6495335da080R154) we don't need to pass `spritemap` anymore.

julien:
@carloslancha I'm not 100% sure, but I think we still need to pass it, because it's not wrapped in a `ClayIconSpriteContext`, meaning that if you don't pass this prop, you're likely to see a warning and no icons.

carloslancha:
@julien we weren't passing it in the first `ClaySelectBox` before because `showArrows` were false. In `ClaySelectBox` the `spritemap` is only needed if `showArrows` is set to true.

It's not important and it's not harmful to pass it, just commented about it to follow the same behavior/pattern as before the change.

julien:
@carloslancha you're right, my bad. I was looking at line 210, not the highlighted one (154), sometimes it's not easy to see everything in GitHub's UI

julien:
All good, thanks @matuzalemsteles

PR #4327, opened 2021-10-06
docs(release): add a note about `lerna version`

julien:
@matuzalemsteles sending this to avoid future errors 
during a release.


matuzalemsteles (approved):
LGTM! Thanks @julien!

PR #4333, opened 2021-10-07
chore(@lexicon-ux): Don't bump the version number used for copyright …

### LINKED ISSUE #4332: @lexicon-ux republishing a failed publish to the maven repo causes copyright text version to be out of sync
Due to https://github.com/liferay/clay/blob/753b7925ed4c10d10335f832338d0cd8dced969e/tasks/release.js#L88

pat270:
…text in `release:build`, it should use the version number in `package.json` as is.

chore(@lexicon-ux): Re-run `release:build` before `release:publish` so copyright text version numbers get updated when `release:git` bumps `package.json` version.

fixes #4332

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -104,13 +104,14 @@ gulp.task(
 						if (answers.publish) {
 							var args = [
 								'release:git',
+								'release:build',
 								'release:publish',
 								cb
 							];
 
 							if (answers.packageManagers) {
-								args.splice(2, 0, 'maven-publish');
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
@pat270 no, I don't think it's necessary to spend a lot of time on it, we can go with it that way without a problem.

pat270:
Thanks @matuzalemsteles I'm going to merge it.

@julien I don't think we can upgrade 1.x to dart-sass. 7.0 is using `libsass` or `ruby-sass` which is deprecated.

PR #4339, opened 2021-10-11
feat(@clayui/css): clay-alert-variant mixin add option to customize alert-indicator and use placeholders for alert variants

### LINKED ISSUE #4337: @clayui/css clay-alert-variant mixin add option to customize alert-indicator
`.alert-indicator` is missing from the `clay-alert-variant` mixin. Dialect has an announcement component that looks similar to an alert. They can leverage unused Clay alert variants to create it (alert-primary) and map their `announcement` component to it.

dialect-announcement

We should also place alert-variants inside Sass placeholders for cleaner extension.

pat270:
fixes #4337

### DIFF
--- a/packages/clay-css/src/scss/mixins/_alerts.scss
+++ b/packages/clay-css/src/scss/mixins/_alerts.scss
@@ -48,6 +48,8 @@
 
 	$alert-indicator: setter(map-get($map, alert-indicator), ());
 
+	$alert-indicator-lead: setter(map-get($alert-indicator, lead), ());
+
 	$alert-btn: setter(map-get($map, alert-btn), ());
 
 	$btn-group: setter(map-get($map, btn-group), ());
@@ -106,6 +108,10 @@
 
 		.alert-indicator {
 			@include clay-css($alert-indicator);
+
+			+ .lead {
+				@include clay-css($alert-indicator-lead);
+			}
 		}
 
 		.alert-btn {


packages/clay-css/src/scss/mixins/_alerts.scss
```diff
@@ -46,6 +46,8 @@
 
 	$hr: setter(map-get($map, hr), ());
 
+	$alert-indicator: setter(map-get($map, alert-indicator), ());
```
matuzalemsteles, line 49: @pat270 have we declared in `$map` the `alert-indicator`? for example in `$alert`, I saw that we have just separated it as a variable, or here we are just adding this option but we are not using it in our implementation for example.

packages/clay-css/src/scss/mixins/_alerts.scss
```diff
@@ -102,6 +104,10 @@
 			@include clay-css($hr);
 		}
 
+		.alert-indicator {
+			@include clay-css($alert-indicator);
+		}
```
matuzalemsteles, line 109: Shouldn't the indicator also change the `+ .lead`?

packages/clay-css/src/scss/mixins/_alerts.scss
pat270: We are just adding an option and not using it. It gives devs a chance to overwrite our styles scoped to a specific variant.

pat270: I thought about it, but decided not to because it would increase the specificity of the original [`.alert-indicator`](https://github.com/liferay/clay/blob/10ca4f01024811869141cea5dfc88a0de1033446/packages/clay-css/src/scss/components/_alerts.scss#L20-L28) to `.alert .alert-indicator`.

matuzalemsteles: Yeah, you're right, I had forgotten that we have the `.alert-indicator` for use in alert-feedback too.

matuzalemsteles (approved):
LGTM

PR #4348, opened 2021-10-18
chore(@clayui/core): add support for keyboard navigation in TreeView

### LINKED ISSUE #4208: Add interactions via Keyboard to TreeView
The TreeView specification covers TreeView interactions via Keyboard, controlling the control of focus and shortcuts. To handle focus control we can use `FocusScope`.

- [Keyboard Spec](https://docs.google.com/document/d/1fggZBRGAIkGPbnkRrGEUOT0hHpkF-YjFmREee0XCoI0/edit#heading=h.5zm374rdzhh)

julien:
This is the initial implementation of keyboard navigation 
for our TreeView component. 

Fixes #4208


packages/clay-core/src/tree-view/TreeViewItem.tsx
```diff
@@ -86,6 +89,20 @@ export const TreeViewItem = React.forwardRef<HTMLDivElement, TreeViewItemProps>(
...
+							const {key} = event;
+							if (group) {
+								if (key === 'ArrowLeft') {
+									close(item.key);
+								}
+								if (key === 'ArrowRight') {
+									open(item.key);
+								}
+							}
+							if (key === 'Delete') {
```
matuzalemsteles, line 102: I'm not able to reproduce this case correctly, I'm not sure it could be because of the keyboard. The "Delete" key on the Macbook Pro keyboard, it's a Backspace too, maybe we should consider both?

matuzalemsteles (commented):
@julien overall it looks like a good initial implementation for interactions, I just have to check because `FocusScope` isn't ignoring elements with negative `tabIndex`, from the document it looks like we should ignore in the focus.

julien:
@matuzalemsteles thanks for the feedback. I'll take a look at how I can ignore those elements with a negative `tabIndex` and push the changes.

packages/clay-core/src/tree-view/TreeViewItem.tsx
```diff
@@ -86,6 +91,23 @@ export const TreeViewItem = React.forwardRef<HTMLDivElement, TreeViewItemProps>(
...
+								if (key === 'ArrowRight') {
+									open(item.key);
+								}
+							}
+							if (key === 'Backspace' || key === 'Delete') {
+								remove(item.indexes);
+							}
+							if ((key === 'R' || key === 'F2') && onRenameItem) {
+								rename(item.indexes, onRenameItem(item.name));
+							}
```
matuzalemsteles, line 109: We could try using a `switch case` here to avoid the number of conditionals, for example:

```js
switch (key) {
	case 'ArrowLeft': {
		if (group) {
			close(item.key);
		}
		break;
	}
	case 'ArrowRight': {
		if (group) {
			open(item.key);
		}
		break;
	}
	case 'Delete':
	case 'Backspace':
		remove(item.indexes);
		break;
	case 'F2':
	case 'R':
		rename(item.indexes);
		break;
	default:
		break;
}
```

packages/clay-core/src/tree-view/useTree.ts
```diff
@@ -126,11 +158,24 @@ type PatchAdd = {
...
+type PatchRemove = {
+	op: 'remove';
+	path: Array<number>;
+};
+
+type PatchRename = {
+	name: string;
+	op: 'rename';
+	path: Array<number>;
+};
```
matuzalemsteles, line 172: @julien we could instead of using the `rename` naming we can go with `replace` which follows the same RFC pattern in section 4.3 and also `replace` sounds more like an immutability oriented pattern. What do you think?

packages/clay-core/src/tree-view/useTree.ts
```diff
@@ -265,6 +310,38 @@ export function createImmutableTree<T extends Array<Record<string, any>>>(
...
+					}
+
+					break;
+				}
+
+				case 'rename': {
+					const {name, path} = patch;
+
+					const nodeToRename = nodeByPath(path);
+					nodeToRename.item.name = name;
```
matuzalemsteles, line 340: I think we can say that `onRenameItem` should return the object instead of changing a property of the object or we would have to make it configurable to remain data agnostic.

packages/clay-core/src/tree-view/TreeViewItem.tsx
julien: We could, but I see the same duplicated `if` condition which makes me say it's not "better" nor is it more readable - change my mind :smile: .

packages/clay-core/src/tree-view/useTree.ts
julien: @matuzalemsteles while trying this out I saw a problem: We can't pass the entire item to the `onRenameItem` callback because it would allow modifying important "hidden" properties like `index`, `indexes` or `key`, but at the same time if we omit passing those properties the operation "fails" when trying to rename a item in the "top" level because we loose that information. Since this was hard to solve, I'm actually passing the entire item in the callback, you can see this in https://github.com/liferay/clay/pull/4348/commits/570dfd501326e68d455902864a9460c58de29b9a

Since it's hard to explain, we can talk about it via Slack if you think it's necessary.

julien: @matuzalemsteles just to be more clear, I'm not sure it's should be the TreeView's responsibility

julien: I tried a bunch of different things, but not that :smile: Thanks for the tip

packages/clay-core/src/tree-view/TreeViewItem.tsx
```diff
@@ -86,6 +93,75 @@ export const TreeViewItem = React.forwardRef<HTMLDivElement, TreeViewItemProps>(
...
+												'.treeview-link'
+											);
+
+										firstItemElement?.focus();
+									}
+								}
+
+								if (key === ' ') {
+									selection.toggleSelection(item.key);
+								}
```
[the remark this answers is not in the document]
julien: Sure, I thought `charCode` wasn't deprecated, but if it is, then I agree we shouldn't use it.
Using those constants seems like the best idea, I think I saw that `Key.ts` file in one of our stories but somehow forgot to make use of it.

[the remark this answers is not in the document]
matuzalemsteles: No problem, I had forgotten about the `Keys` file too.

julien (commented):
LGTM

PR #4352, opened 2021-10-18
fix(@clayui/css): Border utilities `border-color` is the wrong value.…

### LINKED ISSUE #4350: @clayui/css: Atlas Border utility, the border-color should be #e7e7ed instead of #8f8f95
The border utility class colors should match the most common border color used by the theme. For Atlas it is #e7e7ed ($gray-300).

pat270:
… Table Row Variants scoped `$border-color` variable was overwriting global `$border-color` variable

fixes #4350


matuzalemsteles (approved):
LGTM!

PR #4356, opened 2021-10-19
fix(@clayui/css): Global Functions `math-sign` uses deprecated `@else…

### LINKED ISSUE #4353: @clayui/css Dart Sass deprecated @elseif syntax
```
Deprecation Warning on line 81, column 2 of src/scss/functions/_global-functions.scss: 
@elseif is deprecated and will not be supported in future Sass versions.

Recommendation: @else if
   ╷
81 │     @elseif (type-of($num) == 'string') {
   │     ^^^^^^^
   ╵
```

pat270:
…if` syntax, changed to `@else if`

fixes #4353

@matuzalemsteles I thought this one might have been tough, but it turned out to be a non issue.

### DIFF
--- a/packages/clay-css/src/scss/functions/_global-functions.scss
+++ b/packages/clay-css/src/scss/functions/_global-functions.scss
@@ -77,8 +77,7 @@
 @function math-sign($num) {
 	@if (type-of($num) == 'number') {
 		@return -($num);
-	}
-	@elseif (type-of($num) == 'string') {
+	} @else if (type-of($num) == 'string') {
 		@if (starts-with($num, 'var(--')) {
 			@return calc(#{$num} * -1);
 		}


julien (approved):

packages/clay-css/src/scss/functions/_global-functions.scss
```diff
@@ -77,8 +77,7 @@
 @function math-sign($num) {
 	@if (type-of($num) == 'number') {
 		@return -($num);
-	}
-	@elseif (type-of($num) == 'string') {
+	} @else if (type-of($num) == 'string') {
```
matuzalemsteles, line 80: @pat270 the only question regarding what I'm concerned about is if it might fail when the files are built using `node-sass` instead of using `sass` (dart-sass)?

I'm thinking here more in case we ever have to backport a version of Clay with this modification to an older version of DXP that is still using `node-sass`.

matuzalemsteles: Oh perfect, thanks @pat270!

PR #4357, opened 2021-10-19
feat(@clayui/css): Stickers and Cadmin Stickers convert to using Sass maps for more customizability

### LINKED ISSUE #4342: @clayui/css Create clay-sticker-variant mixin for customizing sticker variants
Update sticker component to follow new Sass map pattern, sticker variants can be simplified from:

```scss
$sticker-primary-bg: $primary !default;
$sticker-primary-border-color: null !default;
$sticker-primary-color: color-yiq($sticker-primary-bg) !default;
```

to

```scss
$sticker-primary: () !default;
$sticker-primary: map-deep-merge((
    background-color: $primary,
    color: color-yiq($primary),
), $sticker-primary);
```

pat270:
fixes #4342

Everything should look the same, just the internals were re-arranged to provide more options.

### DIFF
--- a/packages/clay-css/src/scss/components/_stickers.scss
+++ b/packages/clay-css/src/scss/components/_stickers.scss
@@ -65,16 +65,14 @@
 
 // Sticker Sizes
 
-.sticker-sm {
-	@include clay-sticker-variant($sticker-sm);
-}
+@each $selector, $value in $sticker-sizes {
+	%#{$selector} {
+		@include clay-sticker-variant($value);
+	}
 
-.sticker-lg {
-	@include clay-sticker-variant($sticker-lg);
-}
-
-.sticker-xl {
-	@include clay-sticker-variant($sticker-xl);
+	.#{$selector} {
+		@extend %#{$selector} !optional;
+	}
 }
 
 // Sticker Variants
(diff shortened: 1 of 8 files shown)


packages/clay-css/src/scss/components/_stickers.scss
```diff
@@ -47,70 +8,52 @@
...
-
-	bottom: 0;
-	display: flex;
-	justify-content: center;
-	left: 0;
-	overflow: hidden;
-	position: absolute;
-	right: 0;
-	top: 0;
+	@include clay-css($sticker-overlay);
```
matuzalemsteles, line 11: Hey @pat270 it's no big deal but just to understand some of the APIs that we make available and it seems to be a behavior that we do in all components. I see that I have two ways to customize the overlay, using the `$sticker-overlay` variable directly or the `$sticker` variable what's the difference we have here?

I see that the `$sticker-overlay` is not in the scope of the `sticker` class but if I use the variable `$sticker` it is in the scope of the class. Also, I know that with the `$sticker` variable you will be able to modify all the properties besides the overlay here, in that case, it would just be a matter of flexibility to not have to change the whole `$sticker` just to use the overlay?

pat270: `$sticker-overlay` is for the base `.sticker-overlay` component. I couldn't include it all in `$sticker` because of the specificity issue we ran into with alerts. For example:

```scss
$sticker: (
  sticker-overlay: (
    display: none,
  ),
);
```

will output:

```scss
.sticker .sticker-overlay { display: none; }
```

Whereas the base class is just `.sticker-overlay`. We can nest it and everything will work in Clay CSS. I'm just worried about older implementations again.

PR #4358, opened 2021-10-20
feat(@clayui/modal): add the disableAutoClose prop

### LINKED ISSUE #4355: Prevent Modal popup closing on clicking outside area or press the "Esc" key
## What is your proposal?
Is it possible to expose the **_click, kepress/keydown_** handlers so that the default behaviour attached to the **modal** popup can be altered. Or if it is possible to allow props  to stop or prevent the clicking outside area of the modal which result in closing the modal itself.  In some case on the project I am involved in, the requirement is to stop such behaviour. 

It will make the modal popup very flexible to manipulate and interact with as per project requirement.

The alternative is if possible to have a separate **dialogue component** where only the close/cancel button in the dialogue popup can close the dialogue popup and with all the benefits of modal popup.

julien:
The new `disableAutoClose` prop when passed,
will prevent the modal from being closed when either 
the ESC key is pressed or when clicking
outside of the modal's content

Fixes #4355

### DIFF
--- a/packages/clay-modal/src/__tests__/__snapshots__/modal.tsx.snap
+++ b/packages/clay-modal/src/__tests__/__snapshots__/modal.tsx.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ClayModal custom opener does not close automatically with preventAutClose 1`] = `
+exports[`ClayModal custom opener does not close automatically with disableAutoClose 1`] = `
 <body
   class="modal-open"
 >


matuzalemsteles:
@julien yes I know, I've been trying to think of other names but can't think of many other alternatives, maybe `isAutoClose`. What do you think?

julien:
Hey @matuzalemsteles I had another idea, how about `disableAutoClose`?

packages/clay-modal/stories/index.tsx
```diff
@@ -223,6 +223,68 @@ storiesOf('Components|ClayModal', module)
 			</>
 		);
 	})
+	.add('w / preventAutoclose', () => {
```
matuzalemsteles, line 226: Just something you maybe forgot to change

packages/clay-modal/src/__tests__/modal.tsx
```diff
@@ -38,4 +38,46 @@ describe('ClayModal custom opener', () => {
 
 		expect(document.body).toMatchSnapshot();
 	});
+
+	it('does not close automatically with preventAutClose', () => {
```
matuzalemsteles, line 42: The same here!

PR #4388, opened 2021-10-29
fix(@clayui/css): Mixins `clay-label-variant` target [tabindex] inste…

### LINKED ISSUE #4384: @clayui/css: Mixins clay-label-variant should target tabindex instead of tabindex="0"
https://github.com/liferay/clay/blob/1de502fe58ad98c15ed54769812275e6e2ae1695/packages/clay-css/src/scss/mixins/_labels.scss#L634

I assumed we only wanted to target `tabindex="0"` for hover, focus, active, disabled styles. We should target `tabindex` since it can have any value. Negative `tabindex` shouldn't have any hover, focus, active styles since they are usually `disabled`.

pat270:
…ad of [tabindex="0"] so styles still apply when an element is disabled with negative tabindex

fixes #4384

### DIFF
--- a/packages/clay-css/src/scss/mixins/_labels.scss
+++ b/packages/clay-css/src/scss/mixins/_labels.scss
@@ -631,7 +631,7 @@
 
 		&[href],
 		&[type],
-		&[tabindex='0'] {
+		&[tabindex] {
 			@include clay-link($href);
 		}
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM

PR #4397, opened 2021-11-03
fix(@clayui/css): Mixin consolidate card mixins to one `clay-card-variant`

### LINKED ISSUE #4090: @clayui/css Consolidate card mixins to one `clay-card-variant`
We have several card mixins that accomplish similar things. These were made before we had the `clay-css` mixin and had to get each property for each selector. The `clay-css` mixin allows us to pass any CSS property we want, making it less repetitive. We can deprecate and create aliases that map to `clay-card-variant` for:
- clay-card-type-asset
- clay-card-type-asset-variant
- clay-card-type-directory

pat270:
Same as the pr's before. This changes a lot of the internals, but styles should remain the same.

fixes #4090


packages/clay-css/src/scss/mixins/_cards.scss
```diff
@@ -277,146 +412,505 @@
...
 		}
 
 		.card-link {
-			@include clay-link($card-link);
+			@include clay-link(setter(map-get($map, card-link), ()));
+		}
+
+		.card-divider {
+			@include clay-css(setter(map-get($map, card-divider), ()));
+		}
```
matuzalemsteles, line 872: Just one question I'm not sure if I already asked about it, when we don't have the `card-divider` in `$map` the `.card-divider` won't come out in the final build?

matuzalemsteles (approved):
LGTM

packages/clay-css/src/scss/mixins/_cards.scss
pat270: Yeah `clay-css` won't output anything, it would be equivalent of writing:
```scss
.card .card-divider {}
```
Sass doesn't output empty rule-sets.

matuzalemsteles: Oh thanks for clarifying Patrick.

PR #4400, opened 2021-11-04
fix(@clayui/provider): removes dependency from ClayModal to avoid circular dependency issues

### LINKED ISSUE #4399: Remove peerDependency of `@clayui/provider` from `@clayui/shared`
We added the new package from `@clayui/provider` by PR #4162 but to avoid circular dependency issues we need to add the package as a peerDependency of `@clayui/shared` but this has caused some problems when using Clay which forces people to also install `@clayui/provider` explicitly which causes some confusion.

Ideally we want to avoid this and also solve the circular dependency issue that was causing when the package was in `@clayui/shared` dependency.

matuzalemsteles:
Fixes #4399

Unfortunately, this is the only option we have for not using `peerDependencies`. We can for now remove the ClayModalProvider from the ClayProvider and when we can stabilize the package dependencies on `@clayui/core` we can add that support again, it's safe to remove the ClayProvider because it's not used yet.

### DIFF
--- a/packages/clay-provider/docs/provider.mdx
+++ b/packages/clay-provider/docs/provider.mdx
@@ -11,7 +11,6 @@ packageNpm: '@clayui/core'
 -   [Application provider](#application-provider)
     -   [Theme](#theme)
     -   [Icon spritemap](#icon-spritemap)
-    -   [Modal](#modal)
 -   [API](#api)
 
 </div>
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
This is because the packages have decreased in size once we remove the dependence on the Provider's Modal, we have defined a threshold of 5% just to know if this is intentional and to continue controlling. So, we don't need to worry about that in this PR.

bryceosterhaus:
Thanks for update @matuzalemsteles! 🎉 🎉

Issue #4389, opened 2021-10-29
Final adjustments and behavior in TreeView


matuzalemsteles:
Since the TreeView moves into a phase of more performance testing and we have practically all the features implemented, we need to make some adjustments to the component before making it public.

- [x] Add an API to enable and disable DnD 
        (#4396)
- [x] In multi-selection
   - [x] Select nested nodes that are not expanded 
          (#4401)
   - [x] ~If a node that is async is selected, select its children when loading~
- [x] For asynchronous loading of Node, deal with the case of an error or when returning an empty value
       (#4418)
    - [x] For error: we must somehow signal the user or suppress the error
    - [x] For void: this can signal that the Node is not asynchronous, that's because we use a strategy to invoke the method if the Node doesn't have a rendered group, we just deduce that it can be an asynchronous Node.
- [x] Add the option to add Actions to Node
       (#4421)

:warning: Some of these issues are easier to solve, others may take a little longer, we can also create a separate issue if necessary.

julien:
@matuzalemsteles I edited the description to make it more fancy :rofl:

matuzalemsteles:
@julien haha, thanks Julien, it got better.

brunofernandezg:
Test comment to force sync. Ignore

PR #4401, opened 2021-11-04
chore(@clayui/core): select children in collapsed nodes

julien:
Up to now, when a node was collapsed and selected, its
children weren't selected, so in this change we're making
that possible

Related to #4396 and #4389

### DIFF
--- a/packages/clay-core/src/tree-view/useMultipleSelection.tsx
+++ b/packages/clay-core/src/tree-view/useMultipleSelection.tsx
@@ -6,6 +6,10 @@
 import {useInternalState} from '@clayui/shared';
 import {Key, useCallback, useRef} from 'react';
 
+import {ITreeProps, createImmutableTree} from './useTree';
+
+import type {ICollectionProps} from './Collection';
+
 export interface IMultipleSelection {
 	/**
 	 * Handler that is called when the selection changes.
@@ -25,7 +29,10 @@ export interface IMultipleSelectionState {
 	toggleSelection: (key: Key) => void;
 }
 
-export interface IMultipleSelectionProps extends IMultipleSelection {}
+export interface IMultipleSelectionProps<T>
+	extends IMultipleSelection,
+		ITreeProps<T>,
+		Pick<ICollectionProps<T>, 'items'> {}
 
 type LayoutInfo = {
 	children: Set<Key>;
@@ -33,9 +40,13 @@ type LayoutInfo = {
 	parentKey?: Key;
 };
 
-export function useMultipleSelection(
-	props: IMultipleSelectionProps
+export function useMultipleSelection<T>(
+	props: IMultipleSelectionProps<T>
 ): IMultipleSelectionState {
+	const nestedKey = props.nestedKey ?? 'children';
+
+	const tree = createImmutableTree(props.items ?? [], nestedKey);
+
 	const layoutKeys = useRef(new Map<Key, LayoutInfo>());
 
 	const [selectedKeys, setSelectionKeys] = useInternalState<Set<Key>>({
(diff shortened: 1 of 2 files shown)


julien:
@matuzalemsteles in https://github.com/liferay/clay/pull/4401/commits/0b2f601c88e1ac6f0e6c42884c312f82de7cb892 I added the possibility to use single or multiple selection like we spoke about briefly yesterday - The default is multiple selection

packages/clay-core/src/tree-view/TreeView.tsx
```diff
@@ -24,6 +24,7 @@ interface ITreeViewProps<T>
 	expanderIcons?: Icons;
 	onLoadMore?: (item: T) => Promise<unknown>;
 	onRenameItem?: (item: T) => Promise<T>;
+	multipleSelection?: boolean;
```
matuzalemsteles, line 27: @julien I think we can change the name to say that it won't actually disable multi-selection but change the selection type, for example something like:
```
selectionMode: 'multiple' | 'single'
```

But still, I think that could be another name 🤔 I think it's still confusing for the behavior, we have to have an API to say we're not doing recursive selection, maybe `recursiveSelection`.

packages/clay-core/src/tree-view/useMultipleSelection.tsx
```diff
@@ -25,17 +29,28 @@ export interface IMultipleSelectionState {
...
-export function useMultipleSelection(
-	props: IMultipleSelectionProps
+export function useMultipleSelection<T>(
+	props: IMultipleSelectionProps<T>
 ): IMultipleSelectionState {
+	const multipleSelection = props.multipleSelection;
+
+	const nestedKey = props.nestedKey ?? 'children';
+
+	const tree = createImmutableTree(props.items ?? [], nestedKey);
```
matuzalemsteles, line 52: In this case I think we should consider calling this method only where it's being used, because otherwise the `tree` will always be created every time there's a new rendering is a bit unnecessary cost in this case here.

packages/clay-core/src/tree-view/useMultipleSelection.tsx
```diff
@@ -106,12 +189,35 @@ export function useMultipleSelection(
...
+			if (childrenKeyMap) {
+				toggleChildrenSelection(childrenKeyMap, selecteds, select);
+			} else {
+				const parentKeyMatch = key.toString().match(/^(\$\.)(\d)/);
+				let parentKey;
+				if (parentKeyMatch && parentKeyMatch.length === 3) {
+					parentKey = `$.${parentKeyMatch[2]}`;
+				}
+				createPartialLayoutItem(key, parentKey);
+			}
```
matuzalemsteles, line 201: I think this implementation would not be necessary for the parent because since we are clicking on a Node it means that the parent Nodes are also rendered and we have them listed in the `layoutKeys`.

matuzalemsteles (commented):
@julien I just added a comment regarding the API name but I'm still going to look at the rest of the implementation tomorrow, I ended up leaving to review it too late so I'm just taking a quick look here.

packages/clay-core/src/tree-view/useMultipleSelection.tsx
julien: You're right, I actually saw that it wasn't necessary; the only that I didn't like about that is that most of the time we call `createPartialLayoutItem` with two arguments for children (`key` and `parentKey`) so it felt a bit strange calling the same function with only one argument; but since it works and allows us to get rid of that  unnecessary check, I agree that it's better.

packages/clay-core/src/tree-view/useMultipleSelection.tsx
julien: I don't think creating the tree is that expensive, however I agree that we only need it where it's used

packages/clay-core/src/tree-view/TreeView.tsx
julien: I actually like `selectionMode: 'multiple' | 'single'`, and I think that if we used `recursiveSelection` that would imply that it's a `boolean`. We can keep on talking about this.

julien:
@matuzalemsteles thanks for the quick review. I applied suggested changed in https://github.com/liferay/clay/pull/4401/commits/9214f64c169bfedaaec49a11230e6489509e1820, let me know if you see anything else

packages/clay-core/src/tree-view/TreeView.tsx
matuzalemsteles: Yeah, I think if we go with `selectionMode` with the type `multiple` and `single` it might cause a bit of confusion after I thought about it, for example, `single` is more related that it's only allowed to select one node only in every tree.

packages/clay-core/src/tree-view/useMultipleSelection.tsx
matuzalemsteles: Yeah in this case it's not necessarily because we're not doing any operations yet but the problem is more related that this method will create a new reference every time a new rendering happens and internally we just create a partial copy of the first level with destructuring but not much thing too.

matuzalemsteles:
@julien I sent a new commit that follows a slightly different strategy for selecting items recursively when necessary, I added a comment on the commit that has more information, but the main reason I follow another strategy is because:

- The way it was being done would only cover selecting a level instead of selecting recursively.
- The bad part about calling `createPartialLayoutItem` manually in this stream is that we won't know when we need to remove the Node if it doesn't exist anymore or has been moved for example and we will have stale data because the Layout is reactive, I added more comment on this in commit.

The strategy adopted here was to use the items also to know which items need to be selected, the difference is that we do not add the data in the Layout, the Layout informs when the item has lazy children when they are not visible and we use this information to search the remaining Nodes by items, the good thing about using `nodeByPath` is that we're not traversing the tree to get to the node that we don't have more information on and then we just follow the same selection flow that we did before. We also only do this when necessary, so we won't have any performance issues.

An important point is that recursive selection will not work if a Node is asynchronous, let's not make the request and then mark it as selected, this can be a problem for some implementations but that would be too big a problem for us to do it on the runtime side because maybe we would have several requests chained together and wait for the answer to select, that means it would take a long time to make the selection, in these cases what we can advise is to make the recursive selection in the backend.

packages/clay-core/src/tree-view/TreeView.tsx
julien: @matuzalemsteles I think `selectionMode="single"` means that you can select one node at a time. I'm happy with this as it is.

packages/clay-core/src/tree-view/useMultipleSelection.tsx
```diff
@@ -23,22 +24,34 @@ export interface IMultipleSelection {
...
 type LayoutInfo = {
 	children: Set<Key>;
 	intermediate: boolean;
+
+	/**
+	 * Lazy Child means that the current Node has children but they were not
+	 * created because they are not visible in the DOM.
+	 */
+	lazyChild: boolean;
+	loc: Array<number>;
```
julien, line 54: Hey @matuzalemsteles, I'm just reviewing this change in GitHub, and I was wondering, what does `loc` mean here?

matuzalemsteles: Oh sorry @julien, this is actually `indexes`, just short for localization 😅, we can switch to indexes anyway, it's the custom.

julien: Oh I see, if it's `indexes`, then that makes sense. Thanks for clarifying.

packages/clay-core/src/tree-view/TreeView.tsx
matuzalemsteles: Hmm, I always have the intuition that a `selection`, single is related I can just select a Node in the TreeView, for example the `select` tag by default it is a single and you can enable multiple selection of multiple nodes by defining `multiple="multiple"`.

I also find that a single is more intuitive for: I can only select one Node than selecting one at a time, this is also similar to how the user interacts on a form, a Single Selection renders Radios which just means selecting one item within the group and Multiple Selection will render a checkbox which is the multiple selection.

I think the problem here is that this may sound in another sense of what he really wants to do and it can cause some confusion. I'm not saying this is wrong just to try to avoid the API's double meaning.

julien: You might be right, I think if we specify this in the component's documentation it will make it easier for others to understand, but of course we can rename it if you prefer. I'm OK with these last changes.

PR #4405, opened 2021-11-04
fix(@clayui/css): Update map keys that are used in `clay-link` to use newer property names and nesting

### LINKED ISSUE #4402: @clayui/css: Links update map keys to newer pattern
Links Sass maps are still using older keys:
```
btn-focus-box-shadow // should be deprecated
btn-focus-outline // should be deprecated
disabled-active
```
We should update them to the newer pattern:

```
btn-focus-box-shadow // should be deprecated
btn-focus-outline // should be deprecated
disabled: (
    active: (),
),
```

pat270:
fixes #4402


matuzalemsteles (approved):
LGTM

PR #4407, opened 2021-11-05
chore(@clayui/core): add a new performance test example using a very large data

### LINKED ISSUE #4115: Explore the possibility of rendering large tree data
Well, this is an extension for us to be able to create the TreeView component #4111 and it's also part of the collection search #4112 but deeper into the subject of optimally rendering large data in a tree, the TreeView is one of the components that needs it the most, the documentation specification doesn't deal with this issue but we should cover this from the component side, see the ongoing [analysis](https://liferay.atlassian.net/wiki/spaces/ENGFRONTENDINFRA/pages/1586562122/LPS-133328+Scalability+improvements+in+the+TreeView+component) that Daniel is working on from the TreeView.

A crucial part of rendering collections is being able to handle renderings of large amounts of data, depending on the use case, we can work with some strategies like infinite scrolling, virtualization... they are common but for tree, they are more complicated to apply, we can apply load on-demand to subtrees but we will have problems as more items are rendered and a mix of virtualization with load on demand even in memory can be a viable solution but virtualization imposes a lot of rules for this to work as a flat list. So we can study ways how to get this on a tree.

Depending on which solution we follow there is a high possibility of influencing the final component design, so we should be careful to try to align ourselves with #4112.

### Goal

- Rendering strategies
- Solution
- Drawbacks
- Alternatives

matuzalemsteles:
Fixes #4115

This PR only makes a few small changes to stabilize reference to methods that are available through the context but nothing more related to performance improvement, the purpose of this PR is also to bring a TreeView load test analysis and the limitations that exist in which scenarios.

All tests were done using the `performance` example in the Storybook which loads the same XXL data as the Tango team test. I'm just considering the bigger data which is the ultimate goal here.

### [View Rendered Text!](https://github.com/liferay/clay/blob/fc6ea6eda28b2d9b4fa86b4410af687ebe22eab2/docs/treeview_performance.md)

### DIFF
--- a/docs/treeview_performance.md
+++ b/docs/treeview_performance.md
@@ -1,0 +1,136 @@
+# TreeView Performance Analysis
+
+This document covers an analysis of various scenarios of the performance of the `<TreeView />` component in different scenarios and the limitations and more technical details of implementation.
+
+Important information to be considered are the strategies that were adopted to avoid performance losses in the TreeView implementation in different scenarios.
+
+-   Data agnostic
+-   Avoid traverse the tree to DnD, Multiple-selection, and Expandable
+-   Build a tree mirror on a "flat" structure using the React rendering stream
+-   Async Node
+
+### Data agnostic
+
+The main motivation to avoid being rigid by some API of the tree is to avoid developers having to convert their structure to follow the component APIs, this is a component performance degradation point that makes it have to traverse the tree to normalize the data every time it changes on a hot path render.
+
+### Avoid traverse the tree
+
+Avoiding traversing the tree to update some data is always a challenge in different scenarios, in the current implementation in DXP this is often used to update the expandable state and select a Node and create a flat structure to do some operations.
+
+For Expandable and Multi-select solutions instead of adding a unique property as part of each Node, we centralize in one state the unique ids of the Nodes that are expanded and selected respectively to avoid traversing the tree.
+
+```ts
+const [selectedKeys, setSelectionChange] = useState<Set<React.Key>>(
+	new Set(['1', '2', '3'])
+);
+
+const [expandedKeys, setExpandedKeys] = useState<Set<React.Key>>(
+	new Set(['1', '2', '3', '4', '5'])
+);
+```
+
+The DnD implementation would also be a place where we would be using tree traversal to move nodes from one point to another, instead we borrowed the [RFC 6902 (JSON Patch)](https://datatracker.ietf.org/doc/html/rfc6902) implementation to apply the operations to move a Node from a point A to a point B.
+
+```ts
+// Operation of `add` value to the document structure.
+// RFC 6902 (JSON Patch) 4.1
+type PatchAdd = {
+	op: 'add';
+	path: Array<number>;
+	value: unknown;
+};
+
+// Operation of `move` value to the same document structure, removing from
+// `from` and then adding to `path`.
+// RFC 6902 (JSON Patch) 4.4
+type PatchMove = {
+	op: 'move';
+	from: Array<number>;
+	path: Array<number>;
+};
+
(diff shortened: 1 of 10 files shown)


docs/treeview_performance.md
```diff
@@ -0,0 +1,136 @@
...
+
+The API decision to avoid tree traversal to change a property on each Node that would symbolize whether the Node is expanded or not also has the same impact as causing a render on all other Nodes but avoids the cost of traversing the tree.
+
+Since all Nodes consume the same context to know if the Node is expanded any new items added to the state will cause it to re-render.
+
+Ideally, just cause a rendering on the Node that is being expanded, we have some alternatives to avoid that like treating the `expandedKeys` only as initial and using local state on the Node to expand, we can have some drastic performance gains.
+
+This is a common problem for context usage that all-consuming components will cause re-rendering, there is a proposal for this in [React RFC 119 (Context selectors)](https://github.com/reactjs/rfcs/pull/119) that would render the component only if the data it is consuming has changed, a [PR from the React team](https://github.com/facebook/react/pull/20646) was sent some time ago as an experiment but still open.
```
matuzalemsteles, line 120: There are some solutions in userspace that try to implement the idea of Context selectors to not render all components just if the data it's consuming has changed. I'm working on an implementation for this too.

docs/treeview_performance.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Drag and Drop
+
+| Dragging an item | Dropping an item |
+| ---------------- | ---------------- |
+| 5025ms / 3 ops   | 5930ms / 2 ops   |
+
+The implementation of DnD is built on top of the [react-dnd](https://react-dnd.github.io/) library which we ended up depending on the performance of the library but we made our improvements that make the tree change without a high cost, it's down to about a few nanoseconds with the immutability of the tree.
+
+The perception of slowness is totally noticeable to start dragging and then being able to move the Node between the elements is very slow generating animation slowdowns and the drop also has the same effect. The reason is very similar to Select a node and Expand, `react-dnd` causes a lot of rendering in its hooks which are used on all Nodes which makes all components render in one big list takes up a lot of thread time, a list is not phased so it takes a lot of main thread time.
```
matuzalemsteles, line 136: I can't think of a way to smooth DnD for these large data, we rely heavily on react-dnd for that.

matuzalemsteles (commented):
I think to smooth the rendering part the best solution for us here would be to implement virtualization, it seems like a lot has been affected due to a large amount of rendered Nodes because changing something can cause rendering in all components, `react-dnd` it's the most affected by causing a lot of renderings, for multi-selection and expandable, we can smooth it with the idea of Context selectors, I'll send a commit later adding that.

nhpatt:
I think it would be interesting to compare with the performance improvements and tests that Beltrán and Clara did for the older component: https://liferay.atlassian.net/wiki/spaces/ENGTANGO/pages/1890844921/LPS-138166+Frontend+performance+tests+for+the+Treeview+component.

I would try to pass the same tests in the same environment and see if we are improving the performance.

matuzalemsteles:
Hey @nhpatt yeah, I used them as a reference, but yes we are passing the same tests, in most cases getting 4x times as much. I'm using the same data they're using, which is their biggest data. I added a few more scenarios that they didn't cover.

nhpatt:
Nice, good job! :)

docs/treeview_performance.md
matuzalemsteles: Well I used a userspace ([use-context-selector](https://github.com/dai-shi/use-context-selector)) implementation that already implements this idea but I haven't had great performance results, I can see fewer components not being rendered but I can't see a drop in render time.

matuzalemsteles:
Okay I'm going to go ahead with this. Thanks!

PR #4414, opened 2021-11-08
chore(@clayui/alert): adds the indicator and changes the markup for text alignment in the feedback variant

### LINKED ISSUE #4408: Add option to use Indicator Start in ClayAlert
I'm proposing this because right now, when using ClayAlert component with "feedback" variant, it does not have the option to style it with [Alert Indicator Start](https://clayui.com/docs/components/alert/markup.html#css-alert-indicator-start). Adding `alert-indicator-start` class manually makes it work, but it would be great to have a prop for that, or maybe, as suggested by @matuzalemsteles in https://issues.liferay.com/browse/LEXI-1215, adding the class to feedback alerts by default could be enough.

## What is your proposal?
Add the option to style ClayAlert component "feedback" variant with Alert Indicator Start.

We would have the option to use ClayAlert component with that style without adding the CSS class manually

Add `alert-indicator-start` manually

matuzalemsteles:
Fixes #4408

For the text alignment to work well I also needed to change the markup to the feedback variant which requires the Indicator not to be separated from the text due to the CSS implementation.

### DIFF
--- a/packages/clay-alert/src/__tests__/__snapshots__/ClayAlert.tsx.snap
+++ b/packages/clay-alert/src/__tests__/__snapshots__/ClayAlert.tsx.snap
@@ -104,14 +104,14 @@ exports[`ClayAlert renders alert inline with action 1`] = `
 exports[`ClayAlert renders as \`feedback\` variant 1`] = `
 <div>
   <div
-    class="alert alert-feedback alert-info"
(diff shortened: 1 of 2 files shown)


pat270:
LGTM

julien (approved):
LGTM 2

PR #4418, opened 2021-11-09
chore(@clayui/core): suppresses the error from the `onLoadMore` request and returns void does nothing

matuzalemsteles:
See #4389

When a Node is asynchronous, we must handle the request's error, in this case, we only suppress the error but don't show anything to the user, this can be up to whoever is implementing the method, returning an empty value also means that the Node it is not asynchronous or should do nothing.


julien (approved):
LGTM

PR #4421, opened 2021-11-10
chore(@clayui/core): add the feature to add actions to the item in TreeView

matuzalemsteles:
See #4389

Added a new API to `TreeView.Item` so that the developer still has the freedom to add the actions using composition, there's not a very clear way to do this on `children` because `children` is used to building the Node and groups and do this using `children` in a simple way without generating new complexities.

There are some markup issues here though, see example in Storybook of `Actions`, this will work fine with `Button` and `Icon` compositing, but it won't work well using directly eg `DropDown`, because it has a wrapper on the trigger and we need to add the `component-action` to the `Button`. @pat270 is there any way we can get this to work with DropDown's markup, without necessarily being a `Button`?

We also have a limitation for the case of DropDown as an action, the actions are visible and invisible when the mouse is over the Node but the problem is when you try to open DropDown and the trigger will disappear and the Menu will be moved by not find more trigger to reposition the Menu. @pat270 can we do this manually? maybe we have to control the actions show manually via JS to keep the Action visible when the user opens DropDown for example.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -11,6 +11,7 @@ There are two types of documents, repository infrastructure documents, release c
 -   [Developing components](developing_components.md)
 -   [How are clayui.com pages created](creating_pages.md)
 -   [Clay CSS Version History](clay_css_version_history.md)
+-   [TreeView Performance Analysis](treeview_performance.md)
 
 ## RFCs
 
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -91,7 +91,6 @@ export const TreeViewItem = React.forwardRef<
 					}
 					className={classNames('treeview-link', {
 						collapsed: group && expandedKeys.has(item.key),
-						disabled: isDragging,
 						focus,
 						'treeview-dropping-bottom':
 							overTarget && overPosition === 'bottom',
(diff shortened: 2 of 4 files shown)


julien:
@matuzalemsteles I know that this is a draft but I reviewed the code and I like how this is implemented. I also saw the issues you mentioned with the markup, doesn't seem like an easy fix.

matuzalemsteles:
@julien yeah, I'm really worried about the DropDown part, manually controlling the visibility of actions would be the only option for us but it would also be very difficult to manage this visibility because we will have to identify when the user leaves the focus of the Menu 😕.

matuzalemsteles:
Hey @pat270, when you have time, could you help us with that?

matuzalemsteles:
Okay, I didn't realize that Patrick had already added the possibility to handle action focus using classes as well.

I've added the similar implementation we have in ClayList but it's not very efficient but I think it's acceptable since we need to keep focus when opening the DropDown Menu. I've made better support for using `DropDownWithItems` directly without markup issues.

julien:
@matuzalemsteles works like a charm. I'll fix the conflicts, and I might update the "stories" so that there is more than one root item with actions.

julien:
@matuzalemsteles updated the stories and fixed the conflicts (hopefully): this was a rather big conflict, please check if nothing is missing.

matuzalemsteles:
Thanks @julien, just send a commit to remove the `disabled` class that was being added, but otherwise, everything is ok. Thanks for resolving conflicts 🙂

PR #4425, opened 2021-11-15
fix(@clayui/css): Cadmin and Clay Input Group move `.btn` and `.form-control` z-index styles to `input-group-prepend` and `input-group-append`

### LINKED ISSUE #4423: @clayui/css: Chrome Input Group `form-control` inside iframe modal is not visible when focused
https://issues.liferay.com/browse/LPS-142335

This is a strange browser specific bug. Z-index should be ignored if an element is `position: static`. We can work around this with `position: relative`.

input-group

input-group-focus

pat270:
fixes #4423

Input Groups input disappears when focused inside a modal with iframe.

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_input-groups.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_input-groups.scss
@@ -47,6 +47,23 @@
 	> .input-group-text + .btn {
 		margin-left: math-sign($cadmin-input-border-width);
 	}
+
+	.btn {
+		z-index: 1;
+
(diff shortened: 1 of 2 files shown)


julien (approved):
LGTM - no idea why Netlify Jobs fail since they work locally, and there's no good way to debug this

matuzalemsteles:
Netlify build failures don't seem to be related to us, they seem to be Netlify container limits that have been hit.

julien:
thanks for the clarification @matuzalemsteles

PR #4426, opened 2021-11-15
fix(clay-css): 2.x Input Group move `.btn` and `.form-control` z-inde…

pat270:
…x styles to `input-group-prepend` and `input-group-append`. Input Groups input disappears when focused inside a modal with iframe.

#4423


julien (approved):
LGTM

Issue #4424, opened 2021-11-15
@clayui/css: atlas-variables.scss and base-variables.scss should be a Sass partial file


pat270:
https://github.com/liferay/clay/tree/dc5312ed5cdac52662a429624c4628b7ee57eec6/packages/clay-css/src/scss

Prefix the files `atlas-variables.scss` and `base-variables.scss` with a leading underscore so Sass doesn't output the CSS equivalent file.

PR #4428, opened 2021-11-15
fix(clay-css): 2.x `atlas-variables.scss` and `base-variables.scss` shoul…

pat270:
…d be Sass partials. The CSS equivalent shouldn't be created since these files are only meant to be consumed by `.scss` files.

#4424


julien (approved):
LGTM

PR #4432, opened 2021-11-16
feat(@clayui/css): Mixins `clay-css` if parameter is not type map don…

### LINKED ISSUE #4420: @clayui/css: Mixin `clay-css` check if parameter is a map if not don't do anything
We are currently writing `setter(map-get($the-map, the-key-name), ())` every time we pass a parameter to the `clay-css` mixin. This is a workaround for a `map-get` error if the parameter is not type map. We should perform this inside the `clay-css` mixin so we don't repeat it everywhere.

pat270:
…'t do anything

fixes #4420


matuzalemsteles (approved):
LGTM!

PR #4433, opened 2021-11-16
chore(@clayui/core): adds TreeView documentation

matuzalemsteles:
This PR adds the documentation of the TreeView component, still missing add information on how to add Actions, depending on the final implementation #4421.

I'm also exposing the TreeView component in the `@clayui/core` package but under the UNSAFE flag for now.

### DIFF
--- a/packages/clay-core/docs/api-treeview.mdx
+++ b/packages/clay-core/docs/api-treeview.mdx
@@ -1,0 +1,8 @@
+---
+title: 'TreeView'
+description: 'A tree view is a component-based on nodes that are shown in a hierarchical structure.'
+lexiconDefinition: 'https://liferay.design/lexicon/core-components/tree-view/'
+mainTabURL: 'docs/components/treeview.html'
+---
+
+<div>[APITable "clay-core/src/tree-view/TreeView.tsx"]</div>
--- a/packages/clay-core/docs/treeview.mdx
+++ b/packages/clay-core/docs/treeview.mdx
@@ -1,0 +1,378 @@
+---
+title: 'TreeView'
+description: 'A tree view is a component-based on nodes that are shown in a hierarchical structure.'
+packageNpm: '@clayui/core'
+---
+
+import {Example, MultipleSelection} from '$packages/clay-core/docs/treeview';
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
+-   [Example](#example)
+-   [Introduction](#introduction)
+-   [Content](#content)
+    -   [Static](#static)
+    -   [Dynamic](#dynamic)
+        -   [Asynchronous Item](#asynchronous-item)
+    -   [Item](#item)
+        -   [Key](#key)
+        -   [String](#string)
+        -   [Single Level](#single-level)
+        -   [Nested](#nested)
+    -   [Actions](#actions)
+-   [Expander](#expander)
+    -   [Custom expander](#custom-expander)
+-   [Multiple Selection](#multiple-selection)
+    -   [Recursive selection](#recursive-selection)
+-   [Drag and Drop](#drag-and-drop)
+-   [Shortcuts](#shortcuts)
+    -   [Rename Item](#rename-item)
+    -   [Remove Item](#remove-item)
+
+</div>
+</div>
+
+## Example
+
+<Example />
+
+## Introduction
+
+The TreeView component is designed to display data using a hierarchical structure. The component was built with the performance in mind, which influences API design and feature implementations.
+
+This component follows a slightly different thought from our [low-level and high-level philosophy](/docs/get-started/composing.html), we consider this to be a middle-level component that allows you to build the two in one, allowing flexibility and adding features that you wouldn't have in a low-level, it's also easy to move from low-level to high-level quickly and the APIs are more consistent.
+
+## Content
+
+As TreeView is middle-level it allows you to build static or dynamic content if you need to consume data from some service.
+
+It uses a very common technique in React.js components, [Render prop](https://reactjs.org/docs/render-props.html) that whose value of the prop is a function that will receive some data, this is essential for the component to allow it to be data agnostic and we don't need to be rigid as the names the properties to be rendered and avoid creating APIs to change the properties.
+
+```jsx
+<Mouse render={(mouse) => <Cat mouse={mouse} />} />
+```
+
+Another common cause for this is using `children` as a Render prop which allows you to quickly adapt the static component to dynamic just by changing the syntax.
+
+```jsx
+<Mouse>{(mouse) => <Cat mouse={mouse} />}</Mouse>
+```
+
+The main components of the TreeView are `<TreeView.Item />`, `<TreeView.Group />` and `<TreeView.ItemStack />` which abstracts all the necessary markup complexity and allows the granular composition to build different styles of a TreeView, for example, File Explorer, Users, Page Navigation, Page Structure...
+
+### Static
+
+```jsx
+<TreeView>
+	<TreeView.Item>
+		<TreeView.ItemStack>
+			<Sticker displayType="primary" shape="user-icon" size="sm">
+				NS
+			</Sticker>
+			nisi quis eleifend
+		</TreeView.ItemStack>
+		<TreeView.Group>
+			<TreeView.Item key="Victor Valle">
+				<TreeView.ItemStack>
+					<Sticker displayType="primary" shape="user-icon" size="sm">
+						FP
+					</Sticker>
+					fusce ut placerat
+				</TreeView.ItemStack>
+				<TreeView.Group>
+					<TreeView.Item key="susana-vázquez">
+						<Sticker
+							displayType="primary"
+							shape="user-icon"
+							size="sm"
+						>
+							UT
+						</Sticker>
+						ultrices dui sapien
+					</TreeView.Item>
+				</TreeView.Group>
+			</TreeView.Item>
+			<TreeView.Item key="emily-young">
+				<Sticker displayType="primary" shape="user-icon" size="sm">
+					MC
+				</Sticker>
+				maecenas pharetra convallis
+			</TreeView.Item>
+		</TreeView.Group>
+	</TreeView.Item>
+</TreeView>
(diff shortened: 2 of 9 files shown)


packages/clay-core/src/tree-view/TreeView.tsx
```diff
@@ -97,7 +97,7 @@ export function TreeView<T>({
 				ref={rootRef}
 				role="tree"
 			>
-				<DndProvider backend={HTML5Backend}>
+				<DndProvider backend={HTML5Backend} context={window}>
```
julien, line 99: Good catch, I hadn't figured out how to prevent this error

packages/clay-core/src/index.ts
```diff
@@ -19,3 +19,5 @@ export {
 	useModal,
 } from '@clayui/modal';
 export {Provider, useProvider} from '@clayui/provider';
+
+export {TreeView as UNSAFE_TreeView} from './tree-view';
```
julien, line 23: Just a question, why are we exporting this with `UNSAFE_`, it's not that I don't agree, but I thought we've done a pretty good job up to now with this component, so I'd like to understand.

julien:
@matuzalemsteles concerning the documentation, this is a good first step, well done. I'm going to make my change suggestions "online" via GitHub because I don't have access to your fork.

packages/clay-core/docs/api-treeview.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: 'TreeView'
+description: 'A tree view is a component-based on nodes that are shown in a hierarchical structure.'
```
julien, line 3: component-based > component based

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+		</TreeView>
+	);
+}
+```
+
+### Recursive selection
+
+By default when selecting an Item with nested items, the items are recursively selected.
+
+There are some limitations, for static content the recursive selection only works if the item is expanded, ie visible. For dynamic content, it works independently.
```
julien, line 317:
```suggestion
There are some limitations: for static content the recursive selection only works if the item is expanded, (i.e. visible). For dynamic content, it works independently.
```

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+<div class="clay-site-alert alert alert-warning">
+	<strong class="lead">Warning</strong>
+	Recursive selection will not select items from an asynchronous Item.
+</div>
+
+<MultipleSelection />
+
+## Drag and Drop
+
+The Drag and Drop implementation handles items internally and works only for TreeView with dynamic content because it depends on the `items` property and must be enabled using the `dragAndDrop` property.
```
julien, line 328:
```suggestion
The drag and drop implementation handles items internally and works only for TreeView with dynamic content because it depends on the `items` property and must be enabled using the `dragAndDrop` property.
```

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+
+The TreeView handles `items` immutably but follows a partial tree immutability implementation that is more optimized to do at runtime, to access these changes and be able to save in some service, you must control the state using the `items` property and the `onItemsChange` event.
```
julien, line 330:
```suggestion
The TreeView handles `items` immutably but follows a partial tree immutability implementation that is more optimized to do at runtime. To access these changes and be able to save in some service, you must control the state using the `items` property and the `onItemsChange` event.
```

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+		>
+			{...}
+		</TreeView>
+	);
+}
+```
+
+## Shortcuts
+
+The TreeView implements shortcuts and manages the focus, and some shortcuts and focus trigger some shortcuts like renaming, removing, or asynchronous loading if any.
```
julien, line 350: @matuzalemsteles I don't really understand what we mean by
> and some shortcuts and focus trigger some shortcuts like renaming

I would change this to 

The TreeView implements shortcuts and manages the focus. Some shortcuts and focus trigger some actions like renaming, removing or asynchronous loading if any.

But I'd like to know what you think.

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+<div class="clay-site-alert alert alert-warning">
+	<strong class="lead">Warning</strong>
+	If an Item is asynchronous the <code className="gatsby-code-text">
+		onLoadMore
+	</code> method will be called as described in the Asynchronous Item section.
+</div>
+
+### Rename Item
+
+The user can press the shortcut key `R` or `F2` to rename whatever property allows, in that case, feel free to open a Modal or any other experience that allows the user to change something about the Item, for it to work, the `onRenameItem` method must be passed as a property to the component and must be an asynchronous method.
```
julien, line 361:
```suggestion
The user can press the shortcut key `R` or `F2` to rename whatever property. In that case, feel free to open a Modal or any other user interface element that allows the user to change the item. For this to work, the `onRenameItem` method must be passed as a property to the component and must be an asynchronous method.
```

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+		return await openRenameModal(item);
+	}}
+>
+	{...}
+</TreeView>
+```
+
+### Remove Item
+
+If the user presses the `Backspace` or `Delete` hotkey the TreeView removes the Item from the `items` structure, in which case `onItemsChange` is called with the tree updated.
```
julien, line 378:
```suggestion
If the user presses the `Backspace` or `Delete` key, the TreeView removes the item from the `items` structure. In which case `onItemsChange` is called with the tree is updated.
```

julien (commented):
Very good documentation for a first write up! Nice job. 
I made a few suggestions, but feel free to say anything you like or dislike.

matuzalemsteles:
Thanks @julien, I'll look at your suggestions.

> I'm going to make my change suggestions "online" via GitHub because I don't have access to your fork.

Oh yeah, I'll send you an invite to get access to my fork.

packages/clay-core/src/index.ts
matuzalemsteles: Yeah, just adding it as "unstable" to at least have it handy and we can go ahead with the implementations on the DXP side until we're sure nothing's missing, let's work on the next epic of removing TreeView usage from DXP in exchange for Clay, we can remove the flag in the next version anyway.

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+		</TreeView>
+	);
+}
+```
+
+#### Asynchronous Item
+
+Asynchronous Item loading is quite advantageous to decrease the initial payload and decrease memory space, this is more efficient when a tree is very large.
+
+The TreeView doesn't know when an Item is asynchronous so a deduction is made that the developer must deal with on the implementation side to tell whether the Item is asynchronous or not. The `onLoadMore` property is called every time the Item is a leaf of the tree and depending on the method's return it will make some assumptions.
```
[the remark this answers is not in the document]
matuzalemsteles: This got better, I'll still keep the leaf, this normally refers to when node is the last level of the tree because actually, we will only call the method when it is the last level.

packages/clay-core/docs/treeview.mdx
```diff
@@ -0,0 +1,378 @@
...
+>
+	{...}
+</TreeView>
+```
+
+### Item
+
+Item is very flexible and can behave in different ways to simplify usage for different use cases depending on the data.
+
+Item also helps to define the unique `key` of the component across the tree which is also used to tell which items are expanded and selected.
```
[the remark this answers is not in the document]
matuzalemsteles:
```diff
- expanded and selected
+ expanded or selected
```
In that case, I think we should leave `and` because it's used for both and not one or the other, does it make sense? I may be getting it wrong.

packages/clay-core/docs/treeview.mdx
matuzalemsteles: Oh yeah, what I wrote got confused 😅, your suggestion got better to understanding haha

packages/clay-core/src/index.ts
julien: Fair enough, OK ... that said 

> let's work on the next epic of removing TreeView usage from DXP in exchange for Clay, we can remove the flag in the next version anyway.

I'd say: let's help replacing the old TreeView with the new one: if we do it all by ourselves, we're not really helping the team understanding how to use our component. I know we might have to iterate between DXP and Clay to add/fix things that are missing, but I think it's better to involve the team (at least one of their members) on this journey :smile:

packages/clay-core/docs/treeview.mdx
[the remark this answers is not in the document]
julien: No problem. I was thinking that using too many terms: tree, node, leaf might lead to some confusion, if we keep on like that we might end up using "fruit" :joy: But I'm OK with using leaf for now

packages/clay-core/docs/treeview.mdx
julien: Glad you like it

julien (approved):
LGTM

julien:
@matuzalemsteles documentation changes look OK to me.

I had to fix an error with the build, could you check https://github.com/liferay/clay/pull/4433/commits/d4e540894c5ae5de2e87650e56012d05ca7e31de when you get a change? Thanks

packages/clay-core/src/index.ts
matuzalemsteles: haha yeah, I agree.

matuzalemsteles:
> I had to fix an error with the build, could you check d4e5408 when you get a chance? Thanks

@julien oh hadn't noticed, perfect. Thanks

julien:
@matuzalemsteles no worries

julien:
Merging...

Issue #4434, opened 2021-11-17
ClayDropDown.Search navigates when pressing enter in its input


victorg1991:
https://issues.liferay.com/browse/LPS-139049

As explained in the title, the component submits a form when pressing enter which makes a navigation

Here is a codesandbox with and example: https://codesandbox.io/s/hardcore-field-o9zfb?file=/src/index.js

To test that, simply open click the button to open the dropdown, focus the input and press enter

PR #4435, opened 2021-11-17
fix(@clayui/drop-down): Provide a default on submit function to avoid form navigation

victorg1991:
Fixes: https://github.com/liferay/clay/issues/4434

In this PR, a default function is provided for the onSubmit handler in the ClayDropDown.Search component. This prevents de form submission when pressing enter and the input is focused.

This can also be done from the outside passing a custom onSubmit function, but I believe that it is better to control this from the inside since the behaviour is a bit weird.

Let me know what do you think  :)

### DIFF
--- a/packages/clay-drop-down/src/Search.tsx
+++ b/packages/clay-drop-down/src/Search.tsx
@@ -31,17 +31,20 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	value: React.ReactText;
 }
 
+const defaultOnSubmit = (event: React.SyntheticEvent) => event.preventDefault();
+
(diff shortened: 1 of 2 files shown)


julien:
@victorg1991 this change makes sense for me, thanks for the pull request.
I'm still trying to figure out why the "Semantic Pull Request" check wasn't successful because all commits look correctly formatted, so as far as I'm concerned this looks good to me.

victorg1991:
@julien it seemed to be the title of the PR, it is fixed now :)

julien:
@victorg1991 thanks!

matuzalemsteles (approved):
LGTM! It makes sense for me to add this as a default as well.

PR #4440, opened 2021-11-19
 fix(@clayui/css): Atlas `select.form-control` size and multiple styles should match Lexicon

### LINKED ISSUE #4430: I've found that the styles between lexicon and clay in the dual list box are different, is that correct?
## I've found that the styles between lexicon and clay in the dual list box are different, is that correct?

This is the clay snapshot: clay

And the Lexicon snapchot: 
lexicon

pat270:
fixes #4430

The checked unfocused styles have background `#e7e7ed`. Still have to confirm styles for Microsoft Edge browsers.

Chrome (Mac) 96.0.4664.55
chrome-select

Firefox (Mac) 94.0.1
firefox-select

Safari (Mac) 14.1.2
safari-select

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_forms.scss
@@ -371,11 +371,43 @@ $input-select-icon-disabled: clay-icon(
 	$input-select-icon-disabled-color
 ) !default;
 
+// select.form-control[size]
+
+// focus.option.checked: Chrome can't style text, we invert colors to make text white
+// option.checked: hack for changing the background-color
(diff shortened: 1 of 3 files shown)


matuzalemsteles (approved):
LGTM! good solution, I thought we didn't have an option to customize.

pat270:
@matuzalemsteles It's a hack. 

1) Chrome / Edge styles are done by using `filter: invert(100%)`, `background-color: linear-gradient(#ffa700, #ffa700)`. We can't modify text color (always black).
2) FF text is white when checked so we undo the filter and change `background-color: linear-gradient($primary-d2, $primary-d2)`.

matuzalemsteles:
Thanks for clarifying @pat270, this is interesting.

PR #4444, opened 2021-11-23
feat(@clayui/css): Toggle Switch adds option to customize hover, focus, disabled, checked, indeterminate styles

### LINKED ISSUE #4375: @clayui/css: Toggle Switch hover and active states should be customizable
This issue is similar to #4346. We should refactor some of the code to allow more customization through Sass variables.

pat270:
This adds mixins `clay-toggle-switch-bar-variant`, `toggle-switch-check-variant`, and `toggle-switch-variant`. The styles haven't changed, but the option to change them is available.

fixes #4375


marcoscv-work:
It works perfectly! LGTM!

PR #4450, opened 2021-11-24
feat(@clayui/css): Mixins `clay-link` adds customization for `:link` and `:visited`

### LINKED ISSUE #4449: @clayui/css: `clay-link` mixin should allow customizing the `:link` and `:visited` pseudo classes
https://developer.mozilla.org/en-US/docs/Web/CSS/:link
https://developer.mozilla.org/en-US/docs/Web/CSS/:visited

These pseudo classes should be declared before `:hover`, `:focus`, `:active`, etc so those styles will still apply.

pat270:
fix(@clayui/css): Mixins `clay-link` removes setting an empty map if the parameter is null. `clay-css` mixin now checks if the parameter is a map and will no longer error.

docs(@clayui/css): Document link and visited

Regenerates clayui.com Sass API docs

fixes #4449


marcoscv-work:
### LGTM!

PR #4451, opened 2021-11-25
chore: increase Node.js memory usage in Storybook build

matuzalemsteles:
Recently we were seeing the Storybook build failing on Netlify due to lack of memory, I was thinking that the container's memory but doesn't make sense because it looks like the container can use up to 32GB of memory 🙂.

I'm increasing the memory capacity of Node.js to 4 GB, this seems enough because before it was bursting at 2 GB and in some moments finishing the build.


julien (approved):
LGTM

julien:
Thanks @matuzalemsteles

PR #4476, opened 2021-12-02
fix(@clayui/css): Mixin `clay-button-variant` unbundles `.show` from `.active`

### LINKED ISSUE #4475: @clayui/css: `clay-button-variant` mixin split active-class from show and add option to customize ::before and ::after
Currently the `clay-button-variant` mixin bundles `show` class styles with `active` class styles, sometimes designers may want the `show` state to look different from the `active` state (e.g., when dropdown-menu is open).

We should also provide an option to pass in styles to the `::before` and `::after` pseudo elements.

pat270:
- Mixins `clay-link` and `clay-close` Removes the `.show > &` selector
- Mixins `clay-button-variant` adds option to style `::before` and `::after`
- Update `clay-button-variant` documentation

fixes #4475

### DIFF
--- a/packages/clay-css/src/scss/mixins/_buttons.scss
+++ b/packages/clay-css/src/scss/mixins/_buttons.scss
@@ -277,11 +277,6 @@
 		)
 	);
 
-	$active-class: setter(map-get($map, active-class), ());
-	$active-class: map-merge($active, $active-class);
-
-	$active-class-after: setter(map-get($map, active-class-after), ());
-
 	$nested-active-focus: setter(map-get($active, focus), ());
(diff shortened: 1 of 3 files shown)


matuzalemsteles (approved):
LGTM! Good work @pat270, the customization via mixin is getting very interesting, allowing to define everything in the variable map and consistent for all components, it would be interesting later to write a document talking about these principles as a guide.

pat270:
@matuzalemsteles Thanks! Definitely lots of documentation to do. I created https://github.com/liferay/clay/issues/4482 for this.

PR #4478, opened 2021-12-03
fix(@clayui/form): add missing "textarea" type to Input's component prop

### LINKED ISSUE #4477: ClayInput: bug in TypeScript type definition for component prop
## Issue
According to the [docs](https://clayui.com/docs/components/input.html), ClayInput's component prop should accept the string value "textarea". However, it in TypeScript it only accepts "input".

Set up a project with React and Typescript.
Try creating the following element:
```
<ClayInput component="textarea" type="text" value="hello world"  />
```

You'll get the following result:
スクリーンショット 2021-12-03 16 32 55

I should be able to set the value of component to "textarea".
Alternatively, the docs should not provide an example showing that this is possible if it in fact isn't.

## Environment

| Tech           | Version |
| ----------- | ------- |
| clayui/form | 3.40.0  |
| React          | 16.12.0 |
| typescript  | 4.5.2     |

julien:
Fixes #4477

### DIFF
--- a/packages/clay-form/src/Input.tsx
+++ b/packages/clay-form/src/Input.tsx
@@ -157,9 +157,9 @@ ClayInputGroupInsetItem.displayName = 'ClayInputGroupInsetItem';
 
 interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	/**
-	 * Input component to render. Can either be a string like 'input' or a component.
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM! Thanks @julien

PR #4487, opened 2021-12-09
fix(@clayui/form): Hide the symbol.off icon when toggled

### LINKED ISSUE #4486: ClayToggle: the symbol.off icon is visible when toggled

Yes, see [here](https://codesandbox.io/s/cranky-antonelli-18sze)

Use a `ClayToggle` component with the `symbol` prop, and you'll notice that it is rendered like this

It doesn't seem normal to be seeing both icons.

The `symbol.off` icon shouldn't be visible when the toggle is "toggled" (that's a mouthful)

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | latest (2021/12/9) |
| React |   same as above      |

julien:
Fixes #4486

### DIFF
--- a/packages/clay-form/src/Toggle.tsx
+++ b/packages/clay-form/src/Toggle.tsx
@@ -79,31 +79,33 @@ const ClayToggle = React.forwardRef<HTMLLabelElement, IToggleProps>(
 						<span className="toggle-switch-handle">
 							{symbol && (
 								<>
-									<span
-										className={classNames(
-											'button-icon',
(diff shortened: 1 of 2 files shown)


pat270:
@julien we might not need to merge this one. https://github.com/liferay/clay/pull/4485 should fix it.

julien:
@pat270 looking at the React component's code I wasn't sure if this was only about CSS, and I was still seeing this on our `master` branch with #4485, until I checked #4489 so it's all good

PR #4488, opened 2021-12-09
fix(@clayui/css): Mixins `clay-navbar-variant` should be able to output selectors that `clay-navbar-size` can

### LINKED ISSUE #4481: @clayui/css: Update `clay-navbar-variant` to `clay-css` mixin pattern
The mixin [`clay-navbar-variant`](https://github.com/liferay/clay/blob/97e6f321929abb5a630f01eb056ef51beffe8a4d/packages/clay-css/src/scss/mixins/_navbar.scss#L924) is deprecated. It is time to bring it back since the underlying Sass infrastructure better supports customizations.

pat270:
This provides complete customization of all the Navbar spin off components. There are some breaking changes in this pr.

- The mixin `clay-navbar-variant` output `border-color: transparent` and `border-style: solid` by default. This has been removed to reduce unnecessary CSS. You will need to declare them in your Sass map if you relied upon it.
- We previously output `navbar-expand-{sm|md|lg|xl}` for the variants of Application Bar, Management Bar, Navigation Bar, Sidebar Navigation Bar. We only output those that are needed namely, `navbar-expand-md`. If you relied upon other breakpoints you will need to declare them in the respective Sass map.

These changes reduced ~17kb from Atlas. For the future I'd like to move away from using `clay-navbar-size` since it outputs a lot of unused styles. I'm predicting that converting to `clay-navbar-variant` would further reduce Atlas ~43kb.

fixes #4481

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_application-bar.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_application-bar.scss
@@ -9,11 +9,19 @@ $application-bar-size: map-deep-merge(
 $application-bar-base: () !default;
 $application-bar-base: map-deep-merge(
 	(
-		link-border-radius: $border-radius,
-		link-outline: 0,
-		link-transition: box-shadow 0.15s ease-in-out,
-		link-focus-box-shadow: $btn-focus-box-shadow,
-		link-disabled-box-shadow: none,
+		navbar-nav: (
+			nav-link: (
+				border-radius: $border-radius,
+				outline: 0,
+				transition: box-shadow 0.15s ease-in-out,
+				focus: (
+					box-shadow: $btn-focus-box-shadow,
+				),
+				disabled: (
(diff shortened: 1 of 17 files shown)


matuzalemsteles (approved):
Great work here @pat270! Well, I don't see any major problems with not rendering the different sizes of navbars, but it would be good to check if there are use cases for this in DXP before going with that. I'm not sure if Translator Manager is using it.

But good work on consistency between different navbars, this is a really nice DX improvement. File size reduction is also always welcome.

pat270:
@matuzalemsteles Thanks! I searched and we only use `navbar-expand-md`. There are a few cases where we are using `navbar-expand`, but that isn't related to this pr.

matuzalemsteles:
Perfect @pat270, thanks for checking. So let's go ahead with this.

PR #4489, opened 2021-12-09
fix(@clayui/css): Mixin `clay-toggle-switch-bar-variant` adds `.button…

### LINKED ISSUE #4486: ClayToggle: the symbol.off icon is visible when toggled

pat270:
…-icon-off` selector

fixes #4486

@julien This should fix the regression I introduced in #4444.


julien (approved):

julien:
Hey @pat270, yes this does fix #4444, thanks

PR #4500, opened 2021-12-14
fix(clay-link): listens for the `handleClick` event and removes the preventDefault call by default

matuzalemsteles:
Calling the `preventDefault` method every time on Click will change the behavior for all components and we don't want that because we will be breaking all existing behaviors.

I'm also changing the demo to listen for the `handleClick` event, it is called internally due to the explicit listener added in the component but its name is changed due to the ClayComponent implementation.


julien (approved):
LGTM

julien:
@matuzalemsteles thanks!

PR #4508, opened 2021-12-15
feat(@clayui/css): Dropdown adds `dropdown-menu-width-shrink`

### LINKED ISSUE #4501: @clayui/css: Dropdown menu lower min-width to match Lexicon's specs
https://liferay.design/lexicon/core-components/dropdowns/

Lexicon dropdown-menu's show dropdown-menu with short text should be narrower than 240px. We will need to add `white-space: nowrap` to `dropdown-item`. If we allow breaking to new line by default, the menu always defaults to 120px width for relative dropdown-menu's that have a toggle that's narrow (vertical ellipsis) and `lfr-icon-menu` dropdowns.

We will need to add a utility class for `dropdown-menu` to allow breaking text to a new line for items with long text. This would be considered a breaking change imo.

pat270:
fixes #4501

Dropdown Menus in portal are too complicated. There is a huge mix of new and old. Adding a modifier class is the safest way to go, we can update specific places as we see fit.

menu-shrink


matuzalemsteles (approved):
LGTM! @pat270 I added a commit to add the modifier class support to the component as well.

PR #4519, opened 2021-12-20
fix(@clayui/css): Cadmin convert a few places where we are still using slash as division

### LINKED ISSUE #4493: @clayui/css: Cadmin we are still using slash as division operator in a couple places
clay-css/src/scss/cadmin/variables/_alerts.scss 285:52
clay-css/src/scss/cadmin/components/_utilities-functional-important.scss 250:5

convert it to calc or `* 0.5` instead of `/ 2`.

pat270:
fixes #4493 

This uses `calc()` to fix the deprecation warnings for using / as division in Cadmin. This also updates clayui.com to `@use 'sass:math' as *;`. The deprecation warnings shouldn't show when running clayui.com.

### DIFF
--- a/clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
+++ b/clayui.com/plugins/gatsby-plugin-clay-css-tasks/gatsby-node.js
@@ -132,17 +132,20 @@ const generateCSSFiles = (pluginOptions) => {
 	const clayuiSrcDir = path.join(pluginOptions.clayuiSrc, 'styles');
 
 	const atlas = compileSass({
-		file: path.join(scssDir, 'atlas.scss'),
+		file: path.join(clayuiSrcDir, 'dart-sass-atlas.scss'),
+		includePaths: [scssDir],
 		outFile: path.join(cssDir, 'atlas.css'),
 	});
 
 	const base = compileSass({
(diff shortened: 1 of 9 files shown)


julien:
Hey @pat270 not really sure what I'm looking at :rofl: 

When building clay-css I still see

```
Recommendation: math.div($numerator, $denominator)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
13 │         $numerator / $denominator

```
(and a few I left out)


But when running `yarn develop` in the clayui.com directory I don't see any of these warnings.

pat270:
@julien lol no worries. The deprecation warning should still be visible when building clay-css. Dart Sass deprecated using `/` for division and replaced it with a function `div(numerator, denominator)`.  There's no backward compatible way to make `/` work in Libsass and Dart Sass without writing our own division function from scratch.

In Dart Sass, we need to globally include the math library with `@use 'sass:math' as *;` for `div()` to work. The solution was to `@use` wherever we are using Dart Sass (e.g., clayui.com and 7.4) and leave it alone where we are using Libsass (e.g., 7.3).

julien:
@pat270 thanks for the explanation - so this seems good to go

julien (approved):

PR #4551, opened 2022-01-06
fix(@clayui/css): Mixins `clay-button-variant` removes unnecessary `s…

### LINKED ISSUE #4550: @clayui/css: Mixins `clay-button-variant` remove unnecessary setter()
The `setter()` function was used to set `null` parameters to empty maps so the Sass function `map-get` wouldn't throw the error:

```
Error: argument `$map` of `map-get($map, $key)` must be a map
```

We now check to see if the parameter is of type `map` in the `clay-css` mixin and no longer need to do this.

pat270:
…etter()`'s

fixes #4550

### DIFF
--- a/packages/clay-css/src/scss/mixins/_buttons.scss
+++ b/packages/clay-css/src/scss/mixins/_buttons.scss
@@ -635,25 +635,19 @@
 				}
 
 				.inline-item {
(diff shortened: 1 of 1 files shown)


marcoscv-work:
LGTM

PR #4566, opened 2022-01-12
chore: update instructions on how to test local changes in dxp

julien:
I found that using "yarn add" is what works in most cases
(at least on my machine, which is running linux)

This is not a "super important" fix, but it would be nice to have
several people test this before merging.


julien:
Closing, still doesn't work.
With some packages like management-toolbar it does work, but with others (like button) it doesn't.
So I don't know what to do about this, but IMO we should focus on this

PR #4579, opened 2022-01-14
feat(@clayui/popover): close the popover when the ESC key is pressed

### LINKED ISSUE #4567: ClayPopover: the ESC key should close the popover
We recently added a new prop to the popover component, in order
to close the popover when clicking outside of it.

@drakonux suggested that hitting the ESC key should also close
the popover, as this would improve the usability and accessibility of 
the component.

julien:
Fixes #4567

### DIFF
--- a/packages/clay-popover/src/index.tsx
+++ b/packages/clay-popover/src/index.tsx
@@ -187,6 +187,20 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 			}
 		}, [closeOnClickOutside, trigger]);
 
+		useEffect(() => {
+			const handleKeyDown = (event: KeyboardEvent) => {
+				if (event.key === 'Escape') {
+					setShow(false);
+				}
+			};
+
+			window.addEventListener('keydown', handleKeyDown);
+
+			return () => {
+				window.removeEventListener('keydown', handleKeyDown);
+			};
+		}, []);
+
 		let content = (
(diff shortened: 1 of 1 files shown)


packages/clay-popover/src/index.tsx
```diff
@@ -187,6 +187,20 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 			}
 		}, [closeOnClickOutside, trigger]);
 
+		useEffect(() => {
+			const handleKeyDown = (event: KeyboardEvent) => {
+				if (event.key === 'Escape') {
```
diegonvs, line 192: Hey @julien, when reading https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key, I notice we should use `Esc` for representing `Escape` key to make it work with IE/Edge too. 

Nevermind, [it normalizes the key](https://github.com/facebook/react/blob/cae635054e17a6f107a39d328649137b83f25972/packages/react-dom/src/events/SyntheticEvent.js#L300)

julien: :+1: and additionally Edge is Chrome with some fancy UI stuff on top. As for IE, I personally think it's time we stop supporting it - we're in 2022 after all.

diegonvs: > we're in 2022 after all.

future

PR #4590, opened 2022-01-20
feat(@clayui/core): allow passing extra css classes to expander button

### LINKED ISSUE #4589: TreeView: Give options to customize expanders and nodes
## What is your proposal?
It would be cool if the component could get some props like `nodeClassName`, `expanderClassName` or something like that, so we can customize them without overwriting the treeview classes. 
For the expanders, currently the component gets an `expanderIcons` prop in which you can pass a `<ClayIcon />` with a `className`, but in the case of the + button, it's not getting the classes properly (maybe this is a bug).
For the nodes, you define the markup of it, but it would be cool if we could add some class to the `<li class="treeview-item" />`.

We could adapt the style of nodes and expanders without overwriting the treeview classes in our CSS, making our styles independent in case in some moment the treeview classes change.

The alternative is using the treeview classes in our CSS.

julien:
Fixes #4589

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -324,6 +324,7 @@ export function TreeViewItemStack({
 	const {
 		expandOnCheck,
 		expandedKeys,
+		expanderButtonClasses,
 		expanderIcons,
 		nestedKey,
 		open,
@@ -340,6 +341,12 @@ export function TreeViewItemStack({
 
 	const hasChildren = nestedChildren && nestedChildren.length > 0;
 
+	const cssClasses = ['component-expander'];
+
+	Array.isArray(expanderButtonClasses)
+		? expanderButtonClasses.forEach((c) => cssClasses.push(c))
+		: cssClasses.push(expanderButtonClasses!);
+
 	return (
 		<Layout.ContentRow {...otherProps}>
 			{expandable && (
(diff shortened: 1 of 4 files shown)


packages/clay-core/src/tree-view/TreeViewItem.tsx
```diff
@@ -340,14 +341,20 @@ export function TreeViewItemStack({
...
+
 	return (
 		<Layout.ContentRow {...otherProps}>
 			{expandable && (
 				<Layout.ContentCol>
 					<Button
 						aria-controls={`${item.key}`}
 						aria-expanded={expandedKeys.has(item.key)}
-						className={classNames('component-expander', {
+						className={classNames(cssClasses, {
```
matuzalemsteles, line 357: You could have just added it that way wouldn't be a problem either.

```
classNames('component-expander', expanderButtonClasses, {
```

The property doesn't have to be an array, because it has the same effect as `'class other-class'`. It would have the same effect we do for the other components, we could also call it `expanderClassName` to follow the naming of other components.

julien: > You could have just added it that way wouldn't be a problem either.

Sorry I don't understand. 

> The property doesn't have to be an array, because it has the same effect as 'class other-class'. 

Yes, I know, if you want to revert this feel free, the thing is, I needed to get this done and merged - and you were on PTO. 

>  we could also call it expanderClassName to follow the naming of other components.

That's why I did in the first iteration, but I think it's better to be precise, these classes are going to be applied to the "Button" that wraps the "expander"  so `expanderButtonClasses` doesn't seem to be that "wrong" - also don't get me wrong but there is no much "consistency" with prop names in the whole repository.  I'll re submit a change since no one in DXP is using this yet.

PR #4594, opened 2022-01-21
feat(@clayui/css): Ports over Treeview to Clay from Cadmin

### LINKED ISSUE #4588: TreeView: Styles are only in cadmin
I was wondering why the `treeview` related styles are only in 
the [`cadmin`](https://github.com/liferay/clay/blob/bc208f3afc08cd55f7eb2344938851feac762abb/packages/clay-css/src/scss/cadmin/components/_treeview.scss) "theme" (I don't even know if this is a theme).

I'm asking because when integrating the `TreeView` in DXP, a lot of "applications" don't use 
`cadmin` and that causes the `TreeView` to look "broken", for example:

pat270:
fixes #4588

- removed unnecessary `setter()`'s from code
- added docs for `.treeview-item.disabled`, `.treeview-dropping`, `.treeview-dragging-*`
- removed cadmin from examples in the docs

### DIFF
--- a/clayui.com/content/docs/components/markup-treeview.md
+++ b/clayui.com/content/docs/components/markup-treeview.md
@@ -9,11 +9,18 @@ mainTabURL: 'docs/components/treeview.html'
 <div class="nav-toc">
 
 -   [Treeview](#css-treeview)
+
     -   [Group](#css-treeview-group)
     -   [Item](#css-treeview-item)
+    -   [Item](#css-treeview-item-disabled)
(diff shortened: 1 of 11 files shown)


julien (approved):

julien:
Seems we always forget to include  _lx-icons-generated.scss, so I pushed a commit [here](https://github.com/liferay/clay/commit/5ec165d8b2d2f8cbedcd3cb8e201648dcbfc4d04)

PR #4596, opened 2022-01-24
feat(@clayui/core): add onItemClick prop to allow clicking on nodes

### LINKED ISSUE #4568: TreeView: add a onClick prop when clicking on nodes
TreeView has a "onSelectionChange" prop, but that only works
when the component uses a checkbox.

In DXP though, some Trees don't have checkboxes but are selectable, 
for example:

Currently, clicking on one of these nodes marks it as selected.
So we need to think about that an implement it.

julien:
Fixes #4568

@matuzalemsteles if you need more context on why this is needed, please give me a shout on slack. Thanks!

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -46,6 +46,12 @@ interface ITreeViewProps<T>
 	expanderIcons?: Icons;
 
 	/**
+	 * Optional callback which will be invoked if present
+	 * when clicking on a node.
+	 */
+	onItemClick?: (item: T) => void;
+
+	/**
 	 * When a tree is very large, loading items (nodes) asynchronously is preferred to
 	 * decrease the initial payload and memory space. The callback is called every time
 	 * the item is a leaf node of the tree.
@@ -88,6 +94,7 @@ export function TreeView<T>({
 	items,
 	nestedKey = 'children',
 	onExpandedChange,
+	onItemClick,
 	onItemsChange,
 	onLoadMore,
 	onRenameItem,
@@ -123,6 +130,7 @@ export function TreeView<T>({
 		expanderButtonClasses,
 		expanderIcons,
 		nestedKey,
+		onItemClick,
 		onLoadMore,
 		onRenameItem,
 		rootRef,
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -50,6 +50,7 @@ export const TreeViewItem = React.forwardRef<
 		expandedKeys,
 		insert,
 		nestedKey,
+		onItemClick,
 		onLoadMore,
 		onRenameItem,
 		open,
@@ -253,7 +254,26 @@ export const TreeViewItem = React.forwardRef<
 						{typeof left === 'string' ? (
 							<Layout.ContentRow>
 								<Layout.ContentCol expand>
-									<div className="component-text">{left}</div>
+									<div
+										className="component-text"
+										onClick={(event) => {
+											event.stopPropagation();
+
+											const {
+												_indexes,
+												_itemRef,
+												_key,
+												_parentItemRef,
+												...rest
+											} = item;
+
+											if (onItemClick) {
+												onItemClick(rest);
+											}
+										}}
+									>
+										{left} blandndnd
+									</div>
 								</Layout.ContentCol>
 
 								{actions && <Actions>{actions}</Actions>}
@@ -327,6 +347,7 @@ export function TreeViewItemStack({
 		expanderButtonClasses,
 		expanderIcons,
 		nestedKey,
+		onItemClick,
 		open,
 		selection,
 		toggle,
@@ -383,7 +404,31 @@ export function TreeViewItemStack({
 				}
 
 				if (typeof child === 'string') {
-					content = <div className="component-text">{child}</div>;
+					content = React.cloneElement(
+						<div className="component-text">{child}</div>,
+						{
+							onClick: (
+								event: React.MouseEvent<
+									HTMLDivElement,
+									MouseEvent
+								>
+							) => {
+								event.stopPropagation();
+
+								const {
+									_indexes,
+									_itemRef,
+									_key,
+									_parentItemRef,
+									...rest
+								} = item;
+
+								if (onItemClick) {
+									onItemClick(rest);
+								}
+							},
+						}
+					);
 
 					// @ts-ignore
 				} else if (child?.type.displayName === 'ClayIcon') {
--- a/packages/clay-core/src/tree-view/context.ts
+++ b/packages/clay-core/src/tree-view/context.ts
@@ -20,6 +20,7 @@ export interface ITreeViewContext<T> extends ITreeState<T> {
 	expanderButtonClasses?: Array<string> | string;
 	expanderIcons?: Icons;
 	nestedKey?: string;
+	onItemClick?: (item: T) => void;
 	onLoadMore?: (item: any) => Promise<unknown>;
 	onRenameItem?: (item: T) => Promise<any>;
 	rootRef: React.RefObject<HTMLUListElement>;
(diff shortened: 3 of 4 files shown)


matuzalemsteles (commented):
I added some comments to the issue but I think we need to think about it, especially if dealing with multiple selection we won't be able to solve it easily using just `onItemClick`.

I'm seeing the Lexicon team added a new spec for these DXP cases https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1970078399/LEXI-1257+Clarify+behavior+for+DXP+Trees, so maybe we don't need to implement using the same current look in DXP.

There is also the question of whether we should add this for them or should the team deal with it now and just create some demos to demonstrate that we support what the Lexicon proposes.

I think one case we haven't addressed is single selection, I think we can support that built-in, without needing to add a new prop, we can also expose a selection method by rendering props to support the case of having a button to trigger selection like in the proposal.

julien:
@matuzalemsteles yes let's talk about this - I personally think we could have seen that this was needed very easily by analyzing correctly the current usages.

> There is also the question of whether we should add this for them or should the team deal with it now and just create some demos to demonstrate that we support what the Lexicon proposes.

I invite you to migrate an existing component in DXP to realize that  we have to take care of that and not ask the teams to do it.

matuzalemsteles:
@julien this is to be expected that some things will not work mainly that we are developing in a new look. But we didn't analyze all cases, we just used Daniel's document as a reference to understand the performance problem, as discussed in other TreeView development PRs, the idea is to build the component according to the Lexicon and then test how it adapts as for the DXP cases, for sure the use cases that had AlloyUI we will have problems replicating the experience, was another, so maybe we have to do things like this or just support the call of internal methods more, as, in the case of the multi-selection that is abstracted when it has a checkbox, we can pass the methods via render props here too to more flexible to trigger selection, this is another option.

> I invite you to migrate an existing component in DXP to realize that we have to take care of that and not ask the teams to do it.

Yes of course, as I had suggested for us to do this... In this particular case that I'm commenting on, the visual of Category, Page, and Filtering by Type that has this behavior [has a new look that is closer to the Lexicon spec](https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1970078399/LEXI-1257+Clarify+behavior+for+DXP+Trees), just trying to understand if we should also apply this new look, maybe we don't need to implement manually a lot of selection or offer something here.

julien:
@matuzalemsteles, I somewhat agree with what you said, but I have a different point of view concerning some details

> this is to be expected that some things will not work mainly that we are developing in a new look

It might be expected from our side, but teams will not accept this - as I as saying in yesterday's meeting, when you migrate an existing component the goal is to keep any existing functionality, the look and feel can change but not the overall experience - we are there to help teams not to give them more work, unless there is an agreement, and in this case there isn't.

>  But we didn't analyze all cases, we just used Daniel's document as a reference to understand the performance problem

Maybe you didn't, but I did - and mostly with @drakonux: who actually mentioned several times during the implementation period that "click to select" should work - not all "treeviews" have checkboxes, actually most of them don't.

> the idea is to build the component according to the Lexicon and then test how it adapts as for the DXP cases

That's exactly what I'm trying to solve here.

> for sure the use cases that had AlloyUI we will have problems replicating the experience, was another, so maybe we have to do things like this or just support the call of internal methods more, as, in the case of the multi-selection that is abstracted when it has a checkbox, we can pass the methods via render props here too to more flexible to trigger selection, this is another option.

Concerning AlloyUI replacements, I don't see why we'd have a problem since our component supports the same features, concerning the rest, I'm sorry but I really don't understand what you mean.

> I invite you to migrate an existing component in DXP to realize that we have to take care of that and not ask the teams to do it.

> Yes of course, as I had suggested for us to do this... In this particular case that I'm commenting on, the visual of Category, Page, and Filtering by Type that has this behavior has a new look that is closer to the Lexicon spec, just trying to understand if we should also apply this new look, maybe we don't need to implement manually a lot of selection or offer something here.

Concerning that comment, what I wanted to insist on, is that when you migrate an existing component, you quickly realize a few things:

- Teams want you to migrate the component, with **all** the existing functionality (trust me I've done that more than once for several components)

- Testing local changes is near impossible, which is a clear sign that having Clay outside DXP is making us waste lots of time: a component library shouldn't be that hard to use or maintain.

matuzalemsteles:
> It might be expected from our side, but teams will not accept this - as I as saying in yesterday's meeting, when you migrate an existing component the goal is to keep any existing functionality, the look and feel can change but not the overall experience - we are there to help teams not to give them more work, unless there is an agreement, and in this case there isn't.

Yes I agree, we are not going to get in the way, we really have to add all the functions but I think we have to have an alignment in a certain way with the teams, I say more in terms of visuals than functionality.

> Maybe you didn't, but I did - and mostly with @drakonux: who actually mentioned several times during the implementation period that "click to select" should work - not all "treeviews" have checkboxes, actually most of them don't.

Oh Great, I don't remember seeing comments to support this in PRs, really I may have missed something. Looking at the Lexicon documentation do we actually have the selected state, I probably missed or confused it with multiple selection state.

I'm a little confused by that now, @drakonux does this select state mean something like a single selection? ie is it similar to the multiple selection state? in case the user selects the node to do something action, for example in a tree the user can select other nodes or should there always be only one node selected?

Anyway, we can also implement this quickly, using the same multi-selection APIs. I'll create a PR to demonstrate the usage, let me know what you think.

> Teams want you to migrate the component, with all the existing functionality (trust me I've done that more than once for several components)

I'm not disagreeing with you just to remind you, I've done that too.

> Testing local changes is near impossible, which is a clear sign that having Clay outside DXP is making us waste lots of time: a component library shouldn't be that hard to use or maintain.

Well, maybe your flow is different from what we already did in Clay, we don't test Clay using the local link, we always use the published package, if something fails or doesn't work right, we reproduce it in Clay, that's a lot to try to separate if it's a glitch within DXP that is impacting this or is it really in Clay, it's also more comfortable to reproduce the same issue outside of DXP, so we don't waste as much time on that.

Speaking again, I understand that this is being a big pain for you and we should fix it but I don't believe this is real for everyone, especially me and @pat270 who has also been developing on Clay for over 4 years. We've had Clay for a long time and this wasn't at least a problem for us during the development flow of all components. And again, I'm not ignoring this.

julien:
@matuzalemsteles thanks for the feedback, I guess we'll have a call about this, but here are some quick notes

> Oh Great, I don't remember seeing comments to support this in PRs, really I may have missed something. Looking at the Lexicon documentation do we actually have the selected state, I probably missed or confused it with multiple selection state.

> I'm a little confused by that now, @drakonux does this select state mean something like a single selection? ie is it similar to the multiple selection state? in case the user selects the node to do something action, for example in a tree the user can select other nodes or should there always be only one node selected?

> Anyway, we can also implement this quickly, using the same multi-selection APIs. I'll create a PR to demonstrate the usage, let me know what you think.

To make it more simple, we need to be able to select nodes the same way we do with the "checkbox", even when there is no checkbox in the TreeView, and you are right, this could be either "single" or "multiple" selection
I'll check your pull request, because I was thinking we could reuse our "multi-selection" api, we would just need some "markup" changes.

> Well, maybe your flow is different from what we already did in Clay, we don't test Clay using the local link, we always use the published package, if something fails or doesn't work right, we reproduce it in Clay, that's a lot to try to separate if it's a glitch within DXP that is impacting this or is it really in Clay, it's also more comfortable to reproduce the same issue outside of DXP, so we don't waste as much time on that.

Maybe my workflow is different, but I think my it's the one that most developers experience in D.X.P.

Not being able to test local changes means having to cut a release to test in D.X.P. just to realize it "doesn't work". Which is going to take much longer that being able to test the change locally before cutting the release, so when you say that "we don't waste as much time", I disagree and think "we are wasting way too much time".

For someone not familiar with Clay reading this, it almost seems that we are implementing things with no guarantee that they will work, and for me that's a mistake. Regardless of what "we already did in Clay", there are some times when things need to change, and for me this needs to happen **now**.

> Speaking again, I understand that this is being a big pain for you and we should fix it but I don't believe this is real for everyone, especially me and @pat270 who has also been developing on Clay for over 4 years. We've had Clay for a long time and this wasn't at least a problem for us during the development flow of all components. And again, I'm not ignoring this.

It's not a pain just for "me", it's a pain for everyone - or at least any one who tries to use Clay in DXP - which includes you as well, unless you're just focusing on the Clay repository. Let's stop pretending this is easy to use.

PR #4597, opened 2022-01-24
chore: increases storybook memory usage and split files more into chunks

matuzalemsteles:
This is just a test PR, apparently, Storybook is consuming too much memory and maxing out, not sure if it's related to the Node.js limit set before 4GB now to 8GB because the process is being killed with `exit code 137`, so it is more related to the operating system in this case Linux, maybe the container in our plan has a limit of 4GB or less, I will look more at the documentations.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -42,6 +42,14 @@ module.exports = {
 			})
 		);
 
+		config.optimization = {
+			splitChunks: {
+				chunks: 'all',
+				minSize: 30 * 1024,
+				maxSize: 1024 * 1024,
+			},
+		};
+
 		return config;
 	},
 };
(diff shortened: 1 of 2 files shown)


julien:
@matuzalemsteles that's a good start.
Now we should document the whole thing (who has access, how it works). I had another question, do you know why that Netlify "job" needs to install go and python - since this is just a JS project? Seems like a waste of bandwidth

matuzalemsteles:
> I had another question, do you know why that Netlify "job" needs to install go and python - since this is just a JS project? Seems like a waste of bandwidth

Good question, I'm not sure why it's downloading. I have to look at the documentation to see if I can configure this, we don't really configure anything in Netlify, we just plug in and ask to run the command.

I found this https://docs.netlify.com/configure-builds/manage-dependencies/ but it doesn't say much, it seems like it's auto-detected, not sure, I have to read the docs further.

julien (approved):

PR #4599, opened 2022-01-25
fix(@clayui/css): Treeview dropping an item above or below causes items to shift

### LINKED ISSUE #4592: @clayui/css: Treeview dropping an item above or below causes items to shift
https://clayui.com/docs/components/treeview.html

Dragging and dropping an item above or below makes the content below it shift depending on the size of the border. We should use `box-shadow` or a pseudo element to create the indicator instead of border.

pat270:
fix(@clayui/css): Treeview use `box-shadow` instead of border to indicate drop zone
    
fix(@clayui/css): Treeview use `clay-link` mixin for `treeview-dropping-*` it gives more styling options like hover, focus, active, disabled and pseudo elements

fixes #4592

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_treeview.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_treeview.scss
@@ -140,19 +140,19 @@
 	@include clay-link(map-get($cadmin-treeview, link));
 
 	&.treeview-dropping-bottom {
-		@include clay-css(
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
LGTM! Thanks @pat270 for this, I tried to solve it using `border`, but using box-shadow is a great strategy for that.

PR #4600, opened 2022-01-25
fix(@clayui/css): Rename `.treeview-item.disabled` to `treeview-item-dragging`

### LINKED ISSUE #4593: @clayui/css: Treeview rename `.treeview-item.disabled` to `treeview-item-dragging`
~We can still click on links, check form elements, and collapse/expand items inside `.treeview-item.disabled`.~

We should reserve the `disabled` class for disabling pointer-events on links and buttons. `treeview-item-dragging` will highlight the node that is being dragged.

pat270:
- Removes dropping styles and style `treeview-link` directly

@matuzalemsteles I wasn't able to make the disable links when dragging happen since it requires pointer events on the item being dragged. This just frees up the `disabled` class for future use. I think it will come in handy.

fixes #4593

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -87,7 +87,7 @@ export const TreeViewItem = React.forwardRef<
 			<li
 				{...otherProps}
 				className={classNames('treeview-item', className, {
-					disabled: isDragging,
+					'treeview-item-dragging': isDragging,
 				})}
 				role="none"
 			>
--- a/packages/clay-css/src/scss/cadmin/components/_treeview.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_treeview.scss
@@ -120,20 +120,6 @@
 			map-deep-get($cadmin-treeview, treeview-item, last-child)
 		);
 	}
-
-	&.disabled {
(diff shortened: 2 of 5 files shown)


matuzalemsteles (commented):
Hey @pat270 it looks like the Drag and Drop implementation broke with these changes, maybe it's because the `treeview-dropping` was removed here?

pat270:
@matuzalemsteles It's `pointer-events:none` that's preventing the browser from firing mousedown/click/etc. I've been thinking, Clay follows the pattern where we add the `disabled` class to links and buttons and it prevents these events from passing through. The way treeview `disabled` works, breaks this pattern where it highlights the item(s) being dragged. Is it possible to rename it to something like `treeview-item-dragging`? That way we can reserve the `disabled` class for disabling a link the traditional way. 

This would also save some extra logic we would have to write if one day someone wants a specific item in the treeview  to be undraggable. We could just add the class `disabled` to `treeview-link`.

matuzalemsteles:
@pat270 I see, makes sense to me and we really shouldn't break this pattern of disabled, we can go with `treeview-item-dragging`.

pat270:
Thanks @matuzalemsteles I'll go ahead and make the changes.

matuzalemsteles (approved):
Great @pat270, it's working fine. Thanks

PR #4603, opened 2022-01-27
feat(@clayui/core): adds support to trigger selection manually and change the behavior of single selection to select only one Node in the whole TreeView

### LINKED ISSUE #4602: Add single selection state to TreeView and set to default
The feature to select only one Node in the entire TreeView is still missing, this probably got confusing and we ended up losing the tracking of this feature.

We currently have the `selectionMode` API which was with `multiple` and `single` but `single` has become more confusing now in differentiating this state of selecting only one Node in the whole TreeView compared to `single` which has the same effect than `multiple` but not recursively selecting nodes.

We're going to have to change the API to lessen the confusion of states because within the selection all these features are within that same selection group and are not composited together.

We can change the behavior of `selectionMode`:

- `single` allows selecting only one Node in the entire TreeView, if another Node is selected, the previous one is deselected. The single here is much more intuitive with this state because it is closely related to the intuition that we take in our daily lives, of selecting only one product among several products or as we refer to the radio tag as a single and checkbox as the closest multiple example.
- `multiple-recursive` allows you to select multiple nodes and recursively
- `multiple` allows selecting multiple nodes in a TreeView but not recursively.

matuzalemsteles:
Fixes #4602

This PR first implements the proposal to trigger the selection manually using render props like in this example https://github.com/liferay/clay/issues/4568#issuecomment-1022375634, also see the demo in Storybook. The idea here is to use render props and expose a safe API that has a certain barrier in the internal methods so that the selection feature can be used manually without having to add the `<Checkbox />` explicitly in Node.

For the manual selection behavior, I also made some intuitive changes to the `onClick` methods so they can respect the design intuition, see my comment in the commit message which I describe with more information the reason to add it this way https://github.com/liferay/clay/commit/8e03f43d8ad030b6da7695d332d581a2650cb62f.

Also adds the single selection behavior, as I commented in the issue I am changing the behaviors to take the confusion out of the `selectionMode` API and adding the single selection as defined by default according to the [Lexicon spec](https://docs.google.com/document/d/1fggZBRGAIkGPbnkRrGEUOT0hHpkF-YjFmREee0XCoI0/edit#heading=h.e2tswp5i0wxo).

For cases where the TreeView is in a Modal and you select a Node and you need to close it, instead of using a single selection just add an `onClick` method on `TreeView.ItemStack` or `TreeView.Item` to know when you clicked on the node and if you need to prevent the default behavior of expanding, You can use `event.preventDefault()`, something like this:

```jsx
<TreeView
  selectionMode={null}
>
  {(item) => (
    <TreeView.Item>
      <TreeView.ItemStack onClick={() => closeModal(item)}>
        <Icon symbol="folder" />
        {item.name}
      </TreeView.ItemStack>
      <TreeView.Group items={item.children}>
        {(item) => (
          <TreeView.Item onClick={() => closeModal(item)}>
            <Icon symbol="folder" />
            {item.name}
          </TreeView.Item>
        )}
      </TreeView.Group>
    </TreeView.Item>
  )}
</TreeView>
```

I'm still in doubt with the single selection of how the unselected state would work, right now, clicking again on the selected Node will uncheck it. @drakonux could you look at [this example](https://deploy-preview-4603--next-storybook-clayui.netlify.app/?path=/story/components-claytreeview--single-selection) and check if it should be like this?

### ToDo

- [ ] Update the documentation

### DIFF
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -3,11 +3,20 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import React from 'react';
+import React, {Key, useCallback} from 'react';
 
 import {ItemContextProvider, useItem} from './useItem';
+import {useTreeViewContext} from './context';
 
-export type ChildrenFunction<T> = (item: T) => React.ReactElement;
+export type Selection = {
+	toggle: (key: Key) => void;
+	has: (key: Key) => boolean;
+};
+
+export type ChildrenFunction<T> = (
+	item: T,
+	selection: Selection
+) => React.ReactElement;
 
 export interface ICollectionProps<T> {
 	children: React.ReactNode | ChildrenFunction<T>;
@@ -37,13 +46,24 @@ export function Collection<T extends Record<any, any>>({
 	children,
 	items,
 }: ICollectionProps<T>) {
+	const {selection} = useTreeViewContext();
 	const {key: parentKey} = useItem();
+
+	const hasKey = useCallback(
+		(key: Key) => {
+			return selection.selectedKeys.has(String(key));
+		},
+		[selection.selectedKeys]
+	);
 
 	return (
 		<>
 			{typeof children === 'function' && items
 				? items.map((item, index) => {
-						const child = children(item);
+						const child = children(item, {
+							has: hasKey,
+							toggle: selection.toggleSelection,
+						});
 
 						const key = getKey(
 							index,
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -14,7 +14,7 @@ export type Selection = {
 };
 
 export type ChildrenFunction<T> = (
-	item: T,
+	item: Omit<T, 'indexes' | 'itemRef' | 'key' | 'parentItemRef'>,
 	selection: Selection
 ) => React.ReactElement;
 
@@ -42,6 +42,13 @@ export function getKey(
 	return parentKey ? `${parentKey}.${index}` : `$.${index}`;
 }
 
+export function removeItemInternalProps<T extends Record<any, any>>(props: T) {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	const {indexes, itemRef, key, parentItemRef, ...item} = props;
+
+	return item;
+}
+
 export function Collection<T extends Record<any, any>>({
 	children,
 	items,
@@ -60,7 +67,7 @@ export function Collection<T extends Record<any, any>>({
 		<>
 			{typeof children === 'function' && items
 				? items.map((item, index) => {
-						const child = children(item, {
+						const child = children(removeItemInternalProps(item), {
 							has: hasKey,
 							toggle: selection.toggleSelection,
 						});
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -5,8 +5,8 @@
 
 import React, {Key, useCallback} from 'react';
 
-import {ItemContextProvider, useItem} from './useItem';
 import {useTreeViewContext} from './context';
+import {ItemContextProvider, useItem} from './useItem';
 
 export type Selection = {
 	toggle: (key: Key) => void;
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -51,7 +51,7 @@ export function Collection<T extends Record<any, any>>({
 
 	const hasKey = useCallback(
 		(key: Key) => {
-			return selection.selectedKeys.has(String(key));
+			return selection.selectedKeys.has(key);
 		},
 		[selection.selectedKeys]
 	);
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -59,10 +59,11 @@ interface ITreeViewProps<T>
 
 	/**
 	 * Flag changes the Node selection behavior when a checkbox is rendered on the Node.
-	 * - multiple: selects child nodes recursively.
-	 * - single: select only node, ignoring children.
+	 * - single: select only node.
+	 * - multiple: select multiple nodes.
+	 * - multiple-recursive: selects multiple nodes and recursively.
 	 */
-	selectionMode?: 'multiple' | 'single';
+	selectionMode?: 'single' | 'multiple' | 'multiple-recursive' | null;
 
 	/**
 	 * Flag to indicate if the TreeView will show the expander in the hover in the Node.
@@ -93,7 +94,7 @@ export function TreeView<T>({
 	onRenameItem,
 	onSelectionChange,
 	selectedKeys,
-	selectionMode = 'multiple',
+	selectionMode = 'single',
 	showExpanderOnHover = true,
 	...otherProps
 }: ITreeViewProps<T>) {
@@ -126,6 +127,7 @@ export function TreeView<T>({
 		onLoadMore,
 		onRenameItem,
 		rootRef,
+		selectionMode,
 		showExpanderOnHover,
 		...state,
 	};
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -419,7 +419,7 @@ export function TreeViewItemStack({
 					return null;
 				}
 
-				if (typeof child === 'string') {
+				if (typeof child === 'string' || typeof child === 'number') {
 					content = <div className="component-text">{child}</div>;
 
 					// @ts-ignore
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -76,7 +76,7 @@ export const TreeViewItem = React.forwardRef<
 
 	const hasKey = useCallback(
 		(key: Key) => {
-			return selection.selectedKeys.has(String(key));
+			return selection.selectedKeys.has(key);
 		},
 		[selection.selectedKeys]
 	);
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -11,6 +11,7 @@ import {Keys} from '@clayui/shared';
 import classNames from 'classnames';
 import React, {Key, useCallback, useContext, useState} from 'react';
 
+import {removeItemInternalProps} from './Collection';
 import {Icons, useTreeViewContext} from './context';
 import {useItem} from './useItem';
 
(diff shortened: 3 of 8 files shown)


julien:
@matuzalemsteles I think this is going in the right direction, good workd! 

There are still some little details we need to work on. 

When the selection changes after clicking on a node that has children; in the storybook example, this is what I see



- I wanted to ask you if we couldn't "return" the node themselves instead of their "key", because that's what developers work with (i.e. instead of returning `12`, return `{name: 'Documents and Media' , ...}`). 
- Another thing I think could be improved is that when you select a node that has children, you are selecting that node and **not** it's children so why are we returning the "keys" of this children? (This is something we can talk on a call because it's probably easier and faster to understand).
- Maybe I haven't seen it yet, but having the same storybook example with "single" selection would help.

Apart from that I'm very happy with the result

matuzalemsteles:
Hey @julien I added an example to demonstrate the use of manual selection with a single selection, I also made some more changes related to ownership of the properties of the `TreeView.Item` component, similar to the `onClick` change, this is more intuitive in the way of using the `TreeView.Item` component depending on how it is declared.

In the example I also made a small case when you need the entire item data and not just the key, this seems to work well, in this specific case I keep it in the state but in cases where selecting closes a modal for example, it would not be necessary because it could just call the modal's close method passing the given item directly instead of saving in the state or doing the manual selection. This could also be done for the multiple selections but it wouldn't work very well I think, I haven't tested it.

julien:
@matuzalemsteles this is exactly what we needed, I also saw that you took care of what we discussed yesterday - - thanks for the update. I'm going to test how this works in DXP by publishing locally using [verdaccio](https://verdaccio.org/) (as far as I'm concerned, there is no real need to automate this yet), if everything is OK I'll merge this.

julien:
@matuzalemsteles I checked this is DXP and the only issue is that when "toggle" is called, there is a slight delay between the moment when the node is "expanded" (probably due to the animation, at least this is something I had already seen in my first attempt to implement this "feature") .

What I mean is that if you use something like in the example (keep in mind that this is "pseudo code" used to explain the error - please pay attention to the comments in that code), then the TreeView's items (the root ones with "children"), are never expanded, and  after a while the page just leaks memory (at least that's my first guess because everything is unresponsive) and eventually you'll have to close your browser tab

```jsx

const [selectedKeys, setSelectionChange] = useState(new Set());
const [selected, setSelected] = useState({});

const handleItemClick = (item, selection) => {
	if (!selection.has(item.id)) {
		setSelected(item);
	}

	selection.toggle(item.id);

	// Let's say here you want to dispatch an event to 
	// notify an other part of the application (not in React)
	// which item is not selected, for example
	Liferay.Util.getOpener().Liferay.fire(itemSelectorSaveEvent, {
		data: {
			folderId: item.id,
			folderName: item.name,
		},
	});
	// The code just above won't be executed, and the "node" 
	// will say "collapsed"
};


<TreeView
	items={items}
	nestedKey="children"
	onSelectionChange={(keys) => setSelectionChange(keys)}
	selectedKeys={selectedKeys}
	selectionMode="single"
	showExpanderOnHover={false}
>
	{(item, selection) => (
		<TreeView.Item>
			<TreeView.ItemStack 
				onClick={() => handleItemClick(item, selection) } 
				style={{
					backgroundColor: selection.has(item.id) ? 'aliceblue' : ''
				}}>
				<Icon symbol="folder" />

				{item.name}
			</TreeView.ItemStack>

			<TreeView.Group items={item.children}>
				{(item) => (
					<TreeView.Item 
						onClick={() => handleItemClick(item, selection) } 
						style={{
							backgroundColor: selection.has(item.id) ? 'aliceblue' : ''
						}}>
						<Icon symbol="folder" />

						{item.name}
					</TreeView.Item>
				)}
			</TreeView.Group>
		</TreeView.Item>
	)}
</TreeView>
```

I know that in DXP, not everything is using React so you get a mix of React + Liferay specific stuff, but we aren't going to be able to 
- Refactor the entire application
- Ask teams to do that

If you want to understand exactly what I mean, you'll need to replace the previous `TreeView` component located [here](https://github.com/liferay/liferay-portal/blob/e606d943dc44b3263d9720cf007be550b5619712/modules/apps/journal/journal-web/src/main/resources/META-INF/resources/js/SelectFolder.es.js#L81-L86), create a few Web Content Articles, and a Folder, and try to move one of the Web Content Articles into a Folder. 


I've also added a .gif, which might help

matuzalemsteles:
@julien thanks a lot for this information, really this case is very strange to happen, I will test it today in DXP to try to understand what is happening and I will try to reproduce outside of DXP too, thanks for testing.

matuzalemsteles:
Well, I think I figured out the problem here, after spending all afternoon investigating 😄, these are my conclusions about it.

Well, as we are calling the `Liferay.Util.getOpener().Liferay.fire(itemSelectorSaveEvent)` method it will close the Modal and mark the selected item but as the `handleItemClick` method has some asynchronous operations this is causing it to crash everything, for example calling `setSelected` which schedules an asynchronous state update, `selection.toggle` which also has the same effect at the end which updates the state of selected keys. This is causing it to crash, probably due to the implementation of React itself or something related to the React scheduler that is in a constant loop that cannot be executed after it loses all context, this is just a guess I didn't dig deep to find out what It's happening behind the scenes.

So as selecting, in this case, will close the modal, we don't need to save the selected item or do the toggle selection, we also don't need in this case that the node expander is done because it also causes an asynchronous schedule, in this case, we can call `event.preventDefault()` to take out the default expand-on-select behavior.

This is the example I used to test this in Web Content:

```jsx
const handleItemClick = (event, item) => {
  event.preventDefault();

  Liferay.Util.getOpener().Liferay.fire(itemSelectorSaveEvent, {
    data: {
      folderId: item.id,
      folderName: item.name,
    },
  });
};

<TreeView
  items={items}
  showExpanderOnHover={false}
>
  {(item) => (
    <TreeView.Item>
      <TreeView.ItemStack 
        onClick={(event) => handleSelectionChange(event, item)}
        style={{
          backgroundColor: selection.has(item.id) ? 'aliceblue' : ''
        }}
      >
        <Icon symbol="folder" />

        {item.name}
      </TreeView.ItemStack>

      <TreeView.Group items={item.children}>
        {(item) => (
          <TreeView.Item 
            onClick={(event) => handleSelectionChange(event, item)} 
            style={{
              backgroundColor: selection.has(item.id) ? 'aliceblue' : ''
            }}
          >
            <Icon symbol="folder" />

            {item.name}
          </TreeView.Item>
        )}
      </TreeView.Group>
    </TreeView.Item>
  )}
</TreeView>
```

I also noticed that when clicking on our expander it is propagating the event upwards, this is calling `onClick` from the node which is also preventing the user from expanding without selecting, I added a new commit to correct this.

julien:
Hey @matuzalemsteles, I've tried these changes in DXP but I'm running into the same problem - the modal window stays open and after a while the tab crashes. It was also a good opportunity to test "local changes in Clay" using [verdaccio](https://github.com/verdaccio/verdaccio), and I don't think it's the best option. Not because of the tool itself which works well, but for some reason in DXP it only works the first time when installing a new "local" package. I ended having to publish 5 different versions (locally) to make sure I was seeing your changes; but, on the positive side, at least I know this is not going to improve our workflow.

julien:
@matuzalemsteles I was able to solve this issue by doing something like this in the "journal-web" module:

```js
	const handleItemClick = (event, item) => {
		event.preventDefault();

		setTimeout(() => {
			console.log('closing the modal');
			Liferay.Util.getOpener().Liferay.fire(itemSelectorSaveEvent, {
				data: {
					folderId: item.id,
					folderName: item.name,
				},
			});
		}, 1000);
	};
```

I know it's really bad, so I think we could probably trigger the "onClick" once the asynchronous operations end in the TreeView. I'll keep on looking at this just so see if there's an easy way.

matuzalemsteles:
>I've tried these changes in DXP but I'm running into the same problem - the modal window stays open and after a while the tab crashes.

Hmm, I'll try this again today.

> I was able to solve this issue by doing something like this in the "journal-web" module:

Hey @julien did you manage to get it to run without the timeout? just calling the method and `event.preventDefault();`? I was able to do this on Friday with the changes of this PR but I'm not sure because at the end of the day my mind was no longer thinking straight 😅.

Edit: I think you already answered that 😬

> Not because of the tool itself which works well, but for some reason in DXP it only works the first time when installing a new "local" package. I ended having to publish 5 different versions (locally) to make sure I was seeing your changes; but, on the positive side, at least I know this is not going to improve our workflow.

Yeah, I tested it on Friday and it's really not the best option, I'm wondering how we could simplify this flow a bit without using yarn link or a local registry, really having to publish the package every time despite being fast but really, it's not an ideal flow but it can help us for now.

matuzalemsteles:
I did some testing today but can't reproduce the problem when I'm using these changes of the PR, not sure if I missed something, let me know if you need anything else.

matuzalemsteles:
> I spent the afternoon debugging this and I'm 100% sure that I'm still seeing the same behavior as 3 days ago

Hmm, I'll try other strategies today again, maybe clean all my workspace and published packages and do it all over again. Anything, we can call tomorrow and see what might be going on.

matuzalemsteles:
Well, really I'm not able to reproduce this locally, I redid all my space with verdaccio eliminating the published packages, cleaned the clay repository of the builds, and ran the build again on top of this PR, I published a version above the current one with verdaccio 3.45.1 and added in journal-web, using the same example, I simplified it a bit to not use manual selection in this case because I was only using it to change the background.

```jsx
const handleSelectionChange = (event, item) => {
  event.preventDefault();

  Liferay.Util.getOpener().Liferay.fire(itemSelectorSaveEvent, {
    data: {
      folderId: item.id,
      folderName: item.name,
    },
  });
};

<TreeView
  items={nodes}
  showExpanderOnHover={false}
>
  {(item) => (
    <TreeView.Item>
      <TreeView.ItemStack 
        onClick={(event) => handleSelectionChange(event, item)}
      >
        <Icon symbol="folder" />

        {item.name}
      </TreeView.ItemStack>

      <TreeView.Group items={item.children}>
        {(item) => (
          <TreeView.Item 
            onClick={(event) => handleSelectionChange(event, item)}
          >
            <Icon symbol="folder" />

            {item.name}
          </TreeView.Item>
        )}
      </TreeView.Group>
    </TreeView.Item>
  )}
</TreeView>
```

I recorded a short video showing the final result.

https://user-images.githubusercontent.com/13750819/152072382-0edf69d7-948f-464a-accb-65e958784474.mp4

julien:
@matuzalemsteles thanks for the video, - I'll check this again on a "clean" branch based on `master`, I might have some changes producing the "strange" behavior.

julien:
Hi again @matuzalemsteles, I just created a new branch based on `master`, tested these changes and I'm still seeing the same issues (impossible to select a node without making the tab crash) - Let's talk about this over a call if you can.

julien:
@matuzalemsteles since we need this and I already reviewed this, I'm going to merge this. I feel kind of "dumb", not having seen this yesterday since you already cut a release. I guess I'll cut another one tomorrow, just to be able to make progress on the treeview migration.

matuzalemsteles:
> In my case, in my two first failed attempts from this morning I was using 3.44.1 which is already used in DXP and for some reason, it wasn't installing the package from my local registry but from "cache".

This makes sense, I thought I faced something similar for publishing the same version without change, good to know.

julien:
@matuzalemsteles yes, I guess that I might have to update our contributing document just to make it clear that sometimes this happens - will send an update during the day.

PR #4609, opened 2022-01-31
chore: update instructions on how to test local changes

julien:


matuzalemsteles (approved):
LGTM! @julien this is great, I tested it on Friday and it really works, maybe we can improve the experience a little more to not publish every time I make a change but I think it wouldn't be a big problem at least testing for a few hours, I also used the CLI created by Ivan https://github.com/liferay/liferay-frontend-projects/tree/master/support/packages/local-npm helps a little with this registry change.

julien:
@matuzalemsteles I personally don't see how that improves the experience: using `verdaccio` is straight forward. I had a closer look at "local-npm" and the only thing it "automates" is that it "un publishes" packages before publishing them to make sure you can always publish the same version of a package.

Which can easily be done with 
npm unpublish --force --registry http://localhost:4873
npm publish --registry http://localhost:4873

PR #4611, opened 2022-01-31
fix(@clayui/date-picker): fix time picker default value and error when not formatting time in 12hours when passing to time picker

### LINKED ISSUE #4608: @clayui/date-picker: Time is not being correctly shown on the Date Picker input
Hello guys, hope you are doing well!

I am from Bravo Team, responsible for Forms and Data Engine components, and right now we are working on a story that creates a new field called "Date and Time" that involves the Date Picker with the time option enabled. For more details you can see the story ticket [here](https://issues.liferay.com/browse/LPS-143050).

That being said, we are facing some issues that we believe that are clay related so I will explain it to you: 

1. When the user types an input for a language that works with time in AM/PM like english, the Date Picker modal isn't showing the correct value when the time has a PM value. For example, if I type `02/03/2022 09:00 PM` on the field input, when I select the date picker it will show the time value as `21:00 PM`. If the user clicks on a date, the input will be sent as `21:00 PM` which is a problem because this hour doesn't exist and the input will be invalid. A video of this behaviour can be seen [here](https://drive.google.com/file/d/1jbOLppgvrzFSqkJXs-uJA69IMoIzcQZz/view?usp=sharing).

2. When we open the modal for the first time, the the ideal behaviour of the default value for time would be like `--:--` or `--:-- --` as the same of this [Time Picker](https://clayui.com/docs/components/time-picker.html), but right now it shows `00:00` or `00:00 --` making the behaviour a little inconsistent since there isn't a time like `00:00` for AM/PM languages. A video of this behaviour can be seen [here](https://drive.google.com/file/d/15gCFfmVpck9iW22zAULd3fUjQ-D4BV9x/view?usp=sharing).

Please let us know if this is something we could fix on our side, if we are missing something, or if you need more info about this matter. Thanks in advance!

You can test both scenarios here: https://codesandbox.io/s/cranky-platform-47ldv?file=/src/index.js

_For the first scenario (PM case):_
1.  **Type** or **paste** on the field input a date and time like `02/03/2022 09:00 PM`
2. Click on the date picker icon to open the modal
3. Observe the value that is on the time input

_For the second scenario (00:00 case):_
1. Click on the date picker icon to open the modal
2. Observe the value that is on the time input
3. Change **only** the last part of the time, turning `--` into AM or PM
4. Select any date on the date picker
5. Close the modal and observe the value on the field input

_For the first scenario (PM case):_
The time input on the date picker modal should show the 09:00 PM value instead of 21:00 PM.

_For the second scenario (00:00 case):_
The default value for the time on the date picker should be `--:-- --` for AM/PM languages and `--:--` for the others, to avoid sending invalid values to the field input like `00:00 AM` for example.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.43.0  |
| React | v17.0.2 |

matuzalemsteles:
Fixes #4608

Just small fixes related to Date Picker with Time Picker. Some changes are more related to the Time Picker default value which was always starting with the value.

### DIFF
--- a/packages/clay-date-picker/src/Hooks.ts
+++ b/packages/clay-date-picker/src/Hooks.ts
@@ -60,7 +60,7 @@ export const useWeeks = (
 /**
  * Sets the current time
  */
-export const useCurrentTime = (format: string) => {
+export const useCurrentTime = () => {
 	const [currentTime, set] = useState<string>('--:--');
 
 	const setCurrentTime = useCallback(
--- a/packages/clay-date-picker/src/Hooks.ts
+++ b/packages/clay-date-picker/src/Hooks.ts
@@ -61,9 +61,7 @@ export const useWeeks = (
  * Sets the current time
  */
 export const useCurrentTime = (format: string) => {
-	const [currentTime, set] = useState<string>(() =>
-		formatDate(setDate(new Date(), {hours: 0, minutes: 0}), format)
-	);
+	const [currentTime, set] = useState<string>('--:--');
 
 	const setCurrentTime = useCallback(
 		(
--- a/packages/clay-date-picker/src/TimePicker.tsx
+++ b/packages/clay-date-picker/src/TimePicker.tsx
@@ -6,6 +6,8 @@
 import ClayTimePicker, {Input} from '@clayui/time-picker';
 import React from 'react';
 
+import {formatDate, setDate} from './Helpers';
+
 interface IProps {
 	currentTime: string;
 	disabled?: boolean;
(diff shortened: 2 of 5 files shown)


julien (approved):
LGTM. @matuzalemsteles when looking at the issue I was wondering why we aren't using [`Date.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse).

matuzalemsteles:
@julien for this fix exclusively it doesn't cover the case we want, because we want it to format it to AM/PM time and we also don't have a complete date to use `Date.parse` in this case. For the other cases where we parse different formats that can be customized depending on the language and also the format with time, in this case `Date.parse` just accepts an ISO date.

PR #4612, opened 2022-02-01
feat: add plain/common API for button links

carloslancha:
Adding a clearer API. Discussed in https://liferay.slack.com/archives/C3JBR21HA/p1643723253270239


julien (approved):
LGTM, makes sense to add these props since most all of them are being used when the link is not a button. I think this simplifies the use of the component.

matuzalemsteles (approved):
LGTM! it's a good improvement, thanks @carloslancha.

PR #4616, opened 2022-02-01
feat(@clayui/css): Utilities adds `font-size` utilities `fs-*`

pat270:
Quoting Emiliano (can't find gh user name):

> I just want to add that even if programmatically it would make sense to add more steps from both sizes, from the design point of view I wouldn’t recommend going under 10px in any case. Thus, there shouldn’t be a way to scale it downwards, at least from our side.

> I also think a common scenario for this would be a request to extend between steps (like adding one number between XL and XXL) and that’s difficult to provide, but not impossible.

> I think T-Shirt sizes are a good idea if we adapt it for the web, and for this I would recommend aligning the text sizes with the measures the same way we did with colors and is by adding steps, here’s my idea:

> 16px is the base so it should be an unique Medium size (fs-M).
then we can decrease size: 14px (fs-S), 12px (fs-S2), 10px (fs-S3)
we can also Increase size: 18px (fs-L), 20px (fs-L1), 24px (fs-L3)

> AND we can provide extra measures included in the 11 sizes for bigger content: 28px (fs-XL1), 32px (fs-XL2), 36px (fs-XL3), 40px (fs-XL4) (this is actually the only one with 4 text sizes, we can adapt it to be 3 since this scaling system has not been used yet)

> So, in the end each measure would have 3 associated text sizes, it can scale easily and could also extendable by adding more numbers such as fs-L4 or fs-XL5. :smile:
I think we can start with something similar, keep an eye on how people use it, request feedback and iterate.

I modeled the utility classes after this suggestion. Does this provide the space to expand font sizing anywhere along the scale? Is it confusing? If it's good I can send a proper pr for this.

/cc @liferay-lexicon @ethib137 @claraizquierdo @matuzalemsteles @julien

### DIFF
--- a/clayui.com/content/docs/css/utilities/markup-text.md
+++ b/clayui.com/content/docs/css/utilities/markup-text.md
@@ -1,0 +1,211 @@
+---
+title: 'Text'
+lexiconDefinition: 'https://liferay.design/lexicon/foundations/typography/'
+---
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
+-   [Sizes](#css-text-sizes)
+-   [Weights](#css-text-weights)
+-   [Styles](#css-text-styles)
+
+</div>
+</div>
+
+## Sizes(#css-text-sizes)
+
+Utility classes for changing the `font-size` of text.
+
+<div class="sheet-example">
+	<table class="table table-autofit">
+		<thead>
+			<th class="table-cell-ws-nowrap">Utility</th>
+			<th>Rem</th>
+			<th>Px</th>
+			<th class="table-cell-expand">Example</th>
+		</thead>
+		<tbody>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-11</td>
+				<td>2.5rem</td>
+				<td>40px</td>
+				<td class="table-cell-expand"><span class="text-11">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-10</td>
+				<td>2.25rem</td>
+				<td>36px</td>
+				<td class="table-cell-expand"><span class="text-10">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-9</td>
+				<td>2rem</td>
+				<td>32px</td>
+				<td class="table-cell-expand"><span class="text-9">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-8</td>
+				<td>1.75rem</td>
+				<td>28px</td>
+				<td class="table-cell-expand"><span class="text-8">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-7</td>
+				<td>1.5rem</td>
+				<td>24px</td>
+				<td class="table-cell-expand"><span class="text-7">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-6</td>
+				<td>1.25rem</td>
+				<td>20px</td>
+				<td class="table-cell-expand"><span class="text-6">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-5</td>
+				<td>1.125rem</td>
+				<td>18px</td>
+				<td class="table-cell-expand"><span class="text-5">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-4</td>
+				<td>1rem</td>
+				<td>16px</td>
+				<td class="table-cell-expand"><span class="text-4">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-3</td>
+				<td>0.875rem</td>
+				<td>14px</td>
+				<td class="table-cell-expand"><span class="text-3">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-2</td>
+				<td>0.75rem</td>
+				<td>12px</td>
+				<td class="table-cell-expand"><span class="text-2">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+			<tr>
+				<td class="table-cell-ws-nowrap">text-1</td>
+				<td>0.625rem</td>
+				<td>10px</td>
+				<td class="table-cell-expand"><span class="text-1">The quick brown fox jumped over the lazy dog.</span></td>
+			</tr>
+		</tbody>
+	</table>
+</div>
+
+```html
+<span class="text-11">The quick brown fox jumped over the lazy dog.</span>
+<span class="text-10">The quick brown fox jumped over the lazy dog.</span>
+<span class="text-9">The quick brown fox jumped over the lazy dog.</span>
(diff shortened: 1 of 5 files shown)


claraizquierdo:
Hi @pat270 ,

I think that this T-shirt sizes naming with the numbers can be a good solution but I don't know would you add a new 13px size (if necessary). If you add this utility classes I think that it could be a good idea to explain in Clay documentation why we use this naming because for me is not too obvious, but it's not difficult to understand.

claraizquierdo:
Oh, and thank you very much for implementing the solution so fast 😍

pat270:
@claraizquierdo adding 13px does throw the scale off. I wonder if we can just use px values?

```
fs-10
fs-12
fs-14
fs-16
fs-18
fs-20
...
fs-40
```

matuzalemsteles:
Well, the idea of not adding the fixed `px` size to the class name is that the size is fixed, if the scale changes this becomes complicated to change because you can't change the size of an `fs-10` class because it is fixed, so we would have to add a new class and change it in all components that are using `fs-10`.

Using shirt size for example helps to abstract this scale that allows changing in the future, so we don't need to add a new class or change the components, it can be a little confusing at first to assimilate shirt sizes in pixels, but that will be better than adding fixed values.

I like the idea of a shirt for scale but we could try to soften it a little, despite being a nomenclature known to all, I agree that some nomenclatures make it a little difficult to understand, I also like more "obvious" names, for example, `fs` is a good abbreviation it's not bad but `text` is much easier to take in what it's saying. Also, I would prefer that instead of the base being `M` I would go with `fs-base` instead `M` base depends a lot on the context and what we are talking about. Much of what is here is interpretation, it changes from person to person, let me know if that makes sense, I'm fine with this implementation.

ethib137:
I like where we are going with this. One reason we didn't want to use traditional t-shirt sizes was in order to have the possibility of responsive font-sizes. Right now that would look something like this:
```
fs-sm-s
fs-md-s
fs-lg-s
fs-xl-s
fs-sm-s1
fs-md-s1
fs-lg-s1
fs-xl-s1
...
```
`fs-sm-s` has me a little concerned. I agree with @matuzalemsteles that this would be improved by switching to `text-sm-s` improves it but `sm-s` is still confusing. I think we could add consistency and likely clarity by starting our scales with `s0` or `s1` rather than just `s`. This would provide a clearer pattern for the font-size identifier that draws your attention to it's purpose. For example:
```
text-sm-s0
text-md-s0
text-lg-s0
text-xl-s0
text-sm-s1
text-md-s1
text-lg-s1
text-xl-s1
...
```

I still think this has it's issues which is mainly due to us following Bootstraps paradigm of adding the screen size modifier in the middle of the class. In this case I think [tailwind](https://tailwindcss.com/docs/font-size#breakpoints-and-media-queries) has a better method of prepending the screensize to the className in a way that clearly distinguishes what it is: `md:text-base`. Don't worry, I'm not suggesting we switch to this now.

We aren't adding responsive text sizes yet, but it's still good to plan in a way that would work for it. For now I still think it would be improved to have our scales always include a number rather than start without. For example:

```
text-s0
text-s1
text-s2
```

claraizquierdo:
I asked the question about how to add a new font size if needed because Emiliano said:

> I also think a common scenario for this would be a request to extend between steps (like adding one number between XL and XXL) and that’s difficult to provide, but not impossible.

With this classes I guess that if we need to add a new font-size of 13px we would have to change the values for the S sizes, but M, L and XL sizes would stay the same:
```
  fs-S: 14px;
  fs-S1: 13px;
  fs-S2: 12px;
  fs-S3: 10px;
```

Am I right?

I think that using the px value in the name of the class might not be a good idea because if we change any value we would have to change the classes names everywhere.

pat270:
@matuzalemsteles Technically there is no unit on `-14`. It's just a number on a scale. It just happens to be 14px 😆. We can allow modifying these utilities based on scope (we wouldn't do this, just the option is there), `.sidebar .fs/text-14 { font-size: 12px; }` for example. Level 14 text size in sidebar is 12px.

I'm going to be honest about these utilities, they are great when modifying creating your own components. Once you have placed them in OOTB components for reuse, it will cause havoc when someone shifts the scale. Our thought process is we need this font size here and this does the trick. There is no pattern or logic behind our scale. Even if we did apply a pattern, the pattern won't fit another's design system.

Maybe I'm being pedantic and this won't be as bad as I'm making it out to be. I keep going back to the `1-11` scale. Other systems can namespace and add their own scale if shifting scales is too disruptive.

With that said, whatever is agreed upon is good with me.

claraizquierdo:
@pat270 the solution that you have implemented works for me 👍 . I agree with you about changing the scale, is something risky and should be done carefully.

brunofernandezg:
now we all can quote @emiliano-cicero 😃

emiliano-cicero:
Hey guys, sorry I didn't see this. 

I will share some thoughts that might be useful for the thread (or not 😄 ):

**T-shirt sizes:** 
This could be a way of simplifying the use of the classes, it's true that it's difficult to add a 13px value between 12 and 14, but consider also that such a request will not come from the Lexicon scale any time soon. 

**Fixed px values:** 
I think this method would better represent the reality of the scale and allow eventually to include more sizes, I know it can be limited when trying to customize the entire system, but again: from Lexicon it's improbable that we ask for an update to the scale. Even if that happens, it would be a one-time update that will be documented and informed so I don't see the problem there.
The only limitation I see is when someone customizes the 16px base with something different (such as 18px), in that case, all the values will change and the names won't represent the value anymore. But: if that's the case they can also update the name of the classes to represent their system, in the end, we're talking about customization.

To throw a bone to the PX values, that's how we named the scale in Lexicon, we have the number and the weight: 
16-Regular
16-Semibold
16-Bold 

In code it could be organized with 2 utilities FontSize and FontWeight:
fs-16 + fw-R
fs-16 + fw-S
fs-16 + fw-B

I think this would be super easy to use since Figma gives those values in the inspector and no one would have to guess anything or search for anything.

matuzalemsteles:
> I  keep going back to the 1-11 scale. Other systems can namespace and add their own scale if shifting scales is too disruptive. 

@pat270 I don't see a problem with going with a 1-11 scale I get more technically concerned when we're talking about using the same font size value in the class name, so this could be a problem.

For example, if people want to customize the size to `fs-xl` that would be fine but updating to `fs-20` would be more of a consistency issue and the class wouldn't drive with what it's actually saying, so it makes sense to be `fs-xl` or `fs-6` and the size change depending on the component although I don't think it should as there are these DXP cases it makes more sense to me.

> To throw a bone to the PX values, that's how we named the scale in Lexicon, we have the number and the weight:

For these cases I would prefer to provide a utility class to change the weight, this would be more tied to the utility composition and we would have less CSS in this case.

pat270:
I'm going to go with `.text-{1-11}`. Most font/text utilities are written with `.text-*`, the only outlier is `.font-weight-*`. We can alias them to `.text-weight-*` to keep the name consistent.

@ethib137 The responsive naming should be less confusing with this syntax. `text-1-{sm|md|lg}`. You will be able to add any number of font size utilities by adding them to the `$font-sizes` map.

ethib137:
Hey @pat270, yeah this seems reasonable.

Regarding the font-weight utilities would those be using numbers [1-7] or the weight names [lighter|light|normal|semi-bold|bold|bolder]. If we plan to use the weight names it would be more consistent to name them `text-bold` etc. After all, for font-size we aren't using the utility name `text-size-{1-11}`. But, sadly we can't do that since we already have a `text-light` utility class that changes the color. 

I guess `text-weight-*` is probably the best option, but it's too bad it's still slightly inconsistent and deviates slightly from the css property name. Personally I think `text-weight-*` and `font-weight-*` are equally reasonable, so I'm happy with whatever is decided.

pat270:
This is ready for review. I added:
```
text-1
text-2
text-3
text-4
text-5
text-6
text-7
text-8
text-9
text-10
text-11
```
Aliases for consistency:
```
text-weight-lighter
text-weight-light
text-weight-normal
text-weight-semi-bold
text-weight-bold
text-weight-bolder
text-italic
font-monospace
```

matuzalemsteles (approved):
LGTM! Good job @pat270!

PR #4617, opened 2022-02-01
fix(@clayui/shared): fix focus control collision when having FocusScope nested

### LINKED ISSUE #4613: @clayui/date-picker: Tab button ignores minutes field on the Date Picker input
Hello guys, it's me again! Thanks for taking care of those other bugs so fast, we really appreciate it! That said, we need your help again related to the Date Picker.

We found [this](https://issues.liferay.com/browse/LPS-146011) bug where the tab button ignores minutes field on the Date Picker input; it doesn't matter if the `use12Hours` is enabled or not. This is a problem because compromises the accessibility of the product. You can see a video of this behaviour [here](https://issues.liferay.com/secure/attachment/438971/ActualResult_Jan24-LPS-146011.mp4).

If this is something we could fix on our side please let us know, thanks a lot!

You can test this on the [Clay Documentation](https://clayui.com/docs/components/date-picker.html) or here: https://codesandbox.io/s/cranky-platform-47ldv?file=/src/index.js

1. Open the date picker modal
2. Click on the first `00` (hours) on the time input
3. Press the tab key and observe

After pressing the tab key, the selection should go to the other `00` value (minutes), but it goes to the `x` button. This happens for both 24h and 12h time formats.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.43.0  |
| React | v17.0.2  |

matuzalemsteles:
Fixes #4613

There are some cases where FocusScope is declared more than once nested, for example, TimePicker uses FocusScope to manage focus with left and right keys and the component is declared in DatePicker which also has a FocusScope declared.

There is a conflict of who is managing an element aria between the two FocusScopes declared nested and each one can have a configuration, to avoid having conflict between the focus managers, we use the context to know if there is a focus manager parent and we avoid propagation of the event.

### DIFF
--- a/packages/clay-shared/src/FocusScope.tsx
+++ b/packages/clay-shared/src/FocusScope.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import React from 'react';
+import React, {useContext} from 'react';
 
 import {Keys} from './Keys';
 import {useFocusManagement} from './useFocusManagement';
@@ -27,6 +27,13 @@ interface IProps {
 }
 
 /**
+ * The context helps to identify if the FocusScope is being declared nested, to
+ * avoid focus being controlled by more than one focus manager, we stop event
+ * propagation to prevent the parent focus generator from doing anything.
+ */
+const FocusConflictContext = React.createContext<boolean>(false);
+
+/**
  * FocusScope is a component only for controlling focus and listening
  * for children's key down events, since the component handles the `onKeyDown`
  * event.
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
@julien I'm going to merge this PR to go into today's release, this is being urgent for the Bravo team (Forms and Objects), this is not very trivial and shouldn't add any side effects but let me know your opinion anyway, anything we can send another PR with improvements if necessary.

julien:
@matuzalemsteles thanks for the heads up - the change doesn't look too risky for me (and I understand it) - maybe we'll want to update the tests to show this in action, appart from that good work.

matuzalemsteles:
@julien yeah, we need to write more tests for this component, I'll create an issue so we can work on it later.

PR #4625, opened 2022-02-02
fix(@clayui/core): TreeView: fix error when selecting parent nodes and controlling intermediate state in multiple selection

### LINKED ISSUE #4623: Multiple selection should not automatically select parent nodes
After further clarifying this behavior with David in the [document](https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1973518798/Interaction?focusedCommentId=1975288389#Single), multiple selections differs more from recursive multiple selections that automatically when all children's nodes are selected will select the parent recursively as well and handle the intermediate state but this should only apply to recursive multiple selections.

Following that, the behaviors are:

- `multiple-recursive` - Recursively select the child nodes of the node and the parent if all child nodes is not selected set it to intermediate.
- `multiple` - It only allows selecting more than one node in the TreeView

matuzalemsteles:
Fixes #4623

I'm just removing the option to select parent nodes when the `selectionMode` is multiple only, in the code I'm just separating and repeating a bit to avoid the multiple conditions if the `selectionMode` is `multiple-recursive`.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -31,6 +31,13 @@ interface ITreeViewProps<T>
 	dragAndDrop?: boolean;
 
 	/**
+	 * Optional drag and drop context: this is to avoid
+	 * errors when using various drag and drop contexts
+	 * on the same page.
+	 */
+	dragAndDropContext?: Window & typeof globalThis;
+
+	/**
 	 * Flag to expand child nodes when a parent node is checked.
 	 */
 	expandOnCheck?: boolean;
@@ -82,6 +89,7 @@ export function TreeView<T>({
 	className,
 	displayType = 'light',
 	dragAndDrop = false,
+	dragAndDropContext = window,
 	expandOnCheck = false,
 	expandedKeys,
 	expanderClassName,
@@ -143,7 +151,10 @@ export function TreeView<T>({
 				ref={rootRef}
 				role="tree"
 			>
-				<DndProvider backend={HTML5Backend}>
+				<DndProvider
+					backend={HTML5Backend}
+					context={dragAndDropContext}
+				>
 					<TreeViewContext.Provider value={context}>
 						<Collection<T> items={state.items}>
 							{children}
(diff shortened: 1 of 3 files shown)


julien:
Hey @matuzalemsteles, the code changes LGTM, however when switching from "multiple-recursive" to "multiple" in the story, this happens



I know that this is caused by [4faf0ea2049744b0587afec3ca2ee02bf916fe4d](https://github.com/liferay/clay/commit/4faf0ea2049744b0587afec3ca2ee02bf916fe4d), but here's the issue: fixing it here will make it fail in DXP ... So what's the solution? Should we just focus on fixing it here "without" knowing what will happen in DXP; that doesn't sound like the best option.

matuzalemsteles:
> I know that this is caused by [4faf0ea2049744b0587afec3ca2ee02bf916fe4d](https://github.com/liferay/clay/commit/4faf0ea2049744b0587afec3ca2ee02bf916fe4d), but here's the issue: fixing it here will make it fail in DXP ... So what's the solution? Should we just focus on fixing it here "without" knowing what will happen in DXP; that doesn't sound like the best option.

yeah, in DXP we would have to add the same context for all, we had a similar problem in migrating Forms which had some `DndProvider` nested and another one just outside the React tree, ideally, we had to add all `DndProvider` under the same context, in In the case of Forms we just added the `window` where everyone should consume it, it was easier to reference but it is also possible to make it add a `React.Context` also to be isolated only in the application.

Maybe not directly add the `window` as the context here, we can provide a prop to make it configurable, if it has a problem we can let the developers configure where it should consume the state. What do you think?

julien:
@matuzalemsteles thanks for the feedback - Yes, I think adding a prop might be the way to go since it's more flexible as you said. In any case, I'll have a look at this and try to fix the problem both in storybook and DXP.

matuzalemsteles:
@julien ok, do you want to send on top of this PR or do you prefer to send in a separate PR?

julien:
@matuzalemsteles I pushed a new commit https://github.com/liferay/clay/pull/4625/commits/59f9f583a589dfb1494eb82ad2ca0e0fa7e2ed1d, let me know what you think when you get the chance. Thanks

matuzalemsteles:
@julien looks good, thanks for adding this, I will merge.

Issue #4605, opened 2022-01-27
Using pagination components with a big number of pages can cause performance degradation


victorg1991:
When having a lot of pages, the pagination component create links to some of the pages

For example, when having 100 pages, we would have something like this:

![Screenshot 2022-01-27 at 13 31 45](https://user-images.githubusercontent.com/11654230/151359449-94908b68-0423-4214-8250-0d1eb1bfd5f7.png)

When clicking on the ellipsis button, a dropdown appears allowing the user to select a specific page.

This may be fine when having a low number of pages, but in this case for example, the right ellipsis will show a dropdown of 90 items, and having more pages will result in a higher dropdown that will cause performance problem due to React having to render all that items

## Is there an example you can provide via codesandbox.com?

Here: https://codesandbox.io/s/falling-river-3t2i9?file=/src/index.js:257-285

you can change the total pages value

## What are the steps to reproduce?

- Adding any pagination component
- Set a high number of pages

## What is the expected result?

Having a lot of pages shouldn't result in the pagination component creating a dropdown with a link to all the pages

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.40.0  |
| React |  v16.12.0       |

victorg1991:
Talked with @julien about these, we may just simply limit the number of items in the dropdown with a prop, but this will probably require some analysis by lexicon.

I have checked lexicon site but it doesn't seems to specify anything for these cases.

@drakonux what do you think?

julien:
@victorg1991 thanks for opening this - that's definitely a problem (and an interesting example). 

As you said, we might need help from design to help us solve this, but we should also think about what to do with the component. As you said during our call, maybe the team using the component should limit the number of items they pass in, but there's probably something else we need to do internally.

victorg1991:
I've checked again and I don't think we can do anything without changing something in clay. Do you think we can add the aforementioned prop to limit that number? This way we can fix it the problem and design can think of a better way to handle this :)

julien:
@victorg1991 If you have an idea of how to solve this, please feel free to submit a pull request - If @matuzalemsteles has a better idea, he will comment and we can talk about alternative options.

matuzalemsteles:
oh, this is an interesting case, I hadn't thought about these possibilities within DropDown. I didn't quite understand how this limit would work, in that case, we would define a limit for the number of items inside the DropDown but would it appear in the pagination?

[...]

PR #4631, opened 2022-02-03
feat: Allow disabling ellipsis button in pagination

victorg1991:
This is potential quick fix for https://github.com/liferay/clay/issues/4605 until lexicon revise this pattern

Here we simply allow disabling the ellipsis button to avoid having a dropdown with a huge number of items (for bigger pagination)

Let me know what do you think :)


matuzalemsteles (commented):
I think it's a good option, not sure if it affects the user experience but something we can have for now.

julien (approved):

julien:
@matuzalemsteles good point - I also agree that having this option doesn't do any harm so I'm merging this. Thanks @victorg1991

PR #4643, opened 2022-02-07
fix(@clayui/date-picker): Reset actual date when input string is empty

### LINKED ISSUE #4470: @clayui/date-picker: Not possible to reset selected days in calendar when setting value to empty string
Hello,

Not sure if this is a bug, question or change request.
I'm currently using the date-picker component, and there is one functionality that seems to be either not working or missing.

When I want to _reset_ the date value from outside of the component, the actual value of the input gets reset, but the selected days in the calendar remain selected. The _setDaysSelected_ function does not seem to be called when the value is changed to an empty string value. This seems to cause a temporary disconnect between the input value and the visual state of the calendar.

I think a fix could be to have the _inputChange_ function also count an empty value as a valid date, instead of only checking the format of the date.

Some extra information on my setup and a visual of the problem:

calendar-days-not-resetting

simplified code used
```js
<div>
    <ClayDatePicker
        /* ... */
        onValueChange={(value) => { setValue(value); }}
        range
        value={value}
    />
    <Cross
        onClick={() => { setValue(''); }}
    />
</div>
```

Let me know if I'm doing something silly, missing something obvious, or can provide extra information on this.

Thanks!

sergiojimcos:
**@clayui/date-picker:** Reset to actual date when input of the date picker is empty

Fixes #4470

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -350,7 +350,7 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 				setDaysSelected([initialMonth, initialMonth]);
 
 				if (time) {
-					setCurrentTime(0, 0, undefined);
+					setCurrentTime('--', '--', undefined);
 				}
 			} else {
 				const format = time
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -345,23 +345,12 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 			const {value} = event.target;
 
 			if (!value) {
-				const newDaysSelected: [Date, Date] = [NEW_DATE, NEW_DATE];
-
-				const initialTime = {
-					hours: 0,
-					minutes: 0,
-				};
-
 				changeMonth(initialMonth);
 
-				setDaysSelected(newDaysSelected);
+				setDaysSelected([initialMonth, initialMonth]);
 
 				if (time) {
-					setCurrentTime(
-						initialTime.hours,
-						initialTime.minutes,
-						undefined
-					);
+					setCurrentTime(0, 0, undefined);
 				}
 			} else {
 				const format = time
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -346,33 +346,23 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 
 			if (!value) {
 				const newDaysSelected: [Date, Date] = [NEW_DATE, NEW_DATE];
-				let dateFormatted;
 
-				changeMonth(NEW_DATE);
+				const initialTime = {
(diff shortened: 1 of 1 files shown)


packages/clay-date-picker/src/index.tsx
```diff
@@ -344,35 +344,68 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
 		const inputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
 			const {value} = event.target;
 
-			const format = time
-				? `${dateFormat} ${use12Hours ? TIME_FORMAT_12H : TIME_FORMAT}`
-				: dateFormat;
+			if (!value) {
```
matuzalemsteles, line 347: For that case when you don't have a value in the input, you can simplify the implementation here to just reset to the initial values. For example, we don't need to call `onValueChange` and do `dateFormatted` because in that case, we expect the state to be empty.

- Instead of using `NEW_DATE` directly, you can use `initialMonth` because it can be configured by the developer.
- I imagine you only need to call `changeMonth` with `initialMonth`, `setDaysSelected ` and `setCurrentTime` if the time is enabled.

packages/clay-date-picker/src/index.tsx
```diff
@@ -344,36 +344,59 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
+				const newDaysSelected: [Date, Date] = [NEW_DATE, NEW_DATE];
```
matuzalemsteles, line 348: For this one you would need to use `initialMonth` as well because it can happen that the developer sets a date different from the current date he would like, this is common for Forms for example.

```suggestion
				const newDaysSelected: [Date, Date] = [initialMonth, initialMonth];
```

packages/clay-date-picker/src/index.tsx
```diff
@@ -344,36 +344,59 @@ const ClayDatePicker: React.FunctionComponent<IProps> = React.forwardRef<
...
 
-			const [startDate, endDate] = fromStringToRange(
-				value,
-				format,
-				NEW_DATE
-			);
+				const initialTime = {
+					hours: 0,
+					minutes: 0,
+				};
```
matuzalemsteles, line 353: Just a suggestion, just for cases like this where you won't change the variable later, you can pass these values directly to `setCurrentTime` this avoids the allocation of a new variable. You can probably do the same for `newDaysSelected`.

packages/clay-date-picker/src/index.tsx
sergiojimcos: Ohh understood! I thought otherwise, the developer could change `NEW_DATE` constant

matuzalemsteles (approved):
LGTM! @sergiojimcos I just added a commit to reset the time to `--` instead of zero, as this is the initial state of the time.

sergiojimcos:
> LGTM! @sergiojimcos I just added a commit to reset the time to `--` instead of zero, as this is the initial state of the time.

Thanks @matuzalemsteles!

Issue #4368, opened 2021-10-26
Versioning clayui.com documentation or create a lightweight smoke test site


matuzalemsteles:
After we support Node.js 14.x and remove the smoke test site settings from the `clay-css` package we need to decide which direction to go or create a lightweight script to set up a simple smoke test site or create the version for the documentation on clayui.com, as suggested by @pat270 https://github.com/liferay/clay/pull/4317#issuecomment-948869038.

The idea of documentation versioning is a pretty old story that we've discussed here many times and I really like the idea of versioning, especially when we start to release more majors version instead of maintaining a version specific site in another domain.

The smoke test site was important for @pat270 to add many examples, the downside is that we had some examples that weren't on clayui.com, not having this site anymore can help to add more examples on clayui.com but the bad part is clayui.com is a bit slow due to the number of plugins, files, and processes we have.

The issue's idea is to try to see all the points and see what is the best approach to follow here, to discuss the pros and cons.

matuzalemsteles:
LPS Ticket https://liferay.atlassian.net/browse/LPS-190551

PR #4647, opened 2022-02-08
fix(@clayui/date-picker): fix error when formatting time using 12 hours

### LINKED ISSUE #4638: @clayui/date-picker: Time is not being correctly sent on the field input
Hello guys, hope you are doing well!

I'm opening this issue because it seems that the solution of [this issue](https://github.com/liferay/clay/issues/4608) I've opened before didn't cover all the cases. The input of the time value on the date picker is changing correctly, but if I click on a date after that, it still sends the wrong value. For example, if I type `02/02/2022 05:00 PM` on the input field and open the date picker modal, the time input is with the correct value; but if after that I click on a date, it will send the wrong time value as `02/02/2022 17:00 PM`. Also if I do the same steps but with the date and time `02/02/2022 12:00 AM` it sends to the field input the value `02/02/2022 00:00 AM`.

Also a new behaviour is occurring that if I type a date on a language that doesn't use AM/PM, like for example `02/02/2022 13:00` and click on another date on the date picker, it will send to the input the value `02/02/2022 13:00 PM` that isn't a valid hour. If i put `03/02/2022 02:22` and click on another date on the date picker, it will send `03/02/2022 02:22 AM` to the input.

A video of these two behaviours can be seen [here](https://drive.google.com/file/d/1mXA5kORmRZPKcrekNL9YXVqDEqc99BhS/view?usp=sharing).

If need more info or help with anything let me know, thanks!

You can test both scenarios here: https://codesandbox.io/s/cranky-platform-47ldv?file=/src/index.js

_For the first scenario (AM/PM language case):_

1. Type or paste on the field input a date and time like `02/03/2022 09:00 PM`
2. Click on the date picker icon to open the modal
3. Click on any other date
4. Close the modal and observe the value on the field input

_For the second scenario (adding AM/PM case):_

1. Remove the `use12Hours` prop from the `ClayDatePicker` in the code to simulate a language that doesn't use AM/PM
2. Type or paste on the field input a date and time like `02/03/2022 13:00`
2. Click on the date picker icon to open the modal
3. Click on any other date
4. Close the modal and observe the value on the field input

_For the first scenario (AM/PM language case):_
The value on the field input should be the new date with `09:00 PM` and not `21:00 PM`.

_For the second scenario (adding AM/PM case):_
The value on the field input should be the new date with `13:00` and not `13:00 PM`. The same is happening with AM values: If i put `01:00` it will show `01:00 AM`.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | 3.45.1  |
| React |  v17.0.2  |

matuzalemsteles:
Fixes #4638

The time formatting was only being done when it was passed to the TimePicker component but not being formatted correctly when it is saved in the internal state, this caused other actions to change the value to the wrong time when changing the date for example.

The second case was always saving the ampm in the time state, just checking that `use12Hours` is enabled before saving.

### ToDo

- [x] Add tests for this use case

### DIFF
--- a/packages/clay-date-picker/src/Hooks.ts
+++ b/packages/clay-date-picker/src/Hooks.ts
@@ -60,7 +60,7 @@ export const useWeeks = (
 /**
  * Sets the current time
  */
-export const useCurrentTime = () => {
+export const useCurrentTime = (use12Hours: boolean) => {
 	const [currentTime, set] = useState<string>('--:--');
 
 	const setCurrentTime = useCallback(
@@ -73,6 +73,10 @@ export const useCurrentTime = () => {
 
 			if (typeof hours !== 'string') {
 				hours = formatDate(date, 'HH');
+
+				if (use12Hours) {
+					hours = formatDate(setDate(new Date(), {hours}), 'hh');
+				}
 			}
 
 			if (typeof minutes !== 'string') {
--- a/packages/clay-date-picker/src/TimePicker.tsx
+++ b/packages/clay-date-picker/src/TimePicker.tsx
@@ -6,8 +6,6 @@
 import ClayTimePicker, {Input} from '@clayui/time-picker';
 import React from 'react';
 
-import {formatDate, setDate} from './Helpers';
-
 interface IProps {
 	currentTime: string;
 	disabled?: boolean;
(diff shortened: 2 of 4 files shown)


julien:
@matuzalemsteles thanks for this, but I can't see that this change fixes the issue described in #4368:

When pasting a date "string", i.e. 02/03/2022 09:00 PM in the date picker with time this is what I see



I might not be understanding correctly, but I would assume the date and time to be set correctly on the "calendar" pop up.

Since you also added a "TODO" item to add a test, I'll let you have a look at that before merging.

matuzalemsteles:
@julien in this particular case, she configured the Date Picker to accept the Brazilian format, I imagine, in the storybook example I left it default, this did not update because there was no valid date.

Thinking about it a bit more, we're usually letting people add validation on their side for teams to use their own rule, maybe if we add a helper callback to notify that the date is not being validated, what do you think?

Something like that:

```jsx
<ClayDatePicker
  onValidation={() => {}}
/>
```

I'm not sure about this, normally validations are done by whoever is controlling the component. For example when using a Formik with Schema validation.

PR #4648, opened 2022-02-08
fix(@clayui/multi-select): fix error when adding labels with whitespace

### LINKED ISSUE #4637: Multi Select allows White Space Tags
In Multi Select Section, adding a space in the tag section and adding comma will result in an empty tag. This can be done with various empty string sizes.

This can also be seen in https://issues.liferay.com/browse/LPS-146715 which causes issues when trying to publish.

https://user-images.githubusercontent.com/31672551/152584226-b7c913d5-93d0-47cd-9382-3b56d86d18d4.mp4

I believe the expected result would be not allowing white space strings as tags.
## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v?.?.?  |
| React |         |

matuzalemsteles:
Fixes #4637

### DIFF
--- a/packages/clay-multi-select/src/index.tsx
+++ b/packages/clay-multi-select/src/index.tsx
@@ -222,7 +222,7 @@ const ClayMultiSelect = React.forwardRef<HTMLDivElement, IProps>(
 				event.preventDefault();
 			}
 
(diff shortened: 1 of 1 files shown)


julien (approved):
LGTM

PR #4653, opened 2022-02-09
chore(@clayui/core): call `onChange` if declared in Checkbox

### LINKED ISSUE #4652: Adds the possibility to add `onChange` to Checkbox in TreeView
Currently adding the `<Checkbox />` in the TreeView component abstracts the need to add `onChange`, `checked`, `indeterminate` and `disabled`, this works OOTB inside the TreeView but there are some cases inside the DXP that we need know when the Checkbox has been selected to invoke a method passing the item to store the data of the selected items so you don't have to traverse the tree in search of information.

The improvement here would be to be able to call `onChange` and also to be able to override the default behavior to allow for more configuration options.

matuzalemsteles:
Fixes #4652

This PR is adding the possibility to call the `onChange` method if it is declared in the `Checkbox`, ideally, the TreeView abstracts the need to control the Checkbox but some need to listen for the `onChange` event to do other operations.

I'm also adding the same behavior as in `onClick` to be able to prevent the default behavior.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -462,7 +462,20 @@ export function TreeViewItemStack({
 						checked: selection.selectedKeys.has(item.key),
 						disabled: loading,
 						indeterminate: selection.isIntermediate(item.key),
-						onChange: () => {
+						onChange: (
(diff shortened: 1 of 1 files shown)


julien (approved):
LGTM . Seems we'll need this to continue working on the "TreeView migration" epic.

Issue #4641, opened 2022-02-07
Props "Small" in EmptyState component


BarbaraCabrera:
## What is your proposal?

Implement a new props small in EmptyState component.

## Why would adopting this proposal be beneficial?

For the moment it will be used in admin cases, but it is not excluded that it will be used in public sites as well. Also, this was defined in [[LEXI-668](https://issues.liferay.com/browse/LEXI-668)] the (03/Oct/19) for team-tango.

sergiojimcos:
Hi @matuzalemsteles! I will be in charge of this issue!

pat270:
The only difference is the size of the image?

sergiojimcos:
Hi @pat270! In this case the proposal affects the **size of the image** as well as some properties that contains this component like the **title**, **description** and that's all.

PR #4654, opened 2022-02-09
feat(@clayui/empty-state): Include Small property in empty state component

sergiojimcos:
**What is new?**
New feature in clayui/empty-state, `small` property is included in this component.

https://user-images.githubusercontent.com/44723449/153244299-eef1558b-1d40-4804-b462-190904464f9f.mov

**Fixes**
#4641

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
 		"coverage": "jest --coverage",
 		"format": "prettier --write \"**/*.{js,ts,tsx,md,mdx,json,scss}\"",
 		"format:all": "yarn lint:fix && yarn format",
-		"format:changed": "git ls-files -mz \"*.js\" \"**/*.js\" \"**/*.ts\" \"**/*.tsx\" \"*.md\" \"**/*.md\" \"**/*.mdx\" \"*.json\" \"**/*.json\" | xargs -0 prettier --write --",
+		"format:changed": "git ls-files -mz \"*.js\" \"**/*.js\" \"**/*.ts\" \"**/*.tsx\" \"*.md\" \"**/*.md\" \"**/*.mdx\" \"*.json\" \"**/*.json\" \"**/*.scss\" | xargs -0 prettier --write --",
 		"format:check": "prettier --list-different -- \"**/*.{js,ts,tsx,md,mdx,json,scss}\"",
 		"genBuildSize:compare": "NODE_ENV=production node scripts/check-size/index.js --compare",
 		"genBuildSize": "NODE_ENV=production node scripts/check-size/index.js",
--- a/packages/clay-css/src/scss/variables/_empty-state.scss
+++ b/packages/clay-css/src/scss/variables/_empty-state.scss
@@ -13,19 +13,6 @@ $c-empty-state-animation: map-deep-merge(
 	$c-empty-state-animation
 );
 
-$c-empty-state-animation-sm: () !default;
-$c-empty-state-animation-sm: map-deep-merge(
-	(
-		align-items: center,
-		display: flex,
-		flex-direction: column,
-		margin: 40px 32px auto,
-		max-width: 268px,
-		text-align: center,
-	),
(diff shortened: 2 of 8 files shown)


julien:
@matuzalemsteles just so you know, I had to edit our `format:changed` script because it wasn't checking scss files. @sergiojimcos sorry about that.

packages/clay-empty-state/src/index.tsx
```diff
@@ -42,15 +48,26 @@ const ClayEmptyState: React.FunctionComponent<IProps> = ({
...
-				<div className="c-empty-state-image">
+				<div
+					className={classNames(
+						className,
+						{
+							'c-empty-state-image': !small,
+						},
+						{
+							'c-empty-state-image-sm': small,
+						}
```
matuzalemsteles, line 65: This is just a small change 😁 but in these cases, you can group everything into one object, it wouldn't be necessary to separate each condition into an object.

matuzalemsteles (commented):
Well looks good too, good job! I just left a comment for adjustment but very small.

packages/clay-empty-state/src/index.tsx
sergiojimcos: You are right, I haven't seen that before 😄

pat270:
@sergiojimcos Thanks for the pull. I think one improvement we can make is adding the sm class at the top dom element. It would be `c-empty-state-sm`. The Sass would look like:

```scss
$c-empty-state-sm: () !default;
$c-empty-state-sm: map-deep-merge(
  (
    c-empty-state-animation: (),
    c-empty-state-image: (),
    c-empty-state-title: (),
  ),
  $c-empty-state-sm
);

.c-empty-state-sm {
  &.c-empty-state-animation {
    @include clay-css(map-get($c-empty-state-sm, c-empty-state-animation));
  }

  .c-empty-state-image {
    @include clay-css(map-get($c-empty-state-sm, c-empty-state-image));
  }
}
```

packages/clay-css/src/scss/variables/_empty-state.scss
```diff
@@ -13,6 +13,19 @@ $c-empty-state-animation: map-deep-merge(
 	$c-empty-state-animation
 );
 
+$c-empty-state-animation-sm: () !default;
+$c-empty-state-animation-sm: map-deep-merge(
+	(
+		align-items: center,
```
pat270, line 19: A lot of styles are duplicated, we can just inherit them from the original component.

julien:
@sergiojimcos I'll have another look, but for the CSS part, let's ping @pat270

package.json
```diff
@@ -22,7 +22,7 @@
 		"coverage": "jest --coverage",
 		"format": "prettier --write \"**/*.{js,ts,tsx,md,mdx,json,scss}\"",
 		"format:all": "yarn lint:fix && yarn format",
-		"format:changed": "git ls-files -mz \"*.js\" \"**/*.js\" \"**/*.ts\" \"**/*.tsx\" \"*.md\" \"**/*.md\" \"**/*.mdx\" \"*.json\" \"**/*.json\" | xargs -0 prettier --write --",
+		"format:changed": "git ls-files -mz \"*.js\" \"**/*.js\" \"**/*.ts\" \"**/*.tsx\" \"*.md\" \"**/*.md\" \"**/*.mdx\" \"*.json\" \"**/*.json\" \"**/*.scss\" | xargs -0 prettier --write --",
```
julien, line 25: Good catch.

packages/clay-css/src/scss/variables/_empty-state.scss
julien: @pat270 could you please have another look, it seems @sergiojimcos updated this. Thanks!

julien:
@pat270 thanks again

sergiojimcos:
> I updated the scss. This should be gtg.

Thanks!!

Issue #4661, opened 2022-02-11
Parent nodes should be expanded when any of its child is expanded


victorg1991:
Right now if you have a 3 level tree for example, you need and you want to expand the grandChild node, you need to pass the key of all the parents or the node won't be expanded. This behaviour it's not ideal since the selected node is unnoticeable.

## Is there an example you can provide via codesandbox.com?

https://codesandbox.io/s/vibrant-kapitsa-u3ttn?file=/src/index.js

## What is the expected result?

In the codesandox above name1 node and child1 node should be expanded, showing that grandchild1 is selected.

julien:
@victorg1991 makes sense, we will fix this

PR #4664, opened 2022-02-12
fix(@clayui/core): fixes the error of not revalidating the intermediate state of the selected item's parents in the first render and expands the items

### LINKED ISSUE #4660: Pre-selected items in TreeView with recursive multiple selection do not set an intermediate state for the parent
When you render the TreeView with some pre-selected items it doesn't mark the items of the parents of the selected item as intermediate state, that's because this state is not saved or saved on the backend and just a visual effect to show users when we are interacting with the items of TreeView. In this case, we add the intermediate state only to `layoutKeys` and when the user will interact with the TreeView.

As we cannot get this information otherwise, we will have to go through the tree in search of the items that we must expand and mark as intermediate. There are some exchanges here:

- We will traverse the tree only on the first render
- Traversing the tree will only happen after rendering all items, it will not block the initial rendering.
- We will go through the tree until we check all the selected items (This case is good because it depends where the selected items are, even if there are many selected items we don't need to go through the tree completely)

This case was discovered in PR https://github.com/liferay-echo/liferay-portal/pull/7185#issuecomment-1035123189

matuzalemsteles:
Fixes #4660 #4661

This PR fixes the error that when starting the rendering of the TreeView with selected items, it was not keeping the state of intermediate, that's because this data is not saved in the backend and just for visual effect, so we need to check which items should have the state of intermediate.

Also with items selected in depth were not visible, we need to expand their parent items to make them visible.

Correcting this implied having to traverse the items tree to know which items should be expanded, we have some conditions that can optimize this so that it does not traverse the entire tree, first, we traverse the tree in search of the selected items and stop traversing until we find all the items, depending on the data and where the selected items are, this can be very fast but this is just an inference that there is no way to predict.

The problem with this is that it can block initial rendering until this operation is finished if it is called at `expandedKeys` initialization so that the TreeView component renders first with all items expanded and with the intermediate state will degrade performance to the first startup, this, when you have a lot of data, can become a problem, with little data it might not be a very noticeable problem.

We also have the option of doing this operation after the TreeView is rendered, so it does not block the first rendering but it will be noticeable for the user to see two changes in the TreeView, first, the TreeView being rendered and then the items being expanded and marked with the state of intermediate.

That's why we're introducing the `selectionHydrationMode` prop here so that developers can choose which behavior they want, for cases where very large data is rendered, the `render-first` option is recommended to avoid blocking the first render. By default, we are leaving `hydrate-first` as it doesn't have the effect of seeing items expand.

### ToDo

- [x] Adds TreeView performance recommendations in the documentation

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -90,10 +90,10 @@ export function TreeView<T>({
 	displayType = 'light',
 	dragAndDrop = false,
 	dragAndDropContext = window,
-	expandOnCheck = false,
 	expandedKeys,
 	expanderClassName,
 	expanderIcons,
+	expandOnCheck = false,
 	items,
 	nestedKey = 'children',
 	onExpandedChange,
@@ -102,6 +102,7 @@ export function TreeView<T>({
 	onRenameItem,
 	onSelectionChange,
 	selectedKeys,
+	selectionHydrationMode = 'hydrate-first',
 	selectionMode = 'single',
 	showExpanderOnHover = true,
 	...otherProps
@@ -116,6 +117,7 @@ export function TreeView<T>({
 		onItemsChange,
 		onSelectionChange,
 		selectedKeys,
+		selectionHydrationMode,
 		selectionMode,
 	});
 
--- a/packages/clay-core/docs/treeview.mdx
+++ b/packages/clay-core/docs/treeview.mdx
@@ -34,6 +34,8 @@ import {Example, MultipleSelection} from '$packages/clay-core/docs/treeview';
 -   [Shortcuts](#shortcuts)
     -   [Rename Item](#rename-item)
     -   [Remove Item](#remove-item)
+-   [Performance](#performance)
+    -   [Selection hydration](#selection-hydration)
 
 </div>
 </div>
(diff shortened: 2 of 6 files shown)


packages/clay-core/src/tree-view/TreeView.tsx
```diff
@@ -102,6 +102,7 @@ export function TreeView<T>({
 	onRenameItem,
 	onSelectionChange,
 	selectedKeys,
+	selectionHydrationMode = 'hydrate-first',
```
julien, line 105: Shouldn't this also be declared in [`ITreeViewProps`](https://github.com/liferay/clay/blob/a3db213cc1b92ca4b6b35bb0652169c16ea2418d/packages/clay-core/src/tree-view/TreeView.tsx#L19-L79) ?

julien:
@matuzalemsteles LGTM. I can confirm it fixes both https://github.com/liferay/clay/issues/4660 and https://github.com/liferay/clay/issues/4661. Since you added a "TODO" item, I won't merge right now.

matuzalemsteles:
Hey @julien I've added more information on this to the docs, let me know, if we can improve anything else or if you see any spelling improvements.

julien:
Thanks for that @matuzalemsteles. I made slight changes to the documentation, if the CI is happy I'll merge.

PR #4665, opened 2022-02-14
fix(@clayui/core): make it possible to pass items without onItemChanges

### LINKED ISSUE #4662: TreeView: setItems has no effect without onItemsChange

Sure, [here](https://codesandbox.io/s/beautiful-river-w2wqk?file=/src/index.js) you go

Use `TreeView` and try to set the items to another value at some point

The items should be updated, i.e. the component should be re-rendered

julien:
Fixes #4662


julien:
Closing, more details in https://github.com/liferay/clay/issues/4662

PR #4666, opened 2022-02-14
fix(@clayui/core): avoid scrolling when selecting a node

### LINKED ISSUE #4663: Pressing the space button when a tree node is focused shouldn't cause the page to scroll
When having a TreeView with some scroll, focusing a node and pressing the space key is causing the node to scroll. IINM we should only select the node but not scrolling the page.

https://codesandbox.io/s/flamboyant-lewin-zth87?file=/src/index.js

- Focus one node.
- Press space.

- The node is activated but the page doesn't scroll

julien:
Fixes #4663


julien:
:cry: 
Since I know have the required information to understand what might be failing, I'll have a look.

matuzalemsteles (approved):
LGTM! Thanks Julien

PR #4672, opened 2022-02-15
feat(@clayui/date-picker): adds a new `yearsCheck` prop to be able to disable the check if the year is within the years range

### LINKED ISSUE #4624: @clayui/date-picker: Date Picker modal is not showing the typed value on the field
## What is your question?

Hello guys, I'm not sure if this is a bug on our side or not, thats why I'm opening a question issue (if this is the wrong type please let me know so I can use the right one next time).

There is a [bug](https://issues.liferay.com/browse/LPS-146015) open on Forms and Data Engine side where the date picker modal is not showing the typed value on the field (the ticket was opened for Date and Time field but it's valid for the Date field as well).

From what I investigated, this behaviour occurs when we try to input a date where the year isn't in the default year range (-5 years from the year we are now and +5 from this same year); since we are in 2022 now, the default range that is set first and appears in the years dropdown will be from 2017 to 2027. When I try to type something that is out this range, like 01/01/2028, the modal won't reflect this date. This can be a little bit confusing so I made a [video](https://drive.google.com/file/d/1WRgQRX6S1JK-fCPiSkllYOrO58Q4KS4i/view?usp=sharing) with this behaviour happening.

I know that this behaviour of setting the years dropdown -5 years from the year on the field input (current date) and +5 from this same year is something we do on our side, but from what I saw we are sending the right values to the date picker so I'm not sure if this is something we could fix on our side. I also made a [video](https://drive.google.com/file/d/1qf6vf_LZaVTEPk_Ci9gcDYB8vWS7R0Q7/view?usp=sharing) showing this part of our code working. To test this you can go to the portal on the forms section and follow these steps:

1. Create a new Form with a date field
2. Publish the form and open it
3. On the form, type a value like 01/01/2016 on the date field
4. Open the date picker modal and observe the date reflected

I also want to say that this is not urgent, so you guys don't need to worry; thanks again for the other bugs fixed!

PS: If you want to see our code, the path is `modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/DatePicker/DatePicker.es.js`

matuzalemsteles:
Fixes #4624

This API simplifies the use case of issue #4624, but I added it in the API description to only disable it when the year range is dynamic, otherwise, there will be a visual bug in the year selector.

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -148,9 +148,8 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	years?: IYears;
 
 	/**
-	 * Flag to indicate whether the DatePicker should validate whether the year
-	 * entered in the input is included in the range of years. Disable only
-	 * if your range is dynamic.
+	 * Flag to indicate whether the DatePicker should validate if the year
+	 * is included in the range of years. Disable only if your range is dynamic.
 	 */
 	yearsCheck?: boolean;
 
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -148,6 +148,13 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	years?: IYears;
 
 	/**
+	 * Flag to indicate whether the DatePicker should validate whether the year
+	 * entered in the input is included in the range of years. Disable only
+	 * if your range is dynamic.
+	 */
+	yearsCheck?: boolean;
+
+	/**
 	 * Determines if menu is expanded or not
 	 */
 	expanded?: boolean;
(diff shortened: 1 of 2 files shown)


packages/clay-date-picker/src/index.tsx
```diff
@@ -147,6 +147,13 @@ interface IProps extends React.HTMLAttributes<HTMLInputElement> {
 	 */
 	years?: IYears;
 
+	/**
+	 * Flag to indicate whether the DatePicker should validate whether the year
+	 * entered in the input is included in the range of years. Disable only
+	 * if your range is dynamic.
+	 */
+	yearsCheck?: boolean;
```
julien, line 155:
```suggestion
/**
 * Flag to indicate whether the DatePicker should validate if the year
 * is included in the range of years. Disable only if your range is dynamic.
 */
```

julien: Github's suggestion feature seems to put stuff where it want to :rofl:

julien (approved):
Code changes, LGTM. Just added a suggestion for the comment

julien (approved):
LGTM! Thanks @matuzalemsteles

julien:
(I'll still wait until the CI finishes to merge)

PR #4677, opened 2022-02-17
feat(@clayui/localized-input): turn label and resultFormatter into optional properties

miltonmc:
The goal of this changes is to give more flexibility for developers in order to display or not the label or formatted result.

### Some context
I'm currently working on Objects, the only app that's already using this component into DXP, and I've found out that they are hiding the label in order to add a new one supporting the required mark (*).
https://github.com/liferay/liferay-portal/blob/0876c4ff2e20691df829d192039047f26cf63508/modules/apps/object/object-web/src/main/resources/META-INF/resources/js/components/Form/InputLocalized/InputLocalized.tsx#L51-L60

We are also using the taglib `<aui:input />` that does the same as this component, but to port this code to React I need to hide the formated result.
https://github.com/liferay/liferay-portal/blob/0876c4ff2e20691df829d192039047f26cf63508/modules/apps/object/object-web/src/main/resources/META-INF/resources/object_definitions/edit_object_field.jsp#L38

IMHO the proper place to do that is here and once we don't have already other modules relying on this component I've set those properties as optional.

### DIFF
--- a/packages/clay-localized-input/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-localized-input/src/__tests__/__snapshots__/index.tsx.snap
@@ -5,11 +5,6 @@ exports[`ClayLocalizedInput renders 1`] = `
   <div
     class="form-group"
   >
-    <label
-      for="locale1"
-    >
-      Check for translations
-    </label>
     <div
       class="input-group"
     >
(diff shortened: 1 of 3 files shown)


matuzalemsteles (approved):
Makes sense to me! @miltonmc we are scheduling this component to be deprecated soon in favor of implementing Translator Manager in DXP https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-js/frontend-js-components-web/src/main/resources/META-INF/resources/translation_manager, I think you can use it instead. @julien do you know if this is possible, using the React component directly?

julien:
converting to draft to avoid an accidental merge

matuzalemsteles:
I'm closing this as we are going to mark this component as deprecated soon.

PR #4681, opened 2022-02-17
fix(@clayui/core): fix error when not updating parent's loc

### LINKED ISSUE #4680: TreeView with expanded items on the first render does not select item children when the item is not expanded
This is a behavior that @julien found when using the combination of rendering the TreeView with expanded items initially and `selectionMode` with `multiple-recursive`. To reproduce this component you need to close the item that was initially expanded and click on the checkbox to select the item and its children, it will not select the children, only when it is visible.

The reason this happens is that it is related to the mount of `layoutKeys`, that is, how it is built as it renders the components, React calls `useEffect` in a different order when you have nested components, we use `useEffect` to call `createPartialLayoutItem` to add information to `layoutKeys` to have the component mount and unmount workflow automatically manage the `layoutKeys` structure.

For example, when the first render happens if there are no nested components or no expanded items, React calls `useEffect` in the same order they are rendered, ie a regular list. When you have nested components your `useEffect` call order is from bottom to top instead of sequential from top to bottom like in rendering. In `createPartialLayoutItem` to populate the `layoutKeys` we use a linked list of the item's parent and its children, as it is called from bottom to top, it doesn't find its parent in the structure, so we add the parent but with initial values that is a pre-initializes the data in the structure, when the parent of this item is called it will update with its information.

The relationship with `layoutKeys` is that it contains the `loc` which are the indexes where we can find the data of this key in the tree without having to traverse the entire tree, when the pre-initialization of the parent in `layoutKeys` happens we are adding to ` loc` is empty and we don't update this information afterward. Recursive multiple selection relies on this information to know where the item is in the tree and select its children recursively.

matuzalemsteles:
Fixes #4680

Well, the fix is simple but the explanation is contained in the issue.

### DIFF
--- a/packages/clay-core/src/tree-view/useMultipleSelection.tsx
+++ b/packages/clay-core/src/tree-view/useMultipleSelection.tsx
@@ -113,7 +113,7 @@ export function useMultipleSelection<T>(
 					layoutKeys.current.set(parentKey, {
 						children: new Set([key]),
 						lazyChild: false,
-						loc: [],
+						loc: loc.slice(0, -1),
 						parentKey: undefined,
 					});
 				}


julien:
> Well, the fix is simple but the explanation is contained in the issue.

Yes, thanks for the explanation @matuzalemsteles. I'll have another "round of debugging" to make sure I understand this correctly, although the explanation you left in the issue already helped.

julien (approved):
LGTM!

julien:
Thanks for the quick fix @matuzalemsteles

Issue #4676, opened 2022-02-16
Not possible to set disabled state to a tree view item


victorg1991:
Right now, (at least from what I have seen, I may be mistaken) it is not possible to set the disabled state for a tree view item. It would be nice to have a property in the `<ClayTreeView.ItemStack>` or  `<ClayTreeView.Item>` to set this state.

matuzalemsteles:
Hey @victorg1991 what use for `disabled` in this case would be just to disable Node and not its children, correct? or is it a case of disabling Node and its children? Currently in the implementation adding `disabled` also causes the component not to be expanded if it has children. @pat270 maybe now we need to add style to `disabled` in TreeView Node 😄

victorg1991:
Yeah, correct, ideally children should be enabled even if the parent is not. An use case may be selecting a folder to where a document should be moved, you may want to disable the current folder of the document, but not its children :)

victorg1991:
> Currently in the implementation adding disabled also causes the component not to be expanded if it has children. @pat270 maybe now we need to add style to disabled in TreeView Node 😄

I tried that without luck, maybe I wasn't used the latest clay css version, in which component did you add the disabled class?

victorg1991:
I guess in those case we also want to ignore the click event in the item?

matuzalemsteles:
Hey @victorg1991 this behavior only worked for Drag and Drop, you can pass the property it will "work" but it will not disable anything, we will have to add the CSS for this case, maybe we have to rethink because I think we didn't expect this case to still keep expanding even though the Node is disabled.

@drakonux maybe we need to add this use case in the documentation as well.

> I guess in those case we also want to ignore the click event in the item?

Probably when adding the disabled properties in the CSS will do it for us. But today you can do that.

```jsx
<TreeView.ItemStack
  onClick={(event) => item.disabled && event.preventDefault()}
>

<TreeView.Item
  onClick={(event) => item.disabled && event.preventDefault()}
>
```

In this case, it will depend on how your `selectionMode` is, for example in the single selection, it will ignore the selection and expand when you click on Node but clicking on the expand icon will continue to work.

When your selection is multiple or multiple-recursive, as it is via Checkbox, in this case above it will not be necessary because it will still select but you can also ignore it in two ways.

```jsx
<TreeView.ItemStack>
  <Checkbox
    disabled={item.disabled}
  />

  <Checkbox
    onChange={(event) => item.disabled && event.preventDefault()}
  />

  <Checkbox
[...]

PR #4682, opened 2022-02-17
fix(@clayui/css): Treeview adds disabled styles to `treeview-link`

pat270:
issue #4676 

@matuzalemsteles This should cover the disabled styles. We need to disable each clickable element inside a disabled link unfortunately. This mostly applies to a checkbox if it's present. The hover controls won't be shown if disabled.


pat270:
@matuzalemsteles This one ignores disabled styles for the expander. If you want disabled styles on the expander you need to add the `disabled` class on `component-expander`. I forgot to add examples on this pr. I'll do that.

matuzalemsteles (approved):
LGTM!

PR #4685, opened 2022-02-21
feat(@clayui/core): add API to disable item and expander disable optionally

### LINKED ISSUE #4676: Not possible to set disabled state to a tree view item

matuzalemsteles:
Fixes #4676

This PR is adding the two behaviors of disabling the item, disabling all clickable elements inside the item, or making the Expander clickable. By default adding `disabled` to `<Item />` and `<ItemStack />` disables all elements, to always make the expander clickable adds the `expanderDisabled` property to `<ItemStack />`.

I added this information to the documentation as well. I tried to avoid adding an auxiliary prop but without that, we couldn't support both cases by doing inference for example.

### DIFF
--- a/clayui.com/src/components/Typography/Typography.js
+++ b/clayui.com/src/components/Typography/Typography.js
@@ -6,35 +6,38 @@
 import React from 'react';
 
 const getId = (children) => {
-	return (
-		typeof children === 'string' &&
-		children.toLowerCase().split(' ').join('-')
-	);
+	if (typeof children !== 'string') {
+		return;
+	}
+
+	if (children.match(/\(#/)) {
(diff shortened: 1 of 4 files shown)


julien (approved):
Hey @matuzalemsteles LGTM. I made a few edits to the documentation and doc comments, please have a look when you get a chance.

matuzalemsteles:
Thanks @julien for that!

PR #4686, opened 2022-02-22
fix(@clayui/css): Updates `content` property values to conform to wha…

### LINKED ISSUE #4683: Accessibility change request: avoid adding content from CSS
I can't provide an example via codesandbox.com

In order to reproduce, you need to run the accessibility tool described on [LPS-133314](https://issues.liferay.com/browse/LPS-133314) ([OAW](https://administracionelectronica.gob.es/comunidades/verPestanaMas.htm?idComunidad=accesibilidad#.YKZABKj7SHv), owned by the Government of Spain, used by m/any public institutions).

The tool throws some errors in `clay.css` when using `content` + pseudoelements `:after`/`:before`:

| Attribute(s)  | File |
| ----- | ------- |
| `content: "Browse"`  | `clay-css/src/scss/bootstrap/_custom-forms.scss`  |
| `content: attr(data-browse)` |  `clay-css/src/scss/bootstrap/_custom-forms.scss`|
| `content: attr(data-label-min)` / `data-label-max` | `clay-css/src/scss/variables/_range.scss` |
|`content: attr(data-multi-step-icon)` | `clay-css/src/scss/components/_multi-step-nav.scss` |
| `content: attr(data-label-on)` / `data-label-off` | `clay-css/src/scss/components/_toggle-switch.scss` |
|  `content: normal` | `clay-css/src/scss/components/_multi-step-nav.scss` |

**Steps**:

1. Start a DXP Bundle
2. Search for the aforementioned attributes in http://localhost:8080/o/classic-theme/css/clay.css

No content is added from CSS. The accessibility tool does not complain about it.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v1  |
| Clay  | v2  |
| Clay  | v3  |

pat270:
…t is allowed by OAW Tracker

    - Changes `content: normal` to `content: none`
    - Multi Step Nav adds variable `$data-multi-step-icon-before-content`
    - Custom Forms adds variables `$custom-file-label-after-content` and `$custom-file-text-data-browse`
    - Toggle Switch adds variables `$toggle-switch-handle-after-content` and `$toggle-switch-handle-on-after-content` for backward compatibility with 2.x
    - https://issues.liferay.com/browse/LPS-147457

fixes #4683


packages/clay-css/src/scss/components/_multi-step-nav.scss
```diff
@@ -62,7 +62,9 @@
 			@if ($multi-step-icon-active-bg-image) {
 				&[data-multi-step-icon] {
 					&:before {
-						content: normal;
+						// https://issues.liferay.com/browse/LPS-147457
+
+						content: none;
```
matuzalemsteles, line 67: Hey @pat270 just out of curiosity, what effect does this have on `:before`? Is this related to some other value?

matuzalemsteles (commented):
Hey @pat270 from what I understand here we will keep the `content` but let the developers change it via the sass variable.

Seems like a good solution for me to go with this, but I think we need to add this information somewhere so people know what to do when they encounter this issue.

packages/clay-css/src/scss/components/_multi-step-nav.scss
pat270: @matuzalemsteles this is exactly the same as `content: normal`. `content: none` gets a pass on the OAW Tracker for some reason 🤔.

From https://developer.mozilla.org/en-US/docs/Web/CSS/content:
> normal
Computes to none for the ::before and ::after pseudo-elements.

For

> we will keep the content but let the developers change it via the sass variable.

This is correct, but I think the person who reported this issue won't have to change anything after these get merged.

matuzalemsteles: > this is exactly the same as content: normal. content: none gets a pass on the OAW Tracker for some reason 🤔.

Yeah I also found it strange because it seems that the behavior is the same, maybe it has some other effect that we don't know 🤷‍♂️

pat270: Thanks @matuzalemsteles, I'm going to make some minor adjustments. Let's just say the tool isn't perfect.

Issue #4580, opened 2022-01-17
@clayui/color-picker should persist current color when opening gradient selector.


ethib137:
https://storybook.clayui.com/?path=/story/components-claycolorpicker--custom-colors

# Steps to Reproduce:

1. Navigate to https://storybook.clayui.com/?path=/story/components-claycolorpicker--custom-colors
2. Paste or type a new color into the input. 
3. Click on the color swatch to open the picker.
4. Click the color droplet to open the gradient selector.

# Expected Result:

- The gradient selector displays the color added in step 2.

# Actual Result:

- The gradient selector displays the first custom color.

For reference here are 2 gifs. One shows Clay's behavior the other show the Chrome dev tools color picker behavior.

![claycolorpicker](https://user-images.githubusercontent.com/1609196/149730544-c9bc4561-a9fa-4e40-939e-7f88412d4179.gif)
![chromecolorpicker](https://user-images.githubusercontent.com/1609196/149730523-4bd21c6b-80e7-431d-a866-90b11829c5af.gif)

ethib137:
Hey @matuzalemsteles thanks for your response... I really think this behavior is incorrect for custom colors as well, though. In fact, I think only changing the behavior when there are not custom color palettes will add additional inconsistencies between the color picker variants.

> The color typed in the field is never saved among custom colors. The color must be only reflected in the color square.

This should remain true, but can be handled by deselecting a custom color when the color picker input changes. I think this should always be the case that if you change the color picker input value, it should deselect the custom color. Otherwise you end up with strange inconsistent views like this. Also, where is this documentation? I couldn't find this line [here](https://liferay.design/lexicon/core-components/forms/picker-color/#gatsby-focus-wrapper)

![inconsistentInputwithSelectedCustomColor](https://user-images.githubusercontent.com/1609196/153138265-19773a91-04b5-4cf2-84e5-c3a8ffd6ba3d.png)

When you find yourself in the situation pictured above, you can still click on the  already selected custom color in order to apply it to the input, but I think this is unhelpful behavior. It would make more sense if the custom color was not selected, thus implying it is clickable.

Also, even when chrome dev tools includes custom colors, they still always maintain consistency between the current value and the value reflected in the color picker overlay. 

![ChromeColorPickerCustomColors](https://user-images.githubusercontent.com/1609196/153139618-967201fb-541d-476e-9f0f-09e3185ce038.gif)

[...]

Issue #4134, opened 2021-06-16
@clayui/color-picker Should accept 8 digit hexcodes and opacity for the color property


ethib137:
Currently the color picker is unable to properly handle 8 digit hexcodes and opacity. You can see this by completing the following steps.

- Open the code sandbox: https://codesandbox.io/s/ecstatic-feather-rwhw6?file=/src/index.js
- Click on the swatch to open the color picker.
- Click on the droplet to open the advanced picker.
- Notice that the color is displayed as a 6 digit hex code and does not allow for opacity.

Given the current browser support, we should start supporting 8 digit hexcodes and opacity. It may be worth considering 

https://caniuse.com/css-rrggbbaa

matuzalemsteles:
Thanks @ethib137 for bringing this, I think this is super valid and it makes sense to bring it to DatePicker, apparently, the lib we use internally supports this so it wouldn't be a problem to implement or just tweak to support it, but maybe to support `rgba` at the level of Improve user experience in UI than having to type everything in the input, as a native picker in the devtools that you can switch between different formats.

cc @drakonux I think this might be interesting, I think we can support it at the input level but visually I'm not sure if we want to improve the experience here, in the end, everything typed is converted to HEX.

brunofernandezg:
cc @liferay-lexicon

hold-shift:
Hello @ethib137. After studying the current component behavior, we have seen that the input accepts 2 extra digits for opacity, but a UI solution for the panel is necessary. I'm about to open a Lexicon ticket to work on this. We only need a deadline to assign it a priority within our backlog. Thanks!

ethib137:
Hi @hold-shift, I can't provide a deadline, since from my side it's only a nice to have, not a requirement. For determining priority, one thing to consider is the use of the color-picker in Style Books. Style Books is becoming more and more important as we move away from creating custom themes, and having a color picker that properly handles 8 digit hexcodes in the panel will improve that experience. I hope that helps.

matuzalemsteles:
LPS Ticket https://liferay.atlassian.net/browse/LPS-190539

PR #4691, opened 2022-02-23
fix(@clayui/color-picker): Current colour persist and Hue component synchronization

sergiojimcos:
**What is new?**
There are some changes:
-  It saves the current colour in the **Custom Colour Palette**. 
- **Hue** synchronization with the current color and the rest of components.
- **Input** elements in upper case.

## WITH PALETTE

https://user-images.githubusercontent.com/44723449/155372894-1d37400b-86ed-440e-bcf5-bd3a0a3fbfc0.mov

## WITHOUT PALETTE

https://user-images.githubusercontent.com/44723449/155372964-3ad1e707-b7c5-429e-8696-24bc3dfa8abd.mov

**FIXES**
#4580

### DIFF
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -115,6 +115,8 @@ const RGBInput: React.FunctionComponent<IRGBInputProps> = ({
 };
 
 interface IProps {
+	active?: boolean;
+
 	/**
 	 * List of colors that will display as a color splotch
 	 * these can be either hex values, color names or
@@ -160,6 +162,7 @@ const DEFAULT_SPLOTCH_COLOR = 'FFFFFF';
  * Renders the custom color picker
  */
 const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
+	active,
 	colors,
 	editorActive,
 	label,
@@ -247,19 +250,22 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 			setHue(color.toHsv().h);
 			setHexInputValue(color.toHex());
 
-			const index = colors.indexOf(value as string);
-
-			if (index !== -1) {
-				setColors(
-					value!.includes('var(') ? value! : color.toHex(),
-					index
+			if (!active) {
+				const colorIndex = findColorIndex(
+					colors,
+					value!.includes('var(')
+						? getCSSVariableColor(value!)
+						: tinycolor(value)
 				);
-				setActiveSplotchIndex(index);
+
+				if (colorIndex === -1) {
+					setActiveSplotchIndex(undefined);
+				}
 			} else {
-				setActiveSplotchIndex(undefined);
+				setColors(color.toHex());
 			}
 		}
-	}, [value, color]);
+	}, [active, value, color]);
 
 	return (
 		<div ref={customMenuRef}>
@@ -297,11 +303,16 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 										setActiveSplotchIndex(index);
 									}
 
+									// The hexadecimal color `#FFFFFF` is treated as an empty
+									// slot so when the user enters a color that doesn't exist in
+									// the custom, clicking on an empty slot will replace that
+									// slot with the new color if don't have an active slot
+									// being edited.
 									if (hex === DEFAULT_SPLOTCH_COLOR) {
-										const newColor = tinycolor(hex);
-
 										setInternalEditorActive(true);
 
+										// Replaces the slot color with the color entered in the
+										// input if it does not have an active slot being edited.
 										if (
 											previousColorRef.current !==
 												tinycolor(
@@ -315,14 +326,9 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 											setColors(color.toHex(), index);
 											setHexInputValue(color.toHex());
 											setHue(color.toHsv().h);
-										} else if (
-											findColorIndex(colors, newColor) ===
-												-1 ||
-											tinycolor.equals(
-												newColor,
-												tinycolor(DEFAULT_SPLOTCH_COLOR)
-											)
-										) {
+										} else {
+											const newColor = tinycolor(hex);
+
 											setNewColor(newColor, true, index);
 											setHue(newColor.toHsv().h);
 										}
@@ -425,8 +431,10 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 											if (newColor.isValid()) {
 												setHue(newColor.toHsv().h);
 
-												const hasColor =
-													colors.indexOf(newHexValue);
+												const hasColor = findColorIndex(
+													colors,
+													newColor
+												);
 
 												setNewColor(
 													newColor,
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -24,6 +24,20 @@ function findColor(colors: Array<string>, color: tinycolor.Instance): boolean {
 	return false;
 }
 
+function getCSSVariableColor(value: string) {
+	const element = document.createElement('div');
+
+	element.setAttribute('style', `background: ${value};`);
+
+	document.body.appendChild(element);
+
+	const color = tinycolor(getComputedStyle(element).backgroundColor);
+
+	document.body.removeChild(element);
+
+	return color;
+}
+
 interface IRGBInputProps {
 	/**
 	 * Callback function for when the input value is changed
@@ -158,14 +172,29 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 	value,
 }) => {
 	const inputRef = React.useRef(null);
-	const [activeSplotchIndex, setActiveSplotchIndex] = React.useState(0);
+
+	const customMenuRef = React.useRef<HTMLDivElement>(null);
+
+	const [activeSplotchIndex, setActiveSplotchIndex] = React.useState(() => {
+		const index = colors.indexOf(value as string);
+
+		return index !== -1 ? index : undefined;
+	});
+
 	const [internalEditorActive, setInternalEditorActive] = useInternalState({
 		initialValue: !showPalette,
 		onChange: onEditorActiveChange,
 		value: editorActive,
 	});
 
-	const color = tinycolor(value);
+	const color = React.useMemo(
+		() =>
+			value!.includes('var(')
+				? getCSSVariableColor(value!)
+				: tinycolor(value),
+		[value]
+	);
+
 	const [previousColor, setPrevousColor] = React.useState(color);
 
 	const [hue, setHue] = React.useState(color.toHsv().h);
@@ -180,6 +209,16 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 		[b, 'b'],
 	];
 
+	const setColors = (color: string, index?: number) => {
+		const newColors = [...colors];
+
+		index
+			? (newColors[index] = color)
+			: (newColors[activeSplotchIndex as number] = color);
+
+		onColorsChange(newColors);
+	};
+
 	const setNewColor = (
 		colorValue: tinycolor.Instance,
 		setInput = true,
@@ -187,14 +226,7 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 	) => {
 		const hexString = colorValue.toHex();
 
-		const newColors = [...colors];
-
-		index
-			? (newColors[index] = hexString)
-			: (newColors[activeSplotchIndex] = hexString);
-
-		onColorsChange(newColors);
-
+		setColors(hexString, index);
 		onChange(hexString);
 
 		if (setInput) {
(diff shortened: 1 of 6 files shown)


packages/clay-color-picker/src/Custom.tsx
```diff
@@ -177,19 +177,23 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
...
-
-			if (!showPalette) {
+			if (color.isValid()) {
 				setHue(color.toHsv().h);
+
+				if (!colors.includes(color.toHex())) {
+					setNewColor(color);
+				} else {
+					setHexInput(color.toHex());
+				}
```
matuzalemsteles, line 193: It seems that this is having a side effect when interacting with the Hue component and the Gradient selector when moving to a black color and in some situations the Hue is also changing color when moving through the Gradient.

See this video:

https://user-images.githubusercontent.com/13750819/155434784-ceb13a73-785e-4a16-a078-e1457e990f1f.mp4

matuzalemsteles (commented):
Hey @sergiojimcos I think we are deviating from the expected behavior here, apparently when color is added to the input that is not a color that exists when opening the panel the color should only be reflected in the editor but not modify the custom color but if the user clicking on an empty color in the custom color will replace it with the color added by the user.

So if color is manually added in the input that doesn't exist in the custom color, when opening the panel it should not focus/active on a custom color.

See this comment by Emiliano https://github.com/liferay/clay/issues/4580#issuecomment-1034734204. Let me know if you need help!

matuzalemsteles (commented):
I've noticed some inconsistent behavior for some scenarios, but I think we're getting there.

- When typing on the input the Panel is being closed it must remain open
- The input value I think we don't need to leave uppercase for all scenarios, for example `var(--blue)` this can break the implementation, also it's ok to type the color name in lowercase when typing a lowercase hex color is also not a problem.
- Typing a color initially in the input and then opening the panel and clicking on an empty color is not replacing the color typed in the input

packages/clay-color-picker/src/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -1233,7 +1233,7 @@ exports[`Rendering renders w/ var() 1`] = `
             aria-label="Color selection is var(--blue)"
             class="form-control input-group-inset input-group-inset-before"
             type="text"
-            value="VAR(--BLUE)"
+            value="var(--blue)"
```
julien, line 1236: @sergiojimcos do you know why this needs to be updated? not saying it's wrong but I'm curious about the change.

packages/clay-color-picker/src/index.tsx
```diff
@@ -345,12 +345,12 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 								insetBefore
 								onChange={(event) => {
 									onValueChange(event.target.value);
-									setCustomEditorActive(false);
-									setInternalActive(false);
+									// setCustomEditorActive(false);
```
julien, line 348: @sergiojimcos looks like you forgot to "clean up" here

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -229,15 +257,28 @@ const ClayColorPickerCustom: React.FunctionComponent<IProps> = ({
 								onClick={() => {
 									if (hex === 'FFFFFF') {
 										setInternalEditorActive(true);
-									}
+										setActiveSplotchIndex(i);
```
julien, line 260: I'd rather rename `i` to `index` in this scope just to make it clearer

julien (commented):
Looks OK overall, I'm just going to fetch this to see it in an editor, GitHub's UI isn't the greatest thing to review

julien:
@sergiojimcos before submitting a pull request or updating it, you can always run the tests locally to make sure they won't fail

packages/clay-color-picker/src/__tests__/__snapshots__/index.tsx.snap
sergiojimcos: Sure @julien. At first I changed this snapshot when I modified the values of the input into upper case. As I told you yesterday, finally the value of the input must be in lower case, for that reason I changed into into the default one

packages/clay-color-picker/src/index.tsx
sergiojimcos: Oh sorry, I haven't seen that 😅

julien:
@sergiojimcos I've updated everything we did together during our pairing session in https://github.com/liferay/clay/pull/4691/commits/505253aaa0226765be8e988433fe8a77751784c9, let's see if we're luck with the tests :wink:

sergiojimcos:
> @sergiojimcos I've updated everything we did together during our pairing session in [505253a](https://github.com/liferay/clay/commit/505253aaa0226765be8e988433fe8a77751784c9), let's see if we're luck with the tests 😉

Great, many thanks Julien! ❤️

julien:
@matuzalemsteles I helped @sergiojimcos a bit with this and something isn't very clear:

Do we want to let users be able to use things like CSS variables (e.g. `'var(--blue)'`) and "invalid" colors, like `'inherit'`?

If that's the case then we probably need to think about this very carefully, because `tinycolor` doesn't consider those to be "valid" colors, (i.e. `tinycolor('inherit').isValid()` will return false, same with `var(--blue)`). Additionally we also have this [hook](https://github.com/liferay/clay/blob/5591614c1b88a799eab9cc3afbd96c986180faae/packages/clay-color-picker/src/hooks.ts#L38-L56), that not going to work with those values. 

What are your thoughts on this?

matuzalemsteles:
Yeah, it's intentional to support this, this came up with this issue #3681 as well to help support use in stylebooks. We'll probably have to improve our implementation of this to identify when it's using different syntaxes.

Probably for CSS Variables cases, we will have to get the hex just for internal use to show in the gradient and the hue. Apparently, we have the same with the `inherit` and named colors using something like `getComputedStyle(ref).backgroundColor` which returns the rgb, maybe we can try something with that, I would say don't do anything with this here, maybe just avoid uppercase for this scenario and create another issue to work on, we will probably have to rewrite some things I think.

Maybe we can accept any value typed in the input for now so we don't have to use `tinycolor('inherit').isValid()` for example.

julien:
@matuzalemsteles I know this is not "your decision" but  this "feature" doesn't really make any sense ... 

I also think that since @sergiojimcos is new on the team we should try to help him as much as we can when it comes to task definitions and goals - so I'll try to help him on this one.

matuzalemsteles:
> Do you know if this request come from Lexicon or Dialect?

I'm not sure but tracing the conversations, this stemmed from the need for stylebooks.

> It's possible to get the color value when the color is set to inherit, but we'll need to modify the component to add an "alpha input", because for the moment we only support RGB (and not RGBA)

We have a request from Dialect to support RGBA #4134 but we are depending on the Lexicon team because it will need changes to the Panel to support this.

Maybe what we can do here is convert rgba to rgb, I think the `tinycolor` lib can help us with that. It also looks like it can't always come rgba from `inherit` it will depend on the parent declaration.

> We need to get rid of the [useHexInput](https://github.com/liferay/clay/blob/c62913222b959af3ce13f0e00cd10a8406ab4a33/packages/clay-color-picker/src/hooks.ts#L38-L56) hook because it's too restrictive for our this case

Hmm, the problem is that the panel's input works with hex, maybe if it reflects the same thing that is in the input it can make sense to remove it.

> Last but not least, since also need to be able to work with CSS variables, we probably want to add a few CSS variables to the stories in order to check if we can use something like [this](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#values_in_javascript) to retrieve the correct values

Yeah, I think there's a case of `var(--blue)` in the custom color storybook example.

julien:
> I'm not sure but tracing the conversations, this stemmed from the need for stylebooks.

Yes, that's also what I imagined. 

> Maybe what we can do here is convert rgba to rgb, I think the tinycolor lib can help us with that. It also looks like it can't always come rgba from inherit it will depend on the parent declaration.

Yes, we can try that although I'm not convinced it will work with all use cases. 
I agree that sometimes `inherit` won't always be an RGBA value, but sometimes it will and it also seems that we want to support "anything"

> Hmm, the problem is that the panel's input works with hex, maybe if it reflects the same thing that is in the input it can make sense to remove it.

I understand that we want the "main input" to accept either RGB, RGBA, CSS variables (and eventually color "names", like red, blue, green, etc...) and that the input in the panel should show those values converted to hex; at the same time we also want to support RGBA at one point, and that hook won't let us do that. Additionally since we already have `tinycolor` that can validate the color, having a hook based on regular expressions doesn't seem necessary ... We should have one way of validating the color, not two.

julien:
@matuzalemsteles as far as I can see, there's no way to convert from RGBA to RGB with `tinycolor` - and since we're still waiting for the design changes I guess it's better to put this on hold (although the bug was fixed)

matuzalemsteles:
> Additionally since we already have tinycolor that can validate the color, having a hook based on regular expressions doesn't seem necessary ... We should have one way of validating the color, not two.

Yeah, that makes sense to me too, I think this was created just for a way of hex validation, but could have used `tinycolor` for that purpose. Well, I think we can remove that and just use `tinycolor`.

> as far as I can see, there's no way to convert from RGBA to RGB with tinycolor - and since we're still waiting for the design changes I guess it's better to put this on hold (although the bug was fixed)

yeah, it makes sense he will probably put an 8 digit hex that we don't support yet 😔.

sergiojimcos:
Hi @matuzalemsteles! As @julien explains before, I have removed `useHexInput` hook and use `tinycolor` instead to avoid comparing using the regular expression. According to 8 digit hex, I think that we can see later after resolving this issue and investigate how can we support RGBA in color-picker once it works well. Yo can review this.

matuzalemsteles:
Hey @sergiojimcos I'm reviewing this, tomorrow I'll add some interactions on top of this PR to cover some edge cases, no need to worry.

matuzalemsteles:
Hey guys, I pushed some modifications in an attempt to simplify some scenarios and also cover the use cases that was requested by Emiliano. So in these changes we have the following changes:

- Typing a different color in the input than the one in the custom palette and opening the Panel will not overwrite any color until you click on an empty slot
- Fix the problem of typing in the custom input was not modifying the color
- When opening the panel with a color different from the custom color, it is reflected in the editor and it is also possible to change the color without having to save it in the palette
- Adds support for CSS Variables, color is converted to 6-digit Hex in editor so user can view or change

About the case of `inherit` I was thinking about it and it doesn't make much sense we support it in the editor and the slot too, despite the possibility to get the color of `inherit` it doesn't make sense because the `inherit` it will be used in another place that we have no way to infer what the color of the parent will be where it is being used, for example, I think that in this case, the product must have a different behavior for that.

julien:
Hey @matuzalemsteles I think these changes are *almost* OK, I do agree that supporting `inherit` doesn't make much sense as it can be complex to handle - to clarify, I used *almost* because everything work except for when clicking on the white (`FFFFFF`) squares - for some reason the "main" input doesn't get updated. Appart from that LGTM

matuzalemsteles:
Hey @julien could you say what exactly this behavior would be? are you meaning this behavior? Apparently, it's working as expected, I may be missing some other use case.

https://user-images.githubusercontent.com/13750819/157959367-ee98576d-b698-4ecc-b9a6-4b3aa40205a1.mp4

sergiojimcos:
Hey @matuzalemsteles! What @julien means in this scenario is that when we click on a `FFFFFF` splotch, both inputs must be updated. Apart from that, everything is great, LGTM! 

https://user-images.githubusercontent.com/44723449/158130293-baaad763-222e-4e71-b0f7-cfac5033cf83.mov

matuzalemsteles:
Hey guys, I've fixed this use case, let me know if I'm missing anything else. I will also create an issue so we can add more test cases for ColorPicker.

sergiojimcos:
Hey @matuzalemsteles! I've updated the branch and this (https://github.com/liferay/clay/pull/4691#issuecomment-1066486933) is solved and works well.

However, I've been tested several use cases to check that color picker finally works well and it doesn't. There are two use cases that are related:

1. In this case, I think that color picker must store only one value of a particular color, because it doesn't make sense having several instance of the same color. I don't remember well but this particular example was discussed. 

https://user-images.githubusercontent.com/44723449/158224317-702e4796-1dc5-47ce-941e-a95a616564ab.mov

2. On the other hand, it happens only with CSS variable. I think that in this case, the color picker doesn't check the hex values of the CSS variables, because the second value is stored as hex value

https://user-images.githubusercontent.com/44723449/158225248-09f68863-42e3-4ad1-a7ef-8712dd42f83b.mov

matuzalemsteles:
hey @sergiojimcos Thanks for the investigations, I'll work on it! It makes sense to constrain the splotch to not have repeating colors.

matuzalemsteles:
I pushed more changes in this PR to fix the above scenarios, I did several tests too to check if it's working well, but I believe I may have missed something, I also tried to simplify the implementation more and add some comments to help in understanding the edge cases.

I believe that a lot of these problems are a little related to the way the components are structured, Custom is depending a lot on the `value` to define the values, so it's more of a state management problem that is hurting a lot here to define the other states. At another time we can simplify and refactor this component to have better management of the internal states to simplify the final implementation.

julien (approved):

julien:
@matuzalemsteles LGTM thanks. @sergiojimcos could you have a second look at this before we merge?

sergiojimcos:
LGTM! thanks @matuzalemsteles!

Issue #3683, opened 2020-09-03
Consider steps to automate release and portal updating process


bryceosterhaus:
At some point I think it would be helpful if we had some sort of automated process for releasing clay and send a PR to update portal.

Things to consider:
- Knowing when we need to update implementations of clay, we can't really automate this.
- Versioning, is it easier to automate if we go back to "locked" mode in lerna?
- Should this be a local command we run or a gh action?

Anyways, I have been randomly thinking about this and thought I would create an issue to remind us to investigate.

matuzalemsteles:
It would be great! I've been thinking the same thing. I was thinking that we could try to do something using Github Actions but I think they could only be triggered via branch.

jbalsas:
Having something like this would make releasing and integrating probably easier... we could even somehow do nightly PRs just to check that whatever's been merged (even if not released) is working fine or detect issues early before we release and attempt to integrate.

wincent:
> * Should this be a local command we run or a gh action?

My suspicion is that anything related to liferay-portal isn't really suitable for a GitHub action, because you need a copy of liferay-portal in order to do things like create a PR for it, and the repo is too large for that to be feasible (I ran into this trying to set up a GitHub action to auto-update the liferay-frontend/liferay-portal repo; it runs in a VM which is too slow to fetch the repo, and caching doesn't save you...).

Working on the local command isn't a waste of time though. Whatever we can do to make releases "turn-key" will benefit us and in the future if something based on a GitHub action does become feasible, then it's likely our work with the manual process carries over to benefit the automated one too.

Maybe we can do something from the liferay-portal side. Instead of cutting nightly snapshot releases of Clay packages, we could use `yarn add <git remote url>#<branch/commit/tag>` as described [here](https://classic.yarnpkg.com/en/docs/cli/add) to point at a specific commit. Might require us to actually commit build artifacts though, or maybe we can use a `postinstall` script to run the build at install time. Or wrap it all up in a script that we can run from a liferay-portal checkout; it would be a single command (and waiting a few minutes) to get you to the point where you could test `HEAD` of `master` in DXP.

bryceosterhaus:
> we could even somehow do nightly PRs just to check that whatever's been merged (even if not released) is working fine or detect issues early before we release and attempt to integrate.

I really like this idea, whether it's an actual PR or something like @wincent described. This would at least give us a consistent idea of the state of clay in DXP and how difficult an update might be,

[...]

PR #4692, opened 2022-02-23
chore: adds workflow to generate automated release

matuzalemsteles:
Related #3683

I spent a little bit of the day today trying to bring our manual Clay releases workflow into an automated workflow with Github Actions.

This implementation has some implications that not being able to reproduce the `git push $REMOTE master:stable` step was failing in the CI, I think if I do more tests I can get it working too.

Some considerations, this workflow is [triggered manually via the Github CLI or via the UI in the Actions tab](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow), so whenever we need to cut a new release we just run the workflow manually and it will take care of everything, run tests, cut package versions using semantic version, add the changelog and publish to NPM.

> Apparently, we don't have access to configure secrets so we will have to open a ticket to add the `NPM_TOKEN` token.

I made tests in my fork not to pollute the repository, see an Action in action https://github.com/matuzalemsteles/clay/actions/runs/1890172564. Also, the running average is about 9m to 10m, I didn't test the publish step because I didn't really want to publish the packages in NPM.

Leave your concerns and feel free to suggest any improvement.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,8 +14,10 @@ jobs:
             - uses: actions/setup-node@v2
               with:
                 node-version: '16'
-            - name: Clean yarn cache
-              run: yarn cache clean
+            - uses: actions/cache@v2
+              with:
+                path: '**/node_modules'
+                key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
             - name: Install Dependencies
               run: yarn install
             - name: Lint
@@ -24,8 +26,6 @@ jobs:
               run: yarn format:check
             - name: Verify Dependencies
               run: yarn checkDeps
-            - name: Clear jest cache
-              run: yarn test --clearCache
             - name: Run Jest
               env:
                   BROWSERSLIST_CONFIG: './packages/browserslist-config-clay/config'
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
This PR doesn't deal with automating the step of updating the packages in DXP, that would also be interesting but we would probably have work to make it work because the DXP is very big and it would take a long time to run in CI and maybe we have limitations but we can test other strategies as discussed in #3683.

julien:
> This PR doesn't deal with automating the step of updating the packages in DXP, that would also be interesting but we would probably have work to make it work because the DXP is very big and it would take a long time to run in CI and maybe we have limitations but we can test other strategies as discussed in https://github.com/liferay/clay/issues/3683.

Totally agree ... I think this is already a good first step

matuzalemsteles:
Well, I made a few more changes that would break the package publishing phase and added the step of pushing changes from `master` to the `stable` branch. Basically, have all the publishing steps that we do manually on the Cay side are automated.

To publish packages in NPM, we need to add an environment variable to secret in the Clay repository. I don't have access to do this but I'll check who I should ask to add so we can start testing and using it frequently in future releases.

PR #4697, opened 2022-02-25
fix(@clayui/pagination): The current page button must be disabled to avoid screen readers to read it as this is confusing for the final user

jorgediaz-lr:
This issue is linked to https://github.com/liferay-frontend/liferay-portal/pull/1950 and https://issues.liferay.com/browse/LPS-148048

_This issue has been reported by a blind association that is a Liferay customer, for more information see: https://issues.liferay.com/browse/LPS-148048?focusedCommentId=2626115&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2626115_

The pagination button to the current page should be disabled to avoid screen readers read it, as this is confusing for the final users.

After a quick research with several search engines:

- Google: https://www.google.com/search?q=search
- Bing: https://www.bing.com/search?q=search
- Yahoo: https://search.yahoo.com/search?p=search
- Amazon: https://www.amazon.com/s?k=search

All of them are not linking the current page, so it seems we should also disable the button.

For more information see the discussion in PR https://github.com/liferay-frontend/liferay-portal/pull/1950

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
 		"lint:fix": "eslint --fix \".*.js\" \"**/*.{js,ts,tsx}\"",
 		"lint:quiet": "eslint --quiet \".*.js\" \"**/*.{js,ts,tsx}\"",
 		"lint": "eslint \".*.js\" \"**/*.{js,ts,tsx}\"",
-		"site": "cd next.clayui.com && yarn build && yarn serve",
+		"site": "cd clayui.com && yarn build && yarn serve",
 		"sizeLimit": "size-limit",
 		"storybook:build": "NODE_OPTIONS=--max_old_space_size=8192 build-storybook -c .storybook -o .storybook-static",
 		"storybook": "start-storybook -p 8888",
--- a/packages/clay-pagination-bar/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-pagination-bar/src/__tests__/__snapshots__/index.tsx.snap
@@ -58,6 +58,7 @@ exports[`ClayPaginationBar renders 1`] = `
       >
         <button
           class="page-link"
+          disabled=""
           type="button"
         >
           1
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
 		"lint:fix": "eslint --fix \".*.js\" \"**/*.{js,ts,tsx}\"",
 		"lint:quiet": "eslint --quiet \".*.js\" \"**/*.{js,ts,tsx}\"",
 		"lint": "eslint \".*.js\" \"**/*.{js,ts,tsx}\"",
-		"site": "cd next.clayui.com && yarn build && yarn serve",
+		"site": "cd clayui.com && yarn build && yarn serve",
```
julien, line 35: Thanks for fixing this although it has nothing to do with the issue.

julien:
@jorgediaz-lr thanks for submitting this fix.
It seems that the snapshots [haven't been updated](https://github.com/liferay/clay/runs/5331077254?check_suite_focus=true), the code changes look good to me.

jorgediaz-lr:
Hi @julien 

My fault, I executed the yarn test -u inside the pagination folder.

I have executed it again from the root folder and I have added a new commit, see https://github.com/liferay/clay/pull/4697/commits/4367010721180d2a9b5afaa44d88ad72947ebf5f

package.json
jorgediaz-lr: I used it to test the issue following Georgel Pop instructions and it seems the path was wrong.

julien: Yes, exactly. Good catch

julien:
@jorgediaz-lr no problem, I just wanted to give you a heads up. As I wrote above, I'm OK with these changes, but I'll let @matuzalemsteles have a look too just in case (should be merged pretty soon if the sees no issues)

matuzalemsteles:
I just pushed a commit to fix the accessibility issue too for the cases when it's links, thanks for submitting this @jorgediaz-lr!

matuzalemsteles (approved):
LGTM

PR #4698, opened 2022-02-25
fix(lexicon-ux): Update run-sequence it causes publishing to maven to…

pat270:
… fail


pat270:
Going to ninja merge this. This is updating run-sequence to the correct version so the maven-publish task doesn't fail.

PR #4714, opened 2022-03-02
feat(@clayui/css): Adds placeholder comments that can be replaced wit…

### LINKED ISSUE #4713: @clayui/css: Add placeholder comment for `@use 'sass:math' as *`
We should add a placeholder comment that can be replaced with `@use 'sass:math' as *;` for builds that use Dart Sass. It would be similar to [this](https://github.com/liferay/clay/blob/c62913222b959af3ce13f0e00cd10a8406ab4a33/packages/clay-css/src/scss/atlas.scss#L5).

https://issues.liferay.com/browse/LPS-148467

pat270:
…h Sass @use rules

fixes #4713

Edit: ~This one is a little weird, not from our end but from Portal's end. According to the `yarn.lock` file, we're either on Sass [1.29.0](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/yarn.lock#L1716) or [1.30.0](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/yarn.lock#L10068). The slash as division deprecation wasn't implemented until Sass [1.33.0](https://sass-lang.com/documentation/breaking-changes/slash-div).~

Nevermind, I'm going to leave this for future reference. The Sass math module was introduced in Sass [1.23.0](https://sass-lang.com/documentation/modules/math), so we are still good. I was just wondering why I wasn't seeing deprecation warnings when I deployed in Portal.

### DIFF
--- a/packages/clay-css/src/scss/_cadmin-variables.scss
+++ b/packages/clay-css/src/scss/_cadmin-variables.scss
@@ -1,3 +1,5 @@
+// INSERT @use rules
+
 @import '_license-text';
 
 @import 'functions/_global-functions';
--- a/packages/clay-css/src/scss/atlas-variables.scss
+++ b/packages/clay-css/src/scss/atlas-variables.scss
@@ -1,3 +1,5 @@
+// INSERT @use rules
+
 @import '_license-text';
 
 @import 'functions/_global-functions';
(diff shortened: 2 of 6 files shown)


pat270:
@julien no not yet, we need to update build.gradle in unstyled to replace these lines.

julien:
@pat270 thanks for the clarification

matuzalemsteles (approved):
LGTM

PR #4716, opened 2022-03-03
fix(@clayui/css): Forms .form-control-tag-group .label (Multiselect) …

### LINKED ISSUE #4712: @clayui/css: Forms `$form-control-label-size` is referencing deprecated keys
https://github.com/liferay/clay/blob/5591614c1b88a799eab9cc3afbd96c986180faae/packages/clay-css/src/scss/atlas/variables/_forms.scss#L438-L439 is looking for the keys `lexicon-icon-height` and `lexicon-icon-width`. These have been deprecated and should be `map-deep-get($label-lg, lexicon-icon, height)` and `map-deep-get($label-lg, lexicon-icon, width)`.

We can take a step further and define the values instead of inheriting them from the label component. This would make the forms variables more modular, meaning you won't need to import the labels variable file for the forms variables to work. See #4699.

pat270:
…remove references to deprecated keys

    - Define the values for `.label` inside `.form-control` instead of inheriting them from `.label-lg`

fixes #4712

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_forms.scss
@@ -425,20 +425,21 @@ $form-control-label-size: () !default;
 $form-control-label-size: map-deep-merge(
 	(
 		border-width: 0.0625rem,
-		font-size: map-get($label-lg, font-size),
+		font-size: 0.75rem,
 		height: auto,
 		margin-bottom: 0.3125rem,
 		margin-right: 0.625rem,
 		margin-top: 0.3125rem,
 		min-height: 1.5rem,
-		padding-x: map-get($label-lg, padding-x),
-		padding-y: map-get($label-lg, padding-y),
+		padding-bottom: 0.3125rem,
+		padding-left: 0.5rem,
+		padding-right: 0.5rem,
+		padding-top: 0.3125rem,
 		text-transform: none,
-		item-spacer-y: map-get($label-lg, item-spacer-y),
-		lexicon-icon-height: map-get($label-lg, lexicon-icon-height),
-		lexicon-icon-width: map-get($label-lg, lexicon-icon-width),
-		sticker-size: map-get($label-lg, sticker-size),
-		close: map-get($label-lg, close),
+		label-item: (
+			margin-bottom: -0.0625rem,
+			margin-top: -0.0625rem,
+		),
 	),
 	$form-control-label-size
 );
--- a/packages/clay-css/src/scss/cadmin/variables/_forms.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_forms.scss
@@ -318,20 +318,21 @@ $cadmin-form-control-label-size: () !default;
 $cadmin-form-control-label-size: map-deep-merge(
 	(
 		border-width: 1px,
-		font-size: map-get($cadmin-label-lg, font-size),
+		font-size: 12px,
 		height: auto,
 		margin-bottom: 5px,
 		margin-right: 10px,
 		margin-top: 5px,
 		min-height: 24px,
-		padding-x: map-get($cadmin-label-lg, padding-x),
-		padding-y: map-get($cadmin-label-lg, padding-y),
+		padding-bottom: 5px,
+		padding-left: 8px,
+		padding-right: 8px,
+		padding-top: 5px,
 		text-transform: none,
-		item-spacer-y: map-get($cadmin-label-lg, item-spacer-y),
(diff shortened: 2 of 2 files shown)


packages/clay-css/src/scss/atlas/variables/_forms.scss
```diff
@@ -425,20 +425,21 @@ $form-control-label-size: () !default;
 $form-control-label-size: map-deep-merge(
 	(
 		border-width: 0.0625rem,
-		font-size: map-get($label-lg, font-size),
+		font-size: 0.75rem,
```
matuzalemsteles, line 428: I think for typography cases we should start using typography scale variables to make our system easier to maintain. I think the more we do this the better for the design system. what do you think?

matuzalemsteles (approved):
This change makes sense to me, I don't think it fits like breaking change because we still allow modifying the values independently.

packages/clay-css/src/scss/atlas/variables/_forms.scss
pat270: We currently don't have global variables for the typography scale. [These](https://github.com/liferay/clay/blob/c26eedb067b82ce69584120c703885362ae1b1b1/packages/clay-css/src/scss/variables/_utilities.scss#L310-L348) are for setting the utility class font sizes.

We should add the global variables and convert everything in a separate pr. We need to deprecate Bootstrap's `$font-size-*` variables (mobile added by us) and integrate them into the `$text-#` global variables that we need to create.

matuzalemsteles: Hmm entendi, bem podemos deixar isso para outro PR mas porque nesse caso não usamos o `$font-sizes`? se eu entendi bem teremos que criar outras variaveis?

matuzalemsteles: Hmm got it, we might as well leave that to another PR but why in that case don't we use `$font-sizes`? if I understand correctly will we have to create other variables?

pat270: Yes we will have to create other variables. The `font-sizes` map is used to generate the classes `.text-1` - `.text-#`. If we reuse those values all over the framework, someone only wanting to modify the utility classes will have a very bad time. 

Duplicating / layering is a better approach imo, you have the option to modify globally or locally to a component if you don't want to deal with unintended font size changes.

matuzalemsteles: Hmm, well I worry about that I understand this is good for flexibility but bad for a design system because you can have different font sizes and in theory, the scale should be respected. If you change a scale value I imagine that this must be propagated throughout the system, otherwise it causes inconsistency in the system. What do you think?

pat270: We will map the new global variables we create to the `$font-sizes` map that generates the utilities. If you change the globals everything will scale. If you change only the `$font-sizes` map, only the utilities will have the values changed.

Ultimately our design system (Lexicon) is for us and doesn't translate well if at all to other systems (Dialect). There is no definitive pattern in the way we use `font-size` in our components. It's this `font-size` looks good for this component design, let's use it. If we force our scale with no alternative, other systems are forced to fix the font size changes on each component. 

This provides an alternative for those that only want utilities that match their type scale without affecting OOTB components.

pat270: I'm going to add if we can devise a way to logically map font sizes to components, then forcing the scale makes sense.

matuzalemsteles: > We will map the new global variables we create to the $font-sizes map that generates the utilities. If you change the globals everything will scale. If you change only the $font-sizes map, only the utilities will have the values changed.

Lets do this!

> Ultimately our design system (Lexicon) is for us and doesn't translate well if at all to other systems (Dialect). There is no definitive pattern in the way we use font-size in our components. It's this font-size looks good for this component design, let's use it. If we force our scale with no alternative, other systems are forced to fix the font size changes on each component.

I understand that, but if this happens in Dialect it's also a design system problem in not respecting the global scale and causing inconsistency between components.

julien: Hey guys, I'm OK with these changes, but reading @pat270 's last comment it seems you still want to work on this, so I'll wait before merging

pat270: @julien we're going to work on it in a different PR. I'm going to merge it.

PR #4717, opened 2022-03-03
chore(@clayui/core): adds TreeView tests

### LINKED ISSUE #4658: Create a testing strategy for the TreeView component
Well, the TreeView component is on the way to being stable and currently, we only have the test to cover some cases of `useTree` to deal with immutability. We are also considering using more unit tests as they are cheaper than poshi tests.

In v3 we started to adopt some different approaches to handle tests and use less snapshots, some components are unavoidable because they don't have much behavior like a Button, Badge, Label..., but some suite strategies are very valid that we might want to consider for the implementation of these tests, in Date Picker we adopted the suite of `IncrementalInteractions` in order to replicate the behavior of the Lexicon spec and the user as an "end-to-end" test for a suite in the unit test, also some work in ColorPicker, MultiSelect, Panel, TimePicker.

Some suites that we can create with some goals:

- `IncrementalInteractions`: The suite covers the behaviors that are expected using Lexicon as a basis and simulates an "end-to-end" interaction within the component.
- `BasicRendering`: Suite only covers cases of component compositing and customization, eg different compositing of `TreeView.Item`, the tests here may be more markup related than interaction.

Any suggestion is welcome!

matuzalemsteles:
Fixes #4658

This PR is just a draft, but it already includes the `BasicRendering` suite for the TreeView component, I'm still working on the incremental interactions suite, it has good coverage of the TreeView cases, I should probably finish it tomorrow.

It also includes a fix for when the content is static setting the key it was getting lost because it was using the [`React.Children.toArray`](https://reactjs.org/docs/react-api.html#reactchildrentoarray) API which does this intentionally 🤦‍♂️.

Any suggestion is welcome!

### DIFF
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -87,7 +87,7 @@ export function Collection<T extends Record<any, any>>({
 							</ItemContextProvider>
 						);
 				  })
-				: React.Children.toArray(children).map((child, index) => {
+				: React.Children.map(children, (child, index) => {
 						if (!React.isValidElement(child)) {
 							return null;
 						}
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -347,7 +347,7 @@ export const TreeViewItem = React.forwardRef<
 						}}
 						tabIndex={-2}
 					>
-						{typeof left === 'string' ? (
+						{typeof left === 'string' && !right ? (
 							<Layout.ContentRow>
 								<Layout.ContentCol expand>
 									<div className="component-text">{left}</div>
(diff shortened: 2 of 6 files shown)


matuzalemsteles:
Well, that's ready for review! I believe I added most of the TreeView use cases but I didn't cover all edge cases like all recursive selection behaviors or drag and drop tests for example. But we can add with time, we also need to study how to do D&D tests with jest.

With this implementation we reached almost 76% test coverage for the TreeView, I think we can increase it here by adding D&D tests that can cover a lot of the TreeView.

julien (approved):

matuzalemsteles:
@julien as we talked about today in the call, I'm going to merge this and we can work on the D&D tests at another time as it's going to be a lot of work 😕.

julien:
@matuzalemsteles sure, 76% of coverage is already better than we had so it's already good progress.

PR #4723, opened 2022-03-09
feat(@clayui/drop-down): add new prop to align Menu in viewport

### LINKED ISSUE #4720: @clayui/drop-down Drop-down is not showing all items
Hi guys, we're using **drop-down with items** on Forms component, but it was noticed that when it's near the bottom page, the items are being cut. As you can see on the image below, there are more dates fields after Date 6, but they are not able to be selected.

image

We think it occurs because the max-height is set to 500px for dropdown-menu class so the overflow attribute will not work for our case. It was also noticed that there's a media-query for max-width: 991.98px in which the max-height changes and the overflow works showing all the items. 

image

With all that in mind, my question is if this is something we have to adapt on our side or it's possible to make the dropdown height more responsible for cases like this. 

Thanks in advance :smile:

matuzalemsteles:
Fixes #4720

This PR adds a new `alignmentByViewport` prop which is passed to `dom-align` to align the element considering the viewport.

I didn't set it as the default because in most scenarios it doesn't make sense because the user can scroll to view the DropDown content and if considering the viewport it can cause strange behavior and not very smooth alignment, in the case of the issue it does feeling because it is inside another container with scroll and fixed on the screen.

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -29,6 +29,13 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement | HTMLLIElement> {
 	active: boolean;
 
 	/**
+	 * Flag to align the DropDown menu within the viewport.
+	 */
+	alignmentByViewport?: React.ComponentProps<
+		typeof Menu
+	>['alignmentByViewport'];
+
+	/**
 	 * Default position of menu element. Values come from `./Menu`.
 	 */
 	alignmentPosition?: React.ComponentProps<typeof Menu>['alignmentPosition'];
@@ -97,6 +104,7 @@ const ClayDropDown: React.FunctionComponent<IProps> & {
 	Section: typeof Section;
 } = ({
 	active = false,
+	alignmentByViewport,
 	alignmentPosition,
 	children,
 	className,
@@ -178,6 +186,7 @@ const ClayDropDown: React.FunctionComponent<IProps> & {
 						{...menuElementAttrs}
 						active={active}
 						alignElementRef={triggerElementRef}
+						alignmentByViewport={alignmentByViewport}
 						alignmentPosition={alignmentPosition}
 						closeOnClickOutside={closeOnClickOutside}
 						hasLeftSymbols={hasLeftSymbols}
--- a/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
+++ b/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
@@ -18,6 +18,13 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	active?: boolean;
 
 	/**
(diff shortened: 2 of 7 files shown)


packages/clay-drop-down/src/DropDown.tsx
```diff
@@ -28,6 +28,13 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement | HTMLLIElement> {
 	 */
 	active: boolean;
 
+	/**
+	 * Flag to align the DropDown menu within the viewport.
+	 */
+	alignmentByViewport?: React.ComponentProps<
```
julien, line 34: Naming is hard, I'm wondering if this shouldn't be `alignToViewport`

julien:
@matuzalemsteles the code change looks good to me; **but** I wanted to comment a few things:

- Since we removed dependencies from OSGi modules (more info in [this](https://liferay.slack.com/archives/C03CAD9EF/p1645696653581729) Slack thread), it's has become a bit harder to test local changes with `verdaccio`: you'll probably need to use `npx yarn-deduplicate yarn.lock` to make sure you're using the correct version of the package you want to test 

- Here's what I see when testing this in DXP:

Without scrolling to the bottom of the page, the drop down is still not entirely visible



When scrolling to the bottom of the page, is seems that the drop down is positioned correctly



Hope that helps

matuzalemsteles:
Oh thanks @julien I'll take a look, I did some tests reproducing it outside of DXP and apparently, it had passed I'll try to find out what might be different in the Forms use case.

matuzalemsteles:
Well, I found the problem, the version in our `package.json` of `dom-align` was `1.10.2` but in some `yarn.lock` update it went to the latest minor version which is `1.12.2` , the `alwaysByViewport` property was only added in version `1.11.0`, apparently this was not reflected in DXP, which installs version `1.10.2`, it probably didn't install the latest version because our packages don't ship the `yarn.lock` together with the package publishing because we only have one central, this would tell yarn to install the same version of the package which is also not wrong, so I updated the package version here to be explicit in the `package.json`.

packages/clay-drop-down/src/DropDown.tsx
matuzalemsteles: Yeah, this is pretty complicated I thought of leaving `align*` but I saw that we have the `alignmentPosition` API but it also has `alignElementRef`, I got closer to the same name as the `dom-align` API which is `alwaysByViewport`. I don't have very strong arguments so I'm fine with any name.

julien: > ...but I saw that we have the alignmentPosition

You're right ... when reviewing this again I saw it. So I think we can keep this as it is right now.

julien:
@matuzalemsteles Thanks for the detective work. I'll fix the conflict and re-test this.

matuzalemsteles:
Yeah, I think that's the best alignment will do in these situations of not finding space to render 😕.

Issue #4693, opened 2022-02-24
@clayui/css: POC create theme to map Clay's Sass variables to CSS Custom Properties


pat270:
This is just a test to see how big the file will be and how well it will plug into Liferay themes. We are going to do a 1 to 1 mapping of Clay's Sass variables to CSS Custom Properties.

PR #4729, opened 2022-03-15
feat(@clayui/css): Converting Clay CSS to use CSS Custom Properties

pat270:
This is adding a variable theme to Clay CSS that converts the project to use CSS Custom Properties. It's still a work in progress. I tried to create it without defining any Custom Properties in `:root`. It didn't work out the way I expected. I'm in the process of naming and defining variables to make this user friendly.

One drawback from going this route is the size. I've converted about 60% and `atlas.css` is:
|atlas.css|size|
|---|---|
|Uncompressed|981kb|
|Gzip|103kb|

When this is finished it will be over 1meg uncompressed.

#4693


pat270:
@matuzalemsteles yes it's pretty big. It's ~600kb uncompressed for current atlas.css. If we can figure out an easy way to let devs only include the components they need, the file size could become more appealing.

matuzalemsteles:
Yeah this is pretty big. We might as well try something like loading CSS based on the React component being used, which would be lazy and loaded on demand, the problem with this is that it would only work for React components so it wouldn't be as viable for DXP, unfortunately. I can't see any other way for the DXP case that we can't infer this for JSP, elements that can be dynamically embedded.

pat270:
@matuzalemsteles @dsanz I updated this again. You can generate the `frontend-token-definition.json` file by:
1) go to `packages/clay-css/`
2) `yarn compile`
3) `node scripts/compile-tokens.js`

To test it in portal:

1) run `yarn link` in `packages/clay-css`
2) run `yarn link @clayui/css` in `liferay-portal/modules`
3) copy the file from `scripts/frontend-token-definition.json` and overwrite the one in `liferay-portal/modules/apps/frontend-theme/frontend-theme-classic/src/WEB-INF/frontend-token-definition.json`
4) gradle deploy in `liferay-portal/modules/apps/frontend-theme/frontend-theme-classic/`

The script I wrote to do this is a mess. I had to hack and slash until I got something to work. Feedback is appreciated.

Last thing, Stylebooks can't fully support Clay CSS until there are performance improvements. I've only loaded a handful of components (globals, button, alert, input, table) and it's already too slow to use.

matuzalemsteles:
Hey @pat270 I'll take a look at the code later but is the performance issue in stylebooks due to the number of elements being rendered or is it related to the number of tokens in the CSS?

It must be the same problem related to the listing of big data, normally the browser will have problems with this and React itself too, maybe the stylebooks have to adopt list virtualization for this to not be a problem anymore.

pat270:
I'm not sure it could be a combination. It autosaves every time I type in an input. Whatever it's doing in the background makes it laaag, also the dropdown menu that lists each section takes a while to populate. I think I have 33 categories with 10 - 30 variables in each category. It shouldn't lag so bad imo.

matuzalemsteles (commented):
@pat270 regarding the code, I think some can be improved but it is more at the level of formatting and naming, the regex looks a little complex I would say to use a CSS parser to provide an object but that would probably be slow and the regex here will be much more fast in that case.

### Stylebooks performance

Regarding stylebooks performance, it should probably be related to the number of components being rendered, I think there are some ways to optimize this:

- Consider using virtualization for variable listings
- On the Clay side we can add the same optimization that was done in DropDown for ColorPicker to render the menu via lazy only in user interaction, this decreases the number of elements being rendered when it would be multiplied by 4715
- Probably an optimization in the way of dealing with states in stylebooks, maybe this won't be a problem if you have virtualization. I'm just assuming we using a global state that is being manipulated every time it interacts with the input
  1. We can investigate whether using debounce makes sense, maybe not!
  2. We can control the state of the inputs separately and the global state is a reference to prevent it from causing a rendering in the whole tree, so the rendering will be localized, it will take a little more work but a big performance boost.

### CSS Variables

Regarding the variables and the amount I'm a little skeptical to provide this as default not on the Clay side but on the DXP side at the product level, I commented this on another issue that I worry about this granularity, some things don't make sense for me, for example, allow setting `margin`, `width`, `height` in each corner, left, top... this is worrying because although we want to allow this for the client I have doubts that it will use all properties, that is, I would think more about variables like x, y that are consistent at the UI level, hardly anyone will define different values for bottom or top.

I would try to work on reducing the number of css variables more, providing variables at a higher level. I really wanted to see fonts and some more specific things being global than allowing change by component.

I took some numbers, so with this current implementation, we have about 4715 CSS variables, 1355 just to define `padding`, `width`, `height`, `margin`.

Name | Count 
------|-------
Total  | 4715
Width | 254
Height | 254
Padding | 478
Margin | 369

pat270:
@matuzalemsteles Thanks for the feedback! The CSS parser might be easier. I wanted to learn how to regex so that's what I used. You have any suggestions I can take a look at? Speed isn't much of a concern since this is only for generating `frontend-token-definition.json`.

We can pick and choose what goes in Stylebook from the 4000+ variables that can be customized. This was just a test to see how much it could handle. We probably should have a discussion with @dsanz and @drakonux  about what customizations we need and how it should flow through the system.

We also need to find a way to dramatically cut down the size. To be honest, I'd much rather have the customizations take place outside of DXP and be able to import the CSS you want.

matuzalemsteles:
> The CSS parser might be easier. I wanted to learn how to regex so that's what I used. You have any suggestions I can take a look at? Speed isn't much of a concern since this is only for generating frontend-token-definition.json.

I imagine it might simplify the implementation a bit but it will be a trade-off against performance I guess, but if you want to consider it, you can look at the [csstree](https://github.com/csstree/csstree), it will generate an AST from the CSS and provide some utils to traverse the tree looking for something specific, it also provides many other utils to identify properties.

> We can pick and choose what goes in Stylebook from the 4000+ variables that can be customized. This was just a test to see how much it could handle. We probably should have a discussion with @dsanz and @drakonux about what customizations we need and how it should flow through the system.

Yeah makes sense, I think we need to analyze this and identify what should be taken into the stylebook and what is considered token.

PR #4738, opened 2022-03-17
feat(@clayui/color-picker): add color validation in input

### LINKED ISSUE #4694: @clayui/color-picker Color Picker should accept invalid Hexadecimal?
## What is your question?
Hi guys, I had encountered a scenario while testing but I don't know if it's expected behavior or a bug. Can you guys help me?

We're using Color Picker on Forms component and we noticed that it's possible to save an invalid hexadecimal color like `#FFFFFFaaaaaa`.

> image

Even though on the menu dropdown it's set as `#FFFFFF`, the invalid hexadecimal is still being saved.

> image

Obs: I also noticed that is not possible to write invalid hexadecimal on the dropdown menu input field, which would help us in this case.

This was already reported? Is this a bug or expected behavior? Is possible for us prevent to saving this kind of invalid hexadecimal?

Thanks in advance!

matuzalemsteles:
Fixes #4694

This validation does not prevent how much the user is typing at the moment, only after the user leaves the input the validation is done, either by converting the hex, if the color is not valid it is set to an empty value, this allows people to type different types of colors besides hexadecimal.

For example set the color using CSS Variable, color name, hsl, hsv, rgb. I think the Lexicon document is not covering these cases that came more from the stylebook, so some rules that exist in the specification are very related to the use of hexadecimal but in this case, we can't imply some rules or for example, define the maximum size of the input length instead we leave it free and update the input in `onBlur` that has a better DX to cover all these cases.

### DIFF
--- a/packages/clay-color-picker/src/index.tsx
+++ b/packages/clay-color-picker/src/index.tsx
@@ -353,6 +353,23 @@ const ClayColorPicker: React.FunctionComponent<IProps> = ({
 								])}
 								disabled={disabled}
 								insetBefore
+								onBlur={(event) => {
+									const value = event.target.value;
+
+									const newColor = tinycolor(value);
+
+									if (newColor.isValid()) {
+										if (newColor.getFormat() === 'hex') {
+											onValueChange(newColor.toHex());
+										} else if (
+											newColor.toString() !== value
+										) {
+											onValueChange(newColor.toString());
+										}
+									} else if (!value.includes('var(')) {
+										onValueChange('');
+									}
+								}}
 								onChange={(event) => {
 									onValueChange(event.target.value);
 								}}


matuzalemsteles:
cc @liferay-lexicon

matuzalemsteles:

emiliano-cicero:
Hey @matuzalemsteles 
I don't think that as a system we should allow this kind of behavior for the inputs. If this comes from Stylebooks, they're free as a team to allow that behavior but we shouldn't inherit it in the system.
My concern is: how would the users know that they can manually paste any of those parameters such as HSL or RGBA color inside the input? 
IMO it should be clear for the users what the requirements for the input are before starting typing. 

I would suggest opening a Design ticket for @liferay-lexicon to review this.

cc/ @drakonux

matuzalemsteles:
Hey @emiliano-cicero this is pretty hard for us, trying to allow both sides, for example, if stylebooks need it, we have to have a way that stylebooks can bypass our validation system and they create their own.

> My concern is: how would the users know that they can manually paste any of those parameters such as HSL or RGBA color inside the input? IMO it should be clear for the users what the requirements for the input are before starting typing.

Yeah, I agree.

PR #4739, opened 2022-03-17
chore: adds coverage threshold for Clay packages

### LINKED ISSUE #4731: Configure the coverage threshold of each package in the Jest config
Well, we are starting efforts to improve our test coverage based on packages and once we have everything in one package it will be component-based. This is related to the coverage testing initiatives that were written in the [Test coverage](https://github.com/liferay/clay/blob/dc322a085c46b0d5746e682e8ad7c68866c975f4/docs/tests.md) doc.

To do this we can use jest itself which has the [`coverageThreshold`](https://jestjs.io/docs/configuration#coveragethreshold-object) API where we can define these limits by a path, we use [coveralls](https://coveralls.io/github/liferay/clay) which can also be used by the QA team to visualize in real-time the test coverage in the master.

matuzalemsteles:
Fixes #4731

We are adding the limits closer to the current coverage, they will increase organically as we add more features and fix bugs.

Some packages are out of this limit because I think they will no longer be needed, for example, packages in a deprecated state.

- clay-charts
- clay-management-toolbar
- clay-localized-input
- clay-upper-toolbar

This is following our recommendations of the [Test coverage](https://github.com/liferay/clay/blob/dc322a085c46b0d5746e682e8ad7c68866c975f4/docs/tests.md) and QA recommendations. Many of the limits are close to current coverage which are more complex components that we can't lose sight of if coverage drops, simpler components like a button the limits are more rounded because they are simpler components but not that far from current coverage.


matuzalemsteles:

PR #4740, opened 2022-03-18
chore: moves component documents to respective docs folder

### LINKED ISSUE #4202: Organizes component documentation
After we moved all the documents referring to the clayui.com component to the respective package in our folder structure, we had to use the `README.md` of each package as the main file, making `README.mdx` but polluting the information from the package because `.mdx` files contain syntax that Github won't render, some packages like `form` have more than one `.mdx` file and most packages have a `docs` folder which contains extra files related to the component, as a markup, api, and supporting JavaScript file to build the component examples.

Well, after we started working on the `@clayui/core` package, there was a need to not use `README.md` and use the `docs` folder to add the corresponding component documentation, to support this was added this in the clayui.com engine via https://github.com/liferay/clay/pull/4162/commits/5e00f5abb9a5230e396ca472c308849f1b2835f6. The `@clayui/provider` package already uses this conversion as well as a cleaner `README.md`.

The goal is to do the same for the other packages to keep the component documentation organized.

matuzalemsteles:
Fixes #4202

We are moving the docs to the respective `docs` folder in each package to avoid polluting the `README.md` of packages that don't render well on GitHub.

This also cleans up more folders that had a lot of documents like Form, in the root of the package. This change does not affect document URLs.


matuzalemsteles:

PR #4743, opened 2022-03-18
fix(@clayui/core): fixes error when removing parent's intermediate state when item's sibling is intermediate

### LINKED ISSUE #4742: Uncheck the checkbox when a sibling item is indeterminate is removing the indeterminate state of the parent
This is a very specific edge case that is removing the indeterminate state from the parent if no child is selected but some child is in an indeterminate state, in which case it must preserve the parent's state.

```
Root (Intermediate)
├─ Item 0 (Checked) <- Click to uncheck
├─ Item 1 (Intermediate)
│  ├─ Item 2 (Checked)
│  ├─ Item 3
```

matuzalemsteles:
Fixes #4742

Basically, this PR fixes an edge case that was not being considered, considering that in a recursive flow to add or remove the item from the `intermediate` state it was just checking if the children of the parent were selected would decide but exist a case of a tree in-depth than any of the parent's children can have other children and is set to `intermediate`, in which case the recursive multiple selection must preserve the `intermediate` state of the parent if there is any child selected or `intermediate`.

This ASCII represents an illustration of a tree with this edge case:

```
Root (Intermediate)
├─ Item 0 (Checked) <- Click to uncheck
├─ Item 1 (Intermediate)
│  ├─ Item 2 (Checked)
│  ├─ Item 3
```

I'm also adding a test for this selection edge case.

### DIFF
--- a/packages/clay-core/src/tree-view/__tests__/IncrementalInteractions.tsx
+++ b/packages/clay-core/src/tree-view/__tests__/IncrementalInteractions.tsx
@@ -87,6 +87,77 @@ describe('TreeView incremental interactions', () => {
 
 	describe('selection', () => {
 		describe('checkbox', () => {
+			it("uncheck the checkbox when a sibling item is indeterminate preserve the parent's indeterminate state", () => {
+				const {container} = render(
+					<Provider spritemap={spritemap}>
+						<TreeView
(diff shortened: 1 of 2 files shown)


matuzalemsteles:

PR #4744, opened 2022-03-21
feat(clay-modal): improves ClayModal accessibility

### LINKED ISSUE #4736: Improve ClayModal component accessibility
Well, our ClayModal component only has the focus control that persists inside Modal, it was discovered a way for the screen reader to be reading the content from outside Modal when pressing the "up" key, that's because Modal doesn't have the `autoFocus` or we are not moving the focus to the Modal so that the screen reader can read the content as soon as it opens, for example: Open the modal read the dialog title.

The current implementation of ClayModal is also not adding attributes like `role="dialog"`, `aria-hidden="true"` and `aria-labelledby` to improve accessibility.

matuzalemsteles:
Fixes #4736

Basically, it has the same implementation that was done in #4737 but brings it to the version with Metal.js + Soy.


matuzalemsteles:
I'm going ahead and merging this, the implementation is similar to #4737 and I'm also going to release that today.

PR #4758, opened 2022-03-23
fix(@clayui/css): Utilities Close use newer keys in Sass map

### LINKED ISSUE #4734: @clayui/css: Atlas Utilities $close should use newer keys
Atlas `$close` is still using older key names. This is a problem if someone tries to overwrite it using newer keys in `_clay_variables.scss`. The old keys always win for backward compatibility purposes.

https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/atlas/variables/_utilities.scss#L20-L34

pat270:
- Change `hover-*`, `focus-*`, `disabled-*` to `hover: (property: value,)`, `focus: (property: value,)`, and `disabled: (property: value)`, etc
- Removes the use of `$btn-*` variables, to make the Utilities variable file more modular
- Make focus style consistent for `a` or `button`, we have the `c-inner` pattern for removing focus on click

fixes #4734

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_utilities.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_utilities.scss
@@ -21,14 +21,18 @@ $close: map-deep-merge(
 	(
 		color: $gray-900,
 		opacity: 1,
-		hover-color: $gray-900,
-		hover-opacity: 1,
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM

PR #4760, opened 2022-03-23
fix(@clayui/css): Table removes `border-width: 0` from `.table-respon…

### LINKED ISSUE #4724: @clayui/css: Tables outer border is removed from tables nested in `table-responsive`
Don't think this workaround is needed anymore, unless I'm missing an edge case somewhere.

https://github.com/liferay/clay/blob/60eaf7a67f19895798a315dd1f18a4546f151572/packages/clay-css/src/scss/components/_tables.scss#L398-L402

pat270:
…sive > .table-bordered`

    - `fixes table-bordered` wrapped in `table-responsive` remove outer borders

fixes #4724

### DIFF
--- a/packages/clay-css/src/scss/components/_tables.scss
+++ b/packages/clay-css/src/scss/components/_tables.scss
@@ -394,12 +394,6 @@ caption {
 
 				overflow-x: auto;
 				width: 100%;
-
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM

PR #4761, opened 2022-03-23
fix(@clayui/css): Mixin clay-nav-nested use css calc() to calculate l…

### LINKED ISSUE #4721: @clayui/css: Mixin `clay-nav-nested` should use calc to calculate padding-left
Using CSS `calc()` will make this mixin more friendly to CSS Custom Properties. Custom Properties don't work in this mixin due to using Sass arithmetic.

https://github.com/liferay/clay/blob/747b1ec0382c763b17397be2355d95eab168a0db/packages/clay-css/src/scss/mixins/_nav-nested.scss#L14-L33

pat270:
…eft padding

    - We can use CSS Custom Properties instead of number values

fixes #4721


matuzalemsteles (approved):
LGTM

PR #4764, opened 2022-03-24
fix(@clayui/clay-tabs): Insert displayType for nav-tabs and nav-undeline classes

sergiojimcos:
**What is new?**

- Change `modern` property to `displayType` that includes three type:

1. `null`: default, no styles.
2. `tabs`: `nav-tabs`.


packages/clay-tabs/src/index.tsx
```diff
@@ -10,38 +10,35 @@ import Content from './Content';
...
+	 * Determines how to is displayed.
 	 */
-	justified?: boolean;
+	displayType?: DisplayType;
 
 	/**
-	 * Applies a modern style to the tab.
+	 * Justify the nav items according the tab content.
 	 */
-	modern?: boolean;
```
matuzalemsteles, line 22: Sorry if this isn't very clear but unfortunately we still need to support this property to avoid breaking change, so we have to handle with this property still taking effect but using `displayType` overrides the behavior of `modern`.

We are going to remove this API in v4, so you can add this information in the document in the `BREAKING.md` package as a list, we are using this so we don't lose sight of the APIs we need to remove in the next major version.

packages/clay-tabs/src/index.tsx
```diff
@@ -10,38 +10,35 @@ import Content from './Content';
...
-}: IProps) => {
+} = ({children, className, displayType, justified, ...otherProps}: IProps) => {
+	const defaultValue = displayType === null;
+
 	return (
 		<ul
 			className={classNames(
 				'nav',
 				{
+					[`nav-${displayType}`]: !defaultValue,
```
matuzalemsteles, line 40: Since we're just using it here, I think we can avoid creating a variable for this and adding the conditional directly here just to avoid allocating a new variable.

packages/clay-tabs/src/index.tsx
```diff
@@ -10,38 +10,35 @@ import Content from './Content';
 import Item from './Item';
 import TabPane from './TabPane';
 
+export type DisplayType = null | 'tabs' | 'underline';
```
matuzalemsteles, line 13: I think in this case we can go with different names for the display types, maybe just change `tabs` to `basic`, I didn't have much creativity, if you have another name feel free.

matuzalemsteles (commented):
Hey @sergiojimcos, I added some comments regarding these changes as well I think this should be considered a `feat` rather than a `fix` because this behavior was expected so it is not treated as a bug but an API improvement to allow better flexibility for that case. Let me know if anything is unclear.

packages/clay-tabs/src/index.tsx
sergiojimcos: Ohh! Got it! At the first time, I thought that this was a change requested but considering as a bug. So in this case, I have to still support the `modern` behaviour with the new property `displayType`. However, `displayType` must override `modern` one. Yeah, I will add this information in `BREAKING.md`, to record it for the v4.

matuzalemsteles: Yes, exactly sorry for this confusion 😅!

sergiojimcos:
Hey @matuzalemsteles! As I saw in the `CI/test` details, it fails because the coverage of the module `clay-tab` is lower lower than 84% (It is 80%). Should I create a new issue and preparing new test cases to increase the coverage? Or should I continue in this PR?

matuzalemsteles:
> As I saw in the CI/test details, it fails because the coverage of the module clay-tab is lower lower than 84% (It is 80%). Should I create a new issue and preparing new test cases to increase the coverage? Or should I continue in this PR?

Github didn't show me this message 🥲, in that case, I think you can send it here together since I'm going to cut a new release today.

sergiojimcos:
> Github didn't show me this message 🥲, in that case, I think you can send it here together since I'm going to cut a new release today.

No problem @matuzalemsteles! This is the details of the `CI/test` :
CI:test details

Great! I will prepare some tests scenarios to increase the coverage using the new behaviour of the `displayType` prop.

sergiojimcos:
Hey @matuzalemsteles! I've created some test scenarios related to `displayType` prop to increase coverage and `CI/ tests` passed:
- ClayItem renders using `displayType` prop.
- ClayItem renders using `modern` prop and `displayType` as `null`.
- ClayItem renders using `displayType = 'basic'` and overrides `modern` prop.

matuzalemsteles (approved):
Thanks @sergiojimcos!

PR #4767, opened 2022-03-25
feat(@clayui/css): Adds $font-scale map for storing a Design System's font sizes

### LINKED ISSUE #4393: Can't use css variables in $font-size-base and other font size sass variables.
I want to be able to use style books to set font-size sass variables but the theme will not compile.

```
// Fonts

$font-size-base: var(--font-size-base);
$font-size-lg: var(--font-size-lg);
$font-size-sm: var(--font-size-sm);
```

### LINKED ISSUE #4718: @clayui/css: Create `$text-#` global variables for reusable font-size values
From https://github.com/liferay/clay/pull/4716#discussion_r819162725

We need to deprecate `$font-size-*` variables and integrate them into the `$text-#` global variables. This will provide consistency for Design Systems.

Pros:
- Don't need to remember the exact font size value
- Unauthorized font size values will be less likely to appear in the final output

Cons:
- More variables to remember/consider
- Won't propagate logically, these are just reusable values

pat270:
This creates a `$font-scale` map where we can store a Design System's progression of font sizes for reuse. You can reuse them with `map-get($font-scale, #)`. I was hoping to replace `$font-size-base`, `$font-size-sm`, `$font-size-lg`, etc with the font scale, but it would break font sizes for someone who relied on the scaling of Bootstrap's `$font-size-base`.

I ended up fixing #4393 in this pr as well. The font size variables supports CSS Custom Properties now.

/cc @ethib137 

fixes #4718 
fixes #4393

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_globals.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_globals.scss
@@ -362,34 +362,23 @@ $font-scale: () !default;
 $font-scale: map-merge(
 	(
 		1: 0.625rem,
-		1-sm-down: null,
 		2: 0.75rem,
-		2-sm-down: null,
 		3: 0.875rem,
-		3-sm-down: null,
 		4: 1rem,
-		4-sm-down: null,
 		5: 1.125rem,
-		5-sm-down: null,
 		6: 1.25rem,
-		6-sm-down: null,
 		7: 1.5rem,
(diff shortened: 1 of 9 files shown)


matuzalemsteles (approved):
LGTM! @pat270 makes sense, I think we can make these changes in v4 so that everything is using `font-scale` by default.

Issue #4766, opened 2022-03-24
Improves consistency of controlled and uncontrolled component APIs


matuzalemsteles:
Most or almost all of our components allow them to be controlled or uncontrolled using the `useInternalState` hook under the hood, some like DropDown force just use as controlled.

In pull request #4727 we introduced an improvement to avoid the confusion of when the component is behaving like controlled or uncontrolled by adding properties of `default*`, some components use different properties to set the initial value of a value, for example, `initialValue`, some use more friendly naming for the component like `showPalette` to set the initial value of the state or just a flag, in these cases, it might not make sense to change it for now but we should keep the consistency for the rest of the components that follow this behavior.

We don't want to cause breaking changes with this change, so we will alias and mark inconsistent APIs as deprecated and remove them in the next major version.

PR #4774, opened 2022-03-29
docs(RFC): Controlled and uncontrolled components RFC

matuzalemsteles:
This document describes in more detail the final proposal for dealing with the controlled and uncontrolled components in relation to issue #4766.

### [View the rendered text](https://github.com/matuzalemsteles/clay/blob/115b46c965a7a3efca53ecdd2d3177e2ccfd1ee7/docs/proposals/0002-controlled-and-uncontrolled.md)


bryceosterhaus:
Great document! I really like the idea of have a consistent pattern for this.

matuzalemsteles:
Okay, looks like this is ok. I'm going ahead and merging the RFC, I'll start working on that too.

PR #4779, opened 2022-03-29
feat(@clay/empty-state): Enable users to not render title related markup on empty-state

### LINKED ISSUE #4776: EmptyState: hide title markup in JSX if no value is present
## What is your proposal?
Hide the title markup if no value

It should enable a more consistent use of spacings and eliminates unneeded markup

Leave it as it is

image

diegonvs:
## What It Does

When passing `title={null}` it will not render title related markup.

## Worries

Ideally we should:

1. Stop naming props as native DOM attrs 
2. Don't provide default values, turning `title` and `description` properties mandatory and not optional anymore, since they should be i18n with `Liferay.Language.get`.

Can I create tickets for it? Maybe to be updated on v4? /cc @matuzalemsteles @bryceosterhaus 

## How to test 

1. Fetch my changes
2. Run `yarn storybook` on Clay repo
3. Go to ClayEmptyState > no title on Storybook :)

Fixes #4776

### DIFF
--- a/packages/clay-empty-state/src/__tests__/index.tsx
+++ b/packages/clay-empty-state/src/__tests__/index.tsx
@@ -24,6 +24,12 @@ describe('ClayEmptyState', () => {
 		expect(container).toMatchSnapshot();
 	});
 
+	it('when passing `title` property with null, it will not render title section', () => {
+		const {queryByText} = render(<ClayEmptyState title={null} />);
+
+		expect(queryByText('No results found')).toBeNull();
+	});
+
 	it('renders with a children content', () => {
 		const {container} = render(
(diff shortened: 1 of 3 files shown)


matuzalemsteles:
hey @diegonvs maybe we don't need this anymore, Marko and Beltrán are considering following Lexicon's behavior so it will be safe to dismiss this https://github.com/liferay-frontend/liferay-portal/pull/2069

matuzalemsteles:
> Can I create tickets for it? Maybe to be updated on v4?

Yeah it makes sense to me, I think we have to do a component-by-component analysis and see why it was used and come up with some general solution.

matuzalemsteles (approved):
LGTM! Let's go ahead with this change as a flexibility feature.

Issue #4732, opened 2022-03-15
Adds more test cases for the ColorPicker component


matuzalemsteles:
The Color Picker component has changed some behaviors and added new behaviors with Custom, during the development of PR #4691 we did a lot of manual testing to check if we were covering the edge cases.

We need to cover these edge cases in the unit tests to help us not cause regressions.

sergiojimcos:
Hey @matuzalemsteles! I am going to start working on this issue

PR #4781, opened 2022-03-30
feat(@clayui/color-picker): Add more test cases in color-picker

sergiojimcos:
**What is new?**
Adding more test cases due to `Custom Color Picker` component has new behaviour

**New test cases:**

1. `Custom Color Picker` does not update when a splotch is not clicked.
2. `Custom Color Picker` does not update if a splotch (with value different than `FFFFFF`) is clicked.
3. `Custom Color Picker` updates if a splotch (with value of `FFFFFF`) is clicked.
4. `Custom Color Picker` can display name color values in hex using `customHexInput`.
5. `Custom Color Picker` can duplicate a color.
6. `Custom Color Picker` can modify the value of the input using the color editor components like using the `RGB inputs`.

**Issues:**
#4732


matuzalemsteles (commented):
Hey @sergiojimcos looks good, I just added a comment about using a snapshot, also I think we can add some more test scenarios here like:

- Opening the editor with a different color and modifying the editor's input should only change the input of the selected color
- Check if colors are being duplicated
- Check if colors with CSS Variables are being converted to hex in the editor

Let me know if anything is unclear, I can explain in more detail.

packages/clay-color-picker/src/__tests__/index.tsx
```diff
@@ -373,6 +396,86 @@ describe('Interactions', () => {
...
+
+			fireEvent.click(dropdownToggle as HTMLButtonElement, {});
+
+			const colorEditorToggle = (
+				document.body as HTMLElement
+			).querySelector('.clay-color-header button');
+
+			fireEvent.click(colorEditorToggle as HTMLButtonElement, {});
+
+			expect(document.body).toMatchSnapshot();
```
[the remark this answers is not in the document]
sergiojimcos: Make sense! I will change using snapshots and try to make some asserts to the components that define the test case. It would be great to use those custom components that you have mentioned and the input that is on the bottom of the gradient and the hue one (`customHexInput`)

packages/clay-color-picker/src/__tests__/index.tsx
```diff
@@ -467,15 +489,113 @@ describe('Interactions', () => {
...
+
+			fireEvent.change(rInput, {target: {value: '200'}});
+
+			fireEvent.change(gInput, {target: {value: '200'}});
+
+			fireEvent.change(bInput, {target: {value: '200'}});
+
+			expect(input.value).toBe('c8c8c8');
+
+			expect(editorGetByTestId('customHexInput').value).toBe('C8C8C8');
```
matuzalemsteles, line 542: As a final verification test here, you can check if there is any splotch with this color.

packages/clay-color-picker/src/__tests__/index.tsx
```diff
@@ -467,15 +489,113 @@ describe('Interactions', () => {
...
+			fireEvent.change(input, {target: {value: 'DFCAFF'}});
+
+			const purpleSplotchs = (
+				document.body as HTMLButtonElement
+			).querySelectorAll('button[title="dfcaff"]');
+
+			expect(purpleSplotchs.length).toBe(2);
+		});
+
+		it('CSS values are displayed in HEX in the custom hex input', () => {
```
matuzalemsteles, line 579: I think there are two use cases we need to certify, one is named colors like you're doing here, so maybe it's good to define a text that matches that this is testing for named colors and another to check CSS variables with `var(--blue )` for example. I'm not sure if this will work well because maybe in the tests in Jest it's not possible to get the color of the variable, if not we can go without that case.

matuzalemsteles (commented):
Good work, I just added a few comments but nothing that will change much.

packages/clay-color-picker/src/__tests__/index.tsx
sergiojimcos: Hey @matuzalemsteles! I've tred using values like `var(--red)` but Jest doesn't provide this kind of input, that's why I used the `red` one. If you want, I can change the definition of the test case.

matuzalemsteles: Oh I see, I imagined this I was hoping we would make it 🥲. yes, it is better to change the test definition to understand that we are checking named colors here.

matuzalemsteles (approved):
LGTM! Thanks @sergiojimcos

Issue #4773, opened 2022-03-28
Long text in `Clay Navigation Bar.Item` is not being truncated


matuzalemsteles:
Adding long text to the `<Item />` component is causing space issues, the component is not truncating the text by default.

![Screen Shot 2022-03-28 at 2 42 14 PM](https://user-images.githubusercontent.com/13750819/160474393-935b7e6a-88d1-4e07-8b2f-b84c2019febe.png)

This bug was discovered in a Forms use case where the `<Item />` is dynamic, https://issues.liferay.com/browse/LPS-143109.

sergiojimcos:
Hey @matuzalemsteles! I'm going to start working in this issue

PR #4782, opened 2022-03-30
fix(@clayui/navigator-bar): Long text in Item is truncated

sergiojimcos:
**What is new?**
- Adding `text-truncate` in `ClayNavigationBar`.
- Updates snapshots for test cases.

 **Before changes:**

https://user-images.githubusercontent.com/44723449/160881502-82ea1313-08b4-467f-b7ad-88e946ab37ef.mov

**After changes:**

After Changes

**Reproduce in portal:**
Using the Dev tools and applying `text-truncate` in the component, the result is:

https://user-images.githubusercontent.com/44723449/160883404-39f39f3f-b0a5-43e9-8904-2afd793407f0.mov

**Issues:**
#4773

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_navbar.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_navbar.scss
@@ -65,6 +65,7 @@
 		align-items: center;
 		display: flex;
 		word-wrap: normal;
+		max-width: 100%;
 	}
 
 	.nav-item {
--- a/packages/clay-css/src/scss/components/_navbar.scss
+++ b/packages/clay-css/src/scss/components/_navbar.scss
@@ -63,6 +63,7 @@
 		align-items: center;
 		display: flex;
 		word-wrap: normal;
+		max-width: 100%;
 	}
 
 	.nav-item {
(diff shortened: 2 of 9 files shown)


matuzalemsteles:
Hey @sergiojimcos I'm sending some more changes on top of this PR, some things are improvements that simplify the use of the component to avoid for example developers having to manually add the `navbar-text-truncate` instead we are adding OOTB in case it is a component ` ClayLink` or `ClayButton`, it is also no longer necessary to manually add `nav-link` and set `displayType="unstyled"`.

I updated the documentation to remove some code that was wrong in the examples and trying to structure it a bit, maybe we can improve this documentation more in the future.

packages/clay-css/src/scss/cadmin/components/_navbar.scss
```diff
@@ -65,6 +65,7 @@
 		align-items: center;
 		display: flex;
 		word-wrap: normal;
+		max-width: 100%;
```
matuzalemsteles, line 68: Hey @pat270 I added this in `nav-item` because when the text is too big even if it has the truncate class when the navigation-bar is in mobile version it doesn't respect the screen size, let me know if this can break some cases of use.

matuzalemsteles: I talked to Patrick about it, it's ok to go with it.

PR #4786, opened 2022-04-01
feat: standardizes behaviors and APIs for controlled and uncontrolled components

### LINKED ISSUE #4766: Improves consistency of controlled and uncontrolled component APIs

matuzalemsteles:
Fixes #4766

This PR is applying the [RFC 0002 Controlled and uncontrolled components](https://github.com/liferay/clay/blob/4bad9532a7ee689b6add7ba8fa1dbda1fe8b37b3/docs/proposals/0002-controlled-and-uncontrolled.md), with this change we are deprecating some public APIs of the components but still maintaining backward compatibility, also adding the behavior of controlled and uncontrolled for the components that had only the behavior of controlled.

Some components mainly for the `@clayui/form` package, like Checkbox, Toggle, and Select don't allow us to make these changes now because the `onChange` prop has the event argument, we will have to do it in v4.

This PR is still a WIP, still need to add this for other components and also make changes to the documentation. I'm also thinking about adding backward compatibility tests to make sure we don't break anything.

### Discussions

The RFC does not cover this naming part for some edge cases because they are more preferred and optional things, most of the components have the `active` or `open` state of the content, for example: open the DropDown/DatePicker/ColorPicker menu; open the Modal; show the autocomplete menu, we are using the nomenclature of `active` in the vast majority and others like `expanded`, well the idea here is also to define a unique nomenclature for cases like opening/showing a suspended content.

- `active` seems to be a bit confusing that you use for example in a Modal or DropDown, it doesn't seem very predictable to link with the menu.
- `visible` is probably more predictable when content was hidden and is being shown/opened.
- `open` is very similar to `visible`, seems predictable, and perhaps more common for these scenarios.
- `expanded` is less predictable, normally linked to collapse actions like Panel and TreeView, seems to be more aligned for this scenario if you look at the accessibility nomenclature for these `aria-expanded` scenarios.

I don't have a very strong choice on the name of these properties but I'm bringing them here so others can add their opinions as I work on this PR. Probably `open` fits well within these scenarios that are linked to suspended content.

### ToDo

- [x] Update documentation
  - [x] ColorPicker
  - [x] DatePicker
  - [x] DropDown
  - [x] ClayRadioGroup
  - [x] Modal
  - [x] MultiSelect
  - [x] MultiStepNavWithBasicItems
  - [x] ~~VerticalNav~~
  - [x] PaginationWithBasicItems
  - [x] PaginationBarWithBasicItems
  - [x] Panel
  - [x] Popover
  - [x] Slider
  - [x] TimePicker
- [x] Add backward compatibility tests
  - [x] ColorPicker
  - [x] DatePicker
  - [x] DropDown
  - [x] ClayRadioGroup
  - [x] Modal
  - [x] MultiSelect
  - [x] MultiStepNavWithBasicItems
  - [x] ~~VerticalNav~~
  - [x] PaginationWithBasicItems
  - [x] PaginationBarWithBasicItems
  - [x] Panel
  - [x] Popover
  - [x] Slider
  - [x] TimePicker
- [x] Standardize controlled and uncontrolled components
  - [x] ColorPicker
  - [x] DatePicker
  - [x] DropDown
  - [x] ClayRadioGroup
  - [x] Modal
  - [x] MultiSelect
  - [x] MultiStepNavWithBasicItems
  - [x] ~~VerticalNav~~
  - [x] PaginationWithBasicItems
  - [x] PaginationBarWithBasicItems
  - [x] Panel
  - [x] Popover
  - [x] Slider
  - [x] TimePicker

### DIFF
--- a/clayui.com/content/blog/2022-05-02-api-consistency-improvements-for-controlled-and-uncontrolled-components.md
+++ b/clayui.com/content/blog/2022-05-02-api-consistency-improvements-for-controlled-and-uncontrolled-components.md
@@ -1,0 +1,84 @@
+---
+title: 'API consistency improvements for controlled and uncontrolled components'
+author: [matuzalemsteles]
+---
+
+In the last few months, we have been working on identifying points for improvement in Clay components with the mindset of decreasing the learning curve and improving the DX (developer experience) of the components. This requires a lot of research about our components and using existing patterns within the community so we don't reinvent new patterns that can increase the learning curve for the developers but we also want to abstract certain complexities for the component to avoid the developer not having to worry about implementing it on its own.
+
+There are many areas of work that we want to tackle and improve in Clay and we will be making improvements on top of the v3 with compatibility so that current applications continue to work but many APIs are being deprecated which will be removed in v4. In overview, we hope to improve Clay in different arias:
+
+-   [API Consistency](https://github.com/liferay/clay/milestone/47)
+    -   [Standardizing collection components](https://github.com/liferay/clay/issues/4112)
+    -   [Standardizing form-related components](https://github.com/liferay/clay/issues/4741)
+    -   [Improves consistency of controlled and uncontrolled component APIs](https://github.com/liferay/clay/issues/4766) (Current work)
+-   [Solid components foundation](https://github.com/liferay/clay/milestone/50)
+-   [Develop a new structure for documentation](https://github.com/liferay/clay/milestone/48)
+-   Accessibility
+
+These improvements will be progressively implemented in v3 and some may only come in v4 if we are unable to implement compatibility. Now we want to present significant improvements for the consistency of the APIs of the components to the behavior of controlled and uncontrolled.
+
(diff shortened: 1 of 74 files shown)


matuzalemsteles:
This PR is almost ready to go, I'm doing a final review on the code and changes and I'm also going to prepare a blog post to write about these changes.

matuzalemsteles:
Added the blog post adding more information about this change for the developers, any feedback is welcome https://deploy-preview-4786--next-clayui.netlify.app/blog/2022/05/02/api-consistency-improvements-for-controlled-and-uncontrolled-components.html

matuzalemsteles:
I'm going ahead with that and merging it for us to add in today's release.

PR #4793, opened 2022-04-06
feat(@clayui/core): adds the new OverlayMask component

### LINKED ISSUE #4791: Create a new OverlayMask component
Well this is linked with the functionality for the Walkthrough component, at the definition level we don't want to create it just for specific use for Walkthrough but we also want to use this component for the current Modal but with highlighted area support.

This component is more like a utility to allow composing the Walkthrough component and for Modal, we probably won't use CSS here due to the highlighted area which probably wouldn't be possible with CSS alone, so we'll use purely SVG to build this.

For references, see:

- [Reactour](https://reactour.vercel.app/)

matuzalemsteles:
Fixes #4791

Well, this is a new component to handle element highlighting and overlay, I wanted to try using it in Modal but it's a bit overkill for Modal. Also still not sure about that name, I thought `<HighlightOverlay />` seems like a good option.

This component allows [controlling and uncontrolled](https://github.com/liferay/clay/blob/0ac74bb7b16322be916a0dbcc74646c810037aef/docs/proposals/0002-controlled-and-uncontrolled.md) the `bounds` state that defines the highlight aria, otherwise, it will be automatically detected if `children` is defined.

```jsx
// Create the highlight for the Button component
<OverlayMask visible>
  <ClayButton>Button</ClayButton>
</OverlayMask>
```

Another option is to define the children as a function and the `ref` is passed via render props to allow you to define which element should be highlighted if there are too many components being rendered in the `OverlayMask`. See a use case in the storybook using.

```jsx
<OverlayMask visible>
  {(ref) => (
    <>
      <ClayButton>Button A</ClayButton>
      <ClayButton ref={ref}>Button B</ClayButton>
    </>
  )}
</OverlayMask>
```

It is also possible to set the highlight for non-React elements, so you can pass these values to the `bounds` property.

```jsx
const logo = document.body.querySelector('.logo');
const {height, width, x, y} = logo.getBoundingClientRect();

<OverlayMask
  defaultBounds={{
    height,
    width,
    x,
    y,
  }}
  visible
/>
```

### ToDo

- [x] Create documentation
- [x] Create tests

### DIFF
--- a/packages/clay-core/src/overlay-mask/OverlayMask.tsx
+++ b/packages/clay-core/src/overlay-mask/OverlayMask.tsx
@@ -17,9 +17,10 @@ type Bounds = {
 	y: number;
 };
 
-const M = (x: number, y: number) => `M ${x} ${y}`;
-const L = (x: number, y: number) => `L ${x} ${y}`;
-const arc = (toX: number, toY: number) => `A 0,0 0 0 0 ${toX},${toY}`;
+const MoveTo = (x: number, y: number) => `M ${x} ${y}`;
+const LineTo = (x: number, y: number) => `L ${x} ${y}`;
+const EllipticalArcCurve = (toX: number, toY: number) =>
+	`A 0,0 0 0 0 ${toX},${toY}`;
 
 const createClipPath = (
 	bounds: Bounds,
@@ -39,20 +40,20 @@ const createClipPath = (
 	const endY = startY + bounds.height + padding * 2;
 
 	return [
-		M(parentBounds.startX, parentBounds.startY),
-		L(parentBounds.startX, parentBounds.endY),
-		L(parentBounds.endX, parentBounds.endY),
-		L(parentBounds.endX, parentBounds.startY),
+		MoveTo(parentBounds.startX, parentBounds.startY),
+		LineTo(parentBounds.startX, parentBounds.endY),
+		LineTo(parentBounds.endX, parentBounds.endY),
+		LineTo(parentBounds.endX, parentBounds.startY),
 		'z',
-		M(startX, startY),
-		L(startX, endY),
-		arc(startX, endY),
-		L(endX, endY),
-		arc(endX, endY),
-		L(endX, startY),
-		arc(endX, startY),
-		L(startX, startY),
-		arc(startX, startY),
+		MoveTo(startX, startY),
+		LineTo(startX, endY),
+		EllipticalArcCurve(startX, endY),
+		LineTo(endX, endY),
+		EllipticalArcCurve(endX, endY),
+		LineTo(endX, startY),
+		EllipticalArcCurve(endX, startY),
+		LineTo(startX, startY),
+		EllipticalArcCurve(startX, startY),
 	].join(' ');
 };
 
--- a/packages/clay-core/src/overlay-mask/OverlayMask.tsx
+++ b/packages/clay-core/src/overlay-mask/OverlayMask.tsx
@@ -3,7 +3,11 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {TInternalStateOnChange, useInternalState} from '@clayui/shared';
+import {
+	TInternalStateOnChange,
+	observeRect,
+	useInternalState,
+} from '@clayui/shared';
 import React, {useEffect, useLayoutEffect, useRef, useState} from 'react';
 
 type Bounds = {
@@ -124,15 +128,6 @@ export function OverlayMask<T>({
 	const childrenRef = useRef<HTMLElement | null>(null);
 
 	useIsomorphicLayoutEffect(() => {
-		if (childrenRef.current) {
-			const {height, width, x, y} =
-				childrenRef.current.getBoundingClientRect();
-
-			setBounds({height, width, x, y});
-		}
-	}, [visible]);
-
-	useIsomorphicLayoutEffect(() => {
 		if (containerRef.current) {
 			const {height, width, x, y} =
 				containerRef.current.getBoundingClientRect();
@@ -140,6 +135,23 @@ export function OverlayMask<T>({
 			setContainerBounds({height, width, x, y});
 		}
 	}, [visible]);
+
+	useIsomorphicLayoutEffect(() => {
+		if (childrenRef.current) {
+			const updater = () => {
+				if (!childrenRef.current) {
+					return;
+				}
+
+				const {height, width, x, y} =
+					childrenRef.current.getBoundingClientRect();
+
+				setBounds({height, width, x, y});
+			};
+
+			return observeRect(childrenRef.current, updater);
+		}
+	}, [setBounds]);
 
 	return (
 		<>
@@ -168,9 +180,10 @@ export function OverlayMask<T>({
 					style={{
 						bottom: 0,
 						left: 0,
-						position: 'absolute',
+						position: 'fixed',
 						right: 0,
 						top: 0,
+						zIndex: 1040,
 					}}
 				>
 					<svg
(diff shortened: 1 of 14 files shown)


pat270:
I like the name `OverlayMask` more than `HighlightOverlay`. It's less ambiguous to me. Bootstrap uses `*-backdrop` for their overlay. Maybe we can use `BackdropOverlay` if you don't like `OverlayMask`.

packages/clay-popover/src/index.tsx
```diff
@@ -42,6 +42,28 @@ const ALIGNMENTS_MAP = {
...
+	brbl: LEFT_OFFSET,
+	brtr: TOP_OFFSET,
+	clcr: RIGHT_OFFSET,
+	crcl: LEFT_OFFSET,
+	tcbc: BOTTOM_OFFSET,
+	tlbl: BOTTOM_OFFSET,
+	tltr: RIGHT_OFFSET,
+	trbr: BOTTOM_OFFSET,
+	trtl: LEFT_OFFSET,
+};
```
diegonvs, line 65: Hey @matuzalemsteles, why this is not moved to @clayui/shared? I think it can be reused in other domAlign usages around Clay

packages/clay-popover/src/index.tsx
```diff
@@ -69,6 +91,11 @@ interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	 */
 	show?: boolean;
 
+	/**
+	 * Callback for setting the offset of the popover from the trigger.
+	 */
+	offset?: (points: Point) => [number, number];
```
diegonvs, line 97: I will never infer offset will be a callback. ~~Maybe we should apply something like `onOffSet`~~

packages/clay-popover/src/index.tsx
```diff
@@ -114,20 +144,19 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 			ref = popoverRef as React.Ref<HTMLDivElement>;
 		}
 
-		const show = externalShow ? externalShow : internalShow;
+		const show =
+			typeof externalShow === 'boolean' ? externalShow : internalShow;
```
diegonvs, line 149: I don't like naming as internal* external*. Why not `initialShow`?

diegonvs: Also, it looks like you're defining show is a boolean on IProps. why are you checking if it's a boolean or not?

diegonvs: Oh, because it can be `undefined`?

packages/clay-core/src/overlay-mask/OverlayMask.tsx
```diff
@@ -0,0 +1,218 @@
...
+import React, {useEffect, useLayoutEffect, useRef, useState} from 'react';
+
+type Bounds = {
+	height: number;
+	width: number;
+	x: number;
+	y: number;
+};
+
+const M = (x: number, y: number) => `M ${x} ${y}`;
```
diegonvs, line 20: What is M and L?

Machine Learning or Mercado Livre 😂 I'm joking but I would understand what is behind it

packages/clay-core/src/overlay-mask/OverlayMask.tsx
```diff
@@ -0,0 +1,218 @@
...
+};
+
+const initialBounds = {
+	height: 0,
+	width: 0,
+	x: 0,
+	y: 0,
+};
+
+const useIsomorphicLayoutEffect =
```
diegonvs, line 103: Are we rendering this component on server side too? Sorry for the nit but I would know

packages/clay-popover/src/index.tsx
matuzalemsteles: Yes, also if it is `false` it will toggle, in the end this should be replaced by default controlled and uncontrolled in PR #4786

packages/clay-popover/src/index.tsx
matuzalemsteles: That's true, I'm not sure about `onOffset`, maybe yes if I can't find another name I'll go with it.

packages/clay-core/src/overlay-mask/OverlayMask.tsx
```diff
@@ -0,0 +1,218 @@
...
+						<g>
+							<defs>
+								<clipPath id="overlayMask">
+									<path
+										clipRule="evenodd"
+										d={createClipPath(
+											internalBounds,
+											containerBounds,
+											padding
+										)}
```
diegonvs, line 203: SVGs 💘 my soul

packages/clay-core/src/overlay-mask/OverlayMask.tsx
```diff
@@ -0,0 +1,218 @@
...
+								fillOpacity={0.8}
+								height="100%"
+								onClick={onClick}
+								width="100%"
+								x={0}
+								y={0}
+							>
+								<animate
+									attributeName="opacity"
+									dur="0.3s"
```
diegonvs, line 220: Are these animation configs configurable?

packages/clay-popover/src/index.tsx
matuzalemsteles: Yeah, it's good to move there, not sure yet because we only have two places using this, so maybe it's not as necessary now, maybe when everything goes into `@clayui/core` package

packages/clay-core/stories/OverlayMask.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+import {Provider} from '../src';
+import {OverlayMask} from '../src/overlay-mask';
+
+type Props = {
+	onClick: () => void;
+};
+
+// Use only for demo.
+const Dot = ({onClick}: Props) => (
+	<div
```
diegonvs, line 24: image

I think this component is just used on story but ppl use to hard copy and paste those examples. Why we don't use ClaySticker with `sticker-outside` here? 🙈

packages/clay-core/src/overlay-mask/OverlayMask.tsx
matuzalemsteles: 😂  just a shortcut to not replicate but you are right I think I should add the most exact word for it

packages/clay-core/src/overlay-mask/OverlayMask.tsx
matuzalemsteles: Yeah, `useLayoutEffect` doesn't render on the server causes some errors, basically it came from here #3956

packages/clay-core/src/overlay-mask/OverlayMask.tsx
matuzalemsteles: Not for now, following the same overlay definition for Modal, for now, just following the spec.

packages/clay-core/stories/OverlayMask.stories.tsx
matuzalemsteles: I thought about using something else but I was a little afraid of making an impression that people would think we are not following the recommendation, the comment's attempt was to try to prevent people from copying I hope it works, and also the ugliness of it should help 😂

matuzalemsteles:
> I like the name OverlayMask more than HighlightOverlay. It's less ambiguous to me. Bootstrap uses *-backdrop for their overlay. Maybe we can use BackdropOverlay if you don't like OverlayMask.

Oh a good one too, I'm just not sure if this sounds weird but glad you like that nomenclature too, for now, I'll stick with it. Thank you Patrick!

diegonvs (approved):
LNHTM!

Issue #4621, opened 2022-02-02
Add more tests to FocusScope to increase coverage


matuzalemsteles:
Bryce even added some tests for this component to cover the Portals use case, but we have other edge cases, like in #4617, it would be interesting to build more tests for this component, how to handle shift+tab and the focus features using the arrow keys which is also provided.

Some scenarios in mind that we might want to add:

- Interactions with shift+tab
- Move focus using right, left, up, and down arrow keys
- Move focus outside the React tree. This is more the case of DXP when we render only DropDown for example and we need to return the focus control to the browser to the closest element, this has two scenarios, a component that uses `React.Portal` and a second that doesn't use it

sergiojimcos:
Hey @matuzalemsteles! I'm going to start working on this issue. I will take a view of all the tests scenarios that you have mentioned. I think the last one is a little bit confused but if I have any doubt I will ask you, as well as if I have any idea for new test cases I will included it.

PR #4794, opened 2022-04-06
feat(@clayui/clay-shared): Add more test cases to FocusScope

sergiojimcos:
**What is new?**
Adding more test cases to FocusScope:
1. Some interactions using `tab + shift`.
2. Some interactions using `Arrow Up` and `Arrow Down`
3. Some interactions using `ClayPortal` to move the focus outside the React Tree

**Issues:**
#4621

### DIFF
--- a/packages/clay-shared/src/__tests__/FocusScope.tsx
+++ b/packages/clay-shared/src/__tests__/FocusScope.tsx
@@ -3,31 +3,12 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import ClayDropDown from '@clayui/drop-down';
 import {cleanup, fireEvent, render} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 
 import '@testing-library/jest-dom/extend-expect';
 import {ClayPortal, FocusScope} from '..';
-
-const DropDownWithState: React.FunctionComponent<any> = ({
-	children,
-	...others
-}) => {
-	const [active, setActive] = React.useState(false);
-
-	return (
-		<ClayDropDown
-			{...others}
-			active={active}
-			onActiveChange={(val) => setActive(val)}
-			trigger={<button>Click Me</button>}
-		>
-			{children}
-		</ClayDropDown>
-	);
-};
 
 describe('FocusScope', () => {
 	afterEach(() => {
@@ -252,7 +233,7 @@ describe('FocusScope', () => {
 			document.body.innerHTML = '';
 		});
 
-		it('interacts with React.Portal', () => {
+		it('interacts with React.Portal', async () => {
 			const {getByText} = render(
 				<FocusScope>
 					<div>
@@ -264,50 +245,36 @@ describe('FocusScope', () => {
 							</div>
 						</ClayPortal>
 
-						<DropDownWithState>
-							<ClayDropDown.ItemList>
-								{[
-									{href: '#one', label: 'one'},
-									{href: '#two', label: 'two'},
-								].map((item, i) => (
-									<ClayDropDown.Item
-										href={item.href}
-										key={i}
-										spritemap="/foo/bar"
-									>
-										{item.label}
-									</ClayDropDown.Item>
-								))}
-							</ClayDropDown.ItemList>
-						</DropDownWithState>
+						<ClayPortal>
+							<FocusScope>
+								<div id="content">
+									<button>one</button>
+									<button>two</button>
+								</div>
+							</FocusScope>
+						</ClayPortal>
 					</div>
 				</FocusScope>
 			);
 
 			const button1El = getByText('Button 1');
 			const button2El = getByText('Button 2');
-			const clickMeEl = getByText('Click Me');
 
-			userEvent.tab();
+			await userEvent.tab();
 
 			expect(button1El).toHaveFocus();
 
-			userEvent.tab();
+			await userEvent.tab();
 
 			expect(button2El).toHaveFocus();
 
-			userEvent.tab();
-			fireEvent.click(clickMeEl);
-
-			expect(clickMeEl).toHaveFocus();
-
-			userEvent.tab();
+			await userEvent.tab();
 
 			const oneEl = getByText('one');
 
 			expect(oneEl).toHaveFocus();
 
-			userEvent.tab();
+			await userEvent.tab();
 
 			const twoEl = getByText('two');
 
@@ -322,37 +289,23 @@ describe('FocusScope', () => {
 							<button>Button 1</button>
 						</div>
 
-						<DropDownWithState>
-							<ClayDropDown.ItemList>
-								{[
-									{href: '#one', label: 'one'},
-									{href: '#two', label: 'two'},
-								].map((item, i) => (
-									<ClayDropDown.Item
-										href={item.href}
-										key={i}
-										spritemap="/foo/bar"
-									>
-										{item.label}
-									</ClayDropDown.Item>
-								))}
-							</ClayDropDown.ItemList>
-						</DropDownWithState>
+						<ClayPortal>
+							<FocusScope>
+								<div id="content">
+									<button>one</button>
+									<button>two</button>
+								</div>
+							</FocusScope>
+						</ClayPortal>
 					</div>
 				</FocusScope>
 			);
 
 			const button1El = getByText('Button 1');
-			const clickMeEl = getByText('Click Me');
 
 			userEvent.tab();
 
 			expect(button1El).toHaveFocus();
-
-			userEvent.tab();
-			fireEvent.click(clickMeEl);
-
-			expect(clickMeEl).toHaveFocus();
 
 			userEvent.tab();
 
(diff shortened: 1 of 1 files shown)


packages/clay-shared/src/__tests__/FocusScope.tsx
```diff
@@ -108,10 +168,185 @@ describe('FocusScope', () => {
...
+
+			dropdownButton.click();
+
+			userEvent.tab();
+
+			expect(dropdownButton).toHaveFocus();
+
+			userEvent.tab();
+
+			expect(htmlButton1).toHaveFocus();
```
matuzalemsteles, line 309: Hey @sergiojimcos I think this behavior is wrong, it should focus on DropDown links and then focus on `button1`, I think we found an edge case here 😅, I did a test and it seems that I can only reproduce it the first time then it works normally.

I would say switch to this expected behavior and add a `skip` to this test so it doesn't crash this PR and then we work on fixing that behavior and remove the `skip` in other PR.

matuzalemsteles: Could this be because opening DropDown and pressing tab close the dropdown? I think if this happens it makes sense but it was also to focus on `button1` first ah unless the focus is on the button toggle due to the click.

matuzalemsteles (commented):
Hey @sergiojimcos good job here again, I just added a comment about a test case to investigate if this is the behavior, maybe it's a bug.

packages/clay-shared/src/__tests__/FocusScope.tsx
sergiojimcos: Hey @matuzalemsteles! It makes sense because I was testing it manually and the focus closes the Dropdown. I'm going to investigate a bit more about this. If I don't find any fix, I'll update it to `skip` this test and check it later in another PR.

sergiojimcos:
Hey @matuzalemsteles! I have investigated deeply what was the problem and the error was that I clicked in the component (`dropdown.click()`) not dispatching the event using `userEvent`. So now it navigates with the Dropdown links correctly. I've uploaded the last two test cases that uses the Dropdown 😁

packages/clay-shared/src/__tests__/FocusScope.tsx
```diff
@@ -298,15 +297,31 @@ describe('FocusScope', () => {
...
 
 			expect(dropdownButton).toHaveFocus();
 
+			const itemTwo = (document.body as HTMLElement).querySelector(
+				'a[href="#two"]'
+			);
+
 			userEvent.tab();
 
 			expect(htmlButton1).toHaveFocus();
```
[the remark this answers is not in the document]
sergiojimcos: Hey @matuzalemsteles! I've tried using the `FocusScope`  in the Dropdown to maintain the order as you said but it doesn't work. The order passes to the `button1` and continues to the links as the recent commit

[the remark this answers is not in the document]
matuzalemsteles: @sergiojimcos I did a quick review on this, I think this is an unexpected edge case, I'll try to explain what's going on.

Well, as I said we have to add the `FocusScope` as a wrapper of the elements that we want to control the focus, otherwise it will not work, because who does all the control of the focus is this component, the use of `ClayPortal` it changes where the element will be rendered in the DOM, this is to solve problems of dropdown menus that can be cut when some element with `overflow: hidden` exists inside a container, this is a strategy we did but the problem that this breaks the tab because the tab follows the declaration order of elements in the DOM, so we create `FocusScope` to manage the focus i.e. keep the tab order correctly, for example, consider this declaration order:

```jsx
<div>
  <button>Button 1</button>
  <div>
    <button>Button 2</button>
  </div>
  <button>Button 3</button>
</div>
```

This is a very simple case but just to illustrate, the tab order will follow the declaration order, so if you press the tab sequentially the focus order will be, `Button 1 -> Button 2 -> Button 3`.

The problem is that often the order of declaration may not make sense to the user, because the visual positioning may be different depending on the layout of the page, in scenarios where we use `ClayPortal` which breaks with the order of declaration, for example:

```jsx
<div>
  <button>Button 1</button>
  <ClayPortal>
    <button>Button 2</button>
  </ClayPortal>
  <button>Button 3</button>
</div>
```

This represented in the DOM is different:

```html
<body>
  <div id="root">
    <div>
      <button>Button 1</button>
      <button>Button 3</button>
    </div>
  </div>

  <button>Button 2</button>
</body>
```

So if there is no focus manager when pressing the tab sequentially it will be `Button 1 -> Button 3 -> Button 2`, then this is not intuitive. Then the focus manager will correct and maintain the tab order correctly. The scenario from this test case is a little different, as DXP is not an application completely built on top of React, React is used more as loose components and only the products are the application of the page this makes in many scenarios the ` FocusScope` will not act or be able to control the tab order, it only controls what is declared as `children` ie it listens for any tab event if it was pressed within the scope.

To better understand what is happening here, following this test markup, the representation in the DOM will be:

```html
<body>
  <div id="root">
    <div>
        <div class="dropdown dropdown-test">
          <button class="dropdown-toggle">Click Me</button>
        </div>
    </div>
  </div>

  <div id="content">
    <button id="button1"></button>
  </div>

  <div class="dropdown-menu">Menu content here</div>
</body>
```

The focus manager scope here is being the dropdown-toggle, button1 and the dropdown-menu, if you don't have a focus manager the sequential tab order here will be: `Click me -> button1 -> links in the dropdown menu` , we don't want this order because it breaks accessibility and will cause user confusion, so with the focus manager in theory the sequence order should be `button1 -> Click me -> links in the dropdown` the problem with this scenario is that as ClayPortal is rendered first ie it is the first element in the declaration order when the focus comes from outside the FocusScope has no way of detecting that a tab key was pressed inside the scope because the event goes to the window and we only listen for the event within the scope, this causes the focus order to be different than expected, `click me -> links in the dropdown -> button1`, because who controlled this focus was the browser that followed the declaration order in the DOM, this is the problem here because it's not following this behavior.

In short, this is an unexpected edge case because we don't expect the person to add a ClayPortal without some trigger, as this is internal to Clay we have full control over it, so basically every declared ClayPortal has some element before being declared in scope what makes it work well, for example, see this scenario.

```jsx
<FocusScope>
  <div>
    <button id="button1">Button 1</button>

    <ClayPortal>
      <div id="content">
        <button id="button2">Button 2</button>
      </div>
    </ClayPortal>

    <DropDownWithState>
      <ClayDropDown.ItemList>
        {[
          {href: '#one', label: 'one'},
          {href: '#two', label: 'two'},
        ].map((item, i) => (
          <ClayDropDown.Item
            href={item.href}
            key={i}
            spritemap="/foo/bar"
          >
            {item.label}
          </ClayDropDown.Item>
        ))}
      </ClayDropDown.ItemList>
    </DropDownWithState>
  </div>
</FocusScope>
```

This scenario will work as expected, `Button 1 -> Button 2 -> DropDown toggle -> links in the dropdown` this is because when the focus arrives within the scope now who is managing the focus is our component. I hope this is clearer, I'm going to send a commit to switch to this scenario. Another important thing about this is that this component also controls the reverse order as well.

matuzalemsteles:
Hey @sergiojimcos I sent a commit updating the test scenarios to the expected behavior but still, the tests are failing, I think for some reason in Jest the focus management is getting lost on what the next element will be, it doesn't happen by reproducing manually, so it seems to be something more jest related I think. I'll keep investigating this tomorrow but if you want to check it out too feel free.

sergiojimcos:
Hey @matuzalemsteles! thank you so much for all the information that you have offered me ❤️ . I have understood everything even having some knowledges about `ClayPortal` in order to make these test cases. But everything is clear, I will take a view of the tests cases that fail

sergiojimcos:
Hey @matuzalemsteles! I've been investigating what is wrong with the two test cases, as you said they must follow the focus flow, but the `Dropdown` only focus the second item. As well as, I have added more than two items in the Dropdown and it fails too. I think that it is related to a Jest problem. I have also checked the snapshots if they are correct and the methods to `cleanup()` every test case and everything it ok. I also think about if the problem is the combination of `ClayDropdown` with the `FocusScope`, could be that `FocusScope` breaks the `ClayDropDown.Items` focus flow?

matuzalemsteles:
Thanks @sergiojimcos for the investigation, this makes sense I will take a look at it, the strange thing is that this works fine reproducing manually 🥲.

matuzalemsteles:
Hey @sergiojimcos I managed to "solve" the problem related to the test case I added more info in the commit comment on the reasons but basically, the `FocusScope` that is used in DropDown doesn't seem to be the same as the `FocusScope` used in the test this causes a reference problem for the context we use to solve the nested focus manager conflict problem, for example:

```jsx
<FocusScope>
  <div>
    <FocusScope>
      <div></div>
    </FocusScope>
  </div>
</FocusScope>
```

The scope of both can collide at some point causing some problems of skipping the focus twice, for example, we use the context to know if there is a manager declared as a parent to prevent event propagation so we prevent the parent from controlling the focus and avoid effects collaterals.

In summary, the problem seems to be some configuration of our jest that is causing this, for now I refactored the implementation to not use DropDown but I did the same simulation so we got very close here to the cases we want to test.

PR #4797, opened 2022-04-07
feat(@clayui/popover): add new size and style

### LINKED ISSUE #4790: Make flexible popover to allow wider widths and configurable header
This is part of the Delta team's feature request to use Popover to build the Walkthrough component. I did some analysis of the document and to support this component we need some adjustments in Popover.

- The definition doesn't have how wide the popover should be, it doesn't look like it should be `auto` either. We have to check this with Lexicon if there is any recommendation for the maximum size.
- Button with the `times` icon is included in the popover header to allow closing, in the definition it says that `small` size should be used for the Button.
- The Header doesn't have the border and the internal spacing seems to change as well to get closer to the body of the popover.

In general, they are finer adjustments, I was able to reproduce the Popover by doing some Clay comps, I just changed the hardcoded Popover width to check what we have. See this example on [codesandobox](https://codesandbox.io/s/quick-start-clay-finale-forked-fjy3j4?file=/src/index.js).

```jsx
<ClayPopover
  alignPosition="right-top"
  header={
    <ClayLayout.ContentRow noGutters verticalAlign="center">
      <ClayLayout.ContentCol expand>
        <span>Step 1 of 3: Customize logo</span>
      </ClayLayout.ContentCol>
      <ClayLayout.ContentCol>
        <ClayButtonWithIcon
          aria-label="close"
          displayType="unstyled"
          small
          symbol="times"
        />
      </ClayLayout.ContentCol>
    </ClayLayout.ContentRow>
  }
  style={{ maxWidth: "26rem" }}
  trigger={<ClayButton>Logo</ClayButton>}
>
  <p>
    Place your name and logo using this widget and make then a link the
    Homepage.
  </p>

  <ClayLayout.ContentRow noGutters verticalAlign="center">
    <ClayLayout.ContentCol expand>
      <ClayCheckbox label="Don't show this again" />
    </ClayLayout.ContentCol>
    <ClayLayout.ContentCol>
      <ClayButton.Group spaced>
        <ClayButton displayType="secondary" small>
          Previous
        </ClayButton>
        <ClayButton small>Got it</ClayButton>
      </ClayButton.Group>
    </ClayLayout.ContentCol>
  </ClayLayout.ContentRow>
</ClayPopover>
```
Screen Shot 2022-04-04 at 7 06 17 PM

matuzalemsteles:
Fixes #4790

Well, that basically just adds two new classes to the popover, for the header to be borderless and allow a wider size for the popover. We don't have a clear definition if @liferay-lexicon wants to add different widths for the popover, in the Walkthrough component it has two different sizes so I'm not sure, I'll wait for Lexicon to confirm before merging that.

@pat270 I would also like your opinions here, I'm not sure I missed something regarding the patterns, and also feel free to provide other alternatives.


drakonux:
Our idea was more related to giving freedom to designers for setting proper sizes instead of providing a list of predetermined sizes.  

Obviously, the component should provide a minimum and maximum size and we can work on that if you want @matuzalemsteles and provide you a couple of values.

matuzalemsteles:
Hmm got it, makes sense. I think keeping the sizes flexible is a bit bad for us because we would have to ask teams to add specific classes to determine their width, normally we want to avoid shapes like this but if that makes sense we can go with that and offer at least a new width bigger.

packages/clay-css/src/scss/mixins/_popovers.scss
```diff
@@ -85,6 +88,10 @@
 			.inline-scroller {
 				@include clay-css(map-get($map, inline-scroller));
 			}
+
+			&.popover-width-wider {
+				@include clay-css(map-get($map, wider));
+			}
```
matuzalemsteles, line 94: @pat270 maybe the naming for this class could be `.popover-width-lg` what do you think? maybe it should better fit the size standards we have?

pat270: `popover-width-lg` does fit better, we should do that.

pat270:
Hey @matuzalemsteles, I was thinking about creating a variant popover component like `.popover-walkthrough` and applying the styles to that instead of allowing the teams to compose the popovers individually. It gives less flexibility to the teams, but allows us to consistently style it across the board from the CSS.

Individually composing subcomponents in popovers should be done in a separate ticket (if we want it).

matuzalemsteles:
> I was thinking about creating a variant popover component like .popover-walkthrough and applying the styles to that instead of allowing the teams to compose the popovers individually. It gives less flexibility to the teams, but allows us to consistently style it across the board from the CSS.

@pat270 I'm not sure about this because I didn't want to embed this component in Clay, it's a very specific use case that in theory wouldn't be reused by several teams, in fact this will be a global component in DXP the teams wouldn't use directly or build the popover markup or use in your own modules. So I'm not sure. What do you think? I think this could be a less decoupled thing in this case.

pat270:
Ok this makes sense. I think we can get rid of the `borderless` option for `popover-header` in Clay CSS. Bootstrap has a `border-0` that gets rid of borders or we can add `border-transparent`.

matuzalemsteles:
> Ok this makes sense. I think we can get rid of the borderless option for popover-header in Clay CSS. Bootstrap has a border-0 that gets rid of borders or we can add border-transparent.

We'd also have to remove the padding because it's smaller when it doesn't have a border.

pat270:
We could add `transparent` to https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/variables/_utilities.scss#L267. We could even use `border-white`. This still makes me feel uneasy. Any redesign of popover is going to ruin the walkthrough popover styles. Can we also get away with only adding it to cadmin?

matuzalemsteles:
Yeah,this is getting a little misaligned here but adding the borderless might make sense because it adjusts the padding, I think doing a reskin of the entire popover would lead to having a variant for a specific use case that probably wouldn't change much, in that case just the wider width and the header without embroidery and the padding adjustable to avoid a very large white space.

I'm a bit against creating a new variant because we would have to add this to our popover implementation, for example, it would turn into a `displayType` with `walkthrough` or some other name that is disfellowshipped from that component because that would say that Clay is supporting this component which is not our intention, on the contrary, we want people to create it using the popover.

pat270:
Yeah I don't like adding a specific `displayType` either. I was hoping we could just pass it through the `className` attribute instead of making something specific to it.

matuzalemsteles:
@pat270 I see, well, that would be an option the problem is that we can have a certain inconsistency of functionality between CSS and Components, I wanted to try to keep the two close in the future, but I worry a little when we have many but options only in CSS.

Well, let me know what you think is best here, maybe we can go on to implement this style just for CSS but I think it would be good to avoid adding something like `walkthrough` because it implies that we will support all styles that are expected of the component, in addition to the header and width adjustments.

pat270:
Sounds good, I'll pull this down and take a deeper look. I can send another pr and we can figure out what we want and don't want.

matuzalemsteles:
Good, feel free to send it on top of this PR.

pat270:
Hey @matuzalemsteles, I know you didn't want to support `displayType`, but I couldn't get around it due to the `box-shadow` on the walkover popover being blue instead of black.
walkover

I added a variant `popover-secondary`. It styles the header and close button to match. From Clay CSS, we are now able to generate any number of variants from the Sass map for `popover`, `popover-header`, `popover-width-*`, and `popover-body`. I would like the react component to allow for this. Let me know what you think.

pat270:
@matuzalemsteles no problem, is it possible to add `headerDisplayType` and `bodyDisplayType` or something along those lines to allow custom classes to be passed to those components? I also need to update the Clay Admin side of things and add documentation.

matuzalemsteles:
> is it possible to add headerDisplayType and bodyDisplayType or something along those lines to allow custom classes to be passed to those components?

Well it would be possible but I think in this case it would be good to avoid having something like this to define for each element, in this scenario I think I prefer the `displayType` that can change the component's look.

Issue #4783, opened 2022-03-31
Avoid naming properties with the same name of DOM Attributes


diegonvs:
## What is your proposal?

This discussion cames from https://github.com/liferay/clay/pull/4779 where we have a property called `title` and we needed to omit the default `title` DOM attr and consider it as a React Prop using `Omit<...`.

**The idea here is to add a guideline on Clay components for avoiding naming properties with names like DOM attrs.**

## Why would adopting this proposal be beneficial?

More consistency and if users would use a desired DOM attr, they will be able to do it

## What are the alternatives to this proposal?

Add an eslint rule? It should be investigated

sergiojimcos:
Hey @matuzalemsteles! I'm going to start investigating this issue

sergiojimcos:
Hey @matuzalemsteles! I've made a research about the proposal made by @diegonvs. I think that the most useful way to prevent using React props naming the same way as DOM attributes is using an `eslint` rule called:[ `react/forbid-component-props`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/forbid-component-props.md). This rule only applies to Components and not DOM nodes. In order to select which properties you want to avoid, you have to costumize this rule using a list like in this example:

```
'react/forbid-component-props': [
	'error', {
		forbid: ['title'],
	},
],
```

In the array of `forbid` you have to indicate all the props that you want to avoid. Besides, you can also customize the array specifying the property name, an optional custom message, and a component whitelist:

```
{
  "propName": "someProp",
  "allowedFor": [SomeComponent, AnotherComponent],
  "message": "Avoid using someProp"
}
```
However, there is not any `eslint` rule that can avoid props naming all the DOM attributes without specified them in the array

matuzalemsteles:
Cool, this seems like a good way to help avoid this, as we talked in slack I think a good thing now is to try to understand what the alternatives would be for existing components, how this will affect them, and how we can migrate them, so the doc do RFC can help us to understand more deeply what this can cause side effect and how we can avoid it.

PR #4813, opened 2022-04-19
docs: Create proposal to avoid DOM naming props in Clay components

sergiojimcos:
**What is new?**
Creating a new document proposal to avoid DOM naming props in Clay components

**Issues**
#4783

### DIFF
--- a/docs/proposals/0003-avoid-DOM-naming-props.md
+++ b/docs/proposals/0003-avoid-DOM-naming-props.md
@@ -8,7 +8,7 @@
 
 ## Introduction
 
-In Clay we have some components like `<ClayEmptyState />`, `<ClayMultiStepNavWithState />`, `<ClayColorPicker />`... which have props named like DOM attributes (eg. `title`, `value`...)
+In Clay we have some components like `<ClayEmptyState />`, `<ClayMultiStepNavWithState />`, `<ClayColorPicker />`... which have props named like DOM attributes (eg. `title`, `value`, `length`, `size`...)
 
 This document describes the implementation proposal to avoid naming some props like DOM attributes.
 
@@ -83,13 +83,13 @@ In the complete way, you can determine only the chosen DOM attributes in an obje
 'react/forbid-component-props': [
 	'error', {
 		forbid: [{
-        message: "In this file includes title prop",
-        propName: "title",
-    },
-    {
-        message: "In this file includes value prop",
-        propName: "value",
-        }],
+			message: "In this file includes title prop",
+			propName: "title",
+    	},
+    	{
+			message: "In this file includes value prop",
+			propName: "value",
+    	}],
 	},
 ],
 ```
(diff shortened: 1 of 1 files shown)


docs/proposals/0003-avoid-DOM-naming-props.md
```diff
@@ -0,0 +1,123 @@
...
+
+Once you insert this example in `.eslintrc.js` file and save it, then go to the console and execute the following command:
+
+```
+yarn lint
+```
+
+It will show you all the files where Clay components have got prop names like DOM ones with their specific message.
+
+## Effect on API resilience
```
matuzalemsteles, line 105: I wanted to try here to understand more how we could get around the cases that we are already using these attributes, for example, `title`, are we going to rename the current properties? will they be deprecated now or maybe it would be better to move this to v4? I think some properties we might want to treat as an exception for example `value` which we are defaulting to in [RFC 0002 Controlled and uncontrolled components](https://github.com/liferay/clay/blob/a68735bcc31eee6ec9269fc37d6f4dda8f6c7f81/docs/proposals/0002-controlled-and-uncontrolled.md).

docs/proposals/0003-avoid-DOM-naming-props.md
```diff
@@ -0,0 +1,123 @@
...
+		forbid: [{
+        message: "In this file includes title prop",
+        propName: "title",
+    },
+    {
+        message: "In this file includes value prop",
+        propName: "value",
+        }],
+	},
+],
```
matuzalemsteles, line 94: just one typo, I think the formatting broke here.

matuzalemsteles (commented):
Hey @sergiojimcos this turned out really cool, thanks for working on it, I just added a comment to clarify how this affects what we already have and what we can do.

docs/proposals/0003-avoid-DOM-naming-props.md
sergiojimcos: Hey @matuzalemsteles! I've just check it and I think that it is CLI GitHub visualisation problem. I'm going to change a white space in order to maintain the formatting consistency

sergiojimcos: I have made some changes in formatting and the problem occurs GitHub CLI code review that the formatting broke

matuzalemsteles:
Well before merging this, I wanted to try to see the other side of the coin regarding this change to make sure this would be a good decision to make and we are considering the edge cases.

There's the point where we might also want to use DOM properties to extend resources and not necessarily take away from their default behavior, e.g. EmptyState has the `title` property which is common for the `HTMLElement` interface, basically all elements, components as EmptyState is an aggregate of HTML markup to render an element at a high level, in theory using `title` as a DOM resource for this scenario might not make sense, it would be used for the tooltip but this is not a rule for this component, following this reasoning if the API were not title it would be added to the root element of the component which would create a tooltip which would be against the spec, in this scenario we would have to add the `title` using the `title` of the component and the attribute on the text element so that the tooltip is created when it passes exactly over the component's title. Well, this is a scenario where using a DOM attribute would not be a problem.

Also has a scenario like `size` is a very common nomenclature for people to reason about the different sizes of a component when used in a primitive way for an element like `input` is to define the width in visible characters, to `select` the number of options visible in the selector. So it depends on your context and just for those scenarios. We can use another name as an alternative to this name but it needs to be consistent.

I'm not sure if this should really be a general rule of thumb for all components, I can't find a certain pattern. I wanted to hear from @bryceosterhaus also your opinion, maybe @diegonvs can add more stuff here too.

bryceosterhaus (commented):
Thanks for your work on this @sergiojimcos! And sorry I am just seeing this now, @matuzalemsteles, I had lost my notification for my mention on this and never saw it.

I left a comment asking for some more examples of why exactly this is a problem for clay. From my personal experience, I can't remember having run into this being an issue for react components and naming props the same as DOM attrs. I suspect that there are areas of concern where we might need something like this, but I am concerned that if we make this a wide-spread rule then we are going a bit overboard.

I think my main concern is that there are a lot of DOM attributes(170+ I read on one site) and some of those are really useful nomenclature for components(`type`, `size`, `name`, `title`, `onclick`, `disabled`, etc). It seems like overkill to re-think some of these and create other non-dom names.

Additionally, this isn't a common practice in the React world, so are we over complicating it or addressing something that doesn't need to be fixed at this scale? Are there particular edge cases where we do need to create some guidelines or rules instead of a wide-spread change?

matuzalemsteles:
I'm closing this PR because there are still too many points of uncertainty to go forward with this, especially now that Clay is working on accessibility improvements that we likely avoid some of the properties or use it to provide OOTB features.

PR #4826, opened 2022-04-27
fix(@clayui/css): Tables use calc() to make fatter border widths in t…

### LINKED ISSUE #4818: $table-border-width should accept a css variable as a value.
Currently if $table-border-width is set to a css variable clay will not compile due to it being used inside sass calculations. These sass calculations should be switched to using the css calc function to fix this. 

This can be reproduced by setting the following `$table-border-width: var(--spacer-2);` and trying to build.

pat270:
…head and tbody

Tables still need a revamp, but custom properties should work now.

/cc @ethib137 

fixes #4818

### DIFF
--- a/packages/clay-css/src/scss/components/_tables.scss
+++ b/packages/clay-css/src/scss/components/_tables.scss
@@ -107,7 +107,7 @@ caption {
 	}
 
 	tbody + tbody {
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM! We had a file conflict here, I think it was due to the PRs I just merged.

PR #4827, opened 2022-04-28
fix(@clayui/css): Atlas .btn.active:focus should have outline

### LINKED ISSUE #4804: @clayui/css: Mixins clay-button-variant should be able to style .active:focus
`.btn-primary.active` doesn't receive focus border when navigating with the TAB key.

pat270:
- interpolate multiple box-shadows and transitions
- use the global $component-focus-box-shadow where applicable
- Mixins clay-button-variant should be able to style .active:focus

fixes #4804


matuzalemsteles (approved):
LGTM!

PR #4832, opened 2022-05-02
feat(@clayui/css): Globals adds `$c-root` map for outputting custom properties

### LINKED ISSUE #4805: @clayui/css: We should be able to define any custom property in _root.scss through _clay_variables.scss
`_root.scss` only defines a few custom properties. We should be able to define any custom property we like through variables.

https://github.com/liferay/clay/blob/a39732f6268f660a4e19375a70c59c7552a4bbe9/packages/clay-css/src/scss/components/_root.scss#L1-L20

pat270:
Update: We can output custom variables in `:root {}` with:
###### _clay_variables.scss
```
$c-root: (
    --primary-l2: $primary-l2,
    --input-color: $input-color,
    --some-custom-color: green,
);
```

This adds the map `$c-root` for outputting any number of custom properties in the `:root` selector. If `$c-root` is used, we won't output the original custom properties provided by Bootstrap. I think this keeps the api more consistent.

Marking this as a draft for now, I'm waiting for feedback on the private slack thread.

fixes #4805

### DIFF
--- a/packages/clay-css/src/scss/_variables.scss
+++ b/packages/clay-css/src/scss/_variables.scss
@@ -66,3 +66,5 @@
 @import 'variables/_type';
 
 @import 'variables/_utilities';
+
+@import 'variables/_c-root';
--- a/packages/clay-css/src/scss/atlas/_variables.scss
+++ b/packages/clay-css/src/scss/atlas/_variables.scss
@@ -32,6 +32,8 @@
 @import 'variables/_menubar';
 @import 'variables/_navbar';
 
+@import 'variables/_c-root';
+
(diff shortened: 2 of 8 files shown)


marcoscv-work:
I love this idea!

matuzalemsteles:
Hey @pat270 Do you still think this is valid or do you still want to go ahead with it?

pat270:
@matuzalemsteles lets go ahead with it. This will be beneficial for users who want to use Custom Properties.

matuzalemsteles:
Great, merging this!

PR #4840, opened 2022-05-03
docs(RFC): RFC 0004 Collection component

### LINKED ISSUE #4112: Explore the possibility of standardizing collection components

matuzalemsteles:
Closes #4112

This PR is an early draft for the Collection RFC, I still have a lot to write and detail on the implementation design as well as the functionality, alternatives considered and how it affects the implementation, and how we can get to implement this still in v3 or v4.

I worked on a sketch with a diagram to demonstrate some ideas about the design of this internal component, so this helps provide a little more context.

> Better reading when expanding the image

Collection component

I wanted to send this earlier to get some ideas on this as I develop my thinking on this so by the end of this week a lot here could change.

### [View the rendered text](https://github.com/matuzalemsteles/clay/blob/issue-4112/docs/proposals/0000-collection-components.md)

### DIFF
--- a/docs/proposals/0000-collection-components.md
+++ b/docs/proposals/0000-collection-components.md
@@ -127,21 +127,136 @@ Related:
 
 ## Detailed design
 
-![Collection component](https://user-images.githubusercontent.com/13750819/166341856-7c55ab6e-5c82-4071-b9cb-35e720bdd008.png)
+![Collection component](https://user-images.githubusercontent.com/13750819/166605103-fed4553d-0f85-4f27-8fc6-fe87c3044582.png)
 
-Describe the solution design in enough detail for someone familiar with Clay to understand and for someone familiar with the implementation to be able to implement it. If a new API is involved, show the full API and its documentation comments detailing what it does.
+The component's internal implementation must allow for two levels of composition that are principal and essential for Collection, static and dynamic content. The API between dynamic and static content is syntactically close together to avoid a big rewrite when moving from static to dynamic.
 
--   Static data
--   Dynamic data
--   Async loading
-    -   ErrorBoundary
-    -   Suspense
--   Data agnostic
--   Virtualization for large collections
+```jsx
+// Static content
+<Collection>
+  <Item />
+  <Item />
+  <Item />
+</Collection>
+
+// Dynamic content
+<Collection<Items> items={items}>
+  {(item) => <Item />}
+</Collection>
+```
+
+This example above is just a demonstration of the syntax of both possibilities but the developer will not use this component exactly this is what we will implement in the components that need Collection. For example, DropDown:
+
+```tsx
+// (!) Approximate behavior
+
+type Props<Items> = {
+  children: React.ReactNode | ChildrenFunction<Items>;
+  items?: Array<T>;
+  trigger: React.ReactElement;
+};
+
+function DropDown<Items>({trigger, items, children}: Props<Items>) {
+  return (
+    <Root>
+      {trigger}
+      <Menu>
+        <Collection<Items> items={items}>
+          {children}
+        </Collection>
+      </Menu>
+    </Root>
+  );
+}
+
+// (!) Approximate usage example
+/// Static content
+<DropDown>
+  <Item />
+  <Item />
+  <Item />
+</DropDown>
+
+/// Dynamic content
+<DropDown<Items> items={items}>
+  {(item) => <Item />}
+</DropDown>
+```
+
+Dynamic collections offer more features, such as optimizing rendering and implementing list virtualization, and both allow them to be data agnostic without defining a fixed data structure to render some element. This allows us and developers to compose with different hierarchies, e.g. groups or trees like TreeView, and have rich renderings with different markup styles.
+
+### Static collection
+
+Static collections are like the hardcoded list, usually used for elements like a Menu that are not configurable or dynamic or data coming from the user.
+
+```jsx
+<DropDown trigger={<Button>View</Button>}>
+	<Item>List</Item>
+	<Item>Table</Item>
+	<Item>Card</Item>
+</DropDown>
+```
+
+### Dynamic collection
+
+Unlike static collections, dynamic collections are data rendered in the list using a JSX-based interface where the amount of data is unknown and may change during the component's lifecycle.
+
+The `items` property is reactive, this can change according to the implementation and criteria of the component, for example `<TreeView />` data can be modified with user interaction, such as moving, deleting, or renaming the item, in this case, the items property follows [RFC 0002 Controlled and uncontrolled components](./0002-controlled-and-uncontrolled.md).
+
+```tsx
+<DropDown<Items> items={items}>{(item) => <Item>{item.name}</Item>}</DropDown>
+```
+
+#### Virtualization
+
+List virtualization is extremely important to prevent React.js don't lose performance with a large list of data being rendered. The implementation is to render only the visible elements in the viewport or the container where the element is being rendered.
+
+The virtualization implementations normally adjust with the size of the container, we can also provide if the container does not define any size automatically render only the elements considering the viewport.
+
+Virtualization also needs to consider that items can have dynamic sizes, virtualization must be offered in an OOTB way so that it is not necessary to configure any properties, obligatorily also all items being added in the Collection content must accept [`ref`](https://reactjs.org/docs/forwarding-refs.htm) so that it can calculate the element and auto-adjust.
+
+```tsx
+const Item = React.forwardRef(({children}, ref) => (
+	<div ref={ref}>{children}</div>
+));
+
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
This RFC is almost done, there have been few changes since when I opened this PR, so that's good because things are really solid about the proposal. I'll still tweak a few things in the document before getting this ready for review and collecting more feedback, but feel free to add feedback.

matuzalemsteles:
Just saying this PR is ready for review 🙂!

dsanz:
Hey Matu,

I can clearly see the value of standardizing usage of collection-related component APIs. Nevertheless, in order to make this proposal clearer to me, I need help to understand the benefits of this from the point of view of dev exp in DXP and also the impact of this enhancements in terms of performance.

In terms of dev experience, I'd like to see how a FE in DXP can leverage these to reduce time and effort it takes to create end-user robust UIs. For example, you mention the network interactions, the suspense and the error handliing. Looks like proposal would facilitate Product team FE to develop these, in other words, component API would make this easier but that still would be the job for a FE to decide and implement. So, to illustrate this, it'd be very nice to take some DXP usage of a component and turn it into the new pattern, so that the benefits in terms of code consistency could be clearer.

As for the performance, I'd like to know how the proposal can help to create more performant interfaces. So, we talk about lazy loading and virtualization. Is there a DXP use case that could leverage these so that if these patterns are applied we get a significant perf improvement? Maybe an uncontrolled case of treeview?

Finally, when you provide a list of components, you mention TreeView, Autocomplete, DropDown, List, MultiSelect, Nav, NavigationBar, and VerticalNav. Would not the table component be in that list too? I kinda feel that data we put in a table could benefit from many of these things, indeed the FDS could borrow patterns like error boundary to support inline editing. In short, is there a reason why Table is not in the list of components?

Thanks so much!

matuzalemsteles:
@dsanz thanks for adding some comments here, I will clarify some of your questions, others would also be great to hear at the dev level in DXP but we can also infer some things with the data we have and why we are making these decisions based on what already exists so much in the DXP as in React.js standards.

> Impact of this enhancements in terms of performance.

We won't be able to have precise metrics now of how much this improves performance because this is still a proposal but we can already point to where we can use it to improve performance in two recent cases of Clay, TreeView, and Pagination.

For TreeView, see the [performance analysis document](https://github.com/liferay/clay/blob/1c633058690890452cd9e212faf034f1d455875c/docs/treeview_performance.md) we did when we finished the component, we have a lot of optimizations in terms of operation and data manipulation and avoiding rendering blocking, and removing blockers at render time but we don't handle the rendering of many items in the DOM.

For Pagination the scenario is similar but simple because it's a list instead of a nested list like TreeView, I thought we wouldn't have these scenarios but these two performance cases #4605 and #4886 were reported.

Both are related to the amount of items being rendered in the DOM, this is a very common problem in applications when list rendering and solutions already exist to solve this, virtualization. Some of the core of the React team has already created libs to deal with this, they are very old solutions...

- [react-window](https://github.com/bvaughn/react-window)
- [react-virtualized](https://github.com/bvaughn/react-virtualized)

There are other newer libs in the community that create agnostically, but in addition to taking time to render any state update change that causes a rerender, it also impacts how it reflects on the DOM, and can even cause screen crash problems for example.

> For example, you mention the network interactions, the suspense and the error handliing. Looks like proposal would facilitate Product team FE to develop these, in other words, component API would make this easier but that still would be the job for a FE to decide and implement.

Yes, dealing with data and handling errors and the case of Suspense must be decided at the application level and not at the component level, we cannot infer that here, because the way of managing states and dealing with the data request depends on the strategy of each product and how they handle the data. The idea is to facilitate this integration with Clay components, add examples in the documentation as we did with VerticalBar because these are common patterns of React itself and are also native components of React.js.

I've seen some use cases in DXP, using Suspense, actually, I added this when I migrated from Forms to React to load dynamic fields from Forms, so we used the default of Suspense and ErrorBoundary to load the fields on demand and when they are used and capture a network error to give the user the possibility to try again. Also, the VerticalBar implementation that Greg wrote at the time in the PageEditor migration was using both strategies as well.

[This blog post by React.js explains very well the change in the mindset of how to deal with data and create a good user experience](https://reactjs.org/blog/2019/11/06/building-great-user-experiences-with-concurrent-mode-and-suspense.html), I recommend reading it.

The thing is, we don't use it much in DXP either because some developers don't care much about these details of lazy loading, splitting the bundler for the application to load faster, or dealing with the application cycle of loading and dealing with errors. There's a lot here that we're not even dealing with, which are the new scenarios of competing components that React started to introduce in version 18.x.

> Would not the table component be in that list too? I kinda feel that data we put in a table could benefit from many of these things, indeed the FDS could borrow patterns like error boundary to support inline editing. In short, is there a reason why Table is not in the list of components?

I believe so, for sure. There was no reason, I actually forgot to mention it. But yes, you can take advantage of it in several ways to improve performance and UX if necessary.

---

The points related to performance and DX, and as mentioned in the document we want to offer OOTB features, such as virtualization without the developer having to configure extra things and be data agnostic to facilitate the use of components and avoid data being transformed to be passed to the component.

Today DXP developers always tend to use DropDown's low-level or high-level component, the point is that low-level allows them to customize the content and high-level is when they don't want to customize anything but over time we always added APIs to allow them to customize which was a bit against the idea of composition, as I explained in the proposal.

> The API between dynamic and static content is syntactically close together to avoid a big rewrite when moving from static to dynamic.

This phrase defines well the idea of how improving DX, data agnostic, pattern consistency and the same features for all components as well as decreasing the learning curve reflects on the DX of the components. So, in summary, every detail of this proposal and how it was thought are reflected in DX.

A good example of this level is how [VerticalBar](https://clayui.com/docs/components/vertical-bar.html) was developed, following these standards but not implementing all the features, you can have the component working when you write statically, without having to declare a state to deal with the visibility of panels, conditionals, and etc.

The performance as I mentioned above, there are many scenarios and there are different ways to deal with it, bundler split, lazy loading, suspense and etc... the way this is applied in applications depends on how it is structured, here we want to help developers to apply it very well and encourage them to use it with Clay but we can't provide everything OOTB because it becomes opinionated and won't work for all use cases, I believe that this way we can reach the level of abstraction that strikes a balance.

Some Design System cases that are reaching this level are, for example, [Adobe's React Spectrum](https://react-spectrum.adobe.com/index.html), [Radix](https://www.radix-ui.com/) as a visual agnostic solution but with a focus on accessibility.

dsanz:
Thanks for the response Matu, I still need to read it thoroughly, but basically, what I get from it is:

- By standardizing the collection API usage pattern across all clay components using it, we get a more consistent code which facilitates transitioning from static to dynamic collections, and enables usage of  {industry|react community}-standard modern practises to handle with common situations.
- But the job of actually applying these new practises needs a shared effort across DXP frontend developers so that they are encouraged to think about them in the context of the particular use case. That would allow them to improve performance too as clay components will still be low/mid level

matuzalemsteles:
@dsanz I think so, I would add a few more things which is what I expect from this as a result as well.

- Decrease the learning curve and accelerate development. Collection standardization is common for most components, we hope this helps the developer when learning in one component replicates the same learning in the other components, eg DropDown -> Autocomplete
- DX from the APIs. Standardization also helps DX APIs by being consistent across all components, we hope to deliver more OOTB features like virtualization for list optimization and being data agnostic to ease implementation in different products with different data and naming.

drakonux:
It's interesting. I mean if we are doing the same for many components, it makes sense to do this. 

ps. Liferay Front-end developers should be commenting on this thread IMO. Because they are the ones who consume Clay so they will have a clear vision if this would help them or not.

matuzalemsteles:
Just complemented here this is also written in this document, most of the changes are internal this is a proposal for the internal components that can benefit from having the same features and offering this to developers like virtualization and a consistent API pattern, this will not cause a breaking change and allows gradual adoption of developers to opt to use the API benefits, i.e. for low-level components to use these features and use almost the same OOTB features a high-level one would be using APIs like `items` and using the default render props for dynamic content. Components like Tabs can benefit from not having to define visibility states and conditions.

This has been better described in the document, you can see this pattern in action in components like TreeView and VerticalBar (just synthetically because it doesn't have the described features that Collection implements like virtualization). I'll be merging this PR this week and I'll be starting to work on it, once we've had this open for over 2 months waiting for feedback.

PR #4842, opened 2022-05-03
fix(@clayui/css): Mixins convert clay-range-input-variant to use clay-css mixin pattern

### LINKED ISSUE #4172: @clayui/css Mixin clay-range-input-variant convert to use clay-css pattern
The clay-css pattern allows more customization through Sass maps without having to declare a variable name for every property. It also must be backward compatible.

pat270:
fixes #4172


packages/clay-css/src/scss/variables/_range.scss
```diff
@@ -55,55 +55,119 @@ $clay-range-input: map-deep-merge(
...
-		data-label-after-content: unquote("'\\FEFF' attr(data-label-max)"),
-		data-label-after-right: 0,
-		hover-cursor: $link-cursor,
-		focus-outline: 0,
-		focus-thumb-box-shadow: $component-focus-box-shadow,
-		disabled-color: $clay-range-disabled-color,
-		disabled-cursor: $disabled-cursor,
-		disabled-progress-bg: $primary-l2,
-		disabled-track-bg: $gray-200,
+		clay-range-thumb: (
```
matuzalemsteles, line 58: @pat270 why not just keep `range-thumb` as well as the other properties to go without the Clay prefix?

pat270: It keeps it closer to the selector name, but thumb is kind of an outlier since it styles the browser default thumb.

matuzalemsteles: Oh I see, makes sense

matuzalemsteles (approved):
LGTM

PR #4844, opened 2022-05-04
feat(@clayui/modal): Add a children property to IProps as a ReactNode to avoid warning error

kevenleone:


matuzalemsteles (approved):
LGTM! Thanks @kevenleone!

PR #4847, opened 2022-05-06
fix(@clayui/css): Breadcrumbs convert to Clay mixin pattern

### LINKED ISSUE #4702: @clayui/css: Breadcrumb convert to Clay mixin pattern
We should convert breadcrumbs to use the Clay Mixin pattern for better customization via Sass maps. See https://github.com/liferay/clay/blob/c470badcc372b0fef21b97bc09cbd48cd0b251d0/packages/clay-css/src/scss/components/_breadcrumbs.scss

pat270:
fixes #4702


matuzalemsteles (approved):
LGTM

PR #4848, opened 2022-05-06
fix(@clayui/css): Alerts/Table deprecate `$alert-*-level` and `$table-*-level` variables

### LINKED ISSUE #4281: @clayui/css: Alerts Deprecate Not Useful Variables
Color modifiers aren't useful when theming. They are very specific to Bootstrap's colors. We should deprecate them.

```
$alert-bg-level
$alert-border-level
$alert-color-level
```

pat270:
I wanted to remove all uses of Bootstrap's `theme-color-level` function, but might be too breaking.

fixes #4281

### DIFF
--- a/packages/clay-css/src/scss/variables/_alerts.scss
+++ b/packages/clay-css/src/scss/variables/_alerts.scss
@@ -410,8 +410,16 @@ $alert-autofit-row: map-deep-merge(
 
 // Alert Color Levels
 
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM! Yeah, I think during this v3 version we just have to mark as deprecated and in v4 we can remove them.

PR #4849, opened 2022-05-06
fix(@clayui/css): Sidebar long text with no white space should break …

### LINKED ISSUE #3805: @clayui/css Sidebar component-title and component-title should use `overflow-wrap: break-word;`
Long text with no white space overflows the sidebar if `truncate-text-inline truncate-text` isn't used. It also overflows if the `autofit-row` component isn't used.

sidebar-overflow

pat270:
…to new line in component-title and subtitle

fixes #3805


matuzalemsteles (approved):
LGTM

PR #4850, opened 2022-05-06
fix(@clayui/css): Functions add the parameter $fallback to clay color functions

### LINKED ISSUE #4810: @clayui/css: We should be able to override the global setting $enable-clay-color-functions-process-fallback in our color functions
The setting `$enable-clay-color-functions-process-fallback: false` makes Clay return the CSS Custom Property because Sass isn't able to process them through any of its color functions. We should be able to override this setting on a per case basis. 

There are edge cases when we want to process the fallback value of a Custom Property like getting red, green, and blue values to use in `rgba()`.

https://github.com/liferay/clay/blob/ac7c86254ea29e11c040dbb61d64f365cd59f8c4/packages/clay-css/src/scss/functions/_color-functions.scss#L137-L140

should be:

```scss
@function clay-blue($color, $fallback: clay-color-functions-process-fallback()) {
	@if (is-css-var($color) and not $fallback)) {
		@return $color;
	}

	$color: clay-get-fallback($color);

	@return blue($color);
}
```

pat270:
fixes #4810


matuzalemsteles (approved):
LGTM

PR #4865, opened 2022-05-12
feat(@clayui/core): adds new API to expand node via double click in TreeView

### LINKED ISSUE #4863: Add a way to expand node via double click in TreeView
This is a requirement coming from the document https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1970078399/LEXI-1257+Clarify+behavior+for+DXP+Trees.

By default expanding the node is done by clicking on the node or the expander icon, in the document they want a behavior to allow expanding by clicking on the node with a double click only for multi-selection. Currently, we only have one way to bypass default behavior in the same way you bypass default browser behavior.

```jsx
<TreeView
  defaultItems={[...]}
>
  {(item) => (
    <TreeView.Item onClick={(event) => event.preventDefault()}>
      {item.name}
    </TreeView.Item>
  )}
</TreeView>
```

So you can ignore the default expand but we don't have an option for example to expand manually like the selection, so there are two different ways to provide this, one is to add a property to expand via double click and the other is to pass the expander to be manually triggered as the selection.

matuzalemsteles:
Closes #4863

There is another way to provide this like we do for selection but I think the expander is not needed now, we can see if there are use cases in the future and we can think about adding that possibility.

I'm adding the `expandDoubleClick` API that expanding by clicking on the node will only be done by double-click on the node, it is also possible to avoid this behavior but I think it is something that may not even make sense to be added because to have this behavior is enabled via props but I'm adding to keep the same behavior we have for `onClick`.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -38,6 +38,11 @@ interface ITreeViewProps<T>
 	dragAndDropContext?: Window & typeof globalThis;
 
 	/**
+	 * Flag to expand the node's children when double-clicking the node.
+	 */
+	expandDoubleClick?: boolean;
+
+	/**
 	 * Flag to expand child nodes when a parent node is checked.
 	 */
 	expandOnCheck?: boolean;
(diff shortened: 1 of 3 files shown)


matuzalemsteles:
I made some changes to how to handle double-clicking. Using `onDoubleClick` and `onClick` in the element will not work and will have different behavior depending on the browser, another strategy is to use `event.detail` to know how many clicks the click event had, this is according to the browser and the user configuration in the operating system so we don't need to create a setTimeout mechanism to identify the double click.

PR #4867, opened 2022-05-13
feat: add chromatic for visual testing

### LINKED ISSUE #4802: Explore using visual tests for components in the storybook
Our QA team is evaluating adding this to our CI but using ocular, the problem is that the tool is very simple and requires a lot of work from the QA team to update and maintain the baselines as well as it can slightly increase the time of the CI in DXP, we want to try using this outside the DXP repository and using it directly in Clay we can have a better tool to analyze visual tests and it will be easier to maintain, as well as being part of the component development process.

Our QA team is also interested in this move to bring this to Clay.

matuzalemsteles:
Closes #4802

Well I did some exploring with [Chromatic](https://www.chromatic.com/) and we don't really need to do any configuration, this is [very simple and OOTB with storybook](https://storybook.js.org/docs/react/writing-tests/visual-testing) is pretty cool.

The problem is that you need a Cloud service from [Chromatic](https://www.chromatic.com/) itself, you can't run something locally as far as I know but there is a [free plan that offers 5,000 snapshots per month](https://www.chromatic.com/pricing), this is good because so far we have 192 snapshots.

One way to not hit this limit is to run this UI test only on every workflow release instead of running it on every pull request.

I already made an account and it syncs the `liferay/clay` collaborators in [Chromatic](https://www.chromatic.com/) to have access to the build and validate the changes to update the baseline if necessary, you just need to create an account with your GitHub account that collaborates with Clay.

Another point is that I ignored some stories from the storybook so as not to generate visual tests, some are related to the charts and the data provider which is not necessary. I think for better results we need to improve some stories to show for example open menus and dropdowns for DropDown, ColorPicker, and Datepicker, and better examples to show use cases for example typography.

> To run the test in Clay's release automation workflow it is necessary to add the token in the secrets of GitHub as `CHROMATIC_PROJECT_TOKEN`, I will open a ticket for that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,10 @@ jobs:
             BROWSERSLIST_DISABLE_CACHE: 1
             CI: true
         run: yarn test
+      - name: UI Tests
+        env:
+          CHROMATIC_PROJECT_TOKEN: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+        run: yarn chromatic
 
   version:
     needs: test
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
 		"storybook:build": "NODE_OPTIONS=--max_old_space_size=8192 build-storybook -c .storybook -o .storybook-static",
 		"storybook": "start-storybook -p 8888",
 		"test": "jest",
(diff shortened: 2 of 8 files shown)


matuzalemsteles:
cc @john-co

bryceosterhaus:
> One way to not hit this limit is to run this UI test only on every workflow release instead of running it on every pull request.

This is great. Smart way to preserve the limit

matuzalemsteles:
The token has been added in secrets, I'm going ahead and merging.

PR #4872, opened 2022-05-18
feat(@clayui/core): calls onKeyDown on Item and ItemStack if defined

### LINKED ISSUE #4871: Proper way to handle keyboard events in the TreeView
I have a treeView in a modal to select a folder where an item will be moved. Currently the tree handles all the keyboards for interacting with the tree, but for example if I type `enter` nothing happens. 

Currently, I'm doing something like this: https://codesandbox.io/s/modest-turing-ozt23x?file=/src/index.js

It is a bit weird because none of KeyPress and KeyDown works, only KeyUp works.

What I want is the handleSelectionChange function to be called if the user is focusing the node and press either enter or space.

Is there any better way to do this?

matuzalemsteles:
Fixes #4871

This PR I'm adding two new features to the TreeView, one is that it allows to use of the `onKeyDown` callback in the Item and ItemStack components and it is also possible to ignore the default behavior of the component, The second one is adding a new `onSelect` API in the TreeView which is similar to `onSelectionChange` will always be called when there is a selection of an item but passing as a parameter the properties of the item being selected.

It seems to be a default behavior that happens a lot in DXP because cases need item properties to save, this I think could be changed, it's an overload of information being sent between methods unnecessarily as well as being sent to the backend in some At times, the ideal would be to better control the information states and work only with the id of the selected items is less parameter overhead and makes implementations within DXP much simpler, there is a problem is that some components are very isolated, such as selection in Modal, communication between components also becomes more complex with contracts to send all between components but this is something very difficult to solve now and change all cases in DXP.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -17,7 +17,10 @@ import {Icons, TreeViewContext} from './context';
 import {ITreeProps, useTree} from './useTree';
 
 interface ITreeViewProps<T>
-	extends Omit<React.HTMLAttributes<HTMLUListElement>, 'children'>,
+	extends Omit<
+			React.HTMLAttributes<HTMLUListElement>,
+			'children' | 'onSelect'
+		>,
 		ITreeProps<T>,
 		ICollectionProps<T> {
 	/**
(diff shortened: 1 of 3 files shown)


matuzalemsteles:
cc @victorg1991

victorg1991:
Nice, thanks!! 😍

Issue #4861, opened 2022-05-12
Adjust button size for TreeView


matuzalemsteles:
This is a request coming from some document requirements https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1970078399/LEXI-1257+Clarify+behavior+for+DXP+Trees, the Button small component has a smaller size when used in the TreeView.

pat270:
There are some issues with content breaking to newline when the treeview container is small. We have no guidelines on how to handle it. Should everything be on one line? Should it break to new line?

Not related to this issue, but one thing I'd like to change is the quick actions on hover overlaying the content instead of pushing it over, similar to what we are doing in tables https://clayui.com/docs/components/table/markup.html.

matuzalemsteles:
> There are some issues with content breaking to newline when the treeview container is small. We have no guidelines on how to handle it. Should everything be on one line? Should it break to new line?

Well, that's a good question I think the @liferay-lexicon team can help us here.

> Not related to this issue, but one thing I'd like to change is the quick actions on hover overlaying the content instead of pushing it over, similar to what we are doing in tables

Makes sense to me!

matuzalemsteles:
Hey @pat270 it seems that this button also only appears when the mouse hovers over the node or the node is focused, maybe for that, we can create an extra class just to handle this? we can also let the DXP teams handle this because this is not at the core of Lexicon for now.

pat270:
@matuzalemsteles sure we can create a utility for it, that would make it more composable.

emiliano-cicero:
Hey guys, I just want to add my thought on this which is similar to what @matuzalemsteles has written in his last post. The button shouldn't be covered by Lexicon/Clay. As part of the System, we should only provide the component and its extension points. 
For this specific case I'd say we provide the simple Tree View, and then a side extension point where teams can add what they need (Buttons, Badges, Actions, etc) and is not for us to define. 

**Here's an example:** 
<img width="370" alt="image" src="https://user-images.githubusercontent.com/27729173/169273802-870dced4-2b08-4c4f-b629-fd84ab7947d3.png">

As you can see I haven't even added a margin because badge components might need to be closer to the text, that margin should also come from the teams.

pat270:
@emiliano-cicero are you ok if we add a button size for treeview? It will be scoped specifically for treeview, we don't need to add a new button definition to Lexicon.

PR #4875, opened 2022-05-18
chore(@clayui/core): Add usage examples with TreeView

### LINKED ISSUE #4869: Add some examples of TreeView usage in Storybook
This issue is just to add composition examples in the TreeView to demonstrate the different use cases in the document https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1970078399/LEXI-1257+Clarify+behavior+for+DXP+Trees so that the developers can visualize how to implement it for example and how it can reach the different cases with the TreeView.

matuzalemsteles:
Closes #4869

I'm adding some examples of using the TreeView [following the recommendations of the document](https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/1970078399/LEXI-1257+Clarify+behavior+for+DXP+Trees), there are still some details of the button that will be made by issue #4861 and some behavior details like pressing enter on the focused node should select the children if exists a toggle button.

I'm leaving this as a draft to wait for the adaptations of the button to the TreeView.

### DIFF
--- a/packages/clay-core/stories/TreeView.stories.tsx
+++ b/packages/clay-core/stories/TreeView.stories.tsx
@@ -4,6 +4,7 @@
  */
 
 import ClayAlert from '@clayui/alert';
+import Badge from '@clayui/badge';
 import Button from '@clayui/button';
 import {ClayDropDownWithItems as DropDownWithItems} from '@clayui/drop-down';
 import {ClayCheckbox as Checkbox, ClayInput as Input} from '@clayui/form';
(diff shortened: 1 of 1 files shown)


pat270:
I had some ideas about how to add badges and other things inside a `treeview-link`. I mentioned them at the end of https://github.com/liferay/clay/pull/4883#issue-1243701401.

> Lastly I'm going to add another pattern where we use autofit-row with component-text to position buttons, badges, and whatever else right next to the text

I'm assuming we can replace the Drive text with markup at https://clayui.com/docs/components/treeview.html#single-level. This should free us from having to maintain code for that. Drive would be replaced by:
```
<span class="autofit-row">
    <span class="autofit-col autofit-col-shrink">
        <span class="text-truncate-inline">
            <span class="text-truncate">Drive</span>
        </span>
    </span>
    <span class="autofit-col">
        <span class="badge badge-primary">8</span>
    </span>
</span>
```

matuzalemsteles:
> I'm assuming we can replace the Drive text with markup at https://clayui.com/docs/components/treeview.html#single-level. This should free us from having to maintain code for that. Drive would be replaced by:

The problem with this is that we abstract the extra markup from `autofit`, we do it so people don't have to write a lot, so each item added as a child of `TreeView.Item` or `TreeView.ItemStack` is automatically wrapped in a `autofit-col`. https://github.com/liferay/clay/blob/6110c3efaff157cd88dedbae034c845dec54da3d/packages/clay-core/src/tree-view/TreeViewItem.tsx#L539

I think that to achieve the truncate of the text we can leave it via composition that the developer can do, for example:

```jsx
<TreeView.Item>
  <Checkbox />
  <Icon symbol="folder" />
  <Text truncate>Drive</Text>
  <Badge label="8" />
</TreeView.Item>
```

PR #4883, opened 2022-05-20
fix(@clayui/css): Treeview `btn` should be 24px tall and adds show-quick-actions-on-hover

### LINKED ISSUE #4861: Adjust button size for TreeView

pat270:
fixes #4861

@matuzalemsteles I kept going back and forth on how to display these buttons.
component-action
btn

We need to add `show-quick-actions-on-hover` to `treeview` and `quick-action-item` to the buttons/links we want to hide. The markup is below:

The first image:
```
<div class="treeview-link" role="treeitem" tabindex="-1">
    <div class="c-inner" tabindex="-2">
        <div class="autofit-row">
                <span class="component-text">
                    <span class="text-truncate-inline">
                        <span class="text-truncate">Categorization</span>
                    </span>
                </span>
            </div>
            <div class="autofit-col">
                <button class="btn btn-monospaced component-action quick-action-item" tabindex="-1" type="button">
                    <span class="c-inner" tabindex="-2">
                        <svg>...</svg>
                    </span>
                </button>
            </div>
        </div>
    </div>
</div>
```
The second image:
```
<div class="treeview-link show" tabindex="0">
    <div class="c-inner" tabindex="-2">
        <div class="autofit-row">
                <span class="component-text">
                    <span class="text-truncate-inline">
                        <span class="text-truncate">Liferay Drive</span>
                    </span>
                </span>
            </div>
            <div class="autofit-col">
                <button class="btn btn-secondary quick-action-item" tabindex="-1" type="button">
                    <span class="c-inner" tabindex="-2">Check child nodes</span>
                </button>
            </div>
        </div>
    </div>
</div>
```

I wanted to get rid of https://github.com/liferay/clay/blob/28281e32a4743b882def29aa7b0521af96c9502f/packages/clay-css/src/scss/components/_treeview.scss#L137-L151 and https://github.com/liferay/clay/blob/28281e32a4743b882def29aa7b0521af96c9502f/packages/clay-css/src/scss/variables/_treeview.scss#L90. I left them in just in case, but we don't need it if we add `quick-action-item` to `component-action`s.

I decided to go with our current implementation where we place them in `autofit-col` instead of the way we do quick actions on table. The main reason is due to `background-color` on `treeview-link` for hover and active states. Opaque background-colors  don't work too well with overlays which we use on `treeview-dark`.

If for whatever reason we need an overlay, we should be able to add it progressively. The markup would look like:
```
<div class="treeview-link">
    <div class="autofit-row"></div>
    <div class="quick-action-menu">
        <button class="btn btn-secondary quick-action-item" type="button"></button>
    </div>
</div>
``` 

Lastly I'm going to add another pattern where we use `autofit-row` with `component-text` to position buttons, badges, and whatever else right next to the text like:
align-end-text
I'll open another issue regarding that.

### DIFF
--- a/clayui.com/content/docs/components/markup-treeview.md
+++ b/clayui.com/content/docs/components/markup-treeview.md
@@ -23,6 +23,7 @@ mainTabURL: 'docs/components/treeview.html'
 
 -   [Variations](#css-variations)
     -   [Expander on Hover](#css-treeview-expander-on-hover)
+    -   [Quick Actions on Hover](#css-treeview-show-quick-actions-on-hover)
     -   [Light](#css-treeview-light)
     -   [Dark](#css-treeview-dark)
 
@@ -70,24 +71,6 @@ The treeview provides a way to display information in a hierarchical structure b
 							<span class="text-truncate-inline" title="Liferay Drive"><span class="text-truncate">Liferay Drive</span></span>
 						</span>
 					</span>
-					<span class="autofit-col">
-						<button class="btn btn-monospaced component-action" tabindex="-1" type="button">
-							<span class="c-inner" tabindex="-2">
-								<svg class="lexicon-icon lexicon-icon-times-circle-full" focusable="false" role="presentation">
-									<use xlink:href="/images/icons/icons.svg#times-circle-full" />
-								</svg>
-							</span>
-						</button>
-					</span>
-					<span class="autofit-col">
-						<button class="btn btn-monospaced component-action" tabindex="-1" type="button">
-							<span class="c-inner" tabindex="-2">
-								<svg class="lexicon-icon lexicon-icon-ellipsis-v" focusable="false" role="presentation">
-									<use xlink:href="/images/icons/icons.svg#ellipsis-v" />
-								</svg>
-							</span>
-						</button>
-					</span>
 				</span>
 			</span>
 		</div>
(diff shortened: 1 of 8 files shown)


matuzalemsteles (approved):
LGTM! I think the use case of the button is more to be next to the text but I think we already support it here with this implementation it wouldn't have the spacing but I think we don't need to support this directly in the TreeView, I think we can let the developer do that, what I'm imagining here is that the developer can compose this in different ways and the spacing can change. I think just supporting the two cases here like adjusting the size of the button in the treeview and leaving it optional as an action or not is enough.

matuzalemsteles:
We can also get rid of `component-action` if you think that's better, at least for the component, the developer isn't using that markup because we abstracted it into the component

pat270:
@matuzalemsteles I'll get rid of the `component-action` stuff. It was hard to tell where the button should be from the confluence doc. If it needs to be right next to the text, I need to update the docs and provide some spacing. Maybe I can close https://github.com/liferay/clay/issues/4884 and just do everything here.

pat270:
@matuzalemsteles this should be ready to go

matuzalemsteles:
I'll send some commits on top of this PR to update the component as well.

matuzalemsteles:
Hey @pat270 in the case of the Button, couldn't we have it being rendered inside an `autofit-col` without being together with the `component-text`? to do this together with components would create a lot more markup and get a little more complex. For example:

```jsx
<TreeView.Item>
  <TreeView.ItemStack>
    <Icon />
    Root
    <Button>Check</Button>
  </TreeView.ItemStack>
</TreeView.Item>
```
This seems to be possible but the only problem is that as the button will have a higher height when it becomes visible it blinks due to the new spacing.

pat270:
@matuzalemsteles we could, but the user would need to figure out which combo of `autofit-col` utilities they need to use. It wouldn't be one size fits all. For the check all button case:

check-all-btn

```html
<div class="autofit-row">
  <div class="autofit-col"></div>
  <div class="autofit-col autofit-col-shrink">
    <div class="component-text"></div>
  </div>
  <div class="autofit-col">
    <button class="btn btn-secondary" type="button">Check</button>
  </div>
</div>
```
If we apply the above markup to the other quick actions (ellipsis, times-circle-full):

other-quick-actions

To make it layout correctly we need to change shrink to expand:
```html
<div class="autofit-row">
  <div class="autofit-col"></div>
    <div class="component-text"></div>
  </div>
  <div class="autofit-col">
    <button class="btn btn-monospaced component-action quick-action-item" type="button">x</button>
  </div>
  <div class="autofit-col">
    <button class="btn btn-monospaced component-action quick-action-item" type="button">e</button>
  </div>
</div>
```

If we wanted everything together:
everything

```html
<div class="autofit-row">
  <div class="autofit-col"></div>
  <div class="autofit-col autofit-col-shrink">
    <div class="component-text"></div>
  </div>
  <div class="autofit-col">
    <div class="badge badge-primary"></div>
  </div>
    <!-- we need this div below because it turns the button into a flex item + `flex-direction: column` makes the button fill the container -->
    <div class="autofit-section">
      <button class="btn btn-secondary" type="button">Check</button>
    </div>
  </div>
  <div class="autofit-col">
    <button class="btn btn-monospaced component-action quick-action-item" type="button">x</button>
  </div>
  <div class="autofit-col">
    <button class="btn btn-monospaced component-action quick-action-item" type="button">e</button>
  </div>
</div>
```

I felt like the right choice, from a user's perspective, was to nest the `autofit` utilities as it removes having to know all these quirks with flex items.

matuzalemsteles:
@pat270 yeah makes sense, for action cases I follow the same behavior as before so this won't change just for Button's case, I think that way is fine to go, no problem.

PR #4892, opened 2022-05-27
feat(@clayui/core): adds the implementation of the Vertical Bar component

### LINKED ISSUE #4112: Explore the possibility of standardizing collection components

matuzalemsteles:
Fixes #4112

I'm adding the implementation of the Vertical Bar component, there are some details that still need to be adjusted both in the CSS and in the JS part.

The component anatomy looks like this:

```jsx
// Static Content
<VerticalBar>
  <VerticalBar.Content>
    <VerticalBar.Panel>
      <div className="sidebar-header">
        <div className="component-title">Tag</div>
      </div>
    </VerticalBar.Panel>
  </VerticalBar.Content>

  <VerticalBar.Bar>
    <VerticalBar.Item>
      <Button displayType={null}>
        <Icon symbol="tag" />
      </Button>
    </VerticalBar.Item>
  </VerticalBar.Bar>
</VerticalBar>

// Dynamic Content
<VerticalBar defaultActive="Tag">
  <VerticalBar.Content
    displayType={args.contentDisplayType}
    items={items}
  >
    {(item) => (
      <VerticalBar.Panel key={item.title}>
        <div className="sidebar-header">
          <div className="component-title">{item.title}</div>
        </div>
      </VerticalBar.Panel>
    )}
  </VerticalBar.Content>

  <VerticalBar.Bar displayType={args.barDisplayType} items={items}>
    {(item) => (
      <VerticalBar.Item divider={item.divider} key={item.title}>
        <Button displayType={null}>
          <Icon symbol={item.icon} />
        </Button>
      </VerticalBar.Item>
    )}
  </VerticalBar.Bar>
</VerticalBar>
```

The component follows the same idea that was implemented in TreeView with static and dynamic content, so it is possible to render dynamic content too, this is a light implementation of RFC #4840. The implementation is already interactive without having to add states or having to add extra properties on the components to control which panel is active for example.

The detail here is that to define the position of the right and left side, it is necessary to pay attention to the rendering order of the `<Content />` and `<Bar />` components that visually reflect the markup rendering.

There are some CSS details that I don't think have been covered yet. @pat270 I can add this but I would like you to review it later.

- The Vertical Bar in the left position with the white panel, does not have a proper `box-shadow`.
- I don't think we cover the case of adding an item to the bar in the bottom position instead of always being on top.
- The dark color of the panel has only one dark color that is from the application bar, it needs to be the same color as the dark color of the bar.

### ToDo

- [ ] Adds tests
- [ ] Adds documentation
- [ ] Rendering with a panel active by default is not showing


pat270:
> The Vertical Bar in the left position with the white panel, does not have a proper box-shadow.

> The dark color of the panel has only one dark color that is from the application bar, it needs to be the same color as the dark color of the bar.

I can update these.

> I don't think we cover the case of adding an item to the bar in the bottom position instead of always being on top.

I think this is easy to fix as well. We just need to add `height: 100%` to `tbar-nav` inside a `c-slideout` with the modifier class `tbar-item-expand`.  If we want multiple `ul` inside `tbar-stacked` we would need to add some modifier classes for that case.

matuzalemsteles:
I think most of the issues were fixed just missing some CSS (#4904, #4900) tweaks that can be sent in another PR. Marking this PR as ready for review or merge.

matuzalemsteles:
I'm going ahead and merging this PR so we can already include it in this release. It's very likely that the look of this component will change, so that's also in Beta.

PR #4899, opened 2022-06-03
fix(@clayui/css): c-slideout should use display: flex on tbar-stacked for show/transition

### LINKED ISSUE #4896: @clayui/css: Add modifier class to `.tbar-stacked .tbar-nav` should have height 100%
We can't place items at the bottom of a vertical navigation with `tbar-item-expand` due to `tbar-nav` only being as tall as its contents. We need some modifier classes to get what we want.

pat270:
fixes #4896

@matuzalemsteles `tbar-item-expand` should work now for the Vertical Navigation.

### DIFF
--- a/packages/clay-css/src/scss/mixins/_slideout.scss
+++ b/packages/clay-css/src/scss/mixins/_slideout.scss
@@ -18,51 +18,43 @@
 
 @mixin clay-slideout-variant($map) {
 	@if (setter(map-get($map, enabled), true)) {
-		@include clay-css(setter($map, ()));
+		@include clay-css($map);
 
 		&.c-slideout-shown {
-			@include clay-css(setter(map-get($map, c-slideout-shown), ()));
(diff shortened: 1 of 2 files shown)


pat270:
I read the description again.
> I don't think we cover the case of adding an item to the bar in the bottom position instead of always being on top.
Did you mean?
align-to-bottom

matuzalemsteles:
@pat270 yeah, can we have both right? I've only seen a few use cases in the Lexicon documentation, I don't know if it still exists in DXP.

matuzalemsteles (approved):

pat270:
I used `mt-auto` on the `ul` element for that screenshot. `align-items-end` also works.

matuzalemsteles:
Oh I see, seems to work fine at least in this use case:

Screen Shot 2022-06-06 at 12 22 09 PM

PR #4906, opened 2022-06-06
chore(release): Adds input to define specific release version

matuzalemsteles:
I'm adding the possibility that we can set the release version hardcoded instead of following the conventional commit, we have the case that we want some breaking changes for example to update the visual of the loading indicator without increasing the major version.


matuzalemsteles:
I'm going to merge this so I can cut today's release.

PR #4917, opened 2022-06-09
fix(@clayui/css): Sidebar Light positioned on the left should have box-shadow on the right

### LINKED ISSUE #4904: @clayui/css: Sidebar positioned on the left side does not have the correct shadow
From #4892

Vertical Bar component has the possibility to position the Sidebar on the left side but when positioned the `box-shadow` is not corrected because by default it is only for the right side.

Screen Shot 2022-06-06 at 11 34 58 AM

pat270:
fixes #4904

@matuzalemsteles had to re-think this. We have widgets with sidebar's that could be positioned in the middle of the page. Box shadows on both sides would show.
widget-sidebar
I ended up defaulting the box-shadow to be on the right side and moving the box-shadow to the left if it's qualified by `sidenav-right`.


packages/clay-css/src/scss/atlas/variables/_sidebar.scss
```diff
@@ -60,7 +60,10 @@ $sidebar-light: map-deep-merge(
 	(
 		background-color: $white,
 		border-left-width: 0px,
-		box-shadow: -0.25rem 0 0.5rem -0.25rem rgba(0, 0, 0, 0.1),
+		box-shadow: 0.25rem 0 0.5rem -0.25rem rgba(0, 0, 0, 0.1),
+		sidenav-right: (
+			box-shadow: -0.25rem 0 0.5rem -0.25rem rgba(0, 0, 0, 0.1),
+		),
```
matuzalemsteles, line 66: @pat270 maybe the default couldn't be the box-shadow property of `sidenav-right`? It seems that it's more common to have a sidebar on the right side so we don't have to go out and update these components in DXP as well.

I think we do not use `sidenav-right` correctly everywhere in DXP.

pat270: @matuzalemsteles Good point. We will need to add `sidenav-left` or `sidenav-start` in this case.

matuzalemsteles: We might as well use `sidenav-left` since we already have `sidenav-right`, but we also have another nomenclature like `c-slideout-start` for VerticalBar. I don't know if we can use it or is it a separate class for another purpose.

pat270: I can add aliases so we will have `sidenav-left`, `sidenav-start`, `sidenav-right`, `sidenav-end`.

PR #4923, opened 2022-06-10
fix(@clayui/css): Global Variables remove math function parenthesis from $spacers

### LINKED ISSUE #4922: @clayui/css: Negative margin utilities have no value
###### clay.css
```css
.mt-n1,
.my-n1 {
  margin-top: !important;
}
```

We use the `math-sign` function to get the opposite value for negative utilities. 

https://github.com/liferay/clay/blob/f1106e376a0287572b41439187e1f304b9f9a4e5/packages/clay-css/src/scss/components/_utilities.scss#L362-L388

The math sign function only checks if the type is `number` or `string` and returns a `null` value if it's any other type.

https://github.com/liferay/clay/blob/f1106e376a0287572b41439187e1f304b9f9a4e5/packages/clay-css/src/scss/functions/_global-functions.scss#L97-L107

The `$spacers` map has values of type `list` for anything greater than 0. 🤔 

https://github.com/liferay/clay/blob/f1106e376a0287572b41439187e1f304b9f9a4e5/packages/clay-css/src/scss/atlas/variables/_globals.scss#L199-L236

We need to remove the trailing comma in `($spacer * #,)` values. The trailing comma is added by prettier during `yarn format`. The parenthesis for spacer values are used to group together the math function, but prettier mistakes it as a map key value pair.

pat270:
fixes #4922

- throws warning if type is not number, string, and parameter value is equal to null

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_globals.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_globals.scss
@@ -201,36 +201,16 @@ $spacers: () !default;
 $spacers: map-deep-merge(
 	(
 		0: 0,
-		1: (
-			$spacer * 0.25,
-		),
-		2: (
-			$spacer * 0.5,
-		),
-		3: (
-			$spacer * 1,
-		),
-		4: (
-			$spacer * 1.5,
-		),
-		5: (
-			$spacer * 3,
-		),
-		6: (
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
LGTM

PR #4927, opened 2022-06-13
fix(lexicon-ux): 1.x Updates content property values to conform to what is allowed by OAW Tracker

georgel-pop-lr:
- https://issues.liferay.com/browse/LPS-147457
- use fixed versions for dependencies

related to https://github.com/liferay/clay/pull/4688

issue https://github.com/liferay/clay/issues/4683

I have sent a change also on alloy-font-awesome: https://github.com/liferay/liferay-frontend-projects/pull/951

But not sure if we use generated files here, or we get them with some dependency I could not find.

### DIFF
--- a/src/fonts/alloy-font-awesome/scss/_icons-alloy.scss
+++ b/src/fonts/alloy-font-awesome/scss/_icons-alloy.scss
@@ -1,7 +1,7 @@
 /* Use the following CSS code if you want to use data attributes for inserting your icons */
 [data-icon]:before {
 	font-family: 'fontawesome-alloy';
-	content: attr(data-icon);
+	content: unquote("'\\FEFF' attr(data-icon)");
 	speak: none;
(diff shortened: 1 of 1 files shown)


pat270 (changes requested):
@georgel-pop-lr the build directory is the wrong place to change it. It will get overwritten. You should change it at `src/fonts/alloy-font-awesome/scss` instead.

georgel-pop-lr:
> @georgel-pop-lr the build directory is the wrong place to change it. It will get overwritten. You should change it at `src/fonts/alloy-font-awesome/scss` instead.

Hi @pat270,

Yes, my mistake! updated.

Regards.

pat270:
LGTM this is working by passing an html entity to the data-icon attribute (e.g., `data-icon="&#xF006"`).

pat270 (approved):

PR #4940, opened 2022-06-20
chore(clay-css): Use Bootstrap files from src directory instead of no…

### LINKED ISSUE #4939: Clay CSS 2.x use Bootstrap files from src directory instead of node_modules
TS needs custom changes in the Bootstrap source.

pat270:
…de_modules

@matuzalemsteles I think this should update the Bootstrap source on next release, but I'm not sure how the release process works so I didn't confirm it.

fixes #4939

### DIFF
--- a/packages/clay-css/gulpfile.js
+++ b/packages/clay-css/gulpfile.js
@@ -14,8 +14,6 @@ var config = {
 	SRC_GLOB: 'src/**/*'
 };
 
-var bootstrapPath = path.dirname(require.resolve('bootstrap/package.json'));
-
 var license = require('./tasks/copyright_banner');
 var tasks = require('require-dir')('./tasks');
 
@@ -45,9 +43,6 @@ gulp.task(
 		runSequence(
 			'compile:clean',
 			'compile:files',
-			'compile:bootstrap:js',
-			'compile:bootstrap:scss',
-			'compile:popper',
 			'build:svg:scss-icons',
 			'compile:prep-scss',
 			'compile:css',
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
@pat270 I think just generating the build could already confirm if this would fix it. Could it be that we changed a lot in the bootstrap files before but it wasn't reflected because of that and changing that now could have a side effect?

pat270:
No the only changes I saw were the changes made for this specific issue.

packages/clay-css/gulpfile.js
```diff
@@ -14,7 +14,7 @@ var config = {
 	SRC_GLOB: 'src/**/*'
 };
 
-var bootstrapPath = path.dirname(require.resolve('bootstrap/package.json'));
+var bootstrapPath = path.join('.', 'src', 'bootstrap.scss');
```
matuzalemsteles, line 17: I imagine this might break the build https://github.com/liferay/clay/blob/6f4ba2f956ada96130fedaded52b5ec77f2606a2/packages/clay-css/gulpfile.js#L67-L80, can you check if it's breaking the build?

pat270: You're right let me update all the paths.

pat270:
@matuzalemsteles I removed the tasks that only copied Bootstrap from node_modules.

matuzalemsteles:
@pat270 can this break someone using bootstrap directly? Since we are removing this now do we still have the bootstrap modified?

pat270:
No, we still have Bootstrap's source at https://github.com/liferay/clay/tree/2.x/packages/clay-css/src/scss/bootstrap which includes the modification we made. If we need to update Bootstrap to a different version, we will need to fix conflict that will rise with the changes we added.

matuzalemsteles:
Ok I can see the source code but from what I understand the idea is not doing the sass compilation correctly, so I don't understand how this can fix it, there may be something I'm missing here in the build script and we are still building sass even reviewing these steps here in this PR.

pat270:
The compilation is done at https://github.com/liferay/clay/blob/2.x/packages/clay-css/gulpfile.js#L131-L159. The other compile tasks were actually copy from `node_modules` to the `src` directory. I think I named them `compile` to keep the name space the same.

PR #4941, opened 2022-06-20
fix(@clayui/css): Reference Mark vertical alignment when it is used d…

### LINKED ISSUE #4938: The required field icon is not aligned correctly
https://codesandbox.io/s/incorrect-alignment-required-icon-vzv1jf

asterisk2

The asterisk indicating that the field is required is not correctly aligned. It seems that the `vertical-alignment: middle` of the `.lexicon-icon` class is overriding the `vertical-alignment: super` of the `.reference-mark` class.

asterisk

The asterisk should be top aligned.

pat270:
…irectly on the icon

fixes #4938

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_type.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_type.scss
@@ -176,10 +176,11 @@ strong {
 // Reference Mark
 
 .reference-mark {
-	display: inline-block;
-	font-size: $cadmin-reference-mark-font-size;
-	position: relative;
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
Do we need to modify the markup yet? That's how it is for me.

Screen Shot 2022-06-20 at 2 56 48 PM

pat270:
I just made the CSS change so it works with the markup structure that we have. Now that I think about it, it would probably be better to add it to `.reference-mark.lexicon-icon` directly.

matuzalemsteles (approved):
LGTM

PR #4943, opened 2022-06-21
fix(@clayui/css): fix syntax error in conditional declaration

matuzalemsteles:
Hey @pat270 I'm submitting because this was causing the build to fail in Clay's DXP release, I believe that the way in which the conditional syntax was declared in Sass is causing it to break, which is strange because if I'm not mistaken we use sass build to compile.

For more information https://github.com/liferay-frontend/liferay-portal/pull/2394#issuecomment-1161006017


pat270:
LGTM! Thanks @matuzalemsteles. I always mess up the no space.

matuzalemsteles:
No problem, thanks @pat270!

PR #4947, opened 2022-06-22
fix(@clayui/css): Loaders and Cadmin Loaders don't generate placeholders for `$loading-animation-palette`

### LINKED ISSUE #4946: clayui.com: Loading Indicators should only show examples defined by Lexicon
https://clayui.com/docs/components/loading-indicator.html#squares

We currently show both types of loading  indicators in every size. Lexicon defined circle to only be 10px or 16px and squares to be 32px or 64px. We should only show those examples to better communicate Lexicon's intent.

Edit: Found some issues with Reverting to the Dotted Loading Animation code on the docs with Dart Sass. I thought `%loading-animation.loading-animation-light` wouldn't count as a compound selector, but I was wrong.

pat270:
fixes #4946

Dart Sass considers `%selector.selector1` a compound selector and causes Sass build to fail, updated the loading animations scss to not generate placeholders by default.

Updates clayui.com docs to only show sizes, shapes, and colors that are defined by Lexicon.

Updates Reverting to Dotted Animation code to also work on Dart Sass.


matuzalemsteles (approved):
LGTM

PR #4952, opened 2022-06-23
 fix(clay-css): Grid compare the breakpoint positions instead of the `$container-max-widths` values

pat270:
`$container-max-widths` are verified to be in ascending order we can use the positions instead of values

Percent values should now work. Use to test:

```scss
$container-max-widths:(
  xs: 90%,
  sm: 90%,
  md: 95%,
  lg: 95%,
  xl: 1200px
);
```


matuzalemsteles (approved):
LGTM

Issue #4949, opened 2022-06-23
[TreeView] New properties for Selected State and Click event on the container area 


drakonux:
# Context
The TreeView component was designed to cover many cases for Liferay products. So, by default, the component can't solve all the product cases. Instead, it needs to be set depending on the case.

This proposal concluded from a conversation in a Slack thread where two product teams were discussing how a node should behave on click (ExpandOnClick vs CheckOnClick). Initially, we agreed that both teams would use the default interaction, which could be a possible solution. But Lexicon sees benefit for the experience if we invest in additional development to solve the following cases. 

## Checked vs. Selected
If the TreeView is using the Single Selection mode. A node can be selected but not checked. 
If the TreeView is using one of the Multiple Selection modes. A node can be selected and/or checked. 
 
![Select vs Checked](https://user-images.githubusercontent.com/21216422/175260213-0ca439e7-3935-4390-b58d-683509bf5af8.png)

For example, the Selected state makes sense for the Single Selection mode. 

![Single Selection](https://user-images.githubusercontent.com/21216422/175309486-51329dbe-a964-4086-bfc4-24d9c74daab2.png)

But it doesn't work for the Multiple Selection mode except if it's used for a master-detail pattern — Disclaimer; product teams haven't proposed this case.

![Multiple Selection - Master-Detail](https://user-images.githubusercontent.com/21216422/175309523-1da81ff8-3547-4668-a118-148f2db3dbc8.png)

Thus, when the Multiple Selection mode is set, the hover state of a node makes no sense if it triggers the Selected state because, in most cases, the TreeView is not used to apply a Master-Detail pattern. 

![Multiple Selection](https://user-images.githubusercontent.com/21216422/175309570-bd5dc39d-49eb-4a95-b90e-a14db6fda68a.png)

So for the no Master-Detail pattern cases on Multiple Selection, it's weird to see a node selected because the hover state highlights the node but clicking it selects the node, and that's it. So if we prevent the Select state for nodes, we must also avoid the Hover state or **manage the Click event**. If-else we are tricking the user because nothing happens by interacting with the node.  

## Types of Nodes

As a reminder, we proposed the following types of nodes:

- Container: a parent node that can't be selected **nor checked**.
- Parent: a node with one or more child nodes. It can be expanded (open) or collapsed (close).
- End: a node that does not have child nodes; an end node may be either a root node or a child node. 

# Proposal

- Allow turning off the Selected state of nodes
  - Container nodes 
[...]

PR #4958, opened 2022-06-28
feat(@clayui/core): adds new APIs to TreeView to control node active state and expand

matuzalemsteles:
This PR exposes new APIs to allow controlling the TreeView's internal states so that the developer can cover more use cases than the default behavior, read the issue with proposal #4949 to understand the use cases. This PR doesn't have demos to demonstrate how to get to this use case but I'll be adding more in this PR.

Controlling a node's expansion state is similar to controlling a node's selection, APIs are exposed via render props as a low-level API.

```jsx
<TreeView>
  {(item, selection, expand) => (
    <TreeView.Item
      onClick={(event) => {
        clearTimeout(clickTimerRef.current);

        event.preventDefault();

        switch (event.detail) {
          case 1:
            clickTimerRef.current = setTimeout(() => {
              if (!selection.has(item.id)) {
                selection.toggle(item.id);
              }
            }, 200);
            break;
          case 2:
            if (!selection.has(item.id)) {
              selection.toggle(item.id);
            }
            expand.toggle(item.id);
          default:
            break;
        }
      }}
    >
      ...
    </TreeView.Item>
  )}
</TreeView>
```

This is an example of the scenario where the double click has a different behavior than the single click on the same node and the node selection is only considered as selected once.

In the future, we may consider supporting high-level APIs that allow us to deal with this but for now, it is difficult to support this now because they are isolated cases with this behavior and we need to identify patterns to offer a more stable API for this that does not conflict with the default behavior and It's not hard to switch between the behaviors but for now, I can't think of a way to offer that.

## ToDo

- [x] Add documentation
- [x] Add tests
- [x] Add some demos

### DIFF
--- a/packages/clay-core/docs/api-treeview.mdx
+++ b/packages/clay-core/docs/api-treeview.mdx
@@ -11,44 +11,14 @@ mainTabURL: 'docs/components/treeview.html'
 
 ## TreeView.Item
 
-<div class="table-responsive">
-	<table class="table table-autofit table-bordered table-striped">
-		<thead>
-			<tr>
-				<th>Property</th>
-				<th class="table-cell-expand table-cell-minw-150">Type</th>
-				<th>Required</th>
-				<th class="table-cell-expand table-cell-minw-150">Default</th>
-				<th class="table-cell-expand table-cell-minw-250">
-					Description
-				</th>
-			</tr>
-		</thead>
-		<tbody>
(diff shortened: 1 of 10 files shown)


matuzalemsteles:
I'm going ahead with this PR and submitting another PR #4964 that will deal with disabling the hover of a specific item.

PR #4959, opened 2022-06-28
feat(@clayui/drop-down): Allow setting a default selected value for the radiogroup

victorg1991:
This pr uses the value inside item to set a default value for the radio group in DropDownWithItems

### DIFF
--- a/packages/clay-drop-down/src/DropDownWithItems.tsx
+++ b/packages/clay-drop-down/src/DropDownWithItems.tsx
@@ -360,8 +360,9 @@ const RadioGroup: React.FunctionComponent<IItem & IInternalItem> = ({
 	name,
 	onChange = () => {},
 	spritemap,
+	value: defaultValue = '',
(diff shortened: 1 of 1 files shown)


packages/clay-drop-down/src/DropDownWithItems.tsx
```diff
@@ -360,8 +360,9 @@ const RadioGroup: React.FunctionComponent<IItem & IInternalItem> = ({
 	name,
 	onChange = () => {},
 	spritemap,
+	value: initialValue = ''
```
matuzalemsteles, line 363: @victorg1991 I think here this value could be called `defaultValue` just to avoid some confusion, what do you think?

victorg1991: Done :)

matuzalemsteles (approved):
LGTM

PR #4964, opened 2022-07-01
feat(@clayui/css): add support class to remove hover from treeview link

matuzalemsteles:
This is a complement to #4949, just adding in a separate PR because I'm not sure if we should go this way, I'm just adding the support class to the atlas but not adding it to `cadmin` yet, I'll add it later if we agree that this is the best solution.

@pat270 let me know what you think, the intention of this PR is to add a class that allows you to disable hover on a specific node.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -43,6 +43,11 @@ export interface ITreeViewItemProps
 	isDragging?: boolean;
 
 	/**
+	 * Flag to remove the visual of the hover over the item.
+	 */
+	noHover?: boolean;
+
+	/**
 	 * @ignore
 	 */
 	overPosition?: string;
@@ -166,6 +171,8 @@ export const TreeViewItem = React.forwardRef<
 								overTarget && overPosition === 'middle',
 							'treeview-dropping-top':
 								overTarget && overPosition === 'top',
+							'treeview-no-hover':
+								itemStackProps.noHover || nodeProps.noHover,
 						}
 					)}
 					disabled={itemStackProps.disabled || nodeProps.disabled}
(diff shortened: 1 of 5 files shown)


pat270:
@matuzalemsteles is this for the alternative solution?

> Alternative solution
The approach in this case. It's to not innovate. Thus, for the sake of interaction:

> The container area doesn't trigger the hover visual state if the node can't be selected (any node type).
Expand/Collapse a node is triggered via the expander button.
Check/Uncheck a node is triggered via the checkbox.

The only confusing thing to me about the name `treeview-no-hover` is that it sort of implies that quick actions and expander show on hover should be disabled if this class is on the parent element. I'm not too particular about it though.

We will need to add `color: $secondary` to `.treeview-light` and `color: $secondary-l1` to `.treeview-dark`. The `color: inherit` will take on whatever color the nearest parent element is.

matuzalemsteles:
> is this for the alternative solution?

Yeah, I imagine this can be combined in some way with other cases as well.

> The only confusing thing to me about the name treeview-no-hover is that it sort of implies that quick actions and expander show on hover should be disabled if this class is on the parent element. I'm not too particular about it though.

Yeah, that makes sense, do you have another name that could fit better?

> We will need to add color: $secondary to .treeview-light and color: $secondary-l1 to .treeview-dark. The color: inherit will take on whatever color the nearest parent element is.

It's true, I'll update

pat270:
I was thinking about removing hover styles from `treeview-light` and `treeview-dark` and attaching them to `treeview-selectable`.

matuzalemsteles:
Hmm, would `treeview-selectable` be the state of `active`, or are they different things? Because we can have the hover even if it is not selectable with the active state.

pat270:
This is confusing, we can just stick with `treeview-no-hover`.

matuzalemsteles:
Yeah, this is a little confusing. I ended up going and renamed the class to `treeview-no-hover` as well.

Issue #4926, opened 2022-06-13
Should native input validations avoided and be handled individually? 


ambrinchaudhary:
## What is your question?
HTML input offers native validations such as **required**, **min**, **max**, **type=“number”**… that when the user submits the form, this is invalidated and it is not sent to backend, showing native error messages in the wrong inputs. 

If I am not wrong, these validations were handled in portal by taglibs such as `<aui:input>` where attributes like required, min, max … could be set,  along with `<aui:validator>` taglib where you could define common validations like for emails, digits, and so on. 

Now that we should render react components and do not use that taglibs, I don’t see we have anything similar in portal and we should implement these validations in each input or form we use (because we do not want the native messages to be shown). When having a form, the guide is to use Formik (https://github.com/liferay/liferay-frontend-projects/issues/348), but still the validations must be implemented in each component.

Also for the look and feel, we are setting the classes and managing the errors in each component we implement, like: 
```
<form>
      <div class={error ? ‘has-error’: ‘’}
	      <ClayInput {…props} >
      </div>

       {error && (
              <ClayForm.FeedbackGroup>
                  <ClayForm.FeedbackItem>
                      <ClayForm.FeedbackIndicator symbol="exclamation-full" />

                      {error}
                  </ClayForm.FeedbackItem>
              </ClayForm.FeedbackGroup>
          )}
</form>
```

I think this at least should be provided by Clay, avoiding us to implement this everywhere an input needs a validation. 

So my question here is: is this what we really want, to implement every validation whenever we need? Shouldn’t Clay or Portal should provide something for all this like we had previously in portal? Am I doing something wrong and all this should be handled in a different way that I am not aware of?

You can also find some context in this pull, https://github.com/liferay-lima/liferay-portal/pull/3116#issuecomment-1153566605 where I wanted to use the native **min** attribute for the size input but in order to avoid the native messages, I should implement this by myself. 
<img width="773" alt="Screenshot 2022-06-13 at 12 15 55" src="https://user-images.githubusercontent.com/8373764/173332532-f9286808-8408-4469-854e-93b8e2f4a492.png">

<img width="796" alt="Screenshot 2022-06-13 at 12 16 02" src="https://user-images.githubusercontent.com/8373764/173332542-dc0a3b37-9c9b-4ef4-a6d1-b9d4159012a8.png">

@lexicon

matuzalemsteles:
Hey @ambrinchaudhary that's a good question, I'm going to answer some of your questions and an insight I have on it.

[...]

PR #4966, opened 2022-07-01
fix(@clayui/css): Form Validation disables validation icons (HTML5 version)

pat270:
- Bootstrap HTML5 Form Validation adds icons inside the input. We disabled these because it's not in the Lexicon spec. Enable them again if needed with `$enable-validation-icons: true;`

issue #4926
html5-form-validation


matuzalemsteles (approved):
LGTM! Oh that's nice, I didn't know about that option.

PR #4969, opened 2022-07-05
fix(@clayui/css): Alert Indicator inside alert-autofit-row aren't vertically aligned with text

beltranrengifo:
Applies the same explanation we did for this PR
https://github.com/liferay-frontend/liferay-portal/pull/2478

=====

[Jira issue](https://issues.liferay.com/browse/LPS-157502)
[Lexicon](https://liferay.design/lexicon/core-components/alerts)

## Motivation

The icon in the alert tag component seems misaligned with the text

## Solution proposed

Use the baseline alignment for the flex children

## How to verify

In any use case of the tag, check the icon + text alignment.

#### Before

image

#### After

image

### DIFF
--- a/packages/clay-alert/src/index.tsx
+++ b/packages/clay-alert/src/index.tsx
@@ -162,11 +162,7 @@ const ClayAlert: React.FunctionComponent<IClayAlertProps> & {
 			role="alert"
 		>
 			<ConditionalContainer>
-				<ClayLayout.ContentRow
-					className={classNames('alert-autofit-row', {
-						'align-items-baseline': variant !== 'feedback',
-					})}
-				>
+				<ClayLayout.ContentRow className="alert-autofit-row">
 					{!VARIANTS.includes(variant as string) && (
 						<ClayLayout.ContentCol>
 							<ClayLayout.ContentSection>
--- a/packages/clay-alert/src/index.tsx
+++ b/packages/clay-alert/src/index.tsx
@@ -162,7 +162,11 @@ const ClayAlert: React.FunctionComponent<IClayAlertProps> & {
 			role="alert"
 		>
 			<ConditionalContainer>
-				<ClayLayout.ContentRow className="alert-autofit-row">
+				<ClayLayout.ContentRow
+					className={classNames('alert-autofit-row', {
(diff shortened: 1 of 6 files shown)


packages/clay-alert/src/index.tsx
```diff
@@ -162,7 +162,11 @@ const ClayAlert: React.FunctionComponent<IClayAlertProps> & {
 			role="alert"
 		>
 			<ConditionalContainer>
-				<ClayLayout.ContentRow className="alert-autofit-row">
+				<ClayLayout.ContentRow
+					className={classNames('alert-autofit-row', {
+						'align-items-baseline': variant !== 'feedback',
```
pat270, line 167: Instead of adding it here, I think it's better to add it in the CSS. You can add it in this Sass map https://github.com/liferay/clay/blob/a6e08b9db3828c0344aa12254931779b3fd7fa24/packages/clay-css/src/scss/variables/_alerts.scss#L397-L409.

packages/clay-css/src/scss/variables/_alerts.scss
```diff
@@ -177,11 +177,13 @@ $alert-link: map-deep-merge(
 // .alert-indicator
 
 $alert-indicator-font-size: 1.25rem !default;
+$alert-indicator-line-height: 1 !default;
```
pat270, line 180: We don't need to add this specific variable here, just add `line-height: 1,` in the `$alert-indicator` variable.

pat270 (changes requested):

beltranrengifo:
Reviewed the changes needed and pushed.

> Instead of adding it here, I think it's better to add it in the CSS. You can add it in this Sass map

I wonder the reason behind this, what I understand is to keep out of the JS everything is not needed there.

matuzalemsteles:
I just sent a commit to change this in cadmin as well.

pat270:
@beltranrengifo We put the rule in the CSS because Bootstrap utilities use the `!important` flag. It makes it less user friendly since it's more difficult to overwrite. Also markup is harder to customize atm, so we prefer to go with CSS.

PR #4971, opened 2022-07-05
chore(@clayui/css): Fix REUSE/SPDX snippet tags

silverhook:
Recently snippet tags have finally been accepted to SPDX specification, and as such part of REUSE specification too.

This commit updates the licensing info comments in affected snippets to be compliant with the finally standardised tags:
https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/file-tags.md#h3-snippet-tags-format-

Nothing in the code should be affected by this, as it is only touching tags in comments.


pat270:
LGTM! Thanks @silverhook !

PR #4972, opened 2022-07-05
feat: add new property to enable the focus ring visible only when interacting with the keyboard

### LINKED ISSUE #3342: Apply C-inner to Clay components
## What is your proposal?

My proposal is to apply [C-inner](https://clayui.com/docs/css/utilities/c-inner.html) by default.

I think we receive a lot of complaints about the focus-style: very aggressive, thick, etc. I think we took a good decision for the sake of accessibility but on the other hand, we could adjust the experience using it not overwhelming the user if he/she doesn't interact with the keyboard proactively. 

The alternative is to reduce the visual noise of the focus style to a thinner version (1px instead of 2px).

## More context

This proposal becomes from a [slack conversation](https://liferay.slack.com/archives/CNBG06JS3/p1590399390109700) about the focus for the section component in DXP Page Editor.

## Concerns

I really don't see what are the cons of applying c-inner. So I would encourage you to explore and discuss what are the consequences of enabling it by default.

matuzalemsteles:
Fixes #3342

This is a draft to enable the focus ring to be visible only when the user interacts with the keyboard and not be visible on click, there are some details that we need to analyze, for example, some components manually focus on the element and this shows the focus ring, maybe we avoid it.

Another detail is that `input` components cannot easily solve this with just the extra markup and `c-inner` so we would have to control this via JavaScript, for example when interacting via the keyboard we add the focus ring class.

My idea is that we don't allow disabling/enabling this at the component level, we do it for all components is what makes sense. So you need to use `Provider` in the application root to enable it.

```jsx
<Provider focusRing={true}>
  ...
</Provider>
```

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -5,7 +5,7 @@ import {Provider} from '@clayui/provider';
 
 export const decorators = [
 	(Story) => (
-		<Provider spritemap={spritemap}>
+		<Provider spritemap={spritemap} focusRing={true}>
 			<div>
 				<Story />
 			</div>
--- a/jest.config.js
+++ b/jest.config.js
@@ -32,7 +32,7 @@ module.exports = {
 			statements: 97,
 		},
 		'./packages/clay-button/src/': {
-			branches: 90,
+			branches: 87,
 			functions: 100,
 			lines: 100,
 			statements: 100,
@@ -44,7 +44,7 @@ module.exports = {
 			statements: 95,
 		},
 		'./packages/clay-color-picker/src/': {
-			branches: 74,
+			branches: 73,
 			functions: 87,
 			lines: 90,
 			statements: 90,
@@ -98,7 +98,7 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-label/src/': {
-			branches: 100,
+			branches: 93,
 			functions: 100,
 			lines: 100,
 			statements: 100,
@@ -116,7 +116,7 @@ module.exports = {
 			statements: 90,
 		},
 		'./packages/clay-list/src/': {
-			branches: 64,
+			branches: 61,
 			functions: 75,
 			lines: 93,
 			statements: 92,
--- a/packages/clay-alert/src/ToastContainer.tsx
+++ b/packages/clay-alert/src/ToastContainer.tsx
@@ -8,7 +8,8 @@ import React from 'react';
 
 import {IClayAlertProps} from './index';
 
-interface IToastContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+export interface IToastContainerProps
+	extends React.HTMLAttributes<HTMLDivElement> {
 	/**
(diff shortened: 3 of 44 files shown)


matuzalemsteles:
Hey @pat270 what do you think?

pat270:
@matuzalemsteles I'll take a look right now

pat270:
Working well for me, the only nitpick I have is with the `focusRing` attribute. Should the `focusRing` values be swapped? What I mean is `focusRing={false}` should trigger rendering the `c-inner` elements?

I also noticed ColorPicker swatches show focus on mouse down, not sure if it's from Clay or the plugin.

matuzalemsteles:
Hey @pat270 I made a PoC adding the possibility to control the focus ring via JS in the button component, this brings some benefits as we can support input cases, focusing on the element using JS (`element.focus()`) does not add the focus ring it will only add if the user is browsing via keyboard this is very interesting too. You can check these changes in this commit https://github.com/liferay/clay/pull/4972/commits/c9ded7b683e6fd375cfacd63c1be58973d7b7d9c

I made some changes to the Button's CSS to have the `focus-ring` class which removes the focus ring by default and when you add `.focus-ring.show` it sets the focus ring visibility. This was a strategy I thought I could still have focus ring compatibility by default and then be able to override and control it via CSS class. let me know what you think

pat270:
I don't think we need to add the `focus-ring` and `show` classes. We can leverage the browser's `:focus-visible` pseudo class, but we would need to move the focus `box-shadow` style to `.btn:focus-visible`. If someone wants the focus ring on click, maybe we could use a data attribute like `data-show-focus-ring-on-click="true"` and add the `focus` class on click?

matuzalemsteles:
> I don't think we need to add the focus-ring and show classes. We can leverage the browser's :focus-visible pseudo class, but we would need to move the focus box-shadow style to .btn:focus-visible. If someone wants the focus ring on click, maybe we could use a data attribute like data-show-focus-ring-on-click="true" and add the focus class on click?

I thought of something like this about removing by default my concern is that this can generate some problems for those who use only Clay markup because we could keep the default behavior for the components even if we removed the focus ring and left it configurable because we could set it as default.

The issue with wanting to use the class to control the focus is because I'm controlling this via JS to determine when it's focus via the user's keyboard, this also avoids adding the focus ring when we're manually focusing the element, I think using `:focus -visible` would we be going the same way?

pat270:
This seems like an improvement for most cases. For the minority that don't want it, we can provide a way to revert back with a variable like `$enable-focus-visible` or through documentation like https://clayui.com/docs/components/loading-indicator/markup.html#css-loading-animation-reverting-to-the-dotted-animation.

I tested quickly in Chrome and focus with JS doesn't trigger `:focus-visible`. I'll test more thoroughly and get back to you.

matuzalemsteles:
I'll close this PR in favor of implementation #5006

PR #4973, opened 2022-07-06
fix(@clayui/css): Utilities update inline-scroller focus ring

### LINKED ISSUE #4937: Clay Tooltip gets unwanted outline on focus
Hi team! As we discussed, the popover gets outlined if any kb key is pressed.
 1. [Open Clay's SB](https://storybook.clayui.com/?path=/story/design-system-components-popover--default) 
 2. Click on the popover
 3. Press any key in the kb

I wonder if the popover should get outlined or not.

If yes, are this the proper styles?
If not, how should the component handle the focus?

## Environment

Current version

image

pat270:
fixes #4937

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_utilities.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_utilities.scss
@@ -272,14 +272,15 @@
 // Inline Scroller
 
 .inline-scroller {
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
LGTM

PR #4990, opened 2022-07-13
fix(@clayui.css): Removes uses of unnecessary setter functions

### LINKED ISSUE #4983: @clayui/css: Remove unnecessary `setter()` functions
Most of our clay mixins check if the parameter is of type `map`. We don't need to pass an empty map to our mixins in our component files. Example:

https://github.com/liferay/clay/blob/6d7efa37fbc6a74c2a731eb8f26907c98ea9139d/packages/clay-css/src/scss/components/_alerts.scss#L20-L32

pat270:
fix(@clayui/css): Mixins check if parameter is a map to avoid Sass error

I thought it might speed up the compilation time, but I can't really tell the difference.

fixes #4983

### DIFF
--- a/packages/clay-css/src/scss/components/_alerts.scss
+++ b/packages/clay-css/src/scss/components/_alerts.scss
@@ -21,13 +21,11 @@
 	@include clay-css($alert-indicator);
 
 	.lexicon-icon {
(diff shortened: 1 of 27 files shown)


matuzalemsteles (approved):
LGTM! This is good, it might not have a big impact because the size of the variables is not that big but it might be that we are gaining a few seconds here.

Issue #5002, opened 2022-07-20
Critical Accessibility Issue: Not announcing Current Page for nav-link


fortunatomaldonado:
## What are the steps to reproduce?
1. Use a Screen Reader like JAWS
2. Start Liferay and add a Calendar to a page
3. Tab through until you reach the nav-link Calendar

This can also be seen in https://clayui.com/docs/components/navigation-bar/markup.html#clay-sidebar
Current Page is never announced by Screen Reader

## What is the expected result?
Expected:
The Screen Reader should announces Current Page for the nav-link Calendar that is currently open

Actual: 
The Screen Reader does not announce Current Page for the nav-link Calendar that is currently open

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.65.2  |
| React |         |

fortunatomaldonado:
Hello @matuzalemsteles, thank you for taking a look at this.
I believe Lianne was able to provide a fix here: https://github.com/fortunatomaldonado/clay/pull/1

I will forward the fix shortly.

matuzalemsteles:
Oh that's great. No problem, this can be forwarded. After we fix it on our side I'll take care of removing it there so we can leave it on Clay's side, thanks!

matuzalemsteles:
Oh I figured that out now 🤦‍♂️, I thought this was going to be for DXP! Yes, you can send a fix to Clay, thanks!

wanderlast:
Sent a fix here: https://github.com/liferay/clay/pull/5003

Let me know if anything needs adjusting!

PR #5003, opened 2022-07-20
Navigation bar does not inform screen readers of "current page".

wanderlast:
Hi!

This is a fix for the accessibility issue mentioned here: https://github.com/liferay/clay/issues/5002

It's meant to address the client reported issue seen on [LPP-45909](https://issues.liferay.com/browse/LPP-45909) as HRG-06 where screen readers are not informed of what the 'current page' on a navigation bar, even when highlighting it.

I've attempted to format the commits correctly, but please let me know if anything needs adjusting. Thanks!

### DIFF
--- a/packages/clay-navigation-bar/src/Item.tsx
+++ b/packages/clay-navigation-bar/src/Item.tsx
@@ -24,6 +24,8 @@ const ClayNavigationBarIcon = ({
 	className,
 	...otherProps
 }: IItemProps) => {
+	const ariaCurrent = active ? 'page' : false;
+
 	return (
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
Hey @wanderlast, thanks for sending the fix. We don't have the `bug` scope in the commit message, in this case, we use `fix`. You can see the list with full descriptions here https://github.com/liferay/clay/blob/master/CONTRIBUTING.md#type.

I'll take care of changing it for you and I'll make a small change, I'll add the inline conditional instead of creating a variable just for that, in that case, it's good to avoid because we only use it in one place.

PR #5005, opened 2022-07-20
feat(@clayui/drop-down): add new `roleItem` property to set item role

### LINKED ISSUE #5004: Provide a way to change the item's role property in `DropDown.Item`
In issue #4979, we changed the ownership of `role` to be added in the container and #4998 added the default `role` to the item as `menuitem` to improve the accessibility of DropDown OOTB without the DXP developers having to configure it.

There are some scenarios where the Menu is used as `listbox` and the items should be set to `option`, this conflicts with the item being `menuitem` and the item container being `option`, to improve accessibility here we need leave the container with `role="none"` and the item with `role="option"`.

matuzalemsteles:
Fixes #5004

### DIFF
--- a/packages/clay-drop-down/src/Item.tsx
+++ b/packages/clay-drop-down/src/Item.tsx
@@ -30,6 +30,12 @@ export interface IProps
 	innerRef?: React.Ref<any>;
 
 	/**
+	 * Sets the role accessibility property of the item. Set the item's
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
As soon as the tests pass I'll go ahead and merge to make a release of this package, I found this case after updating the Clay packages in DXP.

PR #5007, opened 2022-07-21
feat(@clayui/tooltip): adds new properties to configure the tooltip to be floating

### LINKED ISSUE #4991: New tooltip behavior
## Context
We are currently improving the interaction to live-edit a text fragment in Page Editor. This improvement includes the use of a tooltip that will inform users that a fragment is editable.  

## Our need
What we would need is that the tooltip is shown close to the cursor (Image attached). Something similar to what Jira does in its ticket descriptions (Video attached). If we use the tooltip as it is, the interaction and the tooltip get out of context. Could you study this to see if we can cover this new behavior?   

Image

https://user-images.githubusercontent.com/89843928/179214072-ebb7ac2a-3e31-40a9-8dca-00b9a67ed2f8.mov

## Previous discussion with Lexicon
We opened a thread discussing about this for you to have a look:
https://liferay.slack.com/archives/CKT3L1G58/p1657097834908479

cc\ @mateo-hermosin @drakonux

matuzalemsteles:
Closes #4991

This feature is only available with the `<TooltipProvider />`, as we are already using it in DXP the developers just need to add the `data-tooltip-floating="true"` property to the element with the `title` for the tooltip to be floating according to the mouse position on the element. See the [example in the storybook](https://deploy-preview-5007--next-storybook-clayui.netlify.app/?path=/story/design-system-components-tooltip--floating).

### DIFF
--- a/custom-types/dom-align/index.d.ts
+++ b/custom-types/dom-align/index.d.ts
@@ -20,6 +20,17 @@ declare module 'dom-align' {
 		useCssRight?: boolean
 	}
 
+	type MousePosition = {
+		clientX: number;
+		clientY: number;
+	};
+
+	export function alignPoint(
+		sourceNode: HTMLElement,
+		target: MousePosition,
+		config?: IConfigOptional
+	): void;
+
 	export default function doAlign(
 		sourceNode: HTMLElement,
 		targetNode: HTMLElement,
--- a/packages/clay-shared/src/useMousePosition.ts
+++ b/packages/clay-shared/src/useMousePosition.ts
@@ -25,21 +25,21 @@ function throttle<T>(callback: Function, limit: number) {
 /**
  * Hook to get the current mouse position
  */
-export const useMousePosition = () => {
+export const useMousePosition = (delay: number = 200) => {
 	const [mousePosition, setMousePosition] = useState<MousePosition>([0, 0]);
 
 	useEffect(() => {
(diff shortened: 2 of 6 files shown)


matuzalemsteles:
Hey @carolrodmar could you test this and verify that this is the desired behavior? In this implementation, I made the floating tooltip very fluid that follows the mouse position while the user interacts with the element.

- [Demo](https://deploy-preview-5007--next-storybook-clayui.netlify.app/?path=/story/design-system-components-tooltip--floating)

matuzalemsteles:
I'm going ahead and merging this PR to add this new behavior to the component and we can include it in today's release, if we need to readjust we can create a new follow-up PR.

carolrodmar:
Thanks! @matuzalemsteles 

This solution is a good approach but for me it is a bit weird that the tooltip keeps following the mouse. The idea was to do something similar to this video:

https://user-images.githubusercontent.com/89843928/182169433-bd601764-042c-41e9-a3ac-708b35f4f935.mov

Tooltip appears on mouse hover and if you move it, the tooltip disappears and shows again close to the mouse after few seconds. Do you think we could update this behavior?

matuzalemsteles:
Hey @carolrodmar I see, I think this is possible i will do some more tests and send a new follow up PR. Thanks for the feedback!

PR #5015, opened 2022-07-27
chore(@clayui/css): Fix Apache-2.0 SPDX ID

silverhook:
Noticed a wrong use of Apache-2.0 SPDX License ID, so I fixed it and included the appropriate license text in the right folder too.

Nothing in the code should be affected by this, as it is only touching tags in comments.


pat270:
Thanks again for the update @silverhook !

PR #5023, opened 2022-08-01
fix(@clayui/tooltip): remove the behavior of moving the tooltip when the mouse moves

matuzalemsteles:
https://github.com/liferay/clay/pull/5007#issuecomment-1201277452

Hey @carolrodmar I believe this should be the closest behavior, let me know if something more subtle would be needed. You can test this change in the storybook as soon as the [preview deploy generated by Netlify](https://github.com/liferay/clay/blob/master/docs/netlify_deploy.md#deploy-previews) once the build finishes.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -231,7 +231,7 @@ module.exports = {
 		},
 		'./packages/clay-tooltip/src/': {
 			branches: 52,
-			functions: 61,
+			functions: 60,
 			lines: 70,
 			statements: 69,
 		},
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -403,7 +403,7 @@ const TooltipProvider = ({
 				}
 			);
 		}
-	}, [mousePosition, show, floating]);
+	}, [show, floating]);
 
 	useEffect(() => {
 		if (
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -218,6 +218,7 @@ const TooltipProvider = ({
 
 	// Using `any` type since TS incorrectly infers setTimeout to be from NodeJS
 	const timeoutIdRef = useRef<any>();
+	const floatingTimeoutIdRef = useRef<any>();
 	const targetRef = useRef<HTMLElement | null>(null);
 	const titleNodeRef = useRef<HTMLElement | null>(null);
 	const tooltipRef = useRef<HTMLElement | null>(null);
@@ -265,7 +266,7 @@ const TooltipProvider = ({
 		}
 	}, []);
 
-	const handleHide = useCallback((event?: any) => {
+	const handleHide = useCallback((event?: any, restore: boolean = true) => {
 		if (
 			event &&
 			(tooltipRef.current?.contains(event.relatedTarget) ||
@@ -277,8 +278,11 @@ const TooltipProvider = ({
 		dispatch({type: 'hide'});
 
 		clearTimeout(timeoutIdRef.current);
+		clearTimeout(floatingTimeoutIdRef.current);
 
-		restoreTitle();
+		if (restore) {
+			restoreTitle();
+		}
 
 		if (targetRef.current) {
 			targetRef.current.removeEventListener('click', handleHide);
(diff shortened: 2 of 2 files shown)


matuzalemsteles:
Hey @carolrodmar I made the changes we talked about in Slack, I'll go ahead and merge but feel free to test the behavior via the link https://deploy-preview-5023--next-storybook-clayui.netlify.app/?path=/story/design-system-components-tooltip--floating

carolrodmar:
Hi @matuzalemsteles , I think there was a misunderstanding. What I can see now is that the tooltip disappears even when the mouse is over the editable text. This should not behave like that. I will try to explain the cases here:

-The tooltip shows when the mouse enters in an editable text **and stays visible until he mouse starts moving**.
-If the mouse starts moving, the tooltip disappears.
-The tooltip shows once the mouse stops in the editable text again.

matuzalemsteles:
Hey @carolrodmar I'll try to explain how this works too. The floating tooltip is activated when the mouse is over the entire box and not the text, this is defined by the developer. In this demo I added it to be activated as soon as it enters the whole aria so it doesn't matter if it's over text, this can also be done to have this behavior of appearing when just standing over text, so I believe we supporting this, too if it exits outside the aria that was set will disappear.

I think the only case here is the tooltip disappears when the mouse starts to move, I think this conflicts with @marcoscv-work's comment about the accessibility issue https://github.com/liferay/clay/issues/4991#issuecomment-1201510701 for this behavior, I think we have to reach consensus here before submitting something.

carolrodmar:
Hi @matuzalemsteles I understand but the tooltip now is doing this. It disappears after 1/2 seconds and thats weird. It should stay visible.

https://user-images.githubusercontent.com/89843928/183585584-09c9f799-62fc-415d-90b1-28ca75211bd6.mov

carolrodmar:
Hey @matuzalemsteles I want to clarify a few things after discussing this with @marcoscv-work  

This previous requirement --> "If the mouse starts moving, the tooltip disappears" produced misunderstanding. We are rejecting it as it does not follow WCAG 2.1 AA.

So, the correct behavior should be:
- Tooltip shows when pointer enters an editable area **and it stays visible**.
- The pointer can be moved over the additional content without the additional content disappearing. 
- If pointer leaves the editable area, tooltip should disappear.

matuzalemsteles:
@carolrodmar there are still some things that are a little confusing for me, I think there may also be some mismatch of how this works.

As I said before, the Tooltip floating appears as soon as the mouse enters a designated area, by default for the tooltip to appear it follows the Lexicon recommendation of 500ms if the user remains over the area, for example:

```jsx
<div
  data-tooltip-floating="true"
  style={{
    backgroundColor: '#e6e6e6',
    borderRadius: '10px',
    display: 'inline-block',
    fontWeight: 500,
    padding: '60px 100px',
  }}
  title="Edit Text"
>
  <span>Placeholder</span>
</div>
```

In this scenario to show the tooltip, it doesn't need to be over the text because we can only identify when the mouse is over the box around the text, so there is no differentiation between the space around and the text, to change this we would need to move the attribute `data-tooltip-floating="true"` for the text, for example:

```jsx
<div
  style={{
    backgroundColor: '#e6e6e6',
    borderRadius: '10px',
    display: 'inline-block',
    fontWeight: 500,
    padding: '60px 100px',
  }}
>
  <span data-tooltip-floating="true" title="Edit Text">Placeholder</span>
</div>
```

This way we get the case "Tooltip shows when pointer enters an editable area and it stays visible", this should also cover the case "The pointer can be moved over the additional content without the additional content disappearing.", I imagine you are meaning that when the content here would be a text with many paragraphs I imagine and the tooltip shouldn't disappear, am I right?

To "If pointer leaves the editable area, tooltip should disappear." this would also be covered if it was configured as I showed above, the only thing here is that for the tooltip not to disappear I need to revert this change but I think the tooltip will be stuck in a position, I think you came to comment on this that it looked strange. See these two demos without this change:

https://user-images.githubusercontent.com/13750819/183734267-d6c421f1-f6d1-4dc4-8a6c-f331435608a4.mov

https://user-images.githubusercontent.com/13750819/183735058-3f5f9394-34a1-420e-9810-06e21eec04b8.mov

The case of paragraphs is more complicated to always keep the tooltip visible because the lines have a space between each one and we receive an event that the pointer has left the text. Let me know if you think this is enough.

carolrodmar:
Hi @matuzalemsteles for me these demos work perfect! Thanks a lot!

PR #5037, opened 2022-08-12
fix(@clayui/css): Buttons `$btn-palette` and `$btn-outline-palette` s…

### LINKED ISSUE #5028: @clayui/css: Buttons $btn-palette should support key names prefixed with `btn-`
Currently, the `$btn-palette` map prepends `.btn-` if a key name doesn't start with '.' or '#'. We assume the key names will be `primary`, `secondary`, etc. This was a pattern we inherited from Bootstrap. We are beginning to name our keys closer to the css selector. We should be able to use keys `btn-primary`, `btn-secondary`, etc. Clay will output `.btn-btn-primary` and `.btn-btn-secondary`, respectively. It should output `.btn-primary` and `.btn-secondary`.

https://github.com/liferay/clay/blob/79b6fa6b0661b0e1b2d02b27a0de2f22118348f9/packages/clay-css/src/scss/components/_buttons.scss#L116

https://github.com/liferay/clay/blob/79b6fa6b0661b0e1b2d02b27a0de2f22118348f9/packages/clay-css/src/scss/components/_buttons.scss#L158

pat270:
…hould support keys prefixed with `btn-`

fixes #5028

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_buttons.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_buttons.scss
@@ -119,7 +119,11 @@ input[type='button'] {
 			starts-with($color, '#') or
 			starts-with($color, '%'),
 		$color,
-		str-insert($color, '.btn-', 1)
+		if(
+			starts-with($color, 'btn-'),
+			str-insert($color, '.', 1),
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM

PR #5038, opened 2022-08-15
fix(@clayui/drop-down): Improves group semantics for DropDown and focus with virtual cursor

### LINKED ISSUE #5026: Dropdown menu accessibility issue from the beginning of time
We really need to move the dropdown options to the button of the page and make it visually positioned on the top of any other element, thats a Liferay need for some _corner cases_ (dropdown inside overflow hidden divs, sidebar menus, etc).

It caused issues related to accessibility and we have been fixing the focus during years, currently it is working fine on screen readers like VoiceOver but it still broken for any assistive technology based on the accessibility tree witch still broken because we don't show the dropdown menu options next to the trigger, a good example for this is JAWS, the most popular screen reader.

@victorg1991 has been fighting with it on this PTR: https://issues.liferay.com/browse/PTR-3245
We suggested to use virtual cursor but it could help for JAWS but not for other assistive technology based on the accessibility tree.

Here the accessibility tree showing the dropdown opened:
image

**Proposal**:

Make the positioning on the page button configurable by API and true by default.
The "positioning on the page button"=false should position the dropdown next to the dropdown trigger, then we could change to false the cases one by one and fix this issue for most of the dropdowns

matuzalemsteles:
Closes #5026

Well, this is an accessibility improvement for DropDown that resolves two things, The group semantics mainly for implementation using low-level components and focus control when using the virtual cursor in the screen reader. To better understand this I recommend reading the issue comments https://github.com/liferay/clay/issues/5026#issuecomment-1209895810

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -197,7 +197,7 @@ function ClayDropDown({
 	}, []);
 
 	return (
-		<FocusScope>
+		<FocusScope arrowKeysLeftRight>
 			<ContainerElement
 				{...otherProps}
 				className={classNames('dropdown', className)}
--- a/packages/clay-drop-down/src/DropDownWithItems.tsx
+++ b/packages/clay-drop-down/src/DropDownWithItems.tsx
@@ -249,19 +249,9 @@ const Item = ({
 };
 
 const Group = ({items, label, spritemap}: IItem & IInternalItem) => (
-	<>
-		<ClayDropDownGroup header={label} />
-		{items && (
-			<li role="none">
-				<DropDownContent
-					aria-label={label}
-					items={items}
-					role="group"
-					spritemap={spritemap}
-				/>
-			</li>
-		)}
-	</>
+	<ClayDropDownGroup header={label}>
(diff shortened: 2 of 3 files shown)


matuzalemsteles:
Well, I sent the final commits that make the DropDown menu accessible to the screen reader's virtual cursor, I think we can use this same strategy for Modal for example because it has the same problem. Using `arrowKeysLeftRight` had no effect so I had to go back and go in another direction.

I made some minor role changes since I was reviewing accessibility recommendations for menu https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-links, it seems that we have very different behaviors in some situations, for example navigating the menu using Tab and we need to organize the contents that are part of the menu to be accessible, for now caption, search and others that are outside the list or `role="menu"` but is still accessible if the virtual cursor is used.

I added some more changes like opening the menu using ArrowDown and Spacebar and moving the focus to the menu when opening, there are other shortcuts that were not implemented like typeahead, navigation with the arrow keys being in loop, Home and End shortcuts the other semantics content we are much better now after the recent changes I added to the menu items and now improving the semantics for the group. I'm going to open an issue so we can take care of this and mainly review the tabulation in the menu.

PR #5058, opened 2022-08-30
feat(@clayui/date-picker): add ability to render calendar outside of a toggled dialog

bryceosterhaus:
@matuzalemsteles here is a rough idea for the mini calendar, let me know if you think we should provide it differently, but I thought this was a rough first idea of how to add support for it.

I am going to be helping add this for our calendar portlet, so I'll first test with this implementation and let you know if we need any other APIs.

### DIFF
--- a/packages/clay-date-picker/src/__tests__/BasicRendering.tsx
+++ b/packages/clay-date-picker/src/__tests__/BasicRendering.tsx
@@ -25,6 +25,20 @@ describe('BasicRendering', () => {
 		expect(document.body).toMatchSnapshot();
 	});
 
+	it('renders not in a dialog', () => {
+		render(
+			<ClayDatePicker
+				defaultMonth={new Date(2019, 3, 18)}
+				dialog={false}
+				placeholder="YYYY-MM-DD"
+				spritemap={spritemap}
+				years={{end: 2019, start: 2019}}
+			/>
+		);
+
+		expect(document.body).toMatchSnapshot();
+	});
+
 	it('renders the date picker with the selected day using the defaultMonth', () => {
 		const {getByLabelText} = render(
 			<ClayDatePicker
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
Thanks @bryceosterhaus for working on this! I'm going to do a deeper review today!

matuzalemsteles:
Well, I think it's the beginning. I imagined a little differently how we could offer this but it would also be good to test this in a real case to understand the needs as you are going to do it so it will be great.

I had my mind on being able to offer a component to the Menu part of the Date Picker that was controlled by the developer, something like:

```jsx
// DatePicker
<DatePicker>
  <DatePicker.Input />
  <DatePicker.Menu />
</DatePicker>

// Uncontrolled isolated menu
<DatePicker>
  <DatePicker.Menu />
</DatePicker>

// Controlled isolated menu
<DatePicker.Menu />
```

This is a very premature sketch but something along these lines where `<DatePicker />` has more of a role to control the states and allow the composition of the menu and the trigger which in this case is the input, if the developer wants only the Menu but that the DatePicker still controls the state of all interactions it only remove the input but want to control the states by itself and customize according to the application structure it can just go to directly listen to the Menu interaction events as the last example .

This would still have the usual APIs that exist today in DatePicker but just a different composition that would allow for more flexibility and moving the pieces around. But this would be good to understand if the need for this but I like the idea because it is close to the other components we have, we would just have to understand how this would work being called loosely inside the Calendar for example, but it would be nice to hear from you what this would be like integration into the calendar first before moving forward with any other ideas that require more work and rewriting of some things.

bryceosterhaus:
@matuzalemsteles actually this isn't as urgent any more as we will add the fix via YUI for the calendar portlet. I still think this is a good component to have though, but I'll let you prioritize and design the API for it. You can check the calendar portlet if you ever need a real world use case for it.

bryceosterhaus:
And I liked your API example, providing some low-level components would be helpful where the menu is agnostic of the input

matuzalemsteles:
@bryceosterhaus oh okay, no problems! I'll create an issue to handle with this, thanks Bryce

Issue #5044, opened 2022-08-18
Provide documentation for the clay taglib


kastork:
When you click through on one of the "available tag" links, it takes you to the documentation for the react component.

While this has some value, it does not:

- document the tag attributes
- explain how to write javascript in your JSP that interacts with the component, or that handles component event.

In the "Composing" section of a component's documentaion, there's a JSP tab, but it generally does not provide documentation or examples for using the tag. And most of the content/examples there is still applicable to the javascript component itself.

matuzalemsteles:
Hi @kastork, Thanks for reporting this, we have in mind to work on this and organize a dedicated documentation for the Clay taglibs so that we can provide more detailed information specific to taglib and patterns, you can follow these changes we will implement using the issue #4128. I've linked other related issues to this issue to make it easier to track these changes, feel free to add more suggestions or ideas.

cc @dsanz

matuzalemsteles:
LPS Ticket https://liferay.atlassian.net/browse/LPS-190530

PR #5063, opened 2022-08-30
docs(clayui.com): Panel add warning and work around for collapse plug…

pat270:
…in for 7.4

    - Update examples with panel-secondary and panel-unstyled

Kind of related to #5044. This isn't a Clay tag, but is relevant to using the collapse plugin in JSP's and Fragments.


matuzalemsteles (approved):
LGTM! Makes sense, thanks

PR #5071, opened 2022-09-01
docs(clayui.com): Modal adds documentation about `Liferay.Util.openMo…

pat270:
…dal`

#5044

### DIFF
--- a/packages/clay-modal/docs/markup-modal.md
+++ b/packages/clay-modal/docs/markup-modal.md
@@ -42,6 +42,114 @@ mainTabURL: 'docs/components/modal.html'
 </div>
 </div>
 
+<div class="clay-site-alert alert alert-warning">
+	This page uses Bootstrap's modal plugin which requires JQuery. Liferay 7.4 no longer includes JQuery by default. We have included a standalone modal plugin in 7.4.
+</div>
+
+```html{expanded}
+<button class="btn btn-primary" id="myModalToggle" type="button">Small Modal</button>
+<script>
+	document.getElementById('myModalToggle').addEventListener('click', function(event) {
+		Liferay.Util.openModal({
+			title: 'Modal Title',
+			bodyHTML: '<h4>Modal Body</h4>',
+			buttons: [
+				{
+					displayType: 'secondary',
+					label: 'Cancel',
+					type: 'cancel',
+				},
+				{
+					displayType: 'primary',
+					label: 'Primary',
+					type: 'submit',
+				},
+			],
+			size: 'sm',
+		});
+	});
+</script>
+```
+
+<div class="table-responsive">
+	<table class="table table-autofit table-hover table-list">
+		<thead>
+			<tr>
+				<th class="table-cell-expand-smallest">Property</th>
+				<th class="table-cell-expand-smallest">Type</th>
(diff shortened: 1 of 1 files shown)


packages/clay-modal/docs/markup-modal.md
```diff
@@ -42,6 +42,114 @@ mainTabURL: 'docs/components/modal.html'
 </div>
 </div>
 
+<div class="clay-site-alert alert alert-warning">
```
matuzalemsteles, line 45: I think it would be nice to create a specific session for this to be available in the table of contents and also I think this is more specific to DXP.

I think ideally we should put these types of APIs in specific sessions for DXP and I'm also not sure if we want `openModal` to be public for external use, in addition here we are also saying it is for external use I think so so we need to check with the rest of the team mostly with the DXP squad.

pat270: You're right we probably should consult with the team and it does make sense to have it in the toc. My only concern is that we have no replacement for Bootstrap's modal js, which is why I started documenting this. I wonder if we could add a disclaimer, not officially supported?

matuzalemsteles: > I wonder if we could add a disclaimer, not officially supported?

Yes, this might be left as a suggestion for people who aren't using the React implementation!

> My only concern is that we have no replacement for Bootstrap's modal js

Yeah, just tagging @markocikos and @diegonvs to see if this is something official that we could add here as an alternative.

diegonvs: Maybe we should create a document on liferay docs(is there a SSOT for DXP things?)? where Clay fits on Liferay DXP. 

Like, such as explaining about cadmin, what we can offer for customers to customize using Clay, how ClayCSS fits with DXP CSS like classic-theme, etc. I think these explanations can be diluted around many places like clayui.com, maybe on Liferay docs, but the idea would be to centralize everything and make Clay docs point to this place.

pat270: @diegonvs I always thought https://learn.liferay.com/index.html was the SSOT. I don't see much there for FE devs. I agree this should be where the DXP things live.

diegonvs: Hey @dsanz, what do you think?

markocikos: This does not seem right, `openModal` is specific to DXP. Clay is still platform agnostic, right? docs should reflect that. It may not be a bad idea to add an example of DXP usage, but that should probably be highlighted, either next to the example or even in a new tab.

matuzalemsteles: Yeah, my concern with this is whether this should be used by customers if we leave it here, we're saying customers can use it. I think the best way is things like this should be in a DXP documentation and here we just point to DXP if necessary.

We have to think that Clay is a component base and will not always be the "final component", so the design system so things implemented within DXP will work differently so that it can be reusable within the platform, for example: Modal and Alert.

matuzalemsteles:
I'm going to close this PR since we don't have a final conclusion for this, also I think we agreed that this would be best documented in the DXP docs.

PR #5075, opened 2022-09-02
chore(clayui.com): Only build the files needed when modifying Clay CSS or icons

pat270:
…S or icons

This compiles only the files needed for clayui.com (in development mode) when modifying parts of Clay CSS. It speeds up build time in development depending on what you modify.


marcoscv-work:
Works fine!

PR #5077, opened 2022-09-06
feat(clayui.com): Add key shortcut keys by platform type and hiding it in mobile

marcoscv-work:
Native combination keys should be known by user but a good practice is print the all combination so:
- In mac is "control + option + K"
- In linux is "alt + K"
- In windows is "control + K"

So I've added the combinations:

Linux:
image

Mac:
image

Windows:
image

It is a good exercise if we consider to add native shortcuts without conflicting to screen readers and exposing the key by "aria-keyshortcuts".
If we would use this example in general, probably we could move the `@at-root` clases to a general new `.c-kbd-platform-frefix` class or something similar so then you add the class and the key is preceded by the platform native combination.

Of course, key combinations can be changed in SO configuration but those cases cannot be covered even by native aria, so I think is not a stopper.

In small devices we hide the shortcut since most probably they have not physical keyboard.

Feel free to modify the react part.

### DIFF
--- a/clayui.com/src/components/LayoutNav/Search.js
+++ b/clayui.com/src/components/LayoutNav/Search.js
@@ -3,9 +3,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import React, {useEffect} from 'react';
+import React, {useEffect, useRef} from 'react';
 
 export default (props) => {
+	const autocompleteRef = useRef();
 	useEffect(() => {
 		if (window.docsearch) {
 			window.docsearch({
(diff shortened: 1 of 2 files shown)


marcoscv-work:
it comes from https://github.com/liferay/clay/pull/5074

matuzalemsteles (approved):
LGTM! Thanks for that @marcoscv-work!

pat270 (approved):
LGTM! Thanks @marcoscv-work! If we end up implementing something like this directly in Clay, I wonder if there is a way to figure out the keyboard shortcuts using feature detection instead of User Agent sniffing.

PR #5080, opened 2022-09-07
fix(@clayui/css): Functions `clay-get-fallback()` should return custo…

### LINKED ISSUE #4853: @clayui/css: Function `clay-get-fallback($var)` errors on a custom property without fallback value
`clay-get-fallback(var(--gray-600)` should return `var(--gray-600)` instead of this:

```
Build error:
 Error: $start-at: 2 is not a number.
    ╷
307 │           $fallback-value: str-slice(
    │ ┌──────────────────────────^
308 │ │             $var,
309 │ │             str-index($var, ',') + 2,
310 │ │             str-length($var) - 1
311 │ │         );
    │ └─────────^
    ╵

```

https://github.com/liferay/clay/blob/e1e7ba120b16b5e71cf5a876c775de1e8f83ee9f/packages/clay-css/src/scss/functions/_global-functions.scss#L284-L318

pat270:
…m property if no fallback value

fixes #4853

### DIFF
--- a/packages/clay-css/src/scss/functions/_global-functions.scss
+++ b/packages/clay-css/src/scss/functions/_global-functions.scss
@@ -300,7 +300,7 @@
 /// @param {String} $var - The CSS Custom Property to evaluate.
 
 @function clay-get-fallback($var) {
-	@if (is-css-var($var)) {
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM

PR #5092, opened 2022-09-14
feat(@clayui/css): Custom Checkbox adds readonly styles

### LINKED ISSUE #5085: New read-only style for checkbox
## What is your proposal?
This proposal cames from [DESIGN-729](https://issues.liferay.com/browse/DESIGN-729), I reviewed the current read-only style that is present on the Clay site and was missing from Lexicon. I've created this [Figma file
](https://www.figma.com/file/2GCM8tEbA4xg0NqiN3HkSi/design-729-checkbox-read-only?node-id=1%3A2) with a proposal where I have some examples. 
The idea behind this proposal is to distinguish the read-only style from the default ON/OFF style, and possibly also the disabled style.

The current style of the read-only is the same as the default, and it can confuse users making them feel it's interactive (when it's not). 

none

pat270:
fixes #5085

Nothing needs to be done for the React component, it uses the same markup.

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_custom-forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_custom-forms.scss
@@ -149,6 +149,15 @@ $custom-control-input: map-deep-merge(
 		disabled: (
 			custom-control-label: $custom-control-label-disabled,
 		),
+		readonly: (
+			custom-control-label: (
+				color: $gray-900,
+				before: (
+					background-color: $gray-200,
+					border-color: $secondary-l3,
+				),
+			),
+		),
 		checked: (
 			custom-control-label: (
 				before: (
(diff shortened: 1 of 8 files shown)


matuzalemsteles:
Hey @pat270 I did a quick test with the screen reader with the Checkbox in readOnly mode, I think we need to add `aria-disabled` for the screen reader to inform that the item is dimmed, if we only have `readonly` it announces but informs that it can change the state.

We don't need to do anything for `disabled` because by default the screen reader doesn't announce it, so we don't need `aria-hidden`.

pat270:
@matuzalemsteles thanks for pointing that out. I updated the docs. I also added something I missed, readonly disabled.

matuzalemsteles:
@pat270 I added a few more commits to add test for readOnly case and add `aria-disabled` to `readOnly`.

PR #5094, opened 2022-09-15
fix(@clayui/css): Tables table row variants should style the background

### LINKED ISSUE #4876: @clayui/css: Table Row Variant styles don't work
There are table row variants that don't work. I'd like to get rid of them since they never worked, but they came with Bootstrap 🤔. 
 
```
table-primary
table-secondary
table-success
table-info
table-warning
table-danger
table-light
table-dark
```
table-row-variants

pat270:
fixes #4876

This fixes table row modifiers that don't work. I removed them in Cadmin since we don't use them in our admin interface.

table-row-variants

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_tables.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_tables.scss
@@ -77,14 +77,6 @@ caption {
 
 // Table Row Backgrounds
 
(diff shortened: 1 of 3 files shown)


matuzalemsteles (approved):
@pat270 it makes sense to remove things that we don't use in cadmin, especially things that were inherited from bootstrap, maybe we still keep it in the atlas and we can decide in the next major version if we want to keep that.

PR #5103, opened 2022-09-21
feat(@clayui/core): add paginated data support to TreeView

### LINKED ISSUE #5096: Improve ClayTreeView "load more" functionality
As we are migrating our trees in portal (in this case, specificly the Pages Tree in the product menu), we have found that the current "load more" functionality doesn't exactly fit our requirements. As far as I could see, it allows loading some item's children when expanding it. I think it's very useful, but it does not solve the case of having many items at the same level.

## What is your proposal?
In my opinion, the current "load more" works more like a "load children on expand". It would be great if we could display a certain number of items (for example 20), and then, with a "load more" button, load 20 more, and so on.

It would improve performance in cases in which you have many items at the same level, as you could load them little by little.

matuzalemsteles:
Closes #5096

To support paginated data in TreeView, I had to expose a new API to render props to allow it to be called in composition, this is a bit similar to how it works in `useResource`, the return of `loadMore` tells how it will behave, this works like a contract to allow having the behavior of loading the item's children when expanding a node or loading a node's paginated data. It is also possible to say that you do not want to load the data of a specific item, for example in the stories that are in the Storybook.

```jsx
<TreeView
  onLoadMore={async (item) => {
    // Fetch data

    return {
      cursor: result.next,
      items: result.data.list,
    };
  }}
>
  {(item, selection, expand, loadMore) => (
    <TreeView.Item>
      ...

      {expand.has(item.id) && (
        <Button
          borderless
          displayType="secondary"
          onClick={() => loadMore(item.id, item)}
        >
          Load more results
        </Button>
      )}
    </TreeView.Item>
  )}
</TreeView>
```

I tried to make the `loadMore` API public not need to receive an id and the item but unfortunately I couldn't avoid this due to the nesting system, maybe in the future we can avoid it but for now it had to be passed because internally it doesn't we have visibility of this information when the APIs are passed.

## ToDo

- [x] Add documentation for paginated data with `loadMore`
- [x] Add test for use case

### DIFF
--- a/packages/clay-core/docs/treeview.mdx
+++ b/packages/clay-core/docs/treeview.mdx
@@ -160,7 +160,12 @@ function Example() {
 
 #### Asynchronous Item
 
-When a tree is very large, loading items (nodes) asynchronously is preferred to decrease the initial payload and memory space.
+When a tree is very large, loading items (nodes) asynchronously is preferred to decrease the initial payload and memory space. TreeView provides two ways to load asynchronous data:
+
+-   Load the children of an item when clicking on the item
+-   Load paginated data from an item
+
+**Load the children of an item**
 
 The TreeView doesn't know when an item is asynchronous, so the developer must specify whether the item is asynchronous or not. The `onLoadMore` property is called every time the item is a leaf node of the tree and depending on the method's return value it will behave differently.
 
@@ -183,6 +188,52 @@ When adding a new asynchronous item to the tree, the `onItemsChange` method is r
 	}}
 >
 	{...}
+</TreeView>
+```
+
+**Load paginated data from an item**
+
+The `onLoadMore` API can also be used to load paginated data for a specific item. The signature just needs to be followed so the TreeView can know what to do.
+
+```jsx
+<TreeView
+	onLoadMore={async (item, cursor) => {
+		const result = await fetch(cursor ?? `example.com/tree/item?parent_id=${item.id}`);
+
+		return {
+			cursor: result.next,
+			items: result.data,
+		};
+	}}
+>
+	{...}
+</TreeView>
+```
+
+The `cursor` helps to store the data that will be used to identify which will be the next request, this can be an `offset`, `cursor`, `id` the URL or any other data that represents the item cursor, when there is no more data just sets the cursor to `null`. The TreeView also exposes a public API that can invoke `loadMore` and find out if there is a cursor for a specific item.
+
+```jsx
+<TreeView {...}>
+	{(item, selection, expand, load) => (
+		<TreeView.Item>
+			<TreeView.ItemStack>{item.name}</TreeView.ItemStack>
+			<TreeView.Group items={item.children}>
+				{(item) => (
+					<TreeView.Item>{item.name}</TreeView.Item>
+				)}
+			</TreeView.Group>
+
+			{expand.has(item.id) && load.has(item.id) !== null && (
+				<Button
+					borderless
+					displayType="secondary"
+					onClick={() => load.loadMore(item.id, item)}
+				>
+					Load more results
+				</Button>
+			)}
+		</TreeView.Item>
+	)}
 </TreeView>
 ```
 
--- a/packages/clay-core/src/tree-view/Collection.tsx
+++ b/packages/clay-core/src/tree-view/Collection.tsx
@@ -10,11 +10,11 @@ import {
 	Collection as CollectionBase,
 	excludeProps,
 } from '../collection';
-import {Expand, Selection, useAPI} from './context';
+import {Expand, LoadMore, Selection, useAPI} from './context';
 import {ItemContextProvider, useItem} from './useItem';
 
 export type ChildrenFunction<T extends Record<string, any>> =
-	ChildrenFunctionBase<T, [Selection, Expand]>;
+	ChildrenFunctionBase<T, [Selection, Expand, LoadMore]>;
 
 export interface ICollectionProps<T> {
 	children: React.ReactNode | ChildrenFunction<T>;
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -13,7 +13,7 @@ import {ChildrenFunction, Collection, ICollectionProps} from './Collection';
 import DragLayer from './DragLayer';
 import {TreeViewGroup} from './TreeViewGroup';
 import {TreeViewItem, TreeViewItemStack} from './TreeViewItem';
-import {Icons, TreeViewContext} from './context';
+import {Icons, OnLoadMore, TreeViewContext} from './context';
 import {ITreeProps, useTree} from './useTree';
 
 interface ITreeViewProps<T>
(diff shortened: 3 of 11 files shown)


matuzalemsteles:
Hey @sandrodw3 I would love to get your feedback as to whether this would help you implement your use case. See the example in the storybook https://github.com/liferay/clay/pull/5103/commits/a48f0e672162629af590bddd103df972842b2b66

sandrodw3:
Hey @matuzalemsteles! Sorry for the delay. I've deployed your solution and it looks super cool, thanks! 😄. I've just found two things that I think we should think about:

- As it's implemented right now, we can't have a "Load more results" button in the first level. In your example, it allows loading more children for an item. That's great (and we also need that) 😛 but in our case, we may have a lot of items in the first level, so it would be great if we find a solution for that as well.
- I've seen you are storing the current cursor in a ref, but I found it's common for all items, so if you load more results for item "A", and then you load more results for item "B", the cursor stored belongs to "A". Am I right? If that's the case, I think we should find a way to store a current cursor for each item, because when asking the server for more items, we have to send in the call the item and the current cursor for that item. What do you think?
- Would it be possible to hide the "Load more results" button when there's no more items to load? If I'm not wrong, as it's implemented right now, it will appear even if we don't have more items to be loaded

Please tell me what you think 🙂 thanks a lot for your work!

matuzalemsteles:
> As it's implemented right now, we can't have a "Load more results" button in the first level. In your example, it allows loading more children for an item. That's great (and we also need that) 😛 but in our case, we may have a lot of items in the first level, so it would be great if we find a solution for that as well.

Oh yeah, I think this was more of a problem with my example, let me improve on that.

> I've seen you are storing the current cursor in a ref, but I found it's common for all items, so if you load more results for item "A", and then you load more results for item "B", the cursor stored belongs to "A". Am I right? If that's the case, I think we should find a way to store a current cursor for each item, because when asking the server for more items, we have to send in the call the item and the current cursor for that item. What do you think?

Yeah you're right about that, my initial ideal is to keep a `ref` for each item, I think I have a bug in this implementation that doesn't do this to a certain degree. I'll improve this!

> Would it be possible to hide the "Load more results" button when there's no more items to load? If I'm not wrong, as it's implemented right now, it will appear even if we don't have more items to be loaded.

Oh yeah, that's a good point, I think this pull request implementation doesn't do that, but it makes a lot of sense, I have an idea for this I think it ends up solving the first question as well.

Thanks @sandrodw3 for your feedback this is super important to improve things ❤️

matuzalemsteles:
Send another commit that adds the single cursor per item and I also changed the API that is public from loadMore, now it has the same pattern as the others.

```typescript
type LoadMore = {
	has: (key: Key) => any;
	loadMore: <T>(
		id: React.Key,
		item: T,
		willToggle?: boolean
	) => Promise<void> | undefined;
};
```

The `has` method will help to know if there is still a cursor, otherwise we will not render the `Load more` button.

For the case of having the `Load more` for the root I think we have a certain problem here, the button renders and the render props context is based on an item. To have a `Load more` renderer for the root, we have two options here:

1. Load More and render the rais button outside the TreeView. (I don't really like this because it further increases the implementation for this)
2. Add a root item that groups them all.

I'm very leaning towards the second option because it makes things simpler and the current implementation works as it should too, the only point I'm thinking about here is whether it's worth doing this under the hood or letting this be explicit, maybe it's safer to make this explicit because the developer can add their own data that helps identify which item is root when the `loadMore` function is called, I also think there is always a root item in real cases. I updated the storybook example to follow this strategy. @sandrodw3 let me know what you think. I also think this is valid for the video you showed, where the root node is the `Pages` item.

sandrodw3:
Hey @matuzalemsteles, you are very fast! 😄 thanks again, I've been testing the storybook example, I'm leaving some feedback:

- The cursor per item solution looks and works very good! I only would think about one thing. As I understood, we are using `load.has !== null` to decide whether we show the "Load more" button or not. As all cursors are `undefined` the first time, I guess we are showing the button initially for all items, even if they don't have more children, and then, when clicking it the first time and setting the cursor to null, we are hiding it. It would be good to be able to not show the button initially when the item do not have more children. Maybe it would be enough to add a `paginated` prop to each item that is `true` when that item has more children that haven't been loaded, and then add `item.paginated` to the conditions to show the button. What do you think? As items come from server and the button's JSX is specified in each usage, maybe we don't need to do anything in the TreeView itself to allow this, I'm saying this just in case 😛
- The `load.has` method looks more like a `load.get` to me, I would expect a `has` to return `true` or `false`, and this method is giving us the current cursor value, right?
- About the root node, I don't know what to say 🤔 I've discussed it with our designers. As specified in [Lexicon guidelines](https://liferay.design/lexicon/core-components/tree-view/#type-of-nodes), we won't necessarily have a root node, so they think we should be able to have pagination in first level without it. If it complicates the implementation too much, maybe we can do what you said, secretly add a "virtual" root node always, and make it no-visible and expanded by default. We have another use case in the same tree that I showed you in which we don't have a root node (picture below). Sorry if this increases the implementation 😞

image

Thanks again for your fast and super good work! Let me know what you think 🙂

matuzalemsteles:
@sandrodw3 

> Maybe it would be enough to add a paginated prop to each item that is true when that item has more children that haven't been loaded, and then add item.paginated to the conditions to show the button. What do you think? As items come from server and the button's JSX is specified in each usage, maybe we don't need to do anything in the TreeView itself to allow this, I'm saying this just in case

Yes that makes perfect sense, I don't think we can do something on our side without knowing the data that can handle the condition mainly for this specific use case.

> The load.has method looks more like a load.get to me, I would expect a has to return true or false, and this method is giving us the current cursor value, right?

Yes you are right, the semantics are wrong here. I'll change it to `get` I think it's better for this one because it's better to distinguish from `null` and `undefined` and they are different scenarios.

> About the root node, I don't know what to say 🤔 I've discussed it with our designers...

Oh yeah, that true! Ok, my concern of adding a "virtual" root node is wondering if I would break some implementation, I think we would have to ignore this one in the root as well when it is exposed outside the TreeView, for example in the call of `items`, `onItemsChange`, maybe this makes things a little more complicated and successive errors.

I wanted to do a little bit of a step-by-step implementation of this with you to get more information on the implementation in DXP being used directly so we can figure things out before adding something final here.

I'm thinking the following now, I'm just going to make the change from `has` to `get` and merge this PR so I can include this in today's release. I will create an issue to continue working on how I can handle the implementation of the "virtual" root item, and for now while you work on the DXP side until while this is in feature flag, go for now with the solution of adding a node root just to have the data paginated to the first level and I continue work on that, I would say to add this item root directly in the TreeView definition to not make it complex, as was done in the example. It would also be great to hear back with some feedback as you test these changes, so feel free to message me directly via Slack to discuss the ideas.

sandrodw3:
That's perfect @matuzalemsteles! 😄 When the release is available to use the new changes in portal I'll start with the migration under the feature flag. I think we can go for now with what you have implemented already, using a root node in the TreeView usage, as you suggested. And I'll leave the migration under the feature flag while you work on the virtual root node issue. I'll let you know how goes the migration when I do it. Thanks again for all your work with this ❤️

PR #5105, opened 2022-09-22
feat(@clayui/css): Functions `map-deep-merge` shouldn't merge maps if `merge: false`

### LINKED ISSUE #4898: @clayui/css: Reset a variable containing a Sass map with `$c-unset`
From private slack conversation:

> What's the best way to override a map with an empty map or set it to null? I'm trying to override the following variable coming from Clay:
```
$clay-color-header-component-title: () !default;
$clay-color-header-component-title: map-deep-merge(
	(
		color: $gray-600,
		display: inline-block,
		font-size: 0.875rem,
		font-weight: $font-weight-semi-bold,
		max-width: calc(100% - 2rem),
	),
	$clay-color-header-component-title
);
```
> I've tried:
```
$clay-color-header-component-title: null;

$clay-color-header-component-title: ();

$clay-color-header-component-title: map-new();

$clay-color-header-component-title: $c-unset; // build error because it's not a map due to map-deep-merge
```

We should look into being able to reset variables containing Sass maps with `$c-unset` or providing an easier way to clear styles.

pat270:
This allows resetting a component to a blank state via Sass maps.

```scss
$btn: (
    merge: false,
);
```

will skip merging the `$btn` maps from atlas and base variables. It will let you define your own styles without having to sort through Clay's styles. We probably need to do the same for `map-merge`, but we will need to replace `map-merge` with `clay-map-merge`. I think it's better to have some discussion before we commit to something like that.

fixes #4898


packages/clay-css/src/scss/functions/_global-functions.scss
```diff
@@ -26,7 +26,7 @@
 	@return $toggle-switch-width + $label-spacer-x;
 }
 
-/// A function that returns a new map with all the keys and values including nested keys and values from both `$map1` and `$map2`. If both `$map1` and `$map2` have the same key, `$map2`’s value takes precedence.
+/// A function that returns a new map with all the keys and values including nested keys and values from both `$map1` and `$map2`. If both `$map1` and `$map2` have the same key, `$map2`’s value takes precedence. The key value pair `merge: false,` will skip merging maps. This allows component properties to be reset to a blank state.
```
matuzalemsteles, line 29: @pat270 what would `blank state` mean here?

packages/clay-css/src/scss/functions/_global-functions.scss
```diff
@@ -53,25 +53,29 @@
...
-				)
-			);
-		} @else {
-			$newMap: map-merge(
-				$newMap,
-				(
-					$key: $value,
-				)
-			);
+	@if (map-get($map2, merge) != false) {
```
matuzalemsteles, line 56: Looking at the issue it seems to me that the developer wants to prevent the class from being rendered in the final build, if that's the intention wouldn't that be the same thing as adding the `disabled` property?

pat270: It's almost the same. A `disabled` (we have `enabled` atm) property would prevent a dev from reusing the class name. It would not output any CSS for `.btn-primary` for example. This `merge: false` tells Clay not to merge the values I set for `.btn-primary` with Clay's.

```scss
$btn-primary: (
    merge: false,
    background-color: orange,
);
```
would only output:
```css
.btn-primary {
    background-color: orange;
}
```
instead of:
```css
.btn-primary {
    background-color: orange;
    border-color: transparent;
    color: #fff;
}
.btn-primary:hover {
    background-color: #0053f0;
    border-color: transparent;
    color: #fff;
}
/* etc */
```

I think this is the better way, eventually we should get rid of the `enabled` attribute.

matuzalemsteles: I see, I don't understand if this is really something that would help in most use cases, it's just a little weird because it seems to be a very exclusive thing to remove all nested statements from the output while keeping the first level.

For me it still makes sense to make an `enabled`, because if the person doesn't want everything from Clay he can just disable some classes or variations of a component, for example he can import only what he wants but he can't ignore the classes of a file , I still see value in `enabled` as a way to "remove" specific classes from a file.

I think avoiding nested merge is just a way to go with it to keep the first level style. It seems to me that setting `null` seems more intuitive in subclasses a variable to say it doesn't exit on output.

pat270: The `merge: false` also removes the class from the file. You just need to:

```scss
$btn-primary: (
   merge: false,
);
```
We would have no `.btn-primary {}` in the output.

Avoiding the merge is a QoL improvement for a dev that just wants to start styling `.btn-primary` without baggage from Clay. This isn't really needed, but thought it was a nice to have. I'm ok with adding it later if it gets requested.

matuzalemsteles: I see, yeah this looks like so the same function as `enabled`, I worry about this is that this would be the same thing as making the classes disabled and the developer adding `.btn-primary {...}` without any Clay resource.

I don't really like the idea of having this because the developer at the end is just overriding:

```sass
$btn-primary: (
   merge: false,
   background-color: orange,
);

// vs

$btn-primary: (
   enabled: false,
);

.btn-primary {
   background-color: orange,
}
```

This to me is two more different ways of offering this and not exactly a feature, I prefer the second option because it's explicit that you don't want a variance in the final output.

The first option wraps a lot of logic on top of the Clay only to end up with the same output as the second option, this easily bloats the Clay with resources it wouldn't need.

pat270: Yeah it would essentially be a duplicate of the second option. The only difference is that in the first option we have access to Clay's Sass api. It will be useful when/if we implement design tokens, it will translate directly to CSS. The second option will require the dev to update their `.btn-primary` CSS to be able to take advantage of our tokens.

matuzalemsteles: Hm yeah that sounds like a good point for that, what would that look like directly?

pat270: We would start with some JSON and convert it to Sass variables.

```json
{
  "btn-primary": {
    "merge": {
      "value": "false",
      "type": "other"
    },
    "background-color": {
      "value": "orange",
      "type": "color"
    },
    "border-color": {
      "value": "transparent",
      "type": "color"
    },
    "color": {
      "value": "{$white}",
      "type": "color"
    }
  }
}
```
would output:
```scss
$btn-primary: (
  merge: false,
  background-color: orange,
  border-color: transparent,
  color: $white,
);
```
The JSON format comes from https://www.figma.com/community/plugin/843461159747178978/Figma-Tokens. This would allow a designer using a Figma template we have put together to modify `.btn-primary` in Figma and output it for consumption by Clay. `merge` lets them start over if they wish or they can piggy back off our styles. `enabled` completely takes away the option of using `.btn-primary`.

matuzalemsteles: Yeah makes sense to me, I think that's a good point in favor of `merge` here. Now I'm not sure if `enabled` is equivalent to `merge`, so it looks like `merge` will be used for other purposes whereas `enabled` will just remove it from the final output. Feel free to go ahead with it.

pat270: I'm going to merge it, thanks @matuzalemsteles ! I'm just going to note that we can also do:
```scss
$btn-primary: (
    focus: (
        merge: false,
    ),
);
```
and it won't output the focus styles for `.btn-primary`.

Issue #5100, opened 2022-09-19
Adds the Collection pattern to the Autocomplete component


matuzalemsteles:
From #5032

The new Collection pattern adds consistency to all Clay components to decrease the learning curve, bring React patterns, improve component composition...to read more about the new pattern in Clay, read the [technical implementation of the RFC](https://github.com/liferay/clay/blob/master/docs/proposals/0004-collection-components.md).

matuzalemsteles:
Adding the Collection pattern to Autocomplete requires us to work on remodeling some things on Autocomplete, currently the component is quite verbose and doesn't provide many features like focus control, better integration with data provider, accessibility and needs `DropDown.ItemList` to render properly, some things we can abstract to simplify the implementation, there's a lot left for developers to implement that was a component composition and design.

We need to work on that first before implementing the Collection pattern. I'll create a separate issue to work on this before I get to this issue.

Issue #5041, opened 2022-08-16
Review the accessibility of behavior in DropDown


matuzalemsteles:
After making some accessibility improvements to DropDown, like #4998 which adds semantic role to correctly give context to Menu items and semantics to the DropDown trigger we also sent more improvements in #5038 which improves navigation when using virtual cursor in screen reader and adds the possibility to open DropDown using ArrowDown and Spacebar, move the focus to the first focusable DropDown element, and also correct semantics for groups but there are still some inconsistencies of behavior and shortcuts accordingly https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-links.

See the comment https://github.com/liferay/clay/pull/5038#issuecomment-1216013803 to understand more about it. This issue is mainly to understand some main behaviors that we have for years in the component that deviates from the accessibility rules.

- Tab within DropDown should not be possible
- Navigating using arrow keys should be looped

The idea here is to raise a discussion mainly to understand if this makes sense for the component mainly the tabulation that I remember was added to allow navigation using only keyboard, I think at the beginning we had put the navigation arrows in a loop and to provide for the user exiting the dropdown would be adding the tabulation I think we ended up seeing this the wrong way, the tab would probably not do navigation through the menu but close the menu and go to the next focusable element.

@marcoscv-work I would like to have your comments here too and @drakonux jfyi

marcoscv-work:
About patterns, from issue https://github.com/liferay/clay/issues/5147:

I'm going to try to explain the difference between patterns with focus to the fist element and dont move the focus until press tab or arrows (with menu opened):

Pattern: **Focus auto move to the first item**:

1. SR reads click me, I'm a popup collapsed button:
<img width="755" alt="image" src="https://user-images.githubusercontent.com/7963804/197184534-2d380dcb-578d-4b24-aa6a-a175816461d3.png">
2. We press the button and directly the SR reads: Button (or any first item).
<img width="753" alt="image" src="https://user-images.githubusercontent.com/7963804/197184926-90c58a5b-eac3-4810-b417-62921564b8ed.png">

Pattern:  **Keep the focus on the trigger and reads its change**:

1. SR reads click me, I'm a popup collapsed button:
<img width="755" alt="image" src="https://user-images.githubusercontent.com/7963804/197184534-2d380dcb-578d-4b24-aa6a-a175816461d3.png">
2. We press the button and the SR reads the state change.
[...]

PR #5108, opened 2022-09-24
feat(@clayui/autocomplete): adds new less verbose composition with more OOTB features

### LINKED ISSUE #5106: Review the Autocomplete component and implement improvements
To go ahead with the #5100 Collection pattern for this component we need to review some implementations to provide the Collection pattern, with data provider integration as well. Some things we need to reconsider and design of this component.

- Composition is a bit verbose and provides few OOTB features
- Developer needs to control focus
- The developer needs to control the visibility of the Menu according to the interaction with the input

Currently an Autocomplete component looks like this:

```jsx
<ClayAutocomplete>
  <ClayAutocomplete.Input
    aria-label="Numbers: Enter a number from One to Five"
    onChange={(event: any) =>
      setValue(event.target.value)
    }
    value={value}
  />

  <ClayAutocomplete.DropDown active={!!value}>
    <ClayDropDown.ItemList>
      {['one', 'two', 'three', 'four', 'five']
        .filter((item) => item.match(value))
        .map((item) => (
          <ClayAutocomplete.Item
            key={item}
            match={value}
            onClick={() => setValue(item)}
            value={item}
          />
        ))}
    </ClayDropDown.ItemList>
  </ClayAutocomplete.DropDown>
</ClayAutocomplete>
```

The goal is to redesign the component composition design and resolve the above issues. This way we were able to implement the collection pattern better.

An example of pseudocode:

```jsx
<Autocomplete
  defaultItems={['one', 'two', 'three']}
>
  {(item) => (
    <Autocomplete.Item>
      {item}
    </Autocomplete.Item>
  )}
</Autocomplete>
```

This is just an idea of what the component will look like already using the Collection pattern, it will no longer be necessary to declare the `Input` component it is automatically rendered inside Autocomplete and can be configured by passing properties directly to it.

Observing the usage in DXP, Autocomplete.Input it is more configured to control the focus state and manage the control of the active state, so for this scenario we can still continue providing these callbacks but these functionalities will be controlled by the component itself and reduced implementation on the usage side.

matuzalemsteles:
Closes #5106

Well, I did a more thorough review of Autocomplete and the documentation, our Lexicon documentation is a little old but it did help to add OOTB functionality.

The new composition maintains compatibility with the previous version and to use the new composition, it is no longer necessary to declare the Input or DropDown component, this ends up being more verbose when the developer needed to control all the states and some user behaviors.

```jsx
<Autocomplete
  messages={{
    loading: 'Loading...',
    notFound: 'No results found',
  }}
  placeholder="Enter a number from One to Five"
>
  {['one', 'two', 'three', 'four', 'five'].map((item) => (
    <Autocomplete.Item key={item}>
      {item}
    </Autocomplete.Item>
  ))}
</Autocomplete>
```

[**See example in Storybook**](https://deploy-preview-5108--next-storybook-clayui.netlify.app/?path=/story/design-system-components-autocomplete--default).

The `input` properties are passed directly to the `<Autocomplete />` component instead of declaring the `<Autocomplete.Input />`, the items to be rendered in DropDown are passed as `children` of the `<Autocomplete / >`, it is no longer necessary to render `<Autocomplete.DropDown />` in this way.

- Input states and DropDown active state follow [RFC 0002 Controlled and uncontrolled components](https://github.com/liferay/clay/blob/master/docs/proposals/0002-controlled-and-uncontrolled.md).
- We control the active state of DropDown based on user interactions, click away, exit focus, press Esc and click on the item.
- Automatic filter for static declaration.
- Not found state when no suggestion is found.
- Returns the previous value selected in the input when exiting the autocomplete when typing some value that does not exist.
- Focus control
- Combobox accessibility, virtual cursor support and hotkeys

There is another compositing possibility that makes the APIs more expressive for input and DropDown, there are some places in DXP that Autocomplete has customizations in the DropDown menu, I will check that next week and see if they are necessary, but we can still provide a simplified composition similar to what we have today but with the same features described above.

```jsx
<Autocomplete
  messages={{
    loading: 'Loading...',
    notFound: 'No results found',
  }}
>
  <Autocomplete.Input placeholder="Enter a number from One to Five" />
  <Autocomplete.DropDown>
    {['one', 'two', 'three', 'four', 'five'].map((item) => (
      <Autocomplete.Item key={item}>
        {item}
      </Autocomplete.Item>
    ))}
  </Autocomplete.DropDown>
</Autocomplete>
```

The difference is that the APIs are explicit and easy to customize DropDown for example, but I'm still more inclined to go with the implementation of this PR, that's because DropDown doesn't open doors for more complex use cases that exist in the DXP for not following Lexicon or reusing it for use cases other than autocomplete.

The step of implementing the Collection pattern and better integration with the data provider will be done by the #5100 issue. The data provider integration will provide features like controlling the OOTB loading notification states for the user, normally this was done by the developers and also this has integration with Suspense, ErrorBoundary and paginated data.

## ToDo
- [x] Apply some accessibility `role` rules properly for the dropdown
- [x] Tests
- [x] ~Documentation~


matuzalemsteles:
Well, I sent more final changes to the component mainly applying accessibility to the OOTB component following the specification https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html.

With that, the component also follows the hotkey recommendations but we still have a problem related to the tabulation that happens inside the Menu, according to the specification for both autocomplete and DropDown it shouldn't happen and it shouldn't move the focus between options. An issue about this has already been opened to review this behavior and better understand #5041.

I've also added several tests for the new behavior that cover all major use cases and accessibility rules.

matuzalemsteles:
Going ahead and merging, feel free to add any suggestions, this is not yet public anyway as we don't have documentation yet. We can go back and add changes.

PR #5113, opened 2022-09-28
fix(@clayui/css): Global Functions speed up `map-deep-get`

pat270:
- removes converting null and list to empty map to prevent running map-get on non existing maps
- removes error message and just returns null

Couldn't get it faster with a for loop and chaining `map-get`'s due to needing to use the `nth` function. This came out 500ms faster for 100,000 iterations of `map-deep-get` over the old one.


matuzalemsteles (approved):
LGTM!

PR #5119, opened 2022-09-30
fix(@clayui/css): Reduce the uses of the `setter()` function to speed up build

### LINKED ISSUE #5114: @clayui/css: Convert `setter()` to use Sass ternaries
Clay's `setter()` function adds overhead to the Sass build. The Sass ternary `if(true, true, false)` is faster. We should avoid using `setter()` where we can.

pat270:
fixes #5114

This is a really big change. I think it reduced the build time, but can't say for sure due to processes running on my computer. Here is a really small sample of `yarn compile` I did in the master branch.

```
✨  Done in 35.85s.
✨  Done in 35.51s.
✨  Done in 35.07s.
✨  Done in 34.73s.
```

This is a sample of build times on this branch.

```
✨  Done in 35.98s.
✨  Done in 35.37s.
✨  Done in 35.34s.
✨  Done in 34.21s.
```

My localized performance test said 100,000 iterations of `setter` vs `if` statement is ~900ms slower. I'm not sure if this is worth merging. I'm going to mark this as draft so we can discuss.

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_forms.scss
@@ -223,7 +223,8 @@ $input-danger-readonly: () !default;
 $input-danger-readonly: map-deep-merge(
 	(
 		background-color:
-			setter(
+			if(
+				map-get($input-readonly, bg),
 				map-get($input-readonly, bg),
 				map-get($input-readonly, background-color)
 			),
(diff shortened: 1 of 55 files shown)


matuzalemsteles:
I thought I commented on this, maybe I just commented about this on Slack.

I'm not sure if that could actually be a huge performance improvement, although `setter` is slower it could be that it's just noticeable in a very large amount as you did for 100,000 interactions, perhaps for our codebase this is not noticeable because we have a much smaller amount.

Even though you have other things running on your machine, this is validate the numbers because you compare the two changes in the same environment so this shouldn't be a problem.

pat270:
Closing this one, we can come back to it later if we need to fine tune performance.

PR #5121, opened 2022-10-01
feat(@clayui/autocomplete): add the Collection pattern and loading state

### LINKED ISSUE #5100: Adds the Collection pattern to the Autocomplete component

matuzalemsteles:
Closes #5100

This PR implements the Collection pattern to Autocomplete, this implements the OOTB features of the Collection such as list virtualization (enabled by default), static content and dynamic content.

Dynamic content has a filter by default as does static content. The filter can be customized when `items` is behaved as controlled, for example this is good for cases where the filter is done by an asynchronous request instead of the filter happening in the browser with all the data.

```jsx
<ClayAutocomplete
  defaultItems={['one', 'two', 'three', 'four', 'five']}
  messages={{
    loading: 'Loading...',
    notFound: 'No results found',
  }}
>
  {(item) => (
    <ClayAutocomplete.Item key={item}>
      {item}
    </ClayAutocomplete.Item>
  )}
</ClayAutocomplete>
```

In the example above the filter is enabled because the filter is done directly in the browser instead of an asynchronous request. The developer just passes the items to the `defaultItems` property which has the behavior of uncontrolled. When the filter is done via request the component must be controlled.

```tsx
const [value, setValue] = useState('');

const [networkStatus, setNetworkStatus] = useState<NetworkStatus>(
  NetworkStatus.Unused
);

const {resource} = useResource({
  fetchPolicy: FetchPolicy.CacheFirst,
  link: 'https://rickandmortyapi.com/api/character/',
  onNetworkStatusChange: setNetworkStatus,
  variables: {name: value},
});

<ClayAutocomplete
  isLoading={networkStatus < NetworkStatus.Unused}
  items={resource?.results ?? []}
  messages={{
    loading: 'Loading...',
    notFound: 'No results found',
  }}
  onChange={setValue}
  onItemsChange={() => {}}
  placeholder="Enter a name"
  value={value}
>
  {(item) => (
    <ClayAutocomplete.Item key={item.id}>
      {item.name}
    </ClayAutocomplete.Item>
  )}
</ClayAutocomplete>
```

### ToDo

- [x] Fix tests and add new use cases
- [x] Add new documentation
- [x] Add OOTB paginated data support with `useResource`

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
 		"react": "^16.12.0",
 		"react-dom": "^16.12.0",
 		"react-test-renderer": "^16.6.1",
+		"resize-observer-polyfill": "^1.5.1",
 		"sass-loader": "^7.1.0",
 		"size-limit": "^6.0.3",
 		"source-map-loader": "^0.2.4",
--- a/packages/clay-autocomplete/docs/api-autocomplete.mdx
+++ b/packages/clay-autocomplete/docs/api-autocomplete.mdx
@@ -8,24 +8,49 @@ mainTabURL: 'docs/components/autocomplete.html'
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
+-   [Autocomplete (Deprecated)](#autocomplete-deprecated)
 -   [Autocomplete](#autocomplete)
--   [Autocomplete.DropDown](#autocomplete.dropdown)
--   [Autocomplete.Input](#autocomplete.input)
+-   [Autocomplete.DropDown (Deprecated)](#autocomplete.dropdown)
+-   [Autocomplete.Input (Deprecated)](#autocomplete.input)
 -   [Autocomplete.Item](#autocomplete.item)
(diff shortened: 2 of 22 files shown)


matuzalemsteles:
This is ready to go now, I believe it is still worth adding more tests to test some use cases but the problem is that virtualization relies on browser APIs which are bad to mock in jest but I think to add more tests makes sense for now I will leave to work on that in the future.

I also rewrote the Autocomplete documentation to cover the component's new usage. I don't think it's worth keeping the documentation of the old behavior but I still kept the documentation of the old APIs for people to refer to.

The documentation is also more complete than before, showing all the features and possibilities that you can do with the component, following the TreeView pattern, VerticalBar.

PR #5122, opened 2022-10-03
fix(@clayui/css): Mixins check if parameter is map to avoid Sass error

### LINKED ISSUE #5120: @clayui/css: Some mixins aren't checking if parameter is type map
These will error if something other than a map is passed in.

```
clay-select-variant
clay-form-validation-variant
clay-custom-grid-columns
clay-generate-media-breakpoints
clay-after-highlight-variant
clay-input-group-stacked
clay-list-group-notification-item-variant
clay-loading-animation-variant
clay-menubar-vertical-expand
clay-modal-variant
clay-navbar-size
clay-slideout-variant
clay-sticker-size
clay-table-drag-variant
clay-table-clone-variant
clay-tbar-variant
clay-tbar-inline-down
clay-autofit-row
```

pat270:
fixes #5120

### DIFF
--- a/packages/clay-css/src/scss/mixins/_forms.scss
+++ b/packages/clay-css/src/scss/mixins/_forms.scss
@@ -701,280 +701,315 @@
 /// - Add @link to documentation
 
 @mixin clay-select-variant($map) {
-	$enabled: setter(map-get($map, enabled), true);
+	@if (type-of($map) == 'map') {
+		$enabled: setter(map-get($map, enabled), true);
 
-	$base: setter($map, ());
-	$base: map-merge(
-		$base,
-		(
-			background-color:
-				setter(map-get($map, bg), map-get($map, background-color)),
-			background-clip:
-				setter(map-get($map, bg-clip), map-get($map, background-clip)),
-			background-image:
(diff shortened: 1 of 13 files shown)


packages/clay-css/src/scss/mixins/_forms.scss
```diff
@@ -701,282 +701,317 @@
...
-				),
-			background-size:
-				setter(map-get($map, bg-size), map-get($map, background-size)),
-			color:
-				setter(
-					map-get($map, color),
-					if(
-						variable-exists(input-color),
-						$input-color,
+	@if (type-of($map) == 'map') {
```
matuzalemsteles, line 704: Hey @pat270 can we do an early return to not change the indentation?

```sass
@if (type-of($map) != 'map') {
  @return
}
```

pat270: @matuzalemsteles good idea, unfortunately mixins don't return values. I can't think of another way around it atm.

matuzalemsteles: Oh yes that's true, that's ok, let's go with it!

matuzalemsteles (approved):
LGTM

Issue #5125, opened 2022-10-06
ClayTreeView drag preview does not show name


sandrodw3:
## Is there an example you can provide?
https://storybook.clayui.com/?path=/story/design-system-components-treeview--dynamic

## What are the steps to reproduce?
Just dragging an item.

## What is the expected result?
Item's name should be shown in the drag preview. Instead, an empty white square is being displayed.

sandrodw3:
I've sent a PR for this one https://github.com/liferay/clay/pull/5126 😄
By the way, I've added the possibility to specify a different key for item's name from outside. Let me know what you think, thanks!

PR #5126, opened 2022-10-06
feat(@clayui/core): add possibility to specify item name key

sandrodw3:
This fixes https://github.com/liferay/clay/issues/5125 and also adds the possibility to specify a different key for item's name from outside. I've also added a check in the drag preview to prevent it to be shown in other drag and drop context, as discussed with @matuzalemsteles in Slack 🙂

### DIFF
--- a/packages/clay-core/src/tree-view/DragLayer.tsx
+++ b/packages/clay-core/src/tree-view/DragLayer.tsx
@@ -41,7 +41,7 @@ function getItemStyles(
 	};
 }
 
-const DragLayer = () => {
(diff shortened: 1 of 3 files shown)


matuzalemsteles (approved):
Thanks @sandrodw3 for send this, I just sent a commit to remove the `Item Name Key` from the context and `useTree`, we don't need them there, so it wouldn't be necessary to add more data to the context.

matuzalemsteles:
No problem @sandrodw3, thanks again for submitting the PR!

PR #5127, opened 2022-10-06
feat(@clayui/core): adds API to set heading fontSize independent of tag

### LINKED ISSUE #5110: Header level and Style should not be directly and unbreakable connected
Styles and Header level are not the same, an H1 should not be connected to a specific style.
Lexicon does not determinants this relationship, Lexicon just defines a set of styles:
https://liferay.design/lexicon/foundations/typography/#headings

Currently in heading component we can define the level and the label has associated a style ´text-´.
https://clayui.com/docs/components/heading.html

Initially I thought it could be changed by ´className´ but is not possible.

**Proposal:**
If we add an API new customization to change the style.. it could work, and the default style could persist.

matuzalemsteles:
Closes #5110

### DIFF
--- a/packages/clay-core/src/typography/Heading.tsx
+++ b/packages/clay-core/src/typography/Heading.tsx
@@ -6,7 +6,7 @@
 import classNames from 'classnames';
 import React from 'react';
 
-export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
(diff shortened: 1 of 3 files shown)


marcoscv-work:
Thanks @matuzalemsteles !!  :-D

PR #5129, opened 2022-10-06
feat(@clayui/css): Decouple button variables from other components

### LINKED ISSUE #4699: @clayui/css: Decouple btn variables from other components
We are reusing `$btn-*` variables in other components. This make it difficult to include only the component variable files you need. We can limit the use of `$btn-*` variables only to buttons and create global variables that can be used everywhere (if needed).

pat270:
fixes #4699

This removes the dependency on button variables in other components. I didn't fully update Cadmin, since we don't expect it to be changed much. We can convert if in the future if the need arises.


matuzalemsteles (approved):
LGTM! Makes sense Patrick!

matuzalemsteles:
@pat270 can you take care of fixing the conflicts? feel free to merge when you're done.

PR #5133, opened 2022-10-10
feat(@clayui/core): add possibility to specify that an item has children, so caret is shown even if children haven't loaded

sandrodw3:


sandrodw3:
This solves https://github.com/liferay/clay/issues/5134. With this proposal, adding `hasChildren` to an item will make the expander visible even if that item's children haven't been loaded. Please let me know what you think, thanks in advance 🙂

packages/clay-core/src/tree-view/TreeViewItem.tsx
```diff
@@ -565,7 +565,8 @@ export function TreeViewItemStack({
 	const nestedChildren =
 		nestedKey && (item[nestedKey] as Array<Record<string, any>>);
 
-	const hasChildren = nestedChildren && nestedChildren.length > 0;
+	const hasChildren =
+		(nestedChildren && nestedChildren.length > 0) || item.hasChildren;
```
matuzalemsteles, line 569: Hey @sandrodw3 it's a good idea, I think we can do it in another way so as not to force a property called `hasChildren` to exist in the tree data, I think we can add an API for `TreeView.Item` and `TreeView.ItemStack` this API with that name, so you can use any other data to identify that there are children in the item.

```jsx
<TreeView>
  {(item) => (
    <TreeView.Item expandable={item.children.length}>
      {item.name}
    </TreeView.Item>
  )}
</TreeView>
```

We can use an API that we use internally to make this `expandable` and make it public, I think we also need to improve how this is being done now because when there is no children, we are still rendering a button with no content when the `loadMore` is added, I think this was done wrong, I will send some commits on top of your PR to improve this too and make this logic simpler.

matuzalemsteles:
Hey @sandrodw3 I made some modifications to make the `expandable` API public, this allows you to add some rule to show the expander button even if it has no children loaded in memory, this allows you to use any data to decide this.

I also made some fixes, for example clicking the expander button was not loading data when async mode was set.

I ended up doing some refactors in the implementation to simplify the way to decide if the expander button should appear or not, this was very confusing in the implementation before. I'll go ahead with that PR and merge it to include in today's release, but let me know what you think and we can at any time submit another PR to make changes.

packages/clay-core/src/tree-view/TreeViewItem.tsx
sandrodw3: Hey @matuzalemsteles that's definitely a better solution, thanks! 🙂

sandrodw3:
Hi @matuzalemsteles! Cool, the `expandable` API is definitely a better solution. Thanks again for your fast work. I'll test these changes when they are available in portal and will give you some feedback.

Issue #5135, opened 2022-10-11
ClayTab.Item should set aria-disabled="false" since it hides it from the accessibility tree


victorg1991:
Here we are setting the aria-disabled property to false https://github.com/liferay/clay/blob/58c04f399ddc26e19e17fb571ef67d5351160697/packages/clay-tabs/src/Item.tsx#L55

But this is not needed since the tab that is not selected shouldn't be disabled, we just need to set the aria-selected to false

More info about the tabs pattern here: https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/

PR #5136, opened 2022-10-11
fix(@clayui/tabs) Do not set aria-disabled since this removes the ele…

victorg1991:
…ment from the accesibility tree

Fixes https://github.com/liferay/clay/issues/5135

### DIFF
--- a/packages/clay-tabs/src/Item.tsx
+++ b/packages/clay-tabs/src/Item.tsx
@@ -49,7 +49,11 @@ const Item = React.forwardRef<any, IProps>(
 		}: IProps,
 		ref
 	) => (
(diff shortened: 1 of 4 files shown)


matuzalemsteles (approved):
Hey @victorg1991 thanks for submitting this, it makes a lot of sense. I send some more commits to regenerate the snapshot and also improve accessibility at some points following the https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic.html. 

I didn't add all the accessibility improvements because we would need more changes also in the way it handles the states, so it's quite likely that this component will be simpler and more accessible in an OOTB way when we apply the collection pattern #5032.

PR #5142, opened 2022-10-13
fix(@clayui/css): Removes unused sm, lg, xl breakpoint styles from Navigation, Management and Application Bars

### LINKED ISSUE #3236: Clay CSS Navbar, Application Bar, Management Bar, Navigation Bar only output `.navbar-expand` and `.navbar-expand-md` styles
## What is your proposal?
This is a breaking change. Right now we output styles for `.navbar-expand`, `.navbar-expand-{sm|md|lg|xl}`. The same goes for `.application-bar.navbar-expand-{sm|md|lg|xl}`, `.management-bar.navbar-expand-{sm|md|lg|xl}`, `.navigation-bar.navbar-expand-{sm|md|lg|xl}`. We should only output the base `.navbar-expand` and `.navbar-expand-md` versions since we only use the `md` breakpoint in DXP / Lexicon.

Saves ~40kb of unminified CSS. There will be an option to enable a specific `navbar-expand` breakpoint via Sass map. If you want to enable the `lg` breakpoint:
```
$navbar-expand: (
    xs: true,
    sm: false,
    md: true,
    lg: true,
    xl: false
);
```

Output styles for all breakpoints, but give option to disable output with `$navbar-expand`.

pat270:
fixes #3236

I was hoping to completely remove the dependency to the mixin `clay-navbar-size`, but it's too breaking. I moved as much as I could to use `clay-navbar-variant`. This reduces the size of Clay CSS and Cadmin by 42kb  and 29kb respectively.

### DIFF
--- a/packages/clay-css/src/scss/cadmin/variables/_management-bar.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_management-bar.scss
@@ -32,6 +32,7 @@ $cadmin-management-bar-size: map-deep-merge(
 		min-height: 64px,
 		padding-x: 0,
 		padding-y: 0,
+		font-size-mobile: $cadmin-navbar-font-size,
 		height-mobile: 48px,
 		min-height-mobile: 48px,
 		btn-monospaced-font-size: 16px,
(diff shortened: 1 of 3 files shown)


matuzalemsteles (commented):
@pat270 my only concern with this is that we can break developers who may be using it in themes for example, the bad thing about this is that we can't know if they are being used.

I'm wondering if we could make it so that by default in DXP we could remove this by default but in the 3.x build we still keep it. Do you think this would be an ideal option?

pat270:
@matuzalemsteles my gut tells me no one is using them, but probably better to be safe. We can add the variable changes to classic theme in `_clay_variables.scss`.

matuzalemsteles:
I also imagine this but I have this concern of breaking something and someone is using it. We need some way to detect the usage of Clay stuff by clients, it would help us a lot to make decisions like this.

pat270:
@matuzalemsteles I removed the breaking changes for Clay. I left the changes in for Cadmin. With the mixin updates, we have the option to remove styles for breakpoints we don't use.

matuzalemsteles (approved):
Sounds good @pat270, I think this is an option to remove things in cadmin that we have full control and then move those decisions to the public as atlas in future versions with a better deprecation policy.

PR #5150, opened 2022-10-18
feat(@clayui/drop-down): add collection pattern

### LINKED ISSUE #5101: Adds the Collection pattern to the DropDown component
From #5032

The new Collection pattern adds consistency to all Clay components to decrease the learning curve, bring React patterns, improve component composition...to read more about the new pattern in Clay, read the [technical implementation of the RFC](https://github.com/liferay/clay/blob/master/docs/proposals/0004-collection-components.md).

matuzalemsteles:
Closes #5101

In order to use the internal component of Collection in DropDown, I had to deal with the circular dependency problem, for that I had to remove the DropDown re-export in the `@clayui/core` package. I believe that no one is still using DropDown through the core, if so I'll update in DXP.

The Collection pattern is now available in the DropDown component, providing features such as static and dynamic content. This lets you simplify some simpler cases, for example how to just render a simple list of options.

```jsx
<DropDown
  items={['one', 'two', 'three', 'four']}
  trigger={<Button>Click Me</Button>}
>
  {(item) => <DropDown.Item key={item}>{item}</DropDown.Item>}
</DropDown>
```

When the component renders for example an input to filter the list, it now works in an OOTB way without having to configure anything extra.

```jsx
<DropDown
  trigger={<Button>Click Me</Button>}
>
  <DropDown.Search placeholder="Type to filter" />
  <DropDown.ItemList items={['one', 'two', 'three', 'four']}>
    {(item) => <DropDown.Item key={item}>{item}</DropDown.Item>}
  </DropDown.ItemList>
</DropDown>
```

This also works for groups, see the example in the storybook. This simplifies the composition to behave according to your data and also agnostic.

There are other features that we could add, for example virtualization for DropDown is disabled because the markup hierarchy level is a tree style, for example in groups. We can change this to render as a list but keep accessibility so we can enable virtualization by default and have no issues with it. We would also be able to add infinite scroll support but I haven't seen at least no case in DXP that would use this, normally DropDown is just used for Menu with a small list less than 20 items.

## ToDo

- [x] Update documentation
- [x] Update tests and add new use cases

### DIFF
--- a/packages/clay-autocomplete/package.json
+++ b/packages/clay-autocomplete/package.json
@@ -27,6 +27,7 @@
 	],
 	"dependencies": {
 		"@clayui/core": "^3.78.2",
+		"@clayui/drop-down": "^3.78.2",
 		"@clayui/form": "^3.78.2",
 		"@clayui/loading-indicator": "^3.60.0",
 		"@clayui/shared": "^3.78.2",
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {DropDown, __NOT_PUBLIC_COLLECTION} from '@clayui/core';
+import {__NOT_PUBLIC_COLLECTION} from '@clayui/core';
+import DropDown from '@clayui/drop-down';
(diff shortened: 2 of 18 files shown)


matuzalemsteles:
I'll leave this PR to be merged after we do this week's release to avoid having any issues and because we have a lot of important issues that should get into DXP faster.

PR #5152, opened 2022-10-20
fix(@clayui/drop-down): fix focus control error losing order and wrong behavior of Esc

### LINKED ISSUE #5140: Clay Dropdown Keyboard Accessibility Improvements

https://clayui.com/docs/components/drop-down.html

1. Navigate to https://clayui.com/docs/components/drop-down.html
2. Press `Tab` until the dropdown trigger is focused.
3. Press `Enter` or `Space` to open the dropdown.

When the Menu is open, `Esc`  should exit the dropdown and focus the trigger.
- Actual Result: The focus is moving to the bottom of the page instead of focusing the correct element.
When the Menu is closed and the trigger is focused, `Esc`  should NOT open the dropdown.
- Actual Result: `Esc` will open the menu when it's closed.
When the Menu is open, `Tab` should exit the dropdown and focus the trigger.
- Actual Result: The focus is moving to the bottom of the page instead of focusing the correct element.

https://www.w3.org/WAI/ARIA/apg/patterns/menu/
- Tab and Shift + Tab:
    - When focus is on a menuitem in a menu or menubar, move focus out of the menu or menubar, and close all menus and submenus.

### LINKED ISSUE #5147: @clayui/drop-down: Tabbing through a dropdown menu returns focus to beginning of page
https://clayui.com/docs/components/drop-down.html

1) Open the dropdown menu
2) Press tab to cycle through each item
3) When reaching the end of the menu, focus returns to the browser's tab

Not exactly sure what it's supposed to do.

- Return to the first item in the menu?
- Close the dropdown and focus the next element?

matuzalemsteles:
Fixes #5147, fixes #5140

This PR fixes the two issues related to focus control accessibility losing which the next and previous element should be focused on the Menu, it also fixes the esc is opening the menu when the focus is on the trigger, and when the menu closes the focus is lost.

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -164,12 +164,6 @@ function ClayDropDown({
 		value: active,
 	});
 
-	const handleKeyUp = (event: React.KeyboardEvent<HTMLElement>) => {
-		if (event.key === Keys.Esc) {
-			setInternalActive(!internalActive);
-		}
-	};
-
 	useEffect(() => {
 		if (internalActive) {
 			const onFocus = (event: FocusEvent) => {
@@ -221,7 +215,6 @@ function ClayDropDown({
 				<ContainerElement
 					{...otherProps}
 					className={classNames('dropdown', className)}
-					onKeyUp={handleKeyUp}
 				>
 					{React.cloneElement(trigger, {
 						'aria-controls': ariaControls,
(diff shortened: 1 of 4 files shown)


marcoscv-work:
Looks good, it fixes the issue so go ahead!

(More comments about the focus pattern in the Issue)

matuzalemsteles:
@marcoscv-work I'll leave to make these changes using another issue so we can organize things.

PR #5154, opened 2022-10-21
feat(@clayui/tabs): add new API to configure browsing behavior between tabs

### LINKED ISSUE #5151: Tabs: Tabs are not operable by keyboard
We cannot move along tabs by keyboard using tab or arrows.

image

https://clayui.com/docs/components/tabs.html#content

**Why do we need this functionality?**
WCAG 2.1 - Operable: 2.1 Keyboard Accessible

We should be able to get all tabs by keyboard, there are different ways to get it, most popular is the combination of tab and arrow keys, examples:

- https://getbootstrap.com/docs/5.2/components/navs-tabs/#javascript-behavior
- https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual (manual activation)
- https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic (automatic activation)

If possible, a API property to select if the activation is automatic or manual would be awesome, if not, manual activation is a more common case for liferay

matuzalemsteles:
Fixes #5151

In order to be able to control the automatic activation state of the tab when navigating, we have to have control of the active state, in the current implementation, the developer controls the state. In this PR I am applying the pattern from [RFC 0002 Controlled and uncontrolled components](https://github.com/liferay/clay/blob/master/docs/proposals/0002-controlled-and-uncontrolled.md) to allow the component to have state visibility, this makes it possible during tab navigation when selecting a tab changes the current tab.

```jsx
const [active, setActive] = useState(0);

<>
  <ClayTabs
    active={active}
    onActiveChange={setActive}
  >
    <ClayTabs.Item
      innerProps={{
        'aria-controls': 'tabpanel1',
      }}
    >
      Tab 1
    </ClayTabs.Item>
    <ClayTabs.Item
      innerProps={{
        'aria-controls': 'tabpanel2',
      }}
    >
      Tab 2
    </ClayTabs.Item>
  </ClayTabs>
  <ClayTabs.Content activeIndex={active} fade={args.fade}>
    <ClayTabs.TabPane id="tabpanel1">
      Tab 1
    </ClayTabs.TabPane>
    <ClayTabs.TabPane id="tabpanel2">
      Tab 2
    </ClayTabs.TabPane>
  </ClayTabs.Content>
</>
```

This makes it simpler to handle the active state of tabs. This component can be made simpler and provide some OOTB features, such as not needing to configure `aria-controls` to associate the tab with the panel, but this will only be possible when we start to apply the collection pattern in this component #5032.

Also before that we need to review this component to improve the composition so that the `ClayTabs.Content` component doesn't become like sibling of `ClayTabs` to improve the data flow, in this PR if the `active` is uncontrolled not work because it doesn't have how to pass the state of `active` to `ClayTabs.Content` for it to work correctly needs to be controlled, so this is one of the reasons to refresh the composition of Tabs.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -212,10 +212,10 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-tabs/src/': {
-			branches: 84,
+			branches: 58,
 			functions: 66,
-			lines: 94,
-			statements: 90,
+			lines: 78,
+			statements: 76,
 		},
 		'./packages/clay-time-picker/src/': {
 			branches: 85,
--- a/packages/clay-tabs/docs/index.js
+++ b/packages/clay-tabs/docs/index.js
@@ -13,48 +13,42 @@ const tabsImportsCode = `import ClayTabs from '@clayui/tabs';
 `;
 
 const tabsCode = `const Component = () => {
-    const [activeTabKeyValue, setActiveTabKeyValue] = useState(0);
+    const [active, setActive] = useState(0);
 
     return (
         <>
-            <ClayTabs modern>
+            <ClayTabs active={active} modern onActiveChange={setActive}>
                 <ClayTabs.Item
-                    active={activeTabKeyValue === 0}
                     innerProps={{
                         'aria-controls': 'tabpanel-1',
                     }}
-                    onClick={() => setActiveTabKeyValue(0)}
                 >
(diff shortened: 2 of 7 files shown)


marcoscv-work:
It looks great!

PR #5156, opened 2022-10-24
fix(@clayui/color-picker): Update hue slider to use html range input

### LINKED ISSUE #5137: Color Picker use input type range
https://github.com/liferay/clay/issues/5132

https://clayui.com/docs/components/color-picker.html

The hue selector should use `<input type="range">`.

pat270:
fixes #5137

I finally got this working with the range input. I used the `onKeyUp` and `onPointerUp` events to update the Color Picker. The only problem with this is the map's hue doesn't update while dragging the thumb. Since I have something working, I can probably get it to work with `onChange` only, just need to throttle or debounce the updates so it doesn't lag so much. 

In https://github.com/liferay/clay/commit/87331f613a43444d1a6b527137421774161cd7f0#diff-47b43a95fa966e5189545f95207e5d85c86bb0a2ac02963ec706530728d79af5R39-R41, I had to blur then re-focus the input because FF has a range input issue when using `onPointerUp`. It fails to update the input value for some reason. This problem doesn't exist when using `onMouseUp`.

### DIFF
--- a/packages/clay-color-picker/src/Editor.tsx
+++ b/packages/clay-color-picker/src/Editor.tsx
@@ -186,45 +186,13 @@ export function Editor({
 				</div>
 			</div>
 
-			<div className="clay-color-form-group">
-				<Hue
-					onChange={(hue) => {
-						onHueChange(hue);
-						onColorChange(tinycolor({h: hue, s, v}));
-					}}
-					value={hue}
-				/>
-				<ClayInput.Group>
-					<ClayInput.GroupItem>
-						<ClayInput
-							data-testid="hInput"
-							insetBefore
-							max="360"
-							min="0"
-							onChange={(event) => {
-								const value = event.target.value;
-
-								let newVal = Number(value);
-
-								if (newVal < 0) {
-									newVal = 0;
-								} else if (newVal > 360) {
-									newVal = 360;
-								}
-
-								onHueChange(Math.round(newVal));
-
-								onColorChange(tinycolor({h: newVal, s, v}));
-							}}
-							type="number"
-							value={hue}
-						/>
-						<ClayInput.GroupInsetItem before tag="label">
-							H
-						</ClayInput.GroupInsetItem>
-					</ClayInput.GroupItem>
-				</ClayInput.Group>
-			</div>
+			<Hue
+				onChange={(hue) => {
+					onHueChange(hue);
+					onColorChange(tinycolor({h: hue, s, v}));
+				}}
+				value={hue}
+			/>
 
 			<div className="clay-color-footer">
 				<ClayForm.Group>
--- a/packages/clay-color-picker/src/Editor.tsx
+++ b/packages/clay-color-picker/src/Editor.tsx
@@ -186,13 +186,45 @@ export function Editor({
 				</div>
 			</div>
 
-			<Hue
-				onChange={(hue) => {
-					onHueChange(hue);
-					onColorChange(tinycolor({h: hue, s, v}));
-				}}
-				value={hue}
-			/>
+			<div className="clay-color-form-group">
+				<Hue
+					onChange={(hue) => {
+						onHueChange(hue);
+						onColorChange(tinycolor({h: hue, s, v}));
+					}}
+					value={hue}
+				/>
+				<ClayInput.Group>
+					<ClayInput.GroupItem>
+						<ClayInput
+							data-testid="hInput"
+							insetBefore
+							max="360"
+							min="0"
+							onChange={(event) => {
+								const value = event.target.value;
+
+								let newVal = Number(value);
+
+								if (newVal < 0) {
+									newVal = 0;
+								} else if (newVal > 360) {
+									newVal = 360;
+								}
+
+								onHueChange(Math.round(newVal));
+
+								onColorChange(tinycolor({h: newVal, s, v}));
+							}}
+							type="number"
+							value={hue}
+						/>
+						<ClayInput.GroupInsetItem before tag="label">
+							H
+						</ClayInput.GroupInsetItem>
+					</ClayInput.GroupItem>
+				</ClayInput.Group>
+			</div>
 
 			<div className="clay-color-footer">
 				<ClayForm.Group>
--- a/packages/clay-color-picker/src/Hue.tsx
+++ b/packages/clay-color-picker/src/Hue.tsx
@@ -3,9 +3,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import React, {useState} from 'react';
-
 import ClaySlider from '@clayui/slider';
+import React, {useState} from 'react';
 
 import {usePointerPosition} from './hooks';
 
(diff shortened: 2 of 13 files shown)


packages/clay-color-picker/src/Hue.tsx
```diff
@@ -20,66 +20,56 @@ type Props = {
...
 const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
 	const containerRef = React.useRef<HTMLDivElement>(null);
-	const selectorActive = React.useRef<boolean>(false);
 
-	const {onPointerMove, setXY, x, y} = usePointerPosition(containerRef);
+	const {setXY, y} = usePointerPosition(containerRef);
 
-	const removeListeners = () => {
-		selectorActive.current = false;
+	const [internalValue, setInternalValue] = useState<number>(value);
```
matuzalemsteles, line 31: We don't need another state here, because we also have a state in the parent component, in which case we have access to them which are `value` and `onChange` it is easier to maintain a source of truth than to have multiple states and keep synchronizing them all, this is also bad because of unnecessary rerenders.

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -20,66 +20,56 @@ type Props = {
...
-
 /**
  * Renders Hue component
  */
 const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
 	const containerRef = React.useRef<HTMLDivElement>(null);
-	const selectorActive = React.useRef<boolean>(false);
 
-	const {onPointerMove, setXY, x, y} = usePointerPosition(containerRef);
+	const {setXY, y} = usePointerPosition(containerRef);
```
matuzalemsteles, line 29: We don't need this hook anymore and we don't need to set `setXY` anymore because we are no longer using them here. It used to be just used to get `x` information to deal with the position of the pointer, as we don't have it anymore, we don't need it anymore.

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -20,66 +20,56 @@ type Props = {
...
+		setXY({x: internalValue, y});
 
-	useIsomorphicLayoutEffect(() => {
-		if (containerRef.current && selectorActive.current) {
-			onChange(xToHue(x, containerRef.current));
-		}
-	}, [x]);
+		event.target.blur();
+
+		event.target.focus();
```
matuzalemsteles, line 40: I don't understand very well why we need this 😅, could you explain it better?

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -20,66 +20,56 @@ type Props = {
...
+	};
 
 	React.useEffect(() => {
-		if (containerRef.current) {
-			setXY({x: hueToX(value, containerRef.current), y});
-		}
+		setInternalValue(value);
```
matuzalemsteles, line 44: As we are going to use the parent state `value` and `onChange` we no longer need this to synchronize the values.

packages/clay-color-picker/src/Editor.tsx
```diff
@@ -186,13 +186,45 @@ export function Editor({
...
+								onColorChange(tinycolor({h: newVal, s, v}));
+							}}
+							type="number"
+							value={hue}
+						/>
+						<ClayInput.GroupInsetItem before tag="label">
+							H
+						</ClayInput.GroupInsetItem>
+					</ClayInput.GroupItem>
+				</ClayInput.Group>
```
matuzalemsteles, line 226: As we are dealing with internal components we can move this composition to the Hue component, it is easier to keep the responsibility organized. As it is not an internal component, there is no problem rendering the bar along with the input.

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -20,66 +20,56 @@ type Props = {
...
+	const {setXY, y} = usePointerPosition(containerRef);
 
-	const removeListeners = () => {
-		selectorActive.current = false;
+	const [internalValue, setInternalValue] = useState<number>(value);
 
-		window.removeEventListener('pointermove', onPointerMove);
-		window.removeEventListener('pointerup', removeListeners);
-	};
+	const handleOnChangeEnd = (event: any) => {
```
matuzalemsteles, line 33: Since we won't have the internal state anymore, just add the `onChange` callback as a property of the `ClaySlider`, we won't need that anymore because we will be directly updating the parent state instead of the internal one.

matuzalemsteles (commented):
Hey @pat270 there are some things we need to think about before going ahead with this and maybe improve some things, more in relation to the implementation of React.

It also seems that we have some problems with Slider as you said, I had made some changes to try to understand but it seems that React has some problems with input range, I will continue researching more about this but maybe we will have to not use the input range but we can still make it accessible.

I'll leave some comments here about the improvements that can be made.

packages/clay-color-picker/src/Hue.tsx
pat270: I think there is a bug in FF input range with `onPointerUp`. It keeps dragging the slider thumb even when the left mouse button is released. This didn't happen when I used `onMouseUp`. The only way I could get it to work properly was blur then focus.

pat270:
@matuzalemsteles Thanks for the input, I'll update this pr with your suggestions.

The problem is with the `onChange`, `onHueChange`, etc functions. React ends up re-rendering the slider when the value changes and the mouse loses focus. I was thinking if we could run `onHueChange` in all the other components, but not in slider. It would work.

matuzalemsteles:
> The problem is with the onChange, onHueChange, etc functions. React ends up re-rendering the slider when the value changes and the mouse loses focus. I was thinking if we could run onHueChange in all the other components, but not in slider. It would work.

I see, I'm not sure if that would be the problem because internally the ClaySlider has the `useInternalState` state, so if changing the slider value it causes a re-render, we wouldn't be doing anything different, other than controlling the state would stay in the ColorPicker instead of the slider when we are using the `value` and `onChange` properties.

I had made some changes to just use Hue's `onChange`, but the behavior is very strange that it loses the "focus" of the range when the mouse goes outside the bar, this is strange because the focus is still in the range if you monitor the `document.activeElement`, it doesn't make much sense.

https://user-images.githubusercontent.com/13750819/198695091-22d3c0c9-e420-4244-8371-f36aab2cd07c.mov

pat270:
The recording was exactly what was happening to me. `useState` seems to make this a non issue. It's the way the `onChange` functions work in the Editor imo. I'm trying `useState` with the alpha slider. I'll send a pr once I get swatches updating and we can compare.

packages/clay-color-picker/src/Hue.tsx
pat270: I couldn't do this one due to https://github.com/liferay/clay/pull/5156#issuecomment-1295093670.

packages/clay-color-picker/src/Hue.tsx
matuzalemsteles: Hmm this is weird, I'll do some testing.

matuzalemsteles:
@pat270 I think the most ideal thing here to go with this would be to use `onChange` to update the Hue state so we can see the editor update as the slider changes its value. I'm trying to investigate why the mouse is getting lost when trying to move the slider.

matuzalemsteles:
Hey @pat270 I think we're going to have to refactor how we implement the ClaySlider to behave differently and that the mouse moves work correctly when the onChange changes, maybe the `input[type=range]` maybe it has to be just the thumb and we will have to control movement via JS so it doesn't bug when state update happens.

PR #5158, opened 2022-10-25
fix(@clayui/shared): fix error when fiberNode may come null

matuzalemsteles:
This is a very small fix that happened in a test scenario in DXP, it seems that in some cases the `fiberNode` can return `null` which causes a TypeError to be thrown and the application breaks, from what I noticed this only happens in the test but it doesn't happen in the browser, so it may be some different behavior just in the tests.


matuzalemsteles:
I'm merging this to make a release with this fix.

PR #5162, opened 2022-10-26
feat(@clayui/core): add index data when moving item in TreeView

### LINKED ISSUE #5141: ClayTreeView: Improvements for drag and drop
Hey guys, I've been testing the newest drag and drop improvements in ClayTreeView and I have some feedback!

- Currently we are using the `draggable` prop for both drag and drop. It would be good to have them separated, as it could happen that an item is draggable but not droppable (it’s happening in our use case for example).
- If we do it and add `droppable` prop, it would be awesome if it was a callback that gets the dragged item, or having some way to condition it to the dragged item. Sometimes, an item is droppable for some type of item but not droppable for the rest.
- About `onItemMove`, I have found out that with the current API, we can’t know the new index of the item if we are moving it to a different position inside the same parent. Would it be possible to expose also the new index? So we can pass to the component a function like `(item, parentItem, newIndex) => {…}`

What do you think? 😄 (cc @matuzalemsteles)

matuzalemsteles:
Closes #5141

This is a draft PR, I should add another `onItemHover` API tomorrow and add some usage examples. There is one consideration is that `onItemHover` probably doesn't block any or disable the item from being droppable, so you will have to have the support of a state outside the TreeView to do the conditional using the `draggable` API on the Item. I will study if we can do something internally to improve the DX of use.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -67,6 +67,12 @@ interface ITreeViewProps<T>
 	itemNameKey?: string;
 
 	/**
+	 * The callback is called whenever there is an item dragging over
+	 * another item.
+	 */
+	onItemHover?: (item: T, parentItem: T, index: MoveItemIndex) => void;
+
+	/**
 	 * Callback is called when an item is about to be moved elsewhere in the tree.
 	 */
 	onItemMove?: (item: T, parentItem: T, index: MoveItemIndex) => boolean;
@@ -128,6 +134,7 @@ export function TreeView<T>({
 	items,
 	nestedKey = 'children',
 	onExpandedChange,
+	onItemHover,
 	onItemMove,
 	onItemsChange,
 	onLoadMore,
@@ -171,6 +178,7 @@ export function TreeView<T>({
 		expanderClassName,
 		expanderIcons,
 		nestedKey,
+		onItemHover,
 		onItemMove,
 		onLoadMore,
 		onRenameItem,
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -13,7 +13,7 @@ import {ChildrenFunction, Collection, ICollectionProps} from './Collection';
 import DragLayer from './DragLayer';
 import {TreeViewGroup} from './TreeViewGroup';
 import {TreeViewItem, TreeViewItemStack} from './TreeViewItem';
-import {Icons, OnLoadMore, TreeViewContext} from './context';
+import {Icons, MoveItemIndex, OnLoadMore, TreeViewContext} from './context';
 import {ITreeProps, useTree} from './useTree';
 
 interface ITreeViewProps<T>
@@ -69,7 +69,7 @@ interface ITreeViewProps<T>
 	/**
 	 * Callback is called when an item is about to be moved elsewhere in the tree.
 	 */
-	onItemMove?: (item: T, parentItem: T) => boolean;
+	onItemMove?: (item: T, parentItem: T, index: MoveItemIndex) => boolean;
 
 	/**
 	 * When a tree is very large, loading items (nodes) asynchronously is preferred to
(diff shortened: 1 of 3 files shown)


sandrodw3:
Hey @matuzalemsteles! Sorry for the delay, I've been out of the office last days. Thanks for your work, it looks super good, I'm giving some feedback:
1. The index data in `onItemMove` is just what we needed, thanks! I only would suggest one change. When dropping an item inside another one, I see we are adding it by default in the first position, I think maybe it would make more sense to add it in the last one, what do you think?
2. About using the `accept` property, I think it's a good idea, and I also think it's better to separate the accept to another prop, like you did with `dragAndDropAccept`. This is very useful because we can limit the drag and drop per type before doing it. So we would have:
   * `dragAndDropAccept`: limit drag and drop per item type before doing it.
   * `draggable`: limit drag and drop in a specific item before doing it, depending on whatever we want related to that item.
   * `onItemHover`: limit drag and drop during the drag depending on whatever we want related to the dragging item and the target item.
   
   I think with this we would cover all cases.
3. The implementation of `onItemHover` sounds good! As I understand, the callback is called each time we hover an item during drag and if we return `false` in that callback, the drag will be cancelled on that item, right? That's enough for what we need.

Please let me know what you think 😄

matuzalemsteles:
@sandrodw3 in theory `dragAndDropAccept` it should remove the need for `onItemHover`, not sure it will cover all cases but in theory it would work like `onItemHover`.

>  I think maybe it would make more sense to add it in the last one, what do you think?

I'm not sure what the best direction would be here, actually I don't have a very strong opinion on this, maybe this on an asynchronous load shouldn't work too well?! we also don't have something defined by the Lexicon but I would like their opinion here, maybe it will help us to better understand which direction to go with this @drakonux @emiliano-cicero

sandrodw3:
Hey @matuzalemsteles, about `dragAndDropAccept` and `onItemHover`, I think they have different purposes. For example, we have same cases in which a certain type of item is droppable only when it has certain configuration. We could cover that case with `onItemHover` checking that condition in the callback and returning false when we want, but we wouldn't with `dragAndDropAccept` because there we can only specify de type, and not any other condition.

Anyway, maybe we could cover our case with `onItemMove`, doing the check in that callback, returning false and avoiding the move after the user makes the drop. So if you think it's better to remove `onItemHover` in favor of `dragAndDropAccept`, go with it 😄.

About the second question, you are right, I hadn't thought about async load, maybe it's better to drop it in first position. Let's see what Lexicon thinks.

Thanks!

matuzalemsteles:
@sandrodw3 Hmm I see, it makes sense `onItemHover` will have more freedom. I think we're fine here with the APIs, I'll go ahead and just add the `onItemHover` API, I don't want to add too many APIs to handle this in different ways because it might get more complex, the `onItemHover` here it's more flexible and should support edge cases.

matuzalemsteles:
@sandrodw3 just one thing to clarify about `onItemHover`, it doesn't return a boolean value to decide if the item can be droppable on the item because the item is still moving, so it doesn't make sense to return false to say it can't be dropped on the item, because the user action that wants the item to be moved is only completed when he drops the item and not when the item is still moving, so this API doesn't have the same behavior as `onItemMove`.

The difference between `onItemMove` and `onItemHover` is that it is called when the drop action is completed, whereas `onItemHover` is always called when the user tries to move the item, that is, hovering the mouse over the items will call this method. So the `onItemHover` is something else to help you create another kind of interaction when the user is moving the item and the `onItemMove` is the final action or the user's intention to drop the item somewhere.

Let me know if this is clear, if you have any cases that I'm not seeing could you show the behavior you had before that you want to implement now?

matuzalemsteles:
I'm going ahead and merging this PR to include in today's release so it can be tested in DXP use cases, @sandrodw3 let me know if you have anything else regarding this and we'll keep improving it if needed.

sandrodw3:
Hey @matuzalemsteles thanks for your explanation and for merging this! I had understood `onItemMove` a bit different, but what you say makes sense. Anyway, I think that we will be ok with the point we have reached with the APIs. If I find any tiny thing that can be improved, I'll open a new ticket. Thanks again Matu!

PR #5167, opened 2022-10-28
fix(@clayui/drop-down): fixes Drilldown accessibility issues

matuzalemsteles:
Related #5041

This PR fixes some accessibility issues that exist in Drilldown, some like `aria-hidden="true"` for the active menu and not exposing an API to add aria-label to buttons with unlabeled actions for example.

I also noticed that there are some cases that this component is not covering according to the documentation https://docs.google.com/document/d/1PVQwknqVIOQhf2JL0kEv0k6WHOExKOskiX-CxxWYRqs/edit#, so maybe a review is needed on this component, I should create a separate issue to deal with it.

### DIFF
--- a/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
+++ b/packages/clay-drop-down/src/DropDownWithDrilldown.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {InternalDispatch} from '@clayui/shared';
+import {InternalDispatch, useInteractionFocus, useInternalState} from '@clayui/shared';
 import classNames from 'classnames';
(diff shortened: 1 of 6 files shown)


marcoscv-work:
@drakonux we have here a pending review!

matuzalemsteles:
Just to complement something I did change, this was not clear about it but when you navigate through Drilldown and close the menu, it resets the navigation to the initial state, is that ok or should we keep the old state even when closing the menu?

matuzalemsteles:
Hey @marcoscv-work or @drakonux could you review this? so we can include it in Monday's release.

drakonux:
It's correct. If the user closes the menu, the navigation has to be reset to its initial state.

PR #5174, opened 2022-11-03
feat(@clayui/color-picker): Convert to use 8 digit hex colors and add alpha slider

### LINKED ISSUE #5132: Color Picker: Can we go for hex 8 digits?
Could we jump from rgba() to hex 8 digits?

https://www.w3.org/TR/css-color-4/#hex-notation
https://caniuse.com/?search=%23RRGGBBAA
It seems pretty safe to use

image

image

The patterns is valid and can be written, right now we change the 8 digits value to `rgba()` pattern.
If we try to change it it will be converted to hex 6 digits

image

Some product teams are currently using 8 digits so I think we could use Hex by default and rgba() when we specify by typing

### LINKED ISSUE #5166: @clayui/color-picker: Add alpha transparency slider
Related to https://github.com/liferay/clay/issues/5132

195135184-103c1416-1412-405e-b564-9a83c56c054d

pat270:
fixes #5166 issue #5132

I couldn't get the range input to work nicely with color picker. The alpha value updates when the mouse / key is released. I think if we want the alpha to continuously update with `onChange`, we need to restructure the components. I'm hoping there is a better way though. I included the CSS updates in https://github.com/liferay/clay/pull/5156 in this pr.

I was thinking it might be a good idea to add a setting to color picker to disable 8 digit hex support.

color-picker

### DIFF
--- a/packages/clay-color-picker/src/Alpha.tsx
+++ b/packages/clay-color-picker/src/Alpha.tsx
@@ -1,0 +1,111 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {ClayInput} from '@clayui/form';
+import ClaySlider from '@clayui/slider';
+import React, {useState} from 'react';
+
+type Props = {
+	/**
+	 * Callback function for when the hue value changes
+	 */
+	onChange: (event: any) => void;
+
+	/**
+	 * The inline styles to add on ClaySlider
+	 */
+	style: {
+		color: string;
+	};
+
+	/**
+	 * The value of the alpha transparency
+	 */
+	value: number;
+};
+
+/**
+ * Renders Alpha component
+ */
+const ClayColorPickerAlpha = ({
+	style,
+	value = 0,
+	onChange = () => {},
+}: Props) => {
+	const [alpha, setAlpha] = useState(value);
+
+	const handleOnChangeEnd = (event: any) => {
+		setAlpha(event.target.value);
+
+		onChange(event);
+
+		event.target.blur();
+
+		event.target.focus();
+	};
+
+	React.useEffect(() => {
+		setAlpha(value);
+	}, [value]);
+
+	return (
+		<div className="clay-color-form-group">
+			<ClaySlider
+				className="clay-color-slider clay-color-slider-alpha"
+				max={1}
+				min={0}
+				onChange={setAlpha}
+				onKeyUp={(event) => {
+					const arrowKeys = [
+						'ArrowDown',
+						'ArrowLeft',
+						'ArrowRight',
+						'ArrowUp',
+					];
+
+					if (arrowKeys.indexOf(event.key) > -1) {
+						handleOnChangeEnd(event);
+					}
+				}}
+				onPointerUp={handleOnChangeEnd}
+				showTooltip={false}
+				step={0.01}
+				style={style}
+				value={alpha}
+			/>
+			<ClayInput.Group>
+				<ClayInput.GroupItem>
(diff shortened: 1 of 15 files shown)


packages/clay-color-picker/src/Alpha.tsx
```diff
@@ -0,0 +1,111 @@
...
+	 * Callback function for when the hue value changes
+	 */
+	onChange: (event: any) => void;
+
+	/**
+	 * The inline styles to add on ClaySlider
+	 */
+	style: {
+		color: string;
+	};
```
matuzalemsteles, line 21: In that case since you just want `color` I would say just go with that property instead of having a property that only has the `style`, this is easier to use the component even though it is internal. We don't always need to follow the native attribute pattern to create React components, which makes things simpler, we can simplify some things.

packages/clay-color-picker/src/Alpha.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+import {ClayInput} from '@clayui/form';
+import ClaySlider from '@clayui/slider';
+import React, {useState} from 'react';
+
+type Props = {
+	/**
+	 * Callback function for when the hue value changes
+	 */
+	onChange: (event: any) => void;
```
matuzalemsteles, line 14: The type for React event callback in this case can be: `(event: React.KeyboardEvent<HTMLInputElement>) => void`

packages/clay-color-picker/src/Alpha.tsx
```diff
@@ -0,0 +1,111 @@
...
+	value: number;
+};
+
+/**
+ * Renders Alpha component
+ */
+const ClayColorPickerAlpha = ({
+	style,
+	value = 0,
+	onChange = () => {},
```
matuzalemsteles, line 35: In those cases where `onChange` is required it would not be necessary to add a noop function, although we add a lot in some places in Clay at the beginning I really don't like the idea because it makes it difficult to debug bugs when you have a noop function as a default property.

matuzalemsteles (commented):
> I think if we want the alpha to continuously update with onChange, we need to restructure the components

Yeah, I'm not sure why this is happening, I hope I can delve deeper into it to understand the reasons but the last time I researched this is that using the native range with React doesn't work very well and there are several libraries implementing the slider from scratch probably because of this issue.

It could also be that calling `onChange` here can cause a lot of rendering and lose reference to the element in the DOM which causes the mouse to be lost every time when interact with the slider.

> I was thinking it might be a good idea to add a setting to color picker to disable 8 digit hex support.

I think instead we can make it smarter, because in this implementation we are always converting to hex with 8 digits, I think we should just convert to hex8 when the transparency is less than 1 instead.

I added some initial comments, but I think we can improve a few more things, I'll clone your pr locally and do some more tests to understand some behaviors.

pat270:
@matuzalemsteles I made the changes. Alpha less than 1 will generate hex8 and hex6 for others and the tests _should_ be passing.

matuzalemsteles:
Hey @pat270 I'm taking a look at it now, i'm seeing how we can better handle the states and be able to use `onChange` to change the alpha state and not have problems and also decrease the amount of renderings to keep only one state that has the source of truth.

matuzalemsteles:
@pat270 I tried to make some more changes to try to decrease the number of renders but it seems that React is not working well with the native range, maybe we have to offer a solution as before and add accessibility using ARIA.

Even if we keep this behavior as it is now of just updating at the end, it seems like a bad experience for the user to not be able to see the color change in real time.

I've tried several things to get realtime color change using the native slider but it's still pretty bad because it keeps losing the mouse drag all the time when dragging the slider. I think we'll have to do the same for #5156

pat270:
Putting this on hold, will revisit in the future.

PR #5177, opened 2022-11-04
feat(@clayui/core): add accessibility improvements to VerticalBar

### LINKED ISSUE #5168: VerticalBar: The use of keyboard is not standard / perceptible
**Using keyboard is not standard / perceptible**

_Accessibility improvement_: fix proposed in order to use VerticalBar in DXP.

**Context:**

This VerticalBar component: Visually “Icons > Panels” are just vertical tabs so the behaviour should be like tabs.

¿How are tabs defined in Clay? As a role=“tablist”:
https://clayui.com/docs/components/tabs.html

In aria is pretty similar behaviour:
https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic.html#tablist-1

and the last example of panels use, google (gmail for example right panels):
Following these examples is easy to define the pattern based on clay/aria with some improvements based on google (they invested on accessibility a lot).

image

So, the **detailed description**:

1º We get to the tabs focusing the active tab, if there are not an active tab we go directly to the first item (the rest of them have tabindex=“-1”).
2º If we are in the button-tabs component we can use arrow up/down to move between button-tabs
3º If we are over a button-tab and we press enter/space we show that panel and move the focus to the first focusable item or making a general focus on the panel/sidebar
4º If we get to the componen by focus and we have a panel active we will focus to the button first and next item will be the sidebar/panel
5º We use correct markup and labelling:
https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html#rps_label
6º If we are in the button-tabs component and press tab-key we will go to next element (the sidebar if it is expanded/opened)

matuzalemsteles:
Fixes #5168

This PR adds OOTB accessibility to the VerticalBar component. There are some cases here that maybe we need to discuss if it makes sense, mainly the focus has to go to the tabs and then to the panel, at this moment the focus order is according to the order of declaration of the elements and how it is visible on the screen, for example positioned on the left the order is tab -> focus panel and when rendered on the right because the tab is rendered last and the panel before the tab the order is panel -> tab, according to the order of the document.

Another detail in the issue is that pressing space/enter on the tab, it shows the panel and we need move the focus, I think we can't do this in an OOTB way, so maybe the developer will have to do this, because normally the use cases of DXP is that these panels they are loaded asynchronously.

You can test these changes using the generated Storybook link.


marcoscv-work:
Tested and works perfect! From my side they can be merged. Thank you very much @matuzalemsteles you are implementing these accessibility changes super fast! I am super happy with how everything is working lately even knowing that we are not providing the changes/improvements in the best way and we are generating documentation debts

About the focus and the side and orientation, I think is better if we move the focus directly to the panel like the echo example in DXP, in this way we press and directly go to the section, nd if we need to go back to the verticalBar just press shift + tab. This behaviour could be an API property

marcoscv-work:
> > I think is better if we move the focus directly to the panel like the echo example in DXP, in this way we press and directly go to the section, nd if we need to go back to the verticalBar just press shift + tab. This behaviour could be an API property
> 
> @marcoscv-work I've tried to reproduce this in DXP but I can't do it consistently. If you start the navigation with a tab key, until you go through the entire document it goes first to the bar and then to the panel, the browser also seems to have a behavior of focusing on the element closest to where the last mouse click was (no need to click in a focusable element), apart from that we were also able to do the same because it already seems to be the default behavior, the difference is that in our example we don't have a focusable element in the panel so it always goes to the bar even if you click inside the panel but testing locally with some focusable elements inside the panel the behavior is similar.

It is not implemented in all DXP vertical menu, but yes, looks like is depending of the order of elements, I think we con go ahead and then decide how to redirect that tabulation in inverted sidebars.
About the focus to the panel, it is interesting because some times the panel can have different description than the label (more accurate one). and the reader reads the whole panel.

matuzalemsteles:
> About the focus to the panel, it is interesting because some times the panel can have different description than the label (more accurate one). and the reader reads the whole panel.

Yeah it makes a lot of sense!

marcoscv-work:
I think we can go ahead! thanks @matuzalemsteles

PR #5178, opened 2022-11-04
feat(@clayui/color-picker): Adds splotchProps to pass custom attributes to the input group splotch

### LINKED ISSUE #5155: Title of Heading's color picker is not clear
**Steps to reproduce:**
1. Add a Heading fragment into a content page
2. Navigate to the Styles panel of Heading fragment
3. Hover on the color pick icon of Background Color and Border Color
The title shown is not clear, It only represents numbers and the user may not know what it is.
LPS-166043

## What is your proposal?
My proposal contains two things (choose one):

1. The first one, that the number that appears as a title, we add the symbol "#", because this way the user knows that it is an RGB colour.
2. Secondly, the title could contain the name of the field in question. For example, if the field is "Text", the title should be "Text Color", or if it is "Background", the title should be "Background Color".

This proposal allows us to give clarity to the user, as the numbers represented by the current title do not give us information and can be confusing for the user.

I think that this is the best way we can raise the issue.

pat270:
fixes #5155 

This adds the `#` symbol to hex values in the title. It also adds `splotchProps` so we can pass attributes to the Input Group splotch. This would allow passing custom titles to the Input Group Splotch.

```
<ClayColorPicker
	splotchProps={{
		title: `Background Color: #${color}`,
	}}
	value={color}
/>
```

### DIFF
--- a/packages/clay-color-picker/src/Splotch.tsx
+++ b/packages/clay-color-picker/src/Splotch.tsx
@@ -29,7 +29,7 @@ interface IProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
  * Renders component that displays a color
  */
 const ClayColorPickerSplotch = React.forwardRef<HTMLButtonElement, IProps>(
-	({active, className, size, value, ...otherProps}, ref) => {
+	({active, className, size, title, value, ...otherProps}, ref) => {
 		const color = tinycolor(value);
 
 		const isHex = (color.getFormat() || '').match('hex');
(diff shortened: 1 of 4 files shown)


matuzalemsteles (commented):
It makes sense, I think we can just simplify the API a bit here, have a more straightforward API just to customize the spotch `title` API, I think it doesn't make sense here to expose a `splotchProps` API which is generic, the splotch here is an internal component so people don't need to know that, we also want to avoid more APIs like `*Props` this is a well known problem called prop drilling that we want to solve in these components.

One option here would be to have an API like `splotchTitle?: string;` which just needs to be the title prefix, so it's easier because ColorPicker can be used uncontrolled and doesn't have the visibility of `color` to concatenate in the string, so we did it internally.

matuzalemsteles (approved):
LGTM! Thanks @pat270 for working on it!

PR #5180, opened 2022-11-05
feat(@clayui/nav): add accessibility to the Vertical Nav component

### LINKED ISSUE #5170: Accessibility update for Menubar Vertical Navigation
## What is your proposal?
To improve the accessibility of the Menubar it should be based off of the [W3 Menu Bar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu/).

It should include role="menubar" with aria-orientation="vertical" and menus within that have role="menu" and items with role="menuitem".

This would provide better keyboard functionality, allowing the movement through menubar, menu, and menu items using only arrow keys instead of tab. 

Improved accessibility.

## Needs to be discussed
How should the arrow keys operate moving between top menubar items and submenus.

cc: @marcoscv-work

matuzalemsteles:
Fixes #5170

This PR is a draft to add accessibility to the VerticalNav component, still needs to make adjustments to the interaction with the keyboard is working more or less but I still need to improve the navigation flow of the arrows from top to bottom that is not discarding elements not visible on the screen.

There are also some other interactions using the keyboard, for example for menu with submenus, maybe we can use the same strategy as we do in TreeView:

- pressing the right arrow key on an element with submenu and it has focus, expands the submenu, if pressed again move focus to the first child element
- press the left arrow key on an element with submenu and it has focus, close the menu
- press left arrow key on first submenu item move focus to element parent

### ToDo

- [x] Fix navigation using up and down arrow keys
- [x] Add keyboard interactions to expand/collapse submenu
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/clayui.com/content/docs/components/markup-menubar.md
+++ b/clayui.com/content/docs/components/markup-menubar.md
@@ -1,10 +1,10 @@
 ---
-title: 'Menubar (Vertical Navigation)'
+title: 'Vertical Navigation'
 description: 'An alternative navigation pattern that displays navigation items vertically.'
 order: 5
+lexiconDefinition: 'https://liferay.design/lexicon/core-components/navigation/vertical-nav/'
+mainTabURL: 'docs/components/vertical-nav.html'
 ---
-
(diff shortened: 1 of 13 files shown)


matuzalemsteles:
This is ready for review now, I've added the final interactions for navigating using the keyboard. @marcoscv-work let me know what you think of this interaction before proceeding with it.

marcoscv-work:
We can go ahead with this PR! ;-) !

PR #5193, opened 2022-11-11
fix(@clayui/drop-down): adds keyboard navigation with looped arrow keys

### LINKED ISSUE #5041: Review the accessibility of behavior in DropDown

matuzalemsteles:
Fixes #5041

This PR adds using loop navigation using keyboard arrow keys and disables tabbing in menu, still need to adjust move focus to first menu item which is having some conflicts with FocusScope and useNavigation because we have to disable tab of the Menu items.

### ToDo

- [ ] Fixes tests
- [ ] Move focus to first menu item

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -9,6 +9,7 @@ import {
 	InternalDispatch,
 	Keys,
 	useInternalState,
+	useNavigation,
 } from '@clayui/shared';
 import {hideOthers} from 'aria-hidden';
 import classNames from 'classnames';
@@ -229,8 +230,15 @@ function ClayDropDown<T>({
 		[initialized]
 	);
 
+	const navigationProps = useNavigation({
+		activation: 'manual',
+		containeRef: menuElementRef,
+		loop: true,
+		orientation: 'vertical',
+	});
+
 	return (
-		<FocusScope>
+		<FocusScope arrowKeysUpDown={false}>
 			{(focusManager) => (
 				<ContainerElement
 					{...otherProps}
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
This is ready for review now, I sent more commits which fixes navigation to the contextual menu which allows navigation in submenus just using arrow keys.

matuzalemsteles:
I sent the last remaining interactions to comply with w3c accessibility standards.

PR #5196, opened 2022-11-14
fix(@clayui/date-picker): Map id attribute to .form-control and add i…

### LINKED ISSUE #5195: @clayui/date-picker: The `id` attribute should go on the input
The `id` attribute should map to the `input` element, exactly like `name`. It currently maps to `<div class="input-group">`. This prevents us linking a `<label for="date-picker-input-id">` to the input.

pat270:
…nputGroupId attribute

BREAKING CHANGE: The `id` attribute now maps to `.form-control`, use the `inputGroupId` attribute to add an `id` to `.input-group`

fixes #5195

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -591,12 +591,13 @@ const ClayDatePicker = React.forwardRef<HTMLInputElement, IProps>(
 		return (
 			<FocusScope>
 				<div className="date-picker">
-					<ClayInput.Group id={id} ref={triggerElementRef}>
+					<ClayInput.Group ref={triggerElementRef}>
 						<ClayInput.GroupItem>
 							<InputDate
 								{...otherProps}
 								ariaLabel={ariaLabels.input}
 								disabled={disabled}
+								id={id}
 								inputName={inputName}
 								onChange={inputChange}
 								placeholder={placeholder}


packages/clay-date-picker/src/index.tsx
```diff
@@ -591,12 +597,13 @@ const ClayDatePicker = React.forwardRef<HTMLInputElement, IProps>(
 		return (
 			<FocusScope>
 				<div className="date-picker">
-					<ClayInput.Group id={id} ref={triggerElementRef}>
+					<ClayInput.Group id={inputGroupId} ref={triggerElementRef}>
```
matuzalemsteles, line 600: @pat270 these ids for association I believe we can provide this via OOTB, for example see the `useId()` hook in the VerticalBar component, it will generate a unique incremental id this is better because it is more difficult for DXP developers to deal with the generation of ids and normally they don't add.

pat270: @matuzalemsteles I don't think we need an `id` on `input-group`. I just added `inputGroupId` because we were previously allowed to add an `id` there. I tried `useId()` but it generates it on every change which isn't useful when trying to target it with CSS. I'm going to remove it.

pat270: The use case in portal is:

```html
<label for="idGeneratedByPortal"></label>
<ClayDatePicker id="idGeneratedByPortal" />
```

If we generate for them, they won't be able to pass their custom id.

matuzalemsteles: On second thought I hadn't seen your change, but you're right it doesn't make sense to associate the group with the input just in that case, maybe just leaving the recent change as it is is enough. We can also use `useId` at any time to associate things.

Looking at this component it looks like we have some issues with the menu trigger with the menu, it doesn't correctly associate if the menu is open or not, I think we need to fix that too.

matuzalemsteles (approved):
LGTM!

> Looking at this component it looks like we have some issues with the menu trigger with the menu, it doesn't correctly associate if the menu is open or not, I think we need to fix that too.

I'll leave it to send this in another PR.

Issue #4741, opened 2022-03-18
Explore the possibility of standardizing form-related components


matuzalemsteles:
Well, we want to improve the DX of our form related components, which means we want to have a consistency of APIs, for example `<Select />` doesn't have a `label` API unlike `<ColorPicker />`, that causes inconsistency between Clay components and different ways of using the API.

We also force compositing with `ClayForm` which is good at times but bad at others, this is a big tradeoff of flexibility vs agility, we can go in the direction of trying to keep all form related components consistent with APIs for example, provide `label`, handle OOTB validation in some way that allows it to integrate with form libraries like Formik, required... with the main goal to improve the DX of the components.

The idea here is also to understand how we should do this or we should create something that approximates the composition and improves the DX of the components that deal with the form. The current composition of these use cases means you have to write more and use more components, this can be repeatable in most scenarios and also means more stuff to remember but on the other hand, it means a lot of flexibility, we need to see how far we can provide this flexibility.

The output of this issue is an [RFC](https://github.com/liferay/clay/tree/master/docs#rfcs) that we can describe all the nuances regarding this change and the final proposal for it.

Valid discussions on this:

- #3796

matuzalemsteles:
LPS Ticket https://liferay.atlassian.net/browse/LPS-190556

PR #5198, opened 2022-11-14
feat(@clayui/time-picker): Add `hoursInputId` prop that sets id on `cla…

### LINKED ISSUE #5197: @clayui/time-picker: Allow passing id to clay-time-hours input
We are unable to link `<label for="clayTimePickerHoursInput">` to an input because we can't pass an `id` to `clay-time-hours`.

pat270:
…y-time-hours`

fixes #5197

### DIFF
--- a/packages/clay-time-picker/src/index.tsx
+++ b/packages/clay-time-picker/src/index.tsx
@@ -83,6 +83,11 @@ interface IProps
 	icon?: boolean;
 
 	/**
+	 * Id to be applied to `clay-time-hours` input.
+	 */
+	id?: string;
+
+	/**
 	 * Called when input values change hour, minutes or ampm (controlled).
 	 */
 	onChange?: InternalDispatch<Input>;
@@ -175,6 +180,7 @@ const ClayTimePicker = ({
 		minutes: DEFAULT_VALUE,
 	},
 	icon = false,
+	id,
 	name,
 	onChange,
 	onInputChange,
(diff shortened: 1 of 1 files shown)


packages/clay-time-picker/src/index.tsx
```diff
@@ -77,6 +77,11 @@ interface IProps
 	 */
 	defaultValue?: Input;
 
+	/**
+	 * Id to be applied to `clay-time-hours` input.
+	 */
+	hoursInputId?: string;
```
matuzalemsteles, line 83: I think you can just call `id` and bind the id to the field group instead of just the hour input would be better for the screen reader, we need to add the correct ARIA semantics to the group to create the association correctly and that improves accessibility for that component as well.

Adding `hoursInput*` is that it's a not very intuitive name and we don't want the developer to have to know that the id is going to the hour input just that it's an internal implementation level that the developer doesn't have to worry about it.

pat270: I changed `hoursInputId` to `id`. Clicking on a form label will automatically focus the Time input. The problem with the `for` attribute is that the browser only moves focus on click if it's linked to a proper form element. We have to manually implement a focus change if it maps to a `div`. It's hard to do if the `label` isn't included in the component.

Also it looks like `aria-label` can be customized, we can leave the screen reader announcements to the user.

matuzalemsteles: Oh yes you are right, we would need to add a handler to correctly move the focus so it works properly. Maybe we need to start working on #4741 to redesign the default form related components to provide things like this OOTB.

matuzalemsteles (approved):
LGTM

PR #5202, opened 2022-11-16
feat(@clayui/date-picker): improves semantic accessibility for the component

### LINKED ISSUE #5201: DatePicker component is not accessible
DatePicker component is not accessible following w3c recommendations https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/datepicker-dialog.

Both the ARIA semantic part for the elements and the standard keyboard navigation part in the grid do not exist in the component.

matuzalemsteles:
Fixes #5201

This PR improves DatePicker's semantic accessibility but doesn't add the keyboard navigation recommendations yet, I'll add this in another PR once we have a clearer picture of what it should look like.

This PR adds semantic accessibility for the date grid and `aria-hidden` for elements that should not be assertive by SR when the menu is open, the same as we do for the other menus.

### DIFF
--- a/packages/clay-date-picker/src/DayNumber.tsx
+++ b/packages/clay-date-picker/src/DayNumber.tsx
@@ -27,35 +27,24 @@ const ClayDatePickerDayNumber = ({
 	const {date, nextMonth, previousMonth} = day;
 	const [startDate, endDate] = daysSelected;
 
+	const isStartAndEndDateRange =
+		startDate.toDateString() !== endDate.toDateString() &&
+		isWithinInterval(date, daysSelected);
 	const hasEndDateSelected = date.toDateString() === endDate.toDateString();
 	const hasStartDateSelected =
 		date.toDateString() === startDate.toDateString();
 
-	const classNames = classnames(
-		'date-picker-date date-picker-calendar-item',
-		{
-			active: hasStartDateSelected || (range && hasEndDateSelected),
-			disabled,
-			'next-month-date': nextMonth,
-			'previous-month-date': previousMonth,
-		}
-	);
-
 	return (
 		<div
 			aria-selected={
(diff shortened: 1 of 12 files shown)


packages/clay-date-picker/src/DayNumber.tsx
```diff
@@ -43,6 +43,13 @@ const ClayDatePickerDayNumber = ({
 
 	return (
 		<div
+			aria-selected={
+				(startDate.toDateString() !== endDate.toDateString() &&
```
ethib137, line 47: This check is duplicated at 57... what do you think about adding a const for this boolean like you have for `hasStartDateSelected`?

packages/clay-date-picker/src/index.tsx
```diff
@@ -631,14 +649,12 @@ const ClayDatePicker = React.forwardRef<HTMLInputElement, IProps>(
 							alignElementRef={triggerElementRef}
 							className="date-picker-dropdown-menu"
 							data-testid="dropdown"
+							id={ariaControls}
 							onActiveChange={setExpandedValue}
 							ref={dropdownContainerRef}
+							role="dialog"
```
ethib137, line 655: Might be worthwhile to add an 'aria-label' to the dialog element, since then the screen reader will announce it when the dialog is focused. For example: "Choose Date Modal Dialog" will be read when the aria-label="Choose Date".

ethib137 (commented):
Looks good @matuzalemsteles, just two suggestions for you to consider. :)

packages/clay-date-picker/src/DayNumber.tsx
matuzalemsteles: Oh yes, that would be better.

Issue #5203, opened 2022-11-18
Review Autocomplete navigation behavior


matuzalemsteles:
We recently made some keyboard navigation behavior changes in DropDown which is now following the recommendations of w3c #5193, and autocomplete also had accessibility improvements along with a new composition #5108 the difference is that this improvement came in before we decided on the behavior keyboard navigation for DropDown which means that now we also need to change the keyboard navigation behavior of Autocomplete to just arrow keys and not allow tabs inside the Menu.

Now with the entry of #5193 and the behavior decision we can go with it and apply the keyboard navigation that was missing for Autocomplete following the [w3c recommendations](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html) with virtual focus instead of moving focus.

We also need this to be able to apply the combobox pattern in MultiSelect (#5194), since MultiSelect uses Autocomplete but using its old version, so it affects MultiSelect too, we will need to apply the collection pattern, this will be done when using the new Autocomplete behavior in MultiSelect.

The main changes we should make to the component with this issue are:

- Only allow keyboard navigation using arrow keys
- Navigation using virtual focus
- Add loop navigation
- Adding shortcuts like Home, End, pressing arrow down move the virtual focus to the last element, this should be compatible with virtualization, so we will need to implement more functionality in the `useNavigation` hook that is dealing with this.

PR #5204, opened 2022-11-18
feat(@clayui/multi-select): adds accessibility improvements to MultiSelect

### LINKED ISSUE #5194: Clay Multi Select component is not accessible
https://clayui.com/docs/components/multi-select.html

The current Multi Select component is difficult to use and understand with a screen reader or keyboard navigation. Based on research with @marcoscv-work it appears that the best pattern to follow would be the following:
https://www.w3.org/WAI/ARIA/apg/example-index/grid/LayoutGrids.html#ex2_label

This pattern provides easy keyboard navigation as well as context for what the component does and what part of the component is currently being interacted with.

We also considered applying the [Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) but this did not properly handle the pills aspect of the multi select. The Combobox Pattern should likely still be applied for the case where the Multi Select includes a Select Button for choosing items.

matuzalemsteles:
Fixes #5194

This PR is just a draft of the accessibility improvements for the MultiSelect component and I will just add the w3c grid layout recommendations here, the combobox pattern will be done by issue #5203 which depends on the revision in Autocomplete.

There are still some things missing here to be ready, also a discussion is about the input if it should be part of the `grid` of pills, this in the w3c implementation, it is outside the grid but our markup needs it to be inside the grid to not visually breaking maybe i need help from @pat270 here to try to solve this. I'm also not sure if this can work very well together the combobox pattern and pills, in Lexicon the navigation behaviors allow them to be very integrated with each other, for example pressing backspace when it's empty input twice it deletes the pill, when deleting the pill the focus moves to the input, these behaviors are quite different according to the w3c use case of https://www.w3.org/WAI/ARIA/apg/example-index/grid/LayoutGrids.html#ex2_label.

I was taking a look at the Gmail use case which is very similar to ours when you type the emails in the box to whom you want to send, they use the default combobox and the input and the pills are a `listbox` with multi selection, where the pill is a selected `option`, it seems that this semantically and also behaviorally might work better with the interactions between the input and the pills. cc @ethib137 @marcoscv-work 

If we continue with the behaviors of w3c, here is what still needs to be done to move forward with this PR:

- [x] Add the tab to the pill that had the last focus
- [x] Navigate between pills using arrow keys
- [x] Add keyboard shortcuts for navigation between pills

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -134,10 +134,10 @@ module.exports = {
 			statements: 85,
 		},
 		'./packages/clay-multi-select/src/': {
-			branches: 85,
+			branches: 64,
 			functions: 80,
-			lines: 87,
-			statements: 86,
+			lines: 78,
+			statements: 78,
 		},
 		'./packages/clay-multi-step-nav/src/': {
 			branches: 94,
--- a/jest.config.js
+++ b/jest.config.js
@@ -134,7 +134,7 @@ module.exports = {
 			statements: 85,
 		},
 		'./packages/clay-multi-select/src/': {
-			branches: 64,
+			branches: 62,
 			functions: 80,
 			lines: 78,
 			statements: 78,
--- a/packages/clay-css/src/scss/cadmin/variables/_utilities.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_utilities.scss
@@ -367,8 +367,8 @@ $cadmin-border-theme-colors: map-deep-merge(
 
 // Display
 
-$cadmin-displays: none, inline, inline-block, grid, block, table, table-row,
-	table-cell, flex, inline-flex !default;
+$cadmin-displays: none, inline, inline-block, grid, block, contents, table,
+	table-row, table-cell, flex, inline-flex !default;
 
 // Overflow
 
--- a/packages/clay-css/src/scss/variables/_utilities.scss
+++ b/packages/clay-css/src/scss/variables/_utilities.scss
@@ -357,8 +357,8 @@ $border-theme-colors: map-deep-merge(
 
 // Display
 
-$displays: none, inline, inline-block, block, grid, table, table-row, table-cell,
-	flex, inline-flex !default;
+$displays: none, inline, inline-block, block, grid, contents, table, table-row,
+	table-cell, flex, inline-flex !default;
 
 // Overflow
 
--- a/packages/clay-drop-down/src/Menu.tsx
+++ b/packages/clay-drop-down/src/Menu.tsx
@@ -170,8 +170,14 @@ const Menu = React.forwardRef<HTMLDivElement, IProps>(
 	) => {
 		const setActive = onActiveChange ?? onSetActive;
 
-		const menuRef = useRef<HTMLDivElement | null>(null);
+		const menuInternalRef = useRef<HTMLDivElement | null>(null);
 		const subPortalRef = useRef<HTMLDivElement | null>(null);
+
+		let menuRef = menuInternalRef;
+
+		if (ref) {
+			menuRef = ref as React.MutableRefObject<HTMLDivElement | null>;
+		}
 
 		useOverlayPosition(
 			{
@@ -210,16 +216,7 @@ const Menu = React.forwardRef<HTMLDivElement, IProps>(
 							[`dropdown-menu-width-${width}`]: width,
 							show: active,
 						})}
-						ref={(node) => {
-							menuRef.current = node;
-
-							if (ref instanceof Function) {
-								ref(node);
-							} else if (ref instanceof Object) {
-								// @ts-ignore
-								ref.current = node;
-							}
-						}}
+						ref={menuRef}
 						role={role}
 					>
 						{children}
--- a/packages/clay-multi-select/package.json
+++ b/packages/clay-multi-select/package.json
@@ -30,6 +30,7 @@
 		"@clayui/button": "^3.82.0",
 		"@clayui/drop-down": "^3.82.0",
 		"@clayui/form": "^3.82.0",
+		"@clayui/icon": "^3.56.0",
 		"@clayui/label": "^3.56.0",
 		"@clayui/loading-indicator": "^3.60.0",
 		"@clayui/shared": "^3.82.0",
--- a/packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
@@ -19,6 +19,14 @@ exports[`ClayMultiSelect renders 1`] = `
           value=""
         />
       </div>
+      <div
+        class="sr-only"
+      >
+        <span
+          aria-live="polite"
+          aria-relevant="text"
+        />
+      </div>
     </div>
   </div>
 </body>
@@ -150,6 +158,14 @@ exports[`ClayMultiSelect renders a custom menu 1`] = `
           class="form-control-inset"
           type="text"
           value="one"
+        />
+      </div>
+      <div
+        class="sr-only"
+      >
+        <span
+          aria-live="polite"
+          aria-relevant="text"
         />
       </div>
       <div
@@ -339,6 +355,14 @@ exports[`ClayMultiSelect renders as disabled 1`] = `
           value=""
         />
       </div>
+      <div
+        class="sr-only"
+      >
+        <span
+          aria-live="polite"
+          aria-relevant="text"
+        />
+      </div>
     </div>
   </div>
 </body>
@@ -470,6 +494,14 @@ exports[`ClayMultiSelect renders as not valid 1`] = `
           class="form-control-inset"
           type="text"
           value=""
+        />
+      </div>
+      <div
+        class="sr-only"
+      >
+        <span
+          aria-live="polite"
+          aria-relevant="text"
         />
       </div>
       <div
@@ -622,6 +654,14 @@ exports[`ClayMultiSelect renders with items 1`] = `
           class="form-control-inset"
           type="text"
           value=""
+        />
+      </div>
+      <div
+        class="sr-only"
+      >
+        <span
+          aria-live="polite"
+          aria-relevant="text"
         />
       </div>
       <div
--- a/packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-multi-select/src/__tests__/__snapshots__/index.tsx.snap
@@ -7,18 +7,26 @@ exports[`ClayMultiSelect renders 1`] = `
       class="form-control form-control-tag-group input-group"
     >
       <div
-        aria-describedby="clay-id-2"
-        class="input-group-item input-group-item-shrink input-group-prepend"
-        role="grid"
-      />
-      <div
-        class="input-group-item input-group-prepend"
+        class="input-group-item"
       >
-        <input
-          class="form-control-inset"
-          type="text"
-          value=""
-        />
+        <div
+          class="input-group"
+        >
+          <div
+            aria-describedby="clay-id-2"
+            class="input-group-item d-contents input-group-item-shrink input-group-prepend"
+            role="grid"
+          />
+          <div
+            class="input-group-item input-group-prepend"
+          >
+            <input
+              class="form-control-inset"
+              type="text"
+              value=""
+            />
+          </div>
+        </div>
       </div>
       <div
         class="sr-only"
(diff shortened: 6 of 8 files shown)


ethib137:
Both gmail and w3 are great examples. I'm still leaning towards the w3 one for the following reasons:

- In Gmail focus is moved to the input and there is no acknowledgement of any emails currently being included in the "To" field. 
- The way gmail handles deleting a chip by pressing delete, requires an aria-label="Press delete to remove this chip." which we wouldn't be able to translate and shouldn't require that type of label to passed as a param.
- Additionally gmail's use of chip with an interactive delete seems to actually break the listbox pattern:
    -  "In addition, the interaction model conveyed by the listbox role to assistive technologies does not support interacting with elements inside of an option. Because of these traits of the listbox widget, it does not provide an accessible way to present a list of interactive elements, such as links, buttons, or checkboxes. To present a list of interactive elements, see the [grid](https://www.w3.org/WAI/ARIA/apg/patterns/grid/) pattern."
    
 > it is outside the grid but our markup needs it to be inside the grid to not visually breaking
It should be fine to have the grid, the input, and the clear button as siblings inside a single form-control container styled to look like an input. Right @pat270?

> in Lexicon the navigation behaviors allow them to be very integrated with each other, for example pressing backspace when it's empty input twice it deletes the pill
I don't think these behaviors would conflict with the w3 implementation, they would just be extra. One way this could work, is that when the input is focused, empty, and you press backspace, we would then focus the delete button of the last chip. This would then announce: "Delete Button. Press enter to activate". So enter would delete, but we could also allow for the "Delete" key to delete. This wouldn't contradict with w3 since we already have one default mechanism for the action.

> when deleting the pill the focus moves to the input
I would argue this is the wrong pattern. Notice when we add a pill we keep the focus in the input, rather than move it to the  pill that has just been added. It's helpful to keep the focus where a user can keep doing the same action. IE: adding additional pills, or deleting additional pills. If the want to delete 10 pills it would require hitting "delete" 20 times instead of just 10 times. If this is the pattern set by Lexicon I guess we have to convince them first, though. @marcoscv-work what do you think?

Are there any pieces from Gmail's implementation that you think would still be preferable?

matuzalemsteles:
Thanks @ethib137 for bringing some more insights into the behaviors, well I really don't know but Gmail's visual style is pretty close to our implementation, the w3c example is a bit far from our current use case, the concern I have here is that we have an accessibility issue for SRs and another the component usage experience, so they are different things but they are linked, adopting the case of w3c I think we will have to change some interaction behaviors what is not clear to me now is how it is the interaction between labels and input. For example:

- Should we remove the interaction of pressing backspace moves the focus to the label and pressing again removes? maybe this doesn't break the SR reading, because it can be more of a user experience interaction than something accessible
- Should deleting a label move focus to input? if following w3c the navigation between labels and after deleting it always stays around the labels it never goes to the input, this may seem strange to non SRs users since visually the label belongs to the input, which is different from the case from w3c

I think my question is more how are we going to proceed with these behaviors.

matuzalemsteles:
I added keyboard interactions that are recommended by w3c to interact with labels, let me know what you think.

ethib137:
Hey @matuzalemsteles, I just tested the keyboard accessibility and it feels really nice. Nice Work!

One thing I noticed while testing is that the grid element is not announcing a label when focused. It looks we need a way to pass an aria-label or aria-labelled by prop to the grid element.

matuzalemsteles:
> One thing I noticed while testing is that the grid element is not announcing a label when focused. It looks we need a way to pass an aria-label or aria-labelled by prop to the grid element.

Hmmm yeah, it seems strange that it announces `one Remove one`, I think because the focus is on the label instead of just the text and it will announce the text and the aria-label of the close button. Yes I think `aria-labelled` can help solve this, I'll do some testing.

matuzalemsteles:
I resolved the conflict I was giving and added `aria-labelledby` to associate the text with the label this works better now for readability the SR when focusing on the label, this would also have the same effect when using `aria-label`.

ethib137:
Oh nice.. that's helpful. I was actually referring to the element that has role="grid". Notice that the [grid example](https://www.w3.org/WAI/ARIA/apg/example-index/grid/LayoutGrids.html#ex2_label) has an aria-labelledby attribute on that element which causes the grid to be announced with a name when it is first focused. 

Additionally I wonder if we should place the tabindex="0" on the grid-cell element that contains the label instead of on the grid row like this:
```
<span aria-labelledby="clay-id-1-label-Apple-text" id="clay-id-1-label-Apple-span" role="row" class="label label-secondary">
    <span id="clay-id-1-label-Apple-text" role="gridcell" class="label-item label-item-expand" tabindex="0">Apple</span>
    <span role="gridcell" class="label-item label-item-after">
        <button aria-label="Remove Apple" class="close" id="clay-id-1-label-Apple-close" tabindex="-1" type="button">
            <svg class="lexicon-icon lexicon-icon-times-small" role="presentation"><use xlink:href="static/media/packages/clay-css/lib/images/icons/icons.svg#times-small"></use></svg>
        </button>
    </span>
</span>
```

This would also be following the example, which seems to indicate that we should always focus on the gridcell or an interactive element inside of the cell if it exists: [source](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#gridNav). The only downside of this is that it would only show focus on the label part of the pill... this might actually be for the best since it highlights the fact that there are two cells in each row of the grid. What do you think?

The other thing would be adding roving tab index. Would you be able to include that in this pr?

marcoscv-work:
Thanks for the component update, looks great!

Some confirmations points:

- Ok with keeping the focus on removing previous item, it makes the task of removing faster
- We only need a focus in the label to perform the only one action you can do there. We would like whole item with focus but we can accept is easy to understand the action with the focus on the remove icon
image
image

Visually:
I think gris is breaking lines, the input and the clear all button:
image

Previous one:
image

Interaction:
- Focus should follow logical tab order, for example, if we get out of the grid and then we get in, we should select first label, if we remove a label we should go to the directly previous one, not jump items to go to the first one
- Focus is doing some weird thinks sometimes, in somepoint it are not allowed to get into the grid, sometimes the order jumps items, etc, probably menor issue.

Last accessibility recommendations:
- Add instructions in the grid about how removing items by delete key (not something you expect), example, an `aria-description` with: "Use arrows key to move along labels and delete to remove them", if the text is shown previously the component we could use `aria-describedby`.
- Add feedback to Screen readers about what we have removed what have added, for example: "Label Matu added to the list", "Label Matu removed from the list"

ethib137:
@marcoscv-work 
> if we get out of the grid and then we get in, we should select first label
Is there a reason you would recommend this approach over following the [roving tab index pattern](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex)?

marcoscv-work:
last extra thing (sorry!). About add instructions in the grid about how removing items by delete key, I've seen [this example of aria grid](https://www.w3.org/WAI/ARIA/apg/example-index/grid/LayoutGrids.html) where we have a detailed instructions to use and then the grid uses `aria-describedby` to relate elements

image

Great example!

matuzalemsteles:
Well, a lot of stuff here, let me try to organize the points here a little to know how to apply the changes to the multi select.

> When focusing on the grid elements it is not advertising with the label name

Oh yes, I will fix. One thing to think about this is that we will probably have the two announcements of the label at different times, maybe this will be confusing for the user, one when focusing on the grid and another when focusing on the input, because in our case the label is a reference to the component as a whole or maybe that's a good thing?!

> Focus on text gridcell instead of label row

The problem with focusing on the text `gridcell` is that our markup is different from the example implementation, focusing on the text will not create the focus as it is on the label and visually we will have to change the visual state of the label to support this. I don't think this should affect the SR, we would have to test if it would be a problem to focus on the `row` but the SR announcement is keeping the same behavior so it seems ok.

> The focus inside the grid

I implemented the focus based on the example, it leaves the focus on the last element that had focus on the grid if none had focus yet the first label will have the focus. It also follows the behavior that when deleting it always moves the focus to the next label button.

I think this scenario about adding the wrong label and going back to delete is still possible using the backspace hotkeys.

> in somepoint it are not allowed to get into the grid

Maybe it's the same problem that Evan reported.

> sometimes the order jumps items

@marcoscv-work Could you provide some use cases for me to be able to reproduce this?

### ToDo

- [ ] Fix grid and input alignment when there are too many labels
- [ ] Add `aria-labelledby` to the grid
- [ ] Add description of hotkeys for grid
- [ ] Add feedback to Screen readers about what we have removed what have added
- [ ] Fix loss of label focus when deleting a label https://github.com/liferay/clay/pull/5204#issuecomment-1324676741

marcoscv-work:
I have some random issues I cannot reproduce consistently or provide steps, but I can do it with one! probably a combination of configurations or something with storybook:

**Description**: "Sometimes" I cannot write in input after use clear all button, sometimes even not using it, just get out and get in, and key `D` gets you outside the component

- Turn on VoiceOver
- Write some labels (optional)
- Go to clear all and pulse enter
- Go back to input and try to write using letter `D`. Sometimes it happens sometimes not

The video starts with the input because the input do not interrupt the use of keyboard or the use of letter `D`. the issue can be seen in 00:00:21.

https://user-images.githubusercontent.com/7963804/203747836-fe05ba1a-8c5c-4031-8b23-227beaee74c6.mp4

`D` in storybook is Change add-ons orientation, but I think is not directly related:
storybook

matuzalemsteles:
Thanks @marcoscv-work, I will check it out. I also managed to reproduce it once but I can't anymore 🤷‍♂️, I'm looking at the code but I can't find it related to pressing `D`... pretty weird 😅

marcoscv-work:
> Thanks @marcoscv-work, I will check it out. I also managed to reproduce it once but I can't anymore 🤷‍♂️, I'm looking at the code but I can't find it related to pressing `D`... pretty weird 😅

Pretty weird.. probably the D is coming from some voiceOver key-lapping or something weird like that, we just need to check if we can find out the trigger of the bug.

Thanks Matu

matuzalemsteles:
> Pretty weird.. probably the D is coming from some voiceOver key-lapping or something weird like that, we just need to check if we can find out the trigger of the bug.

Hmm yeah, I'll keep checking that too.

Hey @pat270 I need your help here to deal with the alignment of the input with the labels and buttons, because we need the input to be off the grid, when there are too many labels it is breaking, I tried to adjust some things but I think we will have to create some custom class to handle it. Do you have any idea if we can achieve the same behavior with the classes we have?

pat270:
I'm unable to make it align correctly with the form element outside the `role="grid"`.  I think an alternative solution is to wrap it in:

```html
<span role="row">
  <span aria-colspan="3" role="gridcell">
    <input />
  </span>
</span>
```

This is what Voice Over is reading to me:

voiceover

The `aria-colspan="3"` doesn't seem to register. It could be that my OS is too old and Voice Over is bugged.

matuzalemsteles:
@pat270 well doing some tests the screen reader it just doesn't announce the exit from the grid when moving the focus to the input only when the focus leaves the input, we also have to check if this doesn't fail in some accessibility analysis..., I locally ran axe at least no problem appeared. 

@marcoscv-work what do you think?

pat270:
@matuzalemsteles my suggestion above includes the input as part of the grid. It should announce end of table when you exit the input.

I was thinking could we use something like `list` and `listitem`? I think it fits better since we don't always have a set number of columns in the label. We can have mixed labels without an icon and it doesn't make sense when the screen reader announces row 2 column 1 of 3 when there are only 2 columns.

grid-issues

pat270:
@matuzalemsteles I pushed up a commit with the changes I'm proposing. Feel free to remove the commit if it's no good.

matuzalemsteles:
Hey @pat270 I commented on using `list` as well, I think this would suit our use case better, @ethib137 added some comments about this which are also valid https://github.com/liferay/clay/pull/5204#issuecomment-1319894139

pat270:
[listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) is different from [list](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role), although the way list is implemented here isn't 100% correct either. We could add `aria-hidden="true"` to the thumbnail and close button. It would function more like a list and add a message that says press delete to remove this label.

matuzalemsteles:
> [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) is different from [list](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role), although the way list is implemented here isn't 100% correct either. We could add aria-hidden="true" to the thumbnail and close button. It would function more like a list and add a message that says press delete to remove this label.

I see, I don't have a strong opinion on that... but I'll leave it to @marcoscv-work and @ethib137 to decide. Well, to implement list we would probably have to rewrite all the navigation mechanism that was made for `grid`, the way it navigates is quite different between columns.

It is also worth noting that the input will be a `combobox`, this is depending on completing the improvements to Autocomplete in #5205, so this component is a combination of `grid` and `combobox`.

marcoscv-work:
I would go for keeping the grid and follow the ARIA example.
And looks like grid works better to stop screen reader arrows navigation.

ethib137:
I'm also still think grid is a better option. Could you also check having focus be placed only on gridcell rather than the row. In JAWS at least this really changes how the screen reader announces it. For instances in the example JAWS properly announces the row and column, but with ours, likely because of focusing the row, JAWS announces the delete button as column 1. For testing JAWS announces the row and column when focus moves into the grid, so you can tab out and tab back in to hear the row and column.

matuzalemsteles:
Thanks @ethib137 and @marcoscv-work, I think the problem here that we are facing is the alignment of the input with the labels, as reported by @marcoscv-work, we cannot maintain the correct alignment if the input is sibling the grid, the way we have to have the correct alignment is to leave the input as a child of the grid, the tests we did seem to present no problems but I would like your opinion on this. What do you think?

ethib137:
I do think this is problematic to place the input inside the grid. If the input is in the grid the default keyboard navigation behavior would require that you would need to use arrow keys to navigate to the input as the last item of the grid. This would be annoying when you have many items. You could break this behavior and allow tab to reach the input, but then you break how a grid is normally navigated, which is also not a good idea. 

Do you all have any other ideas for how we could solve this?

matuzalemsteles:
> I do think this is problematic to place the input inside the grid. If the input is in the grid the default keyboard navigation behavior would require that you would need to use arrow keys to navigate to the input as the last item of the grid. This would be annoying when you have many items. You could break this behavior and allow tab to reach the input, but then you break how a grid is normally navigated, which is also not a good idea.

@ethib137 actually the behavior still remains, the arrow keys only become navigable inside the grid of cells, it ignores the input, so the navigation part via keyboard will be the same behavior as the input outside, which can be a problem, maybe be the semantics. I even did some tests with the input inside the grid the problem with the screen reader is just that when moving the focus from the labels to the input it will not announce the exit from the grid only when leaving the input the exit will be announced, maybe that might be confusing for the user, I'm not sure.

Maybe another way to correct the alignment here would be to restructure the input group markups for the multi select, we would have to build a specific one for this component, @pat270 thoughts?

matuzalemsteles:
Screen Shot 2022-12-12 at 2 34 29 PM

Well, I think that's the best we could do with the input outside of the grid, I'm not sure there's another strategy, but unfortunately we couldn't go further than that with CSS.

marcoscv-work:
I think we can go ahead!

marcoscv-work:
Is not working fine if we set display inline CSS for the "table" elements?

matuzalemsteles:
> Is not working fine if we set display inline CSS for the "table" elements?

Well I think table, flex has the same behavior in terms of setting the width.

Screen Shot 2022-12-13 at 11 50 22 AM

Apparently I was able to do this with `display: contents;`, I think this is going to work the way we want it to.

Screen Shot 2022-12-13 at 11 53 00 AM

The difference is that we will have to change the markup here for the "Clear All"  button alignment to work correctly. The way I did it was:

```html
<div className="form-control form-control-tag-group input-group">
  <ClayInput.GroupItem>
    <ClayInput.Group>
      <ClayInput.GroupItem preped shrink role="grid">
        <!-- Label components -->
      </ClayInput.GroupItem>

      <ClayInput.GroupItem prepend>
        <!-- Input component -->
      </ClayInput.GroupItem>
    </ClayInput.Group>
  </ClayInput.GroupItem>


  <ClayInput.GroupItem shrink>
    <!-- Loading Indicator -->
  </ClayInput.GroupItem>

  <ClayInput.GroupItem shrink>
    <!-- Clear All Button -->
  </ClayInput.GroupItem>
</div>
```

The `contents` needs to be on the grid element for it to work properly, I think do not we have a CSS class for that. @pat270, any ideas? on how can we add this in the input, maybe a css class is enough?

matuzalemsteles:
I sent the commit to add `d-contents`, now the alignment is working correctly with the input outside of the grid. The markup part is ok now but we still need to improve the accessibility of the input to be a combobox, as I mentioned this will be done in another PR, the next steps now are:

- #5205
- Add new Autocomplete to MultiSelect

There are some last focus edge cases on the label which are a bit broken due to the conflict with the FocusScope but this will be fixed when we add Autocomplete to the MultiSelect we will no longer need the FocusScope in that part of the MultiSelect.

ethib137:
Great! This looks like a nice way to handle the wrapping issue.

I'm still concerned with the focus being placed on the grid row instead of on the gridcell. So for instance I would propose that `tabindex="-1"` should NOT be on the row like this:
```
<span role="row" tabindex="-1" class="label label-secondary">
    <span role="gridcell" class="label-item label-item-expand">one</span>
    <span role="gridcell" class="label-item label-item-after"></span>
</span>
```
But rather on the `gridcell` like this:
```
<span role="row" class="label label-secondary">
    <span role="gridcell" tabindex="-1" class="label-item label-item-expand">one</span>
    <span role="gridcell" class="label-item label-item-after"></span>
</span>
```

With focus being place on the row, JAWS has some issues with properly announcing which row and column you are in. Is there any reason why we shouldn't make this change?

matuzalemsteles:
> With focus being place on the row, JAWS has some issues with properly announcing which row and column you are in. Is there any reason why we shouldn't make this change?

I think JAWS and VoiceOver have the same problem when entering the gridcell directly it will announce that it is inside a gridcell and gives the descriptions to move using the arrow keys, in this case when being in the row, it still announces that it is inside a table with a row with two columns and the description it announces is the one we added in the description of the grid.

In the end what changes seems to be more the description of the instructions, our description provides that you can use backspace to delete the label, if the focus goes to the gridcell it doesn't read our description of the grid, maybe we would have to specifically add the `aria-describedby` for each gridcell.

The problem with moving to the gridcell is the focus not working properly, for example:

In the row:
Screen Shot 2022-12-15 at 10 48 16 AM

In the gridcell:
Screen Shot 2022-12-15 at 10 49 36 AM

ethib137:
Okay that makes sense... If we use grid, I think we still need to overcome these drawbacks since grid doesn't make sense if we don't place focus on the gridcells. Notice in [Mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role#keyboard_interactions) that it always references focus moving to another cell.

For the instructions on keyboard interactions, it's better if we can let the screen reader announce the the default keyboard interactions for grids... this will be more consistent than always manually including the keyboard interactions when they are following the standard. It should work to use `aria-describedby` like you mentioned for the extra keyboard interaction related to `backspace`. Could be something like: "Press backspace to delete the current row."

Regarding focus, I don't see it as a problem that only the text is focused, followed by the delete button. In order for this to act as a grid, I think this is how it has to be. If we are not okay with only focusing the text, we should consider not using the grid. @marcoscv-work do you see any problem with only focusing the text instead of focusing the entire pill?

matuzalemsteles:
@ethib137 makes sense, I think it seems more correct to also focus on the gridcell of the label's text. I think we would just have to fix the focus visually.

Regarding the description, I also agree that it is better to let the screen reader announce itself according to its default, mainly that the instructions may vary from the operating system.

matuzalemsteles:
I went ahead and moved the focus to stay in the gridcell, now the instructions when moving the focus to the grid are announced according to the screen reader, I also changed the description of the shortcuts to stay in the gridcell because they are not announced being in the grid.

I didn't modify the CSS to fix the focus on the text visually, I had another thought about it that we can get the visual focus on the label but the accessible focus being on the gridcell, internally we can control the `focus` class to add the visual focus on the label when the accessible focus is on the gridcell, I think this we can send in another PR.

I think we can merge this now and ship other improvements in other PRs like the visual focus of the label and continue to improve the accessibility part of this component. I'm going to go ahead and merge that PR so we can have that now in this release.

PR #5205, opened 2022-11-18
feat(@clayui/autocomplete): adds keyboard navigation improvements to Autocomplete

### LINKED ISSUE #5203: Review Autocomplete navigation behavior

matuzalemsteles:
Closes #5203

Well, this PR is still a WIP and requires even more work, mainly because the looped navigation and some shortcuts don't work well when Autocomplete is using virtualization, in case when the content is dynamic.

I still need to implement more features to be able to work keyboard navigation with a virtualized list.

- [ ] Add support for looped browsing with virtualized list
- [ ] Shortcuts like Home and End with virtualized list
- [ ] fix the tests

### DIFF
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -13,6 +13,7 @@ import {
 	useDebounce,
 	useId,
 	useInternalState,
+	useNavigation,
 } from '@clayui/shared';
 import {hideOthers} from 'aria-hidden';
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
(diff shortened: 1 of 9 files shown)


matuzalemsteles:
I'm getting back to working on it again.

matuzalemsteles:
I'm closing this PR because I'll be sending another PR with other approaches.

PR #5207, opened 2022-11-22
feat(@clayui/css): Modularize most Clay CSS components

### LINKED ISSUE #5206: @clayui/css: Modularize components
We should make Clay CSS components more modular by reducing compile errors due to undefined variables. It would work like:

```
@import 'functions/_global-functions';

@import 'atlas/variables/_buttons.scss';
@import 'variables/_buttons.scss';

@import '_mixins';

@import 'components/_buttons.scss';
```

This would only import the CSS for buttons.

pat270:
fixes #5206


matuzalemsteles:
Hey @pat270 looking at this PR I think I had another impression of how this would work, could you go into more detail on how this works and use cases it could cover? From what I'm seeing is that we always check if the variables exist otherwise we add a default value, so we can probably guarantee that when importing the modules they don't break right?

pat270:
This is a more extreme use case where a developer doesn't want to fully import Clay. They can pick and choose what to import by modifying `clay.scss` in DXP. In the example below I'm only importing the CSS reset and Buttons from Clay:

```scss
@import 'functions/_global-functions.scss';

@import 'variables/_globals.scss';
@import 'variables/_globals-z-index.scss';
@import 'variables/_buttons.scss';

@import '_mixins.scss';

@import 'components/_reboot.scss';
@import 'components/_buttons.scss';
```

> From what I'm seeing is that we always check if the variables exist otherwise we add a default value, so we can probably guarantee that when importing the modules they don't break right?

Yes, a lot of the component variables depend on other component variables which makes the pattern above require more work. You will need to manually override the offending variables in `_clay_variables.scss`. The check makes it default to a value that exists. The only files that are required is:

```scss
@import 'functions/_global-functions.scss';

@import 'variables/_globals.scss';
@import 'variables/_globals-z-index.scss';

// individual variable imports go here

@import '_mixins.scss';

// individual component imports go here
```

Everything else can be individually imported. We can probably allow `_mixins.scss` to be individually imported as well, but I think it should be done in a separate issue.

matuzalemsteles:
@pat270 thanks for the detailed comment! This makes sense, the dependencies between component variables are because some depend on other components eg dropdown has some button variables? Is this a deeper issue ie should we rely less on variables across components or is this not an issue?

I'm wondering if we should work on this to simplify our code base. If we have dependency on button variables inside the DropDown, that can mean two things, we are not using global variables to simplify the use, for example font size variables or we are overriding button classes inside the dropdown.

The second case I'm worried about how we deal with it, there are some exception cases like a button inside DropDown that its small size is different from the "normal" small. Other than that, we may be repeating a lot of code or building a lot of exclusive classes for the same elements that are part of different components, mainly buttons and texts. What do you think about this? I'm going to go ahead and merge this PR anyway because that would be a separate discussion.

matuzalemsteles (approved):
LGTM

PR #5208, opened 2022-11-22
feat(@clayui/date-picker): implements keyboard navigation to DatePicker

### LINKED ISSUE #5201: DatePicker component is not accessible

matuzalemsteles:
Fixes #5201

This PR finalizes the accessibility improvements for DatePicker by adding keyboard navigation and hotkeys as per w3c recommendations.

- [x] Keyboard navigation on the grid
- [x] Keyboard shortcuts on the grid
- [x] Focus via tab gets stuck inside Date Picker

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,7 +14,7 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-autocomplete/src/': {
-			branches: 79,
+			branches: 77,
 			functions: 93,
 			lines: 92,
 			statements: 92,
@@ -74,9 +74,9 @@ module.exports = {
 			statements: 82,
 		},
 		'./packages/clay-drop-down/src/': {
-			branches: 60,
-			functions: 57,
-			lines: 73,
+			branches: 58,
+			functions: 53,
+			lines: 72,
 			statements: 72,
 		},
 		'./packages/clay-empty-state/src/': {
--- a/jest.config.js
+++ b/jest.config.js
@@ -70,7 +70,7 @@ module.exports = {
 		'./packages/clay-date-picker/src/': {
 			branches: 56,
 			functions: 89,
-			lines: 82,
+			lines: 81,
 			statements: 82,
 		},
 		'./packages/clay-drop-down/src/': {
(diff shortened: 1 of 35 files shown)


matuzalemsteles:
I'm submitting the implementation of a new internal component is generic to handle overlay for Clay components and have consistent behavior. This makes code maintenance easier so that we don't have to replicate for all components, especially now that we're dealing with accessibility.

This component will also help to remove a lot of duplicate code in dialog and menu related components, it also develops a better mechanism for click out and suppression of non-assertive elements for the screen reader and lock focus within scope. This will benefit components like Modal, MultiSelect, Autocomplete, DropDown, DatePicker, ColorPicker and others.

I'm still going to send another commit that updates the DropDown component and consequently updates the other components that are using to go ahead with this PR and finish it. I will send the changes to Modal using another PR and another issue.

matuzalemsteles:
Hey @marcoscv-work could you review the browsing behavior for the DatePicker?

marcoscv-work:
I looks really great, button text are fantastic!
I think we could go ahead with this since is 1000 better than we already have and then provide improvements, thanks @matuzalemsteles 

We probably could improve the button of "Go to today" moving the focus to the day and probably the three icons could have titles since they are not something recognisable and need to be concrete:
image

So yeah, pretty sure that we would need "Previous month" "Go to today" and "Next month" titles.
Not something needs to be added right now if you need to release but good to add to the task-list.

Thanks @matuzalemsteles

marcoscv-work:
I wanted to test it in Windows/JAWS but preview site is totally blocked in windows :-P

matuzalemsteles:
> So yeah, pretty sure that we would need "Previous month" "Go to today" and "Next month" titles.
> Not something needs to be added right now if you need to release but good to add to the task-list.

Yes, that's really nice, I'll add to these elements it won't be a problem, we just had the `aria-label` label defined.

> I wanted to test it in Windows/JAWS but preview site is totally blocked in windows :-P

Oh no, really?! Is it not rendering or do you really see a blocking message? I can see that, maybe it's Netlify.

PR #5212, opened 2022-11-23
fix(@clayui/css): Replace improper clay-container uses with clay-css mixins

### LINKED ISSUE #5211: @clayui/css: Replace improper uses of clay-container mixin with the newer clay-css mixin
There are many places in Clay CSS components where we are using the `clay-container` mixin. This was used before we created the more robust `clay-css` mixin which accepts many more properties. The `clay-container` mixin should only be used for `.container` type classes.

https://github.com/liferay/clay/blob/1b7095fa95607b82780eb2b673364782919e9713/packages/clay-css/src/scss/components/_date-picker.scss#L53-L55

is one example.

pat270:
fixes #5211

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_alerts.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_alerts.scss
@@ -104,7 +104,7 @@ $alert-indicator: map-deep-merge(
 // .alert-footer
 
 $alert-footer: () !default;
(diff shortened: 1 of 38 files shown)


matuzalemsteles (approved):
LGTM! Good job, it's good to clean up some stuff in Clay.

PR #5223, opened 2022-11-30
feat(@clayui/shared): add component to handle Overlay for Clay components

matuzalemsteles:
This PR is a fork of #5208 where it was originally implemented, I'm sending it separately from this PR so we can have this merged into the master and we can work on other issues that depend on it as well while the PR is under review.

In summary, this PR is implementing a new foundation component to handle accessibility and keyboard focus and navigation interactions for components that need Overlay, for example: DatePicker, DropDown, Autocomplete, Modal, ColorPicker and other components.

Currently we have a duplication of implementation spread across these components, for example Modal has its own focus trap system, click outside the modal and suppression of elements outside the modal, this is implemented in other places too, which increases the duplication and possible inconsistent behavior and likely bugs over time.

Adding this component helps resolve these maintenance issues and inconsistency, also this PR implements better features for focus trap and click away from content which are more responsive.

- Improves the suppression system for elements outside the content. Uses the [native implementation of `inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) when available which causes the browser to ignore input events for the element, including focus events and events from assistive technology. When not available we use the old method of using `aria-hidden`.
  - https://developer.chrome.com/articles/inert/
- Adding a new system to handle clicks outside of content that is adaptive and hardware agnostic, using [`pointer` events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) to handle mouse, touch, and stylus/pen clicks. When the `pointer` is not available it uses the `mouse` and `touch` events.
- A focus trap system integrated with FocusScope, which ensures that the focus is trapped within the content. Good for Modal and DatePicker.
- Closing the Overlay by pressing the Escape key returns focus to the trigger. This makes this behavior consistent for all components that use Overlay.
- When the element has no focus trap, exiting the overlay using focus closes automatically. This was not consistent with some components that had focus inside the overlay but didn't need the focus trap.

### DIFF
--- a/packages/clay-shared/src/Overlay.tsx
+++ b/packages/clay-shared/src/Overlay.tsx
@@ -1,0 +1,172 @@
+/**
+ * SPDX-FileCopyrightText: © 2022 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {hideOthers, suppressOthers} from 'aria-hidden';
+import React, {useCallback, useEffect, useRef} from 'react';
+
+import {Keys} from './Keys';
+import {ClayPortal} from './Portal';
+import {useInteractOutside} from './useInteractOutside';
+
+import type {Undo} from 'aria-hidden';
+
+type Props = {
+	children: React.ReactElement;
+	isCloseOnInteractOutside?: boolean;
+	isKeyboardDismiss?: boolean;
+	isModal?: boolean;
+	isOpen: boolean;
+	menuRef: React.RefObject<HTMLElement>;
+	onClose: () => void;
+	portalRef?: React.RefObject<HTMLElement>;
+	triggerRef: React.RefObject<HTMLElement>;
+};
+
+/**
+ * Inert is a new native feature to better handle DOM arias that are not
+ * assertive to a SR or that should ignore any user interaction.
+ * https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert
+ */
+const hasInertSupport =
+	typeof window !== 'undefined' &&
+	// @ts-ignore
+	typeof document.createElement('div').inert === 'boolean';
+
+/**
+ * Overlay component is used for components like dialog and modal.
+ * For example, Autocomplete, DatePicker, ColorPicker, DropDown are components
+ * that can use and have the same consistent behavior.
+ */
+export function Overlay({
+	children,
+	isCloseOnInteractOutside = false,
+	isKeyboardDismiss = false,
+	isModal = false,
+	isOpen = false,
+	menuRef,
+	onClose,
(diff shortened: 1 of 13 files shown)


packages/clay-shared/src/Overlay.tsx
```diff
@@ -0,0 +1,172 @@
...
+			onClose();
+		},
+		ref: portalRef ?? menuRef,
+		triggerRef,
+	});
+
+	useEffect(() => {
+		if (menuRef.current && isOpen) {
+			if (isModal && hasInertSupport) {
+				unsuppressCallbackRef.current = suppressOthers(menuRef.current);
```
ethib137: Hey @matuzalemsteles, do we need this `hasInertSupport` check here? `suppressOthers` already [checks if inert is supported](https://github.com/theKashey/aria-hidden/blob/master/src/index.ts#L207). Wouldn't this be enough?

Additionally, wouldn't this check: `if (isModal && hasInertSupport)` result in the background being disabled for all menu types not just modals? Is that the behavior we want?

By the way, I really love that this functionality is being moved to a shared location! It's such a benefit for our ability to improve and support Clay! 😄

matuzalemsteles: Hi @ethib137, oh I didn't know about the internal implementation check... but for our case `inert` should still only be used for modal cases, where there is a Dialog and DatePicker that intend to have a focus trap, this is different for a Menu , be it Actions or other common DropDown behavior, where the focus doesn't exist inside the Menu but pressing it should move the focus to the next element, using `inert` here would break these behaviors for those cases, because it also blocks any interaction from the user on these elements, so we only enable it for cases that want the focus trap.

The use of `aria-hidden` for these Menu cases is because the content is rendered in the body and it breaks the navigation based on hierarchy when using the virtual cursor of the screen readers, because it doesn't know which element to move the focus, `aria-hidden` helps us with that because it forces us to move the focus into the menu or the virtual cursor.

We will also still need to check that `inert` is supported, as it disables any user interaction on these elements, we have some behavior like pressing esc when moving focus to the trigger doesn't work, because we need to disable `inert` first and then move, this is not necessary with the use of `aria-hidden`.

ethib137: Thanks @matuzalemsteles. That makes sense. I wasn't sure why we would want to use aria-hidden if it wasn't a modal, but now I get it. It's not a big deal but if you don't want to have to maintain your own method `hasInertSupport` the aria-hidden package also exports it's own [supportsInert](https://github.com/theKashey/aria-hidden/blob/master/src/index.ts#L197) method.

matuzalemsteles: Oh great @ethib137, I'll use this!

PR #5230, opened 2022-12-02
feat(@clayui/css): Adds gap utilities

### LINKED ISSUE #5218: Add gap utility classes to Clay CSS
[As discussed on Slack](https://app.slack.com/client/T03BTCQAJ/threads).

## What is your proposal?

Adding `gap-*`, `row-gap-*` and `column-gap-*` utility classes (similar to `m-*` and `p-*`). They should follow the same spacing values than our existing classes (`gap-0`, `gap-1`, etc.).

https://developer.mozilla.org/en-US/docs/Web/CSS/gap
https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap
https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap

When using flex layouts (and [grid, since it has a nice support](https://caniuse.com/css-grid)) you can use `gap` property to specify space between elements, instead of adding some margin to individual elements, converting this:

```
<div class="d-flex">
  <button class="mr-2" />
  <button class="mr-2" />
  <button />
</div>
```

into this:

```
<div class="d-flex gap-2">
  <button />
  <button />
  <button />
</div>
```

which is much more easy to write and read.

This proposal is just an improvement to bring modern CSS to utility classes (and potentially adding grid too in the future).

We already have alternatives to that: most Clay components already add all needed spaces between elements, and using margins (with if statements in loops) just works.

pat270:
fixes #5218

Adds gap utilities:

|Class Name|CSS Property|
|---|---|
|.c-g-{breakpoint}-{#}|gap|
|.c-gy-{breakpoint}-{#}|row-gap|
|.c-gx-{breakpoint}-{#}|column-gap|


matuzalemsteles (approved):
LGTM! I don't have a very strong opinion on nomenclatures but I do have a preference for a clearer rather than short nomenclature, for example `c-g-3` because it's something new I wouldn't understand what that means, `c-gap-3` sounds more accurate.

I know we have a short naming pattern, for `mt`, `m`... maybe those work really well because they're widespread in the community and other frameworks, so it ends up becoming more known.

Let me know what you think.

pat270:
I'm good with `c-gap-#`, `c-gapy-#`, `c-gapx-#` too. @p2kmgcl @marcoscv-work any opinions?

p2kmgcl:
I'm more into verbose variable names, so `c-gapx-#` and variations look perfect to me :smile:  (also not a strong opinion here).

ethib137:
I would also recommend using -gap- instead of just -g- since that is in line with what both [Bootstrap](https://getbootstrap.com/docs/5.0/utilities/spacing/#gap) and [Tailwind](https://tailwindcss.com/docs/gap) are doing.

PR #5232, opened 2022-12-02
fix(@clayui/panel): Remove fixed height after panel is shown

### LINKED ISSUE #5225: @clayui/panel: Height isn't removed from panel-collapse when shown
https://clayui.com/docs/components/panel.html

There is a fixed height when a panel is open. This causes issues when more content is added to the `panel-body`. The height of the panel remains the same instead of adjusting to fit.

panel-fixed

pat270:
fixes #5225


matuzalemsteles (approved):
LGTM! Thanks @pat270

Issue #5235, opened 2022-12-05
@clayui/modal: Modal window closes when double clicking the file in the Select window


rolandpakai:
https://issues.liferay.com/browse/LPS-169353

On Win10 in case in the File open window the selectable file icon is outside of the modal then the double click is registred outside the modal which fires the onClose event in the modal.

**Steps to reproduce in Liferay**

0. Start a Liferay bundle in Win10
1. Go to the Content & Data → Web Content → Structures
2. Click on the 3 dots in the black bar and select Import structure
3. Click 'Select' to open the file selection panel
4. The selecatble file in the file select window should be outside of the modal like in this screenshot.
5. In the file selection panel, double-click the file you want to select

**Expected result:** The structure is imported

**Actual result:** The import structure panel closed itself without importing the JSON file

![image](https://user-images.githubusercontent.com/35058963/205631068-d286b5f1-84a2-4782-9d70-1c49234cc986.png)

PR #5236, opened 2022-12-05
fix(@clayui/modal): fix double click in the file select window in case outside of the modal

rolandpakai:
Hi Team,

could you please check my fix?
This fixes: https://github.com/liferay/clay/issues/5235

Thanks, Roland

cc: @matuzalemsteles


rolandpakai:
Thank you @matuzalemsteles

PR #5238, opened 2022-12-05
docs(@clayui/table): adds caption to table examples

### LINKED ISSUE #5234: Table: caption and summary
We could meet some customer suggestions.

## Context

https://www.w3.org/WAI/tutorials/tables/caption-summary/

_Captions and summaries provide information that can help users find, navigate, and understand tables. While they are not required in every case to meet WCAG, captions and summaries are fairly straightforward ways to provide such information that is often needed._

By helping to add subtitles we can encourage their use and improve overall accessibility.

Right now we can add a for example a <p> with Id and then add `aria-describedby` to table component, but `<caption>` make the description inside the table markup.

```html
<p id="example">This is a summary of the table</p>

<!-- This is the component with aria-describedby -->

<table aria-describedby="example">
</table>
```

matuzalemsteles:
Closes #5234

### DIFF
--- a/packages/clay-table/docs/index.js
+++ b/packages/clay-table/docs/index.js
@@ -15,6 +15,7 @@ const tableCode = `const Component = () => {
 
 	return (
 		<ClayTable>
+			<caption>Table caption</caption>
 			<ClayTable.Head>
 				<ClayTable.Row>
 					<ClayTable.Cell expanded headingCell>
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
Builds are failing for other reasons, there seems to be some issue with Netlify. I'm moving forward with that.

PR #5239, opened 2022-12-06
docs: Adds `aria-label` and `title` to documentation examples

### LINKED ISSUE #5237: clayui.com documentation accessibility improvements
## Use aria-label and titles in examples

If you read the web with a screen reader you can find a lot of buttons/links without description or wrong markup, I think adding them can be positive to help people to copy correct markup.

For example: https://clayui.com/docs/components/modal/markup.html#claySmallModal
image
It has a `aria-labelledby` from a not existing element so it is read as "button". and it has `role="button"` even when the item is a button. So we could probably review and add good practices to components uses and markup examples

If we provide perfect markup/accessibility examples it will help to be reused

matuzalemsteles:
Fixes #5237

This PR is further updating all the examples to add `aria-label` and `title` to elements that only have icons. I also ended up adding a new property to the Breadcrumb, Pagination and PaginationBar component to be able to configure `aria-label` and `title` for the ellipsis trigger.

### DIFF
--- a/packages/clay-breadcrumb/docs/index.js
+++ b/packages/clay-breadcrumb/docs/index.js
@@ -14,6 +14,7 @@ const BreadcrumbCode = `const Component = () => {
 	return (
 		<ClayBreadcrumb
 			ellipsisBuffer={1}
+			ellipsisProps={{'aria-label': 'More', title: 'More'}}
 			items={[
 				{
 					active: true,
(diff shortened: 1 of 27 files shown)


matuzalemsteles:
The preview deploy fails, looks strange and unrelated to the changes, fails because `cpx` is not installed but it looks like some problem with the Netlify build, we are having a lot of build issues like this. But this doesn't seem to be consistent, it just failed the last two PRs I submitted but it's not failing the other PRs and not the PRs with this same glitch that was merged into the master, I'll go ahead with that and monitor the issue.

Issue #5226, opened 2022-12-01
New Component: Custom Select


marcoscv-work:
## What is your proposal?

**Create a new clay component, Custom Select, or Custom Combobox Select.**

![image](https://user-images.githubusercontent.com/7963804/205102104-798c33e1-34f3-443f-b03d-286858287e03.png)

Context:
From some accessibility issues in DXP like this: https://issues.liferay.com/browse/LPS-167535 we decided to init the debate here: https://github.com/liferay/clay/issues/5181
After the debate about change dropdown to make it configurable as a select we finally decided to create a new component, we created a design task:

Design task:
https://issues.liferay.com/browse/DESIGN-4801

Design final documentation:
https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2178057030/Custom+Select

@emiliano-cicero's design phase defines the component quite detailed, here we can highlight some things:

- Now, the component visually looks like a select
- We recommend to provide native one when possible in mobile, in order to make the component much more usable and accessible, here a quick explanation:
  - https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2179170308/Custom+VS+Native
  - If make hybrid component is not possible at the beginning, we can make the same view for mobile and  study how to be visually closer to native select view

## Technical side:
We need to apply roles and markup from https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html
With all improvements we could make: https://github.com/liferay/clay/issues/5181#issuecomment-1316005047
Of course if we can provide in someway (ScreenReader, markup validation, etc). the suggested solution is better.

The componen needs to be defined following 

## Why would adopting this proposal be beneficial?

It will allow us to fix tons of issues in DXP, and will allow designers to provide better solution for 

## What are the alternatives to this proposal?

We already discussed if make this as a part of the dropdown menu could be good idea:
https://github.com/liferay/clay/issues/5181

pat270:
@marcoscv-work The Custom Select default style is slightly different from the regular select style. White background instead of gray, is this intentional?

![regular-select](https://user-images.githubusercontent.com/788266/205185580-04600a93-6b9d-4fa7-83da-db1aea126d14.jpg)

emiliano-cicero:
[...]

PR #5246, opened 2022-12-08
feat(@clayui/css): Adds variants to be used in Custom Select

pat270:
#5226

@matuzalemsteles I added the variants `form-control-select-secondary` and `dropdown-menu-select`. The custom select element markup looks like:

```html
<div class="form-group">
	<label for="formControlSelectSecondary1">A Button Styled Like a Select Element</label>
	<button class="form-control form-control-select form-control-select-secondary" id="formControlSelectSecondary1" type="button">Custom Select</button>
</div>
```

For the dropdown menu, we just need to add the class `dropdown-menu-select`. It should work with all the `dropdown-menu` modifier classes. I'm going to mark this as draft since the specs don't seem to be done.

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_forms.scss
@@ -75,6 +75,10 @@ $input: map-deep-merge(
 	$input
 );
 
+// Form Control Variants
+
+$input-palette: () !default;
+
 $input-height-border: calc(
 	#{$input-border-bottom-width} + #{$input-border-top-width}
 ) !default;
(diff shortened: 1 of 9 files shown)


matuzalemsteles:
Thanks @pat270, I'm going to use this PR to send the commits of the component on top of that.

matuzalemsteles:
Hey @pat270 I ended up sent the implementation in React.js in another PR #5248 to better separate the discussions and comments about it.

I noticed some issues with the css part and would also need some support modifier class, let me know what you think.

- Can we have a modified `hover` class as support? in the implementation in React.js, we control this manually due to keyboard navigation, another option would be to have `focus` to visually not look like `hover`.
- It looks like the font weight of the text on the `button` appears to be bold.

Screen Shot 2022-12-12 at 11 12 29 PM

The text inside the button is center aligned, maybe that should be left aligned text?

Screen Shot 2022-12-12 at 11 13 49 PM

I'm not sure about that but the Lexicon can confirm better, but the width of DropDown in the Confluence documentation seems to be smaller, it's a little weird in the width size with a lot of empty space.

Maybe the button width should be `auto` instead of `100%`? I'm not sure if it's intended to take up the space of the entire container, however this can be adjusted according to the developer's case.

matuzalemsteles (approved):
LGTM!

PR #5248, opened 2022-12-13
feat(@clayui/core): Adds new Picker/Custom Select component

### LINKED ISSUE #5226: New Component: Custom Select

matuzalemsteles:
Closes #5226

I'm submitting it as a separate PR from #5246, so I can add more info on the component's implementation in React.js and also to better separate the discussions. There are some CSS issues that still need to be fixed, I added some comments about that in PR #5246.

Firstly I went ahead with the `<Picker />` name here for combobox-only selection, I can't find another alternative here, names with Custom Select synthetically are bad to read and sound like customization/alternative, in which case I I think we don't want to just leave this as an alternative but be the main component to replace the native Select component.

Well, this component is following the w3c recommendations https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html. It also implements Clay patterns like Collection and Controlled and Uncontrolled components, which makes the component simple to use. This component doesn't have all Collection features implemented, for example virtualization and load more, we still need to improve keyboard navigation in a virtualized list, this is being done in #5205, so I haven't enabled virtualization here, and we do not have the load more/data paginated implemented for a non-virtualized list, I will create an issue to be able to work on that soon.

## Design Principles

As this component only tends to support behaviors of a select with OOTB accessibility and other resources delivered by the Clay foundation, the composition is very close and common to a DropDown, VerticalBar plus some more simplified APIs that do not exist in a DropDown for example.

```jsx
<Form.Group>
    <label htmlFor={pickerId} id={labelId}>
        Choose a fruit
    </label>
    <Picker aria-labelledby={labelId} id={pickerId}>
        <Option key="apple">Apple</Option>
        <Option disabled key="banana">Banana</Option>
        <Option key="mangos">Mangos</Option>
    </Picker>
</Form.Group>
```

The component's design is very close to a native Select, this allows us to implement this component as a hybrid that allows switching between a native selector or the custom implementation.

Data-agnostic composition also allows you to add composition as a custom content of the options, where possible, and still provide support for the native selector with the support of some attributes, for example:

```jsx
<Form.Group>
    <label htmlFor={pickerId} id={labelId}>
        Choose a fruit
    </label>
    <Picker aria-labelledby={labelId} id={pickerId}>
        <Option key="apple" textValue="Apple">
            <Text>Apple</Text>
            <Text size={3} color="secondary">Edible fruit produced by an apple tree.</Text>
        </Option>
        <Option key="banana" textValue="Banana">
            <Text>Banana</Text>
            <Text size={3} color="secondary">Elongated, edible fruit – botanically a berry.</Text>
        </Option>
        <Option key="mangos" textValue="Mangos">
            <Text>Mangos</Text>
            <Text size={3} color="secondary">Edible stone fruit produced by the tropical tree Mangifera indica.</Text>
        </Option>
    </Picker>
</Form.Group>
```

This allows switching to a native picker easily just by setting the attribute on the `<Picker />` component, this part is yet to be implemented.

Unlike a component like DropDown, customizing the trigger is not possible, because that can easily break the accessibility part, but we can improve this over time as it becomes clearer what alternatives people want.

### ToDo

- [x] Add component documentation
- [x] Add tests
- [x] Add native selector support

#### Accessibility

- [x] Printable Characters - add support for typeahead when the Menu is closed, this already works when the Menu is open.
- [x] Navigation with the PageUp & PageDown keys
- [x] Click outside Menu select option with activeDescendant

### DIFF
--- a/packages/clay-core/stories/Picker.stories.tsx
+++ b/packages/clay-core/stories/Picker.stories.tsx
@@ -5,10 +5,13 @@
 
 import Form from '@clayui/form';
 import Icon from '@clayui/icon';
+import Label from '@clayui/label';
+import Layout from '@clayui/layout';
 import {useId} from '@clayui/shared';
 import React from 'react';
 
 import {Option, Picker} from '../src/picker';
+import {Text} from '../src/typography';
 
 export default {
 	component: Picker,
@@ -105,6 +108,40 @@ export const CustomTrigger = () => {
 				</label>
 				<Picker as={Trigger} items={['Liam', 'Noah', 'Oliver']}>
 					{(item) => <Option key={item}>{item}</Option>}
+				</Picker>
+			</Form.Group>
+		</div>
+	);
+};
+
+export const CustomOptions = () => {
+	const pickerId = useId();
+	const labelId = useId();
+
+	return (
+		<div style={{width: '180px'}}>
+			<Form.Group>
+				<label htmlFor={pickerId} id={labelId}>
+					Choose a user
+				</label>
+				<Picker items={['Liam', 'Noah', 'Oliver']}>
+					{(item) => (
+						<Option key={item} textValue={item}>
+							<Layout.ContentRow>
+								<Layout.ContentCol expand>
+									<Text size={3} weight="semi-bold">
+										{item}
+									</Text>
+									<Text color="secondary" size={2}>
+										Description
+									</Text>
+								</Layout.ContentCol>
+								<Layout.ContentCol>
+									<Label displayType="success">Active</Label>
+								</Layout.ContentCol>
+							</Layout.ContentRow>
+						</Option>
+					)}
 				</Picker>
 			</Form.Group>
 		</div>
--- a/packages/clay-core/stories/Picker.stories.tsx
+++ b/packages/clay-core/stories/Picker.stories.tsx
@@ -1,0 +1,81 @@
+/**
+ * SPDX-FileCopyrightText: © 2022 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import Form from '@clayui/form';
+import {useId} from '@clayui/shared';
+import React from 'react';
+
+import {Option, Picker} from '../src/picker';
+
+export default {
+	component: Picker,
+	title: 'Design System/Components/Picker',
+};
+
+export const Default = () => {
+	const pickerId = useId();
+	const labelId = useId();
+
+	return (
+		<div style={{width: '150px'}}>
+			<Form.Group>
+				<label htmlFor={pickerId} id={labelId}>
+					Choose a fruit
+				</label>
+				<Picker aria-labelledby={labelId} id={pickerId}>
+					<Option key="apple">Apple</Option>
+					<Option disabled key="banana">
+						Banana
+					</Option>
+					<Option key="mangos">Mangos</Option>
+					<Option key="blueberry">Blueberry</Option>
+					<Option key="boysenberry">Boysenberry</Option>
+					<Option key="cherry">Cherry</Option>
+					<Option key="cranberry">Cranberry</Option>
+					<Option key="eggplant">Eggplant</Option>
+					<Option key="fig">Fig</Option>
+					<Option key="grape">Grape</Option>
+					<Option key="guava">Guava</Option>
+					<Option key="huckleberry">Huckleberry</Option>
+				</Picker>
+			</Form.Group>
+		</div>
+	);
+};
+
+export const Dynamic = () => {
+	const pickerId = useId();
+	const labelId = useId();
+
+	return (
+		<div style={{width: '150px'}}>
+			<Form.Group>
+				<label htmlFor={pickerId} id={labelId}>
+					Choose a fruit
+				</label>
+				<Picker
+					aria-labelledby={labelId}
+					id={pickerId}
+					items={[
+						'Apple',
+						'Banana',
+						'Mangos',
+						'Blueberry',
+						'Boysenberry',
+						'Cherry',
+						'Cranberry',
+						'Eggplant',
+						'Fig',
+						'Grape',
+						'Guava',
+						'Huckleberry',
+					]}
+				>
+					{(item) => <Option key={item}>{item}</Option>}
+				</Picker>
+			</Form.Group>
+		</div>
+	);
+};
--- a/packages/clay-core/stories/Picker.stories.tsx
+++ b/packages/clay-core/stories/Picker.stories.tsx
@@ -4,6 +4,7 @@
  */
 
 import Form from '@clayui/form';
+import Icon from '@clayui/icon';
 import {useId} from '@clayui/shared';
 import React from 'react';
 
@@ -73,6 +74,36 @@ export const Dynamic = () => {
 						'Huckleberry',
 					]}
 				>
+					{(item) => <Option key={item}>{item}</Option>}
+				</Picker>
+			</Form.Group>
+		</div>
+	);
+};
+
+const Trigger = React.forwardRef<
+	HTMLDivElement,
+	React.HTMLAttributes<HTMLDivElement>
+>(({children, ...otherProps}, ref) => (
+	<div ref={ref} {...otherProps} tabIndex={0}>
+		<Icon className="mr-2" symbol="user" />
+		{children}
+	</div>
+));
+
+Trigger.displayName = 'Trigger';
+
+export const CustomTrigger = () => {
+	const pickerId = useId();
+	const labelId = useId();
+
+	return (
+		<div style={{width: '180px'}}>
+			<Form.Group>
+				<label htmlFor={pickerId} id={labelId}>
+					Choose a user
+				</label>
+				<Picker as={Trigger} items={['Liam', 'Noah', 'Oliver']}>
 					{(item) => <Option key={item}>{item}</Option>}
 				</Picker>
 			</Form.Group>
--- a/packages/clay-core/stories/Picker.stories.tsx
+++ b/packages/clay-core/stories/Picker.stories.tsx
@@ -106,7 +106,12 @@ export const CustomTrigger = () => {
 				<label htmlFor={pickerId} id={labelId}>
 					Choose a user
 				</label>
-				<Picker as={Trigger} items={['Liam', 'Noah', 'Oliver']}>
+				<Picker
+					aria-labelledby={labelId}
+					as={Trigger}
+					id={pickerId}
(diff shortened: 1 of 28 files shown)


pat270:
@matuzalemsteles `d-contents` is good with me.

matuzalemsteles:
I sent more commits today adding the missing features to meet final accessibility requirements. I also made some more internal improvements to the component, I'll try to describe the main changes but the commit history also helps to understand and some of them have very detailed descriptions.

As we need to use accessibility focus with `aria-activedescendant`, we control it not only with keyboard navigation but also with mouse, we can simulate a better user experience by controlling focus and hover according to user interaction. See the example in the video below.

https://user-images.githubusercontent.com/13750819/207781223-b5d99e37-2484-4f91-8887-31eda7f452ca.mov

I've refactored the collection implementation a bit to move the rendering part of the collection to the hook, this is just a part of it, but it helps us initialize the collection before the content is rendered, allowing us to cache the rendering in memory and use it only when rendered in the DOM. For example in the case of Picker, the design of the API is oriented to the properties that are defined in the options at render time, for example the content that is rendered in an option we use this to define the text in the form control when the option is selected, this much simplifies the API and reduce the need to add other APIs.

```tsx
const [isOpen, setIsOpen] = useState(false);

const collection = useCollection<T, unknown>({
  children,
  items,
});

return isOpen && (
  <Menu>
    <Collection collection={collection} />
  </Menu>
);
```

We were able to cache the rendering before the content was mounted and we avoided the computation of the collection every time the component is mounted and unmounted, we were also able to extract information from the collection even though it was not rendered. This is not supported yet for dynamic content with virtualization, there is more work that needs to be done on this, just supporting dynamic and static content.

I even mentioned that the Picker trigger is not customizable but in fact it is possible to customize with the `as` API that allows passing a component that will be used to replace the internal component of button. This is a simple API but it allows you to customize the trigger and still receive all the necessary properties that the trigger needs, as well as the selected content.

```jsx
<Picker
  as={({children, ...otherProps}) => (
    <div {...otherProps}>
      <Icon symbol="user" />
      {children}
    </div>
  )}
>
  <Option key="apple">Apple</Option>
  <Option key="banana">Banana</Option>
  <Option key="mangos">Mangos</Option>
</Picker>
```

This component in terms of implementation is ready, I still need to add all the test cases and add documentation, I added some simple tests for the BasicRendering suite but I still need to write more cases and add the IncrementalInteractions suite.

matuzalemsteles:
I sent the final commits to the component, adding documentation and testing, the testing coverage is pretty thorough and covers all the major use cases for the component, and especially the accessibility and navigation behaviors.

I also added the API that makes the component hybrid, the strategy here is that by enabling the `native` API of `<Picker />` when the component identifies that it is on a mobile device it renders the native selector, this is capable thanks to design of the component's APIs and the collection pattern that help simplify the use of the component.

I've added two examples in the storybook showing how to customize trigger and options, there is one case here that I'm not sure we should cover, the groups of options, this is not supported yet but it would be easy to add support but we need to know what would be the design whether it is similar to DropDown or is something new.

marcoscv-work:
Awesome @matuzalemsteles, I think we can go ahead. (I've tested only VoiceOver)

Just something super minor, I'm not sure if it is happening only in my mac, but the focus is blinking in the combobox and in the option each time I touch an arrow key:

https://user-images.githubusercontent.com/7963804/208005812-6dcafe51-c79d-4c59-91d0-1e6ad62fc968.mov

matuzalemsteles:
> Just something super minor, I'm not sure if it is happening only in my mac, but the focus is blinking in the combobox and in the option each time I touch an arrow key:

Yeah I figured that out too not sure what it might but I'll check it out.

packages/clay-core/stories/Picker.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+				</Picker>
+			</Form.Group>
+		</div>
+	);
+};
+
+const Trigger = React.forwardRef<
+	HTMLDivElement,
+	React.HTMLAttributes<HTMLDivElement>
+>(({children, ...otherProps}, ref) => (
```
ethib137, line 90: Trigger missing id and for attributes.

packages/clay-core/stories/Picker.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+	const pickerId = useId();
+	const labelId = useId();
+
+	return (
+		<div style={{width: '180px'}}>
+			<Form.Group>
+				<label htmlFor={pickerId} id={labelId}>
+					Choose a user
+				</label>
+				<Picker items={['Liam', 'Noah', 'Oliver']}>
```
ethib137, line 127: missing aria-labeledby and id attributes.

ethib137 (changes requested):
Looks great!

packages/clay-shared/src/useNavigation.tsx
```diff
@@ -69,140 +88,177 @@ export function useNavigation<T extends HTMLElement | null>({
 		}
 	}, [visible]);
 
-	const onKeyDown = useCallback((event: React.KeyboardEvent<HTMLElement>) => {
-		if (!containeRef.current) {
-			return;
-		}
+	const onKeyDown = useCallback(
+		(event: React.KeyboardEvent<HTMLElement>) => {
+			if (!containeRef.current) {
```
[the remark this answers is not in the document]
matuzalemsteles: Oh yes 😅, I don't think I realized that when I was writing, thanks!

packages/clay-core/stories/Picker.stories.tsx
matuzalemsteles: It receives all the internal properties that the component passes to the trigger.

For example:

```jsx
<Picker
  aria-labelledby={labelId}
  as={Trigger}
  id={pickerId}
  items={['Liam', 'Noah', 'Oliver']}
>
  {...}
</Picker>
```

Internally we set all the properties for the trigger, so we pass `aria-activedescendant`, `aria-controls`, events and etc. everything that is needed. The `as` API works as a trigger replacement which by default is a `button`, so when you pass a component it passes all the properties that are defined in the trigger to your custom trigger.

```tsx
function Picker<T>({as: As = 'button', ...otherProps}: Props<T>) {
  return (
    <As
      {...otherProps}
      aria-activedescendant={active ? activeDescendant : ''}
      aria-controls={active ? ariaControls : undefined}
    >
      ...
    </As>
  )
}
```

This is an example of how this pattern works, so that your custom receives all the necessary properties to keep the component working and accessible.

matuzalemsteles:
Apliquei os comentarios de Ivan no PR e fiz mais algumas melhorias da collection, isso permite que possa combinar diferentes colleções aninhadas e se comunicarem para montar a estrutura de layout na coleção raiz isso permite que o desenvolvedor possa combinar outros componentes que tenha colleção, por exemplo criar grupos dentro do Picker usando o componente do DropDown `<DropDown.Group />`.

```jsx
```

Screen Shot 2022-12-16 at 5 40 11 PM

matuzalemsteles:
I fixed the minor issues and made some more improvements to the collection, this allow to combine different nested collections and communicate with each other to assemble the layout structure in the root collection this allows the developer to combine other components that have a collection, for example create groups within the Picker using the DropDown component `<DropDown.Group />`.

```jsx
<Form.Group>
  <label htmlFor="picker-s" id="picker-s-label">
    Select an option
  </label>
  <Picker aria-labelledby="picker-s-label" id="picker-s">
    <DropDown.Group
      header="Fruit"
      items={['Apple', 'Banana', 'Mangos']}
    >
      {(item) => <Option key={item}>{item}</Option>}
    </DropDown.Group>
    <DropDown.Group
      header="Vegetable"
      items={['Potatoes', 'Tomatoes', 'Onions']}
    >
      {(item) => <Option key={item}>{item}</Option>}
    </DropDown.Group>
  </Picker>
</Form.Group>

// or if prefer to simplify even more

<Form.Group>
  <label htmlFor="picker-s" id="picker-s-label">
    Select an option
  </label>
  <Picker
    aria-labelledby="picker-s-label"
    id="picker-s"
    items={[
      {
        label: 'Fruit',
        items: ['Apple', 'Banana', 'Mangos'],
      },
      {
        label: 'Vegetable',
        items: ['Potatoes', 'Tomatoes', 'Onions'],
      },
    ]}
  >
    {(group) => (
      <DropDown.Group
        header={group.label}
        items={group.items}
      >
        {(item) => <Option key={item}>{item}</Option>}
      </DropDown.Group>
    )}
  </Picker>
</Form.Group>
```
Screen Shot 2022-12-16 at 5 40 11 PM

I investigated why the voiceover focus blinking when browsing but can't find an answer but it doesn't seem to affect the announcement, maybe it could be a bug but I'll keep an eye on it.

Overall I think this component is ready to go ahead, as we're all good with the implementation I'll go ahead and merge. Theupdates to the documentation about adding examples of customizations, we can send later.

marcoscv-work:
@matuzalemsteles Could we add spaces or something to labels in the examples?

image

It is read like it would be all the same word "LiamDescriptionACTIVE"

matuzalemsteles:
Hmm, I'll check that out maybe I'll have to add an `aria-label` or `aria-labelledby` with multiple ids so it can add the spaces and read as separate words.

PR #5251, opened 2022-12-15
feat(@clayui/pagination): PaginationWithBasicItems should have aria-labels by default

### LINKED ISSUE #5233: Pagination needs some adjustments to make the component 100% accessible
## Context:

Pagination need some adjustments to be totally accesible.

image

Is a pattern/component that from a design point of view is offering a really good new behaviour that can help to reach all pages with a visual reduction of noise when there is a lot of pages, here are running out of the standard here so we will need to try something new, here a list of changes/improvements

- A pagination must be a nav so we we need to add <nav> container
- This <nav> needs an aria-label "Pagination" by default but it could be customised, for example if there is more than one pagination in the page, so please offer the possibility of changing the default aria-label
- We should build the nav by links, so please let's not use buttons, they should be removed from the examples
- Current page link should be present (sometimes is a link without href), aria-current="page" is enough to identify the active page in the <nav>
- Trigger for showing level-2 pagination buttons -> [...] Should have an aria-label and title like: "Show pages links from X to X". In this way we can understand the decision of skipping or showing that group of pages
- Title and aria-label for next/prev pages should be always in the link and provided by default, if not we were creating a double focus, first focus then in label:
image _(Markup from DXP, list of web content)_
- If we need to hide next/last page link, just change the link for a no focusable element like a span, no buttons with disable or links with layers, they will be accessible by rotor and buttons and links lists

### Warning about dropdown/level-2:

- There is not examples about dropdown menus inside a pagination, we are innovating a little bit here, but if we respect the composition of a nav element everything should be Ok
- We cannot use here a clay dropdown, we need to keep the level-2 structure of a navigation always in the page and exposed to accessibility tree, so the structure should be something like (pseudocode):
```
<nav
   <ul level-1
       <li page-1
           "trigger" aria-expanded="" aria-controls=""
          <ul  level 2
               <li page-2
               <li page-3 ...
```
 - Use of aria-expanded="" aria-controls="" is needed

pat270:
fixes #5233

- This wraps the pagination in `<nav aria-label="Pagination" role="navigation">`. You can change the nav attributes with `navProps`.
- Adds a default `aria-label` for previous, items, and next buttons
- Adds a default `aria-label` and `title` "Show pages %START% through %END%"
- Only output a `ClayLink` no buttons for items. The ellipsis menu still contains buttons.

I think [Warning about dropdown/level-2](https://github.com/liferay/clay/issues/5233#issue-1475584259) should be done in another pr.

I still need to write documentation on this. I need feedback before I start.

### DIFF
--- a/packages/clay-pagination/src/Item.tsx
+++ b/packages/clay-pagination/src/Item.tsx
@@ -21,20 +21,16 @@ const ClayPaginationItem = ({
 	href,
 	...otherProps
 }: IPaginationItemProps) => {
-	const ElementTag = href ? ClayLink : 'button';
-
 	return (
 		<li className={classNames('page-item', {active, disabled})}>
-			<ElementTag
-				type={href ? undefined : 'button'}
+			<ClayLink
 				{...otherProps}
 				aria-current={active && href ? 'page' : undefined}
 				className="page-link"
-				disabled={href ? undefined : disabled || active}
 				href={disabled || active ? undefined : href}
 			>
 				{children}
-			</ElementTag>
+			</ClayLink>
 		</li>
 	);
 };
--- a/packages/clay-pagination/src/Item.tsx
+++ b/packages/clay-pagination/src/Item.tsx
@@ -28,6 +28,15 @@ const ClayPaginationItem = ({
 				aria-current={active && href ? 'page' : undefined}
 				className="page-link"
 				href={disabled || active ? undefined : href}
+				onClick={(event) => {
+					if (!href) {
+						event.preventDefault();
+					}
+
+					if (otherProps.onClick) {
+						otherProps.onClick(event);
+					}
+				}}
 			>
 				{children}
 			</ClayLink>
--- a/packages/clay-pagination/src/Pagination.tsx
+++ b/packages/clay-pagination/src/Pagination.tsx
@@ -11,30 +11,15 @@ import Item from './Item';
 
 interface IProps extends React.HTMLAttributes<HTMLUListElement> {
 	/**
-	 * Use this property for defining `otherProps` that will be passed to the `nav` element.
-	 */
-	navProps?: {
-		'aria-label'?: string;
-		role?: string;
-	};
-
-	/**
 	 * The size of pagination element.
 	 */
 	size?: 'lg' | 'sm';
 }
 
 const ClayPagination = React.forwardRef<HTMLUListElement, IProps>(
-	(
-		{children, className, navProps = {}, size, ...otherProps}: IProps,
-		ref
-	) => {
+	({children, className, size, ...otherProps}: IProps, ref) => {
 		return (
-			<nav
-				aria-label={navProps['aria-label'] || 'Pagination'}
-				role={navProps?.role || 'navigation'}
-				{...navProps}
-			>
+			<nav aria-label={otherProps['aria-label'] || 'Pagination'}>
 				<ul
 					{...otherProps}
 					className={classNames(
--- a/packages/clay-pagination/src/Pagination.tsx
+++ b/packages/clay-pagination/src/Pagination.tsx
@@ -11,23 +11,44 @@ import Item from './Item';
 
 interface IProps extends React.HTMLAttributes<HTMLUListElement> {
 	/**
+	 * Use this property for defining `otherProps` that will be passed to the `nav` element.
+	 */
+	navProps?: {
+		'aria-label'?: string;
+		role?: string;
+	};
+
+	/**
 	 * The size of pagination element.
 	 */
 	size?: 'lg' | 'sm';
 }
 
 const ClayPagination = React.forwardRef<HTMLUListElement, IProps>(
-	({children, className, size, ...otherProps}: IProps, ref) => {
+	(
+		{children, className, navProps = {}, size, ...otherProps}: IProps,
+		ref
+	) => {
 		return (
-			<ul
-				{...otherProps}
-				className={classNames('pagination pagination-root', className, {
-					[`pagination-${size}`]: size,
-				})}
-				ref={ref}
+			<nav
+				aria-label={navProps['aria-label'] || 'Pagination'}
+				role={navProps?.role || 'navigation'}
+				{...navProps}
 			>
-				{children}
-			</ul>
+				<ul
+					{...otherProps}
+					className={classNames(
+						'pagination pagination-root',
+						className,
+						{
+							[`pagination-${size}`]: size,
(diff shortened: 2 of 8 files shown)


packages/clay-pagination/src/Pagination.tsx
```diff
@@ -10,24 +10,45 @@ import Ellipsis from './Ellipsis';
...
 		return (
-			<ul
-				{...otherProps}
-				className={classNames('pagination pagination-root', className, {
-					[`pagination-${size}`]: size,
-				})}
-				ref={ref}
+			<nav
+				aria-label={navProps['aria-label'] || 'Pagination'}
+				role={navProps?.role || 'navigation'}
```
matuzalemsteles, line 35: I'm not sure if this should be configurable, I guess because it's a Pagination component it would have another option besides `navigation`?

packages/clay-pagination/src/Pagination.tsx
```diff
@@ -10,24 +10,45 @@ import Ellipsis from './Ellipsis';
 import Item from './Item';
 
 interface IProps extends React.HTMLAttributes<HTMLUListElement> {
+	/**
+	 * Use this property for defining `otherProps` that will be passed to the `nav` element.
+	 */
+	navProps?: {
+		'aria-label'?: string;
+		role?: string;
+	};
```
matuzalemsteles, line 19: I'm not sure of the `navProps` API in fact I want us to avoid APIs like this for components, this is a common prop drilling issue in React.js, these are usually signs that we don't have good compositing or something it's wrong.

Considering this component's scenario, we can avoid `navProps` and we don't need to leave the `role` configurable, I don't think we have another alternative for this component right?! then keep the `...otherProps` in the new `nav` container instead of the list.

packages/clay-pagination/src/PaginationWithBasicItems.tsx
```diff
@@ -105,14 +118,18 @@ const ClayPaginationWithBasicItems = React.forwardRef<HTMLUListElement, IProps>(
 			activePage,
 			alignmentPosition,
 			ariaLabels = {
-				next: 'Next',
-				previous: 'Previous',
+				link: 'Go to page, %PAGE%',
+				next: 'Go to the next page, %NEXT%',
+				previous: 'Go to the previous page, %PREVIOUS%',
```
matuzalemsteles, line 123: There is already a pattern we have for this, instead of the `%PAGE%` syntax, you can use `Go to page, {0}` and use the `sub` utility from the `@clayui/shared` package. It also works for multiple occurrences.

```js
sub(ariaLabels.link, [
  page,
])
```

packages/clay-pagination/src/Item.tsx
```diff
@@ -21,20 +21,16 @@ const ClayPaginationItem = ({
 	href,
 	...otherProps
 }: IPaginationItemProps) => {
-	const ElementTag = href ? ClayLink : 'button';
-
 	return (
 		<li className={classNames('page-item', {active, disabled})}>
-			<ElementTag
-				type={href ? undefined : 'button'}
+			<ClayLink
```
matuzalemsteles, line 26: I think to make sure this doesn't break some applications in DXP, we can add an `onClick` to `ClayLink` to do `event.preventDefault()` when `href` is not set and call `onClick` from `otherProps ` if defined.

This way we guarantee that the applications that were paging on the client side do not break.

packages/clay-pagination/src/Pagination.tsx
pat270: Wouldn't this be a breaking change? We've had otherProps mapped to the `ul` for a long time.

packages/clay-pagination/src/PaginationWithBasicItems.tsx
pat270: It looks like the sub method just replaces any digit wrapped in curly brackets. I might be missing something but is there a way we can specify what to sub? I'm wanting to use it for this label "Show pages %START% through %END%".

packages/clay-pagination/src/Pagination.tsx
matuzalemsteles: Yes, in theory that would be a significant change, but I'd rather do that than add a new API specific to it, if we think that adding `nav` is also considering a bug that the component didn't have, it's not a breaking change. In theory the `aria-label` will be in the right place, in the `nav` instead of the `ul` and just like role for example, what else could affect would be the `className`. It makes sense?

packages/clay-pagination/src/Item.tsx
pat270: @matuzalemsteles this is the only change I couldn't make. It was breaking too many tests and the links example would reload the url instead of updating the pagination.

packages/clay-pagination/src/Pagination.tsx
pat270: I ended up mapping `aria-label` to the `nav` and not applying `role="navigation"`. I looked at several pagination examples and some don't include it. I went with [Mozilla's example](https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Pagination). We can add it later if it's needed.

I think `...otherProps` is better mapped to the `ul` element. All the modifier classes are added on that element and moving it to `nav` would cause issues.

matuzalemsteles: Makes sense @pat270, looks good to me!

packages/clay-pagination/src/Item.tsx
matuzalemsteles: @pat270 hmm, this is weird, I guess it shouldn't break, have you tried something similar to this?

```jsx
<ClayLink
  onClick={(event) => {
    if (!href) {
      event.preventDefault();
    }

    if (otherProps.onClick) {
      otherProps.onClick(event);
    }
  }}
>
  {children}
</ClayLink>
```

matuzalemsteles (approved):
I think overall this looks good, I think we just need to check out the issue of ClayLink being ignored when it doesn't have an `href` defined, if you need help with that let me know.

packages/clay-pagination/src/Item.tsx
pat270: @matuzalemsteles Thanks, I didn't include the `otherProps.onClick`.

matuzalemsteles:
Thanks @pat270, I just made an update on the test threshold for the tests to pass. Looks good to go ahead now, I'll probably have to update the tests in DXP but it shouldn't be a big problem.

PR #5255, opened 2022-12-17
fix(@clayui/button): improves the error message for button accessibility and the ButtonWithIcon component type

### LINKED ISSUE #5249: Suggestions about accessibility warnings in Button with Icon
## What is your question?

First of all, thank you very much for introducing accessibility warnings for the Button with Icon. It's very cool because thanks to this, quite a few warnings have been found on the portal.

I'm working on solving the warnings that exist in the page editor and I have noticed 2 things:

One of them is about the message that it's announce in the warning: 
`Button Accessibility: Component has only the Icon declared. Define an 'aria-label' or 'title' attribute (consult your design team) that labels the interactive button that screen readers can read.` 
It's understood that adding `aria-label` or `title`, the warning will disappear. But only the `aria-label` is taken into account in the code. Could it be modified to make it less confusing for the developer?

Another question that has arisen to me is, why is the `aria-labelledby` attribute not taken into account? Its purpose is the same as `aria-label`, the only thing that changes is that it can be used to reference another element. In fact, if an element has both attributes, the `aria-labeled` attribute will take precedence ([info here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby)). Do you think we could contemplate this attribute?

Again, thank you very much for this! 😄

matuzalemsteles:
Fixes #5249

This improves the message for the button accessibility error which only ignored the message when `aria-label` was set but did not consider `aria-labelledby`. The message is that you have to define either one and the `title` attribute is optional.

It also improves the type of the `<ButtonWithIcon />` component that makes the type conditional, that is, if you don't declare any of the two properties the build will fail but if you declare any of the two the error will disappear.

### DIFF
--- a/packages/clay-button/src/Button.tsx
+++ b/packages/clay-button/src/Button.tsx
@@ -20,7 +20,7 @@ export type DisplayType =
 	| 'info'
 	| 'unstyled';
 
-interface IProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+export interface IProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
 	/**
 	 * Flag to indicate if button is used within an alert component.
 	 */
@@ -94,9 +94,10 @@ const ClayButton = React.forwardRef<HTMLButtonElement, IProps>(
 				childArray.length === 1 &&
 				// @ts-ignore
 				childArray[0].type?.displayName === 'ClayIcon' &&
-				typeof otherProps['aria-label'] !== 'string'
+				typeof otherProps['aria-label'] !== 'string' &&
+				typeof otherProps['aria-labelledby'] !== 'string'
 			),
-			'Button Accessibility: Component has only the Icon declared. Define an `aria-label` or `title` attribute (consult your design team) that labels the interactive button that screen readers can read.'
+			'Button Accessibility: Component has only the Icon declared. Define an `aria-label` or `aria-labelledby` attribute that labels the interactive button that screen readers can read. The `title` attribute is optional but consult your design team.'
 		);
 
 		return (
--- a/packages/clay-button/src/ButtonWithIcon.tsx
+++ b/packages/clay-button/src/ButtonWithIcon.tsx
@@ -8,12 +8,25 @@ import React from 'react';
 
 import ClayButton from './Button';
 
-interface IProps extends React.ComponentProps<typeof ClayButton> {
-	/**
-	 * Define a value that labels the button.
-	 */
-	'aria-label': string;
+import type {IProps} from './Button';
 
+type ButtonAria =
+	| {
+			/**
+			 * Define a value that labels the button.
+			 */
+			'aria-label': string;
+			'aria-labelledby'?: never;
+	  }
+	| {
+			/**
+			 * Define a value that labels the button.
+			 */
+			'aria-label'?: never;
+			'aria-labelledby': string;
+	  };
+
+interface ICommonProps extends Omit<IProps, 'aria-label' | 'aria-labelledby'> {
 	/**
 	 * Path to the location of the spritemap resource.
 	 */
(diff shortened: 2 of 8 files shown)


ethib137:
Thanks for the work on this! I may be misunderstanding this, but it sounds like the code is currently this:
```
// Not considering title.
if (!ariaLabel && !ariaLabelledBy) {
    throw error;
}
```

I would think we would want the logic to be like this:
```
if (!ariaLabel && !ariaLabelledBy && !title) {
    throw error;
}
```

So as long as an `aria-label`, `aria-labelledby`, or `title` attribute is used we won't throw an error. Isn't that what we want?

ethib137:
Additionally this seems like a good one to include tests to highlight what the expected behavior is for different combinations of props.

matuzalemsteles:
> So as long as an aria-label, aria-labelledby, or title attribute is used we won't throw an error. Isn't that what we want?

Actually, the title is like optional here, it depends on each use case, only `aria-label` and `aria-labelledby` is required. We only inform you that the `title` should be considered in the message but they should consult your design team.

> Additionally this seems like a good one to include tests to highlight what the expected behavior is for different combinations of props.

Hmm, I think I can add a test for that, good idea. I'll try to add some cases and see if I can come up with something, maybe I have to suppress the TypeScript errors because it will fail to build the tests.

matuzalemsteles:
> Yes, it's currently optional, but the point I'm trying to make is that we want to require that the user includes one of those 3 (title, aria-label, or aria-labeledby). As long as one of those is present, it will be read by a screen reader. So as an example, we shouldn't require aria-label if there is a title attribute since the screen reader can read the title attribute too.

I think that `title` does not replace `aria-label` or `aria-labeledby`, if I'm not mistaken they are used in different ways and have different behavior in screen readers @marcoscv-work can confirm here. I think the idea is that `title` is used with `aria-label` or `aria-labeledby`.

matuzalemsteles:
@ethib137 I'm going to go ahead with this PR to include this improvement in the release because it fixes a bug where it still shows the message if it has `aria-labeledby` instead of `aria-label` and it got confusing for some developers the error message. We may send another follow-up PR if we decide that `title` can replace `aria-label` and `aria-labeledby` and we can move to the required property.

ethib137:
Hey @matuzalemsteles that makes sense. I'm glad you moved forward this as is. :) 

In our case with buttons, using the title attribute is likely the most common thing that will happen, such as when adding a tooltip to an icon button. Using a title/tooltip is better than just using an aria-label since the text can also clarify the meaning of the icon for sighted users. 

I did some testing of title and aria-label with Jaws and Chrome's Screen Reader. The results were as follows:
|                | JAWS | Chrome Screen Reader |
| -------- | ------------ | --------- |
| Title Only | Reads only Title | Reads only Title |
| Aria Label Only | Reads Only Aria Label | Reads Only Aria Label |
| Title & Aria Label (Different) | Reads Title & Aria Label | Reads Only Aria Label |
| Title & Aria Label (Same) | Reads Title & Aria Label (Duplicated) | Reads Only Aria Label |

So with Jaws requiring aria-label or aria-labelledby even when title is used will result in reading both even if they are identical. This is why I think we should only require one of the 3 (title, aria-label, or aria-labeledby). 

Do any of you see any issue with this?

cc: @marcoscv-work

matuzalemsteles:
Hmm I see, these behaviors they are the same for NVDA and VoiceOver? I think I saw @marcoscv-work comment recently and also in some articles that `title` and `aria-label` have different behavior.

This was also raised in a thread on the accessibility channel https://liferay.slack.com/archives/C04CPMRMXAL/p1671627393080729 I think some things might be clearer now.

marcoscv-work:
We have a JAWS bug when `title` and `aria-label` has in the text the character `&`. Reported to JAWS!
And will ask to ONCE if they still using not-title reading configuration

PR #5257, opened 2022-12-19
fix(@clayui/css): Sidebar adds focus state

### LINKED ISSUE #5185: Vertical Bar: focus style bug inside the panel
With an external focus border, the result is that the sidebar has a focus that is hidden on top, and the right part goes off the container, probably because of the shadow. 
image

The vertical bar seems to also be customized to be buttons instead of our tabs, as a result the focus is a bit different than what it should be. 
image

no

1. Create a new page in portal
2. Open the page with the page editor to edit it 
3. tab inside the interface until you reach the vertical bar 
4. press Enter to access one of the options, it will focus the sidebar

The focus border in this case should be set as inside of the container and not outside in order to fit the sidebar. 
The color of the focus should also be $primary-l1 instead of $primary.
The vertical bar should use tabs instead of buttons to provide a squared item as defined by Lexicon (with also an inner border to avoid hiding parts when focusing).

image

Here's a [Figma](https://www.figma.com/file/i2GxPzLgF22A0VJE8kfSY2/vertical-bar-and-panel-focus-bug?node-id=1%3A2) mockup. 

## Environment
Master

| Tech  | Version |
| ----- | ------- |
| Clay  | v?.?.?  |
| React |         |

pat270:
fixes #5185 

Also added `$sidebar` and `$cadmin-sidebar` maps to style the base `.sidebar` component.

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_sidebar.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_sidebar.scss
@@ -1,9 +1,11 @@
 .sidebar,
 &.sidebar {
-	height: 100%;
-	overflow: auto;
+	@include clay-css($cadmin-sidebar);
 
-	-webkit-overflow-scrolling: touch;
+	&:focus,
+	&.focus {
+		@include clay-css(map-get($cadmin-sidebar, focus));
+	}
 
 	.container-fluid {
 		padding-left: $cadmin-sidebar-padding-left;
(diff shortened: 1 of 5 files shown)


matuzalemsteles (approved):
LGTM! I think we can add the focus to the VerticalBar panel too, it also has the focus. I'll merge this to add to the release, but feel free to send it in another PR.

PR #5259, opened 2022-12-20
feat(@clayui/multi-select): moves the visual focus from the `gridcell` to the label

matuzalemsteles:
From https://github.com/liferay/clay/pull/5204#issuecomment-1358397921

I made a different strategy to keep the visual focus on the label and still keep the accessibility focus on the `gridcell`, we are controlling the CSS class of `focus` manually to do that.

Screen Shot 2022-12-19 at 6 51 26 PM

@pat270 let me know if the CSS part could cause any problems.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -141,9 +141,9 @@ module.exports = {
 		},
 		'./packages/clay-multi-select/src/': {
 			branches: 62,
-			functions: 80,
-			lines: 78,
-			statements: 78,
+			functions: 77,
+			lines: 77,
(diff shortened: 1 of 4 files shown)


pat270:
The CSS LGTM!

matuzalemsteles:
cc @marcoscv-work @ethib137

matuzalemsteles:
Going forward with that and merge. I think from our conversations we want the visual focus on the label but keeping the accessible focus on the text.

PR #5262, opened 2022-12-20
fix(@clayui/css): Mixins clay-select-variant `$option` should be scop…

### LINKED ISSUE #5261: @clayui/css: Dual List Box spacing between options is incorrect
select-multi

This bug was probably introduced by https://github.com/liferay/clay/pull/5246. It used the `$option` variable name inside active and show states. The `$option` variable isn't scoped inside those because it's declared at the top.

pat270:
…ed to the selector

fixes #5261

fixed

### DIFF
--- a/packages/clay-css/src/scss/mixins/_forms.scss
+++ b/packages/clay-css/src/scss/mixins/_forms.scss
@@ -866,8 +866,6 @@
 			)
 		);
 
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM, thanks @pat270

PR #5265, opened 2022-12-20
fix(@clayui/core): computes nested collections to mount the layout

### LINKED ISSUE #5260: Picker component is not working when having a group inside it
https://codesandbox.io/s/angry-resonance-xuvkhd?file=/src/index.js

When using a Picker component that have groups inside there are some things that doesn't work

 - Keyboard navigation doesn't work, the user cannot move inside the dropdown with the arrow keys
 - If the values of the items are raw numbers, there is an error when trying to focus the element (it seems that it uses the value as the id of the item and it is not a valid id)
 - selectedKey doesn't seems to find the correct item for the value

All the above things to work, like when we have a flat list

I have taken the example from the PR where the functionality was added, maybe there is another way to do this

matuzalemsteles:
Fixes #5260

Now this works better with nested collection, it assembles nested collections even if the Menu is not rendered, there are still some improvements we can do to optimize, for example cache the nested collection result and reuse when React renders the component, for now this just builds the structure of the layout. I'll work on that later.

Also fix the error of the `key` being defined with numbers, let's find for the item by the global `id`, so the `key` must be unique in the entire document so we also respect the rule for the `id` attribute.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -33,6 +33,11 @@ const config = {
 		storyStoreV7: true,
 		postcss: false,
 	},
+	webpackFinal: (config) => {
+		config.resolve.mainFields = ['ts:main', ...config.resolve.mainFields];
+
+		return config;
+	},
 };
 
 module.exports = config;
--- a/packages/clay-core/src/collection/useCollection.tsx
+++ b/packages/clay-core/src/collection/useCollection.tsx
@@ -20,6 +20,8 @@ type CollectionContextProps = {
 
 const CollectionContext = React.createContext({} as CollectionContextProps);
 
+const SECTION_NAMES = ['Group', 'Section'];
+
 export function useCollection<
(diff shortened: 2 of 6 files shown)


matuzalemsteles:
Hey @victorg1991 this should fix the use cases for Group in the Picker, there is a case that is not yet supported when `native` is enabled it won't be able to convert the component from `DropDown.Group` to a native type of select, so don't enable that for now. I have some ideas in mind to make this work OOTB I will create an issue to work on it later but it shouldn't be a blocker for you right now.

victorg1991:
So fast! Thanks @matuzalemsteles ! Could you cut a release for this so I can test this out tomorrow? :)

matuzalemsteles:
@victorg1991 yes, is cutting a new release now https://github.com/liferay/clay/actions/runs/3744382358

PR #5272, opened 2022-12-21
fix(@clayui/pagination): aria-label should only be added to the nav e…

### LINKED ISSUE #5269: @clayui/pagination: ClayPaginationWithBasicItems duplicates aria-label on the ul
When passing `aria-label` to` ClayPaginationWithBasicItems`, it's added to the `nav` and the `ul` element. It should only be on the `nav`.

duplicate

pat270:
…lement

fixes #5269

### DIFF
--- a/packages/clay-pagination/src/Pagination.tsx
+++ b/packages/clay-pagination/src/Pagination.tsx
@@ -18,10 +18,13 @@ interface IProps extends React.HTMLAttributes<HTMLUListElement> {
 
 const ClayPagination = React.forwardRef<HTMLUListElement, IProps>(
 	({children, className, size, ...otherProps}: IProps, ref) => {
+		const ariaLabel = otherProps['aria-label'] || 'Pagination';
+
 		return (
-			<nav aria-label={otherProps['aria-label'] || 'Pagination'}>
+			<nav aria-label={ariaLabel}>
 				<ul
 					{...otherProps}
+					aria-label={undefined}
 					className={classNames(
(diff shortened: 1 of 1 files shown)


packages/clay-pagination/src/Pagination.tsx
```diff
@@ -18,10 +18,13 @@ interface IProps extends React.HTMLAttributes<HTMLUListElement> {
...
 const ClayPagination = React.forwardRef<HTMLUListElement, IProps>(
 	({children, className, size, ...otherProps}: IProps, ref) => {
+		const ariaLabel = otherProps['aria-label'] || 'Pagination';
+
 		return (
-			<nav aria-label={otherProps['aria-label'] || 'Pagination'}>
+			<nav aria-label={ariaLabel}>
 				<ul
 					{...otherProps}
+					aria-label={undefined}
```
matuzalemsteles, line 27: @pat270 considering @marcoscv-work's comment I think the best thing here would be to move the `aria-label` to the `ul`, I did a quick test with VoiceOver and really when focusing on the links the `aria-label` in the `nav ` will not be announced.

pat270: @matuzalemsteles I didn't understand the comment, but I do now. I think we should leave it on both, as @marcoscv-work said. When I use `ctrl + option + arrow` to navigate, it just announces navigation when it focus the `nav`. It's more descriptive if Pagination is there.

PR #5280, opened 2022-12-30
fix(@clayui/pagination-bar): PaginationBarWithBasicItems shouldn't re…

### LINKED ISSUE #5279: @clayui/pagination: ClayPaginationBarWithBasicItems errors if totalItems=0
https://clayui.com/docs/components/pagination-bar.html change the `totalPages` value to `{0}`. There is an error:
```
TypeError: Cannot read properties of undefined (reading 'toString')
```

Related:
https://issues.liferay.com/browse/LPS-171906
https://issues.liferay.com/browse/COMMERCE-10546

pat270:
…nder if totalItems is 0

fixes #5279

### DIFF
--- a/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
+++ b/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
@@ -175,6 +175,10 @@ export const ClayPaginationBarWithBasicItems = ({
 	totalItems,
 	...otherProps
 }: IProps) => {
+	if (totalItems === 0) {
+		totalItems = 1;
+	}
+
 	const [internalActive, setActive] = useInternalState({
 		defaultName: 'defaultActive',
 		defaultValue: defaultActive,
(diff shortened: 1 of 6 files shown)


marcoscv-work:
Tested and working!

Do we need to fix the tests to merge, right?

victoroscuro:
ci:test:commerce

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -218,49 +218,53 @@ export const ClayPaginationBarWithBasicItems = ({
 	}, [totalPages]);
 
 	return (
```
matuzalemsteles, line 220: I don't like this approach of not rendering the element internally it can cause a lot of confusion, I think what we can do here is if the `totalItems` is `0` we change it to `1` which would be the minimum acceptable.

This change can be done in Pagination as well because it will have the same error if only it is used.

matuzalemsteles:
Hey @pat270 I went ahead and updated the PR to still render the component but if `totalPages` in the `ClayPagination` component and `totalItems` in the `ClayPaginationBar` component are set to zero it internally changes the value to 1.

If the pagination in the use case of the DXP they don't need to show the pagination as in the case of a filter they should make this conditional but in the case of a common listing where there are only a few items on only one page, will still render the component.

PR #5288, opened 2023-01-09
feat(@clayui/css): Links adds c-link to be used with utility classes

### LINKED ISSUE #5283: Links must ensure contrast / underlining to be distinguishable
## Reference information:

**WCAG 2.1** (updated)
https://www.w3.org/WAI/WCAG21/Understanding/use-of-color
https://www.w3.org/WAI/WCAG21/Techniques/failures/F73
**WCAG 2.0** (older)
https://www.w3.org/TR/WCAG20-TECHS/F73.html
https://www.w3.org/TR/WCAG20-TECHS/G183.html

All links in Clay should be reviwed to follow the new Lexicon pattern:

https://www.figma.com/file/r0oZ3ORLODdHmvvcyCASoQ/Links?node-id=1%3A56&t=C7qv8XUFgWbgUbRk-1
image

**How to test it**:
Check that each link in the page that is identifiable by color (hue) is visually identifiable via some other means (e.g., underlined, bolded, italicized, sufficient difference in lightness, etc).

Since in Liferay is really difficult to ensure the sourounded color because in admin controls we have really heavy text contrast we hacve decided to provide underline in all non-component contextual content.

Potentially it could be solved with general CSS, but probably better if it is fixed in components like clay-link.

There is a task in DXP to cover all what Clay won't be able to style:
https://issues.liferay.com/browse/LPS-172123

pat270:
fixes #5283 

Update: I made most of the changes in this update. I changed `underline` to `decoration` because the underline isn't always set by ClayLink. We can have underlined text coming from CSS and the utility `text-decoration-none` will undo that and vice versa. The attribute `decoration` also has a `null` option to undo the presets in `primary` for example. We can use `decoration={null}` to prevent the primary link from outputting the underline utility.

This one might be a little confusing. I decided to adopt the utility classes since there were so many variants, but didn't want to break the api by changing the `link-primary` and `link-secondary` classes to use text utilities. I also added the base `c-link` class that should be used in every link.

|Regular Links|Classes|
|---|---|
|primary|c-link link-primary text-underline text-4|
|secondary|c-link link-secondary text-underline text-4|
|danger|c-link text-danger text-underline text-4|
|tertiary|c-link text-tertiary text-4|

|Semi Bold Links|Classes|
|---|---|
|primary|c-link link-primary font-weight-semi-bold text-underline text-4|
|secondary|c-link link-secondary font-weight-semi-bold text-underline text-4|
|tertiary|c-link text-tertiary font-weight-semi-bold text-underline text-4|
|danger|c-link text-danger font-weight-semi-bold text-underline text-4|

|With Icon|Classes|
|---|---|
|primary|c-link link-primary font-weight-semi-bold text-4|
|secondary|c-link link-secondary font-weight-semi-bold text-4|
|tertiary|c-link text-tertiary font-weight-semi-bold text-4|
|danger|c-link text-danger font-weight-semi-bold text-4|

These property settings are kind of hard to remember, we should probably bundle it in an easy to remember way.

### DIFF
--- a/packages/clay-link/src/index.tsx
+++ b/packages/clay-link/src/index.tsx
@@ -8,6 +8,34 @@ import React from 'react';
 
 import ClayLinkContext from './Context';
 
+const LINK_PRESETS = {
+	danger: {
+		base: 'c-link text-danger',
+		decoration: 'underline',
+	},
+	primary: {
+		base: 'c-link link-primary',
+		decoration: 'underline',
+	},
+	secondary: {
+		base: 'c-link link-secondary',
+		decoration: 'underline',
+	},
+	tertiary: {
+		base: 'c-link text-tertiary',
+		decoration: null,
+	},
+	unstyled: {
+		base: 'link-unstyled',
+		decoration: null,
+	},
+} as const;
+
+const FONT_WEIGHTS = {
+	normal: 'font-weight-normal',
+	'semi-bold': 'font-weight-semi-bold',
+};
+
 interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
 	/**
 	 * Renders the button as a block element.
@@ -31,24 +59,39 @@ interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
 		  };
 
 	/**
-	 * Determines how the link is displayed.
+	 * Indicates if the text should be underlined
 	 */
-	displayType?: 'primary' | 'secondary' | 'unstyled';
+	decoration?: 'none' | 'underline' | null;
 
 	/**
-	 * Flag to indicate if link should be monospaced.
+	 * Determines how the link is displayed.
+	 */
+	displayType?: 'danger' | 'primary' | 'secondary' | 'tertiary' | 'unstyled';
+
+	/**
+	 * Sets the text size based on a number scale.
+	 */
+	fontSize?: Number;
+
+	/**
+	 * Flag to indicate if the link should be monospaced.
 	 */
 	monospaced?: boolean;
 
 	/**
-	 * Flag to indicate if link need have an outline.
+	 * Flag to indicate if the link should use the outlined style.
 	 */
 	outline?: boolean;
 
 	/**
-	 * Indicates button should be a small variant.
+	 * Indicates whether the button should use the small variant.
 	 */
 	small?: boolean;
+
+	/**
+	 * Determines the font-weight of the link.
+	 */
+	weight?: 'normal' | 'semi-bold';
 }
 
 const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
@@ -59,12 +102,15 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
 			button,
 			children,
 			className,
+			decoration,
 			displayType,
+			fontSize,
 			monospaced,
 			outline,
 			rel,
 			small,
 			target,
+			weight,
 			...otherProps
 		}: IProps,
 		ref
@@ -85,14 +131,28 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
 				[`btn-${displayType}`]: displayType && !outline && !borderless,
 				[`btn-outline-${displayType}`]:
 					displayType && (outline || borderless),
+				[weight ? FONT_WEIGHTS[weight] : '']: weight,
+				[`text-${fontSize}`]: fontSize,
+				[`text-decoration-${decoration}`]: decoration,
 			};
 		} else {
+			if (displayType && LINK_PRESETS[displayType]) {
+				if (decoration !== null && !outline) {
+					decoration =
+						decoration || LINK_PRESETS[displayType]?.decoration;
+				}
+			}
+
 			classes = {
 				'link-monospaced': monospaced,
 				'link-outline': outline,
 				'link-outline-borderless': borderless,
-				[`link-${displayType}`]: displayType && !outline,
+				[displayType ? LINK_PRESETS[displayType]?.base : '']:
+					displayType && !outline,
 				[`link-outline-${displayType}`]: displayType && outline,
+				[weight ? FONT_WEIGHTS[weight] : '']: weight,
+				[`text-${fontSize}`]: fontSize,
+				[`text-decoration-${decoration}`]: decoration,
 			};
 		}
 
--- a/packages/clay-link/src/index.tsx
+++ b/packages/clay-link/src/index.tsx
@@ -131,26 +131,23 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
 				[`btn-${displayType}`]: displayType && !outline && !borderless,
 				[`btn-outline-${displayType}`]:
 					displayType && (outline || borderless),
-				[weight ? FONT_WEIGHTS[weight] : '']: weight,
+				[FONT_WEIGHTS[weight!]]: weight,
 				[`text-${fontSize}`]: fontSize,
 				[`text-decoration-${decoration}`]: decoration,
 			};
 		} else {
-			if (displayType && LINK_PRESETS[displayType]) {
-				if (decoration !== null && !outline) {
-					decoration =
-						decoration || LINK_PRESETS[displayType]?.decoration;
-				}
-			}
+			decoration =
+				decoration === null || outline
+					? undefined
+					: decoration || LINK_PRESETS[displayType!]?.decoration;
 
 			classes = {
 				'link-monospaced': monospaced,
 				'link-outline': outline,
 				'link-outline-borderless': borderless,
-				[displayType ? LINK_PRESETS[displayType]?.base : '']:
-					displayType && !outline,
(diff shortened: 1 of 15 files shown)


packages/clay-link/src/index.tsx
```diff
@@ -85,14 +116,21 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
...
 			classes = {
 				'link-monospaced': monospaced,
 				'link-outline': outline,
 				'link-outline-borderless': borderless,
-				[`link-${displayType}`]: displayType && !outline,
+				[displayType ? linkDisplayTypes[displayType] : '']:
+					displayType && !outline,
 				[`link-outline-${displayType}`]: displayType && outline,
+				[thickness ? fontWeights[thickness] : '']: thickness,
+				[`text-${fontSize}`]: fontSize,
```
matuzalemsteles, line 132: I think we could use the `Text` component here instead of declaring the class. This helps in the future when we have to make changes, the more we reuse the components the better.

packages/clay-link/src/index.tsx
```diff
@@ -33,22 +33,37 @@ interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
...
+
+	/**
+	 * Determines the font-weight of the link.
+	 */
+	thickness?: 'normal' | 'semi-bold';
+
+	/**
+	 * Indicates if the text should be underlined
+	 */
+	underline?: boolean;
```
matuzalemsteles, line 66: I think the underline will be required for all non-button links.

packages/clay-link/src/index.tsx
```diff
@@ -33,22 +33,37 @@ interface IProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
...
 	/**
-	 * Indicates button should be a small variant.
+	 * Indicates whether the button should use the small variant.
 	 */
 	small?: boolean;
+
+	/**
+	 * Determines the font-weight of the link.
+	 */
+	thickness?: 'normal' | 'semi-bold';
```
matuzalemsteles, line 61: Maybe `weight` is better instead of `thickness` which would be more common.

matuzalemsteles (commented):
I like this idea of going with utilities it's much better for handling different compositions. I just left a few comments.

marcoscv-work:
Maybe I'm wrong because I'm not deploying the PR and make a real review, but based on commits adding specific classes... do we need to use a class to make a link looks like a link?
I mean, the defaul link in clay (even without classes) should look like a link, primary blue and underline.

pat270:
@marcoscv-work We've had default links `text-decoration: none` for a long time. If we change the default to be underlined, it will cascade to many components where underline might not be wanted (nav, dropdown, pagination, etc). I feel like this might be "breaking" for existing users. Do we still want to make this change?

packages/clay-link/src/index.tsx
pat270: @matuzalemsteles it doesn't seem like it will always be required according to https://www.figma.com/file/r0oZ3ORLODdHmvvcyCASoQ/Links?node-id=1%3A56&t=C7qv8XUFgWbgUbRk-1. What if we default to true?

packages/clay-link/src/index.tsx
pat270: I've been looking at this and it looks like we would get some markup like:
```html
<span class="text-4">
    <a class="c-link link-primary" href="">Primary</a>
</span>
```

or

```html
<a class="c-link link-primary" href="">
    <span class="text-4">Primary</span>
</a>
```

It makes more sense to me if we can use `ClayLink` as a wrapper for `ClayText` where we can apply the `a` tag, `href`, `tabindex`, or whatever to `ClayText`. We won't have the extra span tag and the classes will all be on the `a` element.

packages/clay-link/src/index.tsx
matuzalemsteles: Yeah, I think true by default is a good idea, `underline` should add to all links but except for some use cases. Figma's example is for some cases that won't need an underline, I think this Slack thread has more information about that https://liferay.slack.com/archives/C04A7NLM7L1/p1672744222205579

packages/clay-link/src/index.tsx
matuzalemsteles: Oh yes that's true, it makes sense to me too I think we can work on that later and add an API of `as={Link}` in the Text. For now, let's go as it is.

marcoscv-work:
> @marcoscv-work We've had default links `text-decoration: none` for a long time. If we change the default to be underlined, it will cascade to many components where underline might not be wanted (nav, dropdown, pagination, etc). I feel like this might be "breaking" for existing users. Do we still want to make this change?

Sounds very strong, yeah.. we don't want propagation to all components, it should be something controlled, so let's discuss this better and merge the PR by now

cc @emiliano-cicero

pat270:
Thanks @marcoscv-work ! I'm still making some changes to this pr, but should be done soon.

marcoscv-work:
just let me know to review! or just make an auto-merge ;-) !

marcoscv-work (approved):

packages/clay-link/src/index.tsx
```diff
@@ -85,14 +131,28 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
...
+
 			classes = {
 				'link-monospaced': monospaced,
 				'link-outline': outline,
 				'link-outline-borderless': borderless,
-				[`link-${displayType}`]: displayType && !outline,
+				[displayType ? LINK_PRESETS[displayType]?.base : '']:
+					displayType && !outline,
 				[`link-outline-${displayType}`]: displayType && outline,
+				[weight ? FONT_WEIGHTS[weight] : '']: weight,
```
matuzalemsteles, line 153: I think the ternary would not be needed here because it will already only render if `weight` is set.

```suggestion
				[FONT_WEIGHTS[weight!]]: weight,
```

packages/clay-link/src/index.tsx
```diff
@@ -85,14 +131,28 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
...
+						decoration || LINK_PRESETS[displayType]?.decoration;
+				}
+			}
+
 			classes = {
 				'link-monospaced': monospaced,
 				'link-outline': outline,
 				'link-outline-borderless': borderless,
-				[`link-${displayType}`]: displayType && !outline,
+				[displayType ? LINK_PRESETS[displayType]?.base : '']:
```
matuzalemsteles, line 150: Same here

```suggestion
				[LINK_PRESETS[displayType!]?.base]:
```

packages/clay-link/src/index.tsx
```diff
@@ -85,14 +131,28 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
...
+				[`text-${fontSize}`]: fontSize,
+				[`text-decoration-${decoration}`]: decoration,
 			};
 		} else {
+			if (displayType && LINK_PRESETS[displayType]) {
+				if (decoration !== null && !outline) {
+					decoration =
+						decoration || LINK_PRESETS[displayType]?.decoration;
+				}
+			}
```
matuzalemsteles, line 144: I think we can make this a little simpler.

```suggestion
		decoration = decoration === null || outline ? undefined : decoration || LINK_PRESETS[displayType!]?.decoration;
```

We must rely on typing here so we don't have to do checks that wouldn't be expected because when the developer tries to compile the build will fail. Also using `?` and trying to access the object with an `undefined` value will have the same result as the above conditionals which are checking if the value exists in `LINKS_PRESETS` and if the `displayType` exists.

matuzalemsteles (commented):
I just added some small comments for adjustment.

packages/clay-link/src/index.tsx
```diff
@@ -59,12 +102,15 @@ const ClayLink = React.forwardRef<HTMLAnchorElement, IProps>(
 			button,
 			children,
 			className,
+			decoration,
 			displayType,
```
matuzalemsteles, line 106: Maybe we have to set the default value of displayType to `primary` to use the presets when nothing is defined?

packages/clay-link/src/index.tsx
pat270: Thanks for this, I was wondering how to get around typescript's complaints.

packages/clay-link/src/index.tsx
pat270: @matuzalemsteles I think it's better not to have a default. We have components like navigation bar and pagination that rely on `ClayLink`. If we default to primary, we will have to go back and undo them in the components. Also from a user's perspective I would expect `ClayLink` to output a standard link if I don't specify a `displayType`.

matuzalemsteles: Hmm yes, this can cause unwanted issues in other components where this is not expected. My concern is that this change is an update to fix an accessibility issue, ideally anyone using the component would not need to update or set a property to "get the fix".

But it really seems like a safer way to go that way without setting the `displayType` by default.

matuzalemsteles (approved):
That's good to go now.

PR #5290, opened 2023-01-10
fix(@clayui/pagination): Adds tabindex to links with no href

### LINKED ISSUE #5286: Pagination: Empty links
In pagination examples many links can be generated as emptly links without `href`, it is not correct since they cannot get focus.
For example, we could be adding `aria-current="page"` but if the link is not tabbaulable we cann know it is the current and selected page by screen reader.

**Recommendation**:
- Make href mandatory for links
or
- Add a warning if you are creatign an emptly link

pat270:
fixes #5286

@marcoscv-work I added `tabindex="0"` to links that are active or don't have a `href` attribute. It should receive focus when tabbing now. Disabled links don't receive focus when tabbing. I think this is a better way than making `href` mandatory. We don't need to make up a placeholder url that may or may not make sense.

### DIFF
--- a/packages/clay-css/src/scss/variables/_pagination.scss
+++ b/packages/clay-css/src/scss/variables/_pagination.scss
@@ -57,7 +57,7 @@ $pagination-link-border-radius: 0px !default;
 
 /// @deprecated as of v2 use the Sass map `$pagination-link` instead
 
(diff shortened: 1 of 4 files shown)


marcoscv-work:
Tested and make sense for me, thanks @pat270

PR #5298, opened 2023-01-16
fix(@clayui/core): fix Vertical Bar component error change behavior from controlled to uncontrolled

### LINKED ISSUE #5292: Controlled VerticalBar: animation for panels don't work?
## What is your question?
Before opening a bug, I prefer to ask a question in case I am doing something wrong. 

I am using a controlled VerticalBar component, but there is no animation when the active panel changes, and the div dissapears inmediatelly (because the [unmountOnExit](https://github.com/ambrinchaudhary/clay/blame/master/packages/clay-core/src/vertical-bar/Panel.tsx#L81) is set) 

https://user-images.githubusercontent.com/8373764/211804180-dae40955-fea6-4f0d-9b40-aa664cc729d6.mp4

This does not happen if I change it to uncrontrolled VerticalBar, but this is not an option in our case. 

Could there be an issue with the timeout parameter [here](https://github.com/ambrinchaudhary/clay/blame/master/packages/clay-core/src/vertical-bar/Panel.tsx#L79-L79)? I guess it is returning 0, because I do not see any of this classes in the dom element (but I see some of them if I make it uncontrolled)
```
                        classNames={{
				enter: 'c-slideout-transition c-slideout-transition-in',
				enterActive: 'c-slideout-show',
				enterDone: 'c-slideout-show',
				exit: 'c-slideout-transition c-slideout-transition-out',
			}}
```

Anyway, it would be great to be able to configure this timeout parameters. In our case in KB we need to set different animation durations (see [prototype](https://www.figma.com/proto/wbAfx5biJhZaIFuuK0PiA5/Improve-KB-navigation?page-id=0%3A1&node-id=96%3A85681&viewport=-65%2C-504%2C0.35&scaling=min-zoom&starting-point-node-id=79%3A81073&show-proto-sidebar=1)) and know this is not possible, so I must set KB animations same like the VerticalBar (`{enter: 300, exit: 200}` I guess) to not have strange movements

matuzalemsteles:
Fixes #5292

The issue with the animation not working correctly was that when the component was configured to be controlled, the state changed from controlled to uncontrolled which broke the implementation internally.

This PR fixes this so that this can be `React.Key` or `null` instead of `undefined` which is the type that `useInternalState` uses to determine controlled or uncontrolled behavior.

### DIFF
--- a/packages/clay-core/src/vertical-bar/Item.tsx
+++ b/packages/clay-core/src/vertical-bar/Item.tsx
@@ -77,14 +77,12 @@ export const Item = React.forwardRef<HTMLLIElement, Props>(function Item(
 						return;
 					}
 
-					onActivePanel(
-						keyValue === activePanel ? undefined : keyValue!
-					);
+					onActivePanel(keyValue === activePanel ? null : keyValue!);
 				},
 				role: 'tab',
 				tabIndex:
-					(activePanel !== undefined && activePanel !== keyValue) ||
-					(activePanel === undefined && index !== 0)
+					(activePanel !== null && activePanel !== keyValue) ||
+					(activePanel === null && index !== 0)
 						? -1
 						: null,
 			})}
--- a/packages/clay-core/src/vertical-bar/Panel.tsx
+++ b/packages/clay-core/src/vertical-bar/Panel.tsx
@@ -46,7 +46,7 @@ export function Panel({children, keyValue, tabIndex}: Props) {
 
 	const isFirst = useIsFirstRender();
 
-	const previousActivePanelRef = useRef<React.Key | undefined>(undefined);
+	const previousActivePanelRef = useRef<React.Key | null>(null);
 
 	useEffect(() => {
 		previousActivePanelRef.current = activePanel;
@@ -71,10 +71,10 @@ export function Panel({children, keyValue, tabIndex}: Props) {
 			role="tabpanel"
 			tabIndex={tabIndex}
 			timeout={
-				(activePanel !== undefined &&
-					previousActivePanelRef.current === undefined) ||
-				(activePanel === undefined &&
-					previousActivePanelRef.current !== undefined)
+				(activePanel !== null &&
+					previousActivePanelRef.current === null) ||
+				(activePanel === null &&
+					previousActivePanelRef.current !== null)
 					? {enter: 300, exit: 200}
 					: 0
 			}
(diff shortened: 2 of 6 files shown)


pat270:
@matuzalemsteles I noticed in the KB vertical bar, it is setting `active` and `onActiveChange` on `VerticalBar`. When I set the `active` and `onActiveChange` attributes in storybooks, the panels don't open.

I have something that "works", but I need to clean it up a bit. I also wanted to discuss if it's possible to get the exit animations to run when closing a panel and opening another panel. I researched CSSTransition, but couldn't find many examples. I'll write more detail in the draft pr I send.

matuzalemsteles:
> I noticed in the KB vertical bar, it is setting active and onActiveChange on VerticalBar. When I set the active and onActiveChange attributes in storybooks, the panels don't open.

Yeah, this PR should fix that. This is because of the behavior change from controlled to uncontrolled which broke animation.

> I also wanted to discuss if it's possible to get the exit animations to run when closing a panel and opening another panel.

What would the animation look like? for now the component will only animate when the panel closes and not open another one, it is intentional not to animate when you just switch tabs.

pat270:
When I first saw the Lexicon proposal, it was animate in and out even when changing tabs. I don't think the spec changed, but my knowledge could be out of date.

I tested this pr in storybooks and the panels in DynamicContent don't open when you update the example with https://github.com/liferay/clay/pull/5300/commits/c57352347233a7397e0ed0bb26d822e904cdef52.

matuzalemsteles:
>When I first saw the Lexicon proposal, it was animate in and out even when changing tabs. I don't think the spec changed, but my knowledge could be out of date.

Hmm I don't remember about that, at least I don't remember if it was explicit in the documentation but I think we kept the same animation that exists in the DXP that had already implemented the VerticalBar but this component should have visual changes soon but I think that it's not prioritized right now, at least when we implemented this component that was the conversation.

> I tested this pr in storybooks and the panels in DynamicContent don't open when you update the example with https://github.com/liferay/clay/commit/c57352347233a7397e0ed0bb26d822e904cdef52.

I can get this to work with the controlled state. I submit a commit updating the example to use the controlled state now.

- https://deploy-preview-5298--next-storybook-clayui.netlify.app/?path=/story/design-system-components-verticalbar--dynamic-content

pat270:
LGTM.

PR #5300, opened 2023-01-17
fix(@clayui/core): VerticalBar should animate the panel open / close in a controlled component

pat270:
#5292

@matuzalemsteles here is the draft pr I mentioned in #5298. I couldn't animate open and close when switching tabs because CSSTransition runs both animations simultaneously. I don't know if there's a way to execute the close animation before running open. I couldn't find anything on it.

I tried [SwitchTransition](http://reactcommunity.org/react-transition-group/switch-transition) but it wasn't what I was expecting.


matuzalemsteles:
Hm I see I think the `SwitchTransition` must be what you are looking for but the problem is probably that you need to refactor the implementation of the panels, for example the container of the panel it should always be visible and we just have to change its content .

PR #5301, opened 2023-01-17
fix(@clayui/css): Atlas Pagination active state should have more cont…

### LINKED ISSUE #5296: Pagination active button contrast needs to be updated
1. Open [this page](https://clayui.com/docs/components/pagination.html)
2. take a look at the active button of the numbered pagination (the 8)

There is a contrast requirement for this active state to be at least 3:1

## Lexicon solution
We have the [LEXI-1287](https://issues.liferay.com/browse/LEXI-1287) where we provide a visual solution inverting the colors of the button, this will be applied as an exception to this case (specifically for the pagination), not all the borderless buttons. If there are other cases that need this, we'll review everything. 

This also fixes the [LPS-169083](https://issues.liferay.com/browse/LPS-169083)

## Environment
Clay & Master

pat270:
…rast

fixes #5296

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_pagination.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_pagination.scss
@@ -29,15 +29,15 @@ $pagination-color: $gray-600 !default;
 $pagination-padding-x: 0.625rem !default; // 10px
 $pagination-padding-y: 0 !default;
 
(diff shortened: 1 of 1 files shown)


marcoscv-work:
Tested an working fine! @pat270 thanks! :-D

PR #5309, opened 2023-01-19
feat: add prop to remove `aria-current` from `VerticalNav` and `NavigationBar` component

### LINKED ISSUE #5307: Add option to disable aria-current in navigation components
## What is your proposal?
My proposal is adding the possibility to disable `aria-current` attribute in navigation components, such as `ClayNavigationBar` and `ClayVerticalNav`. The reason is that sometimes, we have two different navigations in a page, for example we can have a `ClayNavigationBar` and then a `ClayVerticalNav`. As it's incorrect to have two `aria-current="page"` in the same page, it would be great to have the option of disabling it in one of the navigations.

Because we would have a way to avoid the problem of having two different `aria-current="page"`.

matuzalemsteles:
Fixes #5307

This PR adds to the API the possibility of disabling `aria-current` from the `VerticalNav` and `NavigationBar` components when setting `aria-current` to `null`.

```jsx
<ClayNavigationBar>
  <ClayNavigationBar.Item active aria-current={null}>
    <ClayLink href="#">Item 1</ClayLink>
  </ClayNavigationBar.Item>
</ClayNavigationBar>

<ClayVerticalNav
  aria-current={null}
/>
```

I'm not sure if this is the behavior we want when disabling `aria-current={null}` we add `aria-selected` instead, but since they are links I'm not sure if this would be the right behavior. @marcoscv-work could you confirm?

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -152,7 +152,7 @@ module.exports = {
 			statements: 98,
 		},
 		'./packages/clay-nav/src/': {
-			branches: 90,
+			branches: 88,
 			functions: 80,
 			lines: 93,
 			statements: 93,
--- a/packages/clay-nav/src/Vertical.tsx
+++ b/packages/clay-nav/src/Vertical.tsx
@@ -49,6 +49,12 @@ interface IItemWithItems extends IItem {
 
 export interface IProps {
 	/**
+	 * Flag to define if the item represents the current page. Disable this
+	 * attribute only if there are multiple navigations on the page.
+	 */
+	'aria-current'?: 'page' | null;
+
+	/**
 	 * Flag to indicate the navigation behavior in the menu.
 	 *
 	 * - manual - it will just move the focus and menu activation is done just
@@ -94,7 +100,9 @@ export interface IProps {
 	spritemap?: string;
 }
 
-export interface INavItemProps extends IItemWithItems {
+export interface INavItemProps extends Omit<IItemWithItems, 'aria-current'> {
+	'aria-current'?: 'page' | null;
+
 	/**
 	 * Integer to keep track of what nested level the item is.
 	 */
@@ -130,6 +138,7 @@ function findSelectedNested(
 
 function Item({
 	active,
+	'aria-current': ariaCurrent = 'page',
 	href,
 	initialExpanded = false,
 	items,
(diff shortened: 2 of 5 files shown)


marcoscv-work:
I think we where thinking more about a global prop for the component, this is the real reason we need this:

image


As you can see, the current page is "information". And only vertical-nav should be contributing aria-current, for this case we should say to navigation-bar, "hey, please stop contibuting aria-current". so in this format it should be global prop, not being worries about wich link is active in navigation-bar, another example:

image image
In the first image I would need to add the prop to the second item "contanct" and in the second image I would need to add it to "Preferences" in navigation-bar. For me it is more probably to make a mistake if you need adding the individual item prop instead of global component prop.

Make this sense?

marcoscv-work:
`aria-selected` is just for selections in panels, for example for [tabs](https://clayui.com/docs/components/tabs.html) or [vertical-bar](https://clayui.com/docs/components/vertical-bar.html)

sandrodw3:
Thanks for your fast work @matuzalemsteles! 😄 I agree with Marcos that it would be easier to set the property to the `ClayNavigationBar` itself, instead of setting it in all items, I think when you want to disable it, you want it for all items, so it would make sense.

matuzalemsteles:
I'm not sure I got the idea right but do we want a global prop outside of ClayNavigationBar and VerticalNav? If that's the thought, unfortunately we can't do this OOTB on the Clay side, because in DXP most of these components are rendered with different React trees.

As for moving the `aria-current` property of the ClayNavigationBar item to the root component, the problem is that the component has no visibility of the items, yet we managed to do that anyway.

I think this wouldn't be a big problem because normally in DXP, you only declare one item with iteration, for example:

```jsx
<ClayNavigationBar>
  {items.map((item) => (
    <ClayNavigationBar.Item active={item.id === selectedId} aria-current={null}>
      <ClayLink href="#">Item 1</ClayLink>
    </ClayNavigationBar.Item>
  ))}
</ClayNavigationBar>
```

Anyway I'm going to try to move `aria-current` to root.

sandrodw3:
Hi @matuzalemsteles! I'm not sure but I think with "global prop" @marcoscv-work meant a prop in the root component. At least I meant that. What you say about the usages in DXP is right, now that I think. I think it would be more clean and simple in the root component, but if you have any impediment to that, we could go with the prop in the items. Thanks! 😄

matuzalemsteles:
@sandrodw3 makes sense, I ended up going ahead and adding API support in the root component in the `ClayNavigationBar` component. So I think we're good with that here, I'm going to go ahead and merge this PR.

PR #5311, opened 2023-01-19
fix(@clayui/multi-select): prevents items from having the same key

### LINKED ISSUE #5308: ClayMultiSelect - Extend the functionality of the `locator` to allow the same item to be selected from the DropDown list
## What is your proposal?

This problem/functionality is inspired by https://issues.liferay.com/browse/COMMERCE-10636 where the Commerce team wants the functionality to add more of the same product from the DropDown list.

Also related is [this Slack thread](https://liferay.slack.com/archives/C3JBR21HA/p1674123903930749).

As an alternative I have simply filtered out the already selected products, but this is just a solution that disallows adding 2 of the same items that produce a bug to happen where the React complains about items with the same id.

matuzalemsteles:
Closes #5308

This PR prevents the MultiSelect labels from having the same `key`, the `key` cannot use the `index` because it loses the element reference when the labels are ordered, so we create a sequential key to be uid.

Internally when adding new items, we create a new uid for the item and add the `key` property to the item, we use it for the key and the id, the developer can also change which value we will use by defining the key in `locator.id` prop by default is `key`.


matuzalemsteles:
Hey @kresimir-coko, now we have a `locator.id` but internally we are also generating a uid to avoid having this problem, so this should be OOTB but it also allows us to use another property of the item to be the key/uid.

kresimir-coko:
Great, thanks Matu

PR #5316, opened 2023-01-20
fix(@clayui/navigation-bar): Removes classes btn-unstyled and link-unstyled from nav-link

### LINKED ISSUE #5314: @clayui/navigation-bar: btn-unstyled and link-unstyled shouldn't be there
There are extra classes on `nav-link` that shouldn't be there. The classes are `btn-unstyled` and `link-unstyled`. These classes have been there for a while so this might be a breaking change. If it's not that bad, we should try and remove these.

navigation-bar-classes

pat270:
fixes #5314

@matuzalemsteles I think this should be relatively safe. I checked Portal and seems like the `btn-unstyled` and `link-unstyled` for Navigation Bar aren't used.

### DIFF
--- a/packages/clay-css/src/scss/cadmin/variables/_navigation-bar.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_navigation-bar.scss
@@ -31,6 +31,7 @@ $cadmin-navigation-bar-base: map-deep-merge(
 		navbar-nav: (
 			nav-link: (
 				border-radius: 0,
+				border-width: 0,
 				outline: 0,
 				font-size: inherit,
 				focus: (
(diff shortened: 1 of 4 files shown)


matuzalemsteles (commented):
Great! @pat270 I just noticed a little visual regression when the visual is dark.

| Before  | After  |
|---|---|
|Screen Shot 2023-01-20 at 6 49 04 PM|Screen Shot 2023-01-20 at 6 47 02 PM|

pat270:
@matuzalemsteles Thanks for catching this!

matuzalemsteles (approved):
LGTM

PR #5317, opened 2023-01-21
feat(@clayui/core): add API to disable visual state of indeterminate and change selectionMode on manual selection

### LINKED ISSUE #5302: Add support for selecting/deselecting a bunch of items in ClayTreeView
## What is your proposal?
We have a ClayTreeView implementation with multiple selection without recursion, because children of an item can be all selected without necessarily being the parent (with multiple-recursive this can't be done). At the same time we would like to implement a custom command to select (or deselect) an item and all its descendants.

So it would be great to have a way to select or deselect a bunch of items at the same time.

It would help to have standard multiple selection but having a way to recursively select an item and all its descendants.

matuzalemsteles:
Closes #5302

This PR adds a new public API property of `selection.toggle` in render props that allows changing the selection mode on call without affecting the default behavior.

```jsx
<TreeView
  defaultItems={ITEMS_DRIVE}
  selectionMode="multiple"
>
  {(item, selection) => (
    <TreeView.Item onClick={() => selection.toggle(item.id, 'multiple-recursive')}>
      <Icon symbol="folder" />
      {item.name}
      {item.id}
    </TreeView.Item>
  )}
</TreeView>
```

We are also adding a new property to disable the visual state of `indeterminate` when recursive multiple selection is enabled.

```jsx
<TreeView
  defaultItems={ITEMS_DRIVE}
  selectionMode="multiple"
  indeterminate={false}
>
  {(item, selection) => ...}
</TreeView>
```


matuzalemsteles:
hey @sandrodw3 I changed the strategy a little bit to disable the visual state of `indeterminate` instead of calling in the API of `selection.toggle` you disable it by setting the property on `<TreeView />` this seems more secure because the state of `indeterminate` it can come back if the default selectionMode is with `multiple-recursive` in theory in your use case it wouldn't be a problem because the default would be `multiple` but I think it makes more sense to have a global API that allows disabling this behavior. Let me know what you think and if I missed anything.

sandrodw3:
Hey @matuzalemsteles! This is perfect, I also think it's better with the global API for the indeterminate. I would say this completely solves our use case. Thanks a lot! Waiting for it in portal 🥰

matuzalemsteles:
Thanks @sandrodw3, I'll go ahead and merge that PR.

PR #5323, opened 2023-01-26
fix(@clayui/date-picker): Date Range should be selectable with a sing…

### LINKED ISSUE #5318: Double click needed to edit a Date Range
If a user wants to change the range on the date picker they need to use double-click in order to change the first date of the range.

I've been testing with the example in the Clay documentation: https://clayui.com/docs/components/date-picker.html#range

https://user-images.githubusercontent.com/3276218/214044649-d17bd7e1-01f6-4c89-a618-85cd88a6806e.mp4

I should be able to edit the range with just one click.

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.84.0
| React |  16.9.9  |

pat270:
…le click

fixes #5318

I'm not sure why comparing the Date objects? didn't compare correctly even when they printed the same Date. Converting the dates to string seem to fix the problem.

I also found another bug that's easier to replicate with this pr. I'll file a separate issue for that.

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -438,7 +438,7 @@ const ClayDatePicker = React.forwardRef<HTMLInputElement, IProps>(
 			let daysSelectedToString;
 
 			if (range) {
-				if (startDate !== endDate) {
+				if (startDate.toString() !== endDate.toString()) {
 					newDaysSelected = [date, date];
 				} else if (date < startDate) {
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM!

Hmm this is strange I thought it would be a problem because the `daysSelected` are not normalized but they are, another question would be that the dates can be the same but the type comparison in this case of `!==` would be checking the object reference , then it may be that they are different references but with equal values. Anyway this is working great, good work Patrick!

PR #5329, opened 2023-01-27
fix(@clayui/css): Empty State add bigger margin-bottom on image

pat270:
fixes #5328 

This keeps the spacing between the text and image the same even if there is no title.


matuzalemsteles (approved):
LGTM

PR #5347, opened 2023-02-02
fix(@clayui/css): Atlas Dropdown items should be 32px tall

### LINKED ISSUE #5341: Update dropdown style to match the new picker
## What is your proposal?
The dropdown items have a total height of 37px, following the Lexicon grid definition, it should be either 32 or 40. 
With the new Picker, we decided to use 32px in order to allow having smaller items for cases where the panel has a lot number of elements.

The proposal is to update the dropdown style to match the new picker (custom select) and the Lexicon grid. This was shared and approved by Lexicon and DXP's product designers. 

These are the new specs for the dropdown:
image

Both panels will be consistent in text and height.

none

pat270:
fix(@clayui/css): Atlas Dropdown active state should be font-weight: normal

fixes #5341

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_dropdowns.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_dropdowns.scss
@@ -3,6 +3,7 @@
 // @deprecated as of v3.x use map $dropdown-menu instead
 
 $dropdown-bg: $white !default;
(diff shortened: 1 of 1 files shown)


matuzalemsteles (approved):
LGTM

PR #5353, opened 2023-02-04
feat(@clayui/core): adds initial implementation of drag and drop via keyboard in treeview

### LINKED ISSUE #5146: ClayTreeView: keyboard alternative for drag and drop
Right now we can drag and drop and it works great, but we don't have alternate keys so it can be considered not accessible.

**We would need:**

- A way to invoque the action of move (for example, use the dropdown menu to call 'move action', when pressed start the interaction). 
image

- Use arrow keys and enter to determinate the movement, example (only when the movement started):
https://schne324.github.io/dragon-drop/demo/#nested
image

matuzalemsteles:
Fixes #5146

This commit still needs some implementation refinements, we need to change the way we handle the element indicator to improve indicator announcer to SRs.

I also still need to move the announcer messages to be configured for internationalization.

### ToDo

- [x] Refactor the drop indicator implementation
- [x] Add new API to configure announcement messages

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -62,10 +62,10 @@ module.exports = {
 			statements: 91,
 		},
 		'./packages/clay-core/src/tree-view/': {
-			branches: 68,
-			functions: 73,
-			lines: 76,
-			statements: 75,
+			branches: 59,
+			functions: 70,
+			lines: 71,
+			statements: 71,
 		},
 		'./packages/clay-data-provider/src/': {
 			branches: 69,
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -34,6 +34,7 @@
 		"@clayui/provider": "^3.77.0",
 		"@clayui/shared": "^3.87.0",
 		"@tanstack/react-virtual": "3.0.0-beta.18",
+		"aria-hidden": "^1.2.2",
 		"classnames": "^2.2.6",
 		"fuzzy": "^0.1.3",
 		"react-dnd": "^11.1.1",
--- a/packages/clay-core/src/live-announcer/VisuallyHidden.tsx
+++ b/packages/clay-core/src/live-announcer/VisuallyHidden.tsx
@@ -19,14 +19,21 @@ const styles: React.CSSProperties = {
 };
 
 type Props = {
-	children: React.ReactNode;
+	children?: React.ReactNode;
 	liveAnnouncer?: boolean;
-};
+} & React.HTMLAttributes<HTMLDivElement>;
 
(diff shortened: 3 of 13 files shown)


matuzalemsteles:
I need to refine this implementation a lot, especially the accessibility part which, to be better, I need to refactor the indicator so that the accessibility part works, for that I will have more work on this component, I will have to remove it from this release so as not to block it and continue working on it this week.

matuzalemsteles:
I sent new commits to improve keyboard drag and drop accessibility.

When initiating the drag, we announce that the drag has started and provide some instructions, we also add `aria-hidden="true"` to all elements except the drop indicators and the drag element that locks the focus if tab is held down, ie it is possible to just move the focus to the drag element and the most recent drop target with, both with instructions for the actions.

Moving between the indicators we announce to the SR if it is a drop before or after or inside the item also with description of how to cancel the drop or finish the drop. The announcement is also made when the drop has been completed or cancelled.

There are still some details of the indicator not announcing correctly, I have to investigate the reason but it appears to be a VoiceOver bug.

matuzalemsteles:
I sent the final tweaks to that PR, it's now ready for review and moving forward. The messages are now being done correctly on start drag, cancel or completion. I also added the `messages` API that allows changing the message texts in case of internationalization.

This PR implements some details different from the issue, for example the drag preview is not shown because visually it would always be together with the indicator which is visually redundant, also for SR we don't need it because we use the indicator to announce the target, when the user is in a drag and drop via keyboard we suppress the elements of the DOM so that the focus is only on the current indicator and the drag button that provides the information of which element is in drag as it is said in the message when starting the drag.

Another detail is that the drag button is a drag icon instead of being an option within DropDown in an action, this seems better for the SR user and would also be very difficult to implement because the user is in drag when pressing tab we move the focus of the indicator to the element in drag in the case the button that triggers the drag, we do not move to the item because it will announce wrong, also when canceling the drop we move the focus back to the element the button of drag, doing this in a DropDown would become very complex and confusing for the SR user. I think it's worth discussing here whether it's worth leaving the icon on the right or left side of the item along with the action buttons.

Another detail is that we only show the drag button if the user is navigating with the keyboard or using the SR instead of always being visible when using the mouse.

Below I show a small demonstration using drag and drop via keyboard with different use cases.

https://user-images.githubusercontent.com/13750819/218605140-732dffea-be31-4630-a0c0-5e935d0eea86.mov

matuzalemsteles:
Hey @marcoscv-work I would like you to take a look at this PR. I'm going to move forward with this so we can get this functionality merged and we can unlock Lima and go into today's release and send out other follow-up PRs if needed.

PR #5362, opened 2023-02-07
fix(@clayui/css): Cadmin text-# utilities should use px values

### LINKED ISSUE #5361: @clayui/css: Cadmin text-# utilities should use px value not rem value
Rem values are based on the `font-size` declared on the `html` element. Changes to the page's `font-size` affect admin controls that should be isolated. In the example below, the Fragments and Widgets text should be 14px. It is 8.75px due to the base font-size being 10px.

icon-too-small

pat270:
fixes #5361


matuzalemsteles (approved):
LGTM

PR #5365, opened 2023-02-08
fix(@clayui/css): fix build error when `$secondary` variable is a CSS variable

matuzalemsteles:
Follow up #5364

The DXP build with the dialect theme is failing since the `$secondary` variable could be a CSS variable.


matuzalemsteles:
Merging to release a fix patch release.

PR #5369, opened 2023-02-09
fix(@clayui/css): Adjust list-group-text and subtext to match Lexicon specs

### LINKED ISSUE #5350: @clayui/css: List Group text and subtext should match Lexicon specs
Here is what it should look like

list-group

pat270:
fixes #5350


matuzalemsteles (approved):
lgtm

PR #5372, opened 2023-02-14
feat(@clayui/link): add extra aria-label when target is `_blank`

### LINKED ISSUE #5355: Adding an extra label to links when `target="_blank"` is present
### And, what about add an optional icon to make more visual? (patter sometimes we follow in portal)

It could improve the experience and operate in a more predictable way.

[WCAG 2.1 - 3.2 Predictable](https://www.w3.org/TR/WCAG21/#predictable)
Make Web pages appear and operate in predictable ways.

matuzalemsteles:
Closes #5355

This PR adds an extra label when `target` is set to `_blank`, it also adds a `messages` API which allows changing the text to handle internationalization.

Hey @marcoscv-work I didn't add the visual icon, do you have an example of what this should look like? I think we have to add this via a conditional as well because the link is used in several components and I'm not sure if that would be desired in some occasions.

### DIFF
--- a/packages/clay-link/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-link/src/__tests__/__snapshots__/index.tsx.snap
@@ -1,66 +1,82 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`ClayLink renders 1`] = `
-<a
-  className=""
-/>
+<div>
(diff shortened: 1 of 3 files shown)


matuzalemsteles:
I'm going to go ahead with this to include in the release, if we add the icon and understand better how we can implement it we can send it in another PR.

PR #5376, opened 2023-02-15
fix(@clayui/css): nav-underline and navigation-bar-light should use primary-l0 for active highlight

### LINKED ISSUE #5374: Tab and Nav active lines need a change of color
This is an accessibility bug related to any tab or navigation component that uses a line as main element for showing the active state of the tab/nav. 

Following the WCGA rules, the line used for the active state should have a contrast of at least 3:1

## Jira ticket
this issue is related to: https://issues.liferay.com/browse/LPS-168127

## Proposal
My proposal is to add a new color (Primary-L0) between the 2 that we currently have.
To get this new color I applied 14% of Lighten to the $Primary color and has a minimum color contrast of 3.15:1.
image

pat270:
fixes #5374


matuzalemsteles (commented):
LGTM

PR #5396, opened 2023-02-25
refactor(@clayui/core): uses `useNavigation` hook for keyboard navigation in TreeView

### LINKED ISSUE #5334: Tree: Component should be skippable by tabbing
Testing the tree component I just realized that we cannot skip the component and apply roving tabindex, so enter in a tree can be a nightmare and it could need thousen of tab pressings to get to the next focuseable element in the page
https://storybook.clayui.com/?path=/story/design-system-components-treeview--multiple-selection

https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex
https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/#keyboard-interaction-25

After review ARIA patterns it is my **Proposal**:
- Moves focus to the next interactive widget in the current element (tree node).
- If there are no more interactive controls in the current element, moves focus out of the tree.

matuzalemsteles:
Fixes #5334

Initial implementation to improve TreeView focus system, we are using `useNavigation` hook instead of FocusScope to handle focus and hotkeys, not covering all shortcut cases yet.

I still need to implement the TreeView focus control to retrieve the focus on the last item of the component, I will probably implement a component that can handle this since we already have 2 existing use cases in Clay that are starting to follow this behavior.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -63,7 +63,7 @@ module.exports = {
 		},
 		'./packages/clay-core/src/tree-view/': {
 			branches: 59,
-			functions: 70,
+			functions: 69,
 			lines: 71,
 			statements: 71,
 		},
--- a/packages/clay-core/src/aria/index.ts
+++ b/packages/clay-core/src/aria/index.ts
@@ -1,0 +1,6 @@
+/**
+ * SPDX-FileCopyrightText: © 2023 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export {useFocusWithin, FocusWithinProvider} from './useFocusWithin';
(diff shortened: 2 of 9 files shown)


matuzalemsteles:
Submit the commit that adds the focus control implementation for the most recent element within a scope. There are improvements that this implementation can be more generic to reuse in other components and remove their specific implementation.

I'm still investigating a bug where focus on actions isn't working anymore, I'll check what might be happening after implementing this new behavior but it seems to be something related to a race condition problem due to state updates to identify focus.

matuzalemsteles:
Fix bug with focus on item actions if it exists. Going ahead and merging that. This PR will not migrate all accessibility patterns from an ARIA `tree` to `treegrid` because for that we would have to rethink a lot of things in this component and this is more of a feature than a bug because we are following the accessibility standard of `tree` explicit with nested elements.

To follow `treegrid` we first need to work on converting a tree into a list and for that we need work on issue #5030 for TreeView virtualization and then move on to the implementation of migrating accessibility standards from a `tree` to `treegrid` like I commented on the issue #https://github.com/liferay/clay/issues/5334#issuecomment-1442621374. This path will be natural once we complete issue #5030.

PR #5401, opened 2023-03-02
feat(@clayui/css): Atlas and Cadmin update nav-tabs styles

### LINKED ISSUE #5398: Navbar vs Tabs - style update
## What is your proposal?
Update the tabs to a different style in order to differentiate them from the Navbars (and also merge all the current Tabs styles into the new one). 
image

More details in the [Confluence Page](https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2284388363/Tabs+vs+Navbar)

This change should increase affordance for the components, making it easier for the user to understand the differences between a navbar and a tab. 

none

## Info
[Jira ticket](https://issues.liferay.com/browse/DESIGN-5044)
[Figma file](https://www.figma.com/file/SIjiYx5WXdh1WxaFXKLzhZ/Untitled?node-id=1%3A17836&t=1sUrSYoo8vKWNciS-1)
[Confluence page](https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2284388363/Tabs+vs+Navbar)

pat270:
fixes #5398
basic-tabs


marcoscv-work:
`basic` is the default one?
We would like to replace the default style in whole portal at the same time

packages/clay-tabs/docs/tabs.mdx
```diff
@@ -5,12 +5,14 @@ lexiconDefinition: 'https://liferay.design/lexicon/core-components/tabs/'
...
 
-import {Tabs, TabsDropdown} from '$packages/clay-tabs/docs/index';
+import {Tabs, ModernTabs, TabsDropdown} from '$packages/clay-tabs/docs/index';
 
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
 -   [Introduction](#introduction)
+-   [Basic Tabs](#basic-tabs)
+-   [Modern Tabs](#modern-tabs)
```
matuzalemsteles, line 15: Since we no longer want to support the old style or others styles, we can deprecate the `displayType` APIs and remove the old css styles.

ethib137: If we are deprecating the `displayType` api can we also deprecate the `modern` api?

pat270:
I ended up deprecating `modern` and `displayType`. The tabs on clayui.com have little style now since we removed `nav-underline`. Should we replace it with `nav-tabs`?

tabs

marcoscv-work:
Yeah, we should replace all `nav-underline`.

And probably the 12 uses in DXP:
image

marcoscv-work:
And probably we will find some horizontal lines duplication when we use tabs after a navbar or in modal, etc.
We need to take a look on this @emiliano-cicero 🫣
PR preview:
https://deploy-preview-5401--next-storybook-clayui.netlify.app/?path=/story/design-system-components-tabs--default&args=disabledSecondTab:false;displayType:basic

packages/clay-tabs/docs/tabs.mdx
matuzalemsteles: @ethib137 yes, we can do that.

matuzalemsteles:
Hey @marcoscv-work do we want to force this style for all DXP use cases? if so @pat270 we also need to remove the JS part that adds the classes but we still need to keep the types for `displayType` and `modern` but no they don't do anything, just to make sure we don't break the DXP build.

It is worth remembering that forcing this change can affect older versions of DXP if it needs backporting for any corrections.

pat270:
@matuzalemsteles We will need to remove all cases of `modern` in DXP.

matuzalemsteles:
I'm not sure but from what I understand it's intentional to change the look of all tabs to this new look so in theory we would have to remove all tab styles and change to this one.

marcoscv-work:
nav-underline is an accessibility issue by concept since it is pretty similar to our navigation and does looks like an a tab, so I would remove it and I would go ahead with forcing the change, this is a bug fix actually.
Surely dxp we will have some problems with the style, it would probably be more related to changing the style like the bg, line, etc. So I would suggest to add a light version, something like this:

image

cc @emiliano-cicero

emiliano-cicero:
The intention with this change is to make it clearer for users that a Nav will activate a Navigation event and a Tab will simply change between Tabs. There are cases in DXP where they use tabs for navigation or navs for tabs, the solution for those cases should be to switch components when possible and we should help them when the case is not so clear.  
For now, since we're forcing a single style, I would go with the one with the background. We can always add a light style later if the results are too heavy visually.

matuzalemsteles:
Ok, I will remove the styles being added by the component and just keep the API to prevent the build from failing on old versions to force the new style even if `displayType` or `modern` is set. It is quite likely that we will receive messages from other teams.

matuzalemsteles:
Reesend #5424

PR #5406, opened 2023-03-07
fix(@clayui/css): Clay and Cadmin Menubar make collapsed nav 100% wide

### LINKED ISSUE #5397: Vertical navigation and Navigation Bar - Mobile / 400% zoom version
[Vertical Navigation](https://clayui.com/docs/components/vertical-nav.html)
[Navigation Bar](https://clayui.com/docs/components/navigation-bar.html)

In the past we decided to show the [vertical navigation](https://clayui.com/docs/components/navigation-bar.html) mobile version of the menu as a dropdown menu but actually is not a dropdown menu, it still being a navigation, here the **first issue**, users can misunderstand what is happening.
**The second one** is that the menu is positioning over the content and it can be overlapping content even when we get out of the menu (it is normal because was launched by a toggle button).

image

Technically, in mobile/400%zoom, the whole menu is hidden and a button appears, the button is a toggle button with open/close state so when you open the menu the show/hide state does not depends on if the focus is in the menu or no, thats the reason we need to change the positioning absolute and set the width to 100%:

image

Same solution for [Navigation bar](https://clayui.com/docs/components/navigation-bar.html) in mobile / 400% zoom version (it already has width 100%):
image

pat270:
fixes #5397

menubar-mobile

@marcoscv-work There's nothing I can do for [Navigation Bar](https://clayui.com/docs/components/navigation-bar.html). It's already at 100% width. The container it's in is the limitation. We could add some JS that re-calculates the `left` and `right` positions when the  screen size changes, but I'm not sure if we want that.

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_menubar.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_menubar.scss
@@ -3,10 +3,7 @@
 $menubar-vertical-transparent-md: () !default;
 $menubar-vertical-transparent-md: map-deep-merge(
 	(
-		mobile: (
-			background-color: $white,
-		),
-		link: (
+		nav-link: (
 			border-radius: 0.375rem,
 			transition: #{color 0.15s ease-in-out,
 			background-color 0.15s ease-in-out,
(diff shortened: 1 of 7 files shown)


matuzalemsteles:
I think that for the Navigation Bar we wouldn't need a JS implementation because normally it's used with the size of the full screen container instead of how it's being used in the documentation example.

marcoscv-work:
Hey @pat270 it is not only about with 100%, the menu content needs to push all the content down to avoid overlap:

https://github.com/liferay/clay/issues/5397#issuecomment-1455889197

pat270:
@marcoscv-work Thanks for clearing that up, it should do that now.

marcoscv-work:
Thanks @pat270 , that's perfect now! from my side we can merge and fix like 70 issues in portal!! thanks so much!

matuzalemsteles (approved):
LGTM! Let's go ahead and merge.

Issue #5404, opened 2023-03-07
updateDate is memoized without dateFormat dependency


NemethNorbert:
## Connected LPS to see where it can be reproduced within Liferay
https://issues.liferay.com/browse/LPS-177406

## Is there an example you can provide via codesandbox.com?
https://codesandbox.io/s/distracted-engelbart-lmoih7?file=/src/index.js
## What are the steps to reproduce?

- Change the date by using the date picker modal to 02/04/2023
- Notice the change does not cause any issue
- Click on the Change format button (this imitates the change of the date format, like changing locale while the date picker is on the site)
- Note the format change is reflected on the input field
- Open the date picker modal and note the date is wrong, as it shows april the second
- Change the selected date to number 6
- Note the day selection ended up being a month selection, and now your date is 4th of June

## What is the expected result?
Changing the format should effect the date picker modal as well, not just the inputField

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.89.0  |
| React | v16.8.4 |

PR #5413, opened 2023-03-10
fix(@clayui/date-picker): updateDate is memoized without dateFormat dependency

NemethNorbert:
Fixes: #5404

The root cause of the issue was the memoized updateDate, where I had to include the dateFormat in the dependancy array, as changing this value has an effect on the result.
I also made another commit, where I included every dependancy for the useCallback memoization, I left it in a separate commit in case you don't want to merge that.

Please review my changes.

Thanks


matuzalemsteles (approved):
LGTM! Thanks @NemethNorbert for the PR!

Issue #5414, opened 2023-03-10
Cadmin variables: Pagination active button contrast needs to be updated


veroglez:
Related to this issue https://issues.liferay.com/browse/LPS-169083, the atlas variables were changed in this [PR](https://github.com/liferay/clay/pull/5301) but not for cadmin.

<img width="300" alt="" src="https://user-images.githubusercontent.com/9701095/224335887-fb695970-c79d-4d76-b179-8789d86ef3f9.png">

veroglez:
Hi guys! I have sent this [PR](https://github.com/liferay/clay/pull/5415) to fix this 🙂 . Thanks in advance!

PR #5415, opened 2023-03-10
fix(@clayui/css): Cadmin Pagination active state should have more contrast

veroglez:
It fixes https://github.com/liferay/clay/issues/5414


pat270:
LGTM Thanks @veroglez!

PR #5419, opened 2023-03-14
feat(@clayui/core): Vertical Bar adds resizing to VerticalBar.Panel

### LINKED ISSUE #5348: Allow Resize in <VerticalBar.Panel>
## What is your proposal?

In portal we want to add the ability to resize the VerticalBar panel. Currently this is not possible as we cannot pass any other props to the `<VerticalBar.Panel>` component

I think the quick way to allow this would be to add the posibility to pass at least `className` and `style` attributes in the  `<VerticalBar.Panel>`  component. 

The other option is to add this resize feature to the Clay component, which makes sense from my point of view.

pat270:
fixes #5348 

Putting this up for review, I added a Resizer component to VerticalBar. We can enable it with the `resize` attribute on `VerticalBar`. I also added a `panelWidth` prop so we can store the resized width in local storage and set it when rendering the component.

resize

TODO:
- ~keyboard interactions~
- pass in custom function on resize end so we can store the new width in local storage (maybe?)
- aria attributes
- documentation

### DIFF
--- a/packages/clay-core/src/vertical-bar/Resizer.tsx
+++ b/packages/clay-core/src/vertical-bar/Resizer.tsx
@@ -24,8 +24,8 @@ let keyDownCounter = 0;
 
 export function Resizer({nodeRef, ...otherProps}: Props) {
 	const {
-		internalPanelWidth,
 		onPanelWidthChange,
+		panelWidth,
 		panelWidthMax,
 		panelWidthMin,
 		position,
@@ -39,7 +39,7 @@ export function Resizer({nodeRef, ...otherProps}: Props) {
 
 	const decreasePanelWidth = (delta = 1, startWidth = getStartWidth()) => {
 		const width =
-			internalPanelWidth - delta < panelWidthMin
+			panelWidth - delta < panelWidthMin
 				? panelWidthMin
 				: startWidth - delta;
 
@@ -48,7 +48,7 @@ export function Resizer({nodeRef, ...otherProps}: Props) {
 
 	const increasePanelWidth = (delta = 1, startWidth = getStartWidth()) => {
 		const width =
-			internalPanelWidth + delta > panelWidthMax
+			panelWidth + delta > panelWidthMax
 				? panelWidthMax
 				: startWidth + delta;
 
@@ -137,6 +137,6 @@ export function Resizer({nodeRef, ...otherProps}: Props) {
 			}}
 			role="separator"
 			tabIndex={0}
-		></div>
+		/>
 	);
 }
--- a/packages/clay-core/src/vertical-bar/Resizer.tsx
+++ b/packages/clay-core/src/vertical-bar/Resizer.tsx
@@ -38,21 +38,25 @@ export function Resizer({nodeRef, ...otherProps}: Props) {
 	};
 
 	const decreasePanelWidth = (delta = 1, startWidth = getStartWidth()) => {
-		const width =
-			panelWidth - delta < panelWidthMin
-				? panelWidthMin
-				: startWidth - delta;
+		if (panelWidth >= panelWidthMin) {
+			const width =
+				panelWidth - delta < panelWidthMin
+					? panelWidthMin
+					: startWidth - delta;
 
-		onPanelWidthChange(Math.round(width));
+			onPanelWidthChange(Math.round(width));
+		}
 	};
 
 	const increasePanelWidth = (delta = 1, startWidth = getStartWidth()) => {
-		const width =
-			panelWidth + delta > panelWidthMax
-				? panelWidthMax
-				: startWidth + delta;
+		if (panelWidth <= panelWidthMax) {
+			const width =
+				panelWidth + delta > panelWidthMax
+					? panelWidthMax
+					: startWidth + delta;
 
-		onPanelWidthChange(Math.round(width));
+			onPanelWidthChange(Math.round(width));
+		}
 	};
 
 	return (
--- a/packages/clay-core/src/vertical-bar/Resizer.tsx
+++ b/packages/clay-core/src/vertical-bar/Resizer.tsx
@@ -20,6 +20,8 @@ type Props = {
 	nodeRef?: React.RefObject<HTMLDivElement>;
 };
 
+const MAIN_MOUSE_BUTTON = 0;
+
 let keyDownCounter = 0;
 
 export function Resizer({nodeRef, ...otherProps}: Props) {
@@ -133,7 +135,7 @@ export function Resizer({nodeRef, ...otherProps}: Props) {
 					);
 				}
 
-				if (event.button === 0) {
+				if (event.button === MAIN_MOUSE_BUTTON) {
 					document.addEventListener('pointermove', onResizerMove);
 
 					document.addEventListener('pointerup', removeResizerEvents);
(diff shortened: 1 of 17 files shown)


marcoscv-work:
Looks awesome!!

I was not able to test it in: https://deploy-preview-5419--next-storybook-clayui.netlify.app/?path=/story/design-system-components-verticalbar--default
So I'm not sure about my question... :-D
Are implemented the resize controls by keyboard? I think is not in the requirements but we would need it, sidebar in page editor is worling by keyboard, I think is a good exmaple if needed

pat270:
@marcoscv-work It doesn't work with keyboard yet, but it's on the todo list.

packages/clay-core/src/vertical-bar/Resizer.tsx
```diff
@@ -0,0 +1,77 @@
...
+									Math.round(startWidth + delta)
+								);
+							} else if (
+								(event.pageX < startXPos && positionLeft) ||
+								(event.pageX >= startXPos && !positionLeft)
+							) {
+								onPanelWidthChange(
+									Math.round(startWidth - delta)
+								);
+							}
```
matuzalemsteles, line 50: We have to add minimum and maximum values to the width size, an implementation of clamp should take care of this, the minimum will probably have to be the default size.

packages/clay-core/src/vertical-bar/VerticalBar.tsx
```diff
@@ -58,6 +58,16 @@ type Props = {
 	 * Callback is called when the active state changes (controlled).
 	 */
 	onActiveChange?: InternalDispatch<React.Key | null>;
+
+	/**
+	 * Sets a custom width on the sidebar panel.
+	 */
+	panelWidth?: number | undefined;
```
matuzalemsteles, line 65: I'm not sure if it would be a good idea to provide an API to set the width, I think it would be better to provide an API to set the maximum size instead which would be used for the clamp.

matuzalemsteles:
The accessibility part we can create another issue to work on that, I think the JS implementation part is almost complete.

ambrinchaudhary:
Hi guys, just a question here. In our particular case (described in [LPS-156211](https://issues.liferay.com/browse/LPS-156211) with Figma [mockups](https://www.figma.com/file/eYdALYoOnfl5ftxUjftLst/LPS-127216-Design%3A-research-modernization-of-knowledge-base?node-id=829-534455&t=bX1V57ihSVNQR6Dt-0)) we will need to change the right section if the VerticalBar is resized, as we do not want to overlap it. So... would it be possible to have a `onResizeChange` or something similar and get the `panelWidth` value while it's being resized? Thanks!

packages/clay-core/src/vertical-bar/VerticalBar.tsx
matuzalemsteles: Hmm I don't know if I understood correctly, what would be the use case for setting the `panelWidth`? I think we still need to offer the `panelWidthMax` to use in the clamp.

pat270: Yes I'm including `panelWidthMin` and `panelWidthMax`. The use case for panelWidth is remembering the `panelWidth` via `SessionClicks.get(request, "id", "defaultValue")` in jsps.

matuzalemsteles: I think the component can handle this internally by saving to local storage.

pat270:
@matuzalemsteles It should work with the keyboard now. I've also included `panelWidthMax` and `panelWidthMin`. One thing I'm having trouble with is figuring out how to properly implement an `onResizeChange` prop.

packages/clay-core/src/vertical-bar/VerticalBar.tsx
pat270: We might run into a problem by handling this internally. The `id` isn't guaranteed to stay the same. We would lose reference to it.

packages/clay-core/src/vertical-bar/VerticalBar.tsx
```diff
@@ -88,6 +112,9 @@ export function VerticalBar({
 
 	const id = useId();
 
+	const [internalPanelWidth, setInternalPanelWidth] =
+		React.useState(panelWidth);
```
matuzalemsteles, line 116: @pat270 you can use `useInternalState` here to make this property controlled, so developers can use this information to do anything else, your `panelWidth` and `onPanelWidthChange` will now solve both use cases.

packages/clay-core/src/vertical-bar/Resizer.tsx
```diff
@@ -0,0 +1,142 @@
...
+				function removeResizerEvents() {
+					document.removeEventListener('pointermove', onResizerMove);
+
+					document.removeEventListener(
+						'pointerup',
+						removeResizerEvents
+					);
+				}
+
+				if (event.button === 0) {
```
matuzalemsteles, line 132: What specific case is this?

packages/clay-core/src/vertical-bar/Resizer.tsx
```diff
@@ -0,0 +1,142 @@
...
+	const positionLeft = position === 'left';
+
+	const getStartWidth = () => {
+		return nodeRef?.current?.offsetWidth || 320;
+	};
+
+	const decreasePanelWidth = (delta = 1, startWidth = getStartWidth()) => {
+		const width =
+			internalPanelWidth - delta < panelWidthMin
+				? panelWidthMin
```
matuzalemsteles, line 43: I think when the value reaches the minimum value we only need to call it once to avoid calling this function every time.

packages/clay-core/src/vertical-bar/Resizer.tsx
pat270: Right clicking the resizer doesn't remove the pointer events. After you right click and go back to the page, the resizer is stuck to cursor and you have to click it again to unstick it. `event.button === 0` makes sure to add the listeners only when the main mouse button is clicked.

ethib137: Maybe 0 should be declared as a named constant to avoid having a magic number here. IE: `const MAIN_BUTTON = 0;`

matuzalemsteles: I see, makes sense, maybe another strategy to insure it is removed is to stabilize the listener functions with `useCallback` and add a useEffect with zero dependency and a callback function that cleans up the methods, so when the component is unmounted we always make sure we are removing.

packages/clay-core/src/vertical-bar/VerticalBar.tsx
pat270: @matuzalemsteles I'm having trouble implementing this. Do we have any docs on how to use `useInternalState`? The error I'm dealing with right now is:
```
You provided a 'value' prop for a component without a handler 'onPanelWidthChange'. This will render the component with an internal state, if the component is to be uncontrolled, use 'undefined'. Otherwise, set the 'onPanelWidthChange' handler.
```
I really have no idea what I'm doing.

matuzalemsteles: I'll submit a commit implementing this.

matuzalemsteles:
This is the default of [controlled and uncontrolled](https://github.com/liferay/clay/blob/master/docs/proposals/0002-controlled-and-uncontrolled.md), if you declare `defaultPanelWidth` you can't use `onPanelWidthChange`, because `defaultPanelWidth` is uncontrolled and `onPanelWidthChange` and `panelWidth` are controlled, so you have to declare `onPanelWidthChange` and `panelWidth ` to control the state.

pat270:
Ok I think this is finally ready to go. Let me know if you want to ship documentation with this. I can add it to the pull tomorrow.

Issue #5409, opened 2023-03-09
Provide a public focusScope component


ethib137:
## What is your proposal?

Provide a public <FocusScope> component that can handle the following scenarios:

Everything inside FocusScope is part of the trap
```
<FocusScope trap loop>
    <Button 1>
    <Button 2>
    <Button 3>
    <Button 4>
</FocusScope>
```

Only part of the content inside FocusScope is part of the trap
```
<FocusScope trap loop>
    <Button 1>
    <FocusScope.Trap>
        <Button 2>
        <Button 3>
    </FocusScope.Trap>
    <Button 4>
</FocusScope>
```

## Why would adopting this proposal be beneficial?
Managing focus properly is key for accessibility and can be important to include in compositions of Clay components. 

## What are the alternatives to this proposal?
- We could encourage users to manually control focus when needed.
- We could create compositions as part of the Clay Library anytime a new composition of focus scope is needed.

Neither of these options feels ideal.

veroglez:
Working on it! (waiting for access as a collaborator to assign it to me)

PR #5427, opened 2023-03-17
feat(@clayui/focus-trap): Add FocusTrap component

veroglez:
It fixes: [#5409](https://github.com/liferay/clay/issues/5409)

Hi [@matuzalemsteles](https://github.com/matuzalemsteles)! Here is the draft for the new FocusTrap component. I have proposed the following: the component has an active props, which is in charge of activating the focus trap. In this way, it's the responsibility of the user to activate the focus trap. It could be always activated or, for example, activate it with the click of a button.

I have also added stories for Storybook to verify its behavior (video attached). What do you think?

https://user-images.githubusercontent.com/9701095/225911367-b8e9843c-96b6-4a31-b706-e170c076330c.mov

---
**TO DO:**
- tests
- doc

### DIFF
--- a/packages/clay-core/src/focus-trap/FocusTrap.tsx
+++ b/packages/clay-core/src/focus-trap/FocusTrap.tsx
@@ -4,6 +4,7 @@
  */
 
 import {FOCUSABLE_ELEMENTS, FocusScope} from '@clayui/shared';
+import {hideOthers, supportsInert, suppressOthers} from 'aria-hidden';
 import React, {useEffect, useRef} from 'react';
 
 type Props = {
@@ -48,6 +49,19 @@ export function FocusTrap({active = false, children, focusElementRef}: Props) {
 
 			if (focusableElements) {
 				(focusableElements[0] as HTMLDivElement).focus();
+			}
+		}
+	}, [active]);
+
+	useEffect(() => {
+		if (childrenRef.current && active) {
+			// Inert is a new native feature to better handle DOM arias that are not
+			// assertive to a SR or that should ignore any user interaction.
+			// https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert
+			if (supportsInert()) {
+				return suppressOthers(childrenRef.current);
+			} else {
+				return hideOthers(childrenRef.current);
 			}
 		}
 	}, [active]);
--- a/packages/clay-core/src/focus-trap/FocusTrap.tsx
+++ b/packages/clay-core/src/focus-trap/FocusTrap.tsx
@@ -16,6 +16,11 @@ type Props = {
 	 * The elements that will receive the focus within the focus trap.
 	 */
 	children: React.ReactNode;
+
+	/**
+	 * Ref of the element that receives the focus when the focus trap is activated.
+	 */
+	focusElementRef?: React.RefObject<HTMLElement>;
 };
 
 const getFocusableElements = (childrenRef: React.RefObject<HTMLDivElement>) => {
@@ -28,11 +33,17 @@ const getFocusableElements = (childrenRef: React.RefObject<HTMLDivElement>) => {
 	].filter((element) => !element.getAttribute('aria-hidden'));
 };
 
-export function FocusTrap({active = false, children}: Props) {
+export function FocusTrap({active = false, children, focusElementRef}: Props) {
 	const childrenRef = useRef<HTMLDivElement>(null);
 
 	useEffect(() => {
 		if (active) {
+			if (focusElementRef?.current) {
+				focusElementRef.current.focus();
+
+				return;
+			}
+
 			const focusableElements = getFocusableElements(childrenRef);
 
 			if (focusableElements) {
--- a/packages/clay-core/src/focus-trap/FocusTrap.tsx
+++ b/packages/clay-core/src/focus-trap/FocusTrap.tsx
@@ -1,0 +1,67 @@
+/**
+ * SPDX-FileCopyrightText: © 2021 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {FOCUSABLE_ELEMENTS, FocusScope} from '@clayui/shared';
+import React, {useEffect, useRef} from 'react';
+
+type Props = {
+	/**
+	 * Flag to indicate if the focus trap is activated.
+	 */
+	active: boolean;
+
+	/**
+	 * The elements that will receive the focus within the focus trap.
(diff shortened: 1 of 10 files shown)


packages/clay-core/src/focus-trap/FocusTrap.tsx
matuzalemsteles, line 1: I think that by default FocusTrap should already be active, in theory when FocusTrap is used to lock the focus of elements within the scope normally the element is rendered with the lock and removed when it is no longer rendered, for example, the dialog of the DatePicker and Modal.

marcoscv-work: It is a little bit strange from my since the pattern of having a focus trap by default in your markup ready to be focused in a regular tabbing is something can be considered an accessibility issue, the focus have to be init by the user in some way, for example:
Opening a menu, opening a modal, opening a dialog, opening a search box, etc.
it is described by WCAG as "enter in a widget".
https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html

I know that it is just showing a example of use but if we are thinking in a probably clay/lexicon certification is something to have in count.
Perhaps we could even consider excluding storybooks of the certification... but inevitably the markup and examples in lexicon/clay main pages should be accessible. :-D

matuzalemsteles: Yeah, that's why FocusTrap by itself has to be active, it shouldn't be used together with a trigger for example the button that opens a Modal, but inside the content and only rendered when the content is visible, the trap will disappear when the user doing an action that removes from rendering the most common examples of dialog is clicking the close button or saving anything that removes visibility i.e. FocusTrap shouldn't be a state switcher.

marcoscv-work: I see! but If we trap the keyboard when the user dint init the action to trap, it is by definition a keyboard trap, an accessibility issue.
Traps can be activated when the user init an action to trap: Open a modal, open a dropdown, open a searchbox, etc

So if we make the example in this way, it will be a perfect example of a keyboard trap: WCAG 2.1: 2.1.2: No Keyboard Trap (Level A)

matuzalemsteles: Oh I see, yes I also think we should provide an example that is closer to this I agree, this can be confusing as it is.

veroglez: @matuzalemsteles, I have added an example that may be closer to what we are talking about.

https://user-images.githubusercontent.com/9701095/227992390-1050ffc2-4a09-490a-b334-67785dbaf0fb.mov

---
Talking with @marcoscv-work, we think it would be useful to add a new props `focusElementRef`. This props is the reference of the element which the user wants to focus. If nothing is passed to that props, it will focus the first focusable element.

packages/clay-core/stories/FocusTrap.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+	return (
+		<>
+			<ClayButton onClick={onActivateFocusTrap} ref={activateTrapButton}>
+				Activate trap
+			</ClayButton>
+			{active && (
+				<FocusTrap
+					active={active}
+					focusElementRef={lastButton}
+					onDeactivate={onDeactivateFocusTrap}
```
matuzalemsteles, line 81: I think I wouldn't need `active` and `onDeactivate` anymore since the conditional is done outside the component, to restore the focus, I would add a `useEffect` with the dep of `active`.

veroglez: In this case, there may be components like the Management Toolbar that need the props `active` inside the component. For example, on desktop, the management toolbar displays the following search form:

Screen Shot 2023-03-28 at 4 09 03 PM

This search form is the same as on mobile, which here is hidden by css classes. In this case the focus trap should wrap the search form, but we can't hide it because it would also hide on desktop (on desktop there shouldn't be focus trap either).

https://user-images.githubusercontent.com/9701095/228273513-b56c5989-3f21-4f4c-9e7b-b486bac16bfb.mov

<br></br>
About the `onDeactive` props is simply to be executed when the user presses the escape key. Do you think there may be another way to do this?

packages/clay-core/src/focus-trap/FocusTrap.tsx
veroglez: Yeah, that's a good idea! this would simplify the API. Do you think it could be less confusing for the user when using the component?

packages/clay-core/stories/FocusTrap.stories.tsx
matuzalemsteles: Hmm I see, esse é um caso valido com certeza. 
>

packages/clay-core/src/focus-trap/FocusTrap.tsx
matuzalemsteles: > Yeah, that's a good idea! this would simplify the API. Do you think it could be less confusing for the user when using the component?

I think this is closer to the native implementation, sometimes it's hard to get the element's ref if it's nested.

packages/clay-core/src/focus-trap/FocusTrap.tsx
```diff
@@ -0,0 +1,82 @@
+/**
+ * SPDX-FileCopyrightText: © 2021 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {FOCUSABLE_ELEMENTS, FocusScope} from '@clayui/shared';
+import React, {useEffect, useRef} from 'react';
+
+type Props<T> = {
```
matuzalemsteles, line 9: Just noticed that the generic type is not being used, maybe you don't need it anymore now.

matuzalemsteles (approved):
@veroglez looks good now, I just added a comment of a small detail but looks good to go ahead now. Thanks for working on it.

veroglez:
Hi @matuzalemsteles! I have added two commits with the tests and documentation (https://github.com/liferay/clay/pull/5427/commits/3c8a1df89c8ac41cfdd461b9e7f0395440138980, https://github.com/liferay/clay/pull/5427/commits/af2e8f6bf1fc44df02fe3d995a2e37840567e318) , could you take a look? Thanks in advance!

matuzalemsteles (approved):
LGTM! I send a commit to add the functionality to suppress elements outside the scope of the focus trap, so it will work better when using the mouse for example by clicking elsewhere on the page, the focus will go back into scope again.

PR #5428, opened 2023-03-20
refactor(@clayui/core): WIP refactoring the drop implementation

### LINKED ISSUE #5383: Use the new drop indicator in the TreeView
We recently implemented drag and drop via keyboard in TreeView #5353, for the drop to be announced correctly we created an indicator that is only visible to the SR but visually we still use the old way of showing where the drop is, in the current case the indicator will always be the item is above or below the item this causes you to have two drops in the same place when you try to drop between two items for example, one drop below and the other drop above the item.

To solve this we have to use the new drop indicator we used for the SR and make it visible and integrate with react-dnd's `useDrop`.

matuzalemsteles:
Fixes #5383

Just a draft refactoring of the drop implementation for the TreeView, there are still some tweaks to be made. Maybe I have to change some strategies to make the drop interaction more fluid when using the mouse.

This is still not fully working, so don't test it for now.

### DIFF
--- a/packages/clay-core/src/tree-view/DragAndDrop.tsx
+++ b/packages/clay-core/src/tree-view/DragAndDrop.tsx
@@ -5,6 +5,7 @@
 
 import {Keys, useId} from '@clayui/shared';
 import {suppressOthers} from 'aria-hidden';
+import classNames from 'classnames';
 import React, {
 	useCallback,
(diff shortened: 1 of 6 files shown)


matuzalemsteles:
Another implementation #5448

PR #5434, opened 2023-03-23
fix(@clayui/core): Vertical Bar open panel shouldn't disappear on resize

### LINKED ISSUE #5433: @clayui/core: Resizing an open panel in Vertical Bar makes the panel disappear
A panel that is open on first render disappears when trying to resize.

resize-when-open

pat270:
fixes #5433

### DIFF
--- a/packages/clay-core/src/vertical-bar/Panel.tsx
+++ b/packages/clay-core/src/vertical-bar/Panel.tsx
@@ -33,7 +33,7 @@ type Props = {
 	 * Internal property.
 	 * @ignore
 	 */
-	keyValue?: React.Key;
+	keyValue?: React.Key | null;
 
 	/**
 	 * Indicates whether the panel can be focused.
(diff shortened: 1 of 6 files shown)


packages/clay-core/src/vertical-bar/Panel.tsx
```diff
@@ -60,7 +46,7 @@ export function Panel({children, keyValue, tabIndex}: Props) {
 		<CSSTransition
 			aria-labelledby={`${id}-tab-${keyValue}`}
 			className={classNames('sidebar', {
-				'c-slideout-show': activePanel === keyValue && isFirst,
```
matuzalemsteles, line 63: @pat270 we have to ignore this in the first rendering, for example when opening the panel it is not doing animation anymore.

pat270: @matuzalemsteles Yes sort of, React ends up re-rendering which removes `c-slideout-show`.

pat270:
@matuzalemsteles I solved the disappearing panel and also fixed the transition between close and open panels. This seems a bit messy, but can't figure out a better way at the moment.

matuzalemsteles (commented):
> fixed the transition between close and open panels

I'm not sure about this behavior, I think the old behavior was expected but it's also tricky because the design and implementation of VerticalBar is old and will change so we don't know what the ideal behavior would be yet.

pat270:
@matuzalemsteles I remember there being a transition between panels somewhere but I can't seem to find it.

@emiliano-cicero @marcoscv-work how should the Vertical Bar transition between two panels?

1) No sliding transition
vertical-bar-no-transition

2) With sliding transition
vertical-bar-transition

matuzalemsteles:
@pat270 without the transition it shouldn't blink just render new content because the panel is already open. Just as it is in the master.

https://user-images.githubusercontent.com/13750819/228379301-daa7bd28-203b-4557-acf3-de589a7d72a1.mp4

pat270:
@matuzalemsteles The panels should function how it did previously. We can revert to sliding panels in the future if it's needed.

matuzalemsteles (approved):
@pat270 seems to work great, could you just check the tests before going ahead with this?

pat270:
@matuzalemsteles Vertical Bar tests should be passing now

PR #5442, opened 2023-03-28
fix(@clayui/core): always shows the drag and drop icon

### LINKED ISSUE #5430: Drag&Drop in TreeView is activated whenever other element (like dropdown) is focused outside the component
There are two examples in Liferay Portal, see issue https://issues.liferay.com/browse/LPS-178730 
It happens in _Knowledge Base_ (activate FF `feature.flag.LPS-156421=true`) and in _Pages Tree_ component

In knowledge Base: 

1. Open the Folders and Articles section
2. Click on the kebab menu near Home folder, or try to change view from list to table, or open the info panel ("i" icon in the Management toolbar)

In Pages Tree: 
1. Click on the "+" icon near the selector

The Drag&Drop option must be activated only by keyboard (I guess...)

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | current version in master |
| React |         |

matuzalemsteles:
Fixes #5430

I'm going in the direction of always showing the drag and drop icon when interacting via mouse or keyboard but clicking the mouse on the drag and drop button won't do anything only when interacting via keyboard. This still maintains the idea that the drag and drop is done via the mouse.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -62,7 +62,7 @@ module.exports = {
 			statements: 91,
 		},
 		'./packages/clay-core/src/tree-view/': {
-			branches: 59,
+			branches: 58,
 			functions: 69,
 			lines: 71,
 			statements: 71,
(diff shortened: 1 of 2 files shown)


ethib137:
This sounds like a good way to handle this! Thanks.

Issue #5339, opened 2023-02-02
Complement to `prefers-reduced-motion`


marcoscv-work:
We are using `prefers-reduced-motion` and it is great. It can be configurated at OS level.
We will implement soon a menu to configure accessibility improvments like the possibility of stop motion only for DXP context, so. Could we add any other mechanisms to control this in the page? I dont know, something like a global class or conditional

matuzalemsteles:
I tried to check if there is a global css strategy like the dark mode theme with `color-scheme: dark;` that we could use this to change the media query but I couldn't find it of the same genre so we'll have to create a class global to add in the body to toggle the setting.

cc @pat270

pat270:
Yeah it looks like we need to add a class to disable animation/transitions for specific pages or parts of the page.

marcoscv-work:
@pat270 do you think we could work on this at Clay level and use Clay example to propagate the conditional class to DXP?

marcoscv-work:
Not sure if we can use something like:

```css
.noAnimations * {
    animation-duration: 0;
}
```

We also will need changes to emptyState component in order to stop gifs

marcoscv-work:
Since we need to follow the SO property reduce-motion and use a global class (coming to accessibility menu configuration) to be added from Accessibility menu I've done this example:

https://codepen.io/marcoscv/pen/MWqmVrx

https://user-images.githubusercontent.com/7963804/222760611-4270eb80-fe43-4ad3-bfd7-561cb0ffc710.mov

I've used custom properties with the URLs and background-image instead of an `<img>` in order to provide the power to control reduce-motion visibility to the CSS

- We will need to explain/document that if your emptyState is an animation you will need to provide an alternative
- I can work on the alternative for the current's gifs animations in DXP since I've developed them

matuzalemsteles:
I think for some components we need to implement a feature via JS to identify if the animation has been disabled, for some animations are done via JS with the control to add and remove classes.

marcoscv-work:
> I think for some components we need to implement a feature via JS to identify if the animation has been disabled, for some animations are done via JS with the control to add and remove classes.

That's true @matuzalemsteles thanks!
Can we list those animations in some way?

matuzalemsteles:
> Can we list those animations in some way?

I think I ended up missing the notification for that message. But we have some animations on the components:

- Panel
- VerticalBar
- VerticalNav
- NavigationBar
- TreeView
- Modal

PR #5443, opened 2023-03-28
feat(@clayui/css): Utilities adds c-prefers-reduced-motion

pat270:
#5339

We can add the class `c-prefers-reduced-motion` to the whole page or specific regions of the page similar to the `cadmin` class.

### DIFF
--- a/clayui.com/content/docs/css/utilities/c-prefers-reduced-motion.md
+++ b/clayui.com/content/docs/css/utilities/c-prefers-reduced-motion.md
@@ -1,0 +1,84 @@
+---
+title: 'C Prefers Reduced Motion'
+---
+
+<div class="nav-toc-absolute">
+<div class="nav-toc">
+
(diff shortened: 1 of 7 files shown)


clayui.com/content/docs/css/utilities/c-prefers-reduced-motion.md
matuzalemsteles, line 1: I think once we have the JS implementation we can move this to the components section to be something more general that works for JS and CSS as well.

matuzalemsteles (approved):
LGTM

clayui.com/content/docs/css/utilities/c-prefers-reduced-motion.md
pat270: @matuzalemsteles sounds good. This implementation should cover all the use cases listed at https://github.com/liferay/clay/issues/5339#issuecomment-1482084399 since they're all using CSS classes to transition.

This won't cover transitions manually created in DXP though.

PR #5447, opened 2023-03-30
feat(@clayui/css): Cadmin and Clay adds btn-beta and badge-beta variants

### LINKED ISSUE #5446: Beta Button & Badge Variant
In order to accommodate for the new [BETA Indicator](https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2312110481/Identifying+BETA+Features+in+the+product) feature in Liferay, I propose that we create new variants of the button and badge components that will add the correct colors for a beta-button and a beta-badge. I don't think we need to create the BETA Indicator component in Clay. This can be a portal composition of Clay components. 

What do you all think, does that seam like a reasonable way to handle this feature?

I would expect Clay to provide two new classNames, one for buttons (.btn-beta) and one for badges (.badge-beta). The button styling will need to handle Default, Hover, and Active states. The badge styling will only have a default state since it has no interaction.

pat270:
fixes #5446

One thing I forgot to ask is whether `btn-beta` and `btn-badge` will be used in admin only components. Should we remove this from Clay used on sites?

Also @matuzalemsteles you think it's a good idea to add these to the ClayButton and ClayBadge apis?

/cc @ethib137 @marcoscv-work @drakonux

### DIFF
--- a/packages/clay-css/src/scss/cadmin/variables/_buttons.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_buttons.scss
@@ -702,16 +702,16 @@ $cadmin-btn-palette: map-deep-merge(
 		dark: $cadmin-btn-dark,
 		link: $cadmin-btn-link,
 		btn-beta: (
-			background-color: $cadmin-gray-100,
-			color: $cadmin-info,
+			background-color: $cadmin-light-l1,
+			color: $cadmin-info-d1,
 			text-transform: uppercase,
 			hover: (
-				background-color: $cadmin-gray-200,
-				color: $cadmin-info,
+				background-color: $cadmin-light,
+				color: $cadmin-info-d1,
 			),
 			focus: (
-				background-color: $cadmin-gray-200,
-				color: $cadmin-info,
+				background-color: $cadmin-light,
+				color: $cadmin-info-d1,
 			),
 		),
 	),
--- a/packages/clay-css/src/scss/cadmin/variables/_buttons.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_buttons.scss
@@ -203,7 +203,7 @@ $cadmin-btn-sizes: map-deep-merge(
 			padding-top: 0.125rem,
 			inline-item: (
 				font-size: inherit,
-				margin-top: -3px,
+				margin-top: -0.16em,
 			),
 			inline-item-before: (
 				margin-right: 0.25rem,
--- a/packages/clay-css/src/scss/cadmin/variables/_buttons.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_buttons.scss
@@ -701,6 +701,19 @@ $cadmin-btn-palette: map-deep-merge(
 		light: $cadmin-btn-light,
 		dark: $cadmin-btn-dark,
 		link: $cadmin-btn-link,
+		btn-beta: (
+			background-color: $cadmin-gray-100,
+			color: $cadmin-info,
+			text-transform: uppercase,
+			hover: (
+				background-color: $cadmin-gray-200,
+				color: $cadmin-info,
+			),
+			focus: (
+				background-color: $cadmin-gray-200,
+				color: $cadmin-info,
+			),
+		),
 	),
 	$cadmin-btn-palette
 );
--- a/packages/clay-css/src/scss/components/_badges.scss
+++ b/packages/clay-css/src/scss/components/_badges.scss
@@ -113,7 +113,11 @@
 			starts-with($color, '#') or
 			starts-with($color, '%'),
 		$color,
-		str-insert($color, '.badge-', 1)
+		if(
+			starts-with($color, 'badge'),
+			str-insert($color, '.', 1),
+			str-insert($color, '.badge-', 1)
+		)
 	);
 
 	@if (starts-with($color, '%') or map-get($value, extend)) {
--- a/packages/clay-css/src/scss/variables/_buttons.scss
+++ b/packages/clay-css/src/scss/variables/_buttons.scss
@@ -893,16 +893,16 @@ $btn-palette: map-deep-merge(
 		dark: $btn-dark,
 		link: $btn-link,
 		btn-beta: (
-			background-color: $gray-100,
-			color: $info,
+			background-color: $light-l1,
+			color: $info-d1,
 			text-transform: uppercase,
 			hover: (
-				background-color: $gray-200,
-				color: $info,
+				background-color: $light,
+				color: $info-d1,
 			),
 			focus: (
-				background-color: $gray-200,
-				color: $info,
+				background-color: $light,
+				color: $info-d1,
 			),
 		),
 	),
--- a/packages/clay-css/src/scss/variables/_buttons.scss
+++ b/packages/clay-css/src/scss/variables/_buttons.scss
@@ -203,7 +203,7 @@ $btn-sizes: map-deep-merge(
 			padding-top: 0.125rem,
 			inline-item: (
 				font-size: inherit,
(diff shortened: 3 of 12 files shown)


matuzalemsteles:
> Also @matuzalemsteles you think it's a good idea to add these to the ClayButton and ClayBadge apis?

I'm not sure, because this seems to be something quite specific. I think it would be much better to be a component that already renders the BETA without allowing to change the label for example. But we need something for everyone to use, maybe we can create a component just for that.

ethib137:
Hey @matuzalemsteles, yeah I agree that we should have a component for this, but I was thinking it would make more sense to be a shared component in Portal, rather than something owned by Clay. My reasoning for this is that it's more of a complex component composed of pieces from Clay (Button, Badge, Tooltip) to solve a specific business need. 

Do you think that makes sense, or do you think it would make more sense to have the `BETA Indicator` component directly in Clay?

ethib137:
@pat270 I think it's likely this would not only show up in admin. For instance it could easily be needed in a Commerce widget.

packages/clay-css/src/scss/cadmin/variables/_buttons.scss
```diff
@@ -701,6 +701,19 @@ $cadmin-btn-palette: map-deep-merge(
 		light: $cadmin-btn-light,
 		dark: $cadmin-btn-dark,
 		link: $cadmin-btn-link,
+		btn-beta: (
+			background-color: $cadmin-gray-100,
```
ethib137, line 705: Hey @pat270, design has requested that the background colors for these actually use opacity to define the color so that it properly renders in a variety of different contexts where the background color could be light or dark. IE: $info-d2 4%

ethib137:
Screenshot 2023-03-31 at 12 12 44 PM
It looks like the vertical alignment for the text is a little to high.

packages/clay-css/src/scss/components/_badges.scss
```diff
@@ -113,7 +113,11 @@
 			starts-with($color, '#') or
 			starts-with($color, '%'),
 		$color,
-		str-insert($color, '.badge-', 1)
+		if(
+			starts-with($color, 'badge'),
```
ethib137, line 117: Why do we need to have different logic for beta? Can't we remove this change and just let the old logic insert `.badge-` for beta too?

packages/clay-css/src/scss/variables/_buttons.scss
```diff
@@ -892,6 +892,19 @@ $btn-palette: map-deep-merge(
 		light: $btn-light,
 		dark: $btn-dark,
 		link: $btn-link,
+		btn-beta: (
```
ethib137, line 895: Why are we prefixing beta with btn, instead of letting btn be added automatically like it is for the other variants?

matuzalemsteles:
> yeah I agree that we should have a component for this, but I was thinking it would make more sense to be a shared component in Portal, rather than something owned by Clay. My reasoning for this is that it's more of a complex component composed of pieces from Clay (Button, Badge, Tooltip) to solve a specific business need.
>
> Do you think that makes sense, or do you think it would make more sense to have the BETA Indicator component directly in Clay?

@ethib137 yes I agree with you, I am also more inclined to make this component on the DXP side, what I have a little doubt is having the beta nomenclatures as a variant, I think this would be better if the name was generic, makes me think about keeping the component in Clay too because of that.

packages/clay-css/src/scss/cadmin/variables/_buttons.scss
pat270: @ethib137 I tried the opacity, but it fails under dark backgrounds.

packages/clay-css/src/scss/components/_badges.scss
pat270: We've moved to using key names that are closer to the selector name. The pattern to use:
```
$map: (
    primary: (),
    secondary: (),
    ...
);
```
is from Bootstrap. The new way should be:
```
$map: (
    btn-primary: (),
    btn-secondary: (),
);
```
These extra steps just provide backward compatibility for Clay 2+.

packages/clay-css/src/scss/variables/_buttons.scss
pat270: Same reason as https://github.com/liferay/clay/pull/5447#discussion_r1154767684, I wanted to convert these to have the `btn-` prefix, but can't due to backward compatibility.

ethib137:
> @ethib137 yes I agree with you, I am also more inclined to make this component on the DXP side, what I have a little doubt is having the beta nomenclatures as a variant, I think this would be better if the name was generic, makes me think about keeping the component in Clay too because of that.

@matuzalemsteles The main reason to add beta as a variant rather than make it more generic is this. If we made this generic, we would probably add an inverse utility class. This should then modify all of our variants. This would add quite a bit more css that we don't have any requirements for. Additionally the colors required for beta might not match what we would want for an inverse info button or label. In order to be consistent we would be forced to use the beta button/badge as the pattern for all inverse types. As a result, it seems easier and safer to simply add it as a variant.

Adding it in Clay rather than directly in portal is important to make it possible to easily override the styles using sass. 

Is that convincing, or do you still see a reason to implement this differently?

matuzalemsteles:
@ethib137 I see, it makes sense to me. Taking that into account keeping the component on the Clay side would make sense as well.

ethib137:
@matuzalemsteles, why would it make sense to have the Beta Indicator component in Clay? 

By the way I'm expecting the component to be something like [this](https://codesandbox.io/s/beta-button-badge-c0l6bc?file=/src/index.js). There are also a few oddities about the requirements. For instance, all the strings are intended to not be configurable. They will be the same throughout portal, so we would want to set them once in a shared location and then have everyone use it. So even if we created this component in Clay, we would still need to add the translations in portal and then share the translated component with all other modules.

matuzalemsteles:
> why would it make sense to have the Beta Indicator component in Clay?

@ethib137 I like to follow the same line that if we are adding something in CSS we should also have something in React, both are part of the same package and maintenance in my view it is strange to just provide the CSS but the component will be implemented elsewhere, it doesn't do much sense only to keep one side. If we want it to be DXP only I think CSS should also live in DXP only.

> By the way I'm expecting the component to be something like [this](https://codesandbox.io/s/beta-button-badge-c0l6bc?file=/src/index.js). There are also a few oddities about the requirements. For instance, all the strings are intended to not be configurable. They will be the same throughout portal, so we would want to set them once in a shared location and then have everyone use it. So even if we created this component in Clay, we would still need to add the translations in portal and then share the translated component with all other modules.

I see, is the term BETA translatable? I thought it would be the same regardless of the language 😅, at least in Brazil we don't translate it to another name.

ethib137:
> if we are adding something in CSS we should also have something in React
@matuzalemsteles , yes to this. But only to adding the new types to badge and button. Then the Beta Indicator component that is composed of a badge, button, tooltip, and popover could still be built in portal NOT Clay, right? 

> I see, is the term BETA translatable?
Yeah, even Beta should probably be translatable or more really localizable is probably the better term in this case. Some languages will probably keep it as Beta, but I expect Japanese and some other languages to have alternatives. The title and the description in the popover as well as the tooltip will definitely be localized though.

packages/clay-css/src/scss/cadmin/variables/_buttons.scss
matuzalemsteles: Updated the colors!

matuzalemsteles:
I've updated the latest changes, I'll go ahead and merge so we can proceed with creating the component on the DXP side.

Issue #5444, opened 2023-03-30
[Multi Step Nav] Create a New Error State


drakonux:
## What is your proposal?
Due to [DESIGN-5143](https://issues.liferay.com/browse/DESIGN-5143) we would like to create a new state for the Step.

<img width="800" alt="Error State" src="https://user-images.githubusercontent.com/21216422/228770094-f003036d-5392-40f2-835e-8b3510ad1d1e.png">

## Why would adopting this proposal be beneficial?
We need to add a new Error state to allow users to give feedback when a problem or action is required. 

## What are the alternatives to this proposal?
Instead of creating a specific state for the Step, allowing users to customize the icon and background color via API provides a more flexible way of covering their cases (error, warning, others...) 

cc\ @ethib137 @liferay-lexicon

marcoscv-work:
Please, if we can provide in the implementation example: tooltips to the steps and include one with an error message in the error step would be great

matuzalemsteles:
Just a question about this, do we need to create something that allows to customize the indicator or just create an error variant?

PR #5450, opened 2023-03-31
feat(@clayui/css): Multi Step Nav adds error state

pat270:
#5444

@matuzalemsteles This is the CSS implementation for Multi Step Nav error state. I wasn't able to complete the React component part of it today.

### DIFF
--- a/packages/clay-multi-step-nav/docs/multi-step-nav.mdx
+++ b/packages/clay-multi-step-nav/docs/multi-step-nav.mdx
@@ -8,12 +8,15 @@ storybookPath: 'design-system-components-multistepnav'
 
 import {
 	MultiStepNav,
+	MultiStepNavError,
 	MultiStepNavWithBasicItems,
 } from '$packages/clay-multi-step-nav/docs/index';
 
 <div class="nav-toc-absolute">
 <div class="nav-toc">
 
+-   [State](#state)
+-   [Indicator position](#indicator-position)
 -   [Collapsable Steps](#collapsable-steps)
 
 </div>
@@ -24,6 +27,16 @@ It's used when a major or big task has to be divided into smaller task, with the
 Each step can have two different states: `active` or `complete` defined by props as you can see below.
 
 <MultiStepNav />
+
+## State
+
+The current step can currently be represented by three states, `complete`, `active` or `error`.
+
+The active state can be set using the `active` property in the `ClayMultiStepNav.Item` or `ClayMultiStepNavWithBasicItems` component which defines the step index.
+
+The `complete` and `error` states are defined using the `state` property in the `ClayMultiStepNav.Item` which defines whether the item is complete or had an error in the current state, the `ClayMultiStepNavWithBasicItems` component follows the same API with the difference that it sets the state of the item to the current step which is set when using the `active` or `defaultActive` API.
+
+<MultiStepNavError />
 
 ## Indicator position
 
--- a/packages/clay-multi-step-nav/docs/multi-step-nav.mdx
+++ b/packages/clay-multi-step-nav/docs/multi-step-nav.mdx
@@ -30,11 +30,11 @@ Each step can have two different states: `active` or `complete` defined by props
 
 ## State
 
-The current step can currently be represented by three states, `complete`, `active` or `error`.
+Each step can have two different states: `active` or `complete` defined by props as you can see below.
 
 The active state can be set using the `active` property in the `ClayMultiStepNav.Item` or `ClayMultiStepNavWithBasicItems` component which defines the step index.
 
-The `complete` and `error` states are defined using the `state` property in the `ClayMultiStepNav.Item` which defines whether the item is complete or had an error in the current state, the `ClayMultiStepNavWithBasicItems` component follows the same API with the difference that it sets the state of the item to the current step which is set when using the `active` or `defaultActive` API.
+The `complete` and `error` states are defined using the `state` property in the `ClayMultiStepNav.Item` which defines whether the item is complete or has an error in the current state. The `ClayMultiStepNavWithBasicItems` component follows the same API with the difference that it sets the state of the item to the current step which is set when using the `active` or `defaultActive` API.
 
 <MultiStepNavError />
 
--- a/packages/clay-multi-step-nav/src/Item.tsx
+++ b/packages/clay-multi-step-nav/src/Item.tsx
@@ -19,18 +19,19 @@ export interface IProps extends React.HTMLAttributes<HTMLLIElement> {
 
 	/**
 	 * Flag to indicate if `complete` classname should be applied
+	 * @deprecated since v3.91.0 - use `state` instead.
 	 */
 	complete?: boolean;
-
-	/**
-	 * Flag to indicate that the step has an error.
-	 */
-	error?: boolean;
 
 	/**
 	 * Flag to indicate if progress line should expand out from step
 	 */
 	expand?: boolean;
+
+	/**
+	 * Defines the status of the current step.
+	 */
+	state?: 'error' | 'complete';
 }
 
 const ClayMultiStepNavItem = ({
@@ -38,8 +39,8 @@ const ClayMultiStepNavItem = ({
 	children,
 	className,
 	complete,
-	error,
 	expand,
+	state,
 	...otherProps
 }: IProps) => {
 	return (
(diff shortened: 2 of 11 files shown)


matuzalemsteles:
I'll implement the part in React.js

ethib137:
This is looking good! One thing, when the steps collapse into a menu, we currently don't have any visual cue to show that there is an error in one of those steps. Maybe we can change the color, like we do when one of the collapsed steps is active and add an error label or something like that in the dropdown's popup for the item that has an error. @drakonux & @marcoscv-work what do you think?

Screenshot 2023-04-03 at 1 23 31 PM
Screenshot 2023-04-03 at 1 27 57 PM

We still need to update docs too, right?

matuzalemsteles:
> Maybe we can change the color, like we do when one of the collapsed steps is active and add an error label or something like that in the dropdown's popup for the item that has an error.

Makes sense, I think we could add an error icon to the menu item.

> Oh, one other thought... I'm not sure about this one, but since it's not possible to simultaneously have a complete step and a step with an error, would it make sense to switch to a state prop that can accept error or complete as values? Would that align with our other APIs for Clay. Would that be helpful for us in the future if we decided we wanted to add a warning state or something similar?

Good one, makes a lot of sense. I'm going to switch to that and deprecate the `complete` API.

matuzalemsteles:
I sent some more changes, adding documentation and updating some examples, also moving on with the idea of using the `state` API instead of defining an API for `complete` and `error`.

packages/clay-multi-step-nav/src/Item.tsx
```diff
@@ -19,36 +19,37 @@ export interface IProps extends React.HTMLAttributes<HTMLLIElement> {
...
 	/**
-	 * Flag to indicate that the step has an error.
+	 * Flag to indicate if progress line should expand out from step
 	 */
-	error?: boolean;
+	expand?: boolean;
 
 	/**
-	 * Flag to indicate if progress line should expand out from step
+	 * Defines the status of the current step.
```
ethib137, line 32: It's probably more clear to just say "Defines the status of the step." removing "current". Since we are inside the step/item component this step may not always be the "current" step in the multistepnav.

packages/clay-multi-step-nav/src/MultiStepNavWithBasicItems.tsx
```diff
@@ -60,6 +60,11 @@ export interface IProps extends React.ComponentProps<typeof ClayMultiStepNav> {
 	 */
 	spritemap?: string;
 
+	/**
+	 * Defines the status of the current step.
+	 */
+	state?: 'error' | 'complete';
```
ethib137, line 66: I don't think we should be setting state from the parent component. That should only be set on the item since the error could be on another item than the current one. For instance in the example, when the error is set on the parent component, you can then change the current step and the error moves to whichever step is set as current.

packages/clay-multi-step-nav/docs/multi-step-nav.mdx
```diff
@@ -25,6 +28,16 @@ Each step can have two different states: `active` or `complete` defined by props
```
ethib137, line 28: The line above needs to be updated to match the new `state` prop: "Each step can have two different states: `active` or `complete` defined by props as you can see below."

ethib137 (changes requested):

packages/clay-multi-step-nav/src/MultiStepNavWithBasicItems.tsx
matuzalemsteles: This is a high-level component the developers don't have `Item` visibility, although they can define it using the `items` prop but I think it makes sense in a multi step case where the interaction is linear, normally you just move to the next step if you don't get an error I think it makes sense to keep this prop at the component level to define the state of the current step, does it make sense to you?

packages/clay-multi-step-nav/docs/multi-step-nav.mdx
```diff
@@ -25,6 +28,16 @@ Each step can have two different states: `active` or `complete` defined by props
...
+
+The active state can be set using the `active` property in the `ClayMultiStepNav.Item` or `ClayMultiStepNavWithBasicItems` component which defines the step index.
+
+The `complete` and `error` states are defined using the `state` property in the `ClayMultiStepNav.Item` which defines whether the item is complete or had an error in the current state, the `ClayMultiStepNavWithBasicItems` component follows the same API with the difference that it sets the state of the item to the current step which is set when using the `active` or `defaultActive` API.
+
+<MultiStepNavError />
+
 ## Indicator position
 
 Some use cases don't need the title or subtitle to add context to the steps, in those cases you might just want the label to be rendered to the `top` of the indicator, so it can truncate the text if it is long.
```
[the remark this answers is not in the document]
matuzalemsteles: The examples have already been updated, only the `<ClayMultiStepNav.Indicator />` we still maintain the `complete` API because we don't need to set any `error` state for this component just the `complete` state.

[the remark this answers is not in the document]
ethib137: It looks like Indicator would benefit from knowing about error state as well, so it's probably best that we update that api also to accept `state` instead of `complete`. A [this line](https://github.com/liferay/clay/blob/master/packages/clay-multi-step-nav/src/Indicator.tsx#L53) we should be checking if it's an error as well and only showing the label if `(!error && !complete)`. Right now that logic is split with the error check being [over here](https://github.com/liferay/clay/pull/5450/files#diff-ded261bc8c14e3687f770a378db7c7773320c8585ecc28c3a03823f17e33acccR157). It would be better if that was handled together otherwise you can end up with easy bugs like the image below if you're defining the `<ClayMultiStepNav.Indicator>` yourself.

Screenshot 2023-04-05 at 9 51 14 AM

packages/clay-multi-step-nav/src/MultiStepNavWithBasicItems.tsx
ethib137: Yeah I think that's fine. I've played with it a little more and the concern I had about switching steps when there's an error and thus having the error switch to a different step can be easily mitigated by disabling switching:
```
const Component = props => {
  const [active, setActive] = useState(1);
  const [error, setError] = useState(false);

  const steps = [
    {
      subTitle: "SubOne",
      title: "One"
    },
    {
      subTitle: "SubTwo",
      title: "Two Long of a title"
    },
    {
      subTitle: "SubThree",
      title: "Three"
    },
    {
      subTitle: "SubFour",
      title: "Four"
    }
  ];

  return (
  <div>
    <ClayMultiStepNavWithBasicItems
      active={active}
      onActiveChange={i => {
        if (!error) {
          setActive(i);
        }
      }}
      spritemap={spritemap}
      state={error ? 'error' : 'complete'}
      steps={steps}
    />
    
    <button className="btn btn-primary" onClick={() => setError(!error)}>Toggle Error</button>
    </div>
  );
};

render(<Component />);
```

If someone wanted to switch steps while there's an error they can always use `<ClayMultiStepNav.Item>` and manage the state at that level. So yes, I see the value in both. Sounds good to me. 👍🏻

packages/clay-multi-step-nav/docs/multi-step-nav.mdx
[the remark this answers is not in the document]
matuzalemsteles: Yeah, I'm thinking of using the context to add just to `ClayMultiStepNav.Item` so I don't need to add this API to the Indicator and we can deprecate `complete` API.

matuzalemsteles:
I sent the changes to deprecate the Indicator component's `complete` API which is now no longer needed, just need to declare the `state` prop in the Item component.

ethib137:
> I sent the changes to deprecate the Indicator component's `complete` API which is now no longer needed, just need to declare the `state` prop in the Item component.

LGTM! Thanks for the changes.

One last thing... I'm not concerned with this being changed but I'm just curious why the check icon is added in the React component but the `x` icon is added via css? Do we have a preference for how this is done generally? Is there a reason for this being different?

pat270:
>  I'm just curious why the check icon is added in the React component but the x icon is added via css?

@ethib137 The check icon is also added via CSS (They are both there if you inspect). My personal opinion is to go with CSS since it is easier to customize for OOTB components.

matuzalemsteles:
Hey guys, I don't think we have information on how the state of a step that is in error will work when it is only visible in the DropDown menu as mentioned by @ethib137, I'm thinking of moving forward with this PR and including these changes we made now for this release and we can ship the rest in another PR in the next release.

ethib137:
@drakonux or @marcoscv-work would one of you be able to provide designs for "the state of a step that is in error when it is only visible in the DropDown menu" as discussed above?

PR #5453, opened 2023-04-03
fix(@clayui/alert): move close button outside of ARIA

### LINKED ISSUE #5449: Toast Alerts SR reads "close" button
Toast Alerts (and actually all alerts with buttons) are reading the button immediately after the message, not much semantic sense, so the close button should be out of the role="alert/status"

image

matuzalemsteles:
Fixes #5449


marcoscv-work:
Great!

PR #5463, opened 2023-04-07
fix(@clayui/tabs): displayType should output nav-tabs variants

### LINKED ISSUE #5462: @clayui/tabs: The displayType property doesn't output anything and document nav-tabs-light
`displayType="light"` doesn't output the class `nav-tabs-light`.

We added a new variant in https://github.com/liferay/clay/pull/5461. We need to document it.

/cc @ethib137

pat270:
fixes #5462

### DIFF
--- a/packages/clay-tabs/docs/index.js
+++ b/packages/clay-tabs/docs/index.js
@@ -63,6 +63,68 @@ const Tabs = () => {
 	return <Editor code={tabsCode} imports={tabsImportsCode} scope={scope} />;
 };
 
+const tabsLightImportsCode = `import ClayTabs from '@clayui/tabs';
+`;
+
+const tabsLightCode = `const Component = () => {
+	const [active, setActive] = useState(0);
+
+	return (
+		<>
+			<ClayTabs active={active} displayType="light" onActiveChange={setActive}>
+				<ClayTabs.List>
+					<ClayTabs.Item
+						innerProps={{
+							'aria-controls': 'tabpanel-1',
+						}}
+					>
+						Tab 1
+					</ClayTabs.Item>
+					<ClayTabs.Item
+						innerProps={{
+							'aria-controls': 'tabpanel-2',
+						}}
+					>
+						Tab 2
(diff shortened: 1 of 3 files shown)


packages/clay-tabs/docs/index.js
```diff
@@ -63,6 +63,66 @@ const Tabs = () => {
...
+				</ClayTabs.TabPane>
+				<ClayTabs.TabPane aria-labelledby="tab-2">
+					2. Proin efficitur imperdiet dolor, a iaculis orci lacinia eu.
+				</ClayTabs.TabPane>
+				<ClayTabs.TabPane aria-labelledby="tab-3">
+					3. Proin efficitur imperdiet dolor, a iaculis orci lacinia eu.
+				</ClayTabs.TabPane>
+			</ClayTabs.Content>
+		</>
+	);
```
matuzalemsteles, line 109: @pat270 We can update these examples to use the new compositing pattern now

packages/clay-tabs/src/List.tsx
```diff
@@ -114,6 +114,7 @@ export const List = React.forwardRef<HTMLUListElement, IProps>(function List(
 			className={classNames(
 				'nav nav-tabs',
 				{'nav-justified': justified},
+				{[`nav-tabs-${displayType}`]: displayType},
```
matuzalemsteles, line 117: I think it's better to just add when the type is `light` to make sure that previous versions don't add the type to `basic` or `underline`.

```suggestion
				{[`nav-tabs-${displayType}`]: displayType === 'light'},
```

packages/clay-tabs/docs/index.js
pat270: @matuzalemsteles I'm not sure what you mean by new compositing pattern. Can you point me to an example?

matuzalemsteles: You can see the example in the storybook and do the same here https://storybook.clayui.com/?path=/story/design-system-components-tabs--new-default

PR #5466, opened 2023-04-11
fix(clayui.com): Change tab navigation to navigation bar

### LINKED ISSUE #5464: Change tabs by nav in clayui.com
Right now the tabs are refreshing the whole page, they are emulating a navigation bar:

Since tabs need to be used just for context changes we have two option:

- Load the page changes in a `tab-pane`:
image

- Change the tabs for a nav:
image

I think second option is more reasonable and easy to implement.

matuzalemsteles:
Fixes #5464

This PR changes the use of the component from ClayTabs to ClayNavigationBar, I also made some changes in the header to adjust the look to improve the weight of the fonts and spacing to visually improve the reading.

| Before | After |
|--------|--------|
| Screenshot 2023-04-11 at 16 15 48 | Screenshot 2023-04-11 at 16 15 15 |

### DIFF
--- a/clayui.com/src/styles/_guide.scss
+++ b/clayui.com/src/styles/_guide.scss
@@ -58,6 +58,10 @@
 			}
 		}
 
+		.navbar .container-fluid {
+			padding: 0;
+		}
+
 		.docs-title {
 			font-weight: 500;
 			margin: 0;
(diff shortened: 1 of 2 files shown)


ethib137:
Looks great... I like labeling `install` and `version`. Much easier to quickly grasp at a glance!

PR #5467, opened 2023-04-11
feat(@clayui/css): Sheets adds sheet-row for aligning content flush

### LINKED ISSUE #5465: Add sheet-row css class
## Context

I'm opening this issue to ask for a solution for this case:

image

As you can see, we have tabs inside a sheet, and we would like to have a way to make the tabs fit the whole width.

## What is your proposal?

After discussing it with @marcoscv-work, he suggested the ideal solution would be to have a `sheet-row` CSS class that applies the corresponding negative left and right margin (I think it's 24px for desktop and 12px for mobile). That way, we could have something like this:

image

pat270:
fixes #5465


matuzalemsteles (approved):
LGTM

PR #5468, opened 2023-04-13
feat: add motion reduction config

### LINKED ISSUE #5421: Stop JS animatons
Based on https://github.com/liferay/clay/issues/5339

We would need to stop JS animations when:

- `prefers-reduced-motion` is On in OS.
- Conditioned by a global prop, variable, etc. in order to add this conditional in a global configuration like an accessibility menu.

In order to do this and first at all, we would list all JS animations done in Clay.

matuzalemsteles:
Fixes #5421

I'm implementing the motion reduction setting for all components that use animation in JS, you need to use the `ClayProvider` and configure the `reducedMotion` API, there are 3 settings `never` will never reduce the animation regardless of the user's configuration , `always` will reduce the animation regardless of the user setting and `user` will follow the operating system setting.

This API allows that on the DXP side we just need to configure this in the ClayProvider, we can probably add this in the `react` taglib that renders all applications in React and consume the configuration from the Global Menu of accessibility configuration.

The Modal component is not consuming it yet because we cannot depend on the `Provider` package because it causes a circular dependency, we will add this dependency after we start working on solving this circular dependency problem by moving all components to the `@clayui/core` package.


matuzalemsteles:
To proceed with the implementation of reduced motion in Modal we first need to work on issue #4191 which will help resolve the circular dependency issues.

I'll just add a documentation on this, move the documentation https://clayui.com/docs/css/utilities/c-prefers-reduced-motion.html for the components section and further clarify how this can be used with JS as well.

PR #5474, opened 2023-04-17
fix(@clayui/css): Exclude variant styles from being output when using c-unset

### LINKED ISSUE #5471: @clayui/css: We should be able to exclude variants from being output by using the `c-unset` keyword
We use maps like `$btn-palette` to generate several types of variants.

https://github.com/liferay/clay/blob/eba91e6a4cc9698088c92514ac511c746769eacb/packages/clay-css/src/scss/variables/_buttons.scss#L883-L910

We should be able to prevent a variant from being output with something like:

```scss
$btn-palette: (
    info: c-unset,
);
```

Instead of throwing the error ` Error: argument `$map` of `map-get($map, $key)` must be a map`.

pat270:
fixes #5471

We can prevent variant styles from being output with this pattern:

```scss
$alert-palette: (
    primary: c-unset,
);

$btn-sizes: (
    btn-xs: c-unset,
);

$btn-monospaced-sizes: (
    btn-xs: c-unset,
);
```

previously it would throw an error.

### DIFF
--- a/packages/clay-css/src/scss/components/_alerts.scss
+++ b/packages/clay-css/src/scss/components/_alerts.scss
@@ -372,31 +372,33 @@
 // Alert Variants
 
 @each $color, $value in $alert-palette {
-	$selector: if(
-		starts-with($color, '.') or
-			starts-with($color, '#') or
-			starts-with($color, '%'),
-		$color,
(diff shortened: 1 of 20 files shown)


matuzalemsteles (approved):
LGTM

PR #5479, opened 2023-04-19
feat(@clayui/autocomplete): refactors autocomplete implementation with accessibility improvements

### LINKED ISSUE #5203: Review Autocomplete navigation behavior

matuzalemsteles:
Closes #5203

This PR is still a draft of the Autocomplete refactoring, there is a lot of work here to make the accessibility improvements because we need to make changes in the navigation and collection mechanisms, firstly because the navigation in Autocomplete is done with the virtual focus and not with the focus of the browser and second that Autocomplete has virtualization and infinite scroll capabilities this combination means that we need to rebuild the collection virtualization system and the navigation system so that both can work together.

I made some changes to the collection implementation but still incomplete, the first work I did was to finally move the virtualization implementation to the `useCollection` hook and create the `useVirtual` hook that handles the virtualization logic so we can use the two hooks together or just `useCollection` when we don't want virtualization. There are some jobs to be done:

- [x] Separate the logic of creating the layout structure and linked list from the collection to discover the keys and values of the items in the rendering system, this has to work independently otherwise in virtualization we will only register for the visible items but we want all items so the navigation system knows where to go.
- [x] Implement virtual navigation to work with a collection with virtualization
- [x] Implements the infinite scroll functionality in the `useCollection` hook, it has to work when virtualization is enabled.

Once we have these foundations we can use virtualization in Autocomplete with infinite scroll functionality. For now, Autocomplete already complies with the accessibility behaviors but we don't have the virtualization feature enabled and infinite scroll.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,10 +14,10 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-autocomplete/src/': {
-			branches: 77,
-			functions: 93,
-			lines: 92,
-			statements: 92,
+			branches: 71,
+			functions: 87,
+			lines: 89,
+			statements: 89,
 		},
 		'./packages/clay-badge/src/': {
 			branches: 0,
@@ -56,7 +56,7 @@ module.exports = {
 			statements: 91,
 		},
 		'./packages/clay-core/src/picker/': {
-			branches: 81,
+			branches: 80,
 			functions: 84,
 			lines: 90,
 			statements: 90,
@@ -194,10 +194,10 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-shared/src/': {
-			branches: 20,
+			branches: 18,
 			functions: 14,
-			lines: 36,
-			statements: 39,
+			lines: 35,
+			statements: 38,
 		},
 		'./packages/clay-slider/src/': {
 			branches: 92,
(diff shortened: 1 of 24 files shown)


matuzalemsteles:
I sent new commits that added new features to the Collection API and the navigation mechanism, now the `useNavigation` hook can be integrated with the Collection API which will work when the list is virtualized. Also the infinite scroll implementation works for a list that is not virtualized so we can add this feature to all components that implement the Collection API.

This PR is almost ready to go but I need to rewrite all the Autocomplete tests that handle the navigation interactions I used via tab, I need to move to arrow keys and add more test cases to cover the w3c recommendations use cases to avoid regressions.

Issue #5480, opened 2023-04-19
FocusTrap: Using FocusTrap component interferes with the component structure


veroglez:
Due to the structure of how the focus trap works, the `div` containing the two `span` interferes with the structure of the components.

```
<div>
	<span
		aria-hidden="true"
		data-focus-scope-start="true"
		tabIndex={0}
	/>
                {children}
	<span
		aria-hidden="true"
		data-focus-scope-end="true"
		tabIndex={0}
	/>
</div>
```
Example adding FocusTrap component in the Management Toolbar (in DXP). See that the search form does not take up the full width.

<img width="600" alt="Screen Shot 2023-04-19 at 11 00 06 AM" src="https://user-images.githubusercontent.com/9701095/233024495-c972305b-c4c9-42de-9a7d-fc7ada5b9f9d.png">

PR #5481, opened 2023-04-19
fix(@clayui/focus-trap): Add display contents to ignore the element itself

veroglez:
Fixes https://github.com/liferay/clay/issues/5480

Hi @matuzalemsteles, as we discussed yesterday, I added the `display: contents` style to the `div`, this makes the children of this element appear as if they were direct children of the element's parent, ignoring the element itself.

I have also added several changes:
- Take into account when the `focusableElements` array is empty
- A more understandable description of the component


matuzalemsteles (approved):
LGTM. Thanks @veroglez!

PR #5496, opened 2023-04-27
fix(@clayui/css): Atlas Menubar and Pagination update active and hover state to improve contrast

### LINKED ISSUE #5490: Active/Hover state in Vertical Nav and Pagination
The current style of the vertical navigation has some contrast problems for the hover and active states, this document aims to fix those problems and provide a clear way to see which navigation item is selected/hover.

_Note_: The new style has to be the replacement for the old one being applied to DXP, to correct it in DXP and propagate the change.

**Vertical Nav changes:**
Mouse Hover / Active state:

Focus in active and focus in inactive example:

**Pagination**:
Active/hover state examples:

Full documentation:
https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2337374209/Vertical+Nav+-+Review+for+accessibility

Related to LPS:
https://issues.liferay.com/browse/LPS-168530

/cc @emiliano-cicero

pat270:
fixes #5490

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_menubar.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_menubar.scss
@@ -14,6 +14,7 @@ $menubar-vertical-transparent-md: map-deep-merge(
 				background-color: rgba($gray-900, 0.04),
 				box-shadow: inset 0.125rem 0 0 0 $secondary-l0,
 				color: $gray-900,
+				font-weight: $font-weight-semi-bold,
 			),
 			focus: (
 				background-color: c-unset,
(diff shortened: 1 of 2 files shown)


matuzalemsteles (approved):
LGTM! @pat270 just a detail that I noticed the weight of the font seems to be bigger when it is in the active state, maybe we have to increase it.

pat270:
@matuzalemsteles It looks like the `font-weight` is larger for me. What are you seeing?

vert-nav

pat270:
@matuzalemsteles Did you mean hover state? I just noticed it's bolder in the screen shot.

matuzalemsteles:
@pat270 I was looking at the Pagination use case it still looks like it's font weight is less when it's in active state the Vertical Nav looks good.

pat270:
@matuzalemsteles Thanks for pointing that out, should be good to go now. I adjusted some other things like content slightly shifting on hover due to bolder font weight.

PR #5500, opened 2023-05-03
fix(@clayui/css): Modals converts some components to clay Sass map pattern

### LINKED ISSUE #5488: @clayui/css: Modals convert to use clay-css mixin pattern
Customizing components inside Modal is limited. We should convert these to use the clay-css mixin to allow better customizations.

pat270:
fixes #5488

fix(@clayui/css): Mixins clay-modal-variant should use clay-map-to-css where applicable
fix(@clayui/css): Mixins clay-map-to-css should accept parent selector &

This is the first mixin to use `clay-map-to-css`. It gives us more flexibility when customizing variants. We no longer need to declare all the subcomponents in the mixin. It will output classes based on the Sass map.

```scss
$modal-info: (
	modal-footer: (
		modal-item-last: (
			margin: 0,
		),
	),
);
```

will output:

```css
.modal-info .modal-footer .modal-item-last {
    margin: 0;
}
```

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_modals.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_modals.scss
@@ -10,15 +10,7 @@
 
 .modal,
 &.modal {
-	display: none;
-	height: 100%;
-	left: 0;
-	outline: 0;
-	overflow: hidden;
-	position: fixed;
-	top: 0;
-	width: 100%;
-	z-index: $cadmin-zindex-modal;
+	@include clay-css($cadmin-modal);
 }
 
 .modal-dialog {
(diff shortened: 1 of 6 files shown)


matuzalemsteles:
I don't really like this way of allowing customization, I have the feeling that we are just creating a different syntax for declaring CSS classes from a set... because basically we can't impose limits and if we need to maintain compatibility things get more difficult and complex with the time.

pat270:
@matuzalemsteles 
Yes this is doing just that. This is coming from the perspective of an end user wanting to implement a different design with the components that we currently offer. It's very helpful to be able to target specific selectors in our markup to get what you want. The limitation is the HTML we provide. If you want to reuse our components, you need to adhere to the structure. I think in the future it will prove to be useful when editing from Figma or when we transition to css variables for example.

matuzalemsteles (approved):
@pat270 yeah I see, I think the problem is exactly all the flexibility that this opens up by allowing you to generate any final output. But we can add it little by little and see how it grows.

Unrelated to that but I think you can add `feat` here instead of `fix` since that would be more of a feature than a bug fix.

Issue #5487, opened 2023-04-21
@clayui/css: More complex input-groups should accommodate inset focus styles


pat270:
In the search bar example at https://clayui.com/docs/components/autocomplete/markup.html#css-loading, we manage the focus styling by adding the `focus` class to the parent `input-group-item` element. If we use an inset `box-shadow` the focus styles are no longer visible due to it being covered by the background of the input and buttons. We should move the focus style to a pseudo element to accommodate more types of focus borders.

Edit:
The clayui.com example doesn't add the `focus` class on `input-group-item`, but you can see it if you manually add it via dev tools. The bug we want to fix:

```css
.input-group-item.focus {
  box-shadow: inset 0 0 0 0.125rem #528eff, inset 0 0 0 0.25rem #fff;
}
```

doesn't show the focus outline. It's hidden beind the `input`. We can fix it by moving the focus shadow to a pseudo element using:

```css
.input-group-item.focus::after {
  border-radius: inherit;
  bottom: 0;
  box-shadow: inset 0 0 0 0.125rem #528eff, inset 0 0 0 0.25rem #fff;
  content: '';
  display: block;
  left: 0;
  pointer-events: none;
  position: absolute;
  right: 0;
  top: 0;
}
```

The component we need to update is located [here](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/components/_input-groups.scss#L129-L131). It uses [this mixin](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/mixins/_input-groups.scss#L296). The properties are generated from the values [here](https://github.com/liferay/clay/blob/master/packages/clay-css/src/scss/variables/_forms.scss#L1428-L1471).

We will need to move the `box-shadow` property to:

```
focus: (
  border-radius: clay-enable-rounded($input-border-radius),
  after: (
    box-shadow: $input-focus-box-shadow,
  ),
),
```
and it should generate the css:

```css
.input-group-item.focus::after {
  border-radius: inherit;
  bottom: 0;
  box-shadow: 0 0 0 0.125rem #fff, 0 0 0 0.25rem #528eff;
  content: '';
  display: block;
  left: 0;
  pointer-events: none;
  position: absolute;
  right: 0;
  top: 0;
}
```

SelenaAungst:
PR: https://github.com/liferay/clay/pull/5510

PR #5510, opened 2023-05-05
@clayui/css: More complex input-groups should accommodate inset focus styles

SelenaAungst:
Solution for https://github.com/liferay/clay/issues/5487

### DIFF
--- a/packages/clay-css/src/scss/mixins/_input-groups.scss
+++ b/packages/clay-css/src/scss/mixins/_input-groups.scss
@@ -300,6 +300,14 @@
 		@if ($enabled) {
 			@include clay-css($map);
 
+			&::before {
+				@include clay-css(map-get($map, before));
+			}
+
+			&::after {
+				@include clay-css(map-get($map, after));
+			}
+
 			&.focus {
 				@include clay-input-group-item-variant(
 					map-deep-get($map, focus)
--- a/packages/clay-css/src/scss/atlas/variables/_forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_forms.scss
@@ -1,5 +1,6 @@
 $input-bg: $gray-200 !default;
 $input-border-color: $gray-300 !default;
+$input-border-radius: $border-radius !default;
 $input-border-width: 0.0625rem !default; // 1px
 
 $input-border-bottom-width: $input-border-width !default; // 1px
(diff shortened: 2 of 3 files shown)


packages/clay-css/src/scss/mixins/_input-groups.scss
```diff
@@ -304,6 +304,20 @@
 				@include clay-input-group-item-variant(
 					map-deep-get($map, focus)
 				);
+
+				&::after {
```
pat270, line 308: We need to move the `&::after` rule-set to line 302. It should get the properties from the `$map` via `map-get($map, after)`. We should have a `&::before` rule-set also. We need to use the clay-css mixin, `@include clay-css(map-get($map, after))`, to print out the properties listed in `after` and `before`. The nested maps in `focus` will print because we are passing it back into the `clay-input-group-item-variant` mixin.

pat270 (changes requested):

packages/clay-css/src/scss/mixins/_input-groups.scss
SelenaAungst: Hi @pat270,
I attempted to add these changes, please let me know if this is correct!

packages/clay-css/src/scss/mixins/_input-groups.scss
```diff
@@ -299,6 +299,22 @@
 
 		@if ($enabled) {
 			@include clay-css($map);
+			@include clay-css(map-get($map, after));
```
pat270, line 302: @SelenaAungst Thanks for the update, the lines 302 to 317 should be:
```scss
&::before {
	@include clay-css(map-get($map, before));
}

&::after {
	@include clay-css(map-get($map, after));
}
```

What this does is gets the `before` and `after` keys from the map `$input-group-item`. We passed the map contained in `$input-group-item: (focus: ())` right below on line 320. This mixin will print out the CSS as:

```css
.input-group-item.focus::after {
  ...
}
```
since that is all we are passing in from the variables in `_forms.scss`.

If we passed:

```scss
$input-group-item: (
    before: (
        color: red,
    ),
);
```

it would output:

```css
.input-group-item::before {
    color: red;
}
```

pat270 (changes requested):

packages/clay-css/src/scss/mixins/_input-groups.scss
SelenaAungst: Hi @pat270,
Is there anything else I can do for this?

pat270:
@SelenaAungst Thanks for this pr! Sorry I took so long to look at it. I made some slight adjustments. We didn't need to update anything in the Atlas Theme since it inherits everything from the Base Theme. The focus transitions weren't working so I moved most of the `after` code to `.input-group-item::after`.

pat270 (approved):

PR #5526, opened 2023-05-12
feat(@clayui/nav): add new composition to Vertical Nav with collection API and adds new `expandedKeys` API

### LINKED ISSUE #5515: Is there any way to open all or close all panels of the Vertical Nav?
When defining permissions to a role in Liferay, the current behavior when filtering the vertical nav in the sidebar is to open all sub panels while filtering, and close all panels when done. Do we have a way of doing this with the current vertical nav? 
Screenshot 2023-05-09 at 9 01 51 PM

matuzalemsteles:
Close #5515

This PR makes a lot of changes, first we are rewriting the `VerticalNav` component to use the Collection API and exposing a new public `<VerticalNav.Item />` component.

We are also adding a new `expandedKeys` API with the pattern of controlled and uncontrolled, this makes it easier to manage this state for items in the tree like implementing expand all or collapse.

We are also moving the implementation of all this to the `@clayui/core` package and the `@clayui/nav` package just keeps backwards compatibility with the old version of the component and possibility to use the new version when declaring the new composition.

With the Collection implementation I needed to rewrite the component and this brings a new composition with the benefits of the Collection pattern.

```jsx
<VerticalNav items={items}>
    {(item) => (
        <VerticalNav.Item active={item.active} items={item.items}>
            {item.label}
        </VerticalNav.Item>
    )}
</VerticalNav>
```

This brings new possibilities to the component as it is also data agnostic and does not depend on a data structure to force the product to follow. As I said, we still maintain compatibility with the previous version, so it's easy to migrate to use the new composition and the new features that Collection brings, also the `expandedKeys` API will only take effect with the new composition because it depends on which keys in the tree is unique. Consuming the component directly from the `@clayui/core` package will only support the new composition.

### ToDo

- [x] Update docs

### DIFF
--- a/packages/clay-core/src/vertical-nav/Item.tsx
+++ b/packages/clay-core/src/vertical-nav/Item.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {useProvider} from '@clayui/provider';
 import {Keys, setElementFullHeight} from '@clayui/shared';
 import React, {useContext, useMemo, useRef, useState} from 'react';
 import {CSSTransition} from 'react-transition-group';
@@ -92,6 +93,8 @@ export function Item<T extends object>({
 		spritemap,
 		toggle,
 	} = useVertical();
+
+	const {prefersReducedMotion} = useProvider();
 
 	const itemRef = useRef<HTMLButtonElement | HTMLAnchorElement>(null);
 	const menusRef = useRef<HTMLUListElement | null>(null);
@@ -208,7 +211,7 @@ export function Item<T extends object>({
 					onExiting={(element) =>
 						element.setAttribute('style', 'height: 0px')
 					}
-					timeout={250}
+					timeout={prefersReducedMotion ? 0 : 250}
 				>
 					<Nav ref={menusRef} role="menu" stacked>
 						<ParentContext.Provider value={itemRef}>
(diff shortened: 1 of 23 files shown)


matuzalemsteles:
cc @ethib137 I think this will help you implement the collpase mechanism and I added an example in the storybook using the new `expandedKeys` API.

ethib137:
Thanks @matuzalemsteles ! What do you think about adding `expanded` as a prop to `navItem` so that expanded can be totally controlled? I think that might be even simpler and more flexible, since to get all the keys for the expanded keys api I would still have to loop through all nested items. If I'm doing that, I might as well just set expanded myself directly on the item. What do you think?

packages/clay-core/src/vertical-nav/Item.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+	const {prefersReducedMotion} = useProvider();
+
+	const itemRef = useRef<HTMLButtonElement | HTMLAnchorElement>(null);
+	const menusRef = useRef<HTMLUListElement | null>(null);
+	const parentItemRef = useContext(ParentContext);
+
+	// State only for compatibility with the old compositing version where state
+	// was kept on each Item instead of centralized with `expandedKeys`.
+	const [expanded, setExpaned] = useState(initialExpanded);
```
ethib137, line 105: setExpaned -> setExpanded

matuzalemsteles:
> What do you think about adding expanded as a prop to navItem so that expanded can be totally controlled? I think that might be even simpler and more flexible, since to get all the keys for the expanded keys api I would still have to loop through all nested items. If I'm doing that, I might as well just set expanded myself directly on the item. What do you think?

@ethib137I think this would cause more complexity, i don't see this being simpler for your case, for example if you want to expand all items you would traverse the tree when you need to expand to collect all expanded items and when close you just sets the state of `expandedKeys` to `new Set()` with an empty value. You can do this when making the filter to avoid going through more than once.

ethib137:
Okay, sounds good. I'll try it out that way then. Thanks @matuzalemsteles !

PR #5531, opened 2023-05-16
fix(@clayui/css): fix syntax error

matuzalemsteles:
This bug was discovered with build failure when updating Clay in DXP https://github.com/liferay-peds/liferay-portal/pull/53.

It's strange that this is not reproducible in Clay's build, maybe our build is not throwing the console errors or failing the build correctly, we need to investigate if this is an issue in our build or the DXP later.


pat270:
LGTM! Thanks @matuzalemsteles

PR #5532, opened 2023-05-17
fix(@clayui/core): fixes bug with not having a focusable item when there is no active item

### LINKED ISSUE #5516: When Vertical Nav does not have any active items it is not focusable by the keyboard.
https://codesandbox.io/s/cool-cloud-dzkbcx?file=/src/index.js:256-282

1. Go to https://codesandbox.io/s/cool-cloud-dzkbcx?file=/src/index.js:256-282
2. Click on the button
3. Press Tab

**Expected Result**
- Pressing tab should focus the first item of the vertical Nav when no item is active.

**Actual Result**
- It's not possible to focus the vertical nav when there is not an active item.

matuzalemsteles:
Fixes #5516

This PR makes more changes besides fixing the related bug in the issue, as part of being data agnostic I'm deprecating Item's `active` API instead it's just necessary to set the state of `active` in the component root, this makes it simpler than traversing the tree and setting the property to `active` it also lessens the data payload.

I had to make some internal changes to the Collection API because of how Collection nesting works in this component which is different from components like Picker or DropDown. I added comments in the code about some specs.

I still have to do some more backwards compatibility testing and fix the tests.

### DIFF
--- a/packages/clay-core/src/collection/types.ts
+++ b/packages/clay-core/src/collection/types.ts
@@ -46,6 +46,7 @@ export type CollectionState = {
 	getItem: (key: React.Key) => {value: string; index: number};
 	getItems: () => Array<React.Key>;
 	getLastItem: () => {key: React.Key; value: string; index: number};
+	hasItem: (key: React.Key) => boolean;
 	size?: number;
 	virtualize: boolean;
 };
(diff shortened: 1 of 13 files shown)


matuzalemsteles:
I sent the final commits that maintain compatibility when using the item's `active` API i.e. the behavior of having the focus on the first item when there is no active item also works when not using the `active` property in the root of the component. This behavior will be removed in the major release.

PR #5540, opened 2023-05-25
feat(@clayui/multi-select): adds implementation of new autocomplete

### LINKED ISSUE #5266: Uses the new Autocomplete component in MultiSelect
The MultiSelect component is getting more accessibility improvements, this starting with the grid pattern for the labels with issue #5194 but we still need to apply the combobox pattern to the input. Before starting work on this issue we first need #5203 to be resolved.

Try to use the new Autocomplete component if possible otherwise we can use the hooks that deal with navigation and positioning and the Overlay component so it will be easier to apply the combobox pattern to the input without having to modify something in the public API for MultiSelect since it has a different composition.

matuzalemsteles:
Closes #5266

This PR brings the Autocomplete implementation to MultiSelect, so all accessibility benefits are present in MultiSelect as well.

There are some changes here to bring the implementation since Autocomplete has the implementation of Collection MultiSelect also benefits from this but it still doesn't fully affect your composition, for example building a custom menu (`menuRenderer`) has been deprecated in favor of composition with render props in `children`, the examples in the storybook and docs have been updated.

Also MultiSelect's asynchronous loading mechanism is simpler due to the implementation of Autocomplete, bringing new APIs like `loadingState` and `onLoadMore`.

MultiSelect still needs more changes but that will only be possible in the next major version as we won't be able to maintain compatibility of some things. It is very likely that we would need a new composition for MultiSelect to allow the labels to be customized as well.

### ToDo

- [x] Update tests
- [x] Update docs


matuzalemsteles:
I'm going ahead and merging this, there is still a strange bug that I'm not able to identify when clicking on the Autocomplete item needs to be double-clicked, it doesn't happen on Autocomplete outside of MultiSelect, I'm investigating further and sending another follow-up PR as soon as I can find out the problem.

PR #5542, opened 2023-05-26
feat(@clayui/card): moves root `aria-label` to card titles

### LINKED ISSUE #5529: Possibility of specify different aria-label for Cards
## Context
Hi guys! 😄 This is a follow-up of https://github.com/liferay/clay/issues/5425. 
In that issue, I asked for the possibility of adding an `aria-label` for all Cards component. As far as I could see, the fix has been adding as `aria-label` the same value we have for `title`. I think that in some cases, we need to add a different value for the `aria-label`, for example if we want to add extra info that the `title` does not have.

## What is your proposal?
For that reason, I think it would be ideal to be able to specify the `aria-label` as a prop, something like `titleAriaLabel`, so we can give it a different value. 

## Example
I'm giving our case as example. For this card, the title says "Styles from Theme", but for the `aria-label` we would like to have "Styles from Theme (marked as default)". We could achieve that if the `aria-label` itself was a prop for Card components, as I suggested.

image

Please let me know your thoughts, thanks!

matuzalemsteles:
Closes #5529

Instead of creating a new API to change the `aria-label` of the title we are using the existing `aria-label` property instead of adding it to the root of the component which will be a div which will have no effect it makes more sense to move this to the titles of the cards.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -38,7 +38,7 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-card/src/': {
-			branches: 85,
+			branches: 82,
 			functions: 90,
 			lines: 95,
 			statements: 95,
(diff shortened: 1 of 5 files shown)


matuzalemsteles:
cc @sandrodw3

PR #5543, opened 2023-05-26
refactor(@clayui/core): refactors the implementation of `useInternalState` to `useControlledState` to follow React behavior

### LINKED ISSUE #5536: `onChange` callback doesn't fire for "uncontrolled" components
[Internal discussion on slack](https://liferay.slack.com/archives/C049XG98GQL/p1684305368240679)

Codesandbox: https://codesandbox.io/s/epic-sound-cgi4jo?file=/src/index.js

**Expected:**
Clay components that use `useInternalState` will allow for onChange to be called even if a `value` is not provided.

**Actual:**
`onChange` does not fire if a `value` is not provided. See [DatePicker CSB example](https://codesandbox.io/s/epic-sound-cgi4jo?file=/src/index.js)

matuzalemsteles:
Fixes #5536

I renamed the hook name to follow the terminology of controlled and uncontrolled, also making the changes to allow adding the handler without the value and keep calling but using the internal state, we also throw a warning in the console when the state changes from controlled to uncontrolled and uncontrolled to controlled, similar to React.js.


bryceosterhaus:
Thanks for this @matuzalemsteles !

PR #5545, opened 2023-05-31
fix(@clayui/css): Pagination and Menubar update hover and active borders

### LINKED ISSUE #5535: Vertical nav and pagination border to be a straight shape
no

you can see this in any page with a pagination for example, currently the border is taking a circular shape around the angle that is a bit weird as you can see in this picture: 
image

it would be nice to have the internal part to be a straight line and only the external to be a border as shown in this examples:
image

image

## Environment
Master

| Tech  | Version |
| ----- | ------- |
| Clay  | v?.?.?  |
| React |         |

### LINKED ISSUE #5539: Vertical Navigation Dance
We should try to avoid this dance effect, probably adding the side decorator in an after element or something similar.

https://github.com/liferay/clay/assets/7963804/1d88327a-d5cf-48b8-b92f-812be331fe0d

It may be possible to do it at the same time as https://github.com/liferay/clay/issues/5535

pat270:
vertical-bar

pagination-hover

fixes #5535, fixes #5539

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_menubar.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_menubar.scss
@@ -5,16 +5,30 @@ $menubar-vertical-transparent-md: map-deep-merge(
 	(
 		nav-link: (
 			border-radius: $border-radius,
+			letter-spacing: 0.02em,
+			margin-bottom: 0.25rem,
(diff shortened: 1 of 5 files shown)


matuzalemsteles (approved):
LGTM

PR #5548, opened 2023-06-01
chore: update CI status badge

matuzalemsteles:
I don't think we ever got to update the CI status badge, because it's been a while since we moved our CI to GitHub Actions and it was still showing Travis CI information, strange it was still working but now we are showing the CI status from Github Actions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
         </h1>
     </div>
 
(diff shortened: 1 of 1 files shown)


matuzalemsteles:
cc @pat270

Issue #5551, opened 2023-06-05
TreeViewItem is not being read correctly with the screen reader


veroglez:
Hi guys!

I open this bug in relation to this issue https://issues.liferay.com/browse/LPS-186641 that there is in DXP using the TreeView component.

When you navigate with the keyboard to one of the pages, the screen reader reads "Drag Home" and when the drag button is reached, only "button" is read (you can see it in the attached images). 

<img width="300" alt="actual_result_1" src="https://github.com/liferay/clay/assets/9701095/361216e4-2ae6-4655-a345-16c23f663140">

<img width="308" alt="actual_result_2" src="https://github.com/liferay/clay/assets/9701095/59cfcc33-d3da-4602-a314-adfc8519b6e9">

This is incorrect, because when you reach the page, it should read just the name, in this case "Home" and when you reach the button, it should read "Drag Home Button".

veroglez:
Investigating, it seems that this is happening because when the `child`, which in this case is a `<div>` is checked in [this logic](https://github.com/liferay/clay/blob/dbd8d94c315c484918f0377541ce2ee82e830d54/packages/clay-core/src/tree-view/TreeViewItem.tsx#L759), the `labelId` is not added as `id`. So, when the element that has `aria-labelledby=labelId`  is reached, an incorrect read is done.

For this reason, I am sending this PR https://github.com/liferay/clay/pull/5552, where the `child` keeps all its `props` if passed, and if not, `labelId` as id will be applied by default.

Thanks in advance!

PR #5552, opened 2023-06-05
fix(@clayui/core): Add the missing id so that the screen reader makes a correct reading

veroglez:
Fixes https://github.com/liferay/clay/issues/5551

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -822,6 +822,11 @@ export function TreeViewItemStack({
 							}
 						},
 					});
+				} else {
+					content = React.cloneElement(child as React.ReactElement, {
+						...(child as React.ReactElement)?.props,
+						id: (child as React.ReactElement)?.props?.id || labelId,
+					});
 				}
 
 				return (


matuzalemsteles (approved):
LGTM! Thanks @veroglez for catch this use case, do you have the example of the code that is falling in this use case? maybe it's rendering the wrong spacing because it's not using `component-text`.

veroglez:
@matuzalemsteles yes, this is a use case and it seems that it's actually rendered with a wrong spacing. I have also added the `component-text` class to the `child`. Thanks for this!

Without `component-text`
wrong_spacing

With `component-text`
correct_spacing

matuzalemsteles:
@veroglez I think the component may be using the wrong composition to not fall into the condition of being a `string`, I don't know if it's a good idea to add this to `else`, because this could be any component and we can break the implementation with that . Do you have the example of the Page Tree case that is rendered wrong?

veroglez:
Sure! Here is the Page Tree in DXP https://github.com/liferay/liferay-portal/blob/08ccd16f507ed322a427c3d2eab3d5b7964511e3/modules/apps/product-navigation/product-navigation-product-menu-web/src/main/resources/META-INF/resources/js/PagesTree.js#L233

Screen Shot 2023-06-06 at 11 33 12 PM

Screen Shot 2023-06-06 at 11 33 28 PM

matuzalemsteles:
I see, thanks @veroglez! It's something custom I think in this case we can avoid passing the `component-text` because we can't infer in this case what type the element is and it can cause more problems for other use cases, let's just pass the `id` in this case.

I think you can add the best class directly to your custom, for example:

```jsx
{item.regularURL ? (
  <a
    className="component-text text-truncate"
    data-tooltip-floating="true"
    href={item.regularURL}
    ref={itemAnchorRef}
    tabIndex="-1"
    target={item.target}
    title={item.name}
  >
    {item.name}
  </a>
) : item.name}
```

This simplifies the implementation a bit and removes the use of the `div` and also the second condition will go into the string case of the TreeViewItem.

veroglez:
yes, I agree with that. Better not add the class if there are custom elements. I update the PR with the changes. On the other hand, in DXP, I will simplify the node as much as possible. Thanks for the feedback!

matuzalemsteles:
Thanks @veroglez for that!

PR #5556, opened 2023-06-06
fix(@clayui/css): Menubar removes negative margins in mobile and sync…

### LINKED ISSUE #5521: Vertical Nav add negative margin on mobile view with menubar-vertical-expand-lg class
https://codesandbox.io/s/vertical-nav-margin-issues-g9uex8?file=/src/index.js

1. Go to https://codesandbox.io/s/vertical-nav-margin-issues-g9uex8?file=/src/index.js
2. Make sure the window is small enough to display the mobile view.

The menu and trigger should fit inside the window.

Actual Result:
The menu and trigger have negative margin and are overlapped by the window.

Screenshot 2023-05-10 at 4 30 10 PM

pat270:
… up Cadmin styles with Atlas

fixes #5521 

@ethib137 I ended up removing the negative margins based on @emiliano-cicero's screenshot. It looks like this:

vertical-nav

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_menubar.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_menubar.scss
@@ -114,7 +114,7 @@ $menubar-vertical-transparent-md: map-deep-merge(
 					),
 				),
 				menubar-toggler: (
-					color: $gray-900,
+					color: $gray-600,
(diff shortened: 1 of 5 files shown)


ethib137:
Thanks @pat270 ! Yeah, I think that looks much better. 😄

PR #5557, opened 2023-06-07
feat(@clayui/date-picker): add Picker component to select year

### LINKED ISSUE #5537: New design for the year select inside the date picker
## What is your proposal?
Update the year select inside the date picker for a better user experience, currently even if it's possible to include a higher range of years, it's difficult to find those years in the picker, this proposal should fix that. 

It would allow to select a higher range of years and improve the current user experience of this component.
Also it would increase the consistency overall, right now we have at least 3 different styles and they change based on the browser.

none 

## Jira Ticket 
https://issues.liferay.com/browse/LPS-179007

## Design Deliverables 
[Figma Mockup](https://www.figma.com/file/G1GKqOshs3v3c4BKAzP7ky/lps-184297-update-the-design-for-the-date-picker?type=design&node-id=1%3A455&t=EAPZKLJKu8QYhv7N-1)

[Confluence Documentation](https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2363589035/Year+Select)

matuzalemsteles:
Closes #5537

I still need to make some CSS adjustments to better deal with the arrows that should be floating @pat270 maybe you can help me with that here. Still need to understand what happens when clicking on the arrow options there is no description of the behavior of this in the documents cc @emiliano-cicero.

Instead of creating a DatePicker specific component I added an internal property that adds these new behaviors to the Picker component when enabled to avoid duplicating code or creating rules to try and share as resources are minimal, this seems like a better strategy to maintain the code long-term.

### ToDo

- [x] Add the arrow button behaviors
- [x] Adjust CSS of the arrow buttons

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -56,10 +56,10 @@ module.exports = {
 			statements: 91,
 		},
 		'./packages/clay-core/src/picker/': {
-			branches: 80,
-			functions: 84,
-			lines: 90,
-			statements: 90,
+			branches: 59,
+			functions: 66,
+			lines: 77,
+			statements: 77,
 		},
 		'./packages/clay-core/src/tree-view/': {
 			branches: 56,
@@ -74,7 +74,7 @@ module.exports = {
 			statements: 85,
 		},
 		'./packages/clay-date-picker/src/': {
-			branches: 56,
+			branches: 55,
 			functions: 89,
 			lines: 81,
 			statements: 82,
--- a/packages/clay-core/src/picker/Picker.tsx
+++ b/packages/clay-core/src/picker/Picker.tsx
@@ -519,7 +519,7 @@ export function Picker<T>({
 						className={classNames(
 							'dropdown-menu dropdown-menu-indicator-start dropdown-menu-select show',
 							{
-								'dropdown-menu-width-shrink':
+								'dropdown-menu-height-lg dropdown-menu-width-shrink':
 									UNSAFE_behavior === 'secondary',
 							}
 						)}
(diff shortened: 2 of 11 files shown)


pat270:
I'll take a look at this.

pat270:
@matuzalemsteles I wasn't able to figure out a CSS only solution for the floating arrows today. We're losing focus on the scrollbar due to React re-rendering the dropdown menu similar to what we are experiencing with the range sliders at https://github.com/liferay/clay/pull/5174.

matuzalemsteles:
Hey @pat270 I think something just floating that doesn't change the size of the menu when the button disappears I think it would be enough because we have a JS implementation that will handle when to show the button.

Could you show a video of how this losing focus so I can investigate this further?

pat270:
@matuzalemsteles This is what I'm experiencing.

year-picker

Having it float outside the dropdown menu should work, but it feels a bit fragile.

pat270:
@matuzalemsteles Floating the buttons makes the scroll work, but there is a usability issue when clicking the up and down arrows. When the scroll reaches the end of the list, the button disappears and you end up clicking the content underneath.

year-picker-floating

matuzalemsteles:
Yeah, I think I'd have to put the floating buttons on top of the list with a larger `z-index` so the menu size wouldn't be influenced. What do you think?

For example, could buttons with `position: absolute;` and `top: 0;` or `bottom: 0;` be a solution?

pat270:
@matuzalemsteles in my second gif, I'm using absolute with the buttons in place. It works fine except for the issue when the button disappears.

matuzalemsteles:
>  in my second gif, I'm using absolute with the buttons in place. It works fine except for the issue when the button disappears.

Oh I didn't notice that, are the buttons inside the list or outside? maybe that's what makes it lose the scroll also it seems that the size of the menu changes when the buttons disappear I thought that would not happen.

pat270:
I didn't move the buttons, changing the size of the dropdown menu causes the re-render.

Issue #5562, opened 2023-06-09
Passing a `children` function to MultiSelect behaves oddly


bryceosterhaus:
## Is there an example you can provide via codesandbox.com?
https://codesandbox.io/s/quick-start-clay-finale-forked-jq647v?file=/src/index.js

## What are the steps to reproduce?
Open multi-select menu and click on item(not on the checkbox)

## Result
- First click does nothing, other than lose focus of input
- [second click calls internal onClick which selects the item](https://github.com/liferay/clay/blob/5b18df3b0ec2e65c62d88c2cca82c3b9f64d51e6/packages/clay-multi-select/src/index.tsx#L320-L327)

## What is the expected result?
- the onClick should be defined by the user and not provided internally.
- Input shouldn't get blured when clicking on `ClayMultiSelect.Item` unless defined by the user
- User should be responsible for closing the drop-down, possibly through a prop in the children function. This way the user can keep the menu open if needed (clicking multiple items in a row).

I would expect the API to look like this

```jsx
<ClayMultiSelect
	items={items}
	onItemsChange={setItems}
	sourceItems={sourceItems}
>
	{(sourceItem, closeDropDown) => {
		<ClayMultiSelect.Item
                     onClick={()=> {
                         setItems([...items, sourceItem]);
                         closeDropDown();
                     }}
                  >
              // ...
		</ClayMultiSelect.Item>
	}}
</ClayMultiSelect>
```

Maybe I am missing something here...

bryceosterhaus:
One additional issue I noticed, when using the keyboard to navigate multi-select items (even with the OOTB multi-select), when you hit enter, the event propagates up to the form and submits the form.

matuzalemsteles:
@bryceosterhaus Thanks for reporting these issues I'll look into this, most likely this came after we added the implementation of the new Autocomplete accessibility behavior to MultiSelect.

matuzalemsteles:
@bryceosterhaus just a few details about the expected results, some of these behaviors would not be ideal to pass via render props because it would be a prop drilling issue as well because in that case you have control of the state since this is not a default behavior of MultiSelect with Autocomplete , so the default behaviors are close on click because you are selecting an option according to the accessibility rule and the focus should go back to the input with the selected value so we implement `onClick` OOTB because we also control other rules, but you can add your own rule when clicking on the item and ignoring the default behavior, you can add the `onClick` on the item and call the `event.preventDefault()` method that ignores the default behavior of the item.

[...]

PR #5570, opened 2023-06-13
feat(@clayui/multi-select): adds the possibility to ignore the item's default behavior

matuzalemsteles:
Related #5562

Basically this implements some improvements and fixes as I mentioned in issue #5562, adds the behavior of being able to ignore the default behavior of the component when clicking on the `Item` and explicitly adds the types to control the `active` state of the menu.

A small change was to deprecate the `liveRegion` and `hotkeysDescription` APIs to use the `messages` API which has the same goal to add the same API which has the functionality to handle translate. In the future, an idea is to group all this on the `messages` API in all components with the same behavior and we will consume this by ClayProvider so that all components can share and easier to integrate with the DXP internationalization system so that the developers don't have to worry about adding this.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -142,8 +142,8 @@ module.exports = {
 		'./packages/clay-multi-select/src/': {
 			branches: 52,
 			functions: 57,
-			lines: 65,
-			statements: 65,
+			lines: 62,
+			statements: 63,
 		},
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
I'm going ahead and merging this to add in the release

PR #5579, opened 2023-06-15
feat(@clayui/css): Utilities adds more text and bg color utilities

### LINKED ISSUE #5568: Add extra util classes for background colors
Currently, we have some background css classes like `bg-warning`, `bg-success` but there are other clay colors that are not included like `orange`

Would it be possible to include these classes? Currently I'm working on a story [figma](https://www.figma.com/file/gHUmoRO2cqYwBqu5k3X4EV/LPS-179532--Configure-which-fragments-are-automatically-mapped-to-form-containers?type=design&node-id=39-194214&t=j39pyc3z65N8EOvU-0) where I need to set the background of a sticker specifically to orange, and it seems a bit cumbersome to write custom css just for that

Thanks!

pat270:
fixes #5568

### DIFF
--- a/clayui.com/content/docs/css/content/typography.md
+++ b/clayui.com/content/docs/css/content/typography.md
@@ -294,30 +294,100 @@ Milk filter lungo as galão roast that crema
 ## Contextual Texts(#css-contextual-texts)
 
 <div class="sheet-example">
-    <div>
-        <a href="#1" class="text-muted">text-muted: Milk filter lungo as galão roast that crema blue mountain shop turkish.</a>
-    </div>
-    <div>
-        <a href="#1" class="text-primary">text-primary: Milk filter lungo as galão roast that crema blue mountain shop turkish.</a>
-    </div>
-    <div>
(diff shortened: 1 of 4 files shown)


ethib137:
Hey @pat270 , wouldn't it make sense to document this on the [color utilities page](https://deploy-preview-5579--next-clayui.netlify.app/docs/css/utilities/markup-color-utilities.html) as well? Also, if we are adding these colors for text and background, should we add it for border too? It looks like up until now we have kept parity between the utilities for text, bg, and border-color, right?

pat270:
@ethib137 Thanks for pointing that out, I'll update there too.

pat270:
It should be updated now.

ethib137:
Just out of curiosity, what's our reasoning for not including border color utilities?

pat270:
@ethib137 only reason is I don't want to keep increasing the size of Clay CSS. If you think we should add it, we can.

ethib137:
Yeah, that seems like a fine reason for now. If we end up with someone requesting it for a use case we can reconsider it. Thanks @pat270 .

matuzalemsteles (approved):

PR #5587, opened 2023-06-16
fix(@clayui/autocomplete): fixes type inference for properties and generic type

### LINKED ISSUE #5577: Autocomplete and MultiSelect do not have the type inference to the `items` APIs
[Slack thread](https://liferay.slack.com/archives/C049XG98GQL/p1685342236873719)

The `<Autocomplete />` and `<MultiSelect />` components are giving type inference errors when the developer tries to control the states of `items` for example:

```jsx
<MultiSelect
  items={items}
  onItemsChange={setItems} // <- TypeScript error
/>
```

This problem happens because both components have generic types but use `React.forward` which types are `unknown`. We need to improve the typing for these cases so that the inference works correctly without having to import Clay types just inferring the state types.

matuzalemsteles:
Fixes #5577

Generic types for the component along with `React.forwardRef` didn't work as expected so I had to do some different strategies to get the types to work correctly so the `<Autocomplete />` wasn't inferring the correct properties of the component that was fixed now. `<MultiSelect />` and `<Autocomplete />` now have type inference for controlled and uncontrolled state APIs just like other components, for example:

```jsx
<MultiSelect
  items={items}
  onItemsChange={setItems} // <- It works
/>
```

I'm still marking this PR as a draft because it doesn't correct the inference in case the controlled state is an anonymous function, the value is not being inferred, I still don't understand why this is happening, so I'll keep investigating.

```jsx
<MultiSelect
  items={items}
  onItemsChange={(items) => setItems(items)} // <- items is any
/>
```

### DIFF
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -38,12 +38,7 @@ type ItemProps<T> = {
 	keyValue: React.Key;
 };
 
-export interface IProps<T>
-	extends Omit<
-			React.HTMLAttributes<HTMLInputElement>,
-			'onChange' | 'children'
-		>,
-		Omit<Partial<ICollectionProps<T, unknown>>, 'virtualize'> {
+export type IProps<T> = {
 	/**
 	 * Flag to indicate if menu is showing or not.
 	 */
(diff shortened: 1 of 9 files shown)


matuzalemsteles:
I did some investigation and it seems that the type inference for the anonymous function seems to be some TypeScript error when the type is imported, for example the types with `InternalDispatch<T>` for the anonymous function the first parameter it is always being inferred like `any` but when we add the type of `InternalDispatch` in the same component file the type of the parameter is being inferred correctly.

I will continue with this PR that already improves the typing of the entire component for MultiSelect and Autocomplete, for anonymous functions we can improve this later but it would not be a blocker because the developer can add its type based on its data and it would no longer be necessary to import Clay's built-in types.

PR #5593, opened 2023-06-21
fix(@clayui/autocomplete): fix bug not announcing available options when autocomplete is open

### LINKED ISSUE #5590: Announce results of Autocomplete component with LiveAnnouncer
There is a DXP issue related: https://issues.liferay.com/browse/LPS-168537

According to the conversation here https://github.com/liferay/clay/issues/5266#issuecomment-1404895768, we need to announce the result number of Autocomplete results to the SR with LiveAnnouncer.

matuzalemsteles:
Fixes #5590

This PR implements the announcement to the SR of the amount of options available in an Autocomplete, this reflects to the MultiSelect as well, this only happens when there is no item with focus with the exception of the Voice Over that does not announce. We also announce for any SR when the user is typing and the list changes.

### DIFF
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -3,7 +3,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {__NOT_PUBLIC_COLLECTION} from '@clayui/core';
+import {
+	__NOT_PUBLIC_COLLECTION,
+	__NOT_PUBLIC_LIVE_ANNOUNCER,
+} from '@clayui/core';
 import DropDown from '@clayui/drop-down';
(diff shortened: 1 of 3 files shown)


pedro-michel:
Hey @matuzalemsteles , can you please update the Jira Ticket?

matuzalemsteles:
@pedro-michel it has already been updated I used another ticket https://liferay.atlassian.net/browse/LPS-188262 which is linked to Clay's release epic and I also linked to the original ticket. I did this due to the LPS process that when the status is closed the QA already test again on the master and as in this case the closed is because the PR was merged in Clay's master and not in the DXP master this would be confusing for people because of that we used tickets on IFI for example.

pedro-michel:
Oh I see! Thanks for the explanation! 

So I'll update https://liferay.atlassian.net/browse/LPS-168537 to "In progress" 😄

PR #5622, opened 2023-07-05
fix(@clayui/css): Utilities c-prefers-expanded-text shouldn't display…

### LINKED ISSUE #5620: @clayui/css: c-prefers-expanded-text util breaks tab text
https://liferay.atlassian.net/browse/LPS-189635

Add the class `c-prefers-expanded-text` to the body element

Examine the tabs.

pat270:
… text in a column

fixes #5620

@ethib137 `c-prefers-expanded-text` with tabs should look like:

tabs

The page editor tabs use the utility class `flex-nowrap`. It will overflow on the right, we need to remove it if we want the tabs to break to new line with long text.

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_utilities-functional-important.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_utilities-functional-important.scss
@@ -857,6 +857,15 @@
 	visibility: hidden !important;
 }
 
+// C Prefers Expanded Text
+
+%c-prefers-expanded-text {
+	max-width: none !important;
+	overflow-wrap: break-word !important;
+	white-space: normal !important;
+	word-wrap: break-word !important;
+}
+
 @at-root {
 	// C Prefers Link Underline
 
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
In these cases instead of breaking the line maybe it would be better to keep everyone on the same lines but with scroll? this looks a lot better on a mobile UI especially when you can have multiple tabs.

marcoscv-work:
Hey gusy! I would say double scroll is not a good practice, there is not something in WCAG speaking directly about that, but Reflow speaks about two dimensions scrolling https://www.w3.org/WAI/WCAG21/Understanding/reflow.html

If we could predict this problem, probably we could change tabs by pills tabs, something like:
image

Or use lines to delimit the tabs:
image


This could work better, even if we need JS to know if we are showing the component in more than a line.

matuzalemsteles:
@marcoscv-work could we not consider this behavior similar to the behavior of a table with horizontal scrolling for smaller screens so as not to break the view?

pat270:
This has been here too long. I'm going to close for now.

Issue #5624, opened 2023-07-06
Improvements for the ClayPanel component


veroglez:
Hi guys!

From the Echo team we are replacing all our Collapses with ClayPanel component, and doing this migration several questions have arisen:

When an item is clicked a focus appears, could we implement `c-inner` so that the focus only appears when the element is focused?

<img width="250" alt="Screen Shot 2023-07-06 at 2 32 24 PM" src="https://github.com/liferay/clay/assets/9701095/5758915d-8141-443a-9254-5dce25aa312d">

Also, we have also noticed that the styles are not the same as those defined in Lexicon for this component (see the following images). Is it possible to implement them?

_Current style_
<img width="288" alt="Screen Shot 2023-07-06 at 2 33 28 PM" src="https://github.com/liferay/clay/assets/9701095/b08e0523-97aa-48e4-8e4d-e29a657afacf">

_Styles defined by Lexicon_
![image](https://github.com/liferay/clay/assets/9701095/5e84f21d-183b-451c-86d9-0d04ef2fcb40)

Thank you very much in advance!

matuzalemsteles:
> When an item is clicked a focus appears, could we implement c-inner so that the focus only appears when the element is focused?

Yeah, I think it's possible! not sure if this conflicts with the new accessibility menu rule that this will be activated by the user with focus visible. Thoughts @ethib137 @marcoscv-work 

> Also, we have also noticed that the styles are not the same as those defined in Lexicon for this component (see the following images). Is it possible to implement them?

cc @pat270

marcoscv-work:
Yeah, no problem on removing focus for mouse interaction. that's something Lexicon always wanted and we can apply as much as possible.
WCAG 2.2 will be more specific about this, but it also specify it applies for keyboard interaction / screen reader (they have their own focus):
https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum.html

ethib137:
I think the questions we need to be asking are the following:
- What should the default behavior be?
    - ie: focus-visible-only (only display focus for keyboard interaction) or focus-visible (display focus for both keyboard and mouse interactions)
- What should the accessibility menu setting be ([Other settings for Reference](https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2228781375/Accessibility+Menu+and+Shortcut+modal+to+allow+implementing+accessibility+solutions+faster+and+easier+and+with+more+awareness+for+end+users#List-of-initial-configurations-to-add))?
[...]

PR #5635, opened 2023-07-13
feat(@clayui/css): Adds :focus-visible support

pat270:
This adds `:focus-visible` support by default. All the browsers that we support have this feature, we can disable it by setting `$enable-focus-visible: false`. Most items should be covered except form `input` and `select`, these always show the focus outline.

This also adds the class `c-prefers-focus` to force focus on click via the accessibility menu. I wanted to include docs today, but couldn't get to it today.

Lastly, I didn't cover the style issues in #5624. I feel like it's too much change for this PR.

#3342 #4972 #5006

@ethib137 @matuzalemsteles @marcoscv-work @veroglez

### DIFF
--- a/clayui.com/content/docs/css/utilities/markup-accessibility.md
+++ b/clayui.com/content/docs/css/utilities/markup-accessibility.md
@@ -10,6 +10,7 @@ title: 'Accessibility'
 -   [C Prefers Reduced Motion](#css-c-prefers-reduced-motion)
 -   [C Prefers Link Underline](#css-c-prefers-link-underline)
 -   [C Prefers Letter Spacing 1](#css-c-prefers-letter-spacing-1)
+-   [C Prefers Focus](#css-c-prefers-focus)
 
 </div>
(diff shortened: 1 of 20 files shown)


matuzalemsteles (commented):
LGTM! @pat270 how is the use case when we have the focus visible together with the `c-inner`, does it cause any problems? It would be possible for example to have a toggle, for example if the `enable-focus-visible` property is enabled the `enable-c-inner` property is disabled.

matuzalemsteles:
@pat270 I sent a small fix on the breadcrumb that we were using the `c-focus-inset` on the button.

I noticed that some use cases still have the ring perhaps this is expected for the Picker button, right?. I also noticed that we still have the ring in MultiStep.

pat270:
Thanks @matuzalemsteles ! I probably missed several places. I only updated the mixins. I should check the components as well.

matuzalemsteles (approved):
LGTM!

PR #5640, opened 2023-07-18
feat(@clayui/empty-state): Output a reduced motion image by default

### LINKED ISSUE #5638: @clayui/empty-state: Append _reduced_motion to the imgSrc url by default
@matuzalemsteles I was taking a look at https://liferay.atlassian.net/browse/LPS-168263 and wanted to fix the reduced motion images, but it looks like we are using `ClayEmptyState` individually in each component. I assumed we had a tag that each module was calling. What do you think about appending `_reduced_motion` to the `imgSrc` when `imgSrcReducedMotion` isn't defined?

pat270:
fixes #5638

@matuzalemsteles I didn't update the tests yet. I'm having trouble writing it for:

```
Jest: "./packages/clay-empty-state/src/" coverage threshold for statements (100%) not met: 92.85%
Jest: "./packages/clay-empty-state/src/" coverage threshold for lines (100%) not met: 92.85%
Jest: "./packages/clay-empty-state/src/" coverage threshold for functions (100%) not met: 50%
```

I'm not sure what statements, lines, and functions mean.

### DIFF
--- a/packages/clay-empty-state/src/index.tsx
+++ b/packages/clay-empty-state/src/index.tsx
@@ -4,7 +4,7 @@
  */
 
 import classNames from 'classnames';
-import React from 'react';
+import React, {useMemo, useState} from 'react';
 
 interface IProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title'> {
 	/**
@@ -30,7 +30,7 @@ interface IProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title'> {
 	/**
 	 * Source of the image to show when `.c-prefers-reduced-motion` is active
 	 */
-	imgSrcReducedMotion?: string;
+	imgSrcReducedMotion?: string | null;
 
 	/**
 	 * Indicates empty state should be a small variant.
@@ -59,6 +59,27 @@ const ClayEmptyState = ({
 }: IProps) => {
 	const hasImg = imgSrc || imgProps;
 
+	const [error, setError] = useState(false);
+
+	const reducedMotionImage = useMemo(() => {
+		if (error) {
+			console.warn(
+				'The image url defined in `imgSrcReducedMotion` does not exist on the server. You can provide an updated url through the attribute `imgSrcReducedMotion` or set it to `{null}`.'
+			);
+
+			return null;
+		}
+
+		const dotPosition = imgSrc?.lastIndexOf('.');
+
+		if (dotPosition !== -1 && imgSrc && imgSrcReducedMotion !== null) {
+			return `${imgSrc.substr(
+				0,
+				dotPosition
+			)}_reduced_motion${imgSrc.substr(dotPosition!)}`;
+		}
+	}, [error, imgSrcReducedMotion]);
+
 	imgPropsReducedMotion = imgPropsReducedMotion
 		? imgPropsReducedMotion
 		: imgProps;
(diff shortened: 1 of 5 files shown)


packages/clay-empty-state/src/index.tsx
```diff
@@ -93,6 +102,11 @@ const ClayEmptyState = ({
 									imgPropsReducedMotion &&
 										imgPropsReducedMotion.className
 								)}
+								onError={() => {
+									console.warn(
+										`The image url defined in \`imgSrcReducedMotion\` does not exist on the server.`
+									);
```
matuzalemsteles, line 108: We can also create a state to define whether the image was loaded or had an error, we can use the state value to return in the `useMemo` I mentioned above an empty value for `imgSrcReducedMotion` so we don't render a broken image.

For example:

```jsx
const [error, setError] = useState(false);

const fallbackImageSource = useMemo(() => {
  if (imgSrcReducedMotion) {
    return imgSrcReducedMotion;
  }

  if (error) {
    return null;
  }
}, [error, imgSrcReducedMotion]);

<img
  onError={() => setError(true)}
/>
```

matuzalemsteles (commented):
@pat270 I added some comments about some things we can change.

> I'm not sure what statements, lines, and functions mean.

Oh yeah, basically the tests aren't covering all the edge cases that exist, for example if a test case created to test the behavior of the component didn't pass a condition, it called a function or line in that path that decreased the coverage percentage of the component test.

You can run `yarn test --coverage` to get the coverage of the tests, a folder is created in the root of Clay with a minimal static website that you can use to browse and track coverage and check which lines or conditions of code do not are covered by tests.

pat270:
@matuzalemsteles I've been trying to test `onError` and the warn message, but not sure how to go about it to meet the coverage threshold.

matuzalemsteles:
@pat270 no problem, we can decrease the coverage limit, you can edit the value by changing it in the `jest.config.js` file. https://github.com/liferay/clay/blob/master/jest.config.js#L88-L92

packages/clay-empty-state/src/index.tsx
```diff
@@ -59,6 +59,27 @@ const ClayEmptyState = ({
...
+			return null;
+		}
+
+		const dotPosition = imgSrc?.lastIndexOf('.');
+
+		if (dotPosition !== -1 && imgSrc && imgSrcReducedMotion !== null) {
+			return `${imgSrc.substr(
+				0,
+				dotPosition
+			)}_reduced_motion${imgSrc.substr(dotPosition!)}`;
```
matuzalemsteles, line 79: Maybe we can change a little bit here to avoid breaking the URL, for example when we pass a full uri we can change the host value instead of the pathanme, what we can do is try to make sure that we are changing the path of the image. We have some solutions for that here, for example:

- Convert the `imgSrc` to a `let url = new URL(imgSrc)` and we can get just the `url.pathname`
As we are doing an inference just getting the position of the dot can be unsafe, so we can do a regex after doing the pathname to get the position of the extname of the image, for example `url.pathname.match(/.(gif|png| jpeg|jpg)/)` and then we can get the index.

As it can happen that the developer only defines the pathname of the image when we use the `new URL` we have to ensure that the host exists in the string otherwise it will throw an invalid URL error, so we can check this by defining the base URL if it does not exist with `location.host`, for example:

```js
new URL(imgSrc, imgSrc.match(/http::\/\/|https::\/\//) ? undefined : `https://${location.host}`)
```

packages/clay-empty-state/src/index.tsx
```diff
@@ -59,6 +59,34 @@ const ClayEmptyState = ({
...
+
+		if (imgSrc && imgSrcReducedMotion !== null) {
+			const url = new URL(
+				imgSrc,
+				imgSrc?.match(/http:\/\/|https:\/\//)
+					? undefined
+					: `https://${location.host}`
+			);
+
+			const index = url.pathname.match(/.(gif|png| jpeg|jpg)/)?.index;
```
matuzalemsteles, line 81: @pat270 just to avoid use cases where the URL has a link to the image but doesn't have the file extension, we can return null instead of still trying to infer the URL.

matuzalemsteles (approved):
LGTM

PR #5646, opened 2023-07-24
fix(@clayui/css): Modal Height Full should overflow when content is t…

### LINKED ISSUE #5643: @clayui/modal Enable modal-dialog-scrollable
I'm not sure if this request is related to this old discussion https://github.com/liferay/clay/issues/4113 but after 2 years, it seems like we need this functionality again.

I'm working on a bug https://liferay.atlassian.net/browse/LPS-188817, that leads to a change in the ClayModal behavior.

We need to create a modal window (.modal-dialog) that uses all the space available in the viewport and set the content (.modal-body) to occupy the remain space between the modal-header and modal-footer, to have the buttons and the title always visible. 

image

We have something similar in ClayModal.Body API (`scrollable`) but it's limited to a fixed height. It was 120px and now is 350px so I was trying to add this new behavior in the DXP side with some CSS flex-box properties and it's doable. But I noticed we already have what we need in ClayCSS (more precisely in Bootstrap) but we cannot access it from the ClayModal API. The CSS class is called `.modal-dialog-scrollable` and it should be applied to the `.modal-dialog` and it works exactly how we need. https://getbootstrap.com/docs/5.3/components/modal/#scrolling-long-content

## What is your proposal?

- Add `scrollable` to the ClayModal component that add `.modal-dialog-scrollable` to the modal-dialog.
- Evaluate if keeping `scrollable` in the ClayModal.Body still makes sense.

- We need it.

- Duplicate code we already have from ClayCss to DXP. https://github.com/liferay-frontend/liferay-portal/pull/3537

/cc @dsanz @ugeortiz

pat270:
…oo tall

fixes #5643

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_modals.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_modals.scss
@@ -415,6 +415,10 @@
 		}
 	}
 
+	.modal-body {
+		overflow: auto;
+	}
+
 	.modal-content {
 		@include clay-css($cadmin-modal-height-full-modal-content);
 
--- a/packages/clay-css/src/scss/components/_modals.scss
+++ b/packages/clay-css/src/scss/components/_modals.scss
@@ -417,6 +417,10 @@
 		}
 	}
 
+	.modal-body {
+		overflow: auto;
+	}
+
 	.modal-content {
 		@include clay-css($modal-height-full-modal-content);
 


matuzalemsteles:
I'm not sure if it makes sense to add this to the `modal-body` if we already have the `inline-scroller` that can be configured in the `<Modal.Body scrollable />`.

pat270:
@matuzalemsteles Edu and Dani clarified what they are expecting. I think they want the modal to grow and stop when it reaches window height. I'll update this pr and take the scrollable attribute into account. It should work like the gif below.

modal-grow

matuzalemsteles:
@pat270 I'm moving forward with this, maybe `inline-scroller` isn't something valid anymore since this will be the default now.

PR #5651, opened 2023-07-26
fix(@clayui/core): fix indices when moving item in TreeView

### LINKED ISSUE #5506: Drag and drop moves folder to wrong index
https://codesandbox.io/s/fast-darkness-ofp20c?file=/src/index.js

1. Drag Blogs folder and drop it on Documents and Media Folder
2. See that it goes into Pages

Blogs goes into Documents and media

## Environment

| Tech  | Version |
| ----- | ------- |
| Clay  | v3.93.0  |
| React | v18.2.0        |

matuzalemsteles:
Fixes #5506

I still need to fix some edge cases that are still moving to the wrong places in the tree, I should continue investigating tomorrow. I'm trying to find a simpler solution that can fix only the use cases we want but it's still hard to infer due to the different possibilities that exist.


matuzalemsteles:
I'm closing this PR in favor of another more complete solution that I'll be submitting in another PR.

PR #5656, opened 2023-08-01
refactor(accessibility): Add accessibility features to Color Picker hue slider

ilzamcmed:
Jira ticket: https://liferay.atlassian.net/browse/LPS-190649

### DIFF
--- a/packages/clay-color-picker/src/Custom.tsx
+++ b/packages/clay-color-picker/src/Custom.tsx
@@ -46,6 +46,14 @@ type Props = {
 	 */
 	showPalette?: boolean;
 
+	/**
+	 * Message for aria-label
+	 */
+	messages?: {
+		close: string;
+		customColor: string;
+	};
+
 	splotch?: number;
 
 	/**
@@ -56,11 +64,17 @@ type Props = {
 
 const DEFAULT_SPLOTCH_COLOR = 'FFFFFF';
 
+const defaultMessages = {
+	close: 'close',
+	customColor: 'custom color',
+};
+
 const ClayColorPickerCustom = ({
 	color,
 	colors,
 	editorActive,
 	label,
+	messages = defaultMessages,
 	onChange,
 	onColorsChange,
 	onEditorActiveChange,
@@ -79,6 +93,11 @@ const ClayColorPickerCustom = ({
 
 					{showPalette && (
 						<button
+							aria-label={
+								editorActive
+									? messages.close
+									: messages.customColor
+							}
 							className={`${
 								editorActive ? 'close' : ''
 							} component-action`}
--- a/packages/clay-color-picker/src/Hue.tsx
+++ b/packages/clay-color-picker/src/Hue.tsx
@@ -3,8 +3,10 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import React from 'react';
+import {Keys} from '@clayui/shared';
+import React, {useCallback} from 'react';
 
+import {LimitValue} from './Editor';
 import {usePointerPosition} from './hooks';
 import {hueToX, xToHue} from './util';
 
@@ -32,11 +34,26 @@ const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
 
 	const {onPointerMove, setXY, x, y} = usePointerPosition(containerRef);
 
-	const removeListeners = () => {
+	const removeListeners = useCallback(() => {
 		selectorActive.current = false;
 
 		window.removeEventListener('pointermove', onPointerMove);
 		window.removeEventListener('pointerup', removeListeners);
+	}, []);
+
+	const onKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+		event.preventDefault();
+
+		switch (event.key) {
+			case Keys.Left:
+				onChange(value - 1);
+				break;
+			case Keys.Right:
+				onChange(value + 1);
+				break;
+			default:
+				return;
+		}
 	};
 
 	useIsomorphicLayoutEffect(() => {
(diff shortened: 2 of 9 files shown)


ilzamcmed:
Hey @pat270 @matuzalemsteles 
Can you guys review this PR?
Thanks!

packages/clay-color-picker/src/Custom.tsx
```diff
@@ -79,6 +79,7 @@ const ClayColorPickerCustom = ({
 
 					{showPalette && (
 						<button
+							aria-label={editorActive ? 'close' : 'custom color'}
```
ilzamcmed, line 82: This was something out of scope but since there wasn't any description on this buton, I added an aria-label.
Is there any language key we can use on this?

matuzalemsteles: @ilzamcmed we don't have a language key because Clay is distributed outside of DXP too so we don't have `Liferay.Language.get` for example which is just processed in SSR instead we always expose an API to configure aria-labels, in that if we are adopting a pattern of exposing a `messages` API being the object with keys that are related to language, we can do the same here.

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -54,31 +71,44 @@ const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
...
-				window.addEventListener('pointerup', removeListeners);
-			}}
-			ref={containerRef}
-		>
-			<button
-				className="clay-color-pointer clay-color-range-pointer"
-				style={{
-					background: `hsl(${value}, 100%, 50%)`,
-					left: x - 7,
+		<div className="clay-color-form-group">
```
matuzalemsteles, line 74: I think we don't need it here for this element.

matuzalemsteles (commented):
Looks good, just added some comments. The test seems to fail because the snapshot is out of date, can you update them?

matuzalemsteles:
@ilzamcmed apparently the branch is out of date or something happened that commits are coming from master. Could you see it? it also has some typescript errors.

packages/clay-color-picker/src/__tests__/index.tsx
```diff
@@ -393,6 +396,35 @@ describe('Interactions', () => {
...
+			const hueSlider = screen.getByRole('slider');
+
+			userEvent.click(hueSlider);
+			expect(handleColorsChange).toBeCalledTimes(1);
+			expect(handleValueChange).toBeCalledTimes(1);
+
+			userEvent.keyboard('[ArrowRight]');
+
+			expect(handleColorsChange).toBeCalledTimes(2);
+			expect(handleValueChange).toBeCalledTimes(2);
```
ilzamcmed, line 410: Hey @matuzalemsteles 
The reason I had to add two checks on this test was because to set focus on the slider, I had to  use a click and that triggers the `onChange` function and when i press the arrow key, it triggers again.

packages/clay-color-picker/src/__tests__/index.tsx
```diff
@@ -393,6 +396,35 @@ describe('Interactions', () => {
...
+		it('pressing left arrow key descrease hue value', async () => {
+			const hueSlider = screen.getByRole('slider');
+
+			userEvent.click(hueSlider);
+			expect(handleColorsChange).toBeCalledTimes(1);
+			expect(handleValueChange).toBeCalledTimes(1);
+
+			userEvent.keyboard('[ArrowLeft]');
+			expect(handleColorsChange).toBeCalledTimes(2);
+			expect(handleValueChange).toBeCalledTimes(2);
```
ilzamcmed, line 424: Same thing here

packages/clay-color-picker/src/__tests__/index.tsx
```diff
@@ -3,8 +3,11 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+/** @ts-ignore */
```
matuzalemsteles, line 6: Any specific reason for adding this here?

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -70,14 +93,18 @@ const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
...
 			<button
 				className="clay-color-pointer clay-color-range-pointer"
+				onBlur={() => {
+					selectorActive.current = false;
+				}}
 				style={{
 					background: `hsl(${value}, 100%, 50%)`,
 					left: x - 7,
 				}}
-				type="button"
```
matuzalemsteles, line 80: Can you test if adding this again causes the accessibility bug? because maybe if we remove this, pressing the button with the DatePicker inside the Forms will cause a submit. If this causes accessibility problems, I think we'll have to change the tag to just a `div`.

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -55,7 +72,13 @@ const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
 
 	return (
 		<div
+			aria-labelledby="color-range"
+			aria-valuemax={LimitValue.maxHue}
+			aria-valuemin={LimitValue.min}
+			aria-valuenow={value}
 			className="clay-color-range clay-color-range-hue"
+			id="color-range"
```
matuzalemsteles, line 80: is this still needed? maybe we can remove that.

matuzalemsteles (commented):
@ilzamcmed thanks IIza, I just added a few comments.

packages/clay-color-picker/src/__tests__/index.tsx
ilzamcmed: Removed.

matuzalemsteles (approved):
LGTM

PR #5657, opened 2023-08-03
feat(@clayui/navigation-bar): Adds fluidSize prop to customize container-fluid width

pat270:
https://liferay.atlassian.net/browse/LPS-192418

@matuzalemsteles I added the prop `fluidSize` to ClayNavigationBar. We currently allow `ClayLayout.ContainerFluid` to omit outputting the class `container-fluid-max-{size}` if the `size=false`. Should we add a new `fluidSize`, something like `fluid` or `fluid-layout`, in `ClayLayout.Container` to do that instead?

### DIFF
--- a/packages/clay-navigation-bar/src/index.tsx
+++ b/packages/clay-navigation-bar/src/index.tsx
@@ -32,6 +32,13 @@ export interface IProps
 		| React.ReactElement<React.ComponentProps<typeof Item>>;
 
 	/**
+	 * Set a maximum width on container-fluid.
+	 */
+	fluidSize?:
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
> Should we add a new fluidSize, something like fluid or fluid-layout, in ClayLayout.Container to do that instead?

I'm not sure, `ClayLayout.ContainerFluid` is just a wrapper on top of `ClayLayout.Container` with a simple abstraction so I don't think it would be necessary.

matuzalemsteles (approved):
LGTM

PR #5662, opened 2023-08-15
feat(@clayui/css): adds new size small to DatePicker

matuzalemsteles:
Fixes [LPS-192522](https://liferay.atlassian.net/browse/LPS-192522)

Well, we still need to adjust some small details like adding the time picker to the small size, as we haven't implemented it yet, we'll have to do it now in this ticket.

@pat270 could you revise it a little bit, not sure if it should be done that way, i thought about adding a new modified class that changes the size of the component but according to the documentation this has to be automatic and small only for screens smaller than 351px so it makes no sense to do that. I noticed that we only have breakpoints up to 576px, maybe we can add one more breakpoint of 351px? what do you think?


pat270:
@matuzalemsteles Sorry this took so long, I ended up redoing some of the media-query stuffs to reduce the repeated `@each` statements. The screenshots below show what the Date Picker looks like in max-width: 576px.

date-picker-default
date-picker-time

matuzalemsteles:
@pat270 that is great. Maybe we have to create configure the TimerPicker for a small version also from what it looks like in the DatePicker document it doesn't create scroll even in the small size with TimerPicker.

Another detail is that I thought we could set a custom breakpoint, can we add that? I think that in this specific case the DatePicker would have this small version only for the size of 351px.

pat270:
@matuzalemsteles I adjusted the top select inputs to match the design. The 24px wasn't applying. It fits the dropdown menu now. I think we can think about adding a time picker size in another pr.

matuzalemsteles:
I went ahead and updated the remaining mixins to use the new breakpoint system, I also made a few more tweaks to update the Date Picker menu size. We are going to make other tweaks to the TimePicker at a smaller size, we also need to investigate whether we should add one more breakpoint to 351px for the DatePicker use case but for now we are going to merge this as it already solves the accessibility issue.

cc @emiliano-cicero I'm planning on merging this tomorrow, can you take a look and see what you think?

PR #5663, opened 2023-08-16
 feat(clayui.com): adds Algolia's new docsearch

matuzalemsteles:
Well, I got an email from the Security team to move this to the env despite being a public key. We are also using a very old version of Algolia and need to migrate soon to using [docsearch](https://docsearch.algolia.com/docs/migrating-from-legacy) which has a new scraper infrastructure.

### DIFF
--- a/clayui.com/.env.production
+++ b/clayui.com/.env.production
@@ -1,1 +1,0 @@
-GATSBY_CLAY_NIGHTLY=false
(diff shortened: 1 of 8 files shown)


matuzalemsteles:
I went ahead and migrated to Algolia's new infrastructure and new docsearch, data is exposed via environment variable only.

ethib137:
Thanks @matuzalemsteles ... Did you also update the publicKey value so that we aren't using the one that's visible in the git history. I know you mentioned this isn't actually a vulnerability for us, but it seems to make sense to be thorough nonetheless.

matuzalemsteles:
@ethib137 yeah, I did that after i moved to the new version i need to generate a new one, so it looks ok now.

PR #5664, opened 2023-08-17
fix(@clayui/core): fix error when interacting with TreeView using SR JAWS

### LINKED ISSUE #5385: TreeView: Drag and drop via keyboard does not work with JAWS
Evan did a test using JAWS and moving the drop target using the arrow keys does not work because JAWS may be suppressing keyboard events.

### LINKED ISSUE #5628: TreeView: Actions do not open by Enter key with JAWS
Hi guys!

Testing the TreeView component with JAWS it seems that there is a bug when opening the actions using the `enter` key, the dropdown doesn't open like the rest of the dropdowns. The focus returns to the main item. Could you take a look?

Thanks in advance!

https://github.com/liferay/clay/assets/9701095/8775e95c-fefe-4ae8-b688-21cdd48bbff4

cc: @marcoscv-work

matuzalemsteles:
Fixes #5385, fixes #5628

Using [`role="application"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/application_role#background) makes more complex TreeView interactions accessible to [SRs or ATs that intercept all keyboard input and only pass through to the browser those that are accessible according to the accessibility standard](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/application_role#background).

The `role="application"` we add in the root of the TreeView only for non-Apple platforms since mac or IOS SR doesn't have this problem. This allows the keyboard navigation for drag and drop to work properly and also the interactions on actions to work accordingly.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeView.tsx
+++ b/packages/clay-core/src/tree-view/TreeView.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {useNavigation} from '@clayui/shared';
+import {isAppleDevice, useNavigation} from '@clayui/shared';
 import classNames from 'classnames';
 import React, {useRef} from 'react';
 import {DndProvider} from 'react-dnd';
@@ -120,6 +120,10 @@ interface ITreeViewProps<T extends Record<string, any>>
 
 const focusableElements = ['.treeview-link[tabindex]'];
 
+const Application = ({children}: {children: React.ReactNode}) => (
+	<div role="application">{children}</div>
+);
+
 export function TreeView<T extends Record<string, any>>(
 	props: ITreeViewProps<T>
 ): JSX.Element & {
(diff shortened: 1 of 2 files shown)


matuzalemsteles:
Hey @marcoscv-work I remember you already commented on this once but I would like your opinion on this, I was trying to find better solutions to fix this but that means we would have to refactor almost the entire component and it takes a lot of work to bring the treegrid and probably maybe not work all the behaviors we want in the treeview...

The idea of `role="application"` is normally for the entire document but it seems that it makes sense for elements or widgets that are not part of accessibility standards or want to extend mainly because of JAWS which intercepts keyboard input which makes it impossible for us to bring extra interactions, like DnD via keyboard, actions or clicks on the checkbox for multiple selection for example. One strategy I've done here is to only add this for non-Apple devices since we don't have Apple issues. What do you think?

marcoscv-work (approved):
I've test it and it's Ok, I think we can go ahead with this quick fix and see the feedback from real JAWS users.

PR #5665, opened 2023-08-18
fix(@clayui/breadcrumb): updates colors to active state with link and expand button

matuzalemsteles:
I'm updating link colors with active state and also updating expand button color. There are other small details, the item with the active state we are setting the `href` to `#` instead of the link to avoid page refresh.


pat270:
@matuzalemsteles do you have a link to the issue?

matuzalemsteles:
> We should change breadcrumb-item-expand to breadcrumb-toggle. It's more consistent with these kinds of buttons in navigation-bar, management-bar, etc.

Makes sense.

>  How do you feel about converting the collapsing breadcrumbs to be CSS only? We can hide all items except the first and active. Toggling the >> will add a class that shows the items.

I wanted to avoid this because I don't want to render the hidden items in the DOM that can cause performance issues for a large list.

pat270:
LGTM!

PR #5666, opened 2023-08-22
fix(@clayui/pagination): fix bug rendering large list of items with virtualization

### LINKED ISSUE #4605: Using pagination components with a big number of pages can cause performance degradation

matuzalemsteles:
Fixes #4605 

As part of this PR I'm adding a new experimental component that we're using in Pagination to render a DropDown with a virtualized list that works for nested groups, that part isn't done yet but I'll do it another time.

This allows rendering a large list of items without performance issues this would be the evolution of DropDown in the next major version, a simple API and less configuration with OOTB resources to avoid problems with the current DropDown component and have too much coupling. There are still some tweaks to make currently.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -26,7 +26,7 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-breadcrumb/src/': {
-			branches: 75,
+			branches: 72,
 			functions: 85,
 			lines: 95,
 			statements: 96,
@@ -165,9 +165,9 @@ module.exports = {
 		},
 		'./packages/clay-pagination/src/': {
 			branches: 86,
-			functions: 100,
-			lines: 100,
-			statements: 100,
+			functions: 92,
+			lines: 92,
+			statements: 98,
 		},
 		'./packages/clay-pagination-bar/src/': {
 			branches: 100,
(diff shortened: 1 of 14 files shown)


matuzalemsteles:
Only one bug I noticed is that when interacting via keyboard when moving from the beginning to the end of the list and then it comes back the virtual bug and it doesn't calculate the list anymore, I think this is a bug of the virtualization engine, I'll still go with it going forward and will work on investigating this, I'm also considering moving to use another more stable virtualization engine.

PR #5668, opened 2023-08-26
@clayui/color-picker: Add alpha transparency slider

ilzamcmed:
This PR is a refactor from https://github.com/liferay/clay/pull/5174
Jira ticket:  [LPS-193699](https://liferay.atlassian.net/browse/LPS-193699)

### DIFF
--- a/packages/clay-color-picker/src/Editor.tsx
+++ b/packages/clay-color-picker/src/Editor.tsx
@@ -156,18 +156,6 @@ export function Editor({
 		[b, 'b'],
 	];
 
-	const [alpha, setAlpha] = useState<number>(1);
-
-	React.useEffect(() => {
-		let currentAlpha = color.getAlpha();
-
-		if (currentAlpha !== 0 && currentAlpha !== 1) {
-			currentAlpha = +currentAlpha.toFixed(2);
-		}
-
-		setAlpha(currentAlpha);
-	}, [alpha, color]);
-
 	return (
 		<>
 			<Hue
@@ -178,7 +166,9 @@ export function Editor({
 						hue = LimitValue.maxHue;
 					}
 					onHueChange(hue);
-					onColorChange(tinycolor({h: hue, s, v}).setAlpha(alpha));
+					onColorChange(
+						tinycolor({h: hue, s, v}).setAlpha(color.getAlpha())
+					);
 				}}
 				value={hue}
 			/>
@@ -192,7 +182,7 @@ export function Editor({
 								h: hue,
 								s: saturation,
 								v: visibility,
-							}).setAlpha(alpha)
+							}).setAlpha(color.getAlpha())
 						);
 					}}
 				/>
@@ -225,7 +215,7 @@ export function Editor({
 					const newColor = color.clone();
 					onColorChange(newColor.setAlpha(value));
 				}}
-				value={alpha}
+				value={color.getAlpha()}
 			/>
 
 			<div className="clay-color-footer">
@@ -265,11 +255,9 @@ export function Editor({
 									}
 								}}
 								type="text"
-								value={
-									hex.toUpperCase().slice(-2) === 'FF'
-										? hex.toUpperCase().substring(0, 6)
-										: hex
-								}
+								value={hex
+									.toUpperCase()
+									.substring(0, color.getAlpha() < 1 ? 8 : 6)}
 							/>
 
 							<ClayInput.GroupInsetItem before tag="label">
--- a/packages/clay-color-picker/src/Editor.tsx
+++ b/packages/clay-color-picker/src/Editor.tsx
@@ -7,6 +7,7 @@ import ClayForm, {ClayInput} from '@clayui/form';
 import React, {useEffect, useReducer, useRef, useState} from 'react';
 import tinycolor from 'tinycolor2';
 
+import Alpha from './Alpha';
 import GradientSelector from './GradientSelector';
 import Hue from './Hue';
 import {findColorIndex} from './util';
@@ -43,7 +44,7 @@ export function useEditor(
 		);
 
 		return {
-			hex: color.toHex(),
+			hex: color.toHex8(),
 			hue: color.toHsv().h,
 			splotch: index !== -1 ? index : undefined,
 		};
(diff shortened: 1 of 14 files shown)


packages/clay-color-picker/src/Editor.tsx
```diff
@@ -153,8 +156,32 @@ export function Editor({
...
 		<>
+			<Hue
+				onChange={(hue) => {
+					if (hue < LimitValue.min) {
+						hue = LimitValue.min;
+					} else if (hue > LimitValue.maxHue) {
+						hue = LimitValue.maxHue;
+					}
+					onHueChange(hue);
+					onColorChange(tinycolor({h: hue, s, v}).setAlpha(alpha));
```
matuzalemsteles, line 181: I guess to avoid having an internal `alpha` state in the Editor, here you can just get the alpha from the color.

```suggestion
					onColorChange(tinycolor({h: hue, s, v}).setAlpha(color.getAlpha()));
```

packages/clay-color-picker/src/Editor.tsx
```diff
@@ -165,7 +192,7 @@ export function Editor({
 								h: hue,
 								s: saturation,
 								v: visibility,
-							})
+							}).setAlpha(alpha)
```
matuzalemsteles, line 195: Same here

```suggestion
							}).setAlpha(color.getAlpha())
```

packages/clay-color-picker/src/Editor.tsx
```diff
@@ -242,7 +265,11 @@ export function Editor({
 									}
 								}}
 								type="text"
-								value={hex.toUpperCase().substring(0, 6)}
+								value={
+									hex.toUpperCase().slice(-2) === 'FF'
```
matuzalemsteles, line 269: Why are we doing this here? Is there any specific case?

packages/clay-color-picker/src/Editor.tsx
```diff
@@ -242,7 +265,11 @@ export function Editor({
...
+										? hex.toUpperCase().substring(0, 6)
```
matuzalemsteles, line 270: Maybe we need to change this here later as it now supports hexadecimal with alpha, maybe to 8.

packages/clay-color-picker/src/Editor.tsx
```diff
@@ -192,17 +219,13 @@ export function Editor({
...
-					onHueChange(hue);
-					onColorChange(tinycolor({h: hue, s, v}));
+			<Alpha
+				color={`#${color.toHex()}`}
+				onChange={(value: number) => {
+					const newColor = color.clone();
+					onColorChange(newColor.setAlpha(value));
 				}}
-				value={hue}
+				value={alpha}
```
matuzalemsteles, line 228: Maybe we can get the alpha from the color too.

```suggestion
				value={color.getAlpha()}
```

packages/clay-color-picker/src/Editor.tsx
```diff
@@ -153,8 +156,32 @@ export function Editor({
 		[b, 'b'],
 	];
 
+	const [alpha, setAlpha] = useState<number>(1);
+
+	React.useEffect(() => {
+		let currentAlpha = color.getAlpha();
+
+		if (currentAlpha !== 0 && currentAlpha !== 1) {
+			currentAlpha = +currentAlpha.toFixed(2);
```
matuzalemsteles, line 165: Since we can remove that here, you can still do that in the `onChange` of the `<Alpha />` component you are defining in the Editor.

packages/clay-color-picker/src/index.tsx
```diff
@@ -511,17 +523,16 @@ const ClayColorPicker = ({
...
 										if (internalActive) {
 											const newColors = [...customColors];
 
-											newColors[state.splotch!] =
-												color.toHex();
-
-											onColorsChange(newColors);
+											(newColors[state.splotch!] =
+												internalToHex(color)),
+												onColorsChange(newColors);
```
matuzalemsteles, line 535: This formatting is a little strange, maybe just leaving a blank space between the calls is enough here to not create confusion in the reading.

matuzalemsteles (commented):
Good job, I just added some comments to make some state adjustments.

packages/clay-color-picker/src/Editor.tsx
ilzamcmed: Fixed

packages/clay-color-picker/src/Editor.tsx
ilzamcmed: Its checking if hex color has no transparency (represented by the two last 'FF' digits). If we have, there is no need to display it. 

Since we want to add the 8 digit, is better to add the tow last digits?
Add `hex.toUpperCase()`

packages/clay-color-picker/src/Editor.tsx
ilzamcmed: Removed this `useEffect` and moved this logic to `<Alpha />` component

packages/clay-color-picker/src/index.tsx
ilzamcmed: I removed the brackets and added a blank line. Thanks @matuzalemsteles ! :smiley:

packages/clay-color-picker/src/Editor.tsx
matuzalemsteles: Hmm I think this might be hardcoded check for `FF` the alpha in a hex could be other values too maybe something like this?

```jsx
hex.toUpperCase().substring(0, color.getAlpha() < 1 ? 6 : 8)
```

ilzamcmed: Thanks @matuzalemsteles !

matuzalemsteles (approved):
Just a small detail to move on, good work!

packages/clay-color-picker/src/Hue.tsx
```diff
@@ -92,7 +92,6 @@ const ClayColorPickerHue = ({value = 0, onChange = () => {}}: Props) => {
 				window.addEventListener('pointerup', removeListeners);
 			}}
 			ref={containerRef}
-			role="slider"
```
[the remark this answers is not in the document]
ilzamcmed: You're right! Thanks @matuzalemsteles

PR #5670, opened 2023-09-07
feat(@clayui/css): Adds focus-within

pat270:
https://liferay.atlassian.net/browse/LPS-195104

This adds `:focus-within` to our `form-control` elements. This offloads focus interactions to the browser on more complex elements like `input-group`s (e.g., search bar, time picker, rgb inputs in color picker).  We weren't managing the focus class on a lot of input groups and now we no longer need to.

~One thing I don't like is the name of the `form-control-container` class. I was hoping for something better, but can't think of anything. @matuzalemsteles @ethib137 do you think we should add an attribute to the `ClayInput.GroupItem` component to output this? I'm currently adding it via `className="form-control-container"`.~

Edit: Went with `input-group-item-focusable`. Thanks @ethib137 !

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -226,8 +226,8 @@ module.exports = {
 		'./packages/clay-time-picker/src/': {
 			branches: 85,
 			functions: 95,
-			lines: 98,
-			statements: 98,
+			lines: 95,
+			statements: 95,
 		},
 		'./packages/clay-toolbar/src/': {
 			branches: 50,
--- a/packages/clay-color-picker/src/Editor.tsx
+++ b/packages/clay-color-picker/src/Editor.tsx
@@ -85,7 +85,7 @@ const RGBInput = ({name, onChange, value}: RGBInputProps) => {
 	return (
 		<ClayForm.Group>
 			<ClayInput.Group>
-				<ClayInput.GroupItem>
+				<ClayInput.GroupItem className="form-control-container">
 					<ClayInput
 						data-testid={`${name}Input`}
 						insetBefore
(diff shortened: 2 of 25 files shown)


ethib137:
Hey @pat270 can you make sure to get a design review of these changes to how focus is displayed before this is merged.

marcoscv-work:
Should be these the url to test this PR?
https://deploy-preview-5670--next-clayui.netlify.app/docs/components/dual-list-box.html
https://deploy-preview-5670--next-storybook-clayui.netlify.app/?path=/story/design-system-components-duallistbox--default

pat270:
@marcoscv-work yes, that url contains the changes here.

matuzalemsteles:
Well, if we want this automatically, we can add this class directly to `ClayInput.GroupItem. But maybe we don't want this for all cases, for example in the Search case, does it make sense to show the focus on the input when the focus is on the search button? This seems confusing to the user.

pat270:
@matuzalemsteles The focus shouldn't be on the input when the search button is focused. I need to update the selector to account for that. I remember thinking about excluding `input[type="submit"]`, but couldn't think of a use case.

pat270:
@matuzalemsteles where are you seeing the search button focus issue?

matuzalemsteles:
@pat270 No, I ended up misunderstanding.

pat270:
There is an issue, but it's only with `<input type="submit" value="Submit"/>`. These aren't generally used in input groups since it's hard to swap out an icon for text. We can use `<button type="submit">` for that.

pat270:
@matuzalemsteles I'm still thinking about the class name, can you ping me before merging it. I can't find how to convert to draft for some reason.

matuzalemsteles:
@pat270 Of course, I think that after the PR is ready, it is not possible to go back to draft again, just draft > ready.

PR #5672, opened 2023-09-14
fix(@clayui/autocomplete): fix error when navigating via keyboard with the mouse over the item

matuzalemsteles:
Fixes [LPS-194107](https://liferay.atlassian.net/browse/LPS-194107)

Well, I thought fixing this would be more complex but I ended up using some resources we have to prevent hovering over the item from updating the virtual focus when the scroll moves when navigating via keyboard. The strategy here was simpler than I thought at the beginning, so we just avoided updating the `activeDescendant` on hover over the item when the current interaction is being via keyboard, we got this information with the `isFocusVisible` hook that we have to identify this behavior.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,7 +14,7 @@ module.exports = {
 			statements: 100,
 		},
 		'./packages/clay-autocomplete/src/': {
-			branches: 68,
+			branches: 67,
 			functions: 82,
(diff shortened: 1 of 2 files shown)


ethib137:
That's a nice surprise, when it's easier than expected. LGTM. 😄

PR #5677, opened 2023-09-20
fix(@clayui/css): Update and add box-shadow variable to cadmin class

HenriquePereiraa:
[LPS Ticket](https://liferay.atlassian.net/browse/LPS-190644)

### DIFF
--- a/packages/clay-css/src/scss/cadmin/variables/_range.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_range.scss
@@ -120,15 +120,9 @@ $cadmin-clay-range-input: map-deep-merge(
 			z-index: $cadmin-zindex-clay-range-input-form-control,
 			hover: (
 				cursor: $cadmin-link-cursor,
-				clay-range-thumb: (
-					background-color: $cadmin-primary-l3,
-				),
 			),
 			focus: (
 				outline: 0,
-				clay-range-thumb: (
-					box-shadow: $cadmin-component-focus-inset-box-shadow,
-				),
 			),
 			disabled: (
 				color: $cadmin-clay-range-disabled-color,
@@ -166,6 +160,16 @@ $cadmin-clay-range-input: map-deep-merge(
 			after: (
 				content: unquote("'\\FEFF' attr(data-label-max)"),
 				right: 0,
+			),
+		),
+		focus: (
+			clay-range-thumb: (
+				box-shadow: $cadmin-component-focus-inset-box-shadow,
+			),
+		),
+		hover: (
+			clay-range-thumb: (
+				background-color: $cadmin-primary-l3,
 			),
 		),
 	),


packages/clay-css/src/scss/cadmin/variables/_globals.scss
```diff
@@ -329,8 +329,9 @@ $cadmin-component-transition: color 0.15s ease-in-out,
 
 $cadmin-component-focus-box-shadow: #{0 0 0 2px $cadmin-white,
 	0 0 0 4px $cadmin-primary-l0} !default;
-$cadmin-component-focus-inset-box-shadow: #{inset 0 0 0 2px $cadmin-primary-l0,
-	inset 0 0 0 4px $cadmin-white} !default;
+$cadmin-component-focus-inset-box-shadow: #{inset 0 0 0 0.125rem
```
pat270, line 332: We shouldn't change to `rem` value here. How it works is if you change the `font-size` on the `html` element, the value will scale based on it. In cadmin, we are trying to make it hard to overwrite admin styles via site styles. This is why we use `px` values.

packages/clay-css/src/scss/cadmin/variables/_range.scss
```diff
@@ -80,6 +80,11 @@ $cadmin-clay-range-input: map-deep-merge(
 			position: absolute,
 			top: 50%,
 			width: 50%,
+			focus: (
```
pat270, line 83: You are right about something wonky going on with `clay-range-variant` or `clay-range-input-variant`. Something might be off with the way we are merging here https://github.com/liferay/clay/blob/dfe8e4d9789e2e2586bd439db2496c9f64388d39/packages/clay-css/src/scss/mixins/_forms.scss#L2287-L2312.

https://github.com/liferay/clay/blob/dfe8e4d9789e2e2586bd439db2496c9f64388d39/packages/clay-css/src/scss/variables/_range.scss#L128-L144 works but https://github.com/liferay/clay/blob/dfe8e4d9789e2e2586bd439db2496c9f64388d39/packages/clay-css/src/scss/atlas/variables/_range.scss#L29-L32 doesn't.

It's weird and convoluted because we are keeping backward compatibility with Clay 2.x.

pat270 (changes requested):

packages/clay-css/src/scss/cadmin/variables/_range.scss
```diff
@@ -168,6 +162,16 @@ $cadmin-clay-range-input: map-deep-merge(
 				right: 0,
 			),
 		),
+		focus: (
```
pat270, line 165: @HenriquePereiraa This is working for me on cadmin. You probably need to restart the storybook / server to get the CSS to recompile. 

Something that is curious to me here is that we seem to be circumventing a merge problem in the mixin by declaring it this way.

pat270 (changes requested):

pat270:
Closing this one in favor of https://github.com/liferay/clay/pull/5682

PR #5678, opened 2023-09-20
feat(@clayui/css): Tables adds table-nested-rows

pat270:
https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2432237790/Nested+Rows+-+Front-end+DataSet

I'm sending this as draft so we can discuss this implementation. 

table-nested

I had to use nested tables because we need a container element to animate the collapse if needed. The example in https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/#ex_label is clean and concise, but we can't collapse the rows like we do panels. The code for the nested table looks like:

```html
<tr role="presentation">
  <td colspan="6" role="presentation">
    <div class="collapse" id="tableCollapse01">
      <table role="presentation">
        <tbody role="presentation">
          <tr aria-level="2" role="row"></tr>
        </tbody>
      </table>
    </div>
  </td>
</tr>
```

Voice Over announces this like it's the next table row. The `role="presentation"` makes the screen reader ignore them. 

To create the proper spacing, I used an extra element as a spacer. The [indentation levels](https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2432237790/Nested+Rows+-+Front-end+DataSet#Nesting-Indentation) seem quite complicated since it's based on elements contained inside the title column. We also can't foresee the combination of items that will be put in it. I included a code snippet below.

```html
<tr aria-level="2" role="row">
    <td role="gridcell" style="width: 324.38px">
        <button class="component-action component-drag"></button>
        <div class="autofit-row">
            <div class="autofit-col autofit-col-toggle"></div>
            <div class="autofit-col autofit-col-toggle">
                <button class="component-action" data-target="#tableCollapse02" type="button"></button>
            </div>
            <div class="autofit-col autofit-col-checkbox">
                <div class="custom-control custom-checkbox"></div>
            </div>
                <span class="table-title">Second Level Object</span>
            </div>
        </div>
    </td>
    <td role="gridcell" style="width: 119.74px">2</td>
    <td role="gridcell" style="width: 104.48px">3</td>
    <td role="gridcell" style="width: 233.09px">4</td>
    <td role="gridcell" style="width: 101.58px">5</td>
    <td role="gridcell" style="width: 78.73px">6</td>
</tr>
```

This is for "Second Level Object" in the screenshot. We need to output an extra `<div class="autofit-col autofit-col-toggle"></div>` to space it correctly. In "Third Level Object", we need to output:

```html
<div class="autofit-col autofit-col-toggle"></div>
<div class="autofit-col autofit-col-toggle"></div>
<div class="autofit-col autofit-col-checkbox"></div>
```
to account for the two toggles and the checkbox in each parent item. I really don't like outputting empty elements, but this seemed easier and more future proof than managing specific padding for every situation. I guess the other option might be to calculate it with JS and inline it. Let me know your opinions, Thanks!

/cc @ethib137 @matuzalemsteles @ilzamcmed

### DIFF
--- a/packages/clay-css/src/scss/mixins/_tables.scss
+++ b/packages/clay-css/src/scss/mixins/_tables.scss
@@ -366,6 +366,30 @@
 				}
 			}
 
+			.autofit-col-checkbox {
+				@include clay-css(map-get($map, autofit-col-checkbox));
+			}
+
+			.autofit-col-icon {
+				@include clay-css(map-get($map, autofit-col-icon));
+			}
+
+			.autofit-col-toggle {
+				@include clay-css(map-get($map, autofit-col-toggle));
+			}
+
+			.component-action {
+				@include clay-link(map-get($map, component-action));
+			}
+
+			.component-drag {
+				@include clay-link(map-get($map, component-drag));
+			}
+
+			.component-toggle {
+				@include clay-link(map-get($map, component-toggle));
+			}
+
 			.custom-control,
 			.form-check {
 				@include clay-custom-control-variant(
--- a/clayui.com/static/js/docs-site.js
+++ b/clayui.com/static/js/docs-site.js
@@ -25,6 +25,7 @@ if (!Element.prototype.closest) {
 		var t = event.target;
 
 		var a = t.tagName === 'a' || t.tagName === 'button' ? t : t.closest('a') || t.closest('button');
+		var column = t.closest('.autofit-col-toggle') || false;
 
 		if (a) {
 			if (a.getAttribute('href') === '#1') {
@@ -36,6 +37,14 @@ if (!Element.prototype.closest) {
 			if (dataToggle && dataToggle.startsWith('c-prefers')) {
 				document.querySelector('body').classList.toggle(a.getAttribute('data-toggle'));
 			}
+		}
+
+		if (column) {
+			var button = column.querySelector('.component-action');
+
+			button.classList.toggle('show');
+
+			document.querySelector(button.dataset.target).classList.toggle('show');
 		}
 	});
 })();
--- a/packages/clay-css/src/scss/cadmin/components/_tables.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_tables.scss
@@ -418,6 +418,20 @@ caption {
 	}
 }
 
+// .table-nested-rows
+
+.table-nested-rows {
+	@include clay-table-variant($cadmin-c-table-nested-rows);
+
+	.component-action.show .collapse-icon-closed {
+		display: none;
+	}
+
+	.component-action:not(.show) .collapse-icon-open {
+		display: none;
+	}
+}
+
 // Show Quick Action
 
 .show-quick-actions-on-hover {
--- a/packages/clay-css/src/scss/cadmin/variables/_tables.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_tables.scss
@@ -403,6 +403,14 @@ $cadmin-c-table: map-deep-merge(
 				padding-right: 0,
 			),
 		),
+		component-drag: (
+			font-size: 14px,
+			height: 16px,
+			width: 16px,
+			focus: (
+				box-shadow: $cadmin-component-focus-inset-box-shadow,
+			),
+		),
 		custom-control: (
 			margin-bottom: 0,
 		),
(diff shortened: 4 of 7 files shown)


matuzalemsteles:
@pat270 I wanted to avoid nested indentation here instead we can add indentation only to the first column using padding. This is important because we will follow the accessibility standard of [`treegrid`](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/) where is a list instead of a nested tree.

This will also allow us to create list virtualization to avoid performance problems. Another detail is that the design expects the columns to be aligned if we follow this we will not be able to achieve this for the levels below.

ethib137:
Yeah, I think we should stay with the tree-grid markup. That's more important than the transition. Hopefully we will still be able to find a way to add a transition later, but it doesn't have to be our first priority.

pat270:
@matuzalemsteles can you clarify

> Another detail is that the design expects the columns to be aligned if we follow this we will not be able to achieve this for the levels below.

Is adding spacer elements is no good?

matuzalemsteles:
@pat270 Sorry for not being so clear, I'm referring to the example in the documentation when we use the nested strategy.

Screenshot 2023-09-21 at 15 23 02

For example, level 2 has the columns misaligned with the other columns at level 2.

pat270:
@matuzalemsteles Ah I see...yeah. I was looking at the current FDS and it adds a width to every cell when sized. I was assuming that would continue to be the case, but it might have some performance issues for extremely large tables.

I think I like having no transitions when collapsing / expanding. The code is much cleaner.

packages/clay-css/src/scss/mixins/_tables.scss
```diff
@@ -366,6 +366,37 @@
 				}
 			}
 
+			.autofit-col-checkbox {
+				@include clay-css(map-get($map, autofit-col-checkbox));
+			}
+
+			.autofit-col-icon {
+				@include clay-css(map-get($map, autofit-col-icon));
+			}
```
matuzalemsteles, line 375: I don't really like that we add classes like this to deal with spacing between elements, I wish we used more of the spacing classes we have instead this improves reusability and in the future we can have more use of components like `<Stack/ >` which would handle whitespace between elements following the standards.

matuzalemsteles (commented):
I think we're close to merging this, I just added a comment.

packages/clay-css/src/scss/mixins/_tables.scss
pat270: @matuzalemsteles I was debating with myself whether it should be utilities like 1-10 or this above. I chose easier for the theme developer over component developer. My reasoning is checkbox and icon are the same width (currently 16px), but in the future if we want one of them to be larger (24px), it's really hard to target the column with utilities. It will have to be something like 
```css
.table-nested-rows .autofit-col-1 + .autofit-col-1 {}
````

pat270:
Closing this one in favor of https://github.com/liferay/clay/pull/5682

packages/clay-css/src/scss/mixins/_tables.scss
matuzalemsteles: Yes that makes sense too. But we have to understand that the utilities follow a spacing pattern that would be defined by the Lexicon, its white space pattern. For example, we will not see a pattern outside 1.5px but rather 4px, 2px... that is, if a theme developer wants to change the spacing of his application to apply different styles or have more white space instead of doing this to Each component has its own space palette which in theory will reflect this for all components, what we have to think about is that this has to be consistent so that the UI does not create inconsistency of different spaces for each component and this applies to any UI is a theme developer as well as improving the UI for the end user and less maintenance. Does this make sense to you?

pat270: @matuzalemsteles Ok how about we use spacing utils with specific column names? We don't need to use them, but can tie them in with spacing utils if the need arises.

matuzalemsteles: Hmm what would that be like? I'm not sure I understand correctly.

pat270: It would be something like `class="autofit-col-1 checkbox-column"`.

matuzalemsteles: I see. I'm not sure, we can go with that for now then and think more about it. What do you think? so we can discuss it more.

pat270: Ok thanks, I'll send an update today.

pat270:
@matuzalemsteles I changed it to utility classes. Unfortunately the sizing is a bit weird due to using padding for spacing, like 24px with padding turns out to be 28px minimum. We could use margins also, but I keep going back and forth between those two.

matuzalemsteles (approved):
LGTM

PR #5679, opened 2023-09-21
feat(@clayui/core): adds the implementation of the new data-oriented composition for Table

matuzalemsteles:
Closes [LPS-192998](https://liferay.atlassian.net/browse/LPS-192998)

This PR implements new components for Table to make room for new OOTB features for developers, so we can continue improving the UX and making the component more useful and prevent developers from having to implement more features on top of that.

The idea here is to bring a data-oriented composition using the Collection pattern that we already use in other Clay components, this already brings the UX benefits that we have in other components and we can also implement the features such as list virtualization that we need to implement the [treegrid](https://liferay.atlassian.net/browse/LPS-193002) pattern and the [column resizer](https://liferay.atlassian.net/browse/LPS-193001).

Instead of implementing this on top of the Table currently, we are recreating the component parts in a new composition to avoid backward compatibility or mechanisms that have to identify when using the old or new composition to use the right component. So to use the new composition use the Table from `@clayui/core` and we leave the component in `@clayui/table` in deprecation mode.

This is the new UX and composition of the component, it is not yet public. I will make it public only after we implement all the new features, such as treegrid support and column resizer with OOTB accessibility.

```jsx
<Table>
  <Head items={columns}>
    {(column) => <Column key={column.id}>{column.name}</Column>}
  </Head>

  <Body items={rows}>
    {(row) => (
      <Row items={columns}>
        {(column) => <Column>{row[column.id]}</Column>}
      </Row>
    )}
  </Body>
</Table>
```

Just like in other components that use collection, the list can be static or dynamic, for example `<Row />` can be a static list when the column is not dynamic.

```jsx
<Table>
  <Head items={columns}>
    <Column key="name">Name</Column>
    <Column key="type">Type</Column>
  </Head>

  <Body items={rows}>
    {(row) => (
      <Row>
        <Column>{row.name}</Column>
        <Column>{row.type}</Column>
      </Row>
    )}
  </Body>
</Table>
```

### DIFF
--- a/packages/clay-core/src/table/Column.tsx
+++ b/packages/clay-core/src/table/Column.tsx
@@ -1,0 +1,93 @@
+/**
+ * SPDX-FileCopyrightText: © 2023 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import classNames from 'classnames';
+import React from 'react';
+
+import {Scope, useScope} from './ScopeContext';
+
+type Props = {
+	/**
+	 * Aligns the text inside the Cell.
+	 */
+	align?: 'center' | 'left' | 'right';
+
+	/**
+	 * Children content to render content.
+	 */
+	children: React.ReactNode;
+
+	/**
+	 * Sometimes we are unable to remove specific table columns from the DOM
+	 * and need to hide it using CSS. This property can be added to the "new"
+	 * first or last cell to maintain table styles on the left and right side.
+	 */
+	delimiter?: 'start' | 'end';
+
+	/**
+	 * Fills out the remaining space inside a Cell.
+	 */
+	expanded?: boolean;
+
+	/**
+	 * Aligns horizontally contents inside the Cell.
+	 */
+	textAlign?: 'center' | 'end' | 'start';
+
+	/**
+	 * Truncates the text inside a Cell.
+	 */
+	truncate?: boolean;
+
+	/*
+	 * Break the text into lines when necessary.
+	 */
+	wrap?: boolean;
+} & React.ThHTMLAttributes<HTMLTableCellElement> &
+	React.TdHTMLAttributes<HTMLTableCellElement>;
+
+export const Column = React.forwardRef<HTMLTableCellElement, Props>(
+	function ColumnInner(
+		{
+			align,
+			children,
+			className,
+			delimiter,
+			expanded,
+			textAlign,
+			truncate,
+			wrap = true,
+			...otherProps
+		},
+		ref
+	) {
+		const scope = useScope();
+		const As = scope === Scope.Head ? 'th' : 'td';
+
+		return (
+			<As
+				{...otherProps}
+				className={classNames(className, {
+					'table-cell-expand': truncate || expanded,
+					[`table-cell-${delimiter}`]: delimiter,
+					[`table-column-text-${textAlign}`]: textAlign,
+					[`text-${align}`]: align,
+					'table-cell-ws-nowrap': !wrap,
+				})}
+				ref={ref}
+			>
+				{truncate ? (
+					<span className="text-truncate-inline">
+						<span className="text-truncate">{children}</span>
+					</span>
+				) : (
+					children
+				)}
+			</As>
+		);
+	}
+);
+
+Column.displayName = 'Item';
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -33,6 +33,7 @@
 		"@clayui/loading-indicator": "^3.60.0",
 		"@clayui/modal": "^3.104.0",
 		"@clayui/provider": "^3.93.0",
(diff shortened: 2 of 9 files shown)


matuzalemsteles:
cc @ethib137 let me know what you think.

packages/clay-core/src/table/Column.tsx
```diff
@@ -0,0 +1,93 @@
...
+	truncate?: boolean;
+
+	/*
+	 * Break the text into lines when necessary.
+	 */
+	wrap?: boolean;
+} & React.ThHTMLAttributes<HTMLTableCellElement> &
+	React.TdHTMLAttributes<HTMLTableCellElement>;
+
+export const Column = React.forwardRef<HTMLTableCellElement, Props>(
```
ethib137, line 51: Is there a reason why we use 'Column' instead of 'Cell' as the name?

ethib137 (approved):
This is looking great @matuzalemsteles ! Is it fully working with collections now?

packages/clay-core/src/table/Column.tsx
matuzalemsteles: In fact, there isn't much of a reason, just using a slightly more "popular/common" nomenclature.

matuzalemsteles:
> Is it fully working with collections now?

Yes, but we are not yet using all the features it provides. For example, I haven't enabled virtualization yet because I will do that when I'm working on the treegrid implementation. Another feature we probably won't use is infinite scroll since table is typically used with pagination, so I'll enable when we have use cases for that.

matuzalemsteles:
I'm going to merge this to continue working on the other tasks. But if we need to make any changes we can send it in another PR anyway.

packages/clay-core/src/table/Column.tsx
ethib137: Hey @matuzalemsteles can you point me towards where you've seen `column` used? If we were to think about this like google sheets, this component renders a single cell such as `A1`, right? In the context of google sheets a column would refer to `A:A` or all the cells in a column. 

Checking a few other libraries, [MaterialUI](https://mui.com/material-ui/react-table/#api) uses `TableCell`, [NextUI](https://nextui.org/docs/components/table#import) uses `TableColumn` only for cells in the header and `TableCell` for cells in the body. 

Additionally, FDS uses `TableCell` and `TableHeadCell`. Also, most of the classnames being applied to this component are prefixed `table-cell-*`.

So what do you think, does `Cell` make more sense, or do you still see more reasons for `Column`?

matuzalemsteles: @ethib137 These are good points, I hadn't thought about this side of Google Sheets. I really think that the case of Google sheets makes sense mainly because you deal with cells specifically although you call a column like `A` which groups several cells I would say that this is different for Table where sheets have the concept of cells much stronger because the cell can have a relationship between different cells, not necessarily related only to the specific column but can, for example, be diagonal and column is only used as a reference, which is different in a table where the cell has to be associated with a column to have context, which is not true in Google Sheets since the nature of the cell can have its own context and it is a more flexible term we can say like this.

But yes, calling cells for body elements makes sense, I don't like it particularly because the cell seems to not necessarily be "associated" with a column directly... For example Column is also a common name for components not necessarily related to a table , for example a group of elements to be rendered horizontally or vertically. For example:

- [Braid Design System](https://seek-oss.github.io/braid-design-system/components/Columns)
- [Base Web](https://baseweb.design/components/table-semantic/) They don't necessarily use Column as a composition, there are some components as part of their Table that use the term column but their composition is very different.
- [Adobe Spectrum](https://react-spectrum.adobe.com/react-spectrum/TableView.html) - They use the same Column pattern for the header and Cell for the body.
- [Polaris](https://polaris.shopify.com/components/tables/index-table) - Their composition is a little confusing but I would say that they use the same naming pattern of Column for header and cell for cell, in other table components the cell is not mentioned as part of the APIs, only rows.
- [Evergreen](https://evergreen.segment.com/components/table) - Here they already use Cell for anywhere but you can understand that the cell is not necessarily associated with a column so it seems that it can be rendered in different ways which makes sense here.
- [Carbon](https://carbondesignsystem.com/components/data-table/code/) - Composition follows the same pattern of having Column for header but they only call it Header, and Cell just for the body.
- [Radix](https://www.radix-ui.com/themes/docs/components/table) - Composition is based on HeaderCell and Cell for the body only.

I don't really like the idea of creating a Column component to be used only in the Head and a Cell for only the body, I want to reduce the number of components for the composition and make the component more intelligent, which is also one of the reasons because I went with Column instead of just Cell for naming.

Anyway, we can go with Cell because we are already using this term in several places, so it will be simpler for current developers to associate this.

ethib137: Sounds good. Thanks @matuzalemsteles . Can you create a ticket for making this change?

PR #5680, opened 2023-09-22
feat(@clayui/core): adds new API to handle column sorting with accessibility support

matuzalemsteles:
Closes [LPS-192999](https://liferay.atlassian.net/browse/LPS-192999) and [LPS-193000](https://liferay.atlassian.net/browse/LPS-193000)

This PR implements column sorting support for an OOTB table with accessibility support, this also adds the controlled and uncontrolled state pattern to the sort API. There are still some details to be done, such as adding sort support to the Data Provider for asynchronous sorting. DX also makes this very simple, see my example of sorting in the storybook.

```jsx
<Table onSortChange={onSortChange} sort={sort}>
  <Head>
    <Column key="name" sorting>Name</Column>
    <Column key="files" sorting>Files</Column>
    <Column key="type">Type</Column>
  </Head>

  <Body items={items}>
    {(row) => (
      <Row>
        <Column>{row.name}</Column>
        <Column>{row.files}</Column>
        <Column>{row.type}</Column>
      </Row>
    )}
  </Body>
</Table>
```
By default the component does not handle the sorting of table items, this has to be done by the developer, I'm thinking if we can add this OOTB when the `sort` state is uncontrolled so this could be useful for demo cases or for all data when live on the client side.

The component also deals with accessibility, in the header column we add the ARIA attributes of `aria-sort` and `aria-describedby` to inform that the column can be sorted and current state. We also announce the sort action to the SR, with the column name and direction. One detail is that I want to remove the link action and move it to th so that the focus is on the column instead of the header but for that we would have to deal with managing the focus of the columns which we will do this in the treegrid implementation, so in this At the moment I am making the link assertive for the SR with the description, `aria-sort` is only used in th in the column header so it would have no effect on the link.

An initial idea to add sorting support to the DataProvider, it is very difficult to make some OOTB resource for this when we are dealing with data because there is no standard for this, so it will still be necessary for the developer to describe his rules for fetch the data in the server.

```jsx
const {resource, sort, sortChange} = useResource({
  fetchPolicy: FetchPolicy.CacheFirst,
  link: 'https://example.com/api/',

  // sort.column = column key
  // sort.direction = 'ascending' | 'descending'
  async sort(items, sort) {
    // impl here
  }
});

<Table onSortChange={sortChange} sort={sort}>
  {...}
</Table>
```

### DIFF
--- a/packages/clay-core/src/table/Column.tsx
+++ b/packages/clay-core/src/table/Column.tsx
@@ -40,9 +40,9 @@ type Props = {
 	keyValue?: React.Key;
 
 	/**
-	 * Whether the column allows sorting. Only available in the header column.
+	 * Whether the column allows sortable. Only available in the header column.
 	 */
-	sorting?: boolean;
+	sortable?: boolean;
 
 	/**
 	 * Aligns horizontally contents inside the Cell.
@@ -75,7 +75,7 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
 			delimiter,
 			expanded,
 			keyValue,
-			sorting,
+			sortable,
 			textAlign,
 			textValue,
 			truncate,
@@ -86,18 +86,17 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
 	) {
 		const {onSortChange, sort, sortDescriptionId} = useTable();
 		const scope = useScope();
-		const As = scope === Scope.Head ? 'th' : 'td';
+		const isHead = scope === Scope.Head;
+		const As = isHead ? 'th' : 'td';
 
 		return (
 			<As
 				{...otherProps}
 				aria-describedby={
-					scope === Scope.Head && sorting
-						? sortDescriptionId
-						: undefined
+					isHead && sortable ? sortDescriptionId : undefined
 				}
 				aria-sort={
-					scope === Scope.Head && sorting
+					isHead && sortable
 						? sort && keyValue === sort.column
 							? sort.direction
 							: 'none'
@@ -109,16 +108,14 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
 					[`table-column-text-${textAlign}`]: textAlign,
 					[`text-${align}`]: align,
 					'table-cell-ws-nowrap': !wrap,
-					'table-head-title': scope === Scope.Head,
+					'table-head-title': isHead,
 				})}
 				ref={ref}
 			>
-				{scope === Scope.Head && sorting ? (
+				{isHead && sortable ? (
 					<a
 						aria-describedby={
-							scope === Scope.Head && sorting
-								? sortDescriptionId
-								: undefined
+							isHead && sortable ? sortDescriptionId : undefined
 						}
 						className="inline-item text-truncate-inline"
 						href="#"
--- a/packages/clay-core/src/table/Column.tsx
+++ b/packages/clay-core/src/table/Column.tsx
@@ -3,10 +3,12 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import Icon from '@clayui/icon';
 import classNames from 'classnames';
 import React from 'react';
 
 import {Scope, useScope} from './ScopeContext';
+import {useTable} from './context';
 
 type Props = {
 	/**
@@ -32,6 +34,17 @@ type Props = {
 	expanded?: boolean;
 
 	/**
+	 * Internal property.
+	 * @ignore
+	 */
+	keyValue?: React.Key;
+
+	/**
+	 * Whether the column allows sorting. Only available in the header column.
+	 */
+	sorting?: boolean;
+
+	/**
 	 * Aligns horizontally contents inside the Cell.
 	 */
 	textAlign?: 'center' | 'end' | 'start';
@@ -45,6 +58,11 @@ type Props = {
 	 * Break the text into lines when necessary.
 	 */
 	wrap?: boolean;
+
+	/**
+	 * Sets a text value if the component's content is not plain text.
+	 */
+	textValue?: string;
 } & React.ThHTMLAttributes<HTMLTableCellElement> &
 	React.TdHTMLAttributes<HTMLTableCellElement>;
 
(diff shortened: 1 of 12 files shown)


matuzalemsteles:
@ethib137 let me know what you think.

packages/clay-core/src/table/Column.tsx
```diff
@@ -56,29 +74,86 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
...
 					[`table-column-text-${textAlign}`]: textAlign,
 					[`text-${align}`]: align,
 					'table-cell-ws-nowrap': !wrap,
+					'table-head-title': scope === Scope.Head,
 				})}
 				ref={ref}
 			>
-				{truncate ? (
+				{scope === Scope.Head && sorting ? (
+					<a
```
ethib137, line 117: Why is this an anchor instead of a button?

packages/clay-core/src/table/Column.tsx
```diff
@@ -56,29 +74,86 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
...
+											: 'ascending',
+								},
+								textValue!
+							);
+						}}
+						role="button"
+					>
+						<span className="text-truncate">{children}</span>
+
+						{sort && keyValue === sort.column && (
```
ethib137, line 144: This looks like it matches the current Lexicon table designs, but in the future we will likely need to improve this so that there is a visual indicator that shows which columns allow for sorting and which do not. So no change needed for now, but thought it worth mentioning.

ethib137 (changes requested):
Very nice work @matuzalemsteles ! I definitely agree that we don't need to handle the sorting of items in the component. In most real world situations this would be handled by an api call to the server. Thanks.

packages/clay-core/src/table/Column.tsx
matuzalemsteles: It seems that `button` doesn't have the correct styling inside the header, I need to add this to the CSS but I haven't gone ahead with it yet because I will remove the link in the next interactions and keep the action in `th` instead.

packages/clay-core/src/table/Column.tsx
matuzalemsteles: Yeah, this actually seems a bit confusing, I think the only indication that Lexicon adds is when you hover over the header column, change the color and add the underline but there really isn't anything when the component is viewed without interacting with he.

packages/clay-core/src/table/Column.tsx
```diff
@@ -31,6 +33,17 @@ type Props = {
...
+	/**
+	 * Internal property.
+	 * @ignore
+	 */
+	keyValue?: React.Key;
+
+	/**
+	 * Whether the column allows sorting. Only available in the header column.
+	 */
+	sorting?: boolean;
```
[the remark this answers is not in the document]
matuzalemsteles: Oh nice!! haha a great explanation from ChatGPT 😄

matuzalemsteles:
Well, I've been thinking all day about how we can implement OOTB sorting with useResource support for asynchronous data but still no ideas on how this can be done, mainly because there seems to be no API standard on how this is done on the backend like would it be like for pagination.

My idea was to have a `sort` API as support when the `useResource` is connected to the Table and the user clicks on a column to sort it, it would call the `sort` API which is asynchronous the developer can implement a request to the server with the return of new items or just implement it on the client side. The point is that the data lives in the useResource and we only have a single source of truth so that the developer does not have to store it in a different state, for example.

I'm planning to do this later to find out more about how this will be used by the DXP teams. @ethib137 what do you think?

ethib137:
Hey @matuzalemsteles , I don't think we need this component to handle sorting at all. What you've already added, providing a callback `onSortChange` is enough. In general DXP will always handle sorting on the backend because we are always dealing with an indeterminate amount of data. We never want to assume that all the data would exist on the client side, so we have to sort on the backend and only return the current page of sorted data. 

So, don't spend any more time working on a way to have the table itself handle sorting. This will be handled outside of this component. Does that sound okay?

matuzalemsteles:
@ethib137 Yeah, that's really what I think. What I'm referring to here is how to improve the implementation of asynchronous sorting.

For example, when you load table data, if you want to use the `useResource` hook, it will bring you a lot of benefits such as integration with suspense, caching with LRU, solving network problems, retry and other details. So when you need to sort this table if you use the `onSortChange` callback you will have two ways to implement this, do your own fetching implementation and separate the data from `useResource` so you may have the problem of having two different states which is not good... so my idea here is to try to simplify this implementation. This is a pseudocode of how it would be used.

```jsx
const {resource, sort, sortChange} = useResource({
  fetchPolicy: FetchPolicy.CacheFirst,
  link: 'https://example.com/api/',

  // sort.column = column key
  // sort.direction = 'ascending' | 'descending'
  async sort(sort, fetch) {
    // impl here
  }
});

<Table onSortChange={sortChange} sort={sort}>
  {...}
</Table>
```

So when the user clicks to sort the column, the `sort` API of the `useResource` hook is called and upon returning the data we update the internal states of the hook, the question is as this would be something extremely flexible I would have to create a API that can take advantage of hook features, such as retry, caching, etc... I was thinking about the `sort` API of passing the internal fetch callback that in the end calls the `fetcher` API and which deals with retry, caching and etc... So this still allows flexibility as we don't have an API sorting standard but still uses the features of the data provider.

matuzalemsteles:
I went ahead and added a simpler implementation for the asynchronous sortable. Instead of adding an API I am passing `sort` into the `fetch` API call when defined this still allows the flexibility to implement the fetch logic and also have SSOT.

```jsx
const {resource, sort, sortChange} = useResource({
  fetch: (link, init, sort) => {
    const url = new URL(link);

    if (sort) {
      url.searchParams.append('column', String(sort.column));
      url.searchParams.append('direction', sort.direction);
    }

    return fetch(url, init);
  },
  link: 'https://rickandmortyapi.com/api/character/',
});
```

matuzalemsteles:
I'm going to go ahead and merge this.

PR #5684, opened 2023-09-27
fix(@clayui/css): Checkbox and radio text should be vertically aligne…

pat270:
…d when breaking to a new line

These are regular checkbox and radio not in a dropdown-menu

normal-checkbox


matuzalemsteles (approved):
LGTM

PR #5685, opened 2023-10-05
feat(@clayui/core): adds base implementation for rendering a table tree into an accessible list

matuzalemsteles:
Closes [LPS-197645](https://liferay.atlassian.net/browse/LPS-197645)

It's still a work in progress but I've already implemented a good part of the [treegrid accessibility](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/) pattern and also tree rendering in a list. I'm still not sure if this will be the default because we don't know the number of items that can be rendered so I haven't done a stress test to check the performance, the point is that we will always convert a tree into a list every time the items change at rendering time, this is a blocker and can be noticeable to the user if it contains millions or thousands of items in the tree.

> ⚠️ In this implementation I will not work with multiple selection because this will be worked on the MT side instead of the table, as in DXP the MT and the listing components such as list, card and table do not communicate with each other using React and being independent (taglibs), we still don't know how it will work. In theory will work as expected but there will be no interactions like selecting the parent will select all the children, so we have to investigate this first in the development of MT for FDS and then come back here if you need something or some support so we can implement it like we did it in TreeView.
>
> The row DnD implementation I'm not sure we want to implement now, we can work on that later.

I am satisfied that the composition does not change for the developer when he needs a table tree, practically the composition is done in the same way and being data agnostic. The component converts the tree to a list and uses the [treegrid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/) when the `nestedKey` API is defined.

```jsx
<Table aria-label="Table" nestedKey="children">
  <Head items={columns}>
    {(column) => <Cell key={column.id}>{column.name}</Cell>}
  </Head>

  <Body
    items={[
      {
        children: [{id: 10, name: 'WoW', type: 'MMORPG'}],
        id: 1,
        name: 'Games',
        type: 'File folder'
      },
      {
        id: 2,
        name: 'Program Files',
        type: 'File folder',
      },
    ]}
  >
    {(row) => (
      <Row items={columns}>
        {(column) => (
          <Cell key={`${row.id}:${column.id}`}>
            {row[column.id]}
          </Cell>
        )}
      </Row>
    )}
  </Body>
</Table>
```

### ToDo

- [ ] ~Tests~
- [x] Adds the API to expand a node (it will be similar to TreeView)
- [x] Docs


packages/clay-core/src/table/Cell.tsx
```diff
@@ -110,31 +126,40 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
...
-												: 'ascending'
-											: 'ascending',
-								},
-								textValue!
-							);
-						}}
-						role="button"
+						onClick={(event) => event.preventDefault()}
+						role="presentation"
+						tabIndex={-1}
```
matuzalemsteles, line 162: I made some changes here, the Link at this moment is only serving as the hover interaction when hovering the mouse over it to cover this Lexicon behavior but the sortable action is done when clicking on the column header to better deal with accessibility and SR assertiveness .

The idea is to remove this link and add the appropriate style without needing a link

packages/clay-core/src/table/Cell.tsx
```diff
@@ -154,6 +179,14 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
...
 						<span className="text-truncate">{children}</span>
 					</span>
+				) : treegrid && index === 0 ? (
+					<div
+						style={{
+							paddingLeft: (level - 1) * 28,
+						}}
+					>
+						{children}
+					</div>
```
matuzalemsteles, line 189: I will also be adding autofit to allow you to add Checkbox and other elements you want in the cell.

```jsx
<Row items={columns}>
	{(column) => (
		<Cell key={`${row.id}:${column.id}`}>
			<Checkbox />
			<Icon symbol="folder" />
			{row[column.id]}
		</Cell>
	)}
</Row>
```

packages/clay-core/src/table/useTreeNavigation.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+					item =
+						items[
+							event.key === Keys.Up ? position - 1 : position + 1
+						];
+					break;
+				}
+				case Keys.Left: {
+					if (isRow) {
+						// TODO
```
matuzalemsteles, line 81: Will be implemented in the next iteration.

packages/clay-core/src/table/Body.tsx
```diff
@@ -20,14 +21,66 @@ type Props<T> = {
...
+
+		if (Array.isArray(array[i]![nestedKey])) {
+			delete item[nestedKey];
+			yield item;
+			yield* flatten(array[i]![nestedKey], nestedKey, level + 1);
+		} else {
+			yield item;
+		}
+	}
+}
```
matuzalemsteles, line 51: Well I'm still not sure about using a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) here but this is very good for traversing a tree because we can pause and return later and this is great for when we need to phase the iteration when we have a lot of data, now we iterate through the entire tree to put together the list but we can take better advantage of this when we have more use cases.

- Phase the iteration to avoid blocking the main thread, this can work using a `requestIdleCallback` or using the [browser's scheduler](https://developer.mozilla.org/en-US/docs/Web/API/Scheduler) to yield CPU time to other tasks so we can meet an appropriate deadline following the [RAIL Model](https://web.dev/rail/)
- We can ignore non-visible items and recompute when they are visible (We can make a stress test to see the advantage of this)

packages/clay-core/src/table/Cell.tsx
ethib137: Was the plan for this to be inline style, or is there some other utility class from @pat270 for this?

packages/clay-core/src/table/Body.tsx
ethib137: Yeah, I have the same concern. I wasn't aware of how generators worked so I had to read up on that. It's not that complicated after reading up on it though, so I think it's probably okay. We should be aware of it though, and consider changing it out for something simpler if we don't end up needing it's pause capability.

packages/clay-core/src/table/Cell.tsx
```diff
@@ -110,31 +126,40 @@ export const Column = React.forwardRef<HTMLTableCellElement, Props>(
...
+					typeof keyValue === 'number'
+						? `number,${keyValue}`
+						: `string,${keyValue}`
+				}
+				onClick={(event) => {
+					if (!(isHead && sortable)) {
+						return;
+					}
+
+					event.preventDefault();
```
ethib137, line 139: In portal I understand Brian's rule for when to add a blank line between items: If you can't sort the items add a line between them. But, what's our rule in Clay for when to add a blank line?

packages/clay-core/src/table/Row.tsx
```diff
@@ -29,24 +31,78 @@ type Props<T> = {
...
+	 * @ignore
+	 */
+	keyValue?: React.Key;
+
+	/**
+	 * Internal prop.
+	 * posinset
+	 * @ignore
+	 */
+	_index?: number;
```
ethib137, line 46: Why are some of these internal props prefixed with an underscore but `keyValue` isn't?

packages/clay-core/src/table/context.tsx
```diff
@@ -14,10 +14,18 @@ type Context = {
 	onSortChange: (sorting: Sorting | null, textValue: string) => void;
 	sort: Sorting | null;
 	sortDescriptionId: string;
+	nestedKey?: string;
```
ethib137, line 17: Can we sort this?

ethib137 (approved):
This is looking really good @matuzalemsteles . Thanks for the work on this!

packages/clay-core/src/table/Cell.tsx
matuzalemsteles: Under the hood I'm using the `autofit` classes, the difference is that we don't make this visible to the dev so that he doesn't worry about implementing it.

packages/clay-core/src/table/Cell.tsx
matuzalemsteles: @ethib137 we don't adopt the spacing rule so much here, we just use something for visuals that is clear to understand when reading, we prioritize that.

packages/clay-core/src/table/Row.tsx
matuzalemsteles: `keyValue` and `index` are properties coming from collection, while the properties `_index`, `_size`... are properties coming from flatten when we reassemble the items to a list, so they are just to differentiate where these values are coming from.

packages/clay-core/src/table/Body.tsx
```diff
@@ -20,14 +21,83 @@ type Props<T> = {
...
+
+		if (Array.isArray(array[i]![nestedKey])) {
+			delete item[nestedKey];
+			// @ts-ignore
+			item._expandable = true;
+			yield item;
+
+			// TODO: Make it dynamic using the key generated by the developer at
+			// render time.
+			if (expandedKeys.has(array[i]!['id'])) {
```
matuzalemsteles, line 55: I spent some time trying to make this data agnostic for `<Row />` to get the key through the collection so that the developer can configure it regardless of their data structure but I had problems with this because the flatten it must be done before passing the data to the collection for rendering and this was a problem and this way I cannot obtain this information because this happens before and at the time of rendering.

When we move this to be part of the collection API perhaps we can obtain this dynamically, meaning that the layout creation is still in a tree so it would be done before and the flattening would happen at render time. TreeView can benefit from this as well.

matuzalemsteles:
Now in the final stage, I will update the example to contain more data, add documentation and tests.

There is still a small bug in the focus control of the last focused element, for some reason if the last focused element is a cell then it resets the focus and moves to the nearest row, I think this is related to the nested focus system, I will investigate this further tomorrow.

matuzalemsteles:
This is ready to go now but we still need to add CSS styles to handle the focus on the row and cell but maybe we may have a problem with this due to the use of native tags, @pat270 do you think we can do this with this markup?

matuzalemsteles:
I'm going to go ahead and we can submit a new PR to handle the CSS focus for the row and cell in another PR and I'm also going to add the tests using another ticket to split the tasks and have it merged now.

PR #5689, opened 2023-10-09
LPS-197657 Remove xlink from icons

fortunatomaldonado:
https://liferay.atlassian.net/browse/LPS-197657

Xlinkhref does not work correctly when a CDN is used. This leads to an error message in console when using clay/icons.

I found that xlink is Deprecated and no longer recommended: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

Removing it with href should work just fine unless I am missing something.
Please let me know if there are any questions or comments about this.
Thank you!


matuzalemsteles:
This is good so we can get rid of it, I remember we already had a long discussion about this but I don't remember exactly why. @pat270 or @bryceosterhaus do you remember anything about this in the past?

pat270:
@matuzalemsteles I believe it was used only to support older browsers. I think it's pretty safe to do this since we have updated what browsers we support.

pat270 (approved):

john-co:
ci:test:clay

john-co:
I don't anticipate any test regressions based on the code change

bryceosterhaus:
> I remember we already had a long discussion about this but I don't remember exactly why

I had the same recollection. I don't remember why we kept it, other than browser support when it was required.

matuzalemsteles:
Okay, looks good then. I really think this was related to IE support.I go forward with this I'll just update the snapshot tests.

Issue #4763, opened 2022-03-24
Explore changing the icons viewbox from 512px to 16px 


emiliano-cicero:
## What is your proposal?
We are using a viewBox of 512px (0 0 512 512) for our SVG icons but our most common use case is Portal and other digital products from Liferay where we use them at a size of 16px. 
My proposal is to update the viewBox to match the common usage of the icons which is 16px (0 0 16 16).
As they're SVGs we can increase the size without problems so there shouldn't be a problem to remove the 512px.

## Why would adopting this proposal be beneficial?
The main benefit is that we will be able to create 16px icons and implement them directly without having to resize them twice (one to fit the 512 viewBox and a second time to fit the 16px in the products).

## What are the alternatives to this proposal?
The alternative would be to keep it as it is. 

I'm requesting this issue as an Exploration just to see what would be the problems of doing this.

pat270:
I don't see any problem with updating the icon's `viewBox`. The icons will still render the same. We will need to export all the icons again though. We can't change `viewBox="0 0 16 16" on old icons and expect it to work.

PR #5691, opened 2023-10-16
fix(@clayui/css): SVG Icons bookmarks and bookmarks-full should be centered

pat270:
https://liferay.atlassian.net/browse/LPS-199263

@matuzalemsteles I grabbed the icons from the Lexicon Icon library on Figma. They use `viewBox="0 0 16 16"`. I remember we had issues when the viewBox wasn't `viewBox="0 0 512 512"`. I don't remember if it was because of Clay or a tag in DXP. I was hoping you remember.

/cc @drakonux @emiliano-cicero


matuzalemsteles:
I'm not sure if this is specifically a problem with the viewport, it could perhaps be a problem with the generated svg that may be off-center within the 512x512 viewport because we have other icons similar to this.

pat270:
@matuzalemsteles In the `<clay:icon>` tag we are hardcoding the `viewBox`.

https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib-clay/src/main/java/com/liferay/frontend/taglib/clay/servlet/taglib/IconTag.java#L31

Should we swap out all the SVG's to use the `0 0 16 16` from the Lexicon Icon library?

https://github.com/liferay/clay/issues/4763

matuzalemsteles:
@pat270 I'm not sure about this and I don't think this is the problem yet, my assumption is that the misalignment is the problem with the icon not being positioned correctly within the 512x512 viewport, I would check this first before proceeding with 16x16.

pedro-michel:
I believe this is the same problem that we had in last days with X & Canada Flag icon, right @pat270 ?

PR #5693, opened 2023-10-17
fix(@clayui/tooltip) - Check if title is equal to an empty string and…

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPS-196978


matuzalemsteles (approved):
LGTM

PR #5696, opened 2023-10-20
fix(@clayui/css): Checkbox/Radio text should be properly aligned for real this time

pat270:

@matuzalemsteles I had to move `{children}` inside `custom-control-label-text`. I couldn't make it work in all situations because of that without trying to remove the padding for every case.

### DIFF
--- a/packages/clay-card/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-card/src/__tests__/__snapshots__/index.tsx.snap
@@ -172,27 +172,6 @@ exports[`ClayCard renders a ClayCard as a selectable file card 1`] = `
             <span
               class="custom-control-label"
             />
-            <div
-              class="aspect-ratio-item aspect-ratio-item-center-middle aspect-ratio-item-fluid card-type-asset-icon"
-            >
-              <svg
-                class="lexicon-icon lexicon-icon-documents-and-media"
-                role="presentation"
-              >
-                <use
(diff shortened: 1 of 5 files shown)


packages/clay-card/src/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -172,27 +172,6 @@ exports[`ClayCard renders a ClayCard as a selectable file card 1`] = `
...
-            </div>
-            <span
-              class="sticker sticker-danger sticker-bottom-left"
-            >
-              <span
-                class="sticker-overlay"
-              >
-                DOC
-              </span>
-            </span>
```
matuzalemsteles, line 195: How strange this shouldn't be deleted... @pat270 can you check this?

packages/clay-form/src/Checkbox.tsx
```diff
@@ -95,11 +95,11 @@ const ClayCheckbox = React.forwardRef<HTMLInputElement, IProps>(
 						{label && (
 							<span className="custom-control-label-text">
 								{label}
+
+								{children}
```
matuzalemsteles, line 99: I think this will break other use cases like Card, it doesn't need the label but still wants to render a children, we can maybe keep the `children` in place of before if the label is not defined.

matuzalemsteles (commented):
@pat270 It seems that now this broke the Card, I think it's due to the `children` changing position.

PR #5700, opened 2023-10-23
feat(@clayui/form): Checkbox and Radio should use modifier class custom-control-outside if a label is present

pat270:

@matuzalemsteles I added the modifier class `custom-control-outside`. It's output when there is a `label`. I wish there was a better way to do this globally.

Edit: I'll send a docs update in a separate pr.


matuzalemsteles (approved):
LGTM! Thanks for working on this

PR #5703, opened 2023-10-25
chore: remove the use of regenerator in the build

matuzalemsteles:
Well, we are deleting `transform-regenerator` so that we can use the [Generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) feature and consequently we can now also use [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) instead of continuing to use `then` for Promise, these are features that have been supported by the browser for a long time and is also included within our compatibility matrix.

- `async/await` - 95.9% https://caniuse.com/async-functions
- `function*` - 96.25% https://caniuse.com/es6-generators
- `yield*` - 95.15% https://caniuse.com/mdn-javascript_operators_yield_star
- `yield` - 95.15% https://caniuse.com/mdn-javascript_operators_yield

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -26,8 +26,9 @@ module.exports = (api) => {
 		],
 		presets: [
 			[
-				'@babel/env',
+				'@babel/preset-env',
 				{
+					exclude: ['transform-regenerator'],
 					targets: [
 						'last 2 Chrome versions',
 						'last 1 Edge versions',


matuzalemsteles:
cc @bryceosterhaus @ethib137

bryceosterhaus:
This looks like it should fix out problems in DXP. I tested a local build too and verified that it no-longer requires a global of `regeneratorRuntime`

bryceosterhaus:
Feel free to merge and publish a new version of clay and then I can test in DXP tomorrow.

ethib137:
Would this fix the issue I've started seeing in codesandbox as well: https://codesandbox.io/s/language-selector-picker-example-iq811m?file=/src/index.js:0-66

Screenshot 2023-10-26 at 3 21 58 PM

matuzalemsteles:
@ethib137 It's not possible to test on codesandbox yet because we haven't published the packages with this change yet, so we're just checking the build result locally.

PR #5710, opened 2023-11-02
fix(@clayui/css): Adds .table-focus for focus outline around table row

pat270:
This is a follow up for https://liferay.atlassian.net/browse/LPS-197645. We add `.table-focus` to the table row to add a focus outline when needed. I had to use an unconventional box shadow since Safari built their browser to spec for table rows. The spec says table row position is undefined which prevents us from using relative positioning to position our focus outline.

table-focus

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_tables.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_tables.scss
@@ -31,22 +31,10 @@ $table-head-link: map-deep-merge(
 // Table Body
 
 $c-table-tbody: () !default;
-$c-table-tbody: map-merge(
-	(
-		background-color: $white,
-	),
-	$c-table-tbody
-);
 
 // Table Footer
 
 $c-table-tfoot: () !default;
-$c-table-tfoot: map-merge(
-	(
-		background-color: $table-bg,
-	),
-	$c-table-tfoot
-);
 
 $table-caption-color: $gray-900 !default;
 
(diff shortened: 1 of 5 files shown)


packages/clay-css/src/scss/atlas/variables/_tables.scss
```diff
@@ -134,6 +122,41 @@ $table-link: map-deep-merge(
...
 
+// .table-focus
+
+$c-table-focus: () !default;
+$c-table-focus: map-deep-merge(
+	(
+		td: (
+			before: (
+				box-shadow: (
+					inset 0 0.125rem 0 0 $primary-l0,
```
ethib137, line 133: Hey @pat270 this looks annoying to handle, but good job. One question though... do we not have a global variable for focus color that we could use instead of directly using $primary-l0?

pat270: @ethib137 We have two variables for the focus outline. 

```scss
$component-focus-box-shadow: #{0 0 0 0.125rem $white, 0 0 0 0.25rem $primary-l0} !default;
$component-focus-inset-box-shadow: #{inset 0 0 0 0.125rem $primary-l0,
	inset 0 0 0 0.25rem $white} !default;
```

There are two colors, inner ($white) and outer ($primary-l0). For me, I have to reference these two variables not for the color but for the length settings. I think a global variable for focus color would be useful if we didn't want it to change when someone modifies global colors like `$primary`, `$secondary`, etc.

ethib137: Ahh that makes sense... thanks for explaining the details. Looks good as is.

matuzalemsteles:
Hey @pat270 I think we need the focus to be on the cell too because the user can focus on the cell or the row. For example, when moving the focus to the cell, the visual focus should not be on the row but on the cell that is focused.

pat270:
@matuzalemsteles lol great...I think it shouldn't be a problem. I'll update this.

pat270:
@matuzalemsteles I updated this, we can put `table-focus` on the `tr` or `td`.

matuzalemsteles (approved):
LGTM

PR #5714, opened 2023-11-08
chore(@clayui/core): add focus to table

matuzalemsteles:
Follow up #5710

### DIFF
--- a/packages/clay-css/src/scss/variables/_tables.scss
+++ b/packages/clay-css/src/scss/variables/_tables.scss
@@ -390,8 +390,9 @@ $c-table-hover: map-deep-merge(
 $c-tr-table-focus: () !default;
 $c-tr-table-focus: map-deep-merge(
 	(
+		outline: 0,
 		td: (
-			outline: 'none',
+			outline: 0,
 			before: (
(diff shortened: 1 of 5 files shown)


packages/clay-css/src/scss/variables/_tables.scss
```diff
@@ -391,6 +391,7 @@ $c-tr-table-focus: () !default;
 $c-tr-table-focus: map-deep-merge(
 	(
 		td: (
+			outline: 'none',
```
pat270, line 394: @matuzalemsteles This is a nitpick, but we have been using `outline: 0` to remove the outline.

packages/clay-css/src/scss/variables/_tables.scss
```diff
@@ -391,6 +391,7 @@ $c-tr-table-focus: () !default;
 $c-tr-table-focus: map-deep-merge(
 	(
```
pat270, line 392: I think we need to add `outline: 0` here. I'm seeing a double border for `tr` focus.
double-border

pat270 (changes requested):

PR #5716, opened 2023-11-09
LPS-201300 Fixing Accessibility Issue at Clay Panel Component

AkramYoussoufi:
[Ticket](https://liferay.atlassian.net/browse/LPS-201300)

[Thread-Slack](https://liferay.slack.com/archives/C03CAD9EF/p1699525029248659)

### DIFF
--- a/packages/clay-panel/src/index.tsx
+++ b/packages/clay-panel/src/index.tsx
@@ -10,6 +10,7 @@ import {
 	InternalDispatch,
 	setElementFullHeight,
 	useControlledState,
+	useId,
 } from '@clayui/shared';
 import classNames from 'classnames';
 import React from 'react';
@@ -107,13 +108,14 @@ function ClayPanel({
 
 	const {prefersReducedMotion} = useProvider();
 
+	const ariaControlsId = useId();
+
 	return (
 		<div
 			{...otherProps}
 			className={classNames('panel', className, {
 				[`panel-${displayType}`]: displayType,
 			})}
-			role="tablist"
 		>
 			{!collapsable && (
 				<>
@@ -135,6 +137,7 @@ function ClayPanel({
 			{collapsable && (
 				<>
 					<ClayButton
+						aria-controls={ariaControlsId}
 						aria-expanded={internalExpanded}
 						className={classNames(
 							'panel-header panel-header-link',
@@ -148,7 +151,6 @@ function ClayPanel({
 						)}
 						displayType="unstyled"
 						onClick={() => setInternalExpanded(!internalExpanded)}
-						role="tab"
 					>
 						{displayTitle &&
 							(React.isValidElement(displayTitle) ? (
@@ -178,6 +180,7 @@ function ClayPanel({
 					</ClayButton>
 
 					<CSSTransition
+						aria-labelledby={ariaControlsId}
 						className={classNames(
 							'panel-collapse',
 							collapseClassNames,
@@ -190,6 +193,7 @@ function ClayPanel({
 							exit: `show`,
 							exitActive: 'collapsing',
 						}}
+						id={ariaControlsId}
 						in={internalExpanded}
 						onEnter={(element: HTMLElement) =>
 							setElementFullHeight(element)
@@ -201,7 +205,7 @@ function ClayPanel({
 						onExiting={(element) => {
 							element.style.height = '';
 						}}
-						role="tabpanel"
+						role="region"
 						timeout={!prefersReducedMotion ? 250 : 0}
 					>
 						<div>{children}</div>
--- a/packages/clay-panel/docs/markup-panel.md
+++ b/packages/clay-panel/docs/markup-panel.md
@@ -21,7 +21,7 @@ mainTabURL: 'docs/components/panel.html'
 ## Basic Usage(#css-basic-usage)
 
 <div class="sheet-example">
-	<div class="panel" role="tablist">
+	<div class="panel" >
 		<div class="panel-header">
 			<span class="panel-title">Title</span>
 		</div>
@@ -32,7 +32,7 @@ mainTabURL: 'docs/components/panel.html'
 </div>
 
 ```html
-<div class="panel" role="tablist">
+<div class="panel">
 	<div class="panel-header">
 		<span class="panel-title">Title</span>
 	</div>
@@ -45,7 +45,7 @@ mainTabURL: 'docs/components/panel.html'
 ### Secondary(#css-panel-secondary)
 
 <div class="sheet-example">
-	<div class="panel panel-secondary" role="tablist">
+	<div class="panel panel-secondary" >
 		<div class="panel-header">
 			<span class="panel-title">Title</span>
 		</div>
@@ -55,7 +55,7 @@ mainTabURL: 'docs/components/panel.html'
 </div>
 
 ```html
-<div class="panel panel-secondary" role="tablist">
+<div class="panel panel-secondary">
 	<div class="panel-header">
 		<span class="panel-title">Title</span>
 	</div>
@@ -67,7 +67,7 @@ mainTabURL: 'docs/components/panel.html'
 ### Unstyled(#css-panel-unstyled)
 
 <div class="sheet-example">
-	<div class="panel panel-unstyled" role="tablist">
+	<div class="panel panel-unstyled" >
 		<div class="panel-header">
 			<span class="panel-title">Title</span>
 		</div>
@@ -77,7 +77,7 @@ mainTabURL: 'docs/components/panel.html'
 </div>
 
 ```html
-<div class="panel panel-unstyled" role="tablist">
+<div class="panel panel-unstyled">
 	<div class="panel-header">
 		<span class="panel-title">Title</span>
 	</div>
@@ -95,14 +95,14 @@ Collapsable panels provide you with the ability to expand and collapse content a
 </div>
 
 <div class="sheet-example">
-	<div class="panel" role="tablist">
+	<div class="panel" >
 		<button
 			aria-controls="collapsePanel"
 			aria-expanded="false"
 			class="btn btn-unstyled panel-header panel-header-link collapse-icon collapse-icon-middle collapsed"
 			data-target="#collapsablePanel" 
 			data-toggle="collapse"
-			role="tab"
+			
 		>
 			<span class="panel-title">Toggle me for expanding!</span>
 			<span class="collapse-icon-closed">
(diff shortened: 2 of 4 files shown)


ethib137:
Hi @AkramYoussoufi , did you try using this component after your changes with your keyboard and a screen reader? You'll notice not much info is provided when interacting with a panel. We need these roles to make it work properly with accessible technologies.

AkramYoussoufi:
I have uploaded a video in the comment section in the [Ticket](https://liferay.atlassian.net/browse/LPS-201300) to check.

ethib137:
Hey @AkramYoussoufi , this change still won't work. It breaks the proper HTML semantics. Mozilla explains [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role):
> Elements with the role tab must either be a child of an element with the tablist role, or have their id as part of the [aria-owns](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns) property of a tablist.

Since you have removed the `tablist` role entirely, this breaks those requirements.

AkramYoussoufi:
Hello @ethib137, 
I wanted to let you know that the tablist role is implemented in the Group Component. This is actually why I removed it from the `index.tsx` parent element because both elements were identified as the `tablist` during the devTools scan it triggers that the element parent needs a `tab` and `tabpanel` element which is a false alarm.

but I have made a switch to leave the tab list to the clay panel upon this request, I stumbled upon another problem that I didn't recognize at first glance, the `tablist` won't be a good choice for this component for several reasons!

1. the `tabpanel` role is holded by the Clay Body component and the `tablist` with the ClayComponent this cause accessibility issue when scanning with the devTools because `tablist` should not have a `tabpanel` as a child based on the [mozilla doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role) and the structure based on the mentioned [documentation](https://clayui.com/docs/components/panel.html) of clay.
```
 <ClayPanel
    collapsable
    displayTitle={
      <ClayPanel.Title>
        <h3>{"Title"}</h3>
        <span>{"If field "}</span>
        <ClayLabel displayType="success">{"Country"}</ClayLabel>
        <ClayLabel>{"Is Equal To"}</ClayLabel>
        <span>{"value "}</span>
        <ClayLabel displayType="info">{"Brazil"}</ClayLabel>
        <span>{"enable "}</span>
        <ClayLabel displayType="success">{"State"}</ClayLabel>
      </ClayPanel.Title>
    }
    displayType="secondary"
    showCollapseIcon={true}
    spritemap={spritemap}
  >
    <ClayPanel.Body>{"Body!"}</ClayPanel.Body>
  </ClayPanel>
```
2. even moving the `tabpanel` outside the claypanel will cause the rendering of the panel incorrect, this will not solve either the accessibility issue or the component functionality.

AkramYoussoufi:
@ethib137 [New structure](https://i.imgur.com/JFSecew.png)

AkramYoussoufi:
@ethib137 I didn't like the changes I've made so I decided to work on a new changes which are more efficient!

AkramYoussoufi:
@ethib137 in this changes, I have fixed the conflict happens between **ClayGroupPanel** and **ClayPanelComponent**

as I mentioned earlier the GroupPanel Component has its own role="tablist"
```

const ClayPanelGroup = ({
	children,
	className,
	fluid,
	fluidFirst,
	fluidLast,
	flush,
	small,
	...otherProps
}: IProps) => {
	return (
		<div
			{...otherProps}
			aria-orientation="vertical"
			className={classNames('panel-group', className, {
				'panel-group-fluid': fluid,
				'panel-group-fluid-first': fluidFirst,
				'panel-group-fluid-last': fluidLast,
				'panel-group-flush': flush,
				'panel-group-sm': small,
			})}
			role="tablist"
		>
			{children}
		</div>
	);
};
```

and on the ClayPanelComponent also has the `tablist` role 

this cause issue when using **ClayPanelGroup** and **ClayPanel** Component together ex:

```
	<ClayPanel.Group>
		{['One', 'Two', 'Three'].map((item) => (
			<ClayPanel
				collapsable
				displayTitle={item}
				key={item}
				showCollapseIcon
			>
				<ClayPanel.Body>
					{`Here is some content inside for number ${item}`}
				</ClayPanel.Body>
			</ClayPanel>
		))}
	</ClayPanel.Group>
```


in this scenario, the ClayPanelGroup will look for an element with `tab` and `tabpanel` and will not find them because the ClayPanel assigns to it the `tab` and `tabpanel` as its own element causing the `ClayPanel.Group` triggering the issue of missing elements when scanning with devTools.

I have added a checker which check if the GroupPanel is used as parent if yes in this case the ClayPanel will not use `role=tabpanel` and leave it to the  **ClayPanelGroup** to handle the assignment.

AkramYoussoufi:
one problem present now is that **ClayPanel** Component uses the ClayPanel.Body as its own child this caused problems since the ClayPanel.Body has and must have the `tabpanel` role. 

based on [mozilla doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role) and devTools scanner 
image
 the `tabpanel` should not be a child to the tablist but rather it needs to be a sibling! 

and the **ClayPanel** Component is built to receive the **ClayPanel.Body** as it's own child! based on the [code source ](https://github.com/liferay/clay/blob/master/packages/clay-panel/src/index.tsx) and this is the normal implementation :
```
			<ClayPanel
				collapsable
				displayTitle={item}
				key={item}
				showCollapseIcon
			>
				<ClayPanel.Body>
					{`Here is some content inside for number ${item}`}
				</ClayPanel.Body>
			</ClayPanel>
```
this causes an issue in the scanner but for me, the accessibility is working fine like this
I believe we can either ignore the message, or make another role well-suited for this component.

AkramYoussoufi:
@ethib137 [Disclosure Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-faq/) Implemented

AkramYoussoufi:
Hello @ethib137, [disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/) is implemented

packages/clay-panel/src/index.tsx
```diff
@@ -146,9 +146,9 @@ function ClayPanel({
 								show: internalExpanded,
 							}
 						)}
+						data-disclosure
```
ethib137, line 149: What is the purpose of adding this data attribute?

packages/clay-panel/src/index.tsx
```diff
@@ -135,6 +134,7 @@ function ClayPanel({
 			{collapsable && (
 				<>
 					<ClayButton
+						aria-controls="panel-body"
```
ethib137, line 137: The value of `aria-controls` needs to be the id of the element it controls.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

AkramYoussoufi: fixed!

packages/clay-panel/src/index.tsx
AkramYoussoufi: I removed it

AkramYoussoufi:
@ethib137 required pattern is implemented I added a logic that will look for the number of panels within the parent element and give each panel element its needed index within a group of panels.

AkramYoussoufi:
@ethib137 I don't understand why the CI return failed on the test because the test in my local passed successfully:

```
$ npm run test

> @clayui/panel@3.106.1 test
> jest --config ../../jest.config.js --updateSnapshot

ts-jest[versions] (WARN) Version 27.5.1 of jest installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=28.0.0 <29.0.0-0). Please do not report issues in ts-jest if you are using unsupported versions.
 PASS  src/__tests__/index.tsx (6.488 s)
  ClayPanel
    ✓ renders (31 ms)
    ✓ renders with different displayType (7 ms)
    ✓ renders with multiple panels (20 ms)
    ✓ renders with multiple small panels (9 ms)
    ✓ renders with custom displayTitle (5 ms)
    ✓ renders without displayTitle (4 ms)
  ClayPanel Interactions
    ✓ clicking the title should expand and close the content (21 ms)

```

packages/clay-panel/src/index.tsx
```diff
@@ -135,6 +151,7 @@ function ClayPanel({
 			{collapsable && (
 				<>
 					<ClayButton
+						aria-controls={`panel-index-${indexOfElement}`}
```
ethib137, line 152: Hey @AkramYoussoufi can you look at other instances of how we generate an ID for aria-controls and copy that pattern?

ethib137 (changes requested):

AkramYoussoufi:
@ethib137 Done!

packages/clay-panel/src/index.tsx
```diff
@@ -201,7 +204,6 @@ function ClayPanel({
 						onExiting={(element) => {
 							element.style.height = '';
 						}}
-						role="tabpanel"
```
matuzalemsteles, line 204: I was talking to Marcos about this and one of the things we can follow here is to use the w3c accessibility recommendation for accordion which is similar to panel https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/, we can add `role="region"` here instead of `tabpanel`. The other implementations part is ok too.

packages/clay-panel/package.json
```diff
@@ -18,7 +18,7 @@
 		"build": "cross-env NODE_ENV=production babel src --root-mode upward --out-dir lib --extensions .ts,.tsx",
 		"buildTypes": "cross-env NODE_ENV=production tsc --project ./tsconfig.declarations.json",
 		"prepublishOnly": "yarn build && yarn buildTypes",
-		"test": "jest --config ../../jest.config.js"
+		"test": "jest --config ../../jest.config.js --updateSnapshot"
```
matuzalemsteles, line 21: You can remove `updateSnapshot` here, we don't want this command by default because it will make the snapshot unusable we should only use it when we intend to update the markup. If you maintain this, we can update in a snapshot in cases where the markup change is a bug.

AkramYoussoufi:
@matuzalemsteles @ethib137 all the changes have been implemented

matuzalemsteles (approved):
LGTM! These storybook and website errors don't make any sense, I'll go ahead and see if the builds still fail on master.

PR #5717, opened 2023-11-14
fix(@clayui/css): Add increase text spacing on vertical navbar

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPS-201224

This was the solution I found to add the class letter spacing on the vertical nav. I thought that `!important` overrides all CSS applied on the page but what happening is that there is a very specific selector in the nav-vertical adding `letter-spacing` and the Increased Text Spacing style is not applied. 

I'm open to suggestions! 
Thanks!

cc: @pat270 @matuzalemsteles  @ethib137


ethib137:
Hey @ilzamcmed I'm concerned this is too targeted an approach. For instance, I've noticed tooltips also aren't currently being styled by this setting. The goal of this utility class is to apply to everything, so maybe we could do:
```
'.c-prefers-letter-spacing-1 *': (
	letter-spacing: 1px !important,
),
```

@pat270 would that be a workable solution? Are there any downsides to this approach?

ilzamcmed:
Thanks @ethib137!
I was thinking of that solution but according to my research, the universal selector has less priority than specific classes, that's why I chose to go that way.

packages/clay-css/src/scss/variables/_utilities.scss
```diff
@@ -753,6 +753,9 @@ $c-prefers-letter-spacing: map-deep-merge(
 	(
 		'.c-prefers-letter-spacing-1': (
 			letter-spacing: 1px !important,
+			'menubar-transparent .nav-link': (
```
pat270, line 756: I think @ethib137's suggestion using the universal selector is better. It's more invasive ensuring we don't get similar issues popping up if someone modifies `letter-spacing`. We still should have `letter-spacing: 1px !important` on `.c-prefers-letter-spacing-1` though.

```
'.c-prefers-letter-spacing-1, .c-prefers-letter-spacing-1 *': (
	letter-spacing: 1px !important,
),
```

should cover all cases. We don't need to worry about specificity with `!important`. Specificity only matters if the original value also has `!important` which we shouldn't use in our component CSS.

It's not related to this issue, but we can't nest `'*'` inside `.c-prefers-letter-spacing-1` due to Sass failing. It's probably an issue with our mixin, but I'm not 100% sure. It requires some research.

pat270 (changes requested):

packages/clay-css/src/scss/variables/_utilities.scss
ilzamcmed: Thanks for the explanation @pat270. It worked fine within the navbar items but not for the tooltip. 
Is there anything else I can do?

pat270: @ilzamcmed just change the whole thing to
```
'.c-prefers-letter-spacing-1, .c-prefers-letter-spacing-1 *': (
	letter-spacing: 1px !important,
),
```

ilzamcmed: @pat270  I`ve changed it according to your suggestion and is not working for tooltips.

pat270: @ilzamcmed I think it's because tooltips are placed outside of the scope for `.c-prefers-letter-spacing-1` in DXP.

ethib137: @ilzamcmed , where is this not working for tooltips? When I test it, it works fine. 

@pat270 the `c-prefers-letter-spacing-1` class is already being added to the body element, so I'm surprised we are seeing any issues.

ilzamcmed: @ethib137, I was testing on the storybook but maybe I was doing something wrong. I've sent Patrick's suggestion.

ethib137: I just tried on the storybook and first time I tried adding it to the wrong body class. You need to make sure it's added the body element inside of the iframe that is rendering the tooltip component. I'm guessing that's the issue you saw.

pat270 (approved):

ilzamcmed:
Thanks @pat270!!

pat270:
@ilzamcmed Np! This LGTM.

PR #5726, opened 2023-11-20
feat(@clayui/table): updates the table look

matuzalemsteles:
From https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2540961873/Border+Radius

With the changes in the new component and FDS initiative, this brought changes to the look of the table, such as striped by default and bordered which is only in the header, it also brought new sizes. This PR does not apply all these styles because we still need to work on the CSS part. This PR only introduces sections and striped.

### ToDo

- [x] New sizes and current update
- [x] New bordered for header only https://liferay.atlassian.net/browse/LPS-200555

### DIFF
--- a/packages/clay-core/src/table/Cell.tsx
+++ b/packages/clay-core/src/table/Cell.tsx
@@ -114,7 +114,8 @@ export const Cell = React.forwardRef<HTMLTableCellElement, Props>(
 			onFocusChange: setIsFocused,
 		});
 		const scope = useScope();
-		const {expandable, isLoading, key, lazy, level, loadMore} = useRow();
+		const {divider, expandable, isLoading, key, lazy, level, loadMore} =
+			useRow();
 
 		const isHead = scope === Scope.Head;
 		const As = isHead ? 'th' : 'td';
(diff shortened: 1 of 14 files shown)


matuzalemsteles:
cc @pat270

ethib137:
This looks good. We should just be especially careful when it comes time to release this in Portal that there aren't any UI issues as a result of this change. For instance does this play well with all of our other table utility classes. Also, there may be other styles added in some instances of it's usage that is not handled by us.

matuzalemsteles:
> Also, there may be other styles added in some instances of it's usage that is not handled by us.

Yes that is true, maybe we will need to update in some places like old taglibs I believe.

matuzalemsteles:
I sent the updated sizes to Table and also new API to define the size. @pat270 could you review if this is ok?

pat270 (approved):

pat270:
@matuzalemsteles I think I got everything to spec, but I might have missed something. This is what it looks like

fds

ethib137:
@pat270 looks good. Just one thing looks off to me. The last column with the ellipsis buttons should be sized differently, right?

matuzalemsteles:
@pat270 hover does not affect striped, would this be the expected behavior?

matuzalemsteles:
I'm going ahead with this.

pat270:
@matuzalemsteles We can add hover in another pr.

PR #5727, opened 2023-11-21
fix(@clayui/core): fixes bug when rendering LiveAnnouncer on server side

matuzalemsteles:
This is a report made by the Cloud team, they use server side rendering. So we need to check that we are not using the browser API when rendering on the server side.


ygorcosta (approved):
LGTM, thank you for the fix! <3

PR #5730, opened 2023-11-28
LPS-202818 Update version

p2kmgcl:
> Dear reviewer, [there is still an ongoing conversation on slack](https://liferay.slack.com/archives/C049XG98GQL/p1700832971192009) about how to handle this new design, I originally suggested allowing any custom header in the `clay:panel` taglib, like we do with the React component, but I think it is better, in terms on consistency, to keep a reduced number of cases as far as we can.
> 
> I wanted to update the code beforehand to be sure that this was feasible and not an ugly patch, but feel free to close this PR or request an alternative if this is a bad idea.

This PR adds a new displayType for ClayPanel component, matching a new design that we need in portal to implement the following UI:

https://www.figma.com/file/zf4akVPokkJdRKvgQtHxC0/LPS-114700--Translation-improvements-for-web-content?node-id=1144%3A16180&mode=dev

There will be a another PR adding this new option to the clay:panel taglib, which is being used for this case.

image

### DIFF
--- a/packages/clay-css/src/scss/variables/_panels.scss
+++ b/packages/clay-css/src/scss/variables/_panels.scss
@@ -94,6 +94,71 @@ $panel-title-text-transform: null !default;
 $panel-title-small-font-size: null !default;
 $panel-title-small-margin-left: null !default;
 
+$panel: () !default;
+$panel: map-deep-merge(
+	(
+		background-color: $panel-bg,
+		border-color: $panel-border-color,
+		border-radius: clay-enable-rounded($panel-border-radius),
+		border-style: $panel-border-style,
+		border-width: $panel-border-width,
+		box-shadow: clay-enable-shadows($panel-box-shadow),
+		margin-bottom: $panel-margin-bottom,
+		word-wrap: break-word,
+		collapse-icon-middle: (
+			collapse-icon: (
+				font-size: inherit,
+			),
+		),
+		panel-header: (
+			border-bottom: $panel-header-border-bottom-width solid transparent,
+			border-top-left-radius:
+				clay-enable-rounded($panel-header-offset-border-radius),
+			border-top-right-radius:
+				clay-enable-rounded($panel-header-offset-border-radius),
+			display: block,
+			font-size: $panel-header-font-size,
+			line-height: $panel-header-line-height,
+			padding: $panel-header-padding-y $panel-header-padding-x,
+			position: relative,
+			width: 100%,
+			collapsed: (
+				border-bottom-left-radius:
+					clay-enable-rounded($panel-header-offset-border-radius),
+				border-bottom-right-radius:
+					clay-enable-rounded($panel-header-offset-border-radius),
+				link: $panel-header-collapsed-link,
+			),
+			collapse-icon: (
+				padding-left: $panel-header-collapse-icon-padding-left,
+				padding-right: $panel-header-collapse-icon-padding-right,
+			),
+			link: $panel-header-link,
+		),
+		panel-body: (
+			padding: $panel-body-padding-y $panel-body-padding-x,
+		),
+		panel-footer: (
+			border-bottom-left-radius:
+				clay-enable-rounded($panel-footer-offset-border-radius),
+			border-bottom-right-radius:
+				clay-enable-rounded($panel-footer-offset-border-radius),
+			border-top: $panel-footer-border-top-width solid transparent,
+			padding: $panel-footer-padding-y $panel-footer-padding-x,
+		),
+		panel-title: (
+			font-size: $panel-title-font-size,
+			font-weight: $panel-title-font-weight,
+			text-transform: $panel-title-text-transform,
+			small: (
+				font-size: $panel-title-small-font-size,
+				margin-left: $panel-title-small-margin-left,
+			),
+		),
+	),
+	$panel
+);
+
 // Panel Group
 
 $panel-group-panel-margin-bottom: math-sign(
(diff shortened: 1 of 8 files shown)


matuzalemsteles:
Just to note that we are still waiting for the design decision https://liferay.atlassian.net/browse/LPS-203640.

p2kmgcl:
@matuzalemsteles I've updated the code with the renaming requested by David. Now this is large/small instead of primary/secondary.

packages/clay-css/src/scss/variables/_panels.scss
```diff
@@ -159,23 +159,48 @@ $panel-group-flush-panel-unstyled: map-deep-merge(
 
 // Panel Variants
 
-$panel-secondary-header-link: () !default;
+$panel-large-header-link: () !default;
 
-$panel-secondary: () !default;
-$panel-secondary: map-deep-merge(
```
matuzalemsteles, line 165: @p2kmgcl I know we need to change this here by design decision but this is something that has been in Clay for so long that we don't know who is using it here, so I think the best way to handle the case of `panel- secondary` is making an alias for `panel-small` the same for the React Component to ensure that whoever has `panel-secondary` still works but under the hood it will be `panel-small` in the next major release, we will be confident that we can remove this completely. Does it make sense to you?

p2kmgcl: Yes it does :smile:

matuzalemsteles (approved):
LGTM! @pat270 can you take a look at this in the SASS side before going ahead with this?

packages/clay-css/src/scss/variables/_panels.scss
pat270: @p2kmgcl This is too breaking imo. I think it's better if we keep `panel-large` and `panel-small` as separate modifiers that we can use on `panel` or `panel-secondary`. Those variants generally modify colors while `sm`, `md`, `lg` modify sizing such as padding, margins, font-size. We should also rename `panel-large` and `panel-small` to `panel-lg` and `panel-sm` since this is the naming convention used in Clay.

p2kmgcl: I agree that this breaks existing behavior, but using two different variation types for colors and sizes will create variants that we do not need. If I understood it well you're suggesting:

- `displayType: secondary | unstyled`
- `size: lg | md | sm`

but with that options, we could, for example, do this:

`<ClayPanel size="lg" displayType="secondary" />`, which we do not need at all.

Also, we do not have defined styles for `md`, we would need some extra design work for that.

p2kmgcl: Thanks @ethib137, I'll wait for Emiliano then :smile:

ethib137:
@p2kmgcl here are the updated design requirements:

Updated Design

https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/2608890286/Panel 

API

- displayType: ‘default’ | ‘block’
    - We should deprecate ‘unstyled’ and ‘secondary’.
- size: ‘sm’ | ‘regular’ | ‘lg’
    - using ‘sm’ and ‘lg’ matches button and sticker apis.
- When the displayType is `default` only `sm` and `regular` sizes should be allowed. We can limit this with TypeScript and with a check for `if (displayType === 'default' && size === 'lg')`.

p2kmgcl:
thanks @ethib137! Just a doubt. How do we mark `unstyled` and `secondary` as deprecated? Just updating the docs?

ethib137:
> thanks @ethib137! Just a doubt. How do we mark `unstyled` and `secondary` as deprecated? Just updating the docs?

Yeah @p2kmgcl , I think that's all we can do for now. Just updating [the comment here](https://github.com/liferay/clay/blob/master/packages/clay-panel/src/index.tsx#L52) to indicate something like "`unstyled` and `secondary` are deprecated. Please use 'default' and 'block' instead."

p2kmgcl:
Hi there! Sorry for the delay, but I had other priorities and had this PR pending :sweat_smile: 

I've just updated the component as @ethib137 described, but still need to update the styles accordingly. I think I'll need some help with this, because I am not sure how to proceed with the old CSS classes. @pat270 could you help me a little bit with that?

Also, for some reason I tried running the clayui site locally and the documentation is not being updated. Please correct me if I'm wrong, but I just need to update the comments in the IProps interface, right?

ethib137:
Thanks @p2kmgcl . Yeah updating the comment I believe should do it. The site would just have to be rebuilt. Maybe all the tasks weren't run.

@pat270 would you be able to help with the CSS for this?

packages/clay-panel/src/index.tsx
```diff
@@ -49,9 +54,10 @@ export interface IProps extends React.HTMLAttributes<HTMLDivElement> {
 	displayTitle?: React.ReactNode;
 
 	/**
-	 * Flag to indicate the visual variation of the Panel.
+	 * Display type. Deprecation warning: previous "unstyled" and "secondary"
+	 * styles are not longer supported.
```
ethib137, line 58: Wordsmithing: "previous "unstyled" and "secondary" styles are not longer supported." --> ""unstyled" and "secondary" variants are deprecated. They have been renamed to "default" and "block"."

p2kmgcl: Done!

pat270 (commented):
We can't get rid of `panel-unstyled` and `panel-secondary` due to breaking changes. We should eventually get rid of it after we make the proper markup changes in DXP with a fallback for custom implementations that might have depended on those classes.

A use case that's not covered in the Figma or docs is panels inside sheets.

sheet-panel

will the titles (password, organizations, user groups, and roles) be secondary in color or remain the same? Will the spacing be smaller or remain the same? I think we need more in depth discussion on this.

p2kmgcl:
> will the titles (password, organizations, user groups, and roles) be secondary in color or remain the same? Will the spacing be smaller or remain the same? I think we need more in depth discussion on this.

cc @emiliano-cicero I think this question is for you :sweat_smile:

emiliano-cicero:
hey @pat270, the new definition doesn't change the elements inside. 
Input's labels should be $Dark independently of the position. 
Panels should always use $Secondary for the text and $Secondary-L2 for the border to make them subtle

Making the content smaller for panels is currently out of scope for the system, it's something we could explore in the future.

pat270:
@matuzalemsteles I'm going to go ahead and merge this, it shouldn't affect panels in sheets. We can update those when the time comes.

p2kmgcl:
@pat270 awesome! Just one more question. When will this be updated in portal? Because I also need to update the taglib so I can use the new styles :sweat_smile:

matuzalemsteles:
@p2kmgcl I added this to include in the next release but we released one yesterday, so this should be out in 2 weeks after the change to every 2 week cycles. Do you need it now? If necessary, I can cut a release now and send it to DXP. Let me know if this is urgent.

p2kmgcl:
@matuzalemsteles is not urgent. We've been with this story for so long, but waiting a little bit won't harm anyone :smile:

PR #5733, opened 2023-12-01
fix(@clayui/tooltip): Fix close tooltip when pressing Escape key

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPS-165642

### DIFF
--- a/packages/clay-tooltip/src/TooltipProvider.tsx
+++ b/packages/clay-tooltip/src/TooltipProvider.tsx
@@ -131,7 +131,11 @@ const TooltipProvider = ({
 
 	const {close, isOpen, open} = useTooltipState({delay});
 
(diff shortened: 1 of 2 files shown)


packages/clay-tooltip/src/TooltipProvider.tsx
```diff
@@ -137,10 +137,8 @@ const TooltipProvider = ({
 			isHovered.current = false;
 		}, []),
 		onHide: useCallback(() => {
-			if (!isHovered.current && !isFocused.current) {
```
ilzamcmed, line 140: Hey @matuzalemsteles, the reason I removed this was that the `Esc` event wasn't going to work because on of the `isHovered` and `isFocused` is set to `true` whenever the tooltip is displayed.

matuzalemsteles (approved):
lgtm

PR #5736, opened 2023-12-08
fix(@clayui/core): updates the sort icon for table

matuzalemsteles:
Closes [LPS-202866](https://liferay.atlassian.net/browse/LPS-202866)

The icon update is still in draft, I need to make some adjustments, @pat270, would you have any ideas on how we can try to achieve these use cases of the sorting icon state? I don't know if we have something like this in CSS.

Even though I'm using the button, I'm making it invisible to SR because we're using the head for sorting for accessibility reasons, so when you press the whole head it changes the sorting if it's enabled for the column, so we can use `aria -sort`.

### DIFF
--- a/packages/clay-core/src/table/Cell.tsx
+++ b/packages/clay-core/src/table/Cell.tsx
@@ -168,6 +168,7 @@ export const Cell = React.forwardRef<HTMLTableCellElement, Props>(
 					'table-cell-ws-nowrap': !wrap,
 					'table-focus': focusWithinProps.tabIndex === 0 && isFocused,
 					'table-head-title': isHead,
+					'table-sort': isHead && sortable,
 				})}
 				colSpan={
 					divider
@@ -219,17 +220,13 @@ export const Cell = React.forwardRef<HTMLTableCellElement, Props>(
 							</span>
 						</Layout.ContentCol>
 						<Layout.ContentCol>
-							<Button
-								aria-label=""
+							<span
 								className="component-action"
-								displayType="unstyled"
-								monospaced
 								role="presentation"
-								size="xs"
 								tabIndex={-1}
 							>
 								<Icon symbol="order-arrow" />
-							</Button>
+							</span>
 						</Layout.ContentCol>
 					</Layout.ContentRow>
 				) : truncate ? (
--- a/packages/clay-core/src/table/Cell.tsx
+++ b/packages/clay-core/src/table/Cell.tsx
@@ -179,6 +179,16 @@ export const Cell = React.forwardRef<HTMLTableCellElement, Props>(
 						: undefined
 				}
 				className={classNames(className, {
+					'order-arrow-down-active': isSortable
+						? sort &&
+						  keyValue === sort.column &&
+						  sort.direction === 'descending'
+						: undefined,
+					'order-arrow-up-active': isSortable
+						? sort &&
+						  keyValue === sort.column &&
+						  sort.direction === 'ascending'
+						: undefined,
 					'table-cell-expand': truncate || expanded,
 					[`table-cell-${delimiter}`]: delimiter,
 					[`table-column-text-${textAlign}`]: textAlign,
@@ -186,7 +196,6 @@ export const Cell = React.forwardRef<HTMLTableCellElement, Props>(
 					'table-cell-ws-nowrap': !wrap,
 					'table-focus': focusWithinProps.tabIndex === 0 && isFocused,
 					'table-head-title': isHead,
-					'table-sort': isSortable,
 				})}
 				colSpan={
 					divider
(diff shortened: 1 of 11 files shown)


pat270:
@matuzalemsteles can we use a `span` tag for the icon and put `tabindex="0"` on `th` with a click event listener? I'm not sure if it breaks accessibility rules.

In my old examples, sorting in tables is using up or down arrow depending on the state

sort-icon

I can't find the new designs in the issue.

ethib137:
The designs were attached to the parent ticket: https://www.figma.com/file/c5DKGFa5SSkf6a6se0rJrg/LPS-196505-Clay-Table-Styles-and-Variations?type=design&node-id=19-5188&mode=design&t=wzrbh8sc8xLP05Sm-0

matuzalemsteles:
Actually now I'm a little confused the confluence document shows a different Figma icon https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2541093048/Header+Rows+Interactions.

@emiliano-cicero can you tell which one we should follow here?

ethib137:
@matuzalemsteles I think I must be missing something. It looks to me like [confluence](https://liferay.atlassian.net/wiki/spaces/PEDS/pages/2541093048/Header+Rows+Interactions) simply includes a screenshot of the [figma](https://www.figma.com/file/c5DKGFa5SSkf6a6se0rJrg/LPS-196505-Clay-Table-Styles-and-Variations?type=design&node-id=19-5188&mode=design&t=MyV1FNoIbHv5hSeT-0) documentation. What looks different to you?

matuzalemsteles:
Oh @ethib137 now it looks ok, when I looked at the icon in the table it is using `order-list-down` and `order-list-up`.

matuzalemsteles:
I went ahead and made some changes, I followed @pat270's idea of changing the button to span to move the focus to the `th` the idea was that but the focus went to the button which is bad for accessibility here because we need do focus on `th` so that SR announces the change using `aria-sort`. This is being done now it is possible to focus on the column in the head when treegrid is not enabled when the treegrid is enabled the focus is on the management of the first focusable element and it is possible to navigate to the head using the arrow keys, move the focus first to the cell and then navigate to the head.

I also added the CSS implementation to change the icon color depending on the sort. @pat270 let me know if that would be the best solution.

pat270:
> I followed @pat270's idea of changing the button to span to move the focus to the th the idea was that but the focus went to the button which is bad for accessibility here because we need do focus on th so that SR announces the change using aria-sort.


@matuzalemsteles @ethib137 @marcoscv-work  I dug a little deeper into the sortable table accessibility patterns and came upon https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/.

Problems with placing tab focus on `th`:
1) The screenreader doesn't announce we can interact with the `th`. We need `role="button"`.
2) Adding `role="button"` on the `th` clobbers any announcement related to the `th`.
3) In the w3 sortable-table pattern, there is no requirement for announcing `aria-sort` when navigating with tab. 

The announcements are correct when navigating with `ctrl + option + arrows` / `ctrl + alt + arrows`. I think we should revert back to a button for the order arrows with an `aria-label` indicating it changes the order for the table column.

ethib137:
Thanks @pat270 [this pattern](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) makes sense to me. @marcoscv-work does that look right to you? If so, @pat270 can you go ahead and make those updates?

pat270:
@ethib137 Thanks! I'll implement if it's all good with @marcoscv-work. Last thing, do you think we could check if `Liferay.Language.get('what-we-want-to-announce-for-sorting')` exists if not default to something like `aria-label="What we want to to announce for sorting"`?

ethib137:
> Last thing, do you think we could check if Liferay.Language.get('what-we-want-to-announce-for-sorting') exists if not default to something like aria-label="What we want to to announce for sorting"?
No we don't want to depend on Liferay. This is the pattern we normally use. We provide the label in English as a default value for a prop.

pat270:
I added the `sortAriaLabel` attribute for custom announcements on the sort button.

pedro-michel:
Any news here @pat270 ?

packages/clay-core/src/table/Cell.tsx
```diff
@@ -92,6 +97,7 @@ export const Cell = React.forwardRef<HTMLTableCellElement, Props>(
 			expanded,
 			index,
 			keyValue,
+			sortAriaLabel = 'Sort Column',
```
matuzalemsteles, line 100: We're building a standard to use a single API called `messages`, I'm going to move this there so I can just add it in one place.

https://github.com/liferay/clay/blob/c5675c0443b4ec25273a4172589e3656cdaabfd7/packages/clay-core/src/table/Table.tsx#L54

PR #5739, opened 2023-12-13
feat(@clayui/core): adds column visibility feature

matuzalemsteles:
Ticket [LPS-202287](https://liferay.atlassian.net/browse/LPS-202287)

This PR is still a draft but it adds the main functionality of being able to hide the column, by default we are leaving this functionality always activated but it is possible to deactivate it using the `columnsVisibility` API.

There are still visual details to adjust that are related to DropDown, like the DropDown header and the new switch size, I will work on that tomorrow.

### DIFF
--- a/packages/clay-core/package.json
+++ b/packages/clay-core/package.json
@@ -34,6 +34,7 @@
 		"@clayui/provider": "^3.106.1",
 		"@clayui/shared": "^3.107.0",
 		"@clayui/table": "^3.108.0",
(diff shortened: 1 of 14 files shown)


matuzalemsteles:
This PR is ready to be merged, there are still small details missing which is the new Switch small size as presented in Figma I'll let @pat270 work on this in another PR as he probably needs to change the mixins and other implementation details to create variations of size for the switch as is done for other components.

I will still send the commits to add the test cases for this component.

PR #5740, opened 2023-12-13
fix(@clayui/drop-down): Fix Dropdown doesnt close when click outside

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPS-203851

https://github.com/liferay/clay/assets/47724428/b00dd13b-7490-4346-8b88-2b936b25d347


matuzalemsteles:
@ilzamcmed I made some changes. We cannot close the Overlay at start because this way we will break the other behaviors of detecting the click outside correctly, click on input for upload and etc... so I just added support for clicking outside the window with blur but without changing the others events to not duplicate the event call.

PR #5758, opened 2024-02-01
fix(@clayui/core): fixes bug when not focusing on the first item

matuzalemsteles:
Ticket [LPD-16299](https://liferay.atlassian.net/browse/LPD-16299)

This fixes the bug when not being able to navigate via keyboard in the Columns Visibility of the table and improves accessibility, I'm still not sure what would be the best case here for toggle accessibility because DropDown is a menuitem list, so we still need to have the focus of the switch so that it announce correctly through SR, otherwise the announce will only be for the menuitem. Any idea?


matuzalemsteles:
cc @marcoscv-work can you help us with this here?

PR #5759, opened 2024-02-06
fix(@clayui/css): LPD-16364 @page {.css-selector{}} is not valid css

pat270:
- causes CSS to be removed when converting CSS to rtl

https://liferay.atlassian.net/browse/LPD-16364


matuzalemsteles (approved):
LGTM

PR #5764, opened 2024-02-08
LPD-17368 Accessibility issue in Pagination Picker

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-17368

### DIFF
--- a/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
+++ b/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
@@ -105,6 +105,8 @@ interface IProps extends React.ComponentProps<typeof PaginationBar> {
 	 * Use this property for i18n.
 	 */
 	labels?: {
+		itemsPerPagePickerAriaLabel?: string;
+
 		paginationResults: string;
 
 		perPageItems: string;
@@ -148,6 +150,7 @@ interface IProps extends React.ComponentProps<typeof PaginationBar> {
 }
 
 const DEFAULT_LABELS = {
+	itemsPerPagePickerAriaLabel: 'Items Per Page',
 	paginationResults: 'Showing {0} to {1} of {2}',
 	perPageItems: '{0} items',
 	selectPerPageItems: '{0} items',
@@ -209,6 +212,8 @@ export const ClayPaginationBarWithBasicItems = ({
 		value: typeof active === 'undefined' ? activePage : active,
 	});
 
+	labels = {...DEFAULT_LABELS, ...labels};
+
 	if (!activeDelta) {
 		activeDelta = deltas[0]!.label;
 	}
@@ -230,6 +235,7 @@ export const ClayPaginationBarWithBasicItems = ({
 					<Picker
 						activeDelta={activeDelta}
 						aria-describedby={ariaDescribedby}
+						aria-label={labels.itemsPerPagePickerAriaLabel}
 						as={Trigger}
 						items={deltas}
 						labels={labels}
--- a/packages/clay-pagination-bar/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-pagination-bar/src/__tests__/__snapshots__/index.tsx.snap
@@ -13,6 +13,7 @@ exports[`ClayPaginationBar renders 1`] = `
         aria-describedby="clay-id-1"
         aria-expanded="false"
         aria-haspopup="listbox"
+        aria-label="Items Per Page"
         class="dropdown-toggle btn btn-unstyled"
         role="combobox"
         tabindex="0"


packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -230,6 +233,7 @@ export const ClayPaginationBarWithBasicItems = ({
 					<Picker
 						activeDelta={activeDelta}
 						aria-describedby={ariaDescribedby}
+						aria-label="Pagination"
```
matuzalemsteles, line 236: I think this should be `aria-label={labels.pagination}`

ilzamcmed: Thanks @matuzalemsteles 
Fixed!

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -230,6 +233,7 @@ export const ClayPaginationBarWithBasicItems = ({
 					<Picker
 						activeDelta={activeDelta}
 						aria-describedby={ariaDescribedby}
+						aria-label={labels.pagination}
```
ethib137, line 236: We need a better name for this variable. This is the aria label for the pagination delta picker. In general our props that are used only as an aria-label, include aria label in the name. If we use this as the title (tooltip) then we could not include ariaLabel in the name. The [date picker](https://github.com/liferay/clay/blob/af20c5531b71b89cd5b95c5805f8e8d8b6915dfb/packages/clay-date-picker/src/DateNavigation.tsx#L17) and [color picker](https://github.com/liferay/clay/blob/af20c5531b71b89cd5b95c5805f8e8d8b6915dfb/packages/clay-color-picker/src/index.tsx#L331) include an `aria-labels` prop that adds all aria labels to that object. So that's another option, but since we only have one aria-label currently that might be too much. So maybe `pageDeltaPickerAriaLabel` or `itemsPerPagePickerTitle` of it makes sense to add this to a title/tooltip then `pageDeltaPickerTitle` or `itemsPerPagePickerTitle`.

Additionally, we need to think about the default label. Maybe `Items Per Page`. This would result in the screen reader reading: "Items Per Page Combo box, 20 items." @marcoscv-work What do you think?

matuzalemsteles: We used to use `ariaLabels` to group all of this due to internationalization but last year we changed to using a name that could group `ariaLabels` and other texts, so we are using `messages` now.

I wouldn't add it now because it would introduce a new separate API, but I would keep it as is using `labels` and move it to `messages` in the main version.

matuzalemsteles:
> do we already have a process for making sure we apply these additional changes to portal as part of the Clay release?

In fact we have no process for this, when we add a new key, we make this key optional when it is inside the object to avoid a breaking change, for example when the component has already declared `labels` but a new key was added this would cause a build error. So this is treated as breaking change.

We normally do this but we don't update all DXP use cases due to workforce so as not to delay the release. But we can include this or we can start prioritizing the work on integrating `Liferay.Language` with the portal so that we only need to change in one place.

ethib137:
Thanks @matuzalemsteles , that makes sense.

> But we can include this or we can start prioritizing the work on integrating Liferay.Language with the portal so that we only need to change in one place.

I think it makes more sense to prioritize integrating Liferay.Language, like we've discussed before. Something like a ClayMessageProvider. Do you remember where we discussed this previously. I think it was on a ticket or something. I'll create a new story for this and start working on the requirements.

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
marcoscv-work: > 
> Additionally, we need to think about the default label. Maybe `Items Per Page`. This would result in the screen reader reading: "Items Per Page Combo box, 20 items." @marcoscv-work What do you think?

Sounds much better for me "Items Per Page" for the picker, label much more help

ethib137: Thanks @marcoscv-work.

So the variable should probably be called `itemsPerPagePickerAriaLabel` and the default value should be "Items Per Page". Does that sound okay?

matuzalemsteles:
> I think it makes more sense to prioritize integrating Liferay.Language, like we've discussed before. Something like a ClayMessageProvider. Do you remember where we discussed this previously. I think it was on a ticket or something. I'll create a new story for this and start working on the requirements.

Yes, something like that. We have a ticket that discussed this but I think we need to create another one with this type of story. https://liferay.atlassian.net/browse/LPS-190534

matuzalemsteles (approved):
LGTM

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -105,6 +105,8 @@ interface IProps extends React.ComponentProps<typeof PaginationBar> {
 	 * Use this property for i18n.
 	 */
 	labels?: {
+		pageDeltaPickerAriaLabel?: string;
```
ethib137, line 108: Hey @ilzamcmed  can you switch the name of the variable to `itemsPerPagePickerAriaLabel` to align with [this comment](https://github.com/liferay/clay/pull/5764#discussion_r1486105251)? The other variables use perpage instead of delta, so we should be consistent.

ethib137 (changes requested):

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
ilzamcmed: Thanks @ethib137 
I think I missed this comment.  Everything is ok now.

ethib137 (approved):
LGTM. Thanks!

PR #5765, opened 2024-02-09
fix(@clayui/core): LPS-203851 fixes bug when keeping DropDown open in TreeView actions in Firefox

matuzalemsteles:
Ticket [LPS-203851](https://liferay.atlassian.net/browse/LPS-203851)

Well, basically this is a follow-up fix where the original fix only fixed the bug in Chrome and still kept reproducing in Firefox. After some tests it seems that the problem was always happening when the DropDown Item is a link, when clicking on the link the TreeView node's blur event is triggered first than the dropdown item, causing the dropdown trigger to disappear and the menu was repositioned and for some reason it didn't trigger the click on the DropDown item afterwards. I still haven't found a clear explanation about this behavior and because it happens in Firefox, my assumption is still the order of events and perhaps the cancellation midway.

To prevent the DropDown trigger from disappearing, I'm just adding that in the blur of the TreeView node the `event.relatedTarget` is different from null or undefined, this does the trick because the click needs to be on some valid node so we can check if the click it still happens in DropDown or outside.

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -265,6 +265,7 @@ export const TreeViewItem = React.forwardRef<
 					onBlur={(event) => {
 						if (
 							actions &&
+							event.relatedTarget &&
 							!item.itemRef.current?.contains(
 								event.relatedTarget as HTMLElement
 							)
--- a/packages/clay-core/stories/TreeView.stories.tsx
+++ b/packages/clay-core/stories/TreeView.stories.tsx
@@ -242,8 +242,12 @@ export const Actions = () => (
 					</Button>
 					<DropDownWithItems
 						items={[
-							{label: 'One'},
-							{label: 'Two'},
+							{
+								href: '#',
+								label: 'One',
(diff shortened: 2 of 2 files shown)


ethib137:
Thanks @matuzalemsteles! Do we have a sandbox example that we can reproduce the issue on? Have we been able to reproduce this outside of portal?

If we do have a sandbox example that can reproduce this issue, then we should also add that situation as a storybook sample so we can more easily test this in the future.

matuzalemsteles:
> Do we have a sandbox example that we can reproduce the issue on? Have we been able to reproduce this outside of portal?

I managed to reproduce it in the storybook using the same props that are defined in the Page Tree actions, I will create a sandbox to share.

> If we do have a sandbox example that can reproduce this issue, then we should also add that situation as a storybook sample so we can more easily test this in the future.

Yeah, I'll create a new example to illustrate the portal case.

matuzalemsteles:
I updated the storybook example by adding the case that reproduces this problem in Firefox, basically an item that is a link was causing the bug.

matuzalemsteles:
I'm going to merge this so I can include it in today's release.

PR #5766, opened 2024-02-09
fix(@clayui/core): fixes bug when visual focus is lost when items are updated in Picker

matuzalemsteles:
Thread https://liferay.slack.com/archives/C049XG98GQL/p1707334836231499

When the items are updated, the visual focus is stale, meaning that when navigating via the keyboard it will not work because the key does not exist in the list, so we need to update the visual focus when the list is updated during the component life cycle.


ethib137 (approved):
LGTM

PR #5767, opened 2024-02-16
fix(clayui.com) Change tooltip background color

SelenaAungst:
https://liferay.atlassian.net/browse/LPD-2053
Please let me know if anything else is needed for this.


ethib137:
Hey @SelenaAungst the fix looks good, but the commit message looks like it has the wrong scope. Shouldn't this be `fix(@clayui/css)`?

SelenaAungst:
Thanks @ethib137! I updated the commit message.

PR #5776, opened 2024-03-04
fix(@clayui/core): fixes bug when switching column visibility

matuzalemsteles:
Ticket [LPD-19431](https://liferay.atlassian.net/browse/LPD-19431)

### DIFF
--- a/packages/clay-core/src/table/Head.tsx
+++ b/packages/clay-core/src/table/Head.tsx
@@ -54,22 +54,22 @@ function HeadInner<T extends Record<string, any>>(
 ) {
 	const {
 		columnsVisibility,
-		hiddenColumns,
 		messages,
 		onHeadCellsChange,
-		onHiddenColumnsChange,
+		onVisibleColumnsChange,
+		visibleColumns,
 	} = useTable();
 
 	const collection = useCollection<T>({
 		children,
 		items,
 		suppressTextValueWarning: false,
-		visibleKeys: new Set(hiddenColumns.keys()),
+		visibleKeys: new Set(visibleColumns.keys()),
 	});
 
 	useMemo(() => {
-		if (hiddenColumns.size === 0) {
-			onHiddenColumnsChange(collection.getItems(), 0);
+		if (visibleColumns.size === 0) {
+			onVisibleColumnsChange(collection.getItems(), 0);
 		}
 		onHeadCellsChange(collection.getSize());
 	}, []);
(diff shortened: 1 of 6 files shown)


ethib137:
Hey @matuzalemsteles once you select a column to be visible it's rendering it correctly, but the default state is still wrong. When all the columns are shown, they should all be toggled on.
Screenshot 2024-03-05 at 3 53 20 PM

matuzalemsteles:
@ethib137 Oh thank you so that wasn't very clear to me..

matuzalemsteles:
@ethib137 It looks good now, I fixed a bug in the disabled state behavior too, everything was reversed. Let me know if I missed anything.

packages/clay-core/src/table/Head.tsx
```diff
@@ -63,12 +62,15 @@ function HeadInner<T extends Record<string, any>>(
 
 	const collection = useCollection<T>({
 		children,
-		disabledKeys: new Set(hiddenColumns.keys()),
 		items,
 		suppressTextValueWarning: false,
+		visibleKeys: new Set(hiddenColumns.keys()),
```
ethib137, line 67: Hey @matuzalemsteles this looks problematic setting the value of `visibleKeys` to `hiddenColumns.keys()`. We need to use either visible or hidden throughout.

I think `hiddenKeys` makes more sense because we start with all keys as visible and then hide keys one by one. So, the default state is no keys are hidden which means `hiddenKeys` would be empty. When we hide one key, then `hiddenKeys` adds one item. 

If we use `visibleKeys` then the default value would need to include all keys, for it to make sense. That could work, but it still feels less intuitive than tracking which keys are hidden.

What do you think?

ethib137: By the way, the functionality is now working perfectly! 🎉

matuzalemsteles: Yes, I totally agree.

ethib137 (approved):
Hey @matuzalemsteles I was expecting you to use `hidden` for everything NOT `visible`. But I'm assuming that when you were writing the code this way made more sense, so I trust your judgement. It looks consistent now, so this should be fine to merge. Thank you.

matuzalemsteles:
> I was expecting you to use hidden for everything NOT visible. But I'm assuming that when you were writing the code this way made more sense, so I trust your judgement

Oh yes, I tried using `hidden` for everything but it seemed strange due to the behavior we have to start with all items in the object that are visible so that the checkbox state works correctly when the user unchecks it, so as not to create too much logic on top of that, so that `hidden` is empty and some extra conditionals in the checkbox so that it is inverted internally, I preferred to go with visible to just make the component logic easier.

PR #5777, opened 2024-03-04
feat(@clayui/css): LPD-19394 Mark update colors and add a dark version

pat270:
https://liferay.atlassian.net/browse/LPD-19394

highlight-text


matuzalemsteles (approved):
LGTM! @emiliano-cicero I saw that the mockup example uses mark in autocomplete, should we also update this to autocomplete?

ethib137:
@drakonux or @marcoscv-work Could you take a look at @matuzalemsteles 's question above?

matuzalemsteles:
Well, if we need to update this in Autocomplete we can also create a new ticket. I'm going to merge this to include it in today's release and not block the update.

PR #5779, opened 2024-03-06
fix(@clayui/pagination): LPD-1285 accessibility issues on prev, next …

pat270:
…and ellipsis buttons

https://liferay.atlassian.net/browse/LPD-1285

This adds `role="button"` to the prev and next buttons. Axe flags these because there is no href attribute. This also adds `aria-controls="clay-id-#"` on the ellipsis button when it is active. The dropdown-menu is lazy loaded and Axe can't associate `aria-controls` to the dropdown-menu.

Axe Errors
accessibility-errors

After Fix
accessibility-fixed

### DIFF
--- a/packages/clay-pagination/src/PaginationWithBasicItems.tsx
+++ b/packages/clay-pagination/src/PaginationWithBasicItems.tsx
@@ -164,6 +164,11 @@ const ClayPaginationWithBasicItems = React.forwardRef<HTMLUListElement, IProps>(
 					disabled={internalActive === 1}
 					href={previousHref}
 					onClick={() => setActive(previousPage)}
+					role={
+						previousHref || internalActive === 1
+							? undefined
+							: 'button'
+					}
 				>
 					<ClayIcon spritemap={spritemap} symbol="angle-left" />
 				</Pagination.Item>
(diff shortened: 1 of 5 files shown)


packages/clay-pagination/src/PaginationWithBasicItems.tsx
```diff
@@ -164,6 +164,7 @@ const ClayPaginationWithBasicItems = React.forwardRef<HTMLUListElement, IProps>(
 					disabled={internalActive === 1}
 					href={previousHref}
 					onClick={() => setActive(previousPage)}
+					role={internalActive === 1 ? undefined : 'button'}
```
ethib137, line 167: In both these cases shouldn't we be checking if the href is actually empty or null? Don't we sometimes use an href for going to the previous or next page, in which case the role should stay a link and not a button, right?

ethib137 (changes requested):

packages/clay-pagination/src/PaginationWithBasicItems.tsx
pat270: Good point, I'll update it

matuzalemsteles (approved):
LGTM

packages/clay-pagination/src/PaginationWithBasicItems.tsx
```diff
@@ -164,6 +164,11 @@ const ClayPaginationWithBasicItems = React.forwardRef<HTMLUListElement, IProps>(
 					disabled={internalActive === 1}
 					href={previousHref}
 					onClick={() => setActive(previousPage)}
+					role={
+						previousHref || internalActive === 1
```
ethib137, line 168: If there is a previousHref, it will be a string which would be a truthy value. Then this would be set as the role, right? So I don't think this logic wouldn't be correct when previousHref exists.

pat270: If `previousHref` exists, the pagination component outputs `<a class="page-link" href="/url">`. In this case, we don't need `role="button"` because it's a hyperlink.

PR #5785, opened 2024-03-16
docs(clayui.com): Remove secondary, tertiary, and danger link examples.

SelenaAungst:
https://liferay.atlassian.net/browse/LPD-17393
I was not 100% sure that these are all the changes being requested so please let me know if anything is missing/incorrect from this!


matuzalemsteles (commented):
@SelenaAungst It seems that according to the story we need to remove the size examples too.

> Clay needs to remove size examples an color examples. they can still bring text examples about the properties the component accept but we don't need to represent them visually. We need to persist on accessibility examples, like the link with underline.

Apparently it seems that we also don't need to add the example of the link without the underline but we can at least describe it if that's what I understand in this text.

SelenaAungst:
Hi @matuzalemsteles,
Thanks for pointing this out! I removed the Font Size example. Additionally, I was not sure if this line should be changed as well:

image

"The available values are 'primary', 'secondary', 'danger', and 'tertiary'." 
Should this just be changed to 
"The available value is 'primary'." 
or should this part be removed now that the other options are gone?

matuzalemsteles:
@SelenaAungst I think we can remove and leave this documented only in the API part at this point.

SelenaAungst:
Thanks @matuzalemsteles! I removed that part. Does anything else need to be added/removed for this?

matuzalemsteles:
@SelenaAungst I think that's just it. Thanks!

@drakonux or @emiliano-cicero can you guys validate this before merging?

SelenaAungst:
From a [Slack discussion](https://liferay.slack.com/archives/C04A7NLM7L1/p1710912135587739) of this ticket:

> For links, I think that at the content level, and as Matu says, we should delete references to the rest of displayTypes, we can remove the whole reference, even the sentence “You can style a link…“.
> And actually Semi Bold, Outline, Borderless and Monospaced sections could be also removed, if it’s clear on the API tab, it’s enough

I will add a new commit with these changes.

matuzalemsteles (approved):

PR #5789, opened 2024-03-20
feat: add cron job to run clayui.com's nightly build

matuzalemsteles:
Before merging this we need to add the build hook to the secret of the Github repository, I think we need to open a ticket for this.


ethib137 (approved):
This looks great. Thanks @matuzalemsteles. Please follow up with creating the ticket to add a secret. Please add me as a watcher on any related requests. Thanks.

PR #5794, opened 2024-04-01
fix(@clayui/pagination-bar): LPD-17368 Follow up - Add aria-label

ilzamcmed:
This is a follow-up to add aria-label on the pagination bar using DropDown, in addition to the Picker.
https://liferay.atlassian.net/browse/LPD-17368

### DIFF
--- a/packages/clay-drop-down/src/DropDownWithItems.tsx
+++ b/packages/clay-drop-down/src/DropDownWithItems.tsx
@@ -141,6 +141,11 @@ export interface IProps
 	menuWidth?: React.ComponentProps<typeof ClayDropDown>['menuWidth'];
 
 	/**
+	 * Prop to use language keys.
+	 */
+	messages?: string;
+
+	/**
 	 * Function for setting the offset of the menu from the trigger.
 	 */
 	offsetFn?: React.ComponentProps<typeof ClayDropDown>['offsetFn'];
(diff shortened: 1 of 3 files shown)


matuzalemsteles (approved):
LGTM

packages/clay-drop-down/src/DropDownWithItems.tsx
```diff
@@ -140,6 +140,11 @@ export interface IProps
 
 	menuWidth?: React.ComponentProps<typeof ClayDropDown>['menuWidth'];
 
+	/**
+	 * Prop to use language keys.
+	 */
+	messages?: string;
```
pat270, line 146: Can we make this prop `ariaLabel`? I think it's a clearer name for it.

pat270 (changes requested):

matuzalemsteles:
@ilzamcmed can you update pagination test threshold?

packages/clay-drop-down/src/DropDownWithItems.tsx
ilzamcmed: Hey @pat270, the reason I added this was because we have the `messages` prop in other components that have the same purpose, add language keys.

pat270: I see.

pat270 (approved):

PR #5797, opened 2024-04-04
feat(@clayui/card): Radio Card component variant

veroglez:
Hi guys!

According to this issue https://liferay.atlassian.net/browse/LPD-1261, on the Echo team we need a new variant of cards that have radio buttons, so here I'm sending  a solution for this.

A new prop has been added, which is `selectableType`. This prop defines whether the selectable element will have a checkbox or a radio button. In addition, A `radioProps` prop is also added to be able to pass the necessary props to the radio input. Along with all this, a new case in storybook has been added with two radio cards, to see the behavior of these cards:

https://github.com/liferay/clay/assets/9701095/b414ad66-bb10-4c40-9684-cfe88b0cda4a

Finally, to give the user the option of not truncating the text (it's another requirement in our design and in general accessibility), a new `truncate` prop has been added to CardWithInfo, which is a flag that determines if the text is truncate or not.

Let me know what you think about the solution, any feedback is welcome :smile:. 

Thanks in advance!

cc @marcoscv-work

### DIFF
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -114,6 +114,11 @@ export interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	 * Name of the file
 	 */
 	title: string;
+
+	/**
+	 * Flag to indicate if the card text is truncated
+	 */
+	truncate?: boolean;
 }
 
 /**
@@ -156,6 +161,7 @@ export const ClayCardWithInfo = ({
 	stickerProps = {},
 	symbol,
 	title,
+	truncate = true,
 	...otherProps
 }: IProps & (RadioProps | CheckboxProps)) => {
 	const isCardType = {
@@ -263,12 +269,16 @@ export const ClayCardWithInfo = ({
 							disabled={disabled}
 							displayType="title"
 							href={href}
+							truncate={truncate}
 						>
 							{title}
 						</ClayCard.Description>
 
 						{description && (
-							<ClayCard.Description displayType="subtitle">
+							<ClayCard.Description
+								displayType="subtitle"
+								truncate={truncate}
+							>
 								{description}
 							</ClayCard.Description>
 						)}
--- a/packages/clay-card/src/CardWithInfo.tsx
+++ b/packages/clay-card/src/CardWithInfo.tsx
@@ -29,6 +29,12 @@ export interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	checkboxProps?: React.HTMLAttributes<HTMLInputElement>;
 
 	/**
+	 * Props to add to the radio element
+	 */
+
+	radioProps?: React.HTMLAttributes<HTMLInputElement> & {value: string};
+
+	/**
 	 * Description of the file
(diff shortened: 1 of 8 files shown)


packages/clay-card/src/CardWithInfo.tsx
```diff
@@ -83,7 +89,12 @@ export interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	/**
 	 * Callback for when item is selected
 	 */
-	onSelectChange?: (val: boolean) => void;
+	onSelectChange?: (val: string | boolean) => void;
```
matuzalemsteles, line 92: My only concern is that this may break the current components due to the new type... I'm thinking about how we can avoid this.

matuzalemsteles (commented):
Thanks @veroglez for this! Would this perhaps be interesting for other cards as well?

packages/clay-card/src/CardWithInfo.tsx
veroglez: Hey @matuzalemsteles! in order not to break the current components, I have added that the `onSelectChange` type will be different depending on whether the `selectableType` is radio or checkbox (de0d533de063f132d9ef65b21e9584abba0a3860). Let me know me what you think and if you think it's okay for you, I'll do the same for the rest of the cards.

veroglez: In another vein, the tests are failing because the clay-card threshold is not being met (in this PR the value is 88.88% not 90%). Is there any parameter that you take into account to increase this?

matuzalemsteles: > Let me know me what you think and if you think it's okay for you

Oh great, that will solve it. Thanks!

> In another vein, the tests are failing because the clay-card threshold is not being met (in this PR the value is 88.88% not 90%). Is there any parameter that you take into account to increase this?

We don't actually have a set value, but we always try to keep it above 50%, and we always increase it when coverage also increases, it's just a way to always keep coverage high. In this case, as you added new test cases to cover the behavior, you can lower the threshold without any problems.

veroglez:
Hi @matuzalemsteles! I've updated the PR by adding the radio selectable type and truncate props for CardWithUser and CardWithHorizontal. Could you take a look? 

On the other hand, the CI/stats has failed but I'm not sure what it refers to. Do I have to update something?

Thanks in advance! :smile:

matuzalemsteles:
@veroglez thanks for the work on this!

> On the other hand, the CI/stats has failed but I'm not sure what it refers to. Do I have to update something?

Ah, the failure is because the package size limit has been exceeded, you can increase this to 151kb without problems in `package.json` in the root of the repository.

https://github.com/liferay/clay/blob/master/package.json#L10

matuzalemsteles (approved):
LGTM! Thanks again @veroglez for working on this!

veroglez:
> Ah, the failure is because the package size limit has been exceeded, you can increase this to 151kb without problems in `package.json` in the root of the repository.
> 
> https://github.com/liferay/clay/blob/master/package.json#L10

Perfect! It seems that all checks have passed. Thank you very much for your help and your feedback!

PR #5801, opened 2024-04-09
chore(@clayui/form): add regular size to input storybook and make it …

pat270:
…the default value
https://liferay.atlassian.net/browse/LPD-17804

### DIFF
--- a/packages/clay-form/stories/Input.stories.tsx
+++ b/packages/clay-form/stories/Input.stories.tsx
@@ -12,7 +12,7 @@ export default {
 	argTypes: {
 		sizing: {
 			control: {type: 'select'},
-			options: ['lg', 'sm'],
+			options: ['lg', 'regular', 'sm'],
 		},
 	},
 	title: 'Design System/Components/Input',
@@ -27,7 +27,7 @@ export const Default = (args: any) => (
 				id="basicInputText"
 				placeholder="Insert your name here"
 				readOnly={args.readOnly}
-				sizing={args.sizing}
+				sizing={args.sizing === 'regular' ? null : args.sizing}
 				type="text"
 			/>
 		</ClayForm.Group>
@@ -37,7 +37,7 @@ export const Default = (args: any) => (
 Default.args = {
 	disabled: false,
 	readOnly: false,
-	sizing: undefined,
+	sizing: 'regular',
 };
 
 export const InputFeedback = () => (
--- a/packages/clay-button/src/Button.tsx
+++ b/packages/clay-button/src/Button.tsx
@@ -69,7 +69,7 @@ export interface IProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
 	/**
 	 * Determines the size of a button.
 	 */
-	size?: 'xs' | 'sm';
+	size?: 'xs' | 'regular' | 'sm';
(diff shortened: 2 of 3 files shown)


packages/clay-form/stories/Input.stories.tsx
```diff
@@ -12,7 +12,7 @@ export default {
 	argTypes: {
 		sizing: {
 			control: {type: 'select'},
-			options: ['lg', 'sm'],
+			options: ['lg', 'regular', 'sm'],
```
matuzalemsteles, line 15: Maybe this could be an option in the API directly instead of just in the storybook? we can add `regular` and if it is passed, we leave it as the default value. What do you think?

pat270: I don't have a strong opinion on this. Do you think it will make it clearer for everyone to understand sizing if we add a regular value? I always thought the default was implied if the sizing attribute wasn't specified.

matuzalemsteles: Well, I thought that too when `sizing` is not defined it would imply that it is using the regular value but I think that when you are defining the size it is strange for you to look at it and see that there is only lg and sm this is something else implicit than explicit so maybe we can just make it more explicit I think this is more aesthetic too.

ethib137: I think adding regular may make sense, but if we want to add that we should probably be applying that consistently. For instance button currently has the options: `null | xs | sm`. @drakonux were you thinking the `regular` option should be applied everywhere?

matuzalemsteles: Oh yes I agree with that, it makes a lot of sense to do this for all components.

pat270: @matuzalemsteles should we make these updates on this PR?

ethib137: @drakonux do you think it makes sense to make this change across all components?

drakonux: Yes @ethib137 . Let's make a decision and apply it consistently. We are using regular buttons and also for inputs in the Figma library. 

Our top priority is maintaining a consistent approach with Clay and ensuring a one-to-one correspondence at all components.

ethib137: Okay buttons and inputs sound good. Are there any other usages of regular that we should apply this too?

matuzalemsteles: I went ahead and added it to the Input and Button. We have some other components with `size` as well:

- Table
- Sheet
- LoadingIndicator
- Modal
- MultiSelect
- Pagination
- PaginationBar
- Panel
- Sticker

Does it make sense to add to these too? I would say yes since we are talk for consistency here but I will wait for other opinions to see if I'm missing something.

ethib137: @drakonux what do you think?

matuzalemsteles: Okay, I think we can move forward with this and if it looks good we can do this for the other components just to not block this ticket any longer.

PR #5802, opened 2024-04-10
fix(@clayui/css): update footer spacing in the modal

matuzalemsteles:
Ticket [LPD-18361](https://liferay.atlassian.net/browse/LPD-18361)

### DIFF
--- a/packages/clay-css/src/scss/variables/_modals.scss
+++ b/packages/clay-css/src/scss/variables/_modals.scss
@@ -149,8 +149,8 @@ $modal-footer-border-width: $modal-header-border-width !default;
 $modal-footer-box-shadow: null !default;
 $modal-footer-color: null !default;
 $modal-footer-height: null !default;
-$modal-footer-padding-x: 0.75rem !default;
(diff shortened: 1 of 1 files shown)


pat270:
@matuzalemsteles we should update the padding in cadmin too. I was afraid the taller footer height might mess up some older modals in DXP, but it looks like we have DXP specific CSS for those.

pat270 (changes requested):
We need to update in cadmin too.

matuzalemsteles:
> We need to update in cadmin too.

Updated!

> I was afraid the taller footer height might mess up some older modals in DXP, but it looks like we have DXP specific CSS for those.

Hmm, I don't know if I understand very well, do you think this could affect old modals? visually what would be the side effect of this? I think changing the old modals would also be good since we want to standardize but I may be missing something here.

pat270:
The example I'm thinking about is the permissions modal. We have CSS in DXP for that.

permissions

pat270 (approved):

PR #5803, opened 2024-04-10
docs(clayui.com): Move input-group to own section

pat270:
https://liferay.atlassian.net/browse/LPD-22854


matuzalemsteles (approved):
LGTM! Just one doubt I still have regarding whether we should keep these examples that are in the Markup session or if we want to keep just a few use cases like we are doing in other components. @drakonux @emiliano-cicero thoughts? I didn't see anything explicit about this in the ticket or document just to know if this is ok.

drakonux:
Separating these items is recommended, especially since we'll need to provide extensive design documentation for input groups. If you could review how we're currently giving input groups in the Figma library, we may be able to improve the content.

image

After speaking with @marcoscv-work, it seems that input groups are used in multiple cases, so for us, they are just the ones included in the Figma Library.

matuzalemsteles:
Thanks @drakonux. @pat270 you can check this, apparently everything is ok.

pat270:
@matuzalemsteles I updated the examples to better match with examples in the Figma.

matuzalemsteles (approved):
LGTM

PR #5806, opened 2024-04-15
docs(@clayui/form): remove readonly checkbox examples

matuzalemsteles:
Ticket [LPD-17461](https://liferay.atlassian.net/browse/LPD-17461)

I removed all examples of using readonly from the documentation on clayui.com. I believe we also don't want to add support for this further into the component, I think we can mark the API as deprecated as well and remove it completely in the future.


pat270:
LGTM! I wish we never added it :(.

PR #5807, opened 2024-04-17
feat(@clayui/empty-state): LPD-21741 Remove changes on Empty State and add SVG files

ilzamcmed:


packages/clay-empty-state/src/index.tsx
```diff
@@ -103,60 +85,47 @@ const ClayEmptyState = ({
...
 			})}
 			{...otherProps}
 		>
 			{hasImg && (
 				<div className="c-empty-state-image">
-					{imgSrc && (
-						<div className="c-empty-state-aspect-ratio">
+					<div className="c-empty-state-aspect-ratio">
+						<img
+							alt=""
```
matuzalemsteles, line 105: Can you add it as part of the API to customize the alt?

packages/clay-empty-state/src/index.tsx
```diff
@@ -6,35 +6,35 @@
...
 import React, {useMemo, useState} from 'react';
 
-import {States} from './States';
-
 interface IProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title'> {
+	/**
+	 * Provides alternative information for an image if a user for some reason cannot view it
+	 * The default value is 'No results found'
+	 */
+	altProps?: string;
```
matuzalemsteles, line 14: I think we can use the value of `imgProps`.

ilzamcmed: We decided to remove the altProps and continue with `{...imgProps}`

matuzalemsteles (approved):
LGTM

PR #5826, opened 2024-05-27
refactor(@clayui/checkbox): LPD-1815 Remove readOnly property and add disabled

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-1815

### DIFF
--- a/packages/clay-form/src/Checkbox.tsx
+++ b/packages/clay-form/src/Checkbox.tsx
@@ -18,6 +18,11 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	containerProps?: React.HTMLAttributes<HTMLDivElement>;
 
 	/**
+	 * Props to disable checkbox.
+	 */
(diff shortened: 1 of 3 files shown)


packages/clay-form/src/Checkbox.tsx
```diff
@@ -76,13 +81,13 @@ const ClayCheckbox = React.forwardRef<HTMLInputElement, IProps>(
 				<label>
 					<input
 						{...otherProps}
-						aria-disabled={readOnly}
+						aria-disabled={disabled}
```
matuzalemsteles, line 84: As we already have the `disabled` property I think we don't need `aria-disabled` here, it would only be good for the case where we don't have `disabled` and also when we want the component to have a certain interaction and be announced by SR .

ilzamcmed: Thanks @matuzalemsteles!
Removed.

PR #5827, opened 2024-05-27
refactor(@clayui/popover): LPD-15613 Popover: closeOnClickOutside prop  does not work with keyboard interaction

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-15613

### DIFF
--- a/packages/clay-popover/src/index.tsx
+++ b/packages/clay-popover/src/index.tsx
@@ -247,10 +247,18 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 				}
 			};
 
+			const onBlur = (event: FocusEvent) => {
+				if (event.view === undefined) {
+					setShow(false);
+				}
+			};
+
 			window.addEventListener('keydown', handleKeyDown);
+			window.addEventListener('blur', onBlur);
 
 			return () => {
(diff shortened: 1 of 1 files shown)


packages/clay-popover/src/index.tsx
```diff
@@ -247,10 +247,18 @@ const ClayPopover = React.forwardRef<HTMLDivElement, IProps>(
 				}
 			};
 
+			const onBlur = (event: FocusEvent) => {
+				if (event.view === undefined) {
+					setShow(false);
+				}
```
matuzalemsteles, line 253: It doesn't seem to work well in the case when you are browsing via keyboard and the focus moves to the popover and then when you press the tab when leaving the popover it still remains open, it only closes after the focus leaves the viewport. I think the best thing would be to have a reference to the popover and check if the `event.target` is inside the popover so we don't close the popover when there are other focusable elements, we do this in drop down and modal.

ilzamcmed: Thanks @matuzalemsteles, I'll work on those improvements.

ilzamcmed:
resent to https://github.com/liferay/clay/pull/5831

PR #5828, opened 2024-05-31
feat(@clayui/card): LPD-1261 Add name to group all radio type inputs

veroglez:
Hi guys!,

I am sending this pull because while fixing this bug https://liferay.atlassian.net/browse/LPD-1261, I realized the following:

https://github.com/liferay/clay/assets/9701095/51e8abd8-830f-4812-b611-c29b2eecbefe

As you see in the video, when you navigate by keyboard, the navigation does not recognize which radio button is selected, and when you return to the cards, the radio input that is not selected is focused.

This happens because to group the radio buttons in a group it is necessary to add the same `name` to all the inputs [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role#:~:text=When%20using%20HTML%27s%20native%20input%20radio%20button%2C%20%3Cinput%20type%3D%22radio%22%3E%2C%20the%20radio%20buttons%20are%20grouped%20when%20each%20of%20input%20radio%20buttons%20in%20the%20group%20is%20given%20the%20same%20name).

For that reason, for the radio buttons to work correctly I have added the `name` as required in `radioProps` (apart from `value`), this being the correct behavior:

https://github.com/liferay/clay/assets/9701095/4712d646-2b32-422f-8d45-624ffd5991b2

Let me know what you think, thanks in advance! :smiley:

### DIFF
--- a/packages/clay-card/src/CardWithHorizontal.tsx
+++ b/packages/clay-card/src/CardWithHorizontal.tsx
@@ -46,7 +46,10 @@ export interface IProps extends React.BaseHTMLAttributes<HTMLDivElement> {
 	 * Props to add to the radio element
 	 */
 
-	radioProps?: React.HTMLAttributes<HTMLInputElement> & {value: string};
+	radioProps?: React.HTMLAttributes<HTMLInputElement> & {
+		name: string;
+		value: string;
+	};
 
 	/**
 	 * Flag to indicate if card is selected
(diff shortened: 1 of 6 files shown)


matuzalemsteles (approved):
LGTM! Thanks for working on this @veroglez.

PR #5829, opened 2024-06-05
fix(@clayui/css) LPD-27401 - Allow long word links to break into multiple lines.

juanjofgliferay:
## Jira ticket
https://liferay.atlassian.net/browse/LPD-27401

## Issue
Table links with long words cannot break into multiple lines. The problem is the combination of the `display: inline-flex` property of the link elements with long words.

## UI before
word-wrap-before

The second and fifth link samples should behave in the same way as the normal text.

## UI after
word-wrap-after


pat270:
Closing this in favor of https://github.com/liferay/clay/pull/5830

PR #5835, opened 2024-06-20
fix(@clayui/picker): LPD-28522 Fix language selector not displaying correctly on web content

ilzamcmed:
Ticket: https://liferay.atlassian.net/browse/LPD-28522

### Motivation
The language selector is not displaying correctly: the labels of the status are overlapping the language key and the number of fields is not appearing.
image

### Proposed Solution
Be able to set an auto width so that the menu can grow with the content.

### Screenshots
image

### Commits
- **fix(@clayui/picker): LPD-28522 Add option to add auto width**
- **fix(@clayui/picker): LPD-28522 Update picker stories**

### DIFF
--- a/packages/clay-core/src/picker/Picker.tsx
+++ b/packages/clay-core/src/picker/Picker.tsx
@@ -382,16 +382,7 @@ export function Picker<T extends Record<string, any> | string | number>({
 		);
 	}
 
-	const clientWidth = triggerRef.current?.clientWidth;
-	let menuMinWidth = '160px';
-	let menuWidth = 'auto';
-
-	if (typeof width === 'number') {
-		menuMinWidth = `${clientWidth || 0}px`;
-		menuWidth = `${width}px`;
-	} else if ((clientWidth || 0) > 160) {
-		menuMinWidth = `${clientWidth}px`;
-	}
+	const clientWidth = triggerRef.current?.clientWidth || 0;
 
 	return (
 		<>
@@ -548,8 +539,13 @@ export function Picker<T extends Record<string, any> | string | number>({
 						role="presentation"
 						style={{
 							maxWidth: 'none',
-							minWidth: menuMinWidth,
-							width: menuWidth,
+							minWidth: !width
+								? `${Math.max(160, clientWidth)}px`
+								: undefined,
+							width:
+								typeof width === 'number'
+									? `${Math.max(width, clientWidth)}px`
+									: undefined,
 						}}
 					>
 						{UNSAFE_behavior === 'secondary' &&
--- a/packages/clay-core/src/picker/Picker.tsx
+++ b/packages/clay-core/src/picker/Picker.tsx
@@ -127,9 +127,9 @@ export type Props<T> = {
 	selectedKey?: React.Key;
 
 	/**
-	 * Sets the fixed width of the panel.
+	 * Sets the width of the panel.
 	 */
-	width?: number;
+	width?: number | string;
 
 	/**
 	 * Sets the className for the React.Portal Menu element.
@@ -382,6 +382,20 @@ export function Picker<T extends Record<string, any> | string | number>({
 		);
 	}
 
+	const getComponentWidth = () => {
+		if (typeof width === 'string' && width === 'auto') {
+			return width;
+		}
+
+		if (typeof width === 'number' && width > 160) {
+			return `${width}px`;
+		}
+
+		const clientWidth = triggerRef?.current?.clientWidth || 0;
+
+		return clientWidth > 160 ? `${clientWidth}px` : '160px';
+	};
+
 	return (
 		<>
 			<LiveAnnouncer ref={announcerAPI} />
@@ -537,14 +551,7 @@ export function Picker<T extends Record<string, any> | string | number>({
 						role="presentation"
 						style={{
 							maxWidth: 'none',
-							width: `${
-								typeof width === 'number'
-									? width
-									: (triggerRef.current?.clientWidth || 0) >
-									  160
-									? triggerRef.current?.clientWidth
-									: 160
-							}px`,
+							width: getComponentWidth(),
 						}}
 					>
 						{UNSAFE_behavior === 'secondary' &&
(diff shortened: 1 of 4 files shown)


ilzamcmed:
Heu @matuzalemsteles @pat270 @ethib137 
This PR is ready for review!
Thanks

packages/clay-core/src/picker/Picker.tsx
```diff
@@ -382,6 +382,16 @@ export function Picker<T extends Record<string, any> | string | number>({
 		);
 	}
 
+	const getComponentWidth = () => {
+		if (typeof width === 'string' || typeof width === 'number') {
+			return width;
+		}
```
matuzalemsteles, line 388: I think we always need to maintain the 160px limit regardless of whether the user passes their own value.

ilzamcmed: I think we have that coverage on line 392
```javascript
return clientWidth > 160 ? `${clientWidth}px` : '160px';
``` 

The bug is that we need to allow the user to set the width and make them be readjusted from their content If we add this limit, wouldn't it be an impediment to that?

matuzalemsteles: This limit is only minimum so if the size is greater than that it is ok.

packages/clay-core/src/picker/Picker.tsx
```diff
@@ -382,6 +382,20 @@ export function Picker<T extends Record<string, any> | string | number>({
 		);
 	}
 
+	const getComponentWidth = () => {
+		if (typeof width === 'string') {
```
matuzalemsteles, line 386: I would say that we could just support the use of `auto`, otherwise this could be an escape hatch to get around the imposed limits.

ilzamcmed: Sure! Thanks @matuzalemsteles

packages/clay-core/src/picker/Picker.tsx
ethib137: The initial requirements for this allowed for a width smaller than 160px such as in the case of the date picker for years. We shouldn't prevent a smaller width for that type of case.

packages/clay-core/src/picker/Picker.tsx
ethib137: I don't think we need to make 'auto' a part of the api. I think this would actually just be a part of the styles we apply. To make this work I think it will need width and minWidth, like this:

```
	let menuMinWidth = '160px';
	let menuWidth = 'auto';

	if (typeof width === 'number') {
		menuMinWidth = `${triggerRef.current?.clientWidth || 0}px`;
		menuWidth = `${width}px`;
	} else if ((triggerRef.current?.clientWidth || 0) > 160) {
		menuMinWidth = `${triggerRef.current?.clientWidth}px`;
	}
```

This will result in the following situations all being possible which align with [the initial requirements](https://liferay.atlassian.net/browse/LPD-4232):

Screenshot 2024-06-21 at 10 48 30 AM
Screenshot 2024-06-21 at 10 51 53 AM
Screenshot 2024-06-21 at 10 52 05 AM
Screenshot 2024-06-21 at 10 52 43 AM
Screenshot 2024-06-21 at 10 53 28 AM

ilzamcmed: Thanks @ethib137!
I've made the changes you requested and tested to cover all cases.

packages/clay-core/src/picker/Picker.tsx
```diff
@@ -382,6 +382,22 @@ export function Picker<T extends Record<string, any> | string | number>({
...
+			menuMinWidth = `${triggerRef.current?.clientWidth || 0}px`;
+			menuWidth = `${width}px`;
+		} else if ((triggerRef.current?.clientWidth || 0) > 160) {
+			menuMinWidth = `${triggerRef.current?.clientWidth}px`;
+		}
+
+		return {menuMinWidth, menuWidth};
+	};
+
+	const {menuMinWidth, menuWidth} = getComponentWidth();
```
ethib137, line 399: We don't need to declare this as a function and call it. We can just set the variables directly in the render function.

ilzamcmed: Oh! I got it! 
I'll send the fix in a minute

packages/clay-core/src/picker/Picker.tsx
```diff
@@ -382,6 +382,16 @@ export function Picker<T extends Record<string, any> | string | number>({
...
 	}
 
+	let menuMinWidth = '160px';
+	let menuWidth = 'auto';
+
+	if (typeof width === 'number') {
+		menuMinWidth = `${triggerRef.current?.clientWidth || 0}px`;
+		menuWidth = `${width}px`;
+	} else if ((triggerRef.current?.clientWidth || 0) > 160) {
+		menuMinWidth = `${triggerRef.current?.clientWidth}px`;
```
ethib137, line 392: I'm guessing this should also be `triggerRef.current?.clientWidth || 0`

ilzamcmed: I don't think we need to add `|| 0` because the condition has already ensured that triggerRef.current?.clientWidth is greater than 160.
What do you think?
```javascript
else if ((triggerRef.current?.clientWidth || 0) > 160)
```

ethib137: That makes sense to me. It's probably worth creating a variable for `triggerRef.current?.clientWidth` rather than checking that 3 different times.

ilzamcmed: Done!

packages/clay-core/src/picker/Picker.tsx
matuzalemsteles: That's true, I ended up forgetting about the DatePicker use case.

matuzalemsteles:
I sent a commit to simplify the implementation a little bit.

ethib137:
Nice work on this @ilzamcmed and @matuzalemsteles . LGTM.

I think we really need some tests for this to ensure the menu continues behaving as expected in all the examples we listed on this PR. I don't mind if we merge this first and create a separate task to add the tests. What do you guys think? Do tests seem helpful for this?

matuzalemsteles:
I think we can leave the examples in the stories with menu open by default to do this as a chromatic test, the unit test is a little bad for reproducing this because normally the sizes are fake and it is not very reliable because this is normally a mock.

PR #5837, opened 2024-06-23
LPD-29322 - Improve L0 accessible colors to be closer to 3:1 ratio

marcoscv-work:
Related to this Lexicon PR:
https://github.com/liferay-design/liferay.design/pull/1228/

image

### DIFF
--- a/clayui.com/content/docs/css/color.mdx
+++ b/clayui.com/content/docs/css/color.mdx
@@ -188,9 +188,9 @@ import ClayUISwatch from '../../../src/components/ClayUISwatch';
 		varName="$primary"
 	/>
 	<ClayUISwatch
-		hex="#528eff"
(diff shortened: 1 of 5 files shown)


pat270 (changes requested):
@marcoscv-work we need to update the colors in Cadmin also

marcoscv-work:
> @marcoscv-work we need to update the colors in Cadmin also

thanks! I will add those variables!

marcoscv-work:
@pat270 I've already added variables! thanks!

pat270:
@marcoscv-work I decided it's better to use Sass color functions to derive the `L0` colors just in case someone is relying on it. I think we can change the colors to hex in another ticket.

pat270 (approved):

PR #5841, opened 2024-07-09
fix(@clayui/modal): move focus to h1

matuzalemsteles:
Ticket [LPD-31019](https://liferay.atlassian.net/browse/LPD-31019)

Well, that's moving the focus to the modal title instead of the dialog. As we have a configuration that allows us to customize the modal header and it may be that the developer does not configure it correctly or does not use `Modal.Title` we still focus on the dialog as a fallback.

@pat270 do we have any class that removes the visual state of focus on h1?

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_modals.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_modals.scss
@@ -235,7 +235,13 @@
 // Modal Title
 
 .modal-title {
-	@include clay-map-to-css(map-get($cadmin-modal, modal-title));
+	$_modal-title: setter(map-get($cadmin-modal, modal-title), ());
+
+	@include clay-map-to-css($_modal-title);
+
+	&[tabindex='-1'] {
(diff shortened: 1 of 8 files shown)


marcoscv-work:
We are evaluating whether we can remove it for this case, since a header is not an interactive control like a button, form element or link, and it receives focus via a `tabindex="-1"` (only via JS). So guys, please give me a day to evaluate this and give you a final answer. Thanks @pat270 for stopping this so we can evaluate it better.

marcoscv-work:
Thanks for waiting guys!

After analyzing the current status, and see that currently (previously to this PR) the focus is not visible on modal opening, and since the h1 is not a focusable element (it's going to be only focusable VIA JS). We think that only for this particular case we can hide the focus by CSS. Otherwise it would look very ugly and would not contribute much to non-SR users:

image

pat270:
Thanks @marcoscv-work! I think we can use the class `shadow-none` on `modal-title`.

pat270 (approved):

PR #5851, opened 2024-08-06
fix(@clayui/tree-view): LPD-1277 TreeView with checkbox need a label

ilzamcmed:
### **Jira ticket**
https://liferay.atlassian.net/browse/LPD-1277

### **Commits**
- **fix(@clayui/tree-view): LPD-1277 Add id to correlate label with checkbox**
- **fix(@clayui/tree-view): LPD-1277 Add id prop to correlate label with checkbox**
- **fix(@clayui/tree-view): LPD-1277 Update snapshots**

### DIFF
--- a/packages/clay-core/src/tree-view/TreeViewItem.tsx
+++ b/packages/clay-core/src/tree-view/TreeViewItem.tsx
@@ -780,6 +780,7 @@ export function TreeViewItemStack({
 					// @ts-ignore
 				} else if (child?.type.displayName === 'ClayCheckbox') {
 					content = React.cloneElement(child as React.ReactElement, {
+						'aria-labelledby': labelId,
 						checked: selection.selectedKeys.has(item.key),
 						disabled: loading || disabled,
 						indeterminate: selection.isIndeterminate(item.key),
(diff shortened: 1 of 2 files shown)


ilzamcmed:
Hey @matuzalemsteles  @pat270 
This PR is ready for review!
Thanks!

packages/clay-form/src/Checkbox.tsx
```diff
@@ -81,6 +87,7 @@ const ClayCheckbox = React.forwardRef<HTMLInputElement, IProps>(
 				<label>
 					<input
 						{...otherProps}
+						aria-labelledby={id}
```
matuzalemsteles, line 90: I think the ideal here is to use the `aria-labelledby` property because using the id would be used incorrectly since both have different meanings here. This could also cause a breaking change because the id was passed along with the `otherProps` to the input.

My suggestion is to pass the `aria-labelledby={labelId}` in the TreeViewItem so that there would be no need for changes in the Checkbox.

ilzamcmed: Thanks @matuzalemsteles! I'll proceed with this change.

matuzalemsteles (approved):
LGTM

PR #5852, opened 2024-08-08
feat(@clayui/color-picker)

SelenaAungst:
https://liferay.atlassian.net/browse/LPD-32529
feat(@clayui/color-picker): Remove small color-picker example and add small toggle for all examples

### DIFF
--- a/packages/clay-color-picker/src/index.tsx
+++ b/packages/clay-color-picker/src/index.tsx
@@ -5,7 +5,6 @@
 
 import ClayDropDown from '@clayui/drop-down';
 import {ClayInput} from '@clayui/form';
-import ClayIcon from '@clayui/icon';
 import {
 	FocusScope,
 	InternalDispatch,
@@ -561,15 +560,7 @@ const ClayColorPicker = ({
 							/>
 
 							<ClayInput.GroupInsetItem before tag="label">
-								{isHex ? (
-									'#'
-								) : (
-									<ClayIcon
-										spritemap={spritemap}
-										style={{width: 10}}
-										symbol="color-picker"
-									/>
-								)}
+								{isHex ? '#' : ''}
 							</ClayInput.GroupInsetItem>
 						</ClayInput.GroupItem>
 					)}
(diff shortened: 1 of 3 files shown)


pat270:
LGTM!

SelenaAungst:
Thanks @pat270! I added more changes to this, please let me know if these are good too. 
cc: @marcoscv-work

packages/clay-color-picker/src/index.tsx
```diff
@@ -561,15 +560,7 @@ const ClayColorPicker = ({
...
 							<ClayInput.GroupInsetItem before tag="label">
-								{isHex ? (
-									'#'
-								) : (
-									<ClayIcon
-										spritemap={spritemap}
-										style={{width: 10}}
-										symbol="color-picker"
-									/>
-								)}
```
matuzalemsteles, line 572: I think we can still use this icon, if I'm not mistaken, this was for use cases when the user passes a CSS Variables, we can check if it is a CSS variable and continue rendering the icon.

marcoscv-work: I think at least Page Editor uses this picker and two buttons in a group to represent CSS variables, so not sure if someone is using the icon for that.
image

You guys can still having if you thing is useful for some case, but by default it should not be present

SelenaAungst: @matuzalemsteles @marcoscv-work I see, so should we keep it the way it is? Or change so it is only used to represent CSS variables?

matuzalemsteles: Ok, I think the Page Editor case is a use case for ColorPicker so we can take into account that it is not used, if I'm not mistaken this icon is not planned by Lexicon so I believe we can just remove it.

SelenaAungst: Sounds good, thanks! In that case, is everything else with this PR good to go?

matuzalemsteles (approved):

PR #5855, opened 2024-08-09
feat(@clayui/localized-input): LPD-33580 Add menu element attributes to the clay localized Input API

igor-franca:
Hey team, my goal here is to be able to use menuElementAttrs thought the Clay Input Localized.

Related [issue](https://liferay.atlassian.net/browse/LPD-18759)

My ultimate goal is to implement this new property in Frontend-infra's InputLocalized component to set the z-index to dropdown.

I was trying to implement a test like that:

```
it("renders with menu element attributes", () => {
const {container} = render(
	<ClayLocalizedInput
		id="locale1"
		locales={locales}
                menuElementAttrs={{className: 'custom-class'}}
		onSelectedLocaleChange={() => {}}
		onTranslationsChange={() => {}}
		selectedLocale={locales[0]!}
		spritemap="/path/to/svg"
		translations={{
			'en-US': 'Apple',
			'es-ES': 'Manzana',
		}}
	/>
);

fireEvent.click(
	container.querySelector('.dropdown-toggle') as HTMLButtonElement,
	{}
);

const dropdownMenu = document.querySelector('.dropdown-menu') as HTMLButtonElement;

expect(dropdownMenu).toHaveClass('custom-class');
```

But I wasn't successful, my `dropdownMenu` got a null value and I unfortunately didn't have the necessary time to debug. I hope everything is ok with the contribution.

### DIFF
--- a/packages/clay-localized-input/src/index.tsx
+++ b/packages/clay-localized-input/src/index.tsx
@@ -9,6 +9,7 @@ import ClayForm, {ClayInput} from '@clayui/form';
 import ClayIcon from '@clayui/icon';
 import ClayLabel from '@clayui/label';
 import ClayLayout from '@clayui/layout';
+import {IPortalBaseProps} from '@clayui/shared';
 import React from 'react';
 
 interface IItem {
@@ -40,6 +41,11 @@ interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {
 	 * Label of the input
 	 */
 	label?: React.ReactText;
+
+	/**
+	 * Prop to pass DOM element attributes to DropDown.Menu
+	 */
+	menuElementAttrs?: React.HTMLAttributes<HTMLDivElement> & IPortalBaseProps;
 
 	/**
 	 * Content to be prepended in case you want to localize a URL.
@@ -95,6 +101,7 @@ const ClayLocalizedInput = React.forwardRef<HTMLInputElement, IProps>(
 			id,
 			label = 'Check for translations',
 			locales,
+			menuElementAttrs,
 			onSelectedLocaleChange,
 			onTranslationsChange,
 			placeholder = 'Text to translate...',
@@ -146,6 +153,7 @@ const ClayLocalizedInput = React.forwardRef<HTMLInputElement, IProps>(
 					<ClayInput.GroupItem shrink>
 						<ClayDropDown
 							active={active}
+							menuElementAttrs={menuElementAttrs}
 							onActiveChange={setActive}
 							trigger={
 								<ClayButton


ethib137:
Hey @igor-franca , is this also for the zindex issue with the dropdown menu and the topbar. If so, that shouldn't be considered a bug, so this change shouldn't be necessary.

igor-franca:
Hi @ethib137, there are other times when setting the z-index will be useful. For example in this ticket.

Here we need to increase the z-index of the sidebars so that they are above the diagram and its items. In this case, it would also be useful to have greater control over the z-index of the dropdown.

Let me know what do you think 😄

ethib137:
Well, in general we are careful to define zindex's globally in Liferay. We have defined different zindex's for different types of components so that all the layers are correct. Most of the time we don't want to change that on the fly. @pat270 can probably say more to this.

igor-franca:
> Hi @ethib137, there are other times when setting the z-index will be useful. For example in this ticket.
> 
> Here we need to increase the z-index of the sidebars so that they are above the diagram and its items. In this case, it would also be useful to have greater control over the z-index of the dropdown.
> 
> Let me know what do you think 😄

I forgot to link the mentioned [ticket](https://liferay.atlassian.net/browse/LPD-21569).

This is another one.

igor-franca:
For this second ticket, another approach can be decrease the sidebar `z-index`, .`c-slideout`. But this behaviour will be happen:

https://github.com/user-attachments/assets/35272ca6-4502-4181-917d-eabd4f1f934c

This is expected? Exist one rule that I need to follow to undestand what is expected?

igor-franca:
> Well, in general we are careful to define zindex's globally in Liferay. We have defined different zindex's for different types of components so that all the layers are correct. Most of the time we don't want to change that on the fly. @pat270 can probably say more to this.

I understand your point, for this second ticket, another approach can be decrease the sidebar `z-index`, .`c-slideout`. But this behaviour will be happen:

https://github.com/user-attachments/assets/35272ca6-4502-4181-917d-eabd4f1f934c

This is expected? Exist one rule that I need to follow to undestand what is expected?

igor-franca:
I'm facing an issue when using vertical bar in conjunction with localized input.

VerticalBar default z-index is 1035
Default z-index of the InputLocalized entry is 1000. So I have this behavior:

screenshot-1723239532

igor-franca:
If I go the route of decreasing the sidebar z-index, we see this result:

https://github.com/user-attachments/assets/bd3283e9-db52-46dd-90f4-c89489ea4cb3

ethib137:
Thanks for the extra details @igor-franca . This looks like it might just be a problem with our z-index's globally. We might have missed a use case when defining them. The best case is that we can fix this globally in clay. There is the chance that we can't in which case, then we would want to take the approach you are suggesting to make the z-index configurable. 

@pat270 can you take a look at this and see if we can fix the issue with the dropdown menu rendering behind the sidebar? Can this be fixed globally or does it require the dropdown to be configurable?

You can find this in the product by creating an object, then navigating to `Control Panel > Objects` then clicking the "View in Model Builder" button at the top of the page.

igor-franca:
Hello, @ethib137. I would like to add just one more detail here.

The `menuElementAttrs` property is already exposed in `ClayDropDownWithItems` and we are using the Model Builder sidebars to show the Dropdown items, see:

igor-franca:
Hey @ethib137, any news on this topic?

If this topic is complex and needs to go on for a long time, I would like to know, to be able to align expectations with my team in relation to the fix of the highlighted behaviors.

pat270:
@ethib137 I'll take a look now

ethib137:
@pat270 is there ever an instance where the z-index for a menu should be lower than the z-index for a sidebar? I would have thought menu should always be above the sidebar?

pat270:
@ethib137 Most of our sidebars are relative, the `z-index` only applies to absolute / fixed positioned sidebars. The high `z-index` is there so it slides out above fixed / sticky navs for sidebars that are full height of the browser window.

igor-franca:
Hey guys, thanks for the information. I would like to inform you that [another approach](https://github.com/liferay-objects/liferay-portal/pull/3591) is followed that gave me a satisfactory result. I believe that the behavior exemplified in this [video](https://github.com/liferay/clay/pull/5855#issuecomment-2278808623) can be seen as expected.

I chose to lower the `z-index` of the sidebar in the context of the Model Builder, so I could remove all other `z-index` customizations in the other components.

So, as mentioned that the dropdown above the topbar is not a bug, and in the context of the model builder I was able to decrease the z-index of the verticalBar. 

The changes made to this PR are no longer necessary for the context of the tickets mentioned.

However, there may be other scenarios where this discussion could be relevant.

> If I go the route of decreasing the sidebar z-index, we see this result:
> 
>  simplescreenrecorder-2024-08-09_18.56.39.mp4

pat270:
@igor-franca I'll send a pr with a `z-index` update. We probably need to add a class or attribute in the vertical bar also.

pat270:
Closing this in favor of https://github.com/liferay/clay/pull/5859

PR #5856, opened 2024-08-12
LPD-33629 $modal-palette map should output pseudo classes for .close

pat270:
- Added pseudo class and element keywords to filter for in the mixin `clay-map-to-css`

https://liferay.atlassian.net/browse/LPD-33629

Added a work around to make keys like hover, focus, disabled work when recursively generating selectors, but we should start encouraging using key names wrapped in quotes like '&:hover'.


ethib137:
Thanks @pat270 ! In order to encourage this usage and in general to encourage this way of extending Clay, could you create a ticket for updating our documentation related to this?

matuzalemsteles (approved):
LGTM

matuzalemsteles:
I'll merge this to add to today's release.

PR #5859, opened 2024-08-14
fix(@clayui/css): LPD-34095 Reduce the default c-slideout z-index so …

pat270:
…dropdown-menu's aren't hidden

BREAKING CHANGE: c-slideout is no longer above fixed navigation menus. If you need the old z-index please add the class c-slideout-height-full to c-slideout

This shouldn't affect anything in vanilla DXP since our sidebars don't cover our fixed navs.

/cc @igor-franca @ethib137


matuzalemsteles (approved):
LGTM

PR #5862, opened 2024-08-19
docs(clayui.com): LPD-34243 Fixes Focus Trap page syntax error

pat270:
https://liferay.atlassian.net/browse/LPD-34243

### DIFF
--- a/packages/clay-core/docs/focus-trap.js
+++ b/packages/clay-core/docs/focus-trap.js
@@ -11,12 +11,16 @@ const exampleImports = `import {Button, FocusTrap} from '@clayui/core';
 import React, {useState, useRef} from 'react';`;
 
 const exampleCode = `const FocusTrapTrigger = () => {
-    const [active, setActive] = useState(false);
+	const [active, setActive] = useState(false);
 	const activateButtonRef = useRef(null);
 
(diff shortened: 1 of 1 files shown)


packages/clay-core/docs/focus-trap.js
```diff
@@ -11,12 +11,16 @@ const exampleImports = `import {Button, FocusTrap} from '@clayui/core';
...
 	const activateButtonRef = useRef(null);
 
 	useEffect(() => {
 		if (active) {
-			return () => activateButtonRef.current?.focus();
+			return () => {
+				if (activateButtonRef.current) {
+					activateButtonRef.current.focus()
+				}
+			};
```
matuzalemsteles, line 23: Hmm that's strange the previous implementation is also correct.

pat270: @matuzalemsteles idk node? was having trouble with the optional chaining operator. Maybe we are not running these files through babel?

matuzalemsteles: Probably yes... strange that it didn't break before. Well but we can go with this to avoid headache now.

matuzalemsteles (approved):

PR #5863, opened 2024-08-20
fix(@clayui/form): LPD-34319 adds disabled to Dual Listbox

pat270:
https://liferay.atlassian.net/browse/LPD-34319


matuzalemsteles (approved):
LGTM

PR #5864, opened 2024-08-23
fix(@clayui/css): LPD-2001 c-prefers-expanded-text shouldn't use max-…

pat270:
…width

https://liferay.atlassian.net/browse/LPD-2001

https://github.com/liferay-platform-experience/liferay-portal/pull/689

This will only fix the stacked tab item text. We will add the functionality mentioned in https://github.com/liferay/clay/pull/5622 in another pr. I wanted a CSS only solution, but it looks like we will need to implement in it in ClayTabs since we can't tell if tabs will break to new line. The tabs will look like the screenshot when expanded text and increased spacing.

tabs-will-look

### DIFF
--- a/packages/clay-css/src/scss/components/_utilities-functional-important.scss
+++ b/packages/clay-css/src/scss/components/_utilities-functional-important.scss
@@ -834,7 +834,6 @@
 // C Prefers Expanded Text
 
 %c-prefers-expanded-text {
-	max-width: 100% !important;
 	overflow-wrap: break-word !important;
 	white-space: normal !important;
 	word-wrap: break-word !important;


matuzalemsteles (approved):
LGTM! It makes sense to me, I think we can go through this quickly and work on it more carefully. If I'm not mistaken we have something similar to do in tabs but the solution is to consider wrapping the rest of the items that are overflowing and putting them inside a drop down. Maybe we can consider this too, but we will still need JS.

pat270:
@matuzalemsteles I agree, lmk if you want me to create another ticket to address the stacked tabs or dropdown issue.

PR #5869, opened 2024-09-12
fix(@clayui/css): LPD-36088 Table Striped hover states should work

pat270:
https://liferay.atlassian.net/browse/LPD-36088


matuzalemsteles (approved):
LGTM

PR #5871, opened 2024-09-12
feat: Allow props prefixed by "data-" to be rendered as attributes in picker's option elements

cgmonte:
### Jira tasks

Main task: https://liferay.atlassian.net/browse/LPD-32878
Sub-task: https://liferay.atlassian.net/browse/LPD-35826

### Introduction

The main Jira ticket above is a Liferay's DDM Form task to add `data-option-reference` attributes to option elements in Form's DOM.

It worked without having to modify Clay in all of the following Form field types: [Grid](https://github.com/liferay/liferay-portal/blob/master/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/Grid/Grid.es.js#L53-L54), [Multiple Select](https://github.com/liferay/liferay-portal/blob/master/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/CheckboxMultiple/CheckboxMultiple.es.js#L39), [Single Select](https://github.com/liferay/liferay-portal/blob/master/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/Radio/Radio.es.js#L87) and Multiple Select from List, which is not yet merged.

But the Single Select from List field type did not work because, currently, Option's props are [specifically destructured](https://github.com/liferay/clay/blob/0f04be3078a71123882cbd4eb0f22b536134dbc4/packages/clay-core/src/picker/Option.tsx#L68-L76), without assigning remaining props to something like an `otherProps` object.

### Goal

This PR seeks to allow props prefixed by "data-" to be rendered as attributes in picker's option elements.

### Issues

While developing, after allowing other props to be passed down to option elements, not only the `data-option-reference` attribute was rendered, but also a `index` appeared in the option. This happens because `useCollection()` [adds](https://github.com/liferay/clay/blob/0f04be3078a71123882cbd4eb0f22b536134dbc4/packages/clay-core/src/collection/useCollection.tsx#L116C7-L116C20) that to `DynamicCollection` items. To prevent that, which did not seem right, a filter was added to otherProps to programmatically remove object keys not starting to "data-". A unity test was added to check for this case as well.

### DIFF
--- a/packages/clay-core/src/picker/Option.tsx
+++ b/packages/clay-core/src/picker/Option.tsx
@@ -74,6 +74,7 @@ export function Option({
 	disabled,
 	keyValue,
 	textValue,
+	...otherProps
 }: Props) {
 	const {
 		activeDescendant,
@@ -95,6 +96,7 @@ export function Option({
 	if (isMobile) {
 		return (
 			<option
+				{...otherProps}
 				aria-describedby={ariaDescribedby}
 				disabled={disabled}
 				value={keyValue}
@@ -107,6 +109,7 @@ export function Option({
 	return (
 		<li role="presentation">
 			<button
+				{...otherProps}
 				{...hoverProps}
 				aria-describedby={ariaDescribedby}
 				aria-label={ariaLabel}
--- a/packages/clay-core/src/picker/Option.tsx
+++ b/packages/clay-core/src/picker/Option.tsx
@@ -58,6 +58,12 @@ type Props = {
 	keyValue?: React.Key;
 
 	/**
+	 * Internal property.
+	 * @ignore
+	 */
+	index?: number;
+
+	/**
 	 * Sets a text value if the component's content is not plain text. This value
 	 * is used in the combobox element to show the selected option.
 	 */
@@ -72,6 +78,7 @@ export function Option({
 	'aria-setsize': ariaSetSize,
 	children,
 	disabled,
+	index: _index,
 	keyValue,
 	textValue,
 	...otherProps
--- a/packages/clay-core/src/picker/__tests__/BasicRendering.tsx
+++ b/packages/clay-core/src/picker/__tests__/BasicRendering.tsx
@@ -5,6 +5,7 @@
 
 import Icon from '@clayui/icon';
 import {cleanup, render} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
 
 import {Option, Picker} from '../../';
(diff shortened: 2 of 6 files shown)


packages/clay-core/src/picker/Option.tsx
```diff
@@ -92,9 +101,14 @@ export function Option({
 
 	const isFocus = isFocusVisible();
 
+	const filteredOtherProps = Object.fromEntries(
+		Object.entries(otherProps).filter(([key]) => key.startsWith('data-'))
+	) as Record<DataAttributeKey, any>;
```
matuzalemsteles, line 106: @cgmonte I'm thinking about this a bit, I think we don't need a runtime check, in theory since we adopted TypeScript we let TypeScript handle the validation mainly if you pass an undefined value the build will fail.

Another thing is that we can probably add more new APIs soon so I think it's safe here not to need this validation for now.

matuzalemsteles (commented):
@cgmonte thanks for the contribution.

packages/clay-core/src/picker/Option.tsx
cgmonte: The thing is that TS is not strict with attributes that aren't valid JavaScript identifiers, but are valid in JSX. So it does not quite check attribute names like `data-*`, `aria-*` or even `anything-*`, so `data-${string}: any` isn't doing anything for us. For example, after declaring like this:
```
type Props = {
	// ...
	[key: `data-${string}`]: any;
};
```
I can still pass something like this without getting TS errors:
```
	<Picker>
		<Option anything-attribute="any value" />
	</Picker>
```
The above information can be found in [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/jsx.html#attribute-type-checking), where it reads like:

> Note: If an attribute name is not a valid JS identifier (like a data-* attribute), it is not considered to be an error if it is not found in the element attributes type.

I have tried other strategies to make this work with TypeScript, but could not get it right. There is probably a way, using helper types or explicitly declaring `ElementAttributesProperty`, but I am not sure if this is the way to go. 

Let me know if validating this is really necessary. I could not find other examples of such validations, using string templates, in Clay's codebase.

cgmonte:
Hey @matuzalemsteles, I have pushed a new version. I have excluded both the type declaration, because of what I described [here](https://github.com/liferay/clay/pull/5871#discussion_r1759410147), and the runtime filter you mentioned [here](https://github.com/liferay/clay/pull/5871#discussion_r1757791568). Problem is: because `useCollection()` [adds](https://github.com/liferay/clay/blob/0f04be3078a71123882cbd4eb0f22b536134dbc4/packages/clay-core/src/collection/useCollection.tsx#L116C7-L116C20) an `index` prop that is passed together with the added `otherProps`, snapshots had to be updated. This will break other unity tests based on snapshots in the Portal as well.

cgmonte:
I don't think this is the final solution, and having that `index` in the DOM doesn't seem right. It wasn't meant to be there. But I am sending this so we can discuss other options.

packages/clay-core/src/picker/Option.tsx
matuzalemsteles: Interesting, I didn't know about this problem... so this really won't help at all, not even for the description of the type for the developer, unfortunately.

matuzalemsteles:
I sent a commit to remove the index from the DOM is safe now. We use it internally to calculate other things and the component needs to ignore them.

Well, regarding JS, I don't consider runtime prop validations in components to be very useful. It won't help much, even when passing the wrong value. That's why we avoid it for components nowadays. It doesn't have a valuable benefit, especially when TS is used more today and we also want people in DXP to use TS more than JS. In this specific case, I still don't think we need validation.

matuzalemsteles (approved):
LGTM

PR #5873, opened 2024-09-21
docs(clayui.com): LPD-36695 DualListBox arrow buttons should be disab…

pat270:
…led when there are no more elements to move

https://liferay.atlassian.net/browse/LPD-36695

duallistbox


matuzalemsteles (approved):
LGTM

PR #5874, opened 2024-09-23
fix(@clayui/form): LPD-36696 DualListBox adds children and don't disable left right buttons if no items are selected

pat270:
https://liferay.atlassian.net/browse/LPD-36696

I added children to the component so we can pass in things like an error message. I also removed disabling the move buttons if length is 0.

duallistbox-new

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -92,7 +92,7 @@ module.exports = {
 			statements: 80,
 		},
 		'./packages/clay-form/src/': {
-			branches: 62,
+			branches: 61,
 			functions: 70,
 			lines: 85,
 			statements: 85,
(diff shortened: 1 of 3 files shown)


matuzalemsteles (commented):
I think that instead of delegating the implementation of the error to the developer we can add this internally in the component, especially if it is a default behavior.

matuzalemsteles:
Just an update on this, I spoke to Patrick and will get to work on it, should have something this week.

matuzalemsteles:
I'm moving the entire validation system to the component as well as the error message with translation support, also adding new APIs to configure the maximum number of items for both sides.

matuzalemsteles:
Hey @pat270 What do you think of these changes? Is this a rule that can be applied to all cases?

pat270:
@matuzalemsteles It's not working correctly for me for some reason. Here is a gif of what I'm seeing.
duallistbox

pat270:
@matuzalemsteles I got it to work. Some of the left/right error stuff just needed to be swapped.

PR #5877, opened 2024-10-14
feat(@clayui/css): LPD-39299 Adds d4-l1 for colors

pat270:
https://liferay.atlassian.net/browse/LPD-39299

This needs to make it into DXP before we can merge https://github.com/liferay-platform-experience/liferay-portal/pull/753

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_globals.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_globals.scss
@@ -77,16 +77,105 @@ $grays: map-merge(
 	$grays
 );
 
+$blue-d4: #004799 !default;
+$blue-d3: #0053b3 !default;
+$blue-d2: #005fcc !default;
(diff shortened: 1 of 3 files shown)


ethib137:
LGTM... just one question... If we are using the same value for atlas and default variables do we have to duplicate it? Instead can we just leave out the variables from atlas? I'm guessing you are adding the variables to atlas because this is an existing pattern, but I just wanted to check. Thanks.

matuzalemsteles (approved):
LGTM!

ethib137:
@matuzalemsteles Let's make sure this is in the early release as well.

pat270:
@ethib137 yeah we can leave it out from atlas. I duplicate it because it's much easier to copy and paste global variables for developers wanting to extend Atlas theme.

ethib137:
No problem, if thats our existing pattern then thats fine. thanks

PR #5879, opened 2024-10-21
chore(clay): LPD-39867 Update readme with troubleshooting Python

pat270:
https://liferay.atlassian.net/browse/LPD-39867


matuzalemsteles (approved):
@pat270 Maybe this is related to the node version too, have you tested with node version v18? But anyway we also need to understand why this fails with node v20, we can investigate this later so it's good to have these instructions now.

PR #5880, opened 2024-10-24
LPD-38518 - ClayPaginationBarWithBasicItems does not render picker with links when deltas use links

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-38518

- **LPD-38518 - Add prop "href" to handle options with links**
- **LPD-38518 - Add case test**
- **LPD-38518 -Handle options with links on pagination bar**
- **LPD-38518 - Add case test for pagination bar**
- **LPD-38518 - Update jest.conig**

### DIFF
--- a/packages/clay-core/src/picker/Option.tsx
+++ b/packages/clay-core/src/picker/Option.tsx
@@ -64,6 +64,11 @@ type Props = {
 	index?: number;
 
 	/**
+	 * Path or URL
+	 */
+	href?: string;
+
+	/**
 	 * Sets a text value if the component's content is not plain text. This value
 	 * is used in the combobox element to show the selected option.
 	 */
@@ -78,6 +83,7 @@ export function Option({
 	'aria-setsize': ariaSetSize,
 	children,
 	disabled,
+	href,
 	index: _index,
 	keyValue,
 	textValue,
@@ -113,11 +119,14 @@ export function Option({
 		);
 	}
 
+	const As = href ? 'a' : 'button';
+
 	return (
 		<li role="presentation">
-			<button
+			<As
 				{...otherProps}
 				{...hoverProps}
+				{...(href ? {href} : {})}
 				aria-describedby={ariaDescribedby}
 				aria-label={ariaLabel}
 				aria-labelledby={ariaLabelledby}
@@ -142,7 +151,7 @@ export function Option({
 				)}
 
 				{children}
-			</button>
+			</As>
 		</li>
 	);
 }
--- a/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
+++ b/packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
@@ -246,7 +246,7 @@ export const ClayPaginationBarWithBasicItems = ({
 						spritemap={spritemap}
 					>
 						{(item) => (
-							<Option key={item.label}>
+							<Option href={item.href} key={item.label}>
 								{sub(labels.selectPerPageItems, [
 									String(item.label),
 								])}
--- a/jest.config.js
+++ b/jest.config.js
@@ -57,7 +57,7 @@ module.exports = {
 		},
 		'./packages/clay-core/src/picker/': {
 			branches: 59,
-			functions: 66,
+			functions: 65,
 			lines: 76,
 			statements: 76,
 		},
@@ -170,7 +170,7 @@ module.exports = {
 			statements: 98,
 		},
 		'./packages/clay-pagination-bar/src/': {
-			branches: 95,
+			branches: 92,
 			functions: 88,
 			lines: 94,
 			statements: 95,
--- a/packages/clay-core/src/picker/__tests__/BasicRendering.tsx
+++ b/packages/clay-core/src/picker/__tests__/BasicRendering.tsx
@@ -227,4 +227,46 @@ describe('Picker basic rendering', () => {
 			'data-attribute-value'
 		);
 	});
+
+	it('render option a link when item has href', () => {
+		const items = [
+			{
+				href: '#1',
+				label: 1,
+			},
+			{
+				label: 2,
+			},
+			{
+				href: '#3',
+				label: 3,
+			},
+			{
+				label: 4,
+			},
+		];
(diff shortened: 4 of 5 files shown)


ilzamcmed:
Hey guys, this PR is ready for review. Thanks!

@matuzalemsteles  @pat270 @ethib137

packages/clay-core/src/picker/Option.tsx
```diff
@@ -115,34 +121,66 @@ export function Option({
...
+				>
+					{selectedKey === keyValue && (
+						<span className="dropdown-item-indicator-start">
+							<Icon symbol="check-small" />
+						</span>
+					)}
+
+					{children}
+				</button>
+			)}
```
matuzalemsteles, line 183: I think we can simplify this here with the dynamic component to avoid code repetition. For example:

```jsx
const As = href ? 'a' : 'button';

return (
    <As
    	{...otherProps}
    	{...hoverProps}
    	aria-describedby={ariaDescribedby}
    	aria-label={ariaLabel}
    	aria-labelledby={ariaLabelledby}
    	aria-posinset={ariaPosInset}
    	aria-selected={selectedKey === keyValue}
    	aria-setsize={ariaSetSize}
    	className={classNames('dropdown-item', {
    		active: selectedKey === keyValue,
    		focus: activeDescendant === keyValue && isFocus,
    		hover: activeDescendant === keyValue && !isFocus,
    	})}
        href={href.toString()}
    	disabled={disabled}
    	id={String(keyValue)}
    	onClick={() => onSelectionChange(keyValue!)}
    	role="option"
    	tabIndex={-1}
    >
    	{selectedKey === keyValue && (
    		<span className="dropdown-item-indicator-start">
    			<Icon symbol="check-small" />
    		</span>
    	)}
    
    	{children}
    </As>   
);
```

You just have to check if adding the `href` can cause any type problems, if so you can try, `...(href ? {href} : {})`.

packages/clay-core/src/picker/Option.tsx
```diff
@@ -63,6 +63,11 @@ type Props = {
 	 */
 	index?: number;
 
+	/**
+	 * Flag to indicate that the option has href.
+	 */
+	href?: string | boolean | undefined;
```
matuzalemsteles, line 69: I think we don't need to add `boolean` here the same for `undefined` when adding that the property is optional (`href?: string;`) it is already marked as `undefined`. I believe that `boolean` was added here to handle the conditional, so I think it is not necessary because you can already have it with `undefined` or `string` it is enough and also it would not make much sense to have something like `href={true}` which would do nothing and having `href={false}` would be the equivalent of not passing the `href` which is marked as undefined.

matuzalemsteles (commented):
I added some comments that may simplify the code.

packages/clay-core/src/picker/Option.tsx
```diff
@@ -63,6 +63,11 @@ type Props = {
 	 */
 	index?: number;
 
+	/**
+	 * Flag to indicate that the option has href.
```
ethib137, line 67: When we say something is a flag it should be of type boolean. This comment should be identical to our other comments for hrefs in other components: https://github.com/search?q=repo%3Aliferay%2Fclay%20href%3F&type=code

packages/clay-core/src/picker/Option.tsx
```diff
@@ -115,34 +121,66 @@ export function Option({
...
+					{...hoverProps}
+					aria-describedby={ariaDescribedby}
+					aria-label={ariaLabel}
+					aria-labelledby={ariaLabelledby}
+					aria-posinset={ariaPosInset}
+					aria-selected={selectedKey === keyValue}
+					aria-setsize={ariaSetSize}
+					className={classNames('dropdown-item', {
+						active: selectedKey === keyValue,
+						disabled,
```
ethib137, line 136: We shouldn't ever set a disabled class on a link.

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
```diff
@@ -246,7 +246,7 @@ export const ClayPaginationBarWithBasicItems = ({
 						spritemap={spritemap}
 					>
 						{(item) => (
-							<Option key={item.label}>
+							<Option href={item?.href} key={item.label}>
```
ethib137, line 249: Do we need the `?`

Since it's not used for item.label it seem we wouldn't need it for item?.href

ethib137 (changes requested):
Nice work @ilzamcmed ... just a few updates

packages/clay-core/src/picker/Option.tsx
ilzamcmed: Thanks!!

packages/clay-core/src/picker/Option.tsx
ilzamcmed: Thansk @ethib137 !

packages/clay-pagination-bar/src/PaginationBarWithBasicItems.tsx
ilzamcmed: I added it because there might not be a href property on the object.

ethib137: So that notation isn't needed for that case. `item?.href` is designed to handle the situation where `item` might be null or undefined. If `item` is null or undefined, then calling `item.href` will result in an error: "Cannot read properties of undefined|null". It's this situation that `item?.href` is for. This notation says if item is undefined or null, don't thrown an error, just return undefined. You can read more about this [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

ilzamcmed: Ahh, you're right! Thanks, Evan, I'll remove it.

matuzalemsteles (approved):
LGTM

ethib137 (approved):
LGTM

ilzamcmed:
ci:forward

PR #5882, opened 2024-10-29
fix(@clayui/css): LPD-39960 Fix padding between text and icon on table head

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-39960

image

### DIFF
--- a/packages/clay-css/src/scss/variables/_tables.scss
+++ b/packages/clay-css/src/scss/variables/_tables.scss
@@ -324,8 +324,8 @@ $c-table: map-deep-merge(
 		),
 		autofit-col: (
 			justify-content: center,
-			padding-left: nth($table-cell-padding, 2),
-			padding-right: nth($table-cell-padding, 2),
+			padding-left: calc(nth($table-cell-padding, 2) / 2),
+			padding-right: calc(nth($table-cell-padding, 2) / 2),
 			first-child: (
 				padding-left: 0,
 			),
--- a/packages/clay-css/src/scss/variables/_tables.scss
+++ b/packages/clay-css/src/scss/variables/_tables.scss
@@ -246,6 +246,16 @@ $c-table: map-deep-merge(
 					th: (
 						href: $table-head-link,
 					),
+					autofit-col: (
+						padding-left: 0.125rem,
+						padding-right: 0.125rem,
+						first-child: (
+							padding-left: 0,
+						),
+						last-child: (
+							padding-right: 0,
+						),
+					),
 				)
 			),
 		table-column-start: (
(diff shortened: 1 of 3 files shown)


packages/clay-css/src/scss/variables/_tables.scss
```diff
@@ -325,7 +325,7 @@ $c-table: map-deep-merge(
 		autofit-col: (
 			justify-content: center,
 			padding-left: nth($table-cell-padding, 2),
```
pat270, line 327: We should change this one to 0.125rem as well.

pat270 (changes requested):
@ilzamcmed We also need to update it in cadmin.

packages/clay-css/src/scss/variables/_tables.scss
ilzamcmed: Hey @pat270 , I was discussing it with @ethib137 and found best to add `0.5rem` on each padding.

ilzamcmed: I will add this to cadmin. Thanks!

packages/clay-css/src/scss/variables/_tables.scss
```diff
@@ -324,8 +324,8 @@ $c-table: map-deep-merge(
 		),
 		autofit-col: (
 			justify-content: center,
-			padding-left: nth($table-cell-padding, 2),
-			padding-right: nth($table-cell-padding, 2),
+			padding-left: nth($table-cell-padding, 1),
```
ilzamcmed, line 327: @ethib137 , this will add `0.5 rem`, so I removed the `calc()` function

pat270: @ilzamcmed this is referencing the top and bottom padding of a table cell. I don't think we should do this.

pat270: I also found use cases in Portal that are using utilities to put 0.25rem padding left and right on autofit-col's in tables. @ethib137 would 0.25rem be acceptable? I think 4px/8px spacing is what Lexicon generally uses

ilzamcmed: Thank you, @pat270! Evan clarified the function of `nth`, as I had misunderstood it. I will wait for Evan's feedback regarding your suggestion and will send the necessary fix afterward.

ilzamcmed: Hey @marcoscv, what are your thoughts on Patrick's comment above?

marcoscv-work: Hey guys, sorry for the delay, marcoscv is my personal user and I don’t get notifications.
From my side any way to fix it is ok, you guys are the experts! 🎉

pat270 (approved):

ilzamcmed:
Thanks @pat270 ! :bowing_woman:

PR #5885, opened 2024-10-31
feat: adding react compiler plugin

matuzalemsteles:
Well, running lint didn't show any errors, but let's see how this will work. I also think I'll add a new commit here updating to the new version of React 18.x. As this doesn't affect whoever uses it, we can already see how Clay works in a new version of React, but we don't need to use the new React features yet. Maybe we can add compatibility, for example, using `useId` if it's in the new version would be an option.


matuzalemsteles:
Well, I think I'll add the update commit to React v18 in another PR because I'll need to update all the tests that are still using `react-test-renderer` to jest.

PR #5886, opened 2024-10-31
feat(@clayui/core): add new API to set item id key for table

matuzalemsteles:
Ticket [LPD-40850](https://liferay.atlassian.net/browse/LPD-40850)

We are adding a new API that can configure which item object key can be used as a unique identifier when the structure does not have `id` as default. This is useful for cases where you are rendering a tree table with expandable items. I'm also updating the documentation to mention this new property in which cases to use it.


ethib137 (approved):
Awesome! Looks great @matuzalemsteles ! Thanks

PR #5888, opened 2024-11-01
fix(@clayui/vertical-nav): LPD-39579 Vertical Navigation Items height should be 40px by default, not 41px

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-39579

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_menubar.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_menubar.scss
@@ -6,6 +6,7 @@ $menubar-vertical-transparent-md: map-deep-merge(
 		nav-link: (
 			border-radius: $border-radius,
 			letter-spacing: 0.016rem,
+			line-height: 1.25rem,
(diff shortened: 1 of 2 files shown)


packages/clay-css/src/scss/atlas/variables/_navs.scss
```diff
@@ -4,6 +4,7 @@ $nav-font-size: 0.875rem !default;
 
 $nav-link-disabled-color: $gray-500 !default;
 
+$nav-link-line-height: 1.25rem !default; //20px
```
pat270, line 7: Instead of adding a new line height variable we can place it directly in the menubar map.

pat270 (changes requested):

packages/clay-css/src/scss/atlas/variables/_navs.scss
ilzamcmed: Sure! Thanks @pat270

PR #5892, opened 2024-11-06
feat(@clayui/dropdown): LPD-41310 Update Clay Dropdown Component to Include Caret Icon

ilzamcmed:
Jira issue: https://liferay.atlassian.net/browse/LPD-41310

- **feat(@clayui/dropdown): LPD-41310 Update Clay Dropdown Component to Include Caret Icon**
- **feat(@clayui/dropdown): LPD-41310 Update spanshots**
- **feat(@clayui/dropdown): LPD-41310 Add "showDropdownIcon=false" prop to prevent caret-icon to be displayed**

The idea is to add the caret button icon exclusively to the dropdown component. I didn't include it in the dropdownWithItems component, as most instances already manually add the caret icon, including those in the portal.

### DIFF
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -4,6 +4,7 @@
  */
 
 import {__NOT_PUBLIC_COLLECTION} from '@clayui/core';
+import ClayIcon from '@clayui/icon';
 import {
 	FOCUSABLE_ELEMENTS,
 	InternalDispatch,
@@ -123,6 +124,11 @@ export interface IProps<T>
 	renderMenuOnClick?: boolean;
 
 	/**
+	 * Flag indicating if the caret icon should be displayed on the right side.
+	 */
+	triggerIcon?: string;
+
+	/**
 	 * Element that is used as the trigger which will activate the dropdown on click.
 	 */
 	trigger: React.ReactElement & {
@@ -178,6 +184,7 @@ function ClayDropDown<T>({
 	renderMenuOnClick = false,
 	role = 'menu',
 	trigger,
+	triggerIcon = 'caret-bottom',
 	...otherProps
 }: IProps<T>) {
 	const triggerElementRef = useRef<HTMLButtonElement | null>(null);
@@ -231,6 +238,14 @@ function ClayDropDown<T>({
 				'aria-controls': ariaControls,
 				'aria-expanded': internalActive,
 				'aria-haspopup': 'true',
+				children: (
+					<>
+						{trigger.props.children}{' '}
+						{triggerIcon !== 'undefined' && (
+							<ClayIcon className="ml-1" symbol={triggerIcon} />
+						)}
+					</>
+				),
 				className: classNames(
 					'dropdown-toggle',
 					trigger.props.className
--- a/packages/clay-drop-down/src/DropDown.tsx
+++ b/packages/clay-drop-down/src/DropDown.tsx
@@ -126,7 +126,7 @@ export interface IProps<T>
 	/**
 	 * Flag indicating if the caret icon should be displayed on the right side.
 	 */
-	triggerIcon?: string;
+	triggerIcon?: string | null;
 
 	/**
 	 * Element that is used as the trigger which will activate the dropdown on click.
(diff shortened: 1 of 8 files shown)


ilzamcmed:
Hey @matuzalemsteles @pat270  @ethib137 
This PR is ready for review! 
Please see[ this thread ](https://liferay.slack.com/archives/C04A7NLM7L1/p1730916107086749)for more info about the `DropdownWithItems`
Thanks!

packages/clay-drop-down/src/DropDown.tsx
```diff
@@ -122,6 +123,16 @@ export interface IProps<T>
 	 */
 	renderMenuOnClick?: boolean;
 
+	/**
+	 * Flag indicating if the caret icon should be displayed on the right side.
+	 */
+	showDropdownIcon?: boolean;
```
matuzalemsteles, line 129: I think this could be something like `triggerIcon` and maybe we can use that instead of symbol and the undefined value would be used to not render the trigger icon so we can avoid having to have two APIs for this and it makes it a little more intuitive by the API name.

ilzamcmed: Thanks @matuzalemsteles!
I had to make a few changes because of the `DropdownWithItems`.

packages/clay-drop-down/src/DropDown.tsx
```diff
@@ -231,6 +238,14 @@ function ClayDropDown<T>({
...
 				'aria-expanded': internalActive,
 				'aria-haspopup': 'true',
+				children: (
+					<>
+						{trigger.props.children}{' '}
+						{triggerIcon !== 'undefined' && (
+							<ClayIcon className="ml-1" symbol={triggerIcon} />
+						)}
+					</>
+				),
```
matuzalemsteles, line 248: I'm still not sure about this yet, we don't have any certainty that the trigger can be a button, so by doing this we are inferring that it will always be a button, which may not be true because the DropDown can be used for different purposes, such as passing an input or other custom element that could break any application.

I think that to avoid problems like this or reduce them, we can create a conditional before passing our `children` with the icon and check if the `trigger` is a valid component and if it is a ClayButton component, so we can be more certain and reduce possible failures in the DXP.

```js
children: React.isValidElement(trigger)
    // @ts-ignore
    && child?.type.displayName === 'ClayButton' ?
        (
       	    <>
          		{trigger.props.children}{' '}
          		{triggerIcon !== 'undefined' && (
         			<ClayIcon className="ml-1" symbol={triggerIcon} />
          		)}
           	</>
    	),
	: undefined,
```

packages/clay-drop-down/src/DropDown.tsx
```diff
@@ -231,6 +238,14 @@ function ClayDropDown<T>({
 				'aria-controls': ariaControls,
 				'aria-expanded': internalActive,
 				'aria-haspopup': 'true',
+				children: (
+					<>
+						{trigger.props.children}{' '}
+						{triggerIcon !== 'undefined' && (
```
matuzalemsteles, line 244: When I referred to undefined, it would be the type and not the string. Sorry if I wasn't very clear with that.

```suggestion
						{triggerIcon && (
```

packages/clay-drop-down/src/DropDownWithItems.tsx
```diff
@@ -584,6 +589,7 @@ export const ClayDropDownWithItems = ({
 	searchable,
 	searchProps,
 	searchValue = '',
+	triggerIcon = 'undefined',
```
matuzalemsteles, line 592:
```suggestion
	triggerIcon,
```

packages/clay-localized-input/src/index.tsx
```diff
@@ -165,6 +165,7 @@ const ClayLocalizedInput = React.forwardRef<HTMLInputElement, IProps>(
 									</span>
 								</ClayButton>
 							}
+							triggerIcon="undefined"
```
matuzalemsteles, line 168:
```suggestion
							triggerIcon={undefined}
```

matuzalemsteles (commented):
Hey @ilzamcmed, thanks for the work on this again I added some more follow up comments mainly regarding the trigger which I'm still unsure about and worried about the breaking change it might cause.

matuzalemsteles:
@ilzamcmed I sent some changes to fix some of the concerns.

ilzamcmed:
Thanks @matuzalemsteles! 
It was difficult to find a case that fit all the logic we needed.

PR #5894, opened 2024-11-19
fix(@clayui/core): Adapt snapshots

veroglez:
Hi guys!

I am sending you the fix for this issue https://liferay.atlassian.net/browse/LPD-42404

Thanks in advance! :smile:

### DIFF
--- a/packages/clay-navigation-bar/src/Item.tsx
+++ b/packages/clay-navigation-bar/src/Item.tsx
@@ -44,11 +44,7 @@ const ClayNavigationBarIcon = ({
 							'aria-current': active
 								? ariaCurrent ?? undefined
 								: undefined,
-							children: (
-								<span className="navbar-text-truncate">
-									{child.props.children}
-								</span>
-							),
+							children: <span>{child.props.children}</span>,
 							className: classNames(
 								'nav-link',
 								child.props.className?.replace('nav-link', ''),
--- a/packages/clay-navigation-bar/src/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/clay-navigation-bar/src/__tests__/__snapshots__/index.tsx.snap
@@ -19,9 +19,7 @@ exports[`ClayNavigationBar collapses the previously expanded dropdown when trigg
           class="nav-link active"
           href="#1"
         >
-          <span
-            class="navbar-text-truncate"
-          >
+          <span>
             Item 1
           </span>
         </a>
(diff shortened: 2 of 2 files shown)


veroglez:
> Hi @veroglez, I think this could break other use cases in DXP taking into context when we added this it was to fix the Forms use case #4782. I think the best thing here would be to add an API to disable this something like `truncate` would be safer because we've had it for over 2 years so there are probably more use cases expecting this to be truncated. What do you think?

Hi @matuzalemsteles! We discussed this issue on Slack in [this thread](https://liferay.slack.com/archives/C049XG98GQL/p1731952181225079) and we came to the conclusion that the truncated text is an accessibility issue. But I would like to mention @marcoscv  to clarify this.

ethib137:
@marcoscv-work and @veroglez I think it's reasonable to remove the class, but we should still account for [the case that forms raised](https://github.com/liferay/clay/pull/4782). Currently the nav item text will wrap, but only on whitespace. If we are removing text-truncate then it probably makes sense to allow it to break words in order to wrap and not overflow the container and even the screen. What do you all think?

marcoscv-work:
> @marcoscv-work and @veroglez I think it's reasonable to remove the class, but we should still account for [the case that forms raised](https://github.com/liferay/clay/pull/4782). Currently the nav item text will wrap, but only on whitespace. If we are removing text-truncate then it probably makes sense to allow it to break words in order to wrap and not overflow the container and even the screen. What do you all think?

Yes! We probably didn't focus on and review accessibility too much in 2022.
The specific case for Forms is very interesting because they are using the navigation as a Title of the form when is not possible to use the component to navigate, so they shouldn't be using the navigation:

image

They should use a title. @drakonux already proposed how to mix the titles in control menu, but right now, and as an example, they could use any of the existing titles in page details, here is an example of file detail in Documents and Media:
image

As a final conclusion, I find it interesting that we implemented this text truncate for a misuse that Forms makes of a navigation component :-)

marcoscv-work:
Issue created:
https://liferay.atlassian.net/browse/LPD-42665
(Waiting for forms validation)

cc @veroglez @ethib137 @matuzalemsteles

marcoscv-work:
it looks safe to me, let's wait for @matuzalemsteles and @ethib137

matuzalemsteles:
> I think the fix should involve removing a bit more code though. I would expect it to be more of a revert of [this commit](https://github.com/liferay/clay/commit/b8a53a20d81e10fa065997d740fc000f58334a0c) than just removing a class.

Yeah I agree reverting this commit is better in this case.

matuzalemsteles:
I'm going ahead with this actually don't need to revert the commit that added this because we are also adding other props that are important to make using NavigationBar with Link and Button OOTB without needing the dev to add an extra class.

PR #5897, opened 2024-11-27
LPD 41310 Follow up

ilzamcmed:
- **fix(@clayui/dropdown) LPD-41310 remove default value from triggerIcon and update docs**
- **Revert "feat(@clayui/dropdown): LPD-41310 Add "triggerIcon=undefined" prop to prevent caret-icon to be displayed"**

### DIFF
--- a/packages/clay-drop-down/docs/drop-down.mdx
+++ b/packages/clay-drop-down/docs/drop-down.mdx
@@ -253,9 +253,9 @@ Clay provides other ways to use the Clay component, which we call high-level com
 
 ### Icon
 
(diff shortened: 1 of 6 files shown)


matuzalemsteles (approved):
LGTM! Just for the record, we will make this value default again in a future major version. It is very difficult to make an inference of when not to add the icon and this seems to exist in many places in the DXP that should not have the icon so to avoid breaking changes in places we do not know we will gradually adopt the use of caret by manually configuring it in the places where it is necessary.

PR #5898, opened 2024-11-27
feat(@clayui/css): LPD-40160 Adds menubar-primary for CMS Product Menu variant

pat270:
https://liferay.atlassian.net/browse/LPD-40160

vertical-bar-primary

I'm sending this as a draft because we have a problem with the accordion headers. They can only be `button` and in some of the headers we have nested plus buttons. Having a `button` nested in a `button` violates:

> [Phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2), but there must be no [interactive content](https://html.spec.whatwg.org/multipage/dom.html#interactive-content-2) descendant and no descendant with the [tabindex](https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex) attribute specified.

from the w3c spec. I need some help figuring out which component we should place `<div role="button">` to make our markup valid.

### DIFF
--- a/packages/clay-core/src/nav/Link.tsx
+++ b/packages/clay-core/src/nav/Link.tsx
@@ -52,44 +52,6 @@ export const Link = React.forwardRef<
 		},
 		ref
 	) => {
-		if (showIcon) {
-			return (
-				<div
-					{...(otherProps as Omit<
-						IProps,
-						keyof React.AnchorHTMLAttributes<HTMLAnchorElement>
-					> &
-						React.HTMLAttributes<HTMLDivElement>)}
-					className={classNames('nav-link', className, {
-						active,
-						['collapse-icon']: showIcon,
-						collapsed,
-						disabled,
-					})}
-					ref={ref as React.Ref<HTMLDivElement>}
-				>
-					{children}
-					<span className="collapse-icon-closed">
-						<Icon
-							focusable="false"
-							role="presentation"
-							spritemap={spritemap}
-							symbol="angle-right-small"
-						/>
-					</span>
-
-					<span className="collapse-icon-open">
-						<Icon
-							focusable="false"
-							role="presentation"
-							spritemap={spritemap}
-							symbol="angle-down-small"
-						/>
-					</span>
-				</div>
-			);
-		}
-
 		return (
 			<LinkOrButton
 				{...otherProps}
@@ -104,6 +66,28 @@ export const Link = React.forwardRef<
 				ref={ref}
 			>
 				{children}
+
+				{showIcon && (
+					<>
+						<span className="collapse-icon-closed">
+							<Icon
+								focusable="false"
+								role="presentation"
+								spritemap={spritemap}
+								symbol="angle-right-small"
+							/>
+						</span>
+
+						<span className="collapse-icon-open">
+							<Icon
+								focusable="false"
+								role="presentation"
+								spritemap={spritemap}
+								symbol="angle-down-small"
+							/>
+						</span>
+					</>
+				)}
 			</LinkOrButton>
 		);
 	}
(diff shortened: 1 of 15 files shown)


ethib137:
@marcoscv-work any recommendations for this one?

marcoscv-work:
Hey guys @ethib137 @pat270 if we use `role="menuitem"` it should work fine.
An example here: https://codepen.io/marcoscv/full/NWQbJbx
You can change the <a> for a button and validate with Axe or Wave. it should be valid.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
```diff
@@ -242,53 +248,86 @@ function VerticalNav<T extends Record<string, any> | string>({
...
+									expandedKeys,
+									firstKey: collection.getFirstItem().key,
+									open,
+									spritemap,
+									toggle,
+								}}
+							>
+								<Collection<T> collection={collection} />
+							</VerticalNavContext.Provider>
+						</Nav>
```
matuzalemsteles, line 300: You could store this in a constant to avoid repeating the code and reuse it below.

pat270: Thanks @matuzalemsteles . I tried my best to put it in a `const`. I'm having trouble figuring out typescript after moving it.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
```diff
@@ -34,6 +34,16 @@ const Trigger = ({
 	</Button>
 );
 
+const Menubar = ({collection, value}) => {
```
pat270, line 37: @matuzalemsteles `yarn lint:typescript` is complaining here. 

```
packages/clay-core/src/vertical-nav/VerticalNav.tsx:37:19 - error TS7031: Binding element 'collection' implicitly has an 'any' type.

37 const Menubar = ({collection, value}) => {
                     ~~~~~~~~~~

packages/clay-core/src/vertical-nav/VerticalNav.tsx:37:31 - error TS7031: Binding element 'value' implicitly has an 'any' type.

37 const Menubar = ({collection, value}) => {
```
I don't know how to fix it.

packages/clay-core/src/nav/Link.tsx
```diff
@@ -52,6 +52,40 @@ export const Link = React.forwardRef<
 		},
 		ref
 	) => {
+		if (showIcon) {
```
pat270, line 55: This change here will probably break tests in DXP due to `button` being changed to `div`.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
matuzalemsteles: I'll fix this, push a commit to deal with this and the const.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
matuzalemsteles: Hmm let me see, I'll push a commit dealing with that.

ethib137:
Hey @pat270, what's the status on this one?

Also, it looks like keyboard navigation for vertical nav is broken by this PR.

pat270:
@ethib137 I wasn't able to fix the the keyboard navigation. I had to rearrange the markup due to button in button React error. I believe the keyboard navigation isn't working for this because the text is nested inside other elements.

menubar-primary

matuzalemsteles:
I might take a look at it this week if I can't get it moving @pat270.

ethib137:
@matuzalemsteles and @pat270 we need this ASAP. Please prioritize getting this fixed and released. Thanks guys.

pat270:
@ethib137 keyboard navigation works for the other vertical navs. It's only broken for the primary variant.

ethib137:
That's odd... we need it to work for primary too.

matuzalemsteles:
I fixed the navigation bug for primary, it wasn't working because we have a hook that handles navigation for our components but it was only being applied to the old VerticalNav and not to the new variant.

pat270:
Thanks @matuzalemsteles! I'll update the tests

PR #5901, opened 2024-12-12
fix(@clayui/charts): LPD-44040 Update Clay Chart colors for accessibility

pat270:
https://liferay.atlassian.net/browse/LPD-44040


matuzalemsteles (approved):
LGTM

PR #5903, opened 2024-12-17
fix(@clayui/core): fixes bug when updating the state at render time and the number of table columns when the state changes

matuzalemsteles:
From https://github.com/liferay-platform-experience/liferay-portal/pull/865

Well, this PR only makes a small internal adjustment to recalculate the number of columns when the data changes. This is welcome because we can have applications that can change the data while the user interacts, not necessarily linked to column visibility, for example, a table editor.

The bug originated because of column visibility, but I don't believe this is a bug, but because of the way it is done in FDS, which should not change the data that is passed to the Table. It is recommended to always leave the data intact because we deal with this internally.

### DIFF
--- a/packages/clay-core/src/table/Head.tsx
+++ b/packages/clay-core/src/table/Head.tsx
@@ -7,7 +7,7 @@ import Button from '@clayui/button';
 import {ClayToggle as Toggle} from '@clayui/form';
 import Icon from '@clayui/icon';
 import Layout from '@clayui/layout';
-import React, {useMemo} from 'react';
+import React, {useEffect, useLayoutEffect} from 'react';
 
 import {ChildrenFunction, Collection, useCollection} from '../collection';
 import {Item, Menu} from '../drop-down';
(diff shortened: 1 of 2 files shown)


ethib137:
Hey @matuzalemsteles , it looks like I might have been wrong about what was causing the issue. Since I found a way to [reproduce this issue in one of our examples](https://github.com/liferay-platform-experience/liferay-portal/pull/865#issuecomment-2547878063), we can see that this is not fixing the problem. Any idea what is actually causing this problem?

matuzalemsteles:
@ethib137 I was testing always hiding the first column and it seemed to work, I hadn't even thought about the last column. But now I found the problem and it was fixed. There was a problem in always adding the last one to the column visibility calculation, which shouldn't happen when the last cell is not declared, it was kind of the other way around.

ethib137:
Great! Thanks @matuzalemsteles. Let's get this merged and released along with the vertical nav PR Patrick was working on. Our goal should be to release both of these in Portal ASAP.

PR #5905, opened 2024-12-19
feat(@clayui/css): LPD-40080 Add new sticker variants sticker-outline-*

pat270:
https://liferay.atlassian.net/browse/LPD-40080

sticker

### DIFF
--- a/packages/clay-css/src/scss/variables/_stickers.scss
+++ b/packages/clay-css/src/scss/variables/_stickers.scss
@@ -402,6 +402,106 @@ $sticker-dark: map-deep-merge(
 	$sticker-dark
 );
 
+$sticker-outline-0: () !default;
+$sticker-outline-0: map-deep-merge(
+	(
+		background-color: $light,
+		border: 1px solid $dark,
+		color: $dark,
+	),
+	$sticker-outline-0
+);
+
+$sticker-outline-1: () !default;
+$sticker-outline-1: map-deep-merge(
+	(
+		background-color: $purple-l5,
+		border: 1px solid $purple,
+		color: $purple,
+	),
+	$sticker-outline-1
+);
+
+$sticker-outline-2: () !default;
+$sticker-outline-2: map-deep-merge(
+	(
+		background-color: $yellow-l5,
+		border: 1px solid $yellow-d3,
+		color: $yellow-d3,
+	),
+	$sticker-outline-2
+);
+
+$sticker-outline-3: () !default;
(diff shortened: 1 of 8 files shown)


SelenaAungst:
I think this all looks good. 

I noticed on the ticket and in the description that there is no example added for the blue or indigo added colors, does this need to be added too?

ilzamcmed (approved):
LGTM!

packages/clay-sticker/src/index.tsx
```diff
@@ -15,7 +15,17 @@ export type DisplayType =
...
+	| 'space-0'
+	| 'space-1'
+	| 'space-2'
+	| 'space-3'
+	| 'space-4'
+	| 'space-5'
+	| 'space-6'
+	| 'space-7'
+	| 'space-8'
+	| 'space-9';
```
matuzalemsteles, line 28: I'm not sure but this nomenclature seems a bit strange for these variations, I think it would be better to use the color nomenclature here.

pat270:
@SelenaAungst We don't need blue and indigo variants at this time. The base colors were added to keep things consistent.

@matuzalemsteles @marcoscv-work I wish we could give this component a better name other than space.

marcoscv-work:
> @SelenaAungst We don't need blue and indigo variants at this time. The base colors were added to keep things consistent.
> 
> @matuzalemsteles @marcoscv-work I wish we could give this component a better name other than space.

Yes, I agree that the name should not determine its use, in this case they want to use it in CMS Spaces, so if you guys have a better suggestion, for my side there is no problem in giving it a more standard name.
cc @ethib137

pat270:
@ethib137 @marcoscv-work I ended up naming it `sticker-outline-*`. Let me know if you want to rearrange the names.

packages/clay-css/src/scss/variables/_stickers.scss
```diff
@@ -402,6 +402,106 @@ $sticker-dark: map-deep-merge(
 	$sticker-dark
 );
 
+$sticker-outline-primary: () !default;
+$sticker-outline-primary: map-deep-merge(
```
ethib137, line 406: I don't think we should be calling this primary, secondary, or warning, etc... These don't directly use the normal colors we would have for those names. These are just additional accent colors so it's probably simpler if we just use numbers for the variants: ie: sticker-outline-0

What do you think?

matuzalemsteles (approved):
LGTM

PR #5910, opened 2025-01-03
fix(@clayui/date-picker): update date-fns to v4

bryceosterhaus:
We are going to update date-fns to v4 in liferay-portal and realized we should also update it in clay

Let me know if you see any issues with this.


bryceosterhaus:
hmm, there seems to be issues with Jest and ESM, which is the default way date-fns is shipped

matuzalemsteles:
Thanks Bryce!

> hmm, there seems to be issues with Jest and ESM, which is the default way date-fns is shipped.
> Does it make sense to update clay to be pure ESM?

@bryceosterhaus I don't see any problem in doing that.

bryceosterhaus:
@matuzalemsteles I updated this PR to also include a "dual build" for esm and cjs.

I didn't actually test with npm publishing, but just wanted to send it for you to take a look and see.

With this change you can probably remove a whole host of babel dependencies

bryceosterhaus:
Yep! Feel free to put this on hold, I just at least wanted to get something over to you to see and be able to test out. The date-fns update will be merged relatively soon on the portal side, but I believe that should be be fine.

matuzalemsteles:
> FYI. I think we are actually just going to remove date-fns in favor of Intl.DateTimeFormat

Oh that sounds great.

PR #5914, opened 2025-01-15
feat(@clayui/language-picker): LPD-45714 Add security margin

veroglez:
Hi guys!,

I’m sending this pull request where we’re contributing a new component to Clay according to this Epic https://liferay.atlassian.net/browse/LPD-17787. 

It’s a Language Picker and it has two variants:
- A basic language picker.
- A language picker with translations, where each item in the picker is accompanied by a label indicating the translation status.

https://github.com/user-attachments/assets/5f1c5f15-0955-43c2-9681-a3e772b1baf7

As shown in the video, both variants support a small size and the option to hide the trigger text.

A new feature in the translation variant is the addition of a label to indicate the progress of a translation. This can be useful in cases where the language selector considers more than one field, such as in a form. Along with the existing labels (Default, Translated, and Untranslated), this progress label is now included.

In the code I have also added all the tests that I think are necessary, if you consider that I have to add any more, do not hesitate to tell me.

Could you review it? Any feedback is welcome.

Thanks in advance! 😄 

cc @drakonux


veroglez:
@pat270, Could you help me with the styles of the picker options? It seems that the label is misaligned with the rest of the elements (icon, flag, text)

styles

pat270:
@veroglez Sure I'll take a look

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+			hideTriggerText,
+			selectedLocale,
+			small,
+			spritemap,
+			...otherProps
+		}: Record<string, any>,
+		ref
+	) => {
+		return (
+			<ClayButton
```
pat270, line 170: I think we should use a regular `button` tag. `ClayButton` always outputs the `btn` class. If we or someone modifies `.btn` it might cause unintended style changes on this.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+				{...otherProps}
+				aria-label={ariaLabelTrigger}
+				className={classNames(
+					classNamesTrigger,
+					'form-control-select',
```
pat270, line 175: The correct classes to use for this is `form-control form-control-select form-control-secondary`. The styles aren't correct right now. I need to update it via https://liferay.atlassian.net/browse/LPD-46381.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+					{
+						'hidden-label': hideTriggerText,
+					}
+				)}
+				displayType="secondary"
+				ref={ref}
+				size={small ? 'sm' : undefined}
```
pat270, line 182: We should use `form-control-select-sm` instead of `btn-sm`.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+			>
+				<span className="inline-item-before">
+					<ClayIcon
+						spritemap={spritemap}
+						symbol={selectedLocale.symbol}
+					/>
+				</span>
+
+				{!hideTriggerText ? (
+					<span className="font-weight-normal mr-2">
```
pat270, line 192: Without the `btn` class we won't need `font-weight-normal`. Do we need `mr-2` here because the spacing is incorrect between the arrows and text? I can update in CSS if so.

pat270 (changes requested):

packages/clay-language-picker/src/index.tsx
veroglez: Yes, that margin is a safety spacing between the text and the arrows. Updating it in CSS would be ideal.
This change would only be for the case where the text is displayed, since the spacing is correct when only the flag is displayed.

Screenshot 2025-01-17 at 4 02 09 PM

veroglez:
@pat270 I have updated the code with the changes you recommended for the picker button. Thanks a lot! Do you have any recommendations for the picker options?

pat270:
@veroglez I sent a separate pr for the alignment at https://github.com/liferay/clay/pull/5917

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+import React from 'react';
+
+type DisplayType = 'info' | 'secondary' | 'success' | 'warning';
+
+type Item = {
+	displayName?: string;
+	href?: string;
+	id: string;
+	label: string;
+	symbol: string;
```
matuzalemsteles, line 21: Looking at the examples and other cases it seems that we will almost always have the same data for this API, I would say that we can only keep one which would be the `symbol` prop and it could even be `id` as a name. We can use this value as an identifier for both the icons and for the rest of the other APIs, such as selectedLocale... Apparently these other values ​​would be unnecessary for us here.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+import ClayLabel from '@clayui/label';
+import ClayLayout from '@clayui/layout';
+import {InternalDispatch, sub} from '@clayui/shared';
+import classNames from 'classnames';
+import React from 'react';
+
+type DisplayType = 'info' | 'secondary' | 'success' | 'warning';
+
+type Item = {
+	displayName?: string;
```
matuzalemsteles, line 17: Just a suggestion of preference, maybe `name` or `label` for the name of this prop. What do you think?

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+				])}
+			</span>
+
+			<ClayLabel
+				aria-hidden="true"
+				className="c-ml-3"
+				displayType={displayType}
+			>
+				{label}
+			</ClayLabel>
```
matuzalemsteles, line 148: Instead of using a div for the SR I think we could just add an `aria-label` on the ClayLabel with the text instead to avoid adding more elements to the DOM.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+	id,
+	locales,
+	onActiveChange,
+	onSelectedLocaleChange,
+	selectedLocale,
+	small,
+	spritemap,
+	translations = {},
+}: Props) => {
+	const defaultLanguage = locales[0];
```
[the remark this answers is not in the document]
matuzalemsteles: Thinking about it, maybe we will have to have an API to define the default value so as not to infer that the default value is always the first item. If I'm not mistaken, the default can change according to the site's settings on the portal, but I'm not sure if this changes the order of the items.

matuzalemsteles (commented):
Hi Veronica, Thanks for working on this. We are avoiding creating new packages for the new components and instead we are adding the new components under the @clayui/core package, can you move this component to clay-core? Also I added some more comments.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+	/**
+	 * Flag to indicate if the Picker is active or not (controlled).
+	 */
+	active?: boolean;
+
+	/**
+	 * Class names for the trigger.
+	 */
+
+	classNamesTrigger?: string;
```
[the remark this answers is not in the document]
veroglez: yes, in the case of Page Management team, we use this component in a fragment of the Page Editor and we need to pass external classes to adapt the component to the fragment.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+	/**
+	 * Callback that gets called when a selected locale gets changed
+	 */
+	onSelectedLocaleChange: (val: React.Key) => void;
+
+	/**
+	 * Exposes the currently selected locale
+	 */
+	selectedLocale: Item;
```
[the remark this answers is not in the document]
veroglez: I have changed this to take the `id` as you comment and added a `defaultSelectedLocaleId`. In case no `defaultSelectedLocaleId` is passed, it will take the first item from the locals, since there must always be a selected item. On the other hand, I have updated the code a bit so that the uncontrolled way of the component works correctly, it is in this commit https://github.com/liferay/clay/pull/5914/commits/7613c41ed9eb54375a564dd0d607c08f29764334

packages/clay-language-picker/src/index.tsx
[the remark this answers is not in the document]
veroglez: Here I have added a `defaultLocaleId` prop to set a default locale. If the user does not put any value here, the first locale will be set.

packages/clay-language-picker/src/index.tsx
```diff
@@ -0,0 +1,285 @@
...
+import ClayLayout from '@clayui/layout';
+import {InternalDispatch, sub} from '@clayui/shared';
+import classNames from 'classnames';
+import React from 'react';
+
+type DisplayType = 'info' | 'secondary' | 'success' | 'warning';
+
+type Item = {
+	displayName?: string;
+	href?: string;
```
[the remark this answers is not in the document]
veroglez: This is a mistake sorry, I delete it

packages/clay-language-picker/src/index.tsx
veroglez: `symbol` refers to the icon that will be displayed and `label` is the label that we want to be displayed in the picker option.  In my opinion these two props are independent of the `id` prop. I think it is important to maintain and differentiate these keys. Maybe we want to show other icon or other label, and that doesn't have to affect the prop `id`. What do you think?

packages/clay-language-picker/src/index.tsx
veroglez: `name` sounds good

packages/clay-language-picker/src/index.tsx
veroglez: Yep, agree with this. I have updated the `aria-label` of the `Option` component

veroglez:
> Hi @veroglez, Thanks for working on this. We are avoiding creating new packages for the new components and instead we are adding the new components under the @clayui/core package, can you move this component to clay-core? Also I added some more comments.

@matuzalemsteles thanks a lot for the review! I updated the code with your suggestions, let me know what you think

packages/clay-language-picker/src/index.tsx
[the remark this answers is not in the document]
matuzalemsteles: I see, I wanted to avoid this just in case there wasn't a case yet.

packages/clay-language-picker/src/index.tsx
matuzalemsteles: Hmm okay, we can go with that now, I know there are some cases of the same symbols for some flags but if I'm mistaken we repeat the icon but keep the symbol name different.

matuzalemsteles (approved):
@veroglez thanks for working on this, great job. Just a small tweak you can just export the component in the index.ts in the root package file so people can use `import {LanguagePicker} from '@clayui/core';`. Just a little detail in these new components we are removing the Clay naming convention from the component names.

veroglez:
> @veroglez thanks for working on this, great job. Just a small tweak you can just export the component in the index.ts in the root package file so people can use `import {LanguagePicker} from '@clayui/core';`. Just a little detail in these new components we are removing the Clay naming convention from the component names.

Sure! I updated this in https://github.com/liferay/clay/pull/5914/commits/b030db1e788bc943332e3a08b9dac78e3a844501

pat270 (approved):

PR #5916, opened 2025-01-15
fix(@clayui/css): LPD-45799 Cadmin c-prefers-focus should output corr…

pat270:
…ect selector

https://liferay.atlassian.net/browse/LPD-45799

I couldn't find a cleaner way to do this. We can't store the Sass function `clay-insert-before` in a global variable. It gets run immediately and the Sass parent selector, `&`, has no reference unless it's nested inside a ruleset.


SelenaAungst (approved):
I think this all looks good, I don't see any errors in Storybook or in the logs.

ilzamcmed (approved):
LGTM!  :rocket: 
Thanks @pat270

PR #5918, opened 2025-01-17
LPD-46396 - Clay Vertical Navigation should include a prop that allows an aria-label to be added to the nav element.

ilzamcmed:
Ticket: https://liferay.atlassian.net/browse/LPD-46005

## Motivation
The prop aria-label can be passed to the vertical navigation.

## Proposed Solution
Added a prop called `ariaLabel` and an `useEffect` to display the console error when this prop is not passed

## Steps to Verify
1. Go to Clay Vertical Navigation
2. Open Devtools
3. Check if tag `<nav>` has an aria-label property
4. When the propis not passed, there should be a warning logged to the console when in a development build

## Screenshots/videos (if applies)
image
image

### DIFF
--- a/packages/clay-core/src/vertical-nav/VerticalNav.tsx
+++ b/packages/clay-core/src/vertical-nav/VerticalNav.tsx
@@ -11,7 +11,8 @@ import {
 	useNavigation,
 } from '@clayui/shared';
 import classNames from 'classnames';
-import React, {useCallback, useMemo, useRef, useState} from 'react';
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
+import warning from 'warning';
 
 import {Collection, useCollection} from '../collection';
 import {Nav} from '../nav';
@@ -48,6 +49,11 @@ type Props<T extends Record<string, any> | string> = {
 	 * - automatic - moves the focus to the menuitem and activates the menu.
 	 */
 	activation?: 'manual' | 'automatic';
+
+	/**
+	 * Flag to define aria-label.
+	 */
+	'aria-label': string;
 
 	/**
 	 * The component contents.
@@ -140,6 +146,7 @@ function VerticalNav<T extends Record<string, any> | string>(
 function VerticalNav<T extends Record<string, any> | string>({
 	active,
 	activation = 'manual',
+	'aria-label': ariaLabel,
 	children,
 	decorated,
 	displayType,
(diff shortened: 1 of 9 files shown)


ilzamcmed:
Hi @matuzalemsteles @pat270 @SelenaAungst 
This PR is ready for review!
Thanks!

packages/clay-core/src/vertical-nav/VerticalNav.tsx
```diff
@@ -90,6 +95,13 @@ type Props<T extends Record<string, any> | string> = {
 	 */
 	large?: boolean;
 
+	/**
+	 * Messages used for announcement to SR. Use this for internationalization.
+	 */
+	messages?: {
+		error: string;
+	};
+
```
matuzalemsteles, line 104: @ilzamcmed do you think having an API for this would be necessary? I think this is more of a developer use case so we don't need internationalization of the message.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
```diff
@@ -49,6 +49,11 @@ type Props<T extends Record<string, any> | string> = {
 	 */
 	activation?: 'manual' | 'automatic';
 
+	/**
+	 * Flag to define aria-label.
+	 */
+	ariaLabel: string;
```
matuzalemsteles, line 55: If you want to keep the same pattern as React, we can also keep the same `aria-label` naming here. I've been trying to do this to maintain the standard. You can take a look at the picker.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
```diff
@@ -274,9 +290,15 @@ function VerticalNav<T extends Record<string, any> | string>({
 			</VerticalNavContext.Provider>
 		</Nav>
 	);
+	useEffect(() => {
+		if (!ariaLabel) {
+			console.error(messages['error']);
+		}
```
matuzalemsteles, line 296: We have this pattern in other components, you can use the warning package for this.

https://github.com/liferay/clay/blob/master/packages/clay-icon/src/index.tsx#L33-L36

matuzalemsteles (commented):
Thanks @ilzamcmed, I added some comments.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
ilzamcmed: Thanks @matuzalemsteles! You're right, I'll remove that.

packages/clay-core/src/vertical-nav/VerticalNav.tsx
ilzamcmed: Done!

packages/clay-core/src/vertical-nav/VerticalNav.tsx
ilzamcmed: Done!

matuzalemsteles (approved):
LGTM

PR #5919, opened 2025-01-17
fix(@clayui/css): LPD-46381 form-control-select-secondary should only…

pat270:
… be as wide as its contents

https://liferay.atlassian.net/browse/LPD-46381

### DIFF
--- a/packages/clay-css/src/scss/atlas/variables/_forms.scss
+++ b/packages/clay-css/src/scss/atlas/variables/_forms.scss
@@ -408,6 +408,7 @@ $form-control-select-palette: map-deep-merge(
 	(
 		form-control-select-secondary: (
 			background-color: $white,
+			border-color: $secondary-l0,
 			color: $gray-600,
 			hover: (
 				color: $gray-600,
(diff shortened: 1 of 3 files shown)


matuzalemsteles:
Hey @pat270 I think this affects the Picker for width cases, for example see the difference in the master: https://storybook.clayui.com/?path=/story/design-system-components-picker--width and with the changes of this PR: https://deploy-preview-5919--next-storybook-clayui.netlify.app/?path=/story/design-system-components-picker--width.

I think we want to keep it as it is in the master because we can have the size of the picker taking up all the space regardless of the content inside the trigger.

pat270:
@matuzalemsteles sounds good. Can we make the default `min-content` in Picker? I'll create another ticket for the color discrepancies between clay and cadmin.

pat270:
@matuzalemsteles I looked at this more and how about we add a helper class like `form-control-select-shrink`? That way we don't break users who are using the width method.

matuzalemsteles:
> I looked at this more and how about we add a helper class like form-control-select-shrink? That way we don't break users who are using the width method.

@pat270 well it may be but I'm not sure if it will help some people, because the dev can wrap the Picker in a container and control its size but this can be a good utility.

pat270:
@matuzalemsteles lmk if this is good. Pedro wants me to close out some tickets.

matuzalemsteles (approved):
LGTM! Thanks @pat270

PR #5927, opened 2025-01-23
fix(@clayui/css): LPD-47129 Mixins, Improve performance in alerts, ba…

pat270:
…dges, and cards

https://liferay.atlassian.net/browse/LPD-47129

This shaved like 4s off `yarn compile`.

### DIFF
--- a/packages/clay-css/src/scss/mixins/_badges.scss
+++ b/packages/clay-css/src/scss/mixins/_badges.scss
@@ -178,64 +178,106 @@
 		$badge-item-expand: setter(map-get($map, badge-item-expand), ());
 
 		@if ($enabled) {
-			@include clay-css($base);
+			@if (length($base) != 0) {
+				@include clay-css($base);
+			}
 
 			// Empty badges collapse automatically
 
-			&:empty {
-				@include clay-css(map-get($map, empty));
+			$_empty: map-get($map, empty);
+
+			@if ($_empty) {
+				&:empty {
+					@include clay-css(map-get($map, empty));
+				}
 			}
 
-			&[href],
-			&[type] {
-				@include clay-link($href);
+			@if (length($href) != 0) {
+				&[href],
+				&[type] {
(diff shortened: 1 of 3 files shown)


packages/clay-css/src/scss/mixins/_badges.scss
```diff
@@ -178,64 +178,106 @@
...
 			}
 
-			.badge-item-before {
-				@include clay-css(map-get($map, badge-item-before));
+			$_badge-item-before: map-get($map, badge-item-before);
+
+			@if ($_badge-item-before) {
+				.badge-item-before {
+					@include clay-css($_badge-item-before);
+				}
```
matuzalemsteles, line 265: This is very interesting, in JS you can do all this inline, you save memory and could have the help of inline cache, but interestingly here it could be faster. Maybe this is faster because you pass an empty value to `clay-css`?

pat270: You have a point. I'll test in the clay-css mixin and others.

pat270: @matuzalemsteles Sass actually takes a good amount of time processing CSS rule-sets. Processing a simple rule-set:
```
.selector{
    color: red;
}
```
100k times takes 1700ms on my machine. If we wrap it in `@if (false){}` it reduces it to 40ms.

matuzalemsteles: @pat270 That's interesting, the bad thing about it is that we need to replicate this rule everywhere.

pat270: Yeah it's unfortunate, but we only need to apply it to the mixins to get a big boost in build time.

matuzalemsteles (approved):
LGTM

SelenaAungst (approved):
Everything looks good to me!

PR #5932, opened 2025-02-06
fix(@clayui/shared): Only run hide code if menus match in Overlay

pat270:
https://liferay.atlassian.net/browse/LPD-48227

@matuzalemsteles not related to this pr, but should we be using `inert` for all our dropdown menus and popups?


matuzalemsteles (approved):
LGTM! The problem with inert is browser support, so we still maintain support but are enabling it for some components like DatePicker. It has a different behavior that blocks all elements from having focus, this can harm the user experience for cases like DropDown, so we need to evaluate.

PR #5933, opened 2025-02-10
fix(@clayui/shared): LPD-48347 Close overlay on left click only

pat270:
https://liferay.atlassian.net/browse/LPD-48347

dropdown-overlay

### DIFF
--- a/packages/clay-shared/src/Overlay.tsx
+++ b/packages/clay-shared/src/Overlay.tsx
@@ -114,8 +114,11 @@ export function Overlay({
 
 	useInteractOutside({
 		isDisabled: isOpen ? !isCloseOnInteractOutside : true,
-		onInteract: () => {
-			onHide('blur');
+		onInteract: (event) => {
+			// @ts-ignore
+			if (event.button === 0) {
+				onHide('blur');
+			}
 		},
 		onInteractStart: (event) => {
 			if (overlayStack[overlayStack.length - 1] === menuRef) {


SelenaAungst (approved):
LGTM!

matuzalemsteles:
Hey @pat270 I think this would not be ideal to do because it is not common user behavior, especially if we are doing this for debugging. Would that be the intention? The ideal here would be to just add a breakpoint to do the debugging on the js side to avoid the dropdown closing if that were the case. I don't want to add that behavior if that's the case.

pat270:
@matuzalemsteles Not everyone has access to the source code. It's pretty painful having to search through the elements to locate the correct dropdown.

matuzalemsteles:
@pat270 The debugger can be done by the browser itself, there is no need to go into the code to modify it. We can create a guide for this if necessary, but I don't think it's worth having to make this change just to save the developer's experience and harm the user's experience. This is a very common behavior for the user, so I'm not sure about it.

Just to understand what we want to do here, do you want the overlay to be closed only by left-clicking anywhere on the screen, but should it remain open when right-clicking?

pat270:
I mean when you right click on a page, the browser's context menu to appears and a selection shouldn't cause changes on the page. Look at github's dropdown menus for example, it functions correctly imo.

pat270:
@matuzalemsteles This should work now.

matuzalemsteles:
@pat270 I can't see this case being reproduced in the master. Right-clicking still allows me to open the browser menu and not close the dropdown https://storybook.clayui.com/. Is there any step I can reproduce in the master?

pat270:
@matuzalemsteles You have to right click outside the dropdown menu

dd

pat270:
@matuzalemsteles It doesn't break it.

treeview

matuzalemsteles:
@pat270 we need to check the use case in DXP so we will have to monitor this there when the release comes out with this change, if I remember correctly we had some problems in not being able to properly reproduce this bug in the storybook because in DXP there was interference from other events.

ethib137:
Please make sure to test this in DXP before releasing by following this process: https://github.com/liferay/clay/blob/master/CONTRIBUTING.md#testing-local-changes-in-liferay-portal

PR #5939, opened 2025-02-19
docs(@clayui/table): Add code example of table-nested-rows

SelenaAungst:
I am having trouble testing on the new clay site so I was unable to test this, but this is my proposed change for https://liferay.atlassian.net/browse/LPD-47018.

### DIFF
--- a/packages/clay-core/docs/table.mdx
+++ b/packages/clay-core/docs/table.mdx
@@ -238,8 +238,13 @@ It is also possible to implement your own logic on the client side when your dat
 </div>
 
 ```jsx preview
-export function App() {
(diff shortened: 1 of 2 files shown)


matuzalemsteles (commented):
Hi @SelenaAungst, I think we can add examples using the component on the component's main page too.

Maybe we can just replicate the main example from the page that already shows the use of sorting but we can duplicate it with the code https://clayui.com/docs/components/table#sorting.

SelenaAungst:
I am still unable to test this so I am not 100% sure if this will show the preview.

matuzalemsteles (approved):
Thanks @SelenaAungst! I've updated the example to work properly with codesandbox.

PR #5942, opened 2025-02-20
fix(@clayui/core): fixes build bug in applications

matuzalemsteles:
Since we introduced a new API and made it required, this would be a major change because we would have to update all applications that use the component to define the `aria-label`, so the ideal would be to make the API optional. Since we have a warning message, this will notify developers to pay attention to this and configure it according to each use case.


matuzalemsteles:
Merging this to release because it's affecting the master.

PR #5947, opened 2025-02-20
feat(@clayui/management-toolbar): LPD-49362 Add fluidSize prop

pat270:
https://liferay.atlassian.net/browse/LPD-49362

### DIFF
--- a/packages/clay-management-toolbar/src/ManagementToolbar.tsx
+++ b/packages/clay-management-toolbar/src/ManagementToolbar.tsx
@@ -16,12 +16,18 @@ interface IProps extends React.HTMLAttributes<HTMLElement> {
 	 * Flag that identifies when to show the active state of the component.
 	 */
 	active?: boolean;
+
+	/**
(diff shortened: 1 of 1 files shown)


packages/clay-management-toolbar/src/ManagementToolbar.tsx
```diff
@@ -16,12 +16,20 @@ interface IProps extends React.HTMLAttributes<HTMLElement> {
 	 * Flag that identifies when to show the active state of the component.
 	 */
 	active?: boolean;
+
+	/**
+	 * Set a maximum width on container-fluid.
+	 */
+	fluidSize?:
+		| React.ComponentProps<typeof ClayLayout.Container>['fluidSize']
+		| false;
```
matuzalemsteles, line 25: @pat270 I don't think it would be necessary to use `false` here, just inheriting the component type would be enough.

matuzalemsteles (approved):

PR #5951, opened 2025-02-25
feat(@clayui/css): LPD-49901 Icons, adds send.svg and add alias stop to square icon

pat270:
https://liferay.atlassian.net/browse/LPD-49901

@matuzalemsteles I couldn't figure out how to update `clay/www/public/images/icons/icons.svg`. The send icon doesn't show on clayui.com.


matuzalemsteles:
@pat270 we got the icons for codesandbox from the package https://github.com/liferay/clay/blob/master/www/app/_components/Sandpack.server.tsx#L19 maybe we can change to get it directly from the local package build, I think that way it is getting it via npm so the local update will not have any effect.

The same is for the local build https://github.com/liferay/clay/blob/master/www/plugins/scss/index.mjs, maybe we can link the package with the local instance to make the site as part of the workspace, I remember trying to do this but there were many errors due to dependencies colliding with the rest of our packages.

matuzalemsteles (approved):
LGTM

matuzalemsteles:
@pat270 Can you submit this in a separate PR? I'll merge this PR.

PR #5952, opened 2025-02-26
fix: exports the missing types

matuzalemsteles:
Context https://liferay.slack.com/archives/C049XG98GQL/p1740557002709139

With the update of Clay's website we made many changes to the types to make the API Table work properly and some were removed and went unnoticed in the final review, I am adding back here all those that were left out that could break the build of the applications, as soon as I merge this I will make a new release.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -141,7 +141,7 @@ module.exports = {
 		},
 		'./packages/clay-multi-select/src/': {
 			branches: 52,
-			functions: 54,
+			functions: 52,
 			lines: 62,
 			statements: 62,
 		},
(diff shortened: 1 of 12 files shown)


ethib137 (approved):
LGTM... Thanks @matuzalemsteles!

ethib137:
@matuzalemsteles looks like [this error](https://github.com/liferay-platform-experience/liferay-portal/pull/1019#issuecomment-2687059408) is also caused by a change in the types. Dropdown is missing containerProps on the menuElementAttrs.

As Bryce recommended, please run `yarn run node-scripts check:tsc` from `modules` to see what else might need to be fixed. Since this is breaking SF for a lot of modules I'm sending a revert of the previous release.

PR #5954, opened 2025-02-27
fix(@clayui/css): custom-control-outside shouldn't be clickable beyon…

pat270:
…d the text

https://liferay.atlassian.net/browse/LPD-50093

I wanted to remove `custom-control-outside` from our checkboxes, but there might be users targeting it with a CSS selector. What do you think?

### DIFF
--- a/packages/clay-css/src/scss/cadmin/variables/_custom-forms.scss
+++ b/packages/clay-css/src/scss/cadmin/variables/_custom-forms.scss
@@ -538,7 +538,6 @@ $cadmin-custom-control-outside: () !default;
 $cadmin-custom-control-outside: map-deep-merge(
 	(
 		label: (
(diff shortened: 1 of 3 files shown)


SelenaAungst (approved):
I am still having trouble with testing stuff on the new Clay site, but the code looks good to me.

matuzalemsteles:
> I wanted to remove custom-control-outside from our checkboxes, but there might be users targeting it with a CSS selector. What do you think?

I agree, the ideal would be to just remove it in a next major version, you can note this in the package's `BREAKING.md` so we don't forget to remove it later.

matuzalemsteles (approved):

ethib137:
@matuzalemsteles we need this released as soon as possible for an LPP.

PR #5955, opened 2025-02-27
docs(@clayui/charts): adds chart documentation

matuzalemsteles:
Still in WIP, for some reason the chart styles are bugging in codesandbox, I'll work more on this tomorrow and the geomap is also a bit complicated to make work in codesandbox.


matuzalemsteles:
Well this will stay on hold for now. I will start working on a ticket later to refactor our code block with codesandbox to no longer use the iframe but still allow a button that the user can click and load the example in codesandbox.

PR #5963, opened 2025-03-11
 feat(@clayui/core): LPD-47056 Add Keyboard Interactions to the Icon Selector

ilzamcmed:
Ticket: https://liferay.atlassian.net/browse/LPD-47056

## Motivation
Handle keyboard interactions on the dropdown grid

## Proposed Solution
Use the `useNavigation` hook to handle keyboard interactions

## Screenshots/videos 

https://github.com/user-attachments/assets/8edd1570-ad8a-4c0c-a93c-81c7d07e807f

### DIFF
--- a/packages/clay-core/src/icon-selector/IconSelector.tsx
+++ b/packages/clay-core/src/icon-selector/IconSelector.tsx
@@ -50,6 +50,7 @@ export type Props = {
 	messages?: {
 		changeIconButton: string;
 		clearInput?: string;
+		iconSelected: string;
 		notFound: string;
 		placeholder?: string;
 		removeIcon: string;
@@ -71,6 +72,7 @@ export type Props = {
 const defaultMessages = {
 	changeIconButton: 'Change Icon',
 	clearInput: 'Clear search input',
+	iconSelected: 'Icon selected',
 	notFound: 'No items were found.',
 	placeholder: 'Search',
 	removeIcon: 'Remove icon selection',
@@ -393,15 +395,14 @@ export function IconSelector({
 								</ul>
 							)}
 
-							{filteredIcons.length === 0 && (
-								<div
-									aria-live="polite"
-									className="dropdown-caption"
-									id={inputId}
-								>
-									{messages?.notFound}
-								</div>
-							)}
+							<div
+								aria-live="assertive"
+								className="dropdown-caption"
+								id={inputId}
+							>
+								{filteredIcons.length === 0 &&
+									messages?.notFound}
+							</div>
 						</FocusMenu>
 					</div>
 				</Overlay>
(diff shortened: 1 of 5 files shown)


matuzalemsteles:
I made some improvements to the navigation to maintain some behaviors that exist in DropDown and better control the navigation for some rules such as closing the menu with Esc, opening the menu with the up and down arrow keys (the down arrow key should ideally move the focus to the last element of the list but I think we can implement that later.) and space.

I also made some refactorings that should move the focus to the first focusable element of the menu when opening, reusing an existing implementation that we have in some places but I still think that it is possible to improve this implementation so I added a comment for this in the code. I also added a small detail of moving the focus only when the user is interacting with the arrow keys, this does not happen when using the mouse. Moving the focus after choosing the icon was also implemented.

There is a small detail that I still find strange in user interaction via keyboard, as we have a search now, to get to the icons we use Tab but after that we need to use the arrow keys because this is the default behavior of DropDown and if we use Tab again it closes the Menu and moves the focus to the next focusable element. One detail: should we choose the icon in addition to closing it? This happens in the Picker, maybe we can do it here too.

ethib137:
Hey @marcoscv-work can you review the current state of the keyboard interactions and answer Matu's questions in the comment above? Thanks.

marcoscv-work:
> Hey @marcoscv-work can you review the current state of the keyboard interactions and answer Matu's questions in the comment above? Thanks.

Of copurse!
If we have changed the way we manage the focus, probably we will need a complete Screen Reader and combination OS set test. It could take a bit.
@matuzalemsteles do you recommend a specific place to make the testing? Can we test on the PR preview site? for example: https://deploy-preview-5963--next-storybook-clayui.netlify.app/?path=/story/design-system-components-dropdown--default

matuzalemsteles:
@marcoscv-work Yes, we can use this link to do the tests. If you need anything else in the example, we can add more things.

marcoscv-work:
> @marcoscv-work Yes, we can use this link to do the tests. If you need anything else in the example, we can add more things.

Tomorrow I will test all the combinations, in the meantime, it might be a good idea to run the tests on the portal; playwright simulates many clicks and changes in focus states

packages/clay-core/src/icon-selector/IconSelector.tsx
```diff
@@ -423,6 +424,14 @@ export function IconSelector({
 							<ClayInput readOnly value={selectedIcon} />
 						</ClayInput.GroupItem>
 
+						<div
+							aria-live="polite"
```
marcoscv-work, line 428: We can remove this line, role="status" includes "polite"

marcoscv-work:
I've tested **Keyboard** on _Windows/MacOS/Linux_ and **ScreenReaders**: _Windows: JAWS/NVDA (Edge and Chrome) and macOS: VoiceOver (Safari and Chrome)_.

Everything seems pretty similar, but I'm having a few issues with the storybook, but I don't think we're blocking anything that can't be tweaked or improved later, so let's move on!

marcoscv-work (approved):

PR #5969, opened 2025-03-12
fix(www): Fixes and improvements to documentation

matuzalemsteles:
Ticket [LPD-51046](https://liferay.atlassian.net/browse/LPD-51046)

I made some general improvements that are in the ticket and also some things related to the Table API with some spacing errors between properties, I'm also adding the Deprecated flag for properties that weren't being shown.

There are still a few things that need work:

- [ ] Update the breadcrumb markup. We need to rewrite this documentation @pat270 can you take care of that?
- [ ] Update the iframes sizes. Well this is a bit tricky to do because the iframe doesn't adjust according to the internal content, so I'm thinking about how we can do this without having to manually set this for each example.


pat270:
@matuzalemsteles We need to update the breadcrumbs nav to not use utility classes. It makes it difficult to style.

matuzalemsteles:
Well, the resize size will need to be defined, for example, because we can't create something automatic. I'll do this in another PR.

PR #5971, opened 2025-03-14
LPD-50827 New Variant of Clay Modal

BarbaraCabrera:
## Motivation
Due to [this](https://liferay.atlassian.net/browse/LPD-47547) story we are contributing a new Variant of Clay Modal to be used in the CMS and with Marketplace in DXP. 

## Changes
To do so, I have added the “clean” attribute that hides the title and changes the styles to not show the header and keep the close button in place. Also it adds a "sr-only" to the title.

Once we have this, to obtain the modal that we want to use in both cases, we would only have to place the image, the title and the desired text in the body manually.

### DIFF
--- a/packages/clay-css/src/scss/components/_modals.scss
+++ b/packages/clay-css/src/scss/components/_modals.scss
@@ -472,3 +472,10 @@
 	top: -9999px;
 	width: 50px;
 }
+
+.modal-header-clean {
+	position: absolute;
+	right: 0;
+	z-index: 9;
+	border: 0;
+}
--- a/packages/clay-modal/src/Context.ts
+++ b/packages/clay-modal/src/Context.ts
@@ -11,6 +11,12 @@ export interface IContext {
 	ariaLabelledby?: string;
 
 	/**
+	 * Flag indicating if the modal should be the clean variant
+	 * with the title hidden
+	 */
+	clean?: boolean;
+
+	/**
 	 * Callback called to close the modal.
 	 */
 	onClose: () => void;
--- a/packages/clay-modal/stories/Modal.stories.tsx
+++ b/packages/clay-modal/stories/Modal.stories.tsx
@@ -39,7 +39,9 @@ export const Default = (args: any) => {
 					size={args.size}
 					status={args.status}
 				>
-					<ClayModal.Header>{args.title}</ClayModal.Header>
+					<ClayModal.Header clean={args.clean}>
+						{args.title}
+					</ClayModal.Header>
 					<ClayModal.Body
 						iFrameProps={{
 							'aria-label': 'Hello World',
@@ -95,6 +97,7 @@ export const Default = (args: any) => {
 Default.args = {
 	autoClose: false,
 	center: false,
+	clean: false,
 	scrollable: false,
 	size: 'lg',
 	status: undefined,
--- a/packages/clay-modal/src/Header.tsx
+++ b/packages/clay-modal/src/Header.tsx
@@ -22,6 +22,12 @@ export const ItemGroup = ({
 
 export interface IItemProps extends React.HTMLAttributes<HTMLDivElement> {
 	/**
+	 * Flag indicating if the modal should be the clean variant
+	 * with the title hidden
+	 */
+	clean?: boolean;
+
+	/**
 	 * Flag for indicating if item should autofitting the width
 	 */
 	shrink?: boolean;
@@ -120,12 +126,13 @@ const ICON_MAP = {
 
 const HighLevel = ({
 	children,
+	clean,
 	onClose,
 	spritemap,
 	status,
 }: IContext & {children?: React.ReactNode}) => (
 	<>
-		<Title>
+		<Title className={classNames({'sr-only': clean})}>
 			{status && (
 				<TitleIndicator>
 					<ClayIcon spritemap={spritemap} symbol={ICON_MAP[status]} />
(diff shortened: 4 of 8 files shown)


packages/clay-modal/src/Context.ts
```diff
@@ -10,6 +10,12 @@ import {Status} from './types';
 export interface IContext {
 	ariaLabelledby?: string;
 
+	/**
+	 * Flag indicating if the modal should be the clean variant
+	 * with the title hidden
+	 */
+	clean?: boolean;
```
pat270, line 17: I couldn't find an example of a modal with an image in the Figma document. The way it looks is similar to Cards. We can add a status type called image at https://github.com/liferay/clay/blob/master/packages/clay-modal/src/types.ts#L6 instead of `clean` here. It will add the class `modal-image` to `modal-dialog`.

packages/clay-css/src/scss/components/_modals.scss
```diff
@@ -472,3 +472,10 @@
 	top: -9999px;
 	width: 50px;
 }
+
+.modal-header-clean {
```
pat270, line 476: We should scope all styles in `.modal-asset`.

```
.modal-asset {
  .modal-header {
    height: auto;
    ...

    .close {
       position: absolute;
       ...
    }
  }

  .modal-body {
    ...
  }
}
```

packages/clay-modal/stories/Modal.stories.tsx
```diff
@@ -39,7 +39,9 @@ export const Default = (args: any) => {
 					size={args.size}
 					status={args.status}
 				>
-					<ClayModal.Header>{args.title}</ClayModal.Header>
+					<ClayModal.Header clean={args.clean}>
```
pat270, line 42: This demo doesn't make sense. There should be an image in the header?

pat270 (changes requested):

matuzalemsteles:
I don't like this approach because you can achieve this with the current component using composition. You can get something close to that.

You can achieve the same thing with this example but with a bit of CSS, maybe what we can add here is to allow the floating close button.

```jsx
const {observer, onOpenChange, open} = useModal();

if (!open) {
    return null;
}

return (
  <Modal observer={observer}>
    <Button
      aria-label="close"
      className="close new-class-to-make-floating"
      displayType="unstyled"
      onClick={() => onOpenChange(false)}
    >
      <Icon symbol="times" />
    </Button>
    <Modal.Body>
      Content
    </Modal.Body>
    <Modal.Footer
      last={
        <ClayButton.Group spaced>
          <Button
            displayType="secondary"
            onClick={() => onOpenChange(false)}
          >
            Cancel
          </Button>
          <Button onClick={() => onOpenChange(false)}>
            Save changes
          </Button>
        </Button.Group>
      }
    />
  </Modal>
);
```

packages/clay-modal/src/Context.ts
BarbaraCabrera: Looks good, I think it's a good idea to do it from the status. Thanks!

packages/clay-modal/stories/Modal.stories.tsx
marcoscv-work: It's probably better in the body, as it seems like adding it to the header adds it to the `<h1>`, but we don't want to add the semantic information of the image to the title

BarbaraCabrera:
> I don't like this approach because you can achieve this with the current component using composition. You can get something close to that.
> 
> You can achieve the same thing with this example but with a bit of CSS, maybe what we can add here is to allow the floating close button.
> 
> ```js
> const {observer, onOpenChange, open} = useModal();
> 
> if (!open) {
>     return null;
> }
> 
> return (
>   <Modal observer={observer}>
>     <Button
>       aria-label="close"
>       className="close new-class-to-make-floating"
>       displayType="unstyled"
>       onClick={() => onOpenChange(false)}
>     >
>       <Icon symbol="times" />
>     </Button>
>     <Modal.Body>
>       Content
>     </Modal.Body>
>     <Modal.Footer
>       last={
>         <ClayButton.Group spaced>
>           <Button
>             displayType="secondary"
>             onClick={() => onOpenChange(false)}
>           >
>             Cancel
>           </Button>
>           <Button onClick={() => onOpenChange(false)}>
>             Save changes
>           </Button>
>         </Button.Group>
>       }
>     />
>   </Modal>
> );
> ```


The reason to contribute it is not to use custom CSS and that it will be used both in the CMS and DXP in the marketplace integration with the page-management components to replace the existing one (`modules/apps/layout/layout-js-components-web/src/main/resources/META-INF/resources/js/components/modals/MarketplacePresentationModal.tsx`).

Here is an example of use in the [CMS](https://www.figma.com/design/fvBs3CAVkx7NlroWEQjwH8/LPD-50245---Allow-users-to-configure-the-experience-of-their-structures-for-content-creators?node-id=1-2)

packages/clay-modal/stories/Modal.stories.tsx
pat270: I see, putting everything in body is fine too. You will run into an issue where there will be no close button if you don't include an empty `<ClayModal.Header><ClayModal.Header/>`. We would need to implement a close button on `ClayModal.Body`.

marcoscv-work: That's why the idea was to make the `.modal-title` sr-only and then float the close button. The conceptualization exercise was done here: https://codepen.io/marcoscv/pen/azbwEeV
But guys, you know best how to adapt these exercises to a real-world clay implementation!

pat270: @marcoscv-work I think it's better to have an empty `<ClayModal.Header><ClayModal.Header/>` and don't output `h1` if empty. We can absolute position the close button.

matuzalemsteles:
@marcoscv-work @pat270 @BarbaraCabrera I see, we can still do it with composition without needing to render the Moda.Header additionally because we won't need it. So we can still use the Modal.Title which will still render the h1 along with the id for the dialog so we can add the sr-only to it and render the button later. I think we can create a new class for this button case.

```jsx
<Modal observer={observer}>
    <Modal.Title className="sr-only">Modal Title</Moda.Title>
    <Modal.Body>
        <Button
            aria-label="close"
            className="close new-class-to-make-floating"
            displayType="unstyled"
            onClick={() => onOpenChange(false)}
        >
            <Icon symbol="times" />
        </Button>
      Content
    </Modal.Body>
</Modal>
```

What do you think?

pat270:
Wouldn't the modal title already be in the body? In this image:

modal-with-image

the modal title would be "Marketplace is now in Page Builder".

marcoscv-work:
@pat270 It's not necessarily the title of the modal, but it could be. If so, it would have to be related via `aria-labelledby`.
We are evaluating other uses like the example in the codepen.

If you guys think the current component can handle those compositions, that would be perfect!. If you could provide detailed examples, that would also be great, and a good example of the component flexibility!

matuzalemsteles:
I think the way we can go about this is to replicate these use cases in the storybook with the examples I showed and I also think we can add some CSS to make the close button floating in this case.

@pat270 @marcoscv-work what do you think?

BarbaraCabrera:
Then I guess I can close this task and you guys can take care of it from now. 
If so, thanks a lot guys! 😄

pat270:
@matuzalemsteles @marcoscv-work sounds good

PR #5983, opened 2025-03-25
LPD-51873 Include a visible label on Clay Picker input fields

ilzamcmed:
- **fix(@clayui/date-picker): LPD-51873 improve accessibility and update types**
- **fix(@clayui/date-picker): LPD-51873 Update stories**
- **fix(@clayui/date-picker): LPD-51873 Update tests and snapshots**

Ticket: https://liferay.atlassian.net/browse/LPD-51873

## Motivation
 Improve the accessibility of the date picker by making it more usable for screen reader users

## Proposed Solution

- Add `aria-label` on input fields
- Better description once the dropdown to select a date is open

## Screenshots/videos (if applies)

https://github.com/user-attachments/assets/c18f837c-7083-485b-88cb-582a5af2703d

### DIFF
--- a/packages/clay-date-picker/src/DateNavigation.tsx
+++ b/packages/clay-date-picker/src/DateNavigation.tsx
@@ -66,6 +66,7 @@ const ClayDatePickerDateNavigation = ({
 			<div className="date-picker-nav">
 				<div className="date-picker-nav-item input-date-picker-month">
 					<Select
+						aria-label={ariaLabels.selectMonth}
 						disabled={disabled}
 						name="month"
 						onChange={(event) =>
@@ -84,6 +85,7 @@ const ClayDatePickerDateNavigation = ({
 				<div className="date-picker-nav-item input-date-picker-year">
 					<Picker
 						UNSAFE_behavior="secondary"
+						aria-label={ariaLabels.selectYear}
 						className="form-control-sm"
 						data-testid="year-select"
 						disabled={disabled}
--- a/packages/clay-date-picker/src/types.ts
+++ b/packages/clay-date-picker/src/types.ts
@@ -18,8 +18,12 @@ export interface IAriaLabels {
 	buttonDot: string;
 	buttonNextMonth: string;
 	buttonPreviousMonth: string;
-	input?: string;
+	chooseDate?: string;
 	dialog?: string;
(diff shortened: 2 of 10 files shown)


pat270 (approved):
LGTM! Works well on voice over. Not sure if this is an issue, but when YYYY-MM-DD gets focused, voice over reads Y-Y-Y-Y-M-M-D-D Edit text. I think the announcement could be clearer like "type the year, month, and date separated by a dash". We would clear the label once the user enters a date or selects via dropdown so it announces the correct date.

Also in the storybook, the `label` isn't associated with the input. We should have a `for` attribute linked to an `id`.

packages/clay-date-picker/src/DateNavigation.tsx
```diff
@@ -30,7 +30,15 @@ type Props = {
...
-	ariaLabels,
+	ariaLabels = {
+		buttonChooseDate: 'Choose date',
+		buttonDot: 'Select current date',
+		buttonNextMonth: 'Select the next month',
+		buttonPreviousMonth: 'Select the previous month',
+		chooseDate: 'Use the calendar to choose a Date. Current selection {0}',
+		selectMonth: 'Select a month',
+		selectYear: 'Select a year',
+	},
```
matuzalemsteles, line 41: Since we already have the predefined values ​​in the root component we don't need to do it here again.

packages/clay-date-picker/src/types.ts
```diff
@@ -18,8 +18,11 @@ export interface IAriaLabels {
 	buttonDot: string;
 	buttonNextMonth: string;
 	buttonPreviousMonth: string;
-	input?: string;
+	chooseDate: string;
```
matuzalemsteles, line 21: This should be optional as well to avoid build issues.

matuzalemsteles (approved):
LGTM! Just a few small changes before we move on with this.

ilzamcmed:
Thanks @pat270!
We have an `ariaLabel.input ` that can be added, but it is optional.
On the storybook, we are passing the placeholder value, so that's why it reads YYYY...

matuzalemsteles (approved):
LGTM

pat270 (approved):

PR #5991, opened 2025-03-31
fix(@clayui/css): LPD-48267 Label close button should be 24px x 24px …

pat270:
…for accessiblity

https://liferay.atlassian.net/browse/LPD-48267


matuzalemsteles:
@pat270 I think we just need to adjust the close button spacing a bit.

Screenshot 2025-04-01 at 2 34 51 AM

pat270:
@matuzalemsteles Good catch, I updated the spacing. I think it should be exactly the same as before. I don't know why it still looks off to me.

label-new-spacing

matuzalemsteles:
@pat270 this is what it looks like to me, looks good.

Default:
Screenshot 2025-04-01 at 8 30 06 PM

Large:
Screenshot 2025-04-01 at 8 29 38 PM

PR #6008, opened 2025-04-21
fix(@clayui/css): LPD-53627 Modals scope splash image to modal-banner-img instead of aspect-ratio

pat270:
https://liferay.atlassian.net/browse/LPD-53627


matuzalemsteles (approved):
LGTM

PR #6011, opened 2025-04-24
feat(@clayui/clay-css): LPD-53576 Add thumbs-up 16px icons and generated files

ilzamcmed:


packages/clay-css/src/images/icons/thumbs-up-full.svg
```diff
@@ -1,10 +1,8 @@
+
```
pat270, line 1: Don't need the new line here.

pat270 (changes requested):

matuzalemsteles (approved):

PR #6049, opened 2025-05-02
feat(@clayui/localized-input): LPD-50722 Use Language Picker for switching languages

pat270:
https://liferay.atlassian.net/browse/LPD-50722

localized-input

### DIFF
--- a/packages/clay-core/src/language-picker/LanguagePicker.tsx
+++ b/packages/clay-core/src/language-picker/LanguagePicker.tsx
@@ -14,7 +14,7 @@ import {Option, Picker} from '../picker';
 
 type DisplayType = 'info' | 'secondary' | 'success' | 'warning';
 
-type Item = {
+export type Item = {
 	id: string;
 	label: string;
 	name?: string;
@@ -133,18 +133,20 @@ const getTranslationLabel = ({
 	let displayType: DisplayType = 'warning';
 	let label: string = messages.untranslated;
 
-	if (localeId === defaultLocaleId) {
-		displayType = 'info';
-		label = messages.default;
-	} else if (translation) {
+	if (translation) {
 		const {total, translated} = translation;
 
-		if (total && total === translated) {
-			displayType = 'success';
-			label = messages.translated;
-		} else {
-			displayType = 'secondary';
-			label = sub(messages.translating, [translated, total]);
+		if (total !== 0) {
+			if (localeId === defaultLocaleId) {
+				displayType = 'info';
+				label = messages.default;
+			} else if (total === translated) {
+				displayType = 'success';
+				label = messages.translated;
+			} else {
+				displayType = 'secondary';
+				label = sub(messages.translating, [translated, total]);
+			}
 		}
 	}
 
--- a/packages/clay-core/src/language-picker/LanguagePicker.tsx
+++ b/packages/clay-core/src/language-picker/LanguagePicker.tsx
@@ -141,7 +141,7 @@ const getTranslationLabel = ({
 		const {total, translated} = translation;
 
 		if (translated !== 0) {
-			 if (total === translated) {
+			if (total === translated) {
 				displayType = 'success';
 				label = messages.translated;
 			} else {
--- a/packages/clay-core/src/language-picker/LanguagePicker.tsx
+++ b/packages/clay-core/src/language-picker/LanguagePicker.tsx
@@ -7,6 +7,7 @@ import ClayIcon from '@clayui/icon';
 import ClayLabel from '@clayui/label';
 import ClayLayout from '@clayui/layout';
 import {InternalDispatch, sub} from '@clayui/shared';
+import {ClayTooltipProvider} from '@clayui/tooltip';
 import classNames from 'classnames';
 import React from 'react';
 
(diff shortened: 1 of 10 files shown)


packages/clay-core/src/language-picker/LanguagePicker.tsx
```diff
@@ -133,18 +133,20 @@ const getTranslationLabel = ({
...
+	if (translation) {
 		const {total, translated} = translation;
 
-		if (total && total === translated) {
-			displayType = 'success';
-			label = messages.translated;
-		} else {
-			displayType = 'secondary';
-			label = sub(messages.translating, [translated, total]);
+		if (total !== 0) {
```
ethib137, line 139: This logic doesn't seem right. This means the only way to display the untranslated message would be if total is 0. But I think total refers to "total available translations", so this isn't correct. It should probably be `if (translated !== 0)` since translated refers to which of the available translations have been translated.

packages/clay-localized-input/src/index.tsx
```diff
@@ -111,6 +115,36 @@ const LocalizedInput = React.forwardRef<HTMLInputElement, IProps>(
...
+
+			return languagePickerTranslations;
+		}, []);
+
+		useMemo(() => {
+			languagePickerTranslations[selectedLocale?.id!] = {
+				total: 1,
+				translated: translations[selectedLocale.label] ? 1 : 0,
+			};
+		}, [translations]);
```
ethib137, line 146: This can all be updated to something like this:

```
const languagePickerTranslations = useMemo(() => {
	const languagePickerTranslations = {};

	locales.forEach((locale) => {
		languagePickerTranslations[locale.symbol] = {
			total: 1, // this should always be one because we just have one input being translated.
			translated: translations[locale.symbol] ? 1 : 0],
		};
	}

	return languagePickerTranslations;
}, [locales, translations]);
```

For localized input the status label should never be "translating", so make sure to check that.

Also, an empty string should be rendered as untranslated, not translated. Test this by adding a translation and then removing it.

ethib137 (changes requested):
Thanks Patrick... just a couple things.

Also, the ticket is missing this, but we had discussed that we should use the [trigger-without-language-label](https://clayui.com/docs/components/language-picker#trigger-without-language-label) version of the Language selector since it takes up less space. Can you update to use that? With that change it will also be important to ensure that version of the Language Selector has a tooltip so that it's clear for accessibility. 

Thanks.

packages/clay-localized-input/src/index.tsx
pat270: @ethib137 in the storybook, `translations` keys match `locales` label values. The symbol values are lower case. `translations[locale.symbol]` returns undefined because `en-us` should be `en-US` should mixed case keys still be valid or should it be an exact match?

pat270:
I'm also getting an error when trying to apply `ariaLabels` https://github.com/liferay/clay/blob/a9af679de74b34fac8f826aec407c1fd96047d28/packages/clay-localized-input/src/index.tsx#L27-L32 to LanguagePicker `messages`.https://github.com/liferay/clay/blob/a9af679de74b34fac8f826aec407c1fd96047d28/packages/clay-core/src/language-picker/LanguagePicker.tsx#L24-L31 

```
TypeError: Cannot read properties of undefined (reading 'split')
    at sub (http://localhost:8888/packages_clay-shared_src_index_tsx.iframe.bundle.js:1508:26)
    at children (http://localhost:8888/packages_clay-core_src_language-picker_index_ts.iframe.bundle.js:222:75)
    at http://localhost:8888/packages_clay-core_src_collection_index_ts.iframe.bundle.js:467:125
    at http://localhost:8888/packages_clay-core_src_collection_index_ts.iframe.bundle.js:583:5
    at mountMemo (http://localhost:8888/vendors-node_modules_storybook_addon-a11y_preview_js-node_modules_storybook_addon-actions_pre-a27f80.iframe.bundle.js:71290:19)
    at Object.useMemo (http://localhost:8888/vendors-node_modules_storybook_addon-a11y_preview_js-node_modules_storybook_addon-actions_pre-a27f80.iframe.bundle.js:71586:16)
    at useMemo (http://localhost:8888/vendors-node_modules_storybook_addon-a11y_preview_js-node_modules_storybook_addon-actions_pre-a27f80.iframe.bundle.js:84626:21)
    at useCollection (http://localhost:8888/packages_clay-core_src_collection_index_ts.iframe.bundle.js:575:50)
    at Picker (http://localhost:8888/packages_clay-core_src_language-picker_index_ts.iframe.bundle.js:779:79)
    at renderWithHooks (http://localhost:8888/vendors-node_modules_storybook_addon-a11y_preview_js-node_modules_storybook_addon-actions_pre-a27f80.iframe.bundle.js:70651:18)
```
I get the same error when trying to apply directly to the LanguagePicker story.
https://github.com/liferay/clay/blob/a9af679de74b34fac8f826aec407c1fd96047d28/packages/clay-core/stories/LanguagePicker.stories.tsx#L145-L154

I can't figure out how the data should be organized.

pat270:
@ethib137 I figured out the issue for `TypeError: Cannot read properties of undefined (reading 'split')`. It's caused by the sub method in `/clay/packages/clay-shared/src/sub.ts` due to option and translating being missing from the messages prop in LocalizedInput.

I was thinking maybe we can let the sub method do nothing if it receives undefined / null arguments.

PR #6055, opened 2025-05-13
feat(www): implements the new CodeBlock

matuzalemsteles:
Well, I refactor the CodeBlock implementation that removes the use of the sandpack and brings many user experience improvements. This implementation still keeps the button to open the example in codesandbox, which is great because it is a very important feature.

Live: https://next.clayui.com/docs/components/alert

Some highlights of this new codeblock in the documentation:

- Uniform code blocks for the examples and snippets
- Processed locally instead of running on a codesandbox server
- Much faster loading of the pages
- Easier to test the examples since we no longer have an iframe
- Automatically collapses for very large code
- The example code has intellisense of the types on hover

Screenshot 2025-05-13 at 12 41 46 AM

Screenshot 2025-05-13 at 12 35 34 AM

As part of the new CodeBlock I also needed to update the `renoun` dependency and I would need to refactor part of the documentation to use the new APIs. This also brought performance improvements to local testing so the pages load faster than before.

This change solves some bugs such as when opening the example in codesandbox the icons now load normally and it is also now possible to add the chart examples again which will be done in another PR #5955.

## ToDo

- [ ] Fix React v16 internal dependency bug with React v19 for TreeView examples
- [ ] Add remaining dependencies for all examples
- [ ] Fixes the bug where some examples still don't render


ethib137:
Thanks Matu! This looks like a great improvement. 

Would it be difficult to also make the code blocks for examples editable on the site?

PR #6056, opened 2025-05-13
feat(@clayui/css): LPD-53483 Adds inline-text-input

pat270:
https://liferay.atlassian.net/browse/LPD-53483

inline-text

### DIFF
--- a/packages/clay-form/src/Input.tsx
+++ b/packages/clay-form/src/Input.tsx
@@ -200,27 +200,27 @@ const InlineText = React.forwardRef<HTMLDivElement, IInlineTextProps>(
 		{
 			children,
 			className,
-			defaultInputValue,
-			id,
-			inputValue: externalInputValue,
-			onInputValueChange,
+			defaultValue,
+			onValueChange,
 			placeholder,
-			readOnly,
 			spritemap,
+			value: externalValue,
 			...otherProps
 		}: IInlineTextProps,
 		ref: React.Ref<HTMLDivElement>
 	) => {
 		const inputRef = React.useRef<HTMLSpanElement>(null);
 
-		const [inputValue = '', setInputValue] = useControlledState({
-			defaultName: 'defaultInputValue',
-			defaultValue: defaultInputValue,
-			handleName: 'onInputValueChange',
-			name: 'inputValue',
-			onChange: onInputValueChange,
-			value: externalInputValue,
+		const [value = '', setValue] = useControlledState({
+			defaultName: 'defaultValue',
+			defaultValue: defaultValue,
+			handleName: 'onValueChange',
+			name: 'value',
+			onChange: onValueChange,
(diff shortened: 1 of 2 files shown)


ethib137:
Shouldn't we be adding this to our docs site too? https://clayui.com/docs/components/input

packages/clay-form/src/Input.tsx
```diff
@@ -155,6 +156,97 @@ const GroupInsetItem = React.forwardRef<
...
+		className,
+		id,
+		placeholder,
+		readOnly,
+		spritemap,
+		...otherProps
+	}: IInlineTextProps) => {
+		const inputRef = React.useRef<HTMLDivElement>(null);
+
+		const [inputValue, setInputValue] = React.useState('');
```
matuzalemsteles, line 199: To follow the pattern of other components we need the state to be controlled or uncontrolled you can see how we implemented this in Autocomplete for example.

packages/clay-form/src/Input.tsx
```diff
@@ -155,6 +156,97 @@ const GroupInsetItem = React.forwardRef<
...
+>(
+	({
+		children,
+		className,
+		id,
+		placeholder,
+		readOnly,
+		spritemap,
+		...otherProps
+	}: IInlineTextProps) => {
```
matuzalemsteles, line 196: We need to pass the `ref` forward to the span element of the component and in this case we just need to set the ref type to `HTMLSpanElement` or just pass it to the root element which might be better for the developer.

matuzalemsteles (commented):
I added some comments but overall it looks good.

ethib137 (changes requested):
Hey @pat270 , I think with this current implementation we don't have anyway to display a saved value in the inline input on the first load? So, the inline input will always be empty when it loads even if a value is passed to it. As long as we use contentEditable I don't think this will work, since content editable does not play nicely with React.

What was the reason for using content editable instead of an unstyled input?

pat270:
@ethib137 Good catch, I'll update it.

Standard inputs are a fixed width and can't grow with its content. There is some discussion at https://liferay.atlassian.net/browse/LPD-53482?focusedCommentId=2911231

pat270:
@ethib137 I've been trying to get the value to display in the contenteditable element, but I'm running into some React quirks.  Using `{value}` in the contenteditable span causes the text cursor to move to the front which makes the text print out right to left.
cursor-to-front
I got it to work in uncontrolled mode by using `defaultValue`. The controlled component doesn't work. I have a question about why we need to support both types. Is it really necessary?

> What was the reason for using content editable instead of an unstyled input?

I was hoping Lexicon could revisit the design / interactions and accommodate the limitations of the standard HTML input.

pat270:
Closing this since it was discarded, will reopen if revisited.

PR #6058, opened 2025-05-14
feat(@clayui/core): adds SidePanel implementation

matuzalemsteles:
Ticket [LPD-48055](https://liferay.atlassian.net/browse/LPD-48055)

This is an initial implementation of the SidePanel without CSS implementation and the markup may also change but talking to Patrick ideally we will reuse the `.sidebar` and `.c-slideout` classes but there should be adjustments to meet the ticket requirements, the CSS part has a separate ticket for that.

This initial implementation contains the component structure and the necessary states, also the trigger focus recovery in an OOTB way.

There are some cases to implement still like when the title is rendered in the header under the covers we can add the `aria-labelledby` to the sidepanel OOTB when an `aria-label` is not defined in the root component. I also need to adjust the SidePanel animation when opening, @pat270 you can check that I'm not sure if the order is right here.

### DIFF
--- a/packages/clay-core/src/side-panel/SidePanel.tsx
+++ b/packages/clay-core/src/side-panel/SidePanel.tsx
@@ -150,25 +150,31 @@ export function SidePanel({
 			document.addEventListener('keydown', onKeyDown, true);
 
 			return () => {
-				// Remove Side Panel content from DOM focus management
-				ref.current?.setAttribute('inert', '');
-
 				document.removeEventListener('keydown', onKeyDown, true);
 			};
+		} else {
+			// Remove Side Panel content from DOM focus management
+			ref.current?.setAttribute('inert', '');
 		}
 	}, [open]);
 
 	useLayoutEffect(() => {
 		const performSlideout = (position: string | number) => {
 			if (ref.current) {
-				animate(
-					ref.current,
-					{[direction]: position},
-					{
-						duration: 0.3,
-						ease: 'easeInOut',
-					}
-				);
+				// Only use animate in the browser environment
+				if ('animate' in ref.current) {
+					animate(
+						ref.current,
+						{[direction]: position},
+						{
+							duration: 0.3,
+							ease: 'easeInOut',
+						}
+					);
+				} else {
+					// @ts-ignore
+					(ref.current as HTMLDivElement).style[direction] = position;
+				}
 			}
 		};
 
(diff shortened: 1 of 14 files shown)


packages/clay-core/src/side-panel/SidePanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+			<As
+				className={classnames('sidebar c-slideout-active', className, {
+					'c-slideout-show': open,
+					'c-slideout-transition': transition,
+					'c-slideout-transition-in': transition && open,
+					'c-slideout-transition-out': transition && !open,
+					'sidebar-dark': displayType === 'dark',
+					'sidebar-light': displayType === 'light',
+				})}
+				onTransitionEnd={() => setTransition(false)}
```
matuzalemsteles, line 176: I wanted to move this animation to [motion](https://motion.dev/) which handles animations better and also reduced motion for example but unfortunately it is only compatible with React v19. I need to check if old versions are compatible with React v16.

pat270: @matuzalemsteles the slide in isn't animating because there needs to be a slight delay for the browser to apply `c-slideout-transition c-slideout-transition-in` styles. I made some little tweaks and it's working for me, but it's hacky.

https://github.com/pat270/clay/commit/55216977e43447042c728323a5deb794791cafaf

matuzalemsteles: Thanks @pat270! I'm going to try using `motion` for this, I think it will simplify and handle edge cases better.

matuzalemsteles:
Well I've submitted some more changes now with animation working better and also removing focus from sidebar elements when it's closed. I've also added the title reference to the sidebar when `aria-label` and `aria-labelledby` are not defined, a nice feature to have.

To finish the implementation on the React side, I need to create the documentation and add tests, I'll work on that on Monday before I go on vacation. It will be necessary to adjust or build a new CSS for the component to get the desired look for the SidePanel, we'll use the [LPD-55065](https://liferay.atlassian.net/browse/LPD-55065) ticket for that.

@pat270 the animation implementation is now using `motion` I believe we would no longer need a `c-slideout` wrapper for the sidebar nor the `c-slideout-fixed` tools could we add some new CSS classes that handle all this directly from the sidebar instead. what do you think?

ethib137:
Resent at https://github.com/liferay/clay/pull/6072

PR #6059, opened 2025-05-14
feat(@clayui/css): LPD-51801 Adds .blockquote variant for .form-text

pat270:
https://liferay.atlassian.net/browse/LPD-51801

localized-input1

~Not sure if I implemented this correctly, but it looks like they just want a style change?~

### DIFF
--- a/packages/clay-css/src/scss/cadmin/components/_form-validation.scss
+++ b/packages/clay-css/src/scss/cadmin/components/_form-validation.scss
@@ -20,6 +20,22 @@
 
 .form-text {
 	@include clay-css($cadmin-form-text);
+
+	$_blockquote: map-get($cadmin-form-text, blockquote);
+
+	@if ($_blockquote) {
+		&.blockquote {
+			@include clay-css($_blockquote);
+
+			$_before: map-get($_blockquote, before);
+
+			@if ($_before) {
+				&::before {
+					@include clay-css($_before);
+				}
+			}
+		}
+	}
 }
 
 // Non HTML5 Form Validator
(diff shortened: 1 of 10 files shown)


ethib137 (changes requested):
Thanks @pat270 ... In looking at the gif, it appears that spacing between the input and the text does not match what is in the figma. Could you update that? Also make sure the spacing is correct when there is an input description or an error as well.

Additionally, this should be added to the story book and to the [forms documentation page.](https://clayui.com/docs/components/form)

Thanks.

ethib137:
> Thanks @pat270 ... In looking at the gif, it appears that spacing between the input and the text does not match what is in the figma. Could you update that? Also make sure the spacing is correct when there is an input description or an error as well.
> 
> Additionally, this should be added to the story book and to the [forms documentation page.](https://clayui.com/docs/components/form)
> 
> Thanks.

@pat270 can you add this to the forms documentation page?

pat270:
@ethib137 I removed the use of `ClayForm.Group` from LocalizedInput. It should only be used by the dev when composing a form group. Teams are overriding the `margin-bottom` in the DXP use case.

form-group

ethib137:
Hey @pat270 this still looks like it hasn't been added to the [forms documentation page](https://clayui.com/docs/components/form). It's only been added to the [LocalizedInput docs page](https://clayui.com/docs/components/localized-input). Please could you add it to the [forms documentation page](https://clayui.com/docs/components/form), since this is built using the `<ClayForm.Text>` component.

PR #6069, opened 2025-05-22
LPD-48055 - DRAFT PR

ethib137:
Hey @pat270, this is mostly working now with css transitions but it is adding horizontal scroll bars during the transition. I'm guessing that's because I'm doing something wrong. Can you take a look at it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
 	"size-limit": [
 		{
 			"path": "all.js",
-			"limit": "159 kb",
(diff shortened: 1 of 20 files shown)


pat270:
@ethib137 I added the classes `c-slideout-push-start` and `c-slideout-push-end` for the parent container. I also made some updates to the component and storybook. You can see my commits at https://github.com/pat270/clay/commits/LPD-48055/.

PR #6072, opened 2025-05-26
feat(@clayui/core): adds SidePanel implementation 

ethib137:
Updated to use css transitions instead of motion js.

_Initial Message_

Ticket [LPD-48055](https://liferay.atlassian.net/browse/LPD-48055)

This is an initial implementation of the SidePanel without CSS implementation and the markup may also change but talking to Patrick ideally we will reuse the .sidebar and .c-slideout classes but there should be adjustments to meet the ticket requirements, the CSS part has a separate ticket for that.

This initial implementation contains the component structure and the necessary states, also the trigger focus recovery in an OOTB way.

There are some cases to implement still like when the title is rendered in the header under the covers we can add the aria-labelledby to the sidepanel OOTB when an aria-label is not defined in the root component. I also need to adjust the SidePanel animation when opening, @pat270 you can check that I'm not sure if the order is right here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
 	"size-limit": [
 		{
 			"path": "all.js",
-			"limit": "159 kb",
+			"limit": "160 kb",
 			"ignore": [
 				"react",
 				"react-dom"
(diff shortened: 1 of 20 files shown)


pat270 (approved):

pat270:
@ethib137 I'll add a subtask to the story for these changes

PR #6074, opened 2025-05-28
fix(@clayui/css): LPD-55065 Update c-slideout and sidebar to adhere …

pat270:
…to Lexicon Side Panel specs

https://liferay.atlassian.net/browse/LPD-55065
https://liferay.atlassian.net/browse/LPD-33184

### DIFF
--- a/packages/clay-core/src/side-panel/Header.tsx
+++ b/packages/clay-core/src/side-panel/Header.tsx
@@ -4,6 +4,7 @@
  */
 
 import Icon from '@clayui/icon';
+import classnames from 'classnames';
 import React from 'react';
 
 import {useSidePanel} from './context';
@@ -13,14 +14,32 @@ type Props = {
 	 * Children content to render a content.
 	 */
 	children: React.ReactNode;
+
+	/**
+	 * Sets the CSS className for the component.
+	 */
+	className?: string;
+
+	/**
+	 * Property to make the Header sticky.
+	 */
+	sticky?: boolean;
 };
 
-export function Header({children}: Props) {
+export function Header({children, className, sticky}: Props) {
 	const {onOpenChange} = useSidePanel();
 
 	return (
-		<div className="sidebar-header">
-			<div className="autofit-row sidebar-section">
+		<div
+			className={classnames(
+				'sidebar-header',
+				{
+					'sticky-top': sticky,
+				},
+				className
+			)}
+		>
+			<div className="autofit-row">
 				<div className="autofit-col autofit-col-expand">{children}</div>
 				<div className="autofit-col">
 					<button
--- a/packages/clay-core/src/side-panel/Header.tsx
+++ b/packages/clay-core/src/side-panel/Header.tsx
@@ -21,7 +21,9 @@ type Props = {
 	className?: string;
 
 	/**
-	 * Property to make the Header sticky.
+	 * Property to make the Header sticky. Absolutely positioned SidePanel's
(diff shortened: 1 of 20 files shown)


pat270:
@ethib137 I wasn't able to get something working for fixed header and footer without using `position: fixed`. I think we need to use position fixed and calculate the top bar heights if we want fixed header and footer.

c-slideout

ethib137:
Hey @pat270 , 

Thanks for the info on fixed vs absolute. I'll check with product and design and see what we have to do.

Regarding styling it looks like we are still missing a border for the header and footer. Additionally, it would be good to add a new storybook example that includes the additional elements in the header like we see [here](https://www.figma.com/design/vLPRyGECU7lFC86wsbd5Hy/design-5654-side-panel?node-id=47-17263&t=vAV4yIc4moFlZGCT-0). Thanks.

pat270:
@ethib137 We might not need the position fixed. I misinterpreted the height.

> When we talk about “Component Height”, we are talking about the actual height of the component, not the total height of the browser.

I went ahead and added a min-height to the container since container queries for height don't work. See https://liferay.atlassian.net/browse/LPD-45006?focusedCommentId=2939028.

I also added a sticky prop to the Header and Footer so the dev can choose to have one or both fixed.

sidePanelSticky

packages/clay-css/src/scss/cadmin/components/_sidebar.scss
```diff
@@ -121,3 +145,23 @@
...
+	position: sticky;
+	top: 0;
+	z-index: $cadmin-zindex-sticky;
+}
+
+.sidebar-footer-sticky {
+	bottom: 0;
+	position: sticky;
+	z-index: $cadmin-zindex-sticky;
+}
```
ethib137, line 159: Hey @pat270 we already have a sticky-top utility class. Why do we need the sidebar-header-sticky class? Also instead of sidebar-footer-sticky class, maybe copy the current pattern and add a sticky-bottom class instead. These would be usable in more situations.

packages/clay-core/src/side-panel/Header.tsx
```diff
@@ -13,14 +14,32 @@ type Props = {
...
 	const {onOpenChange} = useSidePanel();
 
 	return (
-		<div className="sidebar-header">
-			<div className="autofit-row sidebar-section">
+		<div
+			className={classnames(
+				'sidebar-header',
+				{
+					'sidebar-header-sticky': sticky,
```
ethib137, line 37: Let's use the sticky-top utility class instead.

packages/clay-core/stories/SidePanel.stories.tsx
```diff
@@ -94,14 +120,83 @@ export const Default = () => {
 					onOpenChange={setOpen}
 					open={open}
 				>
-					<SidePanel.Header>
-						<SidePanel.Title>Right Panel</SidePanel.Title>
+					<SidePanel.Header className="asdfadsf" sticky>
```
ethib137, line 123: This looks like an unintended added className

ethib137 (changes requested):
Hey @pat270 this looking better. We are still going to need fixed positioning as an option though because we want to have independent scroll for the cms use case. I'm going to merge my PR so that you can rebase off of mine and make sure fixed and absolute positioning is working as expected. Thanks.

packages/clay-core/stories/SidePanel.stories.tsx
pat270: oops thanks for getting this

PR #6077, opened 2025-05-30
fix(@clayui/core): Fix the SidePanel so that starting in an open state works.

ethib137:
https://liferay.atlassian.net/browse/LPD-48055


pat270:
@ethib137 I cherry picked this commit to https://github.com/liferay/clay/pull/6074 and updated the snapshots. I also made a minor change so the SidePanel doesn't animate when it's already open.

open-instantly

ethib137:
Thanks @pat270 !

PR #6083, opened 2025-06-05
fix(@clayui/css): LPD-57411 SVG Icon slash.svg should use fill instea…

pat270:
…d of stroke

https://liferay.atlassian.net/browse/LPD-57411


SelenaAungst (approved):
LGTM!

PR #6095, opened 2025-06-24
feat(@clayui/core): LPD-58729 - Allow attaching onClick handlers to table cell

juanjofgliferay:
https://liferay.atlassian.net/browse/LPD-58729

Current @clayui/core Table Cell expects `onClick` events only for header cells to trigger column sorting, preventing any other custom `onClick` event from running.

With the proposed change, we keep the current behavior while allowing users to attach custom `onClick` events to any other table cell.


ethib137 (approved):
Thank you @juanjofgliferay . LGTM

PR #6097, opened 2025-06-27
feat(@clayui/autocomplete): LPD-54854 Allow adding new items to the autocomplete dropdown

pat270:
https://liferay.atlassian.net/browse/LPD-54854

This is still a draft PR. I'm having issues with the React component type (controlled / uncontrolled). From my experimentation, controlled autocomplete isn't functioning, but it could be a knowledge issue here. The storybook example I added is an uncontrolled component.

This PR adds the prop `onAddNewItem`. It's called when a new item is added to the list via the enter key or clicking on the dropdown item button. I'm pretty sure I've implemented it incorrectly. I'm looking for some help writing it the right way.

I've added the keys `allowsCustomValue` and `setAsHTML` to the `messages` prop. `allowsCustomValue` is used to pass in the text to display for the Create New Menu Item interaction, https://liferay.atlassian.net/wiki/spaces/ENGLEXICON/pages/3811442825/Create+New+Option#%E2%9C%A8-Interaction.

The `setAsHTML` key is used to enable the ability to pass in markup as well as text for the Create New Menu Item interaction.

### DIFF
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -114,16 +114,23 @@ export interface IProps<T>
 	 * Messages for autocomplete.
 	 */
 	messages?: {
+		allowsCustomValue?: string;
 		listCount?: string;
 		listCountPlural?: string;
 		loading: string;
 		notFound: string;
+		setAsHTML?: boolean;
 	};
 
 	/**
 	 * Callback for when the active state changes (controlled).
 	 */
 	onActiveChange?: InternalDispatch<boolean>;
+
+	/**
+	 * Callback called when an item is added to the autocomplete list.
+	 */
+	onAddNewItem?: () => void;
 
 	/**
 	 * Callback called when input value changes (controlled).
@@ -189,10 +196,12 @@ function hasItem<T extends Item>(
 const ESCAPE_REGEXP = /[.*+?^${}()|[\]\\]/g;
 
 const defaultMessages = {
+	allowsCustomValue: 'Add {0}',
 	listCount: '{0} option available.',
 	listCountPlural: '{0} options available.',
 	loading: 'Loading...',
 	notFound: 'No results found',
+	setAsHTML: false,
 };
 
 function AutocompleteInner<T extends Item>(
@@ -214,6 +223,7 @@ function AutocompleteInner<T extends Item>(
 		menuTrigger = 'input',
 		messages,
 		onActiveChange,
+		onAddNewItem,
 		onChange,
 		onItemsChange,
 		onLoadMore,
@@ -279,6 +289,8 @@ function AutocompleteInner<T extends Item>(
 			) !== null,
 		[value]
 	);
+
+	const allowsCustomValueMessage = messages?.allowsCustomValue!;
 
 	useEffect(() => {
 		// Validates that the initial value exists in the items.
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
 		),
 		items: filteredItems,
 		notFound: (
-			<DropDown.Item
-				aria-disabled="true"
-				className="disabled"
-				roleItem="option"
-			>
-				{messages.notFound}
-			</DropDown.Item>
+			<>
+				{allowsCustomValue && (
+					<>
+						<DropDown.Item
+							className=""
+							id={value}
+							key={value}
+							onClick={() => {
+								if (allowsCustomValue && items && onAddNewItem) {
+									onAddNewItem();
+
+									items.push(value);
+
+									inputElementRef.current?.focus();
+
+									setValue('');
+								}
+							}}
+							roleItem="option"
+						>
+							{messages.setAsHTML &&
+							typeof allowsCustomValueMessage === 'string' ? (
+								<span
+									dangerouslySetInnerHTML={{
+										__html: sub(allowsCustomValueMessage, [
+											value,
+										]),
+									}}
+								/>
+							) : (
+								sub(allowsCustomValueMessage, [value])
+							)}
+						</DropDown.Item>
+						<li className="dropdown-divider"></li>
+					</>
+				)}
+				<DropDown.Item
+					aria-disabled="true"
+					className="disabled"
+					roleItem="option"
+				>
+					{messages.notFound}
+				</DropDown.Item>
+			</>
 		),
 		suppressTextValueWarning: false,
 		virtualizer: items ? virtualizer : undefined,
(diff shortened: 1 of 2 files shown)


packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
+									items.push(value);
+
+									inputElementRef.current?.focus();
+
+									setValue('');
+								}
+							}}
+							roleItem="option"
+						>
+							{messages.setAsHTML &&
```
ethib137, line 439: We don't need to allow setting html. The designs don't provide this flexibility so we shouldn't either..

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
+					<>
+						<DropDown.Item
+							className=""
+							id={value}
+							key={value}
+							onClick={() => {
+								if (allowsCustomValue && items && onAddNewItem) {
+									onAddNewItem();
+
+									items.push(value);
```
ethib137, line 430: We shouldn't be adding the value to the list of items. That behavior can be handled if it's a controlled component by having the parent component add it to the list when onAddNewItem is called.

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
+			<>
+				{allowsCustomValue && (
+					<>
+						<DropDown.Item
+							className=""
+							id={value}
+							key={value}
+							onClick={() => {
+								if (allowsCustomValue && items && onAddNewItem) {
+									onAddNewItem();
```
ethib137, line 428: This should probably call `onAddNewItem(value);` passing in value.

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
+
+									items.push(value);
+
+									inputElementRef.current?.focus();
+
+									setValue('');
```
ethib137, line 434: We shouldn't be setting the value to an empty string. When this is selected we are confirming the value of the input, so the value should stay.

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -189,10 +196,12 @@ function hasItem<T extends Item>(
 const ESCAPE_REGEXP = /[.*+?^${}()|[\]\\]/g;
 
 const defaultMessages = {
+	allowsCustomValue: 'Add {0}',
```
ethib137, line 199: This should probably be called something more like "addNewItem" or "addCustomValue"

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
-				className="disabled"
-				roleItem="option"
-			>
-				{messages.notFound}
-			</DropDown.Item>
+			<>
+				{allowsCustomValue && (
+					<>
+						<DropDown.Item
+							className=""
```
ethib137, line 423: We shouldn't need to pass empty className

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
+							id={value}
+							key={value}
```
ethib137, line 425: Why are you adding id and key props?

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -554,6 +603,14 @@ function AutocompleteInner<T extends Item>(
 							break;
 						}
 						case Keys.Enter: {
+							if (allowsCustomValue && items && onAddNewItem) {
+								onAddNewItem();
+
+								items.push(value);
```
ethib137, line 609: Same here, we shouldn't be adding to items.

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -404,13 +416,50 @@ function AutocompleteInner<T extends Item>(
...
+									}}
+								/>
+							) : (
+								sub(allowsCustomValueMessage, [value])
+							)}
+						</DropDown.Item>
+						<li className="dropdown-divider"></li>
+					</>
+				)}
+				<DropDown.Item
```
ethib137, line 455: Shouldn't this only render if `allowsCustomValue` is false?

packages/clay-autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -506,3 +506,69 @@ export const AsyncData = () => {
...
+							Categories
+						</label>
+						<ClayAutocomplete
+							allowsCustomValue
+							aria-labelledby="clay-autocomplete-label-add-new-item"
+							defaultItems={items}
+							id="clay-autocomplete-add-new-item"
+							messages={{
+								allowsCustomValue:
+									'<span class="text-primary">{0} <i>(Add New Category)</i></span>',
```
ethib137, line 544: Adding text-primary should be in our component rather than our example.

packages/clay-autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -506,3 +506,69 @@ export const AsyncData = () => {
...
+								setAsHTML: true,
+							}}
+							onAddNewItem={() => {
+								if (items.indexOf(value) < 0) {
+									setItems(() => {
+										const newItems = [
```
ethib137, line 550: To make this controlled, you need to add the `items` prop to the autocomplete passing in your items.

ethib137 (changes requested):
Hey @pat270 this is in the right direction. A few changes are needed to simplify it and clean up the api. Thanks!

packages/clay-autocomplete/src/Autocomplete.tsx
pat270: @ethib137 in some of the documentation there were some differences that required flexibility.

[Case 1](https://www.figma.com/design/8iouiTNmXq56OcuhWQEGry/LPD-51807-STOR-Autocomplete---Dropdown-with-fixed-button?node-id=1-476&p=f&t=VIO5MLaTjCeTP6FY-0): Has parts of the message italicized and message text after the value

case-1

[Case 2](https://liferay.atlassian.net/wiki/spaces/PRDDXP/pages/3811442825/Create+New+Option#%F0%9F%8E%A8--Design): No italics and message is before the value

case-2

HTML passed in the message prop gets escaped. I believe the only way to support this is through setting HTML.

packages/clay-autocomplete/src/Autocomplete.tsx
pat270: @ethib137 I was under the impression that `allowsCustomValue === true` is for adding items not on the list.

```
	/**
	 * Flag to allow an input value not corresponding to an item to be set.
	 */
```

We could make the description better.

PR #6099, opened 2025-07-09
chore: use node 20 to match DXP version

bryceosterhaus:
Not sure if there was a reason we pinned at node 18... I updated our Netlify settings to use node 20 as well


bryceosterhaus:
It looks like Node 18 is required for lerna when using vercel. So we would need to update that before using node 20

bryceosterhaus:
@ethib137 yeah we might need to, it's probably long overdue to move off of it since I think we can just use yarn workspaces.

However, I don't think this is critical at the moment, so we can leave 18.x for now

PR #6101, opened 2025-07-10
chore: get clayui.com site working on vercel

bryceosterhaus:
These commits get the clay site working with auto-builds on vercel(tested on my own account). 

I am just waiting on legal and IT to get the proper vercel account so that we can link the domain.

Note: we may need to tweak lerna when we release next after merging this. I tested it a bit, but not a full publish.


pat270:
I'm getting an error when running `yarn lerna`.

```
$ lerna bootstrap -- --no-optional --no-package-lock
ERR! bootstrap The "bootstrap" command was removed by default in v7, and is no longer maintained.
ERR! bootstrap Learn more about this change
```

bryceosterhaus:
@pat270 do you use that command specifically for anything? I noticed we don't use `yarn lerna` in our release process, so I wonder if we can just remove that script. I believe `yarn` with workspaces supersedes `lerna bootstrap`

PR #6103, opened 2025-07-11
LPD-60233 Add accessible text to clay table's column visibility header cell

ethib137:
Resent from: https://github.com/liferay/clay/pull/6102

_Original Message:_

**Bug ticket:** https://liferay.atlassian.net/browse/LPD-59500
**Subtask:** https://liferay.atlassian.net/browse/LPD-60233

---

- This adds sr-only text to the columns visibility header to fix an accessibility issue where header cells can't be empty.
  - Using the WAVE tool extension mentioned in the bug ticket this is how it appears:  Screenshot 2025-07-10 at 2 05 30 PM

- This also adds a new message `columnsVisibilityCellLabel` to allow customizing the message for cases when this column is also used as the "row actions"
  - For example: Screenshot 2025-07-10 at 4 44 35 PM

### DIFF
--- a/packages/clay-core/src/table/Head.tsx
+++ b/packages/clay-core/src/table/Head.tsx
@@ -86,6 +86,11 @@ export function Head<T extends Record<string, any>>(
 
 					{columnsVisibility && (
 						<Cell keyValue="visibility" width="72px">
+							<span className="sr-only">
+								{messages['columnsVisibilityCellLabel'] ||
+									messages['columnsVisibility']}
+							</span>
+
 							<Menu
 								UNSAFE_focusableElements={[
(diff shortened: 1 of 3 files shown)


ethib137:
@pat270 can you please merge this if it looks okay to you?

pat270 (approved):
@ethib137 it looks good to me. I don't like how it's redundant, but I can't think of another work around with the button in the table header. Can we go back to design and have them put the columns visibility button outside the table?

This isn't related to this issue, but I noticed the switches don't have a label associated with them. Voice Over announces switch on.

switches-not-linked

I'll file an issue for it.

PR #6108, opened 2025-07-16
fix(date-picker): create a new date object for initialMonth if none is provided and create new variable so that we aren't re-assigning a prop value

bryceosterhaus:
https://liferay.atlassian.net/browse/LPD-61062

one of the rules in react is that we shouldn't ever manually change a prop value or reassign it. So I moved initialMonth to a newly named variable and set it to `new Date()` if no prop is assigned.

### DIFF
--- a/packages/clay-date-picker/src/index.tsx
+++ b/packages/clay-date-picker/src/index.tsx
@@ -259,7 +259,7 @@ const DatePicker = React.forwardRef<HTMLInputElement, IProps>(
 			footerElement,
 			id,
 			initialExpanded = false,
-			initialMonth = NEW_DATE,
+			initialMonth,
 			inputName,
 			months = [
 				'January',
@@ -299,7 +299,7 @@ const DatePicker = React.forwardRef<HTMLInputElement, IProps>(
 	) => {
 		// For backward compatibility `defaultMonth` is just an alias for
 		// `initialMonth`.
-		initialMonth = defaultMonth ?? initialMonth;
+		const initialMonthInternal = defaultMonth ?? initialMonth ?? new Date();
 
 		const [internalValue, setValue, isUncontrolled] = useControlledState({
 			defaultName: 'defaultValue',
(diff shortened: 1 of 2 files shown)


ethib137:
Hey @bryceosterhaus it looks like we are no longer getting the storybook preview here. Do you know why that is?

packages/clay-date-picker/src/index.tsx
```diff
@@ -554,32 +557,34 @@ const DatePicker = React.forwardRef<HTMLInputElement, IProps>(
 		 * is no date selected.
 		 */
 		const handleDotClicked = () => {
+			const currentTime = defaultMonth || new Date();
```
bryceosterhaus, line 560: defaultMonth favored over _now_

ethib137: Can't we just use `initialMonthInternal` here? If not because it's not always getting now correctly, why not use a function instead of `initialMonthInternal`.
```
const getDefaultMonth = useCallback(() => defaultMonth ?? initialMonth ?? new Date(),[defaultMonth, initialMonth]);
```
This would keep us from duplicating logic and still handle the fallback to the deprecated prop as well.

packages/clay-date-picker/src/index.tsx
```diff
@@ -488,10 +489,12 @@ const DatePicker = React.forwardRef<HTMLInputElement, IProps>(
 
 		const updateDate = useCallback(
 			(value: string) => {
+				const currentDate = getDefaultMonth();
```
ethib137, line 492: Should this also be called currentDateTime?

ethib137 (approved):
Yep, this looks good to me... using a function keeps the logic in one place which looks good to me. Thanks.

packages/clay-date-picker/src/index.tsx
ethib137: Nah, that's fine. I was just checking that it was intentional.

PR #6115, opened 2025-07-24
fix: selecting an option in multi-select should not close the dropdown menu

cgmonte:
Hi team!

One of our customers complained that [dropdowns in Forms' "Select from List" field close after selecting an item](https://liferay.atlassian.net/browse/LPP-59701). The behavior is reproducible with the field setting "Allow Multiple Selections" enabled.

After investigating, I found that `ClayMultiSelect` does not have a `closeOnClick` prop like `ClayDropDown` [does](https://github.com/liferay/clay/blob/master/packages/clay-drop-down/src/DropDown.tsx#L176). This PR implements `closeOnClick` on `ClayMultiSelect`, allowing the user to select multiple options without having to retrigger the dropdown menu.  

Why `false` as default? Because, [according to Lexicon](http://liferay.design/lexicon/core-components/dropdowns/#gatsby-focus-wrapper):

> To close a dropdown, use one of the following methods:
> 
> - Select an option in the panel, if they are mutually exclusive.
> - Click on the button that triggered it.
> - Click outside the panel.

I'm assuming that, because in `MultiSelect` **options are not mutually exclusive**, clicking them should not close the dropdown. Think about it: if the user can select multiple options, it is more convenient if the selection dropdown stays available while the user clicks the items.

This will change the default behavior for every current usage of `ClayMultiSelect` in Liferay Portal. After this change, If there is any case where closing the dropdown after clicking is better, then the caller must pass `closeOnClick={true}`. Some tests might break, but I believe it is a change for the better. If you think we should be more conservative and keep the current behavior as default, let me know and I can update the PR, defaulting to `true`.

Thanks!

### DIFF
--- a/packages/clay-multi-select/src/MultiSelect.tsx
+++ b/packages/clay-multi-select/src/MultiSelect.tsx
@@ -79,6 +79,11 @@ export interface IProps<T extends Record<string, any> = Item>
 	closeButtonAriaLabel?: string;
 
 	/**
+	 * Flag that indicates whether to close DropDown when clicking on the item.
+	 */
+	closeOnClick?: boolean;
+
+	/**
 	 * The initial value of the active state (uncontrolled).
 	 */
 	defaultActive?: boolean;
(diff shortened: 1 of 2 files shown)


ethib137:
Hi @cgmonte , thank you for the PR. The current behavior is intended and at this time changing this behavior would be more effort than it is likely worth. The requirement that the dropdown menu stay open while adding items is a specific behavior that could be considered an improvement in certain cases, but it shouldn’t be considered an issue.
Adding this variation to the pattern would mean we would need to improve the visual feedback and the screen reader feedback. I'll add a comment on the LPP as well. Thank you.

PR #6123, opened 2025-08-08
fix(@clayui/core): LPD-62562 Improve panel transition handling by also taking into account the previous value of panelNext

veroglez:
Hi guys! I'm sending the fix for https://liferay.atlassian.net/browse/LPD-62562. Could your review it? Thanks in advance!

### DIFF
--- a/packages/clay-core/src/vertical-bar/Panel.tsx
+++ b/packages/clay-core/src/vertical-bar/Panel.tsx
@@ -58,27 +58,26 @@ export function Panel({children, keyValue = null, tabIndex}: Props) {
 
 	const nodeRef = useRef<HTMLDivElement | null>(null);
 
-	const previousActivePanelRef = useRef<React.Key | null>(null);
-
 	const {prefersReducedMotion} = useProvider();
 
 	const isFirst = useIsFirstRender();
 
+	const previousActivePanel = usePrevious(activePanel);
+	const previousPanelNext = usePrevious(panelNext);
+
 	const isPanelOpen = () => {
 		if (isFirst) {
 			return activePanel !== null;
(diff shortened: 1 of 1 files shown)


packages/clay-core/src/vertical-bar/Panel.tsx
```diff
@@ -52,32 +52,36 @@ export function Panel({children, keyValue = null, tabIndex}: Props) {
...
 			return activePanel !== null;
-		} else if (previousActivePanelRef.current === activePanel) {
+		}
+
+		if (previousActivePanel === activePanel) {
 			return true;
-		} else if (activePanel === null) {
+		}
+
+		if (activePanel === null) {
```
pat270, line 77: @veroglez I don't think we need this block, since we are returning false by default.

pat270 (changes requested):
Looks mostly good to me! Just need to make a small change, see my comment. Could you also cherry pick https://github.com/liferay/clay/pull/6124/commits/2aec02f95ae2ed7914ef6a60056574beb3416b68 to this pr? I updated the storybook example so we can keep track of it in the future.

packages/clay-core/src/vertical-bar/Panel.tsx
veroglez: Uppss! That slipped past me 😅

veroglez:
> Looks mostly good to me! Just need to make a small change, see my comment. Could you also cherry pick [2aec02f](https://github.com/liferay/clay/commit/2aec02f95ae2ed7914ef6a60056574beb3416b68) to this pr? I updated the storybook example so we can keep track of it in the future.

@pat270 Thanks for the review! I cherry pick the commit you mentioned

pat270 (approved):

PR #6140, opened 2025-08-22
chore: fix how package.json specifies type exports

bryceosterhaus:


packages/clay-alert/package.json
```diff
@@ -7,8 +7,11 @@
 	"main": "lib/cjs/index.js",
 	"module": "lib/esm/index.js",
 	"exports": {
-		"import": "./lib/esm/index.js",
-		"require": "./lib/cjs/index.js"
+		".": {
+			"import": "./lib/esm/index.js",
+			"require": "./lib/cjs/index.js",
+			"types": "./lib/index.d.ts"
```
bryceosterhaus, line 13: This is used in TS 4.7+, whereas before that it relied only on the "types" field at the root. So this way we allow for both

PR #6145, opened 2025-08-26
feat(@clayui/core): LPD-48310 Add tests

veroglez:
Hi team,

I’m sending a PR for this story https://liferay.atlassian.net/browse/LPD-47740, which includes a variant of the SidePanel with drilldown.

https://github.com/user-attachments/assets/7c12a67e-2316-4104-ab81-ae178a61a246

Additionally, I’ve also done a modification of the SidePanel header, as specified in this task https://liferay.atlassian.net/browse/LPD-60257. 

@pat270, I tried to adjust the back button in the header using `btn-outline-borderless btn-secondary ml-n2`. However, the negative margin should be 12 px instead of 8 px, and we currently don’t have a class for that margin. Could you help me with this?

Thanks in advance!

### DIFF
--- a/packages/clay-core/src/side-panel/Header.tsx
+++ b/packages/clay-core/src/side-panel/Header.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+import {ClayButtonWithIcon} from '@clayui/button';
 import Icon from '@clayui/icon';
 import classnames from 'classnames';
 import React from 'react';
@@ -23,9 +24,7 @@ type Props = {
 	/**
 	 * Messages for the Side Panel Header.
 	 */
-	messages?: {
-		closeAriaLabel?: string;
-	};
+	messages?: Messages;
 
 	/**
 	 * Property to make the Header sticky. Absolutely positioned SidePanel's
@@ -33,15 +32,27 @@ type Props = {
 	 * for any fixed or sticky navigation bars on the page.
 	 */
 	sticky?: boolean;
+
+	/**
+	 * Function to execute when the back button is pressed.
+	 */
+	onBack?: () => void;
+};
+
+export type Messages = {
+	backAriaLabel?: string;
+	closeAriaLabel?: string;
 };
 
 export const Header = ({
 	children,
 	className,
 	messages = {
+		backAriaLabel: 'Go back.',
 		closeAriaLabel: 'Close the side panel.',
 	},
 	sticky,
+	onBack,
 }: Props) => {
 	const {onOpenChange} = useSidePanel();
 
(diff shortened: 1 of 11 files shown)


packages/clay-core/src/side-panel/Header.tsx
```diff
@@ -56,6 +66,18 @@ export const Header = ({
 			)}
 		>
 			<div className="autofit-row">
+				{onBack ? (
+					<div className="autofit-col">
+						<button
+							aria-label={messages.backAriaLabel}
+							className="btn-outline-borderless btn-secondary ml-n2"
```
pat270, line 73: @veroglez We can use the class `component-action` to get it to be the right size. We need to add `displayType={null}` to get rid of `btn-primary`. I really wish we didn't default to primary.

pat270 (changes requested):
I can't seem to get the storybook to work. I'm getting this error
storybook-error

veroglez:
> I can't seem to get the storybook to work. I'm getting this error storybook-error

I moved `usePrevious` to another directory so I could use it in more components. It's quite possible that the packages haven't been updated. Can you delete `node_modules` and re-run `yarn`?

packages/clay-core/src/side-panel/Header.tsx
veroglez: Updated! Thanks a lot!

veroglez: On the other hand, I'm waiting for @drakonux confirmation regarding the margins for this task https://liferay.atlassian.net/browse/LPD-60257, so for now we'll leave it on hold.

ethib137 (changes requested):
Looks great @veroglez ! Just a couple questions. Thanks.

packages/clay-core/src/side-panel/SidePanelWithDrilldown.tsx
```diff
@@ -0,0 +1,117 @@
...
+	 * Selected panel key
+	 */
+
+	selectedPanelKey: keyof Panels;
+} & Omit<SidePanelProps, 'children'>;
+
+type Panel = {
+	component: React.ReactNode;
+	headerMessages?: Messages;
+	parent?: string;
```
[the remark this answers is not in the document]
veroglez: Done! I've updated it to `parentKey`

packages/clay-core/src/side-panel/Header.tsx
veroglez: Finally, I’ve moved this task https://liferay.atlassian.net/browse/LPD-60257 to the [next story](https://liferay.atlassian.net/browse/LPD-47742), since adjusting the header paddings requires taking the side panel’s resizable behavior into account.

ethib137:
Thanks @veroglez

PR #6146, opened 2025-08-26
LPD-63668 - Remove unnecessary LG and SM versions

marcoscv-work:
This is coming directly from Bootstrap 4.X but it is not needed in Clay.

If we cannot remove the CSS reference in bootstrap feel free to keep it, but we dont want / need the properties to use those sizes.

There are no uses for these in any Liferay product, and we do not plan to use them as they are not recommended and are disproportionate.


pat270:
@marcoscv-work I think we can deprecate it and put it behind a variable `$enable-pagination-sizes: false;`.

marcoscv-work:
> @marcoscv-work I think we can deprecate it and put it behind a variable `$enable-pagination-sizes: false;`.

@pat270 It sounds good to me. Do you want me to do it? or can you adapt the PR?

pat270:
@marcoscv-work I resent here https://github.com/liferay/clay/pull/6152

PR #6152, opened 2025-09-02
fix(@clayui/css): LPD-63668 Deprecate and remove styles for paginatio…

pat270:
…n-sm and pagination-lg

BREAKING CHANGE: This removes styles for pagination-sm and pagination-lg. If you need this, set `$enable-pagination-sizes: true;`

https://liferay.atlassian.net/browse/LPD-63668


marcoscv-work:
Awesome @pat270 !!
Thanks!!!

PR #6154, opened 2025-09-05
fix(@clayui/multi-select): LPD-64387 Investigate test failures in ClayMultiselect

gabsprates:
Ticket: https://liferay.atlassian.net/browse/LPD-64387

### DIFF
--- a/packages/clay-multi-select/src/MultiSelect.tsx
+++ b/packages/clay-multi-select/src/MultiSelect.tsx
@@ -370,10 +370,18 @@ export const MultiSelect = React.forwardRef(function MultiSelectInner<
 				setActive(false);
 
 				if (
-					items.find(
-						(dropdownItem) =>
-							dropdownItem['value'] === item['value']
-					)
+					items.find((dropdownItem) => {
+						return (
+							getLocatorValue({
+								item: dropdownItem,
+								locator: locator.value,
+							}) ===
+							getLocatorValue({
+								item,
+								locator: locator.value,
+							})
+						);
+					})
 				) {
 					return;
 				}
--- a/packages/clay-multi-select/src/MultiSelect.tsx
+++ b/packages/clay-multi-select/src/MultiSelect.tsx
@@ -433,7 +433,7 @@ export const MultiSelect = React.forwardRef(function MultiSelectInner<
 				</AutocompleteItem>
 			);
 		},
-		[children, locator]
+		[children, locator, items, setItems]
 	);
 
 	return (


pat270:
@gabsprates The test failure is located in `liferay-portal/portal-web/test/functional/com/liferay/portalweb/tests/enduser/frontendinfrastructure/uiinfrastructure/clay/ClayMultiselect.testcase`. The multi-select in the Clay Sample portlet isn't adding a label when clicking on the dropdown item.

multi-select

gabsprates:
Hi @pat270,

> The multi-select in the Clay Sample portlet isn't adding a label when clicking on the dropdown item.

That's correct, and that happens because Clay is running a comparison on a fixed item property (`.value`), instead the one that was provided to be used as `value`, thought the `locator`.

In objects like `{data=1, name=one, label=label, value=value}` and `{data=2, name=two, label=label, value=value}`, where `data` was supposed to be used as the **value**, it is not, it uses the `value` property, leading on comparing `'value' === 'value'`, what is always true. That stops the execution and skips the `setItems([...items, item]);` step.

I can send a PR with the following change, to `liferay-portal`, but Clay would still be wrong.

https://github.com/liferay/liferay-portal/blob/2e1dd3cea588015764bf5a5005a399bd6dc6204a/modules/apps/frontend-taglib/frontend-taglib-clay-sample-web/src/main/java/com/liferay/frontend/taglib/clay/sample/web/internal/display/context/MultiselectDisplayContext.java?plain=1#L112-L133

```diff
liferay-portal/modules/apps/frontend-taglib/frontend-taglib-clay-sample-web/src/main/java/com/liferay/frontend/taglib/clay/sample/web/internal/display/context/MultiselectDisplayContext.java

@@ -112,22 +112,22 @@ public class MultiselectDisplayContext {
                _sourceMultiselectItemsWithCustomProperties =
                        MultiselectItemListBuilder.add(
                                multiselectItem -> {
-                                       multiselectItem.put("data", "1");
+                                       multiselectItem.put("value", "1");
                                        multiselectItem.put("name", "one");
                                }
                        ).add(
                                multiselectItem -> {
-                                       multiselectItem.put("data", "2");
+                                       multiselectItem.put("value", "2");
                                        multiselectItem.put("name", "two");
                                }
                        ).add(
                                multiselectItem -> {
-                                       multiselectItem.put("data", "3");
+                                       multiselectItem.put("value", "3");
                                        multiselectItem.put("name", "three");
                                }
                        ).add(
                                multiselectItem -> {
-                                       multiselectItem.put("data", "4");
+                                       multiselectItem.put("value", "4");
                                        multiselectItem.put("name", "four");
                                }
                        ).build();

```

cc @ethib137

PR #6163, opened 2025-09-15
fix(@clayui/css): LPD-65265 c-slideout unsticky sidebar header and footer in small height screens

pat270:
@veroglez I scoped header / footer changes here. You can cherry-pick the top two commits in your pr.

### DIFF
--- a/packages/clay-core/docs/side-panel.mdx
+++ b/packages/clay-core/docs/side-panel.mdx
@@ -226,8 +226,10 @@ export default function App() {
 									the navigation context.
 								</SidePanel.Body>
 							),
-							headerMessages: {
-								backAriaLabel: 'Back to previous panel',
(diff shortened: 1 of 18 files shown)


pat270:
@veroglez I did see that. It's because header, body, footer is wrapped in `display: contents`. I thought it would pass down background colors from the parent, but I read it wrong. I'll see what I can do. As a last resort, I can set the colors on header and footer.

veroglez:
> @veroglez I did see that. It's because header, body, footer is wrapped in `display: contents`. I thought it would pass down background colors from the parent, but I read it wrong. I'll see what I can do. As a last resort, I can set the colors on header and footer.

@pat270 your changes are working correctly. From my side, we can go ahead and move this forward. Thanks so much for your help!

PR #6164, opened 2025-09-16
LPD-65537 - Update props for clay components that include aria labels that cannot be translated. 

ethib137:
https://liferay.atlassian.net/browse/LPD-65537

The goal is that all of these extra props are optional so that we don't break any existing usages.


ethib137:
@pat270 since this is needed for an LPP, can we start a new clay release to Portal?

PR #6169, opened 2025-09-19
feat: add a primaryAction API for autocomplete

bryceosterhaus:
This new API allows for configuring an action in the autocomplete dropdown that will always appear as the first item in the list and when selecting it will call the onClick callback. This also works in MultiSelect as well

https://github.com/user-attachments/assets/7f16d754-d071-4b5a-b19f-bd3db7d44059

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 const config = {
 	stories: ['../packages/*/stories/*.stories.@(js|jsx|ts|tsx)'],
 	addons: [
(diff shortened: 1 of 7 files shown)


ethib137:
Thanks @bryceosterhaus , this looks really good. Can you update the component so that the create new item never gets highlighted by the auto complete. 

Screenshot 2025-09-22 at 12 57 03 PM

Also, being able to add icons is nice, but was there a requirement for that? If not, it's probably best to leave it out for now.

bryceosterhaus:
@ethib137 I actually now can't find where the icon design I saw was... maybe I am hallucinating

PR #6180, opened 2025-10-03
feat(@clayui/autocomplete): LPD-66883 Adds displayType prop to mimic picker input

pat270:
https://liferay.atlassian.net/browse/LPD-66883

autocomplete

autocomplete-loading

@ethib137 I don't think we need to add `insetItemAfter`. We can use the classes we use for Picker on the input. Let me know if this works.

### DIFF
--- a/packages/clay-autocomplete/src/Autocomplete.tsx
+++ b/packages/clay-autocomplete/src/Autocomplete.tsx
@@ -24,6 +24,7 @@ import {
 	useNavigation,
 	useOverlayPosition,
 } from '@clayui/shared';
+import classNames from 'classnames';
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 
 import {AutocompleteContext} from './Context';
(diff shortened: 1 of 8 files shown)


ethib137:
@pat270 does this also allow us to display the sticker inside the input at the start like the description of the ticket shows? If so, can you make sure to add an example of that?

packages/clay-autocomplete/src/Autocomplete.tsx
```diff
@@ -217,7 +228,9 @@ function AutocompleteInner<T extends Item>(
 		defaultItems,
 		defaultValue,
 		direction = 'bottom',
+		displayType,
 		filterKey,
+		insetItemBefore,
```
ethib137: Hey @pat270 we don't want to add new props to the autocomplete for this. We just need to update how the input can be rendered. The autocomplete and other components like it let us pass our own component to be used as the input by using the `As` prop.

ethib137: Great! So we shouldn’t need any changes to autocomplete… likely only changes to input css and maybe some other utility classes.

packages/clay-form/docs/input-group.mdx
```diff
@@ -167,6 +167,48 @@ export default function App() {
...
+							<ClayInput
+								aria-label="Select a space"
+								className="form-control-select"
+								id="selectASpace"
+								insetBefore
+								placeholder="Select a Space"
+								type="text"
+								value="Space Program"
+							/>
+							<ClayInput.GroupInsetItem before>
```
ethib137: @pat270 Is there a reason why this doesn't come before the input? It seems odd since it's an before insetItem.

